PR #8218, opened 2018-10-10
Update cross-platform-targeting.md

ctyar:
I think it would better if we target the `netstandard2.0` and `net461` in this sample, this way it follows the point made earlier about `consider adding a target for net461 when you're offering a netstandard2.0 target`.


BillWagner:
I looked at the change, and it's fine to me.  @JamesNK what do you think?

JamesNK (approved):
Looks good

BillWagner (approved):

PR #8269, opened 2018-10-11
Fix header level

pkulikov:


rpetrusha (approved):
Thanks, @pkulikov, for noticing the incorrect header level. I'll merge your PR now.

PR #8284, opened 2018-10-12
Update gacutil-exe-gac-tool.md

drewjcooper:
Added a note explaining the "file name is too long" error.

## Summary

Added a note explaining the "file name is too long" error.

### DIFF
--- a/docs/framework/tools/gacutil-exe-gac-tool.md
+++ b/docs/framework/tools/gacutil-exe-gac-tool.md
@@ -89,10 +89,20 @@ myAssembly3,Version=1.1.0.0,Culture=en,PublicKeyToken=874e23ab874e23ab
 ```  
 
 > [!NOTE]
->  Attempting to install an assembly with a name longer than 85 characters (excluding the file extension) will result in the following error:
+>  Attempting to install an assembly with a filename longer than between 79 and 91 characters (excluding the file extension) can result in the following error:
 > ```
 > Failure adding assembly to the cache:   The file name is too long.
 > ```
+> This is because internally Gacutil.exe constructs a path of up to MAX_PATH characters that consists of the following elements:
+> - GAC Root - 34 chars (ie. `C:\Windows\Microsoft.NET\assembly\`)
+> - Architecture - 7 or 9 chars (ie. `GAC_32\`, `GAC_64\`, `GAC_MSIL`)
+> - AssemblyName - Up to 91 chars, depending on the size of the other elements (eg. `System.Xml.Linq\`)
+> - AssemblyInfo - 31 to 48 chars or more consisting of:
+>   - Framework - 5 chars (eg. `v4.0_`)
+>   - AssemblyVersion - 8 to 24 chars (eg. `9.0.1000.0_`)
+>   - AssemblyLanguage - 1 to 8 chars (eg. `de_`, `sr-Cyrl_`)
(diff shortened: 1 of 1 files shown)


rpetrusha (changes requested):
Thanks, @drewjcooper. The maximum length of the assembly name is an important detail to add to the documentation. I'm not certain, however, that 85 characters is the limit. When testing with the .NET Framework 4.7.2 version of GacUtil, I can only reproduce the message with a filename (excluding extension) greater than 89 characters.


@karelz do you know who might know the maximum length of the assembly name supported by GacUtil?

drewjcooper:
@rpetrusha I've done some more testing based on your comment and it's more nuanced than I realized when I proposed the changed.  The max length is influenced by the path segment containing the version/locale/publickey info.

The file path in the GAC is of the form <prefix><assemblyName><versionSegment><assemblyName>.dll.  The longest <prefix> is the MSIL path, ie. "C:\windows\Microsoft.NET\assembly\GAC_MSIL\" for 43 characters.  The minimum length I've seen the <versionSegment> be is of the form "\v4.0_1.0.0.0__89845dcd8080cc91\" at 32 chars.  The path excluding the <assemblyName> is "<prefix><versionSegment>.dll" which is, at minimum, 79 characters.  So for a max path of 260 including a null terminator we're left with (260 - 79 - 1) / 2 = 90 chars for the <assemblyName> piece.

However, the <versionSegment> can be longer that the above.  On my system I can <versionSegment>s like "v4.0_14.0.0.0_sr-Cyrl_89845dcd8080cc91" (40 chars) and "\v4.0_9.0.1000.0__31bf3856ad364e35\" (35 chars).  An assembly with both a long version number and long locale identifier could theoretically have a <versionSegment> at around 48 chars (or more?).  This would reduce the max length for the assembly name by 8 chars to 82 chars.  

So, the absolute ceiling for Assembly Name length seems to me to be 90 chars, But this is dependent on the assembly version and locale information that will also be included in the path.  Given the potential length of that information the max length for Assembly Name could be as low as 80 chars.

karelz:
cc @jkotas @jeffschwMSFT 

I would base the answer based on experiments.

jeffschwMSFT:
I agree with @karelz , I think that the write-up that @drewjcooper has done in the comment is great context for anyone that is concerned about max length of a gac'd assembly.

rpetrusha:
Thanks for your effort in investigating this issue, @drewjcooper. I think that the note could be changed to something like:

> Attempting to install an assembly with a filename longer than between 80 and 90 characters (excluding the file extension) can result in the following error: 
> ....
> 
> This is because internally Gacutil.exe constructs a path of up to MAX_PATH characters that consists of the following elements;
> 
> 

Then list the elements in a bulleted list with approximate ranges.

drewjcooper:
@rpetrusha I like it. Sounds good.

drewjcooper:
I'll update the PR.  I was on vacation last week. Give me a couple of days
to get over my backlog.

Andrew

--------------------------------------
There are 10 types of people in this world, those who understand binary,
and those who don't.

For Mole Problems, Call Avogadro: 602-1023

rpetrusha (approved):

drewjcooper:
No worries.  Thanks for the suggestions for improvement.

Andrew

--------------------------------------
There are 10 types of people in this world, those who understand binary,
and those who don't.

For Mole Problems, Call Avogadro: 602-1023

PR #8265, opened 2018-10-11
Add instructions for linking to the dependent repositories

BillWagner:
The dependent repositories for the language specification requires special handling for the links to resolve correctly by the build system.

See https://github.com/dotnet/docs/pull/8221#issuecomment-428943541 for details.

/cc @Mackiovello

PR #8298, opened 2018-10-12
Respond to later feedback on links

BillWagner:
This addresses suggestions on #8265

/cc @pkulikov

### DIFF
--- a/styleguide/template.md
+++ b/styleguide/template.md
@@ -102,16 +102,18 @@ Use for inline code, language keywords, NuGet package names, command-line comman
 To link to a header in the same Markdown file (also known as anchor links), you'll need to find out the id of the header you're trying to link to. To confirm the ID, view the source of the rendered article, find the id of the header (for example, `id="blockquote"`), and link using # + id (for example, `#blockquote`).
 The id is auto-generated based on the header text. So, for example, given a unique section named `## Step 2`, the id would look like this `id="step-2"`.
 
-- Example: `[Chapter 1](#chapter-1)` produces [Chapter 1](#chapter-1).
+- Example: `[Guidance on file names](#file-name)` produces [Guidance on file names](#file-name).
 
 To link to a Markdown file in the same repo, use [relative links](https://www.w3.org/TR/WD-html40-970917/htmlweb.html#h-5.1.2), including the ".md" at the end of the filename.
 
-- Example: `[Readme file](../readme.md)` produces [Readme file](../readme.md).
+- Example: `[Readme file](../README.md)` produces [Readme file](../README.md). (Note that links are case-sensitive.)
 - Example: `[Welcome to .NET](../docs/welcome.md)` produces [Welcome to .NET](../docs/welcome.md).
 
 To link to a header in a Markdown file in the same repo, use relative linking + hashtag linking.
 
 - Example: `[.NET Community](../docs/welcome.md#community)` produces [.NET Community](../docs/welcome.md#community).
+
+In most cases, we use the relative links and discourage the use of `~/` in links because relative links resolve in the source on GitHub. However, whenever we link to a file in a dependent repo, we'll use the `~/` character to provide the path. Because the files in the dependent repo are in a different location in GitHub the links won't resolve correctly with relative links regardless of how they were written.
 
 The C# language specification and the Visual Basic language specification are included in the .NET docs by including the source from the language repositories. The markdown sources are managed in the [csharplang](https://github.com/dotnet/csharplang) and [visual basic](https://github.com/dotnet/vblang) repositories.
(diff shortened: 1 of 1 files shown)


pkulikov:
Cool! Thanks for update @BillWagner

styleguide/template.md
```diff
@@ -102,32 +102,34 @@ Use for inline code, language keywords, NuGet package names, command-line comman
 To link to a header in the same Markdown file (also known as anchor links), you'll need to find out the id of the header you're trying to link to. To confirm the ID, view the source of the rendered article, find the id of the header (for example, `id="blockquote"`), and link using # + id (for example, `#blockquote`).
 The id is auto-generated based on the header text. So, for example, given a unique section named `## Step 2`, the id would look like this `id="step-2"`.
 
-- Example: [Chapter 1](#chapter-1)
+- Example: `[Chapter 1](#chapter-1)` produces [Chapter 1](#chapter-1).
```
rpetrusha: This link should be a valid one (have a real target in this document, like `file-names`).

styleguide/template.md
```diff
@@ -102,32 +102,34 @@ Use for inline code, language keywords, NuGet package names, command-line comman
...
 
 To link to a Markdown file in the same repo, use [relative links](https://www.w3.org/TR/WD-html40-970917/htmlweb.html#h-5.1.2), including the ".md" at the end of the filename.
 
-- Example: [Readme file](../readme.md)
-- Example: [Welcome to .NET](../docs/welcome.md)
+- Example: `[Readme file](../readme.md)` produces [Readme file](../readme.md).
```
rpetrusha: The link is broken -- links are case-sensitive, which probably should be added to this section. In this case, it's README.md.

styleguide/template.md
```diff
@@ -102,32 +102,34 @@ Use for inline code, language keywords, NuGet package names, command-line comman
...
+- Example: `[Welcome to .NET](../docs/welcome.md)` produces [Welcome to .NET](../docs/welcome.md).
 
 To link to a header in a Markdown file in the same repo, use relative linking + hashtag linking.
 
-- Example: [.NET Community](../docs/welcome.md#community)
+- Example: `[.NET Community](../docs/welcome.md#community)` produces [.NET Community](../docs/welcome.md#community).
```
rpetrusha: There is no community anchor. Perhaps open-source?

rpetrusha: The relative links can be frustrating and error-prone for a deeply nested topic. (e.g., `../../../../standard/something.md`). I'm wondering if we should add the `~` for the root of the repo? (like `~/docs/standard/index.md`). The problem, though, is that since it's an extension, the link will 404, so that may be a bad idea.

styleguide/template.md
```diff
@@ -102,32 +102,34 @@ Use for inline code, language keywords, NuGet package names, command-line comman
...
 
-- Example: [GitHub](http://www.github.com)
+- Example: `[GitHub](https://www.github.com)` produces [GitHub](https://www.github.com).
 
 If a URL appears in a Markdown file, it will be transformed into a clickable link.
 
-- Example: <http://www.github.com>
+- Example: `<https://www.github.com>` produces <https://www.github.com>.
+
+Whenever possible, use `https` for all external links.
```
rpetrusha: I'm not sure whether this is the best guidance. https to an http site fails. Perhaps recommend that the URL accurately reflect the protocol (http or https) used by the target?

rpetrusha (changes requested):
Adding examples using fenced text and the actual links really helps here, @BillWagner. I do have a number of comments and suggestions, though.

styleguide/template.md
BillWagner: I added language here based on the practices we **usually** follow. I don't know that we have a clear rule here.

/cc @mairaw if she wants to add her thoughts.

styleguide/template.md
BillWagner: I re-worded to provide a preference for https, whenever both protocols are supported.

styleguide/template.md
rpetrusha: Did you miss this comment, @BillWagner? It would be good for this anchor (as well as the anchor on line 105) to link to a real header in the document.  A good choice for the first, since it links to a multi-word header, is \#inline-code-blocks-with-language-identifier. For welcome.md, \#open-source.

rpetrusha (approved):

rpetrusha:
I've left one additional comment for you to consider, @BillWagner. You can merge when you're ready.

styleguide/template.md
BillWagner: I did miss that. Just updated.

PR #8165, opened 2018-10-08
Ported changes from similar article to this article.

douglaslMS:
## Summary

Porting the changes made describing connection string syntax in PR #7972 to fix issue #7628 to this second, similar article about connection strings.

Fixes #7646

PR #8328, opened 2018-10-14
Improve description of "Provider Connection String"

ant-222:
The current version advises to surround the value of `Provider Connection String` with double quotation marks, which is not generally correct. I have replaced this advise with a reference to the general connection-string syntax, but have kept the example, explaining why it this *specific* case double quotation marks work.

## Summary

Fulfills this [request](https://github.com/dotnet/docs/pull/8165#issuecomment-428669401)


douglaslMS:
@ant-222 Just for the sake of discussion - isn't a connection string always going to contain a semicolon, because it will always (I think) contain more than one keyword/value pair, and therefore - in fact if not in principle - always need to be surrounded with double quotes?

ant-222:
@douglaslMS A connection string to a file-based data source may easily contain a single parameter—the name of the file. Nothing forbids an empty connection string to some global, default, or predefined data source (such a as the table with environment variables). If a connection string does contain more than a single parameter separated by a semicolon, the double quotation mark is not *the only* possible enclosing character: the single quotation mark is another option. Furthermore, the latter seems more useful for *EF* because the internal connection string is more likely to use double quotation marks.

douglaslMS:
@ant-222 Thanks for the clarification. I just want to make sure that we don't obscure the most common use case.

ant-222:
@douglaslMS The example in my version still reflects the common use case:

    `Provider Connection String ="Server=serverName; User ID = userID";`

douglaslMS (approved):

douglaslMS:
@ant-222 Thank you for your revisions. The updated page should be published by tomorrow.

ant-222:
@douglaslMS Thanks to you, Douglas. Please, check the syntax of my reference—I couldn't test it in the github preview.

PR #8353, opened 2018-10-15
chore: Remove en-us from docs.microsoft.com links

nschonni:
The docs styleguide says that these links should be language neutral. These mostly appear internal to the docset, so they could also be done as relative links


rpetrusha:
Build is stuck; closing and reopening to begin new build.

rpetrusha (approved):
Thanks so much, @nschonni, for removing "en-us" from links for docs.microsoft.com.  As you point out, many of these could be relative links; I'll open a separate issue to convert them. I'll also merge your PR now. Your changes should be live on docs.microsoft.com in the next day or two.

Issue #6507, opened 2018-07-17
Eliminate cast or use `var`


johnnyagerard:
Those lines contain a useless cast, we could replace them:
IEnglishDimensions eDimensions = (IEnglishDimensions)box1;
IMetricDimensions mDimensions = (IMetricDimensions)box1;

IEnglishDimensions eDimensions = box1;
IMetricDimensions mDimensions = box1;

However, if we would use the var keyword, the cast would be useful:
var eDimensions = (IEnglishDimensions)box1;
var mDimensions = (IMetricDimensions)box1;

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 594cb1b7-c9c9-5f33-fca7-2f1872c095a9
* Version Independent ID: 2ed35b23-7ffd-0d95-a2f0-02e3ccc4f8a7
* Content: [How to: Explicitly Implement Members of Two Interfaces (C# Programming Guide)](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/how-to-explicitly-implement-members-of-two-interfaces)
* Content Source: [docs/csharp/programming-guide/interfaces/how-to-explicitly-implement-members-of-two-interfaces.md](https://github.com/dotnet/docs/blob/master/docs/csharp/programming-guide/interfaces/how-to-explicitly-implement-members-of-two-interfaces.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

aditimantri2196:
@johnnyagerard You are right, this will work without explicit casting, but it seems to had been added for more readability to readers of microsoft docs. Adding an explicit casting won't harm.

johnnyagerard:
Yes, you're right. Thanks.

BillWagner:
You are both correct: the cast works, but is unnecessary.  Because the sample does use an explicit variable declaration, I'd prefer that the unnecessary cast was removed.

That would clearly show readers that the `Box` type can be implicitly converted to either interface that it implements. 

I've added this to our backlog, and added the 'up for grabs' label.

PR #8375, opened 2018-10-15
Remove unnecessary cast syntax.

adyavanapalli:
## Summary

The cast syntax is unnecessary here, which is consistent with changes introduced in #6507.


rpetrusha (approved):
Thanks for removing the unnecessary casts, @adyavanapalli. We'll merge your PR now. Your changes should be live on docs.microsoft.com in the next day or two.

PR #8376, opened 2018-10-15
Removed duplicate text from child topic, linked to source.

douglaslMS:
## Summary

The section about connection string syntax was previously duplicated in two places in the ADO.NET docs, with the risk of not matching and of being incomplete or out of sync. I removed the duplicate section from this child topic in the EF docs and pointed to the section in the main Connecting Strings article at the parent ADO.NET level.


rpetrusha (approved):
I left one comment for you to consider, @douglaslMS. Once you decide whether to address it, you can merge.

PR #8390, opened 2018-10-16
Include note about not changing the strong name key

JamesNK:
Include recommendation to not change the SN key also on the strong naming page.

Some devs may not visit the breaking changes page and will be unaware.


mairaw (approved):
Thanks @JamesNK. merging now.

PR #8398, opened 2018-10-16
Fix XML comment

andyedinborough:
## Summary

The opening tag of the XML comment was not correct.


BillWagner (approved):
Thanks for spotting this and fixing it @andyedinborough We appreciate it.

I've reviewed the changes, and I'll :shipit: now. You should see the changes in the next day or so on our regular publish cycle.

Thanks again for contributing to docs. 🎆

PR #8401, opened 2018-10-16
Asynchronous programming patterns: removed duplication

pkulikov:
The list at the beginning of the async programming patterns overview:
https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/

duplicates the table at the end of the article. Thus, the table can be removed.

Also reordered listing of patterns, such that it matches TOC and the recommend TAP goes first.


rpetrusha (approved):
Thanks, @pkulikov. I like the way that you've streamlined and consolidated this topic. I'll merge your PR now.

PR #8449, opened 2018-10-17
Revised the Numerics in .NET article

pkulikov:
This PR updates the [Numerics in .NET](https://docs.microsoft.com/en-us/dotnet/standard/numerics) article:
- When talking about floating-point types, it's more interesting to know the precision, than min/max value. Thus, replace the floating-point types table with the one from the C# reference: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/floating-point-types-table

- Fixed typos in the section about SIMD-enabled types and mentioned that only .NET Framework needs a NuGet package for `Vector<T>`.

- As most probably, this article aims at those who wants to do some calculations, I've shortened paragraphs that describe the set of available operations for integral and floating-point types.

### DIFF
--- a/docs/csharp/language-reference/keywords/double.md
+++ b/docs/csharp/language-reference/keywords/double.md
@@ -14,7 +14,7 @@ The `double` keyword signifies a simple type that stores 64-bit floating-point v
 
 |Type|Approximate range|Precision|.NET type|
 |----------|-----------------------|---------------|-------------------------|
-|`double`|±5.0 × 10<sup>−324</sup> to ±1.7 × 10<sup>308</sup>|15-16 digits|<xref:System.Double?displayProperty=nameWithType>|
+|`double`|±5.0 × 10<sup>−324</sup> to ±1.7 × 10<sup>308</sup>|15-17 digits|<xref:System.Double?displayProperty=nameWithType>|
 
 ## Literals
 
--- a/docs/csharp/language-reference/keywords/double.md
+++ b/docs/csharp/language-reference/keywords/double.md
@@ -14,7 +14,7 @@ The `double` keyword signifies a simple type that stores 64-bit floating-point v
 
 |Type|Approximate range|Precision|.NET type|
 |----------|-----------------------|---------------|-------------------------|
-|`double`|±5.0 × 10<sup>−324</sup> to ±1.7 × 10<sup>308</sup>|15-17 digits|<xref:System.Double?displayProperty=nameWithType>|
+|`double`|±5.0 × 10<sup>−324</sup> to ±1.7 × 10<sup>308</sup>|~15-17 digits|<xref:System.Double?displayProperty=nameWithType>|
 
 ## Literals
 
--- a/docs/standard/numerics.md
+++ b/docs/standard/numerics.md
@@ -1,6 +1,6 @@
 ---
 title: "Numerics in .NET"
-ms.date: "10/12/2018"
+ms.date: "10/17/2018"
 ms.technology: dotnet-standard
 helpviewer_keywords: 
   - "SIMD"
@@ -16,11 +16,11 @@ ms.author: "ronpet"
 ---
 # Numerics in .NET
 
-.NET provides the standard numeric integral and floating-point primitives, as well as <xref:System.Numerics.BigInteger?displayProperty=nameWithType>, an integral type with no theoretical upper or lower bound, <xref:System.Numerics.Complex?displayProperty=nameWithType>, a type that represents complex numbers, and a set of SIMD-enabled vector types in the <xref:System.Numerics> namespace.
+.NET provides the standard numeric integral and floating-point primitives, as well as <xref:System.Numerics.BigInteger?displayProperty=nameWithType>, which is an integral type with no theoretical upper or lower bound, <xref:System.Numerics.Complex?displayProperty=nameWithType>, which represents complex numbers, and a set of SIMD-enabled vector types in the <xref:System.Numerics> namespace.
   
 ## Integral types
 
-.NET supports both signed and unsigned integers ranging from one byte to eight bytes in size. The following table lists the integral types and their size, indicates whether they are signed or unsigned, and documents their range. All integers are value types.  
+.NET supports both signed and unsigned integers ranging from one byte to eight bytes in size. The following table lists .NET integer types:
   
 |Type|Signed/Unsigned|Size (in bytes)|Minimum value|Maximum value|  
 |----------|----------------------|--------------------|-------------------|-------------------|  
@@ -33,13 +33,20 @@ ms.author: "ronpet"
 |<xref:System.UInt32?displayProperty=nameWithType>|Unsigned|4|0|4,294,967,295|  
 |<xref:System.UInt64?displayProperty=nameWithType>|Unsigned|8|0|18,446,744,073,709,551,615|  
   
- Each integral type supports a standard set of arithmetic, comparison, equality, explicit conversion, and implicit conversion operators. Each integer also includes methods to perform equality comparisons and relative comparisons, to convert the string representation of a number to that integer, and to convert an integer to its string representation. Some additional mathematical operations beyond those handled by the standard operators, such as rounding and identifying the smaller or larger value of two integers, are available from the <xref:System.Math?displayProperty=nameWithType> class. You can also work with the individual bits in an integer value by using the <xref:System.BitConverter?displayProperty=nameWithType> class.  
-  
- Note that the unsigned integral types are not CLS-compliant. For more information, see [Language Independence and Language-Independent Components](language-independence-and-language-independent-components.md).  
+Each integral type supports standard arithmetic operators. The <xref:System.Math?displayProperty=nameWithType> class provides methods for a broader set of mathematical functions.
+
+You can also work with the individual bits in an integer value by using the <xref:System.BitConverter?displayProperty=nameWithType> class.  
+
+> [!NOTE]  
+> The unsigned integral types are not CLS-compliant. For more information, see [Language Independence and Language-Independent Components](language-independence-and-language-independent-components.md).
+
+## BigInteger
+
+The <xref:System.Numerics.BigInteger?displayProperty=nameWithType> structure is an immutable type that represents an arbitrarily large integer whose value in theory has no upper or lower bounds. The methods of the <xref:System.Numerics.BigInteger> type closely parallel those of the other integral types.
   
 ## Floating-point types
 
-.NET includes three primitive floating-point types, which are listed in the following table.  
+.NET includes three primitive floating-point types, which are listed in the following table:
   
 |Type|Size (in bytes)|Approximate range|Precision|  
 |----------|--------|---------------------|--------------------|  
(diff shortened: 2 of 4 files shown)


docs/standard/numerics.md
```diff
@@ -33,44 +33,58 @@ The .NET Framework supports the standard numeric integral and floating-point pri
...
+
+- The <xref:System.Numerics.Plane> type, which represents a plane in three-dimensional space.
+
+- The <xref:System.Numerics.Quaternion> type, which represents a vector that is used to encode three-dimensional physical rotations.
+
+- The <xref:System.Numerics.Vector%601> type, which represents a vector of a specified numeric type and provides a broad set of operators that benefit from SIMD support.
+  > [!NOTE]
+  > The <xref:System.Numerics.Vector%601> type is not included into the .NET Framework. You must install the [System.Numerics.Vectors](https://www.nuget.org/packages/System.Numerics.Vectors) NuGet package to get access to this type.
   
+The SIMD-enabled vector types are implemented in IL, which allows them to be used on non-SIMD-enabled hardware and JIT compilers. To take advantage of SIMD instructions, your 64-bit apps must be compiled by the new 64-bit JIT Compiler for managed code, which is included with the .NET Framework 4.6; it adds SIMD support when targeting x64 processors.
```
pkulikov: No changes in this paragraph. However, I wonder if it's still accurate?

rpetrusha: included with the .NET Framework Framework 4.6 --> included in .NET Core and in .NET Framework 4.6 and later versions. "it adds SIMD support when targeting x64 processors" can be a separate sentence -- though I'm not sure whether or not it should be 64-bit instead of x64.

rpetrusha (changes requested):

rpetrusha (changes requested):
This looks really good, @pkulikov. I thought the Ryujit compiler section could be revised slightly -- but I'm also not completely certain about it. @jkotas, could you review the paragraph on the 64-bit compiler, please?

jkotas:
cc @tannergooding

jkotas:
This looks good to me, modulo Ron's existing comments.

docs/standard/numerics.md
pkulikov: @rpetrusha @jkotas the following change would be correct: "the new 64-bit JIT Compiler for managed code" --> "RyuJIT compiler"?

jkotas: Yes.

docs/standard/numerics.md
```diff
@@ -14,15 +14,15 @@ ms.assetid: dfebc18e-acde-4510-9fa7-9a0f4aa3bd11
 author: "rpetrusha"
 ms.author: "ronpet"
 ---
-# Numerics in the .NET Framework
-The .NET Framework supports the standard numeric integral and floating-point primitives, as well as <xref:System.Numerics.BigInteger>, an integral type with no theoretical upper or lower bound, <xref:System.Numerics.Complex>, a type that represents complex numbers, and a set of SIMD-enabled vector types in the <xref:System.Numerics> namespace.  
-  
- In addition, System.Numerics.Vectors, the SIMD-enabled library of vectory types, was released as a NuGet package.  
+# Numerics in .NET
+
+.NET provides the standard numeric integral and floating-point primitives, as well as <xref:System.Numerics.BigInteger?displayProperty=nameWithType>, which is an integral type with no theoretical upper or lower bound, <xref:System.Numerics.Complex?displayProperty=nameWithType>, which represents complex numbers, and a set of SIMD-enabled vector types in the <xref:System.Numerics> namespace.
```
tannergooding: Is just saying "standard" sufficient here? Many modern systems count `int128` (for example) as "standard", because it has special handling at the ABI layer.

docs/standard/numerics.md
```diff
@@ -33,44 +33,58 @@ The .NET Framework supports the standard numeric integral and floating-point pri
...
   
--   <xref:System.Numerics.Vector2>,  <xref:System.Numerics.Vector3>, and  <xref:System.Numerics.Vector4> types, which are 2-, 3-, and 4-dimensional vectors of type <xref:System.Single>.  
+Each floating-point type supports standard arithmetic operators. The <xref:System.Math?displayProperty=nameWithType> class provides methods for a broader set of mathematical functions. .NET Core 2.0 and later includes the <xref:System.MathF?displayProperty=nameWithType> class that provides methods which accept arguments of the <xref:System.Single> type.
+
+You can also work with the individual bits in <xref:System.Double> and <xref:System.Single> values by using the <xref:System.BitConverter?displayProperty=nameWithType> class. The <xref:System.Decimal?displayProperty=nameWithType> structure has its own methods, <xref:System.Decimal.GetBits%2A?displayProperty=nameWithType> and <xref:System.Decimal.%23ctor%28System.Int32%5B%5D%29?displayProperty=nameWithType>, for working with a decimal value's individual bits, as well as its own set of methods for performing some additional mathematical operations.
```
tannergooding: Should we have a comment about the methods on `System.Single` and `System.Double` that allow checking for special values or categories (i.e. `IsNaN`, `IsFinite`, etc)

docs/standard/numerics.md
```diff
@@ -33,44 +33,58 @@ The .NET Framework supports the standard numeric integral and floating-point pri
...
   
--   Two matrix types, <xref:System.Numerics.Matrix3x2>, which represents a 3x2 matrix; and <xref:System.Numerics.Matrix4x4>, which represents a 4x4 matrix.  
+The <xref:System.Double> and <xref:System.Single> types are intended to be used for values that by their nature are imprecise (such as the distance between two stars) and for applications in which a high degree of precision and small rounding error is not required. You should use the <xref:System.Decimal?displayProperty=nameWithType> type for cases in which greater precision is required and rounding error is undesirable.
```
tannergooding: I've encountered a few people that think `System.Decimal` completely removes rounding error. I wonder if we could clarify that its internal representation merely helps to reduce rounding error instead (the System.Decimal docs page has some examples)...

docs/standard/numerics.md
```diff
@@ -33,44 +33,58 @@ The .NET Framework supports the standard numeric integral and floating-point pri
...
+
+- The <xref:System.Numerics.Vector2>, <xref:System.Numerics.Vector3>, and <xref:System.Numerics.Vector4> types, which represent vectors with 2, 3, and 4 <xref:System.Single> values.
+
+- Two matrix types, <xref:System.Numerics.Matrix3x2>, which represents a 3x2 matrix, and <xref:System.Numerics.Matrix4x4>, which represents a 4x4 matrix.
+
+- The <xref:System.Numerics.Plane> type, which represents a plane in three-dimensional space.
+
+- The <xref:System.Numerics.Quaternion> type, which represents a vector that is used to encode three-dimensional physical rotations.
+
+- The <xref:System.Numerics.Vector%601> type, which represents a vector of a specified numeric type and provides a broad set of operators that benefit from SIMD support.
```
tannergooding: Is it worth noting that `Vector<T>` can be variable sized, depending on the support provided by the underlying hardware?

docs/standard/numerics.md
rpetrusha: possibly instead of standard, we should say 8-bit, 16-bit, 32-bit, and 64-bit

docs/standard/numerics.md
rpetrusha: Yes. And the last sentence ("rounding error is undesirable") implies there is none. Better is something like "and rounding error must be minimized".

docs/standard/numerics.md
pkulikov: I would replace "standard" with "range of". Actual integer sizes are mentioned in the **Integral types** section.

pkulikov:
@tannergooding @jkotas @rpetrusha thanks a lot for your feedback. I've addressed it, please review the latest commit.

@tannergooding @jkotas please check if I've updated the section about SIMD-enabled types correctly.

docs/csharp/language-reference/keywords/double.md
```diff
@@ -14,7 +14,7 @@ The `double` keyword signifies a simple type that stores 64-bit floating-point v
 
 |Type|Approximate range|Precision|.NET type|
 |----------|-----------------------|---------------|-------------------------|
-|`double`|±5.0 × 10<sup>−324</sup> to ±1.7 × 10<sup>308</sup>|15-16 digits|<xref:System.Double?displayProperty=nameWithType>|
```
tannergooding: I think this should be explicit that it is approx 15-17 digits (same for float on 6-9 digits). There may be some values that aren't "exact" even for 15 digit inputs (depending on the exact input, its magnitude, etc).

docs/standard/numerics.md
```diff
@@ -14,15 +14,15 @@ ms.assetid: dfebc18e-acde-4510-9fa7-9a0f4aa3bd11
 author: "rpetrusha"
 ms.author: "ronpet"
 ---
-# Numerics in the .NET Framework
-The .NET Framework supports the standard numeric integral and floating-point primitives, as well as <xref:System.Numerics.BigInteger>, an integral type with no theoretical upper or lower bound, <xref:System.Numerics.Complex>, a type that represents complex numbers, and a set of SIMD-enabled vector types in the <xref:System.Numerics> namespace.  
-  
- In addition, System.Numerics.Vectors, the SIMD-enabled library of vectory types, was released as a NuGet package.  
+# Numerics in .NET
+
+.NET provides the range of numeric integer and floating-point primitives, as well as <xref:System.Numerics.BigInteger?displayProperty=nameWithType>, which is an integral type with no theoretical upper or lower bound, <xref:System.Numerics.Complex?displayProperty=nameWithType>, which represents complex numbers, and a set of SIMD-enabled types in the <xref:System.Numerics> namespace.
```
tannergooding: > provides the range

Should this be `provides a range`?

docs/standard/numerics.md
```diff
@@ -33,44 +33,63 @@ The .NET Framework supports the standard numeric integral and floating-point pri
...
-  
-## BigInteger  
- <xref:System.Numerics.BigInteger?displayProperty=nameWithType> is an immutable type that represents an arbitrarily large integer whose value in theory has no upper or lower bounds. The methods of the <xref:System.Numerics.BigInteger> type closely parallel those of the other integral types.  
-  
-## Complex  
- The <xref:System.Numerics.Complex> type represents a complex number, that is, a number with a real number part and an imaginary number part. It supports a standard set of arithmetic, comparison, equality, explicit conversion, and implicit conversion operators, as well as mathematical, algebraic, and trigonometric methods.  
+Each integer type supports standard arithmetic operators. The <xref:System.Math?displayProperty=nameWithType> class provides methods for a broader set of mathematical functions.
```
tannergooding: I think "a set of standard  arithmetic operators" reads better. Thoughts?

docs/standard/numerics.md
```diff
@@ -33,44 +33,63 @@ The .NET Framework supports the standard numeric integral and floating-point pri
...
+|<xref:System.Double?displayProperty=nameWithType>|8|±5.0 × 10<sup>−324</sup> to ±1.7 × 10<sup>308</sup>|15-17 digits|  
+|<xref:System.Decimal?displayProperty=nameWithType>|16|±1.0 x 10<sup>-28</sup> to ±7.9228 x 10<sup>28</sup>|28-29 digits|  
   
--   <xref:System.Numerics.Vector2>,  <xref:System.Numerics.Vector3>, and  <xref:System.Numerics.Vector4> types, which are 2-, 3-, and 4-dimensional vectors of type <xref:System.Single>.  
+Both <xref:System.Single> and <xref:System.Double> types support special values that represent non-a-number and infinity. For example, the <xref:System.Double> type provides the following values: <xref:System.Double.NaN?displayProperty=nameWithType>, <xref:System.Double.NegativeInfinity?displayProperty=nameWithType>, and <xref:System.Double.PositiveInfinity?displayProperty=nameWithType>. You use the <xref:System.Double.IsNaN%2A?displayProperty=nameWithType>, <xref:System.Double.IsInfinity%2A?displayProperty=nameWithType>, <xref:System.Double.IsPositiveInfinity%2A?displayProperty=nameWithType>, and <xref:System.Double.IsNegativeInfinity%2A?displayProperty=nameWithType> methods to test for these special values.
```
tannergooding: `non-a-number` should be `not-a-number`

tannergooding (approved):
A couple of nits/questions. Otherwise, LGTM

docs/standard/numerics.md
pkulikov: @rpetrusha please tell which variant is correct

rpetrusha: "provides a range" is correct, @pkulikov.

pkulikov:
@tannergooding thanks for another check. I've addressed your feedback.

rpetrusha (approved):

rpetrusha:
This looks really good, @pkulikov. I'll merge your PR now.

PR #8468, opened 2018-10-18
Missing call to base serialization constructor

csrowell:
Without the call to the base class constructor, the class does not deserialize correctly.

## Summary

For example, if you serialize and deserialize a "`new InvalidDepartmentException("Department '123' is invalid")`" object, the Message property will say "Exception of type 'MyApp.InvalidDepartmentException' was thrown." instead of "Department '123' is invalid". Other properties, such as the Source and StackTrace, will just be initialized to their default values.


BillWagner (approved):
Thanks for spotting this and fixing it @csrowell We appreciate it.  

This is a good change, and I'll :shipit: now.

Thanks again.

PR #8470, opened 2018-10-18
Updating the New Project text and image.

sayedihashimi:
## Summary

The text and images showed creating a standard library project from the Multiplatform node.
It's better to point users to the .NET Core node instead because:
 - It is simpler (less templates shown)
 - It's closer to other project templates they are likely to use for this solution


mairaw (approved):
Thanks @sayedihashimi!

PR #8479, opened 2018-10-18
Moved Northwind first, fixed installation instructions.

douglaslMS:
## Summary

Based on usage metrics, I moved the para about Northwind higher than the para about AdventureWorks. I also removed one of the installation methods which does not in fact work.


rpetrusha (approved):
LGTM, @douglaslMS. You can merge when you're ready.

PR #8503, opened 2018-10-19
Remainder assignment operator: definition first

pkulikov:
Nit pick: first say what operator does and only then that it's supported by numeric types.


rpetrusha (approved):
This is a good change, @pkulikov. I'll merge your PR now.

PR #8523, opened 2018-10-20
Improved async.md Cancellation Token example.

isaacabraham:
Improve Cancellation Token sample.

## Summary

Having reviewed the Cancellation Token source example, I'm not sure it is currently a particularly good example:

1. It doesn't do anything F# asynchronous inside it - it actually calls the "original" Async web client upload async, not the F# async version or even the C# Async / Await version. This is confusing, as it doesn't block and can't be let bound.
2. The workflow returned immediately - there was no opportunity to see cancellation working in action.

I've replaced it with a simpler example that simpler prints to the console every 1 second. The cancellation token can be called, which will stop the workflow. This works very well from within a script.


cartermp (approved):
Thanks @isaacabraham this is much better.

Thraka:
@isaacabraham Hi Isaac! Looks like something came through and caused a merge conflict. You'll need to fix this before we can merge. Thanks.

isaacabraham:
Oops - will do.

isaacabraham:
@mairaw fixed. @cartermp it should be ok now - it uses the second (simplified) version.

PR #8576, opened 2018-10-23
Add example to unwrap DU in function parameter

theprash:
I've added an example that shows a more common and less verbose way to unwrap single-case DUs.


cartermp (approved):
Sorry for taking so long to get to this @theprash, thanks so much.

PR #8588, opened 2018-10-24
Remove version number

terrajobst:
Let's avoid using specific version numbers here as it means this topic will continuously get outdated by shipping newer versions.

Also, none of the mentioned version numbers add anything really.

/cc @mairaw @BethMassi


mairaw (approved):
LGTM

PR #8608, opened 2018-10-24
Noted that fibers are not supported for p/invoke

rpetrusha:
## Noted that fibers are not supported for p/invoke

Given the fact that interop documentation has been scaled back, with much of its content archived, it seemed to me that [Consuming Unmanaged DLL Functions](https://docs.microsoft.com/dotnet/framework/interop/consuming-unmanaged-dll-functions) is the best place to surface the fact that fibers are not supported.

//cc @jkotas


pkulikov:
@rpetrusha is the following topic worth the same update?
https://docs.microsoft.com/en-us/dotnet/standard/threading/managed-and-unmanaged-threading-in-windows

jkotas:
I think https://docs.microsoft.com/en-us/dotnet/standard/threading/managed-and-unmanaged-threading-in-windows would be more appropriate doc to talk about this. @pkulikov Thanks for the hint!

I do not think this needs to be mentioned in Consuming Unmanaged DLL Functions

rpetrusha:
Great suggestion, @pkulikov. I agree, @jkotas. I'll revise this PR to modify the managed and unmanaged threading in Windows article instead.

jkotas (approved):

PR #8610, opened 2018-10-24
Update write-safe-efficient-code.md

pkulikov:
This PR fixes the following compile-time error:
>CS8057	Block bodies and expression bodies cannot both be provided


BillWagner (approved):
Thanks for these updates @pkulikov 

I'll :shipit: now.

PR #8613, opened 2018-10-24
List alternate alias for Razor Pages 2.1 project template

scottaddie:
As of .NET Core SDK 2.1.300 (2.1 RTM), there are 2 valid aliases for using the ASP.NET Core 2.1 Razor Pages project template: `razor` and `webapp`. The CLI only lists the first alias in the array, due to a limitation in its current design. Adding the other alias here to raise awareness. Related discussion: https://github.com/aspnet/Docs/pull/9159#issuecomment-432518051.


mairaw (approved):
Good to know @scottaddie. Thanks! Do you also wanna change ms.date here?

PR #8614, opened 2018-10-24
ref.md: clarify ref readonly return

pkulikov:
Updated the **Ref readonly locals** section, because one cannot use `return readonly ref`. Readonliness of ref is defined by the signature:
```csharp
public ref readonly Point3D Origin() { return ref origin; }
```

Also made updates to follow the "For more information, see..." guidelines.


BillWagner (approved):
Thanks for those important updates @pkulikov 

I'll :shipit: now.  

Thanks again for working with us so often in docs.

Issue #8386, opened 2018-10-15
Dead link


ChrisMaddock:
https://www.myget.org/opensource no longer exists.

---
#### Document details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1e51d820-1f85-192c-9169-cd965581b4f1
* Version Independent ID: 9e7ea8c7-ae47-eda8-4ba0-d5f59a3ade3c
* Content: [Publishing a NuGet package](https://docs.microsoft.com/en-gb/dotnet/standard/library-guidance/publish-nuget-package)
* Content Source: [docs/standard/library-guidance/publish-nuget-package.md](https://github.com/dotnet/docs/blob/master/docs/standard/library-guidance/publish-nuget-package.md)
* Product: **dotnet**
* GitHub Login: @JamesNK
* Microsoft Alias: **mairaw**

JamesNK:
Hmmm, MyGet may no longer offer a plan for open source projects.

tbolon:
It does, but it is not displayed anymore on their website, see comments on the [acquisition announcement](https://blog.myget.org/post/2018/08/31/assembla-acquisition.html).

ChrisMaddock:
Good find @tbolon. It would be great if there was a more authoritative source though.

I wonder if @maartenba or @xavierdecoster could advise?

laithassembla:
Hi everyone, we certainly do still support the open source community (we just removed the page from the site). Our free plan includes the same features as the old open source plan. You can create and share public feeds. You can simply sign up for a free trial here. The trial includes all paid features for 14 days then automatically reverts to a free (open source) plan after 14 days. No credit card or payment required. Hope this helps. Thank you! https://www.myget.org/Account/Register?ReturnUrl=%2FSubscription%2FSignUp%3Fsubscription_type%3Dtrial

JamesNK:
Good news! We'll need to remove the dead link and reword the article in our docs.

Thraka:
@laithassembla @JamesNK @giotavekiari 

I think the myget website should make this clear, not our docs here. This will just cause confusion because we'll be saying there is something free behind a subscription model that really could change at any moment (as it did in this case of removing the page from the myget site) 

We should just simply state that "MyGet is available for public and private feeds." We shouldn't advertise more than what is actually printed on the website.

ChrisMaddock:
Foxed in https://github.com/dotnet/docs/pull/8615 - thanks!

PR #8615, opened 2018-10-24
MyGet Plan&Link Change

giotavekiari:
Url and description of MyGet, changed

Fixes Issue #8386


Thraka (changes requested):
Hi! Thank you very much for this pull request. Please see the discussion in the linked issue #8386 but I'm thinking we should simplify this paragraph:

1. We should probably remove the ! in the first sentence and just use a period.
2. I think we should remove the link entirely (since the link at the start of the bullet point works just fine) and just state that public and private feeds are available.

Let's see what everyone else thinks on the issue though.

giotavekiari:
Actually it's kind of confusing, because they don't state anything about the free plan, even in their pricing page: https://www.myget.org/pricing

Do they really support free plans? How can we redirect people there stating that they do have a free plan if MyGet doesn't even mention it on their site...

Thraka:
@giotavekiari According to the post by someone associated with MyGet (in the linked issue) they do convert trial accounts to free accounts. I agree it does not make that clear on their website, but I don't see why we can't link to it. Even if it was a pay-only service, I don't think it matters. We're just letting users know the options they can use. This is why I was suggesting we simplify the text for MyGet and not mention the free option.

giotavekiari:
Made some changes as discussed! Please check again! Thank you!

Thraka (approved):

PR #8624, opened 2018-10-25
Update `Search()` formatting to be consistent with earlier usage

eonarheim:
## Summary

Update formatting of `Search()` to be consistent with earlier usage

Fixes: N/A

### DIFF
--- a/docs/csharp/event-pattern.md
+++ b/docs/csharp/event-pattern.md
@@ -181,7 +181,7 @@ EventHandler<FileFoundArgs> onFileFound = (sender, eventArgs) =>
 ## Adding Another Event Declaration
 
 Let's add one more feature, and demonstrate other language idioms
-for events. Let's add an overload of the `Search()` method that
+for events. Let's add an overload of the `Search` method that
 traverses all subdirectories in search of files.
 
 This could get to be a lengthy operation in a directory with many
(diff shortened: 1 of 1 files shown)


eonarheim:
I also noticed that the `SearchDirectoryArgs` sample is a struct doesn't extend `EventArgs`, I can open another pr for this or include it in this one. Not sure if that was intentional when other samples are classes and extend `EventArgs`.

image

rpetrusha (changes requested):
Thank you for contributing to the dotnet/docs repo and standardizing the use of `Search`, @eonarheim. A better choice, though, would be to use `Search` rather than `Search()`; I'd appreciate it if you could make that change. As you noted, `SearchDirectoryArgs` should be a class that inherits from `EventArgs`, If you'd like, you can also make that change in this PR.

eonarheim:
@rpetrusha I've made the requested changes, looks like the sample are stored in a separate repo so I've made a PR there as well https://github.com/dotnet/samples/pull/423

rpetrusha (approved):

rpetrusha:
Thank you, @eonarheim, for submitting the additional PR in the dotnet/samples repo, as well as for making the additional changes to this PR. I'll merge your PR now. Your changes should be live on docs.microsoft.com in the next day or two.

PR #8654, opened 2018-10-26
Include languages

terrajobst:
@KathleenDollard gave feedback and said we should include languages there, which makes sense to me.


JamesNK (approved):

rpetrusha (approved):
This is a good addition, @terrajobst. I'll merge your PR now.

PR #8679, opened 2018-10-26
Mark IClassFixture as code in prose

MartinWilkerson:
`IClassFixture<WebApplicationFactory<TEntry>>` was rendering as `IClassFixture<WebApplicationFactory>`. Adding backquotes gets it to render the missing `<TEntry>` part at least.

## Summary

In the functional testing section, `IClassFixture<WebApplicationFactory<TEntry>>` was not rendering correctly. The `<Tentry>` part was being removed by the markdown renderer. Putting backquotes around it fixes it not rendering, however rendering in fixed width may not be what you want, but it was the quickest fix. I'm happy to play around with escaping to get it to render in non-fixed-width if preferred.


rpetrusha (changes requested):
Thanks for contributing to the dotnet/docs repo and making this correction, @MartinWilkerson. Although your change restores the missing \<TEntry>, it isn't consistent with the remainder of the text in the topic, which doesn't fence type names except in two bulleted lists earlier in the article. A better fix that's consistent with the topic is to escape the opening bracket: IClassFixture\\<WebApplicationFactory\\\<TEntry>>. Would you mind making that change?

Also, completely unrelated to your PR, there's a typoi

MartinWilkerson:
I have made the requested change and it looks like it renders correctly. Thank you for your help.

rpetrusha (approved):

rpetrusha:
Thanks, @MartinWilkerson, for making the additional change. I'll merge your PR now. The change should be live on docs.microsoft.com in the next day or two.

PR #8683, opened 2018-10-27
Consider keeping obsolete API

ejball:
## Summary

A reminder that it isn't always necessary or wise to eventually remove obsolete types and methods.

### DIFF
--- a/docs/standard/library-guidance/breaking-changes.md
+++ b/docs/standard/library-guidance/breaking-changes.md
@@ -88,6 +88,10 @@ public class Document
 }
 ```
 
+**✔️ CONSIDER** keeping types and methods with the <xref:System.ObsoleteAttribute> indefinitely in low and middle-level libraries.
+
+> Removing an obsolete type or method is a binary breaking change. Not removing them can help avoid the worst-case scenarios mentioned above.
+
 ## See also
 
 * [Version and update considerations for C# developers](../../csharp/whats-new/version-update-considerations.md)


rpetrusha:
//cc @JamesNK @terrajobst

docs/standard/library-guidance/breaking-changes.md
```diff
@@ -88,6 +88,10 @@ public class Document
 }
 ```
 
+**✔️ CONSIDER** keeping types and methods with the <xref:System.ObsoleteAttribute> indefinitely in low and middle-level libraries.
+
+> Removing an obsolete type or method is a binary breaking change. Not removing them can help avoid the worst-case scenarios mentioned above.
```
JamesNK: Overall I like the advice. I think it does depend on the situation. If the obsolete API is rarely used, or keeping it creates a lot of technical debt then I think removing it after some time is a good idea.

```suggestion
> Removing APIs is a binary breaking change. Considering keeping the obsolete types and methods if maintaining them doesn't add lot of technical debt to your library. Not removing types and methods can help avoid the worst-case scenarios mentioned above.
```

JamesNK:
@terrajobst re: https://github.com/dotnet/docs/pull/8683#issuecomment-434025063

Need to include what exactly `[EditorBrowsable(EditorBrowsableState.Never)]` does, i.e. how does it clean up code competition.

rpetrusha (approved):
Thank you, @ejball, for contributing to the dotnet/docs repo, for submitting this PR, and for making the siuggested changes. We'll merge your PR now. Your changes should be live on docs.microsoft.com in the next day or two.

ejball:
Thanks for making it so easy!

PR #8746, opened 2018-10-30
Update F# optional parameters docs

gsomix:
## Summary

+ add information about passing existing optional values to method with optional parameters
+ add description of F# 4.1 `DefaultParameterValue` feature
+ snippet https://github.com/dotnet/samples/pull/429/

@cartermp Review, please!

### DIFF
--- a/docs/fsharp/language-reference/parameters-and-arguments.md
+++ b/docs/fsharp/language-reference/parameters-and-arguments.md
@@ -119,7 +119,7 @@ The output is as follows.
 Baud Rate: 9600 Duplex: Full Parity: false
 Baud Rate: 4800 Duplex: Half Parity: false
 Baud Rate: 300 Duplex: Half Parity: true
-Baud Rate: 9600 Duplex: Full Parity: false4
+Baud Rate: 9600 Duplex: Full Parity: false
 Baud Rate: 9600 Duplex: Full Parity: false
 Baud Rate: 4800 Duplex: Half Parity: false
 ```
--- a/docs/fsharp/language-reference/parameters-and-arguments.md
+++ b/docs/fsharp/language-reference/parameters-and-arguments.md
@@ -137,11 +137,11 @@ type C =
 The value given as argument to `DefaultParameterValue` must match the type of the parameter, i.e. the following is not allowed:
 
 ```fsharp
-type C() =
+type C =
     static member Wrong([<Optional; DefaultParameterValue("string")>] i:int) = ()
 ```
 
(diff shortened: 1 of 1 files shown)


docs/fsharp/language-reference/parameters-and-arguments.md
```diff
@@ -117,6 +119,19 @@ The output is as follows.
 Baud Rate: 9600 Duplex: Full Parity: false
 Baud Rate: 4800 Duplex: Half Parity: false
 Baud Rate: 300 Duplex: Half Parity: true
+Baud Rate: 9600 Duplex: Full Parity: false
+Baud Rate: 9600 Duplex: Full Parity: false
+Baud Rate: 4800 Duplex: Half Parity: false
+```
+
+Since F# 4.1 you can use the attributes `[<Optional; DefaultParameterValue<(...)>]` in F# for C# and VB interop, so that C#/VB callers will see an argument as optional. This is equivalent to defining the argument as optional in C# as in `MyMethod(int i = 3)`.
```
cartermp: Given that F# 4.1 was released in early 2017, I think we can drop the "Since F# 4.1" and just say, "You can...". It's probably also worth mentioning that the value passed into the constructor of `DefaultParameterValue` needs to match the type of the argument to the method.

cartermp (approved):
I think this is good to merge as-is, but one suggestion could make it better. Thanks @gsomix!

gsomix:
@cartermp Done!

docs/fsharp/language-reference/parameters-and-arguments.md
```diff
@@ -134,6 +134,15 @@ type C =
...
 ```
 
+The value given as argument to `DefaultParameterValue` must match the type of the parameter, i.e. the following is not allowed:
+
+```fsharp
+type C() =
+    static member Wrong([<Optional;DefaultParameterValue("string")>] i:int) = ()
+```
+
+In this case compiler generate a warning, and ignore both attributes altogether. Note that the default value `null` needs to be type-annotated, as otherwise the compiler infers the wrong type, i.e. `[<Optional;DefaultParameterValue(null:obj)>] o:obj`.
```
cartermp: Change to: "In this case, the compiler generates a warning and will ignore both attributes altogether." or some variant, as the current first sentence is not grammatically correct.

cartermp (changes requested):
Just one small change and this is good to go.

docs/fsharp/language-reference/parameters-and-arguments.md
gsomix: My bad.

cartermp (approved):
Thanks!

gsomix:
@cartermp Thank you!

PR #8752, opened 2018-10-30
Update net-standard.md

Soundug:


Soundug:
Closes Soundug/standard#736

docs/standard/net-standard.md
```diff
@@ -29,7 +29,7 @@ To find the highest version of .NET Standard that you can target, do the followi
 
 1. Find the row that indicates the .NET implementation you want to run on.
 2. Find the column in that row that indicates your version starting from right to left.
-3. The column header indicates the .NET Standard version that your target supports (and any lower .NET Standard versions will also support it).
+3. The column header indicates the .NET Standard version that your target supports. You may also target any lower .NET Standard. Higher .NET Standard will also support your implementation.
```
rpetrusha: lower .NET Standard --> lower .NET Standard version
Higher .NET Standard --> Higher .NET Standard versions

rpetrusha (changes requested):
Thanks, @Soundug, for contributing to the dotnet/docs repo and clarifying this paragraph. I've suggested a few changes.

rpetrusha (approved):

docs/standard/net-standard.md
```diff
@@ -29,7 +29,7 @@ To find the highest version of .NET Standard that you can target, do the followi
 
 1. Find the row that indicates the .NET implementation you want to run on.
 2. Find the column in that row that indicates your version starting from right to left.
-3. The column header indicates the .NET Standard version that your target supports (and any lower .NET Standard versions will also support it).
+3. The column header indicates the .NET Standard version that your target supports. You may also target any lower .NET Standard version. Higher .NET Standard versions will also support your implementation.
```
richlander: How about:

```
The column header indicates the .NET Standard version that your chosen .NET implementation version supports. You may also target lower .NET Standard versions.
```

This part is true: `Higher .NET Standard versions will support your implementation.` I think it is almost an orthogonal topic, for two different reasons: compatibility is primary a function of the .NET implementation and less the standard and what the future looks like is a policy issue that isn't represented by this table.

PR #8827, opened 2018-11-03
Replace media14.png

RedasP:
Replace media14.png to make it more readable in Dark mode


BillWagner:
Tagging @CESARDELATORRE 

We've had several concerns with the images in dark mode.  This is a good solution. If we merge this, we should update the other images in the same section to match.

What are your thoughts?

CESARDELATORRE:
Not sure about it. Grey doesn't make it a clear image... What about forcing white for the background?
Adding @nishanil so he'll track down these. 👍

PR #8891, opened 2018-11-06
WPA also includes msmq

Styxxy:
Windows Process Activation, root page: "in this section" was missing MSMQ activation, though it is also a page in this section.

Docs url: https://docs.microsoft.com/en-us/dotnet/framework/wcf/samples/windows-process-activation


rpetrusha (approved):
Thanks, @Styxxy, for updating this article. We'll merge your PR now. The addition will be live on docs.microsoft.com in the next day or two.

PR #8911, opened 2018-11-07
Added in the command for creating a NUnit test project

OsirisTerje:
## Summary

dotnet new nunit
creates a similar nunit test project as the mstest and xunit.  Added that to the list of new commands


BillWagner (approved):

BillWagner:
Thanks for spotting this and fixing it @OsirisTerje We appreciate it.

Thanks again for contributing to docs. 🎆

PR #8922, opened 2018-11-07
Removed the Interlocked conceptual docs in favour of API reference

pkulikov:
Contributes to #8121

The [Interlocked](https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked?view=netframework-4.7.2) API page covers the main part of the current `Interlocked` [conceptual article](https://docs.microsoft.com/en-us/dotnet/standard/threading/interlocked-operations), contains the same `CompareExchange` example. As for the pre-generic era [example](https://docs.microsoft.com/en-us/dotnet/standard/threading/interlocked-operations#untyped-overloads-of-exchange-and-compareexchange), similar example was removed from the API docs earlier (see dotnet/dotnet-api-docs#162), so I see no need to keep it.


rpetrusha (approved):
This looks good, @pkulikov; thanks for doing it. I'll merge your PR now. 
There was nothing here that is worth incorporating in the Interlocked documentation, though I noticed a few issues, so I'm going to open a quick PR to correct them. I'll cc you.

PR #8938, opened 2018-11-08
Change "destructor" to "finalizer" in comment

zanedp:
The word "finalizer" is used everywhere else on the page.

## Summary

Fixes a comment in the sample code to use the same term used everywhere else on the page.


rpetrusha (changes requested):
Thanks, @zanedp, for submitting this PR to make the usage of finalizer and destructor consistent in the finalizer article. There are, however, eight more instances of **destructor** in the article's code. Can you change those as well?

rpetrusha:
Thanks, @zanedp, for making the additional changes. I'll merge your PR now. It should be live on docs.microsoft.com in the next day or two.

rpetrusha (approved):

PR #8985, opened 2018-11-12
Fix Remarks table broken in fsharp "constraints"

vbfox:
I don't think this line should be part of the table, github markdown formatter integrate it but it look weird and it completely break formatting on the real website.

image


cartermp (approved):
Thanks @vbfox!

PR #8989, opened 2018-11-12
Wrong namespace in the xml

neilbgr:
In Xml the namespace was "www.example.com". Everywhere is "www.contoso.com"


BillWagner (approved):
Thanks for spotting this and fixing it @neilbgr We appreciate it.

I'll :shipit: now. You will see the changes on the live site in the next day or two, on our regular publishing cycle.

Thanks again!

PR #8993, opened 2018-11-12
Update com-callable-wrapper.md

sharpist:
line 81-93: fixed code syntax
line 138-148: fixed code syntax
line 167-177: fixed code syntax

### DIFF
--- a/docs/framework/interop/com-callable-wrapper.md
+++ b/docs/framework/interop/com-callable-wrapper.md
@@ -84,8 +84,6 @@ End Class
 // Implicitly extends System.Object.
 public class Mammal
 {
-    public Mammal() {}
-
     public void Eat() {}
     public void Breathe() {}
     public void Sleep() {}
@@ -141,9 +139,7 @@ End Class
 [ClassInterface(ClassInterfaceType.None)]
 public class LoanApp : IExplicit
 {
-    public LoanApp() {}
-
-    Int32 IExplicit.M() { return 0; }
+    int IExplicit.M() { return 0; }
 }
 ```
   
@@ -170,9 +166,7 @@ End Class
 [ClassInterface(ClassInterfaceType.AutoDispatch)]
 public class LoanApp
 {
-    public LoanApp() {}
-
-    public Int32 M() { return 0; }
+    public int M() { return 0; }
 }
 ```
   
--- a/docs/framework/interop/com-callable-wrapper.md
+++ b/docs/framework/interop/com-callable-wrapper.md
@@ -134,8 +134,6 @@ End Class
 ```  
   
 ```csharp
-// The CLR does not expose a class interface for this type.
-// COM clients can call the members of this class using the methods from the IExplicit interface.
 [ClassInterface(ClassInterfaceType.None)]
 public class LoanApp : IExplicit
 {
@@ -161,8 +159,6 @@ End Class
 ```  
   
 ```csharp
-// Have the CLR expose a class interface (derived from IDispatch) for this type.
-// COM clients can call the members of this class using the Invoke method from the IDispatch interface.
 [ClassInterface(ClassInterfaceType.AutoDispatch)]
 public class LoanApp
 {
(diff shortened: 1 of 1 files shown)


richlander:
@AaronRobinsonMSFT and I are going to review this PR and document together in the new week. I read most of the doc and found multiple things that could be improved.

This is also a good time to decide what we're going to do w/rt COM Interop docs for the upcoming .NET Core 3.0 release, which will support this feature/scenario.

sharpist:
Hello, @richlander and @AaronRobinsonMSFT !
Yes, I want to be part of this. Any of my notes that are useful can be supplemented or corrected and used.
COM Interop documentation may be improved with time including new features in .NET Core 3.0.
Thanks!

docs/framework/interop/com-callable-wrapper.md
```diff
@@ -84,6 +84,8 @@ End Class
 // Implicitly extends System.Object.
 public class Mammal
 {
+    public Mammal() {}
```
svick: If a class doesn't define any other constructor, it will have a default public parameterless constructor. So I don't think adding the constructor here and also below is necessary.

docs/framework/interop/com-callable-wrapper.md
```diff
@@ -134,9 +136,14 @@ End Class
...
+// The CLR does not expose a class interface for this type.
+// COM clients can call the members of this class using the methods from the IExplicit interface.
 [ClassInterface(ClassInterfaceType.None)]
-public class LoanApp : IExplicit {
-    public void M() {}
+public class LoanApp : IExplicit
+{
+    public LoanApp() {}
+
+    Int32 IExplicit.M() { return 0; }
```
svick: I think it's more idiomatic to use `int`, rather than `Int32`. Also, you could use expression-bodied method here:

```suggestion
    int IExplicit.M() => 0;
```

The same applies for the block below.

docs/framework/interop/com-callable-wrapper.md
sharpist: You are right, this default public parameterless constructor has been added only as a visual example for perception. It can be deleted.

docs/framework/interop/com-callable-wrapper.md
sharpist: Here I would choose a classic implementation. Because the 'return' instruction is more recognizable and identifiable.

AaronRobinsonMSFT: @sharpist I think the 'classic implementation' argument is valid here, but the type really should be `int`.

docs/framework/interop/com-callable-wrapper.md
```diff
@@ -134,9 +134,12 @@ End Class
 ```  
   
 ```csharp
+// The CLR does not expose a class interface for this type.
+// COM clients can call the members of this class using the methods from the IExplicit interface.
```
pkulikov: This comment is in the C# snippet, so readers with Visual Basic as the selected language will miss that comment. Moreover, comments are not translated, so this content won't be localized. @sharpist please consider moving this paragraph into the body of the article. 

The same is applicable to another snippet update.

sharpist: @pkulikov I think these both accompanying comments can be excluded from the code examples. At your discretion. Because the main article already contains a detailed description.

pkulikov: @sharpist 
>I think these both accompanying comments can be excluded from the code examples.

If the article contains enough explanations, I also don't see the need in the code comments.

richlander:
@AaronRobinsonMSFT and I were going to meet at 10AM PST on Monday (tomorrow). That looks like the best time (10AM) we can select. DM me on twitter @runfaster2000 and we can work out particulars.

docs/framework/interop/com-callable-wrapper.md
pkulikov: @Thraka yes
@sharpist thanks for additional updates

Thraka (approved):
To me this looks like a good improvement and corrects the problem of not actually implementing the explicit interface in the example

PR #8996, opened 2018-11-12
Lower Dispose SEO ranking on WCF specific article

BillWagner:
This article is specific to WCF and we want to discourage search ranking when developers are searching for general information on the C# using statement.

relates to #8065


mairaw (approved):
LGTM. Should we also update incoming links to use the new title? It seems that 6 files in our repo link to this topic.

PR #8999, opened 2018-11-13
Removed discussion of lazy quantifier

### LINKED ISSUE #8216: explaination is wrong regrading your statement about  .*?(es)
I tried out in regex101.com to check the debug steps, so .\*? is not greedy pattern, it's lazy pattern, it only takes characters left behind by (es) in each round, therefore it makes no differences between .\*?(es) and (es) to your input "Essential services are provided by regular expressions."
because in the 1st round it can find the match 'es' in 'services' already. 

Your explanation is correct when there are no matches, as I mentioned above what I have observed is that each rounds .\*? will take all the characters left behind by (es), so instead of saying it's tracebacking, I would say .\*? actually hold the position pointer in string for each round, so it moves very slowly, each round is an inside loop, and if there is no matches, the .\* will spit out all the characters and move forward the position by 1 character and then start a new round of loop, that's why 'more than twice' quoted your statement.

In summary, if there is a match, (es) and .\*?(es) make no much differences, but if there is no matches:
(es) just scan through the string once(one round), 
.\*?(es) will have many rounds(rounds==length of the string in this case), and in each round move forward one position/character

---
#### Document Details

* ID: 3ce579cf-f3af-6050-895b-d00ae083c26b
* Version Independent ID: 89943c6e-1d5e-737d-cac8-ef87e8d42cf6
* Content: [Backtracking in Regular Expressions](https://docs.microsoft.com/en-us/dotnet/standard/base-types/backtracking-in-regular-expressions)
* Content Source: [docs/standard/base-types/backtracking-in-regular-expressions.md](https://github.com/dotnet/docs/blob/master/docs/standard/base-types/backtracking-in-regular-expressions.md)
* Product: **dotnet**
* GitHub Login: @rpetrusha
* Microsoft Alias: **ronpet**

rpetrusha:
## Removed discussion of lazy quantifier

Fixes #8216

### DIFF
--- a/docs/standard/base-types/backtracking-in-regular-expressions.md
+++ b/docs/standard/base-types/backtracking-in-regular-expressions.md
@@ -1,6 +1,6 @@
 ---
 title: "Backtracking in Regular Expressions"
-ms.date: "03/30/2017"
+ms.date: "11/12/2018"
 ms.technology: dotnet-standard
 dev_langs: 
   - "csharp"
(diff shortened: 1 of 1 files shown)


mairaw (approved):
You're the expert on this so LGTM 😎

PR #9048, opened 2018-11-15
Fix style formatting

pkulikov:
The current display of the alarm box at [Composite formatting](https://docs.microsoft.com/en-us/dotnet/standard/base-types/composite-formatting):
image

My guess, it's caused by two white spaces after the `>` characters
Does removal of excessive white spaces fix rendering?


BillWagner (approved):
This LGTM @pkulikov Thanks for spotting this and making the fix.

I'll :shipit: now.

PR #9161, opened 2018-11-21
Add dotnet Pack the projec using a .nuspec example

devlead:
## Summary

Added an example of how you can use a .nuspec file to pack your project along with SDK project file and link to where it's documented. MSBuild targets probably isn't where people go looking for this, so this example makes it a bit more discoverable.


rpetrusha (approved):
Thanks, @devlead, for adding this example. We'll merge your PR now. The example should be live on docs.microsoft.com in the next day or two.

Issue #9070, opened 2018-11-16
Formatting guidelines: no rule for formatting literals (constants), nor any rule for attributes


abelbraaksma:
This article describes using `camelCase` and `CamelCase` for must of the identifiers and/or classnames etc, but there is no mention of what the preferred style for literals is:

```f#
let [<Literal>] iceCube = "cold"   // camelCase
let [<Literal>] IceCube = "cold"  // CamelCase
let [<Literal>] ICECUBE = "cold"   // ALLCAPS
let [<Literal>] ICE_CUBE = "cold"   // ALL_CAPS
let [<Literal>] ice_cube = "cold"  // snake_case
```

Also, do we have any preference for placement of attributes? I think the rule ought to be that anything _but_ the `Literal` attribute should be placed on its own line and should not span multiple lines. Each attribute should be on its own line.

```f#
[<SomeAttribute>] 
class GetIce() ...

[<SomeAttribute>]
let Foo = "foo"

let [<Literal>] Foo = "foo"

// NOT OK (raises warning due to a compiler bug)
[<Literal>] let Foo = "foo"
[<Literal>] let Boo = "foo"
[<Literal>] let Hoo = "foo"
```

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 8c09996f-24ac-cf39-ef43-d7909cbc2b89
* Version Independent ID: 1efdd523-8552-6421-e149-74f76826ef8f
* Content: [F# code formatting guidelines](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting)
* Content Source: [docs/fsharp/style-guide/formatting.md](https://github.com/dotnet/docs/blob/master/docs/fsharp/style-guide/formatting.md)
* Product: **dotnet-fsharp**
* GitHub Login: @cartermp
* Microsoft Alias: **phcart**

PR #9168, opened 2018-11-21
Add attribute formatting guidelines

cartermp:
Adding formatting for attributes, to address #9070

### DIFF
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -604,3 +604,45 @@ let makeStreamReader x = new System.IO.StreamReader(path=x)
 // Not OK
 let makeStreamReader x = new System.IO.StreamReader(path = x)
 ```
+
+## Formatting attributes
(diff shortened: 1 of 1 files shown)


docs/fsharp/style-guide/formatting.md
```diff
@@ -604,3 +604,45 @@ let makeStreamReader x = new System.IO.StreamReader(path=x)
...
+
+```fsharp
+// Defines a class that takes an optional value as input defaulting to false.
+type C() =
+    member __.M([<Optional; DefaultParameterValue(false)>] doSomething: bool)
+```
+
+### Formatting multiple attributes
+
+When attributes are applied to a construct that is not a parameter, they should be placed such that there is one attribute per line:
```
rpetrusha: When attributes --> When multiple attributes

rpetrusha (approved):
This looks good, @cartermp. I've just left one suggested change.

PR #9186, opened 2018-11-22
Update with-visual-studio-code.md

kennethjohnsen:
## Summary

The "continue" arrow in VS code is actually blue, and is more of a pause/play symbol.
The only green things are the "Start debugging" button and the "restart" button, which both do other things than continue debugging.
Not sure my edit is correct either, as there are several blue arrows at the top...


kennethjohnsen:
> I think that the number 1 problem is that there is a screenshot missing with the debugger controls, show below. It is also the case that the VS Code screenshots are for the old UI model and has been updated. It is probably time we re-update them all.
> 
> Thoughts?
> 
> image

At least have an image with the debug-controls shown. Otherwise it's pretty hard to describe what the "continue" button looks like.

richlander:
I think we need to re-do all of the images as opposed to a super scoped image of he debug controls. The user base of VS Code is very high, so it warrants great docs. Are you interested in doing this? It is OK if you are not.

kennethjohnsen:
Yes, I can do that. I will have a look at it tomorrow.
Kinda new to this whole contributing/github thing, so do I just do it in this pull request, or submit a new one?

richlander:
Thanks!

Fine to continue using this PR. Just make sure you are using the latest VS Code.

kennethjohnsen:
Added the new images. Let me know what you think.
Btw. the "continue.png" does not seem to be used, and I think everything is shown in the new "rundebug.png" anyway, so I don't know if it's needed.
Also, the note on dotnet restore in .NET Core 2.0 looks like this in my fork: [!INCLUDEDotNet Restore Note], but I guess that's related to where the file to include is located?
Lastly, I'm still not sure changing "green" to "blue" as in my initial commit is sufficient explanation, but I don't know how else to describe the "Continue" button.

richlander:
This looks great to me. Thanks much!!

richlander (approved):

docs/core/tutorials/with-visual-studio-code.md
```diff
@@ -77,7 +77,7 @@ You can also watch a short video tutorial for further setup help on [Windows](ht
```
mairaw: can you please also update ms.date on line 6 to be 11/28/2018?

docs/core/tutorials/with-visual-studio-code.md
```diff
@@ -77,7 +77,7 @@ You can also watch a short video tutorial for further setup help on [Windows](ht
 
     ![Run and Debug](media/with-visual-studio-code/rundebug.png)
```
mairaw: given how the blue arrow and red square are now highlighted in this image, it'd probably be better to move this down to step 7 then.

mairaw (changes requested):
This is awesome @kennethjohnsen! Left just two minor comments for you to consider. Thanks!

docs/core/tutorials/with-visual-studio-code.md
kennethjohnsen: Fixed.

docs/core/tutorials/with-visual-studio-code.md
kennethjohnsen: I have moved it to step 7.
However, there's also a highlight around the debug-sidebar to show locals and watches etc. Is it OK to keep the images as-is?

mairaw: Yes, let's keep it like that and if we get any complaints, we can fix it.

mairaw (approved):
Thanks for applying the changes @kennethjohnsen. Your contribution should be live within a day or so.

PR #9194, opened 2018-11-23
Update supportedruntime-element.md

ForNeVeR:
## Summary

Add quotes around runtime values.

### DIFF
--- a/docs/framework/configure-apps/file-schema/startup/supportedruntime-element.md
+++ b/docs/framework/configure-apps/file-schema/startup/supportedruntime-element.md
@@ -71,12 +71,12 @@ The `sku` attribute uses a target framework moniker (TFM) to indicate the versio
 |----------------------------|---------------------|  
 |4.0|".NETFramework,Version=v4.0"|  
 |4.0, Client Profile|".NETFramework,Version=v4.0,Profile=Client"|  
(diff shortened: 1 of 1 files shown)


richlander:
Why would we add quotes? That seems odd to me. The quotes might be there for two reasons:

- Demonstrate that the data is a string.
- Convenience for copy pasting it into XML

I'd suggest that the first goal isn't necessary and if it is, we can describe this in a different way. Given that the quotes might already be in a target XML file, I'd suggest that the second goal is questionable.

Thoughts

ForNeVeR:
Yes, the quotes seem odd indeed, but the neighbour lines have the quotes already. Would you prefer me to remove the quotes from the whole table?

richlander:
> Yes, the quotes seem odd indeed, but the neighbour lines have the quotes already. Would you prefer me to remove the quotes from the whole table?

That's my feeling. It would be good to hear an opinion from someone else.

rpetrusha (approved):
Thanks, @ForNeVeR, for contributing to the dotnet/docs repo.
 Quotation marks are used with attribute values with the version attribute, and they're used with eleven of the 17 attribute values for the sku attribute. Since your PR adds quotation marks to the remaining six possible values of the sku attribute, I'll merge your PR now.

PR #9206, opened 2018-11-23
Update cs0116.md

pkulikov:
According to the [namespace](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/namespace) article, namespace can contain another namespace and no mention of unions.

What is "unions" in 
>Inside a namespace, the compiler only accepts classes, structures, unions, enumerations, interfaces, and delegates. 

by the way?


docs/csharp/language-reference/compiler-messages/cs0116.md
```diff
@@ -11,7 +11,7 @@ ms.assetid: 4cb137b5-ec29-4c1a-adde-9f8424cb9496
 
 A namespace does not directly contain members such as fields or methods  
   
- Inside a [namespace](../../../csharp/language-reference/keywords/namespace.md), the compiler only accepts classes, structures, unions, enumerations, interfaces, and delegates. This error is often generated by developers from a C/C++ background who forget that in C#, methods and variables must be declared and defined within a struct or class. For more information, see [General Structure of a C# Program](../../../csharp/programming-guide/inside-a-program/general-structure-of-a-csharp-program.md).  
+ Inside a [namespace](../keywords/namespace.md), the compiler only accepts another namespaces, classes, structures, enumerations, interfaces, and delegates. This error is often generated by developers from a C/C++ background who forget that in C#, methods and variables must be declared and defined within a struct or class. For more information, see [General Structure of a C# Program](../../../csharp/programming-guide/inside-a-program/general-structure-of-a-csharp-program.md).
```
richlander: > another namespaces

This phrasing isn't quite right (English) grammatically. "another" is expecting a singular noun but you've provided a plural one.

I don't have any sense of the C# grammar that is being discussed.

pkulikov: @richlander thanks for the correction; I think "another" can be dropped at all.

svick:
> What is "unions" […] by the way?

In C++, apart from `class`es, `struct`s and `enum`s, you can also have `union`s as another type kind. And since the article says:

> This error is often generated by developers from a C/C++ background who forget that in C#, methods and variables must be declared and defined within a struct or class.

My guess is that:

> This error was generated by a writer from a C/C++ background who forgot that in C#, unions don't exist.

😉

docs/csharp/language-reference/compiler-messages/cs0116.md
```diff
@@ -11,7 +11,7 @@ ms.assetid: 4cb137b5-ec29-4c1a-adde-9f8424cb9496
 
 A namespace does not directly contain members such as fields or methods  
   
- Inside a [namespace](../../../csharp/language-reference/keywords/namespace.md), the compiler only accepts classes, structures, unions, enumerations, interfaces, and delegates. This error is often generated by developers from a C/C++ background who forget that in C#, methods and variables must be declared and defined within a struct or class. For more information, see [General Structure of a C# Program](../../../csharp/programming-guide/inside-a-program/general-structure-of-a-csharp-program.md).  
+ Inside a [namespace](../keywords/namespace.md), the compiler only accepts namespaces, classes, structures, enumerations, interfaces, and delegates. This error is often generated by developers from a C/C++ background who forget that in C#, methods and variables must be declared and defined within a struct or class. For more information, see [General Structure of a C# Program](../../../csharp/programming-guide/inside-a-program/general-structure-of-a-csharp-program.md).
```
mikedn: Perhaps it would be better to simply say that it only accepts namespaces and type definitions rather than enumerating the various kind of types C# has. If new types are added to C# (e.g. records) then this kind of description will get stale.

pkulikov: @mikedn that's a good idea; thanks for sharing!

docs/csharp/language-reference/compiler-messages/cs0116.md
```diff
@@ -9,9 +9,11 @@ ms.assetid: 4cb137b5-ec29-4c1a-adde-9f8424cb9496
 ---
 # Compiler Error CS0116
 
-A namespace does not directly contain members such as fields or methods  
+A namespace cannot directly contain members such as fields or methods
```
pkulikov: Updated this line based on this code:
https://github.com/dotnet/roslyn/blob/1fbd48157d9196ad227f120b731d3f80b1b162de/src/Compilers/CSharp/Portable/CSharpResources.resx#L652

docs/csharp/language-reference/compiler-messages/cs0116.md
```diff
@@ -9,9 +9,11 @@ ms.assetid: 4cb137b5-ec29-4c1a-adde-9f8424cb9496
 ---
 # Compiler Error CS0116
 
-A namespace does not directly contain members such as fields or methods  
+A namespace cannot directly contain members such as fields or methods  
   
- Inside a [namespace](../../../csharp/language-reference/keywords/namespace.md), the compiler only accepts classes, structures, unions, enumerations, interfaces, and delegates. This error is often generated by developers from a C/C++ background who forget that in C#, methods and variables must be declared and defined within a struct or class. For more information, see [General Structure of a C# Program](../../../csharp/programming-guide/inside-a-program/general-structure-of-a-csharp-program.md).  
+A namespace can contain only other namespaces and type declarations. For more information, see the [namespace keyword](../keywords/namespace.md) article.
```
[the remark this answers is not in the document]
pkulikov: @richlander then we need to define the term "child namespace". Moreover, docs rather use "nested namespace" than "child namespace"; I'm not sure if the "child namespace" is an "official" term (language spec uses "nested namespace").

What about:
> A namespace can contain nested namespaces and type declarations

?? (though, "contain nested" is a duplication)

BillWagner (approved):
This LGTM @pkulikov 

Thanks for updating based on @richlander 's comments.

I'll :shipit: now.

PR #9214, opened 2018-11-24
Fixed inaccuracies in the C# types overview

pkulikov:
Rather than fixing the "value types are also nullable" statement (`int` is not nullable), I've removed the whole paragraph, as this paragraph doesn't fit the intention of the page: provide very high level overview of the C# types (note, the page is a part of C# keywords language reference). The [Value types](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/value-types) article does a better job to mention nullable value types.

### DIFF
--- a/docs/csharp/language-reference/keywords/types.md
+++ b/docs/csharp/language-reference/keywords/types.md
@@ -16,13 +16,11 @@ The C# typing system contains the following categories:
 
 - [Pointer types](../../programming-guide/unsafe-code-pointers/pointer-types.md)
 
- Variables that are value types store data, and those that are reference types store references to the actual data. Reference types are also referred to as objects. Pointer types can be used only in [unsafe](unsafe.md) mode.
+ Variables that are value types store data, and those that are reference types store references to the actual data. Instances of reference types are also referred to as objects. Pointer types can be used only in [unsafe](unsafe.md) mode.
 
  It's possible to convert a value type to a reference type, and back again to a value type, by using [boxing and unboxing](../../../csharp/programming-guide/types/boxing-and-unboxing.md). With the exception of a boxed value type, you cannot convert a reference type to a value type.
 
  This section also introduces [void](void.md).
-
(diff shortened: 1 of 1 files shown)


docs/csharp/language-reference/keywords/types.md
```diff
@@ -16,19 +16,17 @@ The C# typing system contains the following categories:
 
 - [Pointer types](../../programming-guide/unsafe-code-pointers/pointer-types.md)
 
- Variables that are value types store data, and those that are reference types store references to the actual data. Reference types are also referred to as objects. Pointer types can be used only in [unsafe](unsafe.md) mode.
+ Variables that are value types store data, and those that are reference types store references to the actual data. Instances of reference types are also referred to as objects. Pointer types can be used only in [unsafe](unsafe.md) mode.
```
richlander: This explanation, in general, isn't really working for me. These concepts are fundamental to both C# and .NET yet this description is very narrow and (IMO) confusing. I clicked on the description provided for reference types. It's a much better description. IMO, it's not a great description for reference types because it re-describes value and reference types again. It should focus on reference types.

I'd suggest we merge this PR and create an issue to address the feedback I've given. Thoughts?

pkulikov: > I'd suggest we merge this PR and create an issue to address the feedback I've given.

Agree that the page under consideration can be improved further. Given its place in the TOC, it should rather describe and reference descendant articles.

The description of value/reference types seems to be more a concern of the "Concepts" part of the docs, not the "Language reference", though of course, they should link each other.

richlander (approved):

docs/csharp/language-reference/keywords/types.md
richlander: @pkulikov Want me to run with my feedback or did you intend to do that?

pkulikov: @richlander please create an issue with your feedback

PR #9215, opened 2018-11-24
Value types: define simple types

pkulikov:
While the `char` type is an integral type, I think it's still worth an explicit mention in the list of structs.

Other changes are minor style updates.

### DIFF
--- a/docs/csharp/language-reference/keywords/value-types.md
+++ b/docs/csharp/language-reference/keywords/value-types.md
@@ -1,5 +1,5 @@
 ---
-title: "Value Types (C# Reference)"
+title: "Value types (C# Reference)"
 ms.date: 07/20/2015
 f1_keywords: 
   - "cs.valuetypes"
@@ -9,26 +9,28 @@ helpviewer_keywords:
   - "C# language, value types"
 ms.assetid: 471eb994-2958-49d5-a6be-19b4313f80a3
 ---
-# Value Types (C# Reference)
-The value types consist of two main categories:  
+# Value types (C# Reference)
+
+The value types consist of two main categories:
+
+- [Structs](struct.md)
+
+- [Enumerations](enum.md)
+ 
+ Structs fall into these categories:
+
+- Numeric types and [char](char.md)
+
+    - [Integral types](integral-types-table.md) (including the [char](char.md) type)
+
+    - [Floating-point types](floating-point-types-table.md)
+
+- [bool](bool.md)
+
+- User-defined structs.
   
--   [Structs](../../../csharp/language-reference/keywords/struct.md)  
-  
--   [Enumerations](../../../csharp/language-reference/keywords/enum.md)  
-  
- Structs fall into these categories:  
-  
--   Numeric types  
-  
-    -   [Integral types](../../../csharp/language-reference/keywords/integral-types-table.md)  
-  
-    -   [Floating-point types](../../../csharp/language-reference/keywords/floating-point-types-table.md)  
-  
--   [bool](../../../csharp/language-reference/keywords/bool.md)  
-  
--   User defined structs.  
-  
-## Main Features of Value Types  
+## Main features of value types
+
  Variables that are based on value types directly contain values. Assigning one value type variable to another copies the contained value. This differs from the assignment of reference type variables, which copies a reference to the object but not the object itself.  
   
  All value types are derived implicitly from the <xref:System.ValueType?displayProperty=nameWithType>.  
@@ -37,16 +39,18 @@ The value types consist of two main categories:
   
  Unlike reference types, a value type cannot contain the `null` value. However, the [nullable types](../../../csharp/programming-guide/nullable-types/index.md) feature does allow for value types to be assigned to `null`.  
   
- Each value type has an implicit default constructor that initializes the default value of that type. For information about default values of value types, see [Default Values Table](../../../csharp/language-reference/keywords/default-values-table.md).  
+ Each value type has an implicit default constructor that initializes the default value of that type. For information about default values of value types, see [Default values table](../../../csharp/language-reference/keywords/default-values-table.md).  
   
-## Main Features of Simple Types  
- All of the simple types -- those integral to the C# language -- are aliases of the .NET Framework System types. For example, [int](../../../csharp/language-reference/keywords/int.md) is an alias of <xref:System.Int32?displayProperty=nameWithType>. For a complete list of aliases, see [Built-In Types Table](../../../csharp/language-reference/keywords/built-in-types-table.md).  
+## Main features of simple types
+
+ All of the simple types -- those integral to the C# language -- are aliases of the .NET <xref:System> types. For example, [int](int.md) is an alias of <xref:System.Int32?displayProperty=nameWithType>. For a complete list of aliases, see [Built-in types table](../../../csharp/language-reference/keywords/built-in-types-table.md).  
   
  Constant expressions, whose operands are all simple type constants, are evaluated at compilation time.  
   
- Simple types can be initialized by using literals. For example, 'A' is a literal of the type `char` and 2001 is a literal of the type `int`.  
+ Simple types can be initialized by using literals. For example, `'A'` is a literal of the type `char` and `2001` is a literal of the type `int`.  
   
-## Initializing Value Types  
+## Initializing value types
+
  Local variables in C# must be initialized before they are used. For example, you might declare a local variable without initialization as in the following example:  
   
 ```csharp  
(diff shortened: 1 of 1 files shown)


docs/csharp/language-reference/keywords/value-types.md
```diff
@@ -9,26 +9,28 @@ helpviewer_keywords:
...
+
+The value types consist of two main categories:
+
+- [Structs](struct.md)
+
+- [Enumerations](enum.md)
+ 
+ Structs fall into these categories:
+
+- Numeric types and [char](char.md)
```
richlander: I think it is odd to mention char twice, particularly in the title. "Numeric types and char" doesn't seem like a category to me. I'd stick with "Numeric types". Char is either a numeric type of not.

On the topic of numeric types, I am still not convinced that char is a numeric type. This is the best code I can write that makes me consider it as numeric.

```csharp
char ten = (char)10;
char twenty = (char)20;
Console.WriteLine(ten * twenty);
```

The first two files rely on an explicit cast. If it was implicit, I'd feel more compelled. The thing that makes it numeric type is that it uses ints as both a backing data structure and it has a well-defined conversion with ints. So, to my mind, calling it numeric is more academic than practical.

I think that characterizing structs is difficult. Are ValueTask and Span in the user-defined structs category? If so, then I'd switch your categorization of structs to be: primitive, special , and user-defined. "special" is a bad name but I cannot think of something better ATM. Span is neither primitive or user-defined, in my mind. Another framing might be: provided (as in "in-box") and user-defined.

docs/csharp/language-reference/keywords/value-types.md
```diff
@@ -37,16 +39,18 @@ The value types consist of two main categories:
...
+ Each value type has an implicit default constructor that initializes the default value of that type. For information about default values of value types, see [Default values table](default-values-table.md).  
   
-## Main Features of Simple Types  
- All of the simple types -- those integral to the C# language -- are aliases of the .NET Framework System types. For example, [int](../../../csharp/language-reference/keywords/int.md) is an alias of <xref:System.Int32?displayProperty=nameWithType>. For a complete list of aliases, see [Built-In Types Table](../../../csharp/language-reference/keywords/built-in-types-table.md).  
+## Main features of simple types
+
+ All of the simple types -- those integral to the C# language -- are aliases of the .NET <xref:System> types. For example, [int](int.md) is an alias of <xref:System.Int32?displayProperty=nameWithType>. For a complete list of aliases, see [Built-in types table](built-in-types-table.md).
```
richlander: Here we have a different category "simple types". The categorization and this section should match.

We shouldn't use the word "integral". It is too close to integer, and there are better synonyms available.

I would remove the article in "are aliases of the .NET System types"

docs/csharp/language-reference/keywords/value-types.md
```diff
@@ -77,26 +81,26 @@ int myInt = new int();
...
 ```  
   
- Using the [new](../../../csharp/language-reference/keywords/new.md) operator calls the default constructor of the specific type and assigns the default value to the variable. In the preceding example, the default constructor assigned the value `0` to `myInt`. For more information about values assigned by calling default constructors, see [Default Values Table](../../../csharp/language-reference/keywords/default-values-table.md).  
+ Using the [new](new.md) operator calls the default constructor of the specific type and assigns the default value to the variable. In the preceding example, the default constructor assigned the value `0` to `myInt`. For more information about values assigned by calling default constructors, see [Default values table](default-values-table.md).  
   
- With user-defined types, use [new](../../../csharp/language-reference/keywords/new.md) to invoke the default constructor. For example, the following statement invokes the default constructor of the `Point` struct:  
+ With user-defined types, use [new](new.md) to invoke the default constructor. For example, the following statement invokes the default constructor of the `Point` struct:
```
richlander: Should new be in a `code` block?

pkulikov: As far as I observe and remember, code fencing is not applied in the link texts. though sometimes one can notice the link text with code fencing. @mairaw what is the rule here?

docs/csharp/language-reference/keywords/value-types.md
pkulikov: @richlander 
> I think that characterizing structs is difficult. 

Definitely. Thanks a lot for your response. It made me look at the article more critically.

I think the struct categorization is not necessary at all here. How does it help to reason about value types? What I've noticed is that the current "categorization" is about simple types and all other structs, which are called user-defined structs (yes, even [BigInteger](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.biginteger?view=netframework-4.7.2) becomes a user-defined struct). So, I've removed the categorization and, instead, provided the definition (through enumeration) of the *simple types*. Why is it necessary to have such a category: simple types? Because those differ from all other structs and those differences are then in the article.

I've updated the article based on the [C# language spec](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/types#simple-types) and omitted the last "feature" of the simple types. Please let me know if you think that it should go to the docs as well.

docs/csharp/language-reference/keywords/value-types.md
pkulikov: Please check new updates; I've updated this section a lot.

docs/csharp/language-reference/keywords/value-types.md
richlander: I agree. We shouldn't use code fencing for links (it looks strange), however, I think it is important for keywords to be code fenced. I noticed another usage where the first entry was code-fenced and then a later entry was linked. I found that to be the best model as a reader.

docs/csharp/language-reference/keywords/value-types.md
```diff
@@ -9,26 +9,16 @@ helpviewer_keywords:
...
+
+The value types consist of two main categories:
+
+- [Structs](struct.md)
+
+- [Enumerations](enum.md)
+
+## Main features of value types
+
  Variables that are based on value types directly contain values. Assigning one value type variable to another copies the contained value. This differs from the assignment of reference type variables, which copies a reference to the object but not the object itself.
```
richlander: I am having trouble with the "variables that are based on" phrase. How about something like:

> Variables in C# are declared or inferred to be a certain type. This type is is either a value or reference type. Variables that are a value type, like `int`, contain the value of the type, like `42`. This differs from variables that are a reference type, like `string`, which contain a reference (the memory location) to a value (otherwise known as an "object").

richlander:
I was looking at this text ...

> Unlike reference types, a value type cannot contain the `null` value. However, the nullable types feature does allow for value types to be assigned to `null`.

It would be good to make nullability seem more integrated. I think it is written this way since nullability came later. When we write the nullable reference types docs, we'll want them written a similar way, I think.

How about:

> Value type variables cannot be `null` by default. However, value type variables can be declared to be nullable, with the `?` syntax, and can then hold the `null` value.

docs/csharp/language-reference/keywords/value-types.md
```diff
@@ -37,16 +27,30 @@ The value types consist of two main categories:
...
- Each value type has an implicit default constructor that initializes the default value of that type. For information about default values of value types, see [Default Values Table](../../../csharp/language-reference/keywords/default-values-table.md).  
-  
-## Main Features of Simple Types  
- All of the simple types -- those integral to the C# language -- are aliases of the .NET Framework System types. For example, [int](../../../csharp/language-reference/keywords/int.md) is an alias of <xref:System.Int32?displayProperty=nameWithType>. For a complete list of aliases, see [Built-In Types Table](../../../csharp/language-reference/keywords/built-in-types-table.md).  
-  
- Constant expressions, whose operands are all simple type constants, are evaluated at compilation time.  
+ Each value type has an implicit default constructor that initializes the default value of that type. For information about default values of value types, see [Default values table](default-values-table.md).  
   
- Simple types can be initialized by using literals. For example, 'A' is a literal of the type `char` and 2001 is a literal of the type `int`.  
+## Simple types
```
richlander: - It seems like all of the following text should go into the structs topic not the more general value types topic. All of this is specific to structs, right?
- The "simple types" section seems to beg for a matching user-defined types section.

pkulikov: > The "simple types" section seems to beg for a matching user-defined types section

What would such a section contain? Is there any property of user-defined structs that is not valid for a simple type?

> It seems like all of the following text should go into the structs topic not the more general value types topic. All of this is specific to structs, right?

Though I agree with this, I don't think this PR should handle it. Simply because what is the structs topic? There are several candidates:
- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/struct
- https://docs.microsoft.com/en-us/dotnet/csharp/structs
- https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/structs

#966 is tracking the duplication issue.

docs/csharp/language-reference/keywords/value-types.md
```diff
@@ -9,44 +9,48 @@ helpviewer_keywords:
...
+
+There are two kinds of value types:
+
+- [Structs](struct.md)
+
+- [Enumerations](enum.md)
+
+## Main features of value types
+
+A variable of a value type contains a value of the type. For example, a variable of the `int` type might contain the value `42`. This differs from a variable of a reference type, which contains a reference to an instance of the type, also known as an object. When you assign a new value to a variable of a value type, that value is copied. When you assign a new value to a variable of a reference type, the reference is copied, not the object itself.
```
pkulikov: @richlander @BillWagner please check this one rigorously

docs/csharp/language-reference/keywords/value-types.md
richlander: > What would such a section contain? Is there any property of user-defined structs that is not valid for a simple type?

There may not be differences. It's more that there is a section dedicated to one form of structs and not to the other.

> Though I agree with this, I don't think this PR should handle it.

Sounds good.

richlander:
LGTM. Thanks!

richlander (approved):

docs/csharp/language-reference/keywords/value-types.md
pkulikov: > It's more that there is a section dedicated to one form of structs and not to the other.

I see the whole picture as one set of structs (simple types) is a subset of the set of all struct types. So, what should be described is (1) structs in general; (2) simple types. Though, it's good idea to say that a user-defined struct is a struct, which is not a simple type. I don't know where to put it yet in docs. So, come back to it later (together with another review of this topic, which can still be improved).

PR #9216, opened 2018-11-24
Revised a pointer member access operator article

pkulikov:
Like with the [address-of operator `&`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/and-operator#unary-address-of-operator), I've added the link to the dedicated programming guide article instead of embedding the example into the reference article.


BillWagner (approved):
This LGTM @pkulikov 

I'll :shipit: now.

Thanks again for all the updates.

PR #9247, opened 2018-11-26
Made it more clear page is for SQL Server; added Access info.

douglaslMS:
## Summary

The CSAT for this page has continued to decline (from 60% to 43%) even after I simplified the page and updated the links. 85% of visitors come from search, and most have searched for "Northwind database." I can only guess that lots of customers want it for Access, not for SQL. This seems to be the de facto Sample Databases page since there isn't one in the SQL or Access content. So I have done 2 things: (1) make it more clear that this page is for SQL Server, and (2) add brief info for Access users, since this info is not available except on third-party sites.


rpetrusha (approved):
This looks good, @douglaslMS. You can merge when you're ready.

PR #9266, opened 2018-11-27
Update cs0283.md: added missing types

pkulikov:
Half of the integer types was missing from the list of the types that are allowed in the constant declaration.


BillWagner (approved):
This is a good addition @pkulikov 

I'll :shipit: now.  Thanks again.

PR #9311, opened 2018-11-29
Add language identifiers to code blocks under linq

mikkelbu:
Relates to #2192


mairaw (approved):
That is great. Thank you @mikkelbu!

Issue #8330, opened 2018-10-14
Need some clarification or an example of unboxing


volkan-yazilim:
"Attempting to unbox a reference to an incompatible value type causes an InvalidCastException. "

I have coded and tried to any other value type than the original one. See below.
Should this sentence be changed to "Attempting to unbox a reference to the original value type causes an InvalidCastException." or what?

Thanks

int i = 1;
object o = i;
int a = (int)o; // OK
bool a = (bool)o; // NOT OK. any value type other than int
long l = (long)o; // NOT OK. Even no loss.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 75a88bb6-664f-99b3-07be-22a666888d11
* Version Independent ID: bea53588-8e8d-5f80-d572-178cd8ba48cb
* Content: [Boxing and Unboxing (C# Programming Guide)](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/types/boxing-and-unboxing#unboxing)
* Content Source: [docs/csharp/programming-guide/types/boxing-and-unboxing.md](https://github.com/dotnet/docs/blob/master/docs/csharp/programming-guide/types/boxing-and-unboxing.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Your analysis is correct @volkan-yazilim The unboxing conversion must have an identity conversion to the type of object that has been boxed.

volkan-yazilim:
https://github.com/dotnet/docs/pull/9319

msschl:
@BillWagner What do you think about adding your eample from the PR#9319 to the [docs](https://github.com/dotnet/docs/blob/master/docs/csharp/programming-guide/types/boxing-and-unboxing.md)?  
  
```csharp
public enum Color
{
    Red,
    Green,
    Blue
}

Color color = Color.Red; // a value type
object r = color;        // boxing
int i = (int) r;         // unboxing to a compatible type
```

BillWagner:
@msschl I like that idea.  I want to be careful of the language used to describe the conversions that are allowed. It's hard to make sure it's accurate.

PR #9319, opened 2018-11-30
Update boxing-and-unboxing.md

ghost:
" to other than the original value type "
https://github.com/dotnet/docs/issues/8330

## Summary

Describe your changes here.


BillWagner:
Thanks for submitting this change @volkan-yazilim  We appreciate all community contributions.

I'm going to reject this PR, because the original language is more correct.  Consider this `enum` type:

```csharp
public enum Color
{
  Red,
  Green,
  Blue
}
```

The following code compiles and runs without any exceptions:

```csharp
object r = Color.Red;
int i = (int) r;
Console.WriteLine(r);
Console.WriteLine(i);
```

`o` is being unboxed to a different type, but it is unboxed to a compatible type.

PR #9340, opened 2018-12-02
Remove redundant word.

ericmutta:
In the section `Defining a Time-out Interval` the second paragraph has a sentence which reads `A RegexMatchTimeoutException exception indicates that the regular expression engine was unable to find a match within in the...`. The words at the end should either be `...within the...` or `...in the...`


rpetrusha (approved):
Thank you, @ericmutta, for contributing to the dotnet/docs repo and for eliminating the redundant word. We'll merge your PR now. The change should be live on docs.microsoft.com in the next day or two.

PR #9353, opened 2018-12-03
Update await.md: mention that `Main` can be async

### LINKED ISSUE #9349: "updated example but incorrect description"
it is stated in the example section "Because the use of async and await in an application entry point is not supported, we cannot apply the async attribute to the Main method" but the example clearly shows an async task main.

The use of async/await is now supported (in DotNetCore at least) and the example shows a correct usage so this statement should be revised.

---
#### Document Details

* ID: 61f81774-3755-817c-edb0-8f3b36ddcc92
* Version Independent ID: 8ed3997b-3dd6-fcfd-67ad-6426e2292eb6
* Content: [await (C# Reference)](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/await)
* Content Source: [docs/csharp/language-reference/keywords/await.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/keywords/await.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

pkulikov:
Fixes #9349


docs/csharp/language-reference/keywords/await.md
```diff
@@ -59,7 +59,7 @@ The following example returns the total number of characters in the pages whose
...
+The preceding example uses C# 7.1, which supports the [`async` `Main` method](../../programming-guide/main-and-command-args/index.md). Because earlier C# versions don't support the use of `async` and `await` in an application entry point, you cannot apply the `async` modifier to the `Main` method, nor can you await the `GetPageLengthsAsync` method call. In that case, you can ensure that the `Main` method waits for the async operation to complete by retrieving the value of the <xref:System.Threading.Tasks.Task%601.Result?displayProperty=nameWithType> property. For tasks that do not return a value, you can call the <xref:System.Threading.Tasks.Task.Wait%2A?displayProperty=nameWithType> method. For information about how to select the language version, see [Select the C# language version](../configure-language-version.md).
```
BillWagner: It would be more accurate to say that "earlier C# versions don't support application entry points that return `Task` or `Task<T>`. If you can make that one small change, I'll :shipit:.

BillWagner (changes requested):
Thanks for spotting this and making this change @pkulikov 

I am requesting one small fix to make the change more accurate, and then it's ready to merge.

docs/csharp/language-reference/keywords/await.md
pkulikov: @BillWagner thanks for review; I've addressed your feedback.

BillWagner (approved):

PR #9378, opened 2018-12-04
fixing 404 due to tilde not being interpolated

gortok:
## Summary
The link to the create custom template tutorial would 404 due to the tilde.  Not sure Github appropriately interpolates that?  Also not sure that my addition is the 'right' URL format; only that it works?


rpetrusha:
Thanks for contributing to the dotnet/docs repo, @gortok. The tilde is an shortcut for an absolute link from the dotnent/docs repo. It is expected to render in the article on [docs.microsoft.com](https://docs.microsoft.com/en-us/dotnet/core/tools/custom-templates), but not in GitHub. There are certainly hundreds, if not thousands, occurrences of such tildes throughout our doc set.

Since the link renders successfully in docs.microsoft.com, I'll close your PR.

PR #9380, opened 2018-12-04
C53994: "|" is breaking formatting

v-makoud:
Hello, @cartermp, @mairaw,
Many thanks in advance.

## Summary

Describe your changes here.


cartermp (changes requested):
This change turns the symbol into raw HTML:
image

mairaw:
This is not a source issue but a translation issue so I'll be closing. FYI @JasonCard @nokura 

See how the translated version doesn't contain the first backtick and is using a single quote instead:
image

I've opened https://github.com/dotnet/docs.pt-br/pull/98 to fix this for pt-br as an example.

JasonCard:
@mairaw I also looked into this and it appears to be MD transformer issue rather than source issue. I'll follow up on this bug and agree for this PR to be closed. cc @v-makoud @nokura

v-maudel:
Thanks @mairaw and @JasonCard

PR #9417, opened 2018-12-05
Update cs1626.md

### LINKED ISSUE #8601: Compiler Error CS1626: mistake in the docs?
Compiler Error CS1626 states: "A yield statement is not allowed in a try block if there is a catch clause associated with the try block. To avoid this error, move the yield statement out of the catch clause."
This does not make sense: if the problematic yield statement is in a TRY block, then how can it be moved out of a CATCH?  I think the writer confused the words TRY and CATCH.  (The code sample seems to confirm that this is a mistake.)

---
#### Document Details

* ID: 6c0911c4-db28-56c2-9e10-db98404bca40
* Version Independent ID: 0822b9c2-3b52-0384-5991-f3e2f00a45b0
* Content: [Compiler Error CS1626](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs1626)
* Content Source: [docs/csharp/misc/cs1626.md](https://github.com/dotnet/docs/blob/master/docs/csharp/misc/cs1626.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

brian-fors:
## Summary

Updated error description and added finally{} block to provide context for option of removing the catch block to resolve the compile error

Fixes #8601

### DIFF
--- a/docs/csharp/misc/cs1626.md
+++ b/docs/csharp/misc/cs1626.md
@@ -10,7 +10,7 @@ ms.assetid: 3ba03383-eb24-4fd8-bf40-8b0f7d6baf0d
 # Compiler Error CS1626
 Cannot yield a value in the body of a try block with a catch clause  
   
- A yield statement is not allowed in a try block if there is a catch clause associated with the try block. To avoid this error, either move the yield statement out of the try block or remove the catch block. 
+ A yield statement is not allowed in a try block if there is a catch clause associated with the try block. To avoid this error, either move the yield statement out of the try/catch/finally block, or remove the catch block.
   
  The following sample generates CS1626:  
   
--- a/docs/csharp/misc/cs1626.md
+++ b/docs/csharp/misc/cs1626.md
@@ -10,7 +10,7 @@ ms.assetid: 3ba03383-eb24-4fd8-bf40-8b0f7d6baf0d
 # Compiler Error CS1626
 Cannot yield a value in the body of a try block with a catch clause  
   
- A yield statement is not allowed in a try block if there is a catch clause associated with the try block. To avoid this error, move the yield statement out of the catch clause.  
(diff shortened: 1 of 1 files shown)


docs/csharp/misc/cs1626.md
```diff
@@ -10,7 +10,7 @@ ms.assetid: 3ba03383-eb24-4fd8-bf40-8b0f7d6baf0d
 # Compiler Error CS1626
 Cannot yield a value in the body of a try block with a catch clause  
   
- A yield statement is not allowed in a try block if there is a catch clause associated with the try block. To avoid this error, move the yield statement out of the catch clause.  
+ A yield statement is not allowed in a try block if there is a catch clause associated with the try block. To avoid this error, either move the yield statement out of the try block or remove the catch block.
```
rpetrusha: The question here is where the yield statement can be moved. It can't be moved to the catch block (generates CS1631: Cannot yield a value in the body of a catch clause) or to the finally block (generates CS1625: Cannot yield in the body of a finally clause). I'd rephrase this as: "move the yield statement out of the try/catch/finally block, or remove the catch block.

rpetrusha (changes requested):
Thanks for noticing the error, @brian-fors, and opening this PR to correct. I've suggested an alternate wording for the first workaround.

docs/csharp/misc/cs1626.md
brian-fors: Yes, that is much clearer.  Thanks for reviewing and suggesting, I will make the change.

brian-fors: @rpetrusha I have completed the change.  Please take a look when you have time.

rpetrusha (approved):

rpetrusha:
Thanks for making the additional changes, @brian-fors. I'll merge your PR now. The changes should be live on docs.microsoft.com in the next day or two.

Issue #8882, opened 2018-11-06
Move Resources documentation to .NET Guide


rpetrusha:
## Move Resources documentation to .NET Guide

[Resources in .NET Apps](https://docs.microsoft.com/dotnet/framework/resources/) and its child articles are currently located in the .NET Framework guide. However, support for resources is largely identical across .NET implementations. This content:

- Should be moved to the .NET Guide. This involves moving the topics, as well as redirecting from the old URLs to the new URLs.
- Revising existing topics to replace references to the .NET Framework with references to .NET.

#9421 is a closed PR that incorporates these changes; it was closed because it will be easier to create a new PR than to address merge conflicts. The changes in #9421 should be reintroduced in a new PR.

See also #8881

PR #9421, opened 2018-12-05
Changed resources article title

rpetrusha:
## Changed resources article title

After merging #8920, I realized that we hadn't changed the title to reflect that the article covers both .NET Framework and .NET Core.


jzeferino (approved):

mairaw (approved):
LGTM. I'm guessing this will move to a new place later?

rpetrusha:
Yes, @mairaw, it's #8882, which I've scheduled for Sprint 147.

PR #9430, opened 2018-12-05
Clarify the recommendations on try / catch and try / finally

### LINKED ISSUE #7327: "Use try/catch/finally blocks around code that can potentially generate an exception" - bad advice
IMHO the first advice that is mentioned: "Use try/catch/finally blocks around code that can potentially generate an exception" is a bad one. Instead, I would say: "Don't catch exceptions if you don't have something useful to do with them".
One of the most common problems that I see in junior's code is that they wrap everything with try/catch because they're "afraid" that exceptions will be thrown from their code. This often makes them swallow exceptions. Even if they add a log entry in the catch block, but don't re-throw it, then practically it's pretty much the same as swallowing it. And even if they rethrow it, they either don't do it correctly (using `ExceptionDispatchInfo`, or at least `throw;`), or it's just redundant and makes the code more cumbersome. After all, if there's really an unexpected exception being thrown, then you would probably want the application to crash and display the entire exception details, which is what happens anyway.

---
#### Document Details

* ID: b1ea31f6-516f-100d-046e-1a0445d92824
* Version Independent ID: a713e8c9-8210-761d-9ae7-238793825ff3
* Content: [Best Practices for Exceptions](https://docs.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions)
* Content Source: [docs/standard/exceptions/best-practices-for-exceptions.md](https://github.com/dotnet/docs/blob/master/docs/standard/exceptions/best-practices-for-exceptions.md)
* Product: **dotnet**
* GitHub Login: @mairaw
* Microsoft Alias: **mairaw**

BillWagner:
Also, formatting to code fence `null`.
Code formatting (there were hard tabs)
Finally, change header to be more explicit about restoring state when methods exit because of exceptions.

Fixes #7327

### DIFF
--- a/docs/standard/exceptions/best-practices-for-exceptions.md
+++ b/docs/standard/exceptions/best-practices-for-exceptions.md
@@ -1,6 +1,6 @@
 ---
 title: "Best Practices for Exceptions"
-ms.date: "03/30/2017"
+ms.date: "12/05.2018"
 ms.technology: dotnet-standard
 dev_langs: 
   - "csharp"
@@ -9,20 +9,16 @@ dev_langs:
 helpviewer_keywords: 
   - "exceptions, best practices"
 ms.assetid: f06da765-235b-427a-bfb6-47cd219af539
-author: "mairaw"
-ms.author: "mairaw"
 ---
 # Best practices for exceptions
 
 A well-designed app handles exceptions and errors to prevent app crashes. This section describes best practices for handling and creating exceptions.
 
-## Use try/catch/finally blocks
+## Use try/catch/finally blocks to recover or clean resources
 
-Use `try`/`catch`/`finally` blocks around code that can potentially generate an exception. 
+Use `try`/`catch` blocks around code that can potentially generate an exception ***and*** your code can recover from that exception. In `catch` blocks, always order exceptions from the most specific to the least specific. When your code cannot recover from an exception, don't catch that exception. Enable methods further up the call stack to recover if possible.
 
-In `catch` blocks, always order exceptions from the most specific to the least specific.
-
-Use a `finally` block to clean up resources, whether you can recover or not.
+Clean up resources allocated with either `using` statements, or `try`/`finally` blocks. Prefer `using` statements to automatically clean up resources when exceptions are thrown. Use  `try`/`finally` blocks to clean up resources that don't implement <xref:System.IDisposable>.
 
 ## Handle common conditions without throwing exceptions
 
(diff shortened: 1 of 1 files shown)


docs/standard/exceptions/best-practices-for-exceptions.md
```diff
@@ -9,20 +9,16 @@ dev_langs:
...
 ms.assetid: f06da765-235b-427a-bfb6-47cd219af539
-author: "mairaw"
-ms.author: "mairaw"
 ---
 # Best practices for exceptions
 
 A well-designed app handles exceptions and errors to prevent app crashes. This section describes best practices for handling and creating exceptions.
 
-## Use try/catch/finally blocks
+## Use try/catch/finally blocks to recover or clean resources
```
rpetrusha: clean resources? perhaps free resources or release resources?

docs/standard/exceptions/best-practices-for-exceptions.md
```diff
@@ -9,20 +9,16 @@ dev_langs:
...
 
-Use `try`/`catch`/`finally` blocks around code that can potentially generate an exception. 
+Use `try`/`catch` blocks around code that can potentially generate an exception ***and*** your code can recover from that exception. In `catch` blocks, always order exceptions from the most specific to the least specific. When your code cannot recover from an exception, don't catch that exception. Enable methods further up the call stack to recover if possible.
```
rpetrusha: most specific to the least specific: "specific" is unclear. For example, is FileNotFoundException more or less specific that OverflowException? I think that it's better to say "most derived to least derived", and to add a brief explanation (all exceptions are derived from Exception, and the more derived exceptions are not handled if they're preceded by base class exceptions).

docs/standard/exceptions/best-practices-for-exceptions.md
```diff
@@ -9,20 +9,16 @@ dev_langs:
...
 
-In `catch` blocks, always order exceptions from the most specific to the least specific.
-
-Use a `finally` block to clean up resources, whether you can recover or not.
+Clean up resources allocated with either `using` statements, or `try`/`finally` blocks. Prefer `using` statements to automatically clean up resources when exceptions are thrown. Use  `try`/`finally` blocks to clean up resources that don't implement <xref:System.IDisposable>.
```
rpetrusha: isn't it preferable to describe this as simply a finally block, rather than a try/finally block? And perhaps note that the finally block is (generally) always executed.

rpetrusha (approved):
This is a major improvement, @BillWagner. I've left a number of suggestions. You can merge when you're ready.

docs/standard/exceptions/best-practices-for-exceptions.md
BillWagner: Great choice. Thanks.

PR #9450, opened 2018-12-06
SEO adjustments for .NET Core Docker

Thraka:
Contributes to #9440

### DIFF
--- a/docs/core/docker/index.md
+++ b/docs/core/docker/index.md
@@ -1,4 +1,4 @@
 ---
 redirect_url: intro-net-docker
-ms.custom: "mvc", "seodec18"
+ms.custom: "mvc, seodec18"
 ---
--- a/docs/core/docker/index.md
+++ b/docs/core/docker/index.md
@@ -1,18 +1,4 @@
 ---
-title: Docker and .NET Core
-description: Learn about using Docker with .NET Core using these fun tutorials.
-author: jralexander
-ms.author: johalex
-ms.date: 01/10/2018
+redirect_url: intro-net-docker
+ms.custom: "mvc", "seodec18"
 ---
-
-# Docker and .NET Core 
-
-The following tutorials are available for learning about using Docker with .NET Core.
-
-- [Introduction to .NET and Docker](intro-net-docker.md)
-- [Learn Docker Basics with .NET Core](docker-basics-dotnet-core.md)
-- [Building Docker Images for .NET Core Applications](building-net-docker-images.md)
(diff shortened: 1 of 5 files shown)


Thraka:
@BillWagner I did an index change akin to what you did in a diff PR. I'm curious how it renders

docs/core/docker/index.md
```diff
@@ -1,18 +1,4 @@
 ---
-title: Docker and .NET Core
-description: Learn about using Docker with .NET Core using these fun tutorials.
-author: jralexander
-ms.author: johalex
-ms.date: 01/10/2018
+redirect_url: intro-net-docker
```
BillWagner: Something is wrong here. It's not redirecting on the review site. It's showing me the empty page.

docs/core/docker/intro-net-docker.md
```diff
@@ -1,11 +1,12 @@
 ---
-title: Introduction to .NET and Docker
-description: Understanding Docker and .NET Core
+title: Introduction to Docker - .NET Core
+description: This article provides an introduction and overview to Docker in the context of a .NET Core application.
 author: jralexander
 ms.author: johalex
 ms.date: 11/06/2017
-ms.custom: mvc
+ms.custom: "mvc", "seodec18"
```
BillWagner: This should be:

```yaml
ms.custom: "mvc, seodec18"
```

It's not really YAML. It's a string.

BillWagner (changes requested):
The redirect isn't working correctly. We need to get that fixed before we ship it.

docs/core/docker/intro-net-docker.md
Thraka: Doug Erickson said in the chat on the SEO training that you need to use multiple strings separated with a , to prevent it from being seen as a single string (they want multiple string values)

BillWagner: Yes, but then he and Carl Rabeler investigated. It needs to be a single YAML string with the values separated by commas.

Thraka: I ran it though an online yaml validator and it hated the `", "` syntax.. So you're probably right. I'll fix it up 😄

docs/core/docker/index.md
Thraka: Could be because of the invalid yaml header. Hold on I'll fix that and retest

BillWagner (approved):

BillWagner:
@Thraka Fixing the header issues fixed the redirection. :shipit: when ready.

PR #9484, opened 2018-12-07
Remove confusing recommendation re: default globs

chwarr:
Removes a confusing recommendation about removing default globs.

* The preceding section outlines the two ways that the "Duplicate
  Compile items were included." can be resolved.
* Only projects that were migrated from project.json will have globs in
  the .csproj. Older .csproj files will have explicit <Compile> items
  for each file.


mairaw (approved):
Since @dsplaisted already approved this change offline, I'm approving this here. Thanks!

PR #9494, opened 2018-12-08
Update F# namespaces doc a bit

### LINKED ISSUE #2973: F# Documentation on namespaces and namespace syntax lacks some details
*Originally reported in visualfsharp repo: https://github.com/Microsoft/visualfsharp/issues/3485.*

When I found out that a file can contain multiple namespaces (I have the policy of having one namespace per file) I [looked up the documentation](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/namespaces) and found a few key points missing:

* The syntax does not include the syntax for qualified modules, which can also declare a namespace (though this is explained later on, the subject is namespaces and the syntax should be complete)
* The main syntax bulb should show the full syntax, or show two variants (pre and post 4.1), my preference is to just show the new syntax in full and add a remark that the `rec` keyword is supported from F# 4.1.
* The introduction says that once you have a namespace declaration, the entire file will be in that namespace. This suggests that there can be only namespace declaration per file and raises the wrong assumptions for readers not going through the all sections: it is in fact only until the next namespace declaration (a remark can link to the section on nested namespaces).
* I wouldn't object, or in fact, welcome, a clear enumeration on the syntax peculiarities, right after the intro-line under Remarks, for instance:
    * Namespaces cannot be nested (link to section on nested namespaces)
    * Namespaces can be preceded by comments, but XML doc or simple documentation comments are allowed but ignored
    * Namespaces can be preceded by compiler directives
    * Namespaces can be recursive (link to that section)
    * Namespaces need not be unique, they can span multiple files or even projects
    * A qualified module declaration (i.e., implicit namespace) cannot have other namespace or module declarations at the top-level
    * A nested module cannot have an implicit namespace

I'd be happy to fix this myself (provided there's some consensus here). I'd just need a pointer to where I find this in the source (sorry, it'll be my first commit, a little new to that).

cartermp:
Fixes #2973

* Addresses issues
* Clarifies a few points made in the issue

### DIFF
--- a/docs/fsharp/language-reference/namespaces.md
+++ b/docs/fsharp/language-reference/namespaces.md
@@ -111,7 +111,7 @@ module BananaHelpers =
         | Down -> b |> peelSides
 ```
 
-Note that the exception `DontSqueezeTheBananaException` and the class `Banana` both refer to each other.  Additionally, the module `BananaHelpers` and the class `Banana` also refer to each other.  This would not be possible to express in F# if you removed the `rec` keyword from the `MutualReferences` namespace.
+Note that the exception `DontSqueezeTheBananaException` and the class `Banana` both refer to each other.  Additionally, the module `BananaHelpers` and the class `Banana` also refer to each other. This wouldn't be possible to express in F# if you removed the `rec` keyword from the `MutualReferences` namespace.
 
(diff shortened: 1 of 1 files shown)


mairaw (approved):
LGTM. Left some comments for you to consider.

PR #9499, opened 2018-12-09
Fix typo

lindexi:
## Summary

We need to add space between two word


mairaw (approved):
Thank you @lindexi for adding the missing space. Your contribution should go live within a day or so.

PR #9518, opened 2018-12-10
Csharp branding title suffix

BillWagner:
The C# reference and C# programming guide both had a suffix in parentheses, not following the "- C# reference" or "- C# Programming Guide" format.

The advantage of that was that this is a simple change.
The disadvantage is that it's unclear how much of an effect that will have on SEO.


rpetrusha (approved):
LGTM, @BillWagner. This is ready to merge.

PR #9537, opened 2018-12-11
Fix capitalization of MSBuild

kkm000:
## Summary
Capitalize MSBuild properly.


BillWagner (approved):
Thanks for spotting this and submitting a fix @kkm000  We appreciate it.

I'll :shipit: now.  Thanks again for contributing to docs!

PR #9564, opened 2018-12-12
remove duplicate core version history

BillWagner:
It's outdated, and doesn't help most customers.

/cc @KathleenDollard @Thraka


docs/core/versions/index.md
```diff
@@ -10,7 +10,7 @@ ms.custom: "seodec18"
 
 .NET Core refers to the .NET Core Runtime and the .NET Core SDK, which contains the tools you need to develop applications. .NET Core SDKs are designed to work with any previous version of the .NET Core Runtime. This article explains the runtime and the SDK version strategy. An explanation of version numbers for .NET Standard can be found in the article introducing [.NET Standard](../../standard/net-standard.md#net-implementation-support).
 
-The .NET Core Runtime and .NET Core SDK add new features at a different rate - in general the .NET Core SDK provides updated tools more quickly than the .NET Core Runtime changes the runtime you use in production. Unfortunately, this problem has resulted in several versioning strategies over the last few years. You can learn about the history in the article on [.NET Core versioning](version-history.md).
+The .NET Core Runtime and .NET Core SDK add new features at a different rate - in general the .NET Core SDK provides updated tools more quickly than the .NET Core Runtime changes the runtime you use in production. Unfortunately, this problem has resulted in several versioning strategies over the last few years.
```
Thraka: nit: I think we can remove

>Unfortunately, this problem has resulted in several versioning strategies over the last few years.

as it really has no reason to be in the article now that there isn't a link to explain that statement. And, the fact that the Runtime and SDK version at different intervals isn't a "problem" in the sense the article is making it out to be. Plenty of products do that.

Thraka (approved):
Looks good to me

PR #9570, opened 2018-12-12
"length" instead of "rank"

ghost:
IMHO. because "the number of elements in the initialization list" can give only length not rank.

## Summary

"rank" word is replaced with "length" word.


rpetrusha (approved):
Thank you, @volkan-yazilim, for contributing to the dotnet/docs repo and correcting the confusion between rank and length. We'll merge your PR now. Your change should be live on docs.microsoft.com in the next day or two.

PR #9572, opened 2018-12-12
Redirected from Dispose guidelines

rpetrusha:
## Redirected from Dispose guidelines

This PR removes the Dispose pattern article from the .NET design guidelines and redirects to *Implementing a Dispose method*.


BillWagner (approved):
This LGTM @rpetrusha 

You can :shipit: when ready

Issue #9578, opened 2018-12-13
Confused


volkan-yazilim:
Even though I know how to use indexers, I got confused with the usage ot terms.
Indexer, index value, index type, indexed value, parameter,...
I believe those should be defined at the beginning. IMHO.
Thanks. 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 5323b38b-8dc5-8f50-f5e7-b58ca0e88ad7
* Version Independent ID: 3d9a83cd-98e3-fc6e-727b-32c8e9228c69
* Content: [Using Indexers - C# Programming Guide](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/indexers/using-indexers#example-2)
* Content Source: [docs/csharp/programming-guide/indexers/using-indexers.md](https://github.com/dotnet/docs/blob/master/docs/csharp/programming-guide/indexers/using-indexers.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

PR #9579, opened 2018-12-13
"indexed value" or "indexer value"

ghost:
I did understand this way. If you did mean this way, it is a reasonable change. IMHO.
https://github.com/dotnet/docs/issues/9578

## Summary

#9578


BillWagner:
@volkan-yazilim Thanks for working to make this article more clear.  I"m going to close this PR without merging it, because the original term was correct. (See this section of the [spec](https://docs.microsoft.com/dotnet/csharp/language-reference/language-specification/classes#indexers)). 

That said, regarding #9578, defining the terms "indexer", "indexer parameter" and "indexer value" would be a great addition. I'll leave that issue open, and if you want to submit a PR with that addition, we'd appreciate it.

PR #9580, opened 2018-12-13
"indexer parameter type" instead of "index type"

ghost:
I did understand this way. If you did mean this way, it is a reasonable change. IMHO.
https://github.com/dotnet/docs/issues/9578

## Summary

#9578


BillWagner (approved):
Thanks for making this change @volkan-yazilim 
The updated text is more accurate.

Thanks for contributing to docs. 🎆

PR #9596, opened 2018-12-13
Revised (in)equality operator articles

pkulikov:
Snippets are moved to the samples repo: dotnet/samples#512


BillWagner:
closing and reopening to build with new snippets

docs/csharp/language-reference/operators/equality-comparison-operator.md
```diff
@@ -11,16 +10,46 @@ helpviewer_keywords:
...
+
+The equality operator `==` returns `true` if its operands are equal, `false` otherwise.
+
+## Value types equality
+
+Operands of the [built-in value types](../keywords/value-types-table.md) are equal if their values are equal:
+
+[!code-csharp-interactive[value types equality](~/samples/snippets/csharp/language-reference/operators/EqualityAndNonEqualityExamples.cs#ValueTypesEquality)]
+
+By default, the `==` operator is not defined for a user-defined [struct](../keywords/struct.md) type. A user-defined type can [overload](#operator-overloadability) the `==` operator.
```
BillWagner: This should state that `==` is defined for enum types. That slipped through the cracks in the new (more clear) description.

BillWagner (changes requested):
Overall, this is much improved @pkulikov The language is more clear and easier to understand.

I had one comment to address, and then this is ready to merge. I don't think we need a new sample for an enum comparison. I think it would be enough to say that `==` and `!=` for `enum` types obey the same rules as the underlying integral type.

docs/csharp/language-reference/operators/equality-comparison-operator.md
pkulikov: @BillWagner indeed, as behavior for `struct`s is described, `enum` case also should be stated explicitly. Thank you for the remark. I've added the clarification, please check the latest commit.

BillWagner: I'll suggest this small word change:

"Two operands of the same [enum](../keywords/enum.md) type are equal if the ..."

It's a small change, but the text you have could be misinterpreted to support something like the following:

```csharp
    enum Color
    {
        Red,
        Green,
        Blue
    }

    enum Status
    {
        Red,
        Yellow,
        Green
    }

 bool equal = (Status.Red == Color.Red); // CS0019 (operator == cannot be applied to ...)
```

pkulikov: @BillWagner that's a good catch! Thanks. Fixed.

BillWagner (approved):

BillWagner:
Thanks @pkulikov 

This looks great. I'll :shipit: as soon as the build finishes.

PR #9603, opened 2018-12-14
Extra explanation

ghost:
To explain why removing method call does not any effect, when the confusing sentence hits.

## Summary

added an explanation for partial methods returning always void. 


rpetrusha (changes requested):
Thank you, @volkan-yazilim, for adding an important detail about the required syntax of a partial method. I thought, though, that the real problem with the section is that the phrase "similar to an event" is totally confusing (what this means, evidently, is that event syntax is similar to partial method syntax), and that the syntactical requirements of a partial method are presented much too late. As a result, your change duplicates the first bulleted list item that follows the code block. It implies that the bulleted list is presented much too late.

What I would suggest is that the bulleted list be moved to follow the "Partial methods enable the implementer..." sentence, and that that sentence also be modified to read, "Partial methods enable an implementer of one part of a class to provide a method definition with no implementation. The signature of the partial method must conform to the following guidelines:

Would you like to make those changes?

BillWagner (approved):
Thanks for spotting this and submitting the fix @volkan-yazilim We appreciate it.

I updated the wording to be consistent with the description in the C# spec, and I'll approve this PR. I'll :shipit: after the build finishes.

Thanks again for contributing. 🎆

volkan-yazilim:
Thank you. I did not have time to do what you asked Mr. @rpetrusha , even though I did want to do.

PR #9632, opened 2018-12-17
Fix typo

pkulikov:
Shame on me...


BillWagner (approved):
Thanks for fixing this @pkulikov I can't believe we all missed that in the last update.

I'll :shipit: now.

PR #9648, opened 2018-12-18
F# Async programming rewrite

### LINKED ISSUE #666: Async F# Article Re-architecting.
I'm not really happy with how I wrote this a while back.  While it gets some necessary information across, it is misleading, the section comparing to C#/VB is unnecessary, and the article doesn't describe what's going on under the covers at all.  There's also nothing about using `MailboxProcessor<'T>`, but it's a critical sibling component of the language that deserves mentioning here.  It should also get its own article for teaching how to do concurrent programming in F#, but that's another issue.  Here's what I'm thinking as an outline:
1. Start with the motivation - why write async code?  Why in F#?
2. Explain the three core components: `Async<'T>`, `async` as a special kind of computation expression, and the `Async` starting functions.
3. Simple example.
4. Example combining workflows.
5. Parallel example.
6. Example showing interop with `Task`, and `Task<T>`.
7. Explain what's going on under the covers:
   - What does the compiler transform the async workflow into?
   - How is work scheduled on a thread?  How are callbacks handled?
   - What do each of the starting functions do?
   - Thread trampolining to handle tons of calls that aren't taken as tailcalls by the runtime.  What, why, and how?
8. An aside on Agents and why they matter when writing asynchronous code.  Link to alternative documentation for now, since we don't have anything official.

cartermp:
Fixes #666 

- [x] Starting functions section
- [x] "awaiting" function section
- [x] Write .NET interop section (existing functions, converting to/from `Task`, etc.)
- [x] Cancellation and propagation of tokens

### DIFF
--- a/docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
+++ b/docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
@@ -235,7 +235,7 @@ What to watch out for:
 
 ### Async.Catch
 
-Creates an asychronous computation that executes a given `Async<'T>`, returning a `Async<Choice<'T, exn>>`. If the given `Async<'T>` completes successfully, then a `Choice1Of2` is returned with the resultant value. If it raises an exception before it completes, then a `Choice2of2` is returned with the raised exception. If it is used on an asynchronous computation that is itself composed of many computations, and one of those computations throws an exception, then encompassing computation will be stopped entirely.
+Creates an asychronous computation that executes a given `Async<'T>`, returning an `Async<Choice<'T, exn>>`. If the given `Async<'T>` completes successfully, then a `Choice1Of2` is returned with the resultant value. If it raises an exception before it completes, then a `Choice2of2` is returned with the raised exception. If it is used on an asynchronous computation that is itself composed of many computations, and one of those computations throws an exception, then encompassing computation will be stopped entirely.
 
 Signature:
 
--- a/docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
+++ b/docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
@@ -7,8 +7,8 @@ ms.date: 12/17/2018
 
 Asynchronous programming is a mechanism that is essential to modern applications for diverse reasons. There are two primary use cases that most developers will encounter:
 
-* Presenting a server process that can service a significant number of concurrent incoming requests, while minimizing the system resources occupied while request processing awaits inputs from systems or services external to that process
-* Maintaining a responsive UI or main thread while concurrently progressing background work
+- Presenting a server process that can service a significant number of concurrent incoming requests, while minimizing the system resources occupied while request processing awaits inputs from systems or services external to that process
+- Maintaining a responsive UI or main thread while concurrently progressing background work
 
 Although background work often does involve the utilization of multiple threads, it's important to consider the concepts of asynchrony and multi-threading separately. In fact, they are separate concerns, and one does not imply the other. What follows in this article will describe this in more detail.
 
@@ -16,9 +16,9 @@ Although background work often does involve the utilization of multiple threads,
 
 The previous point - that asynchrony is independent of the utilization of multiple threads - is worth explaining a bit further. There are three concepts that are sometimes related, but strictly independent of one another:
 
-* Concurrency; when multiple computations execute in overlapping time periods.
-* Parallelism; when multiple computations or several parts of a single computation run at exactly the same time.
-* Asynchrony; when one or more computations can execute separately from the main program flow.
+- Concurrency; when multiple computations execute in overlapping time periods.
+- Parallelism; when multiple computations or several parts of a single computation run at exactly the same time.
+- Asynchrony; when one or more computations can execute separately from the main program flow.
 
 All three are orthogonal concepts, but can be easily conflated, especially when they are used together. For example, you may need to execute multiple asynchronous computations in parallel. This does not mean that parallelism or asynchrony imply one another.
 
@@ -37,9 +37,9 @@ The main takeaway you should have is that asynchronous computations are independ
 
 In F#, asynchronous programming is centered around three core concepts:
 
-* The `Async<'T>` type, which represents a composable asynchronous computation.
-* The `Async` module functions, which let you schedule asynchronous work, compose asynchronous computations, and transform asynchronous results.
-* The `async { }` [computation expression](../../language-reference/computation-expressions.md), which provides a convenient syntax for building and controlling asynchronous computations.
+- The `Async<'T>` type, which represents a composable asynchronous computation.
+- The `Async` module functions, which let you schedule asynchronous work, compose asynchronous computations, and transform asynchronous results.
+- The `async { }` [computation expression](../../language-reference/computation-expressions.md), which provides a convenient syntax for building and controlling asynchronous computations.
 
 You can see these three concepts in the following example:
 
@@ -47,7 +47,7 @@ You can see these three concepts in the following example:
 open System
 open System.IO
 
-let countFileBytes path =
+let printTotalFileBytes path =
     async {
         let! bytes = File.ReadAllBytesAsync(path) |> Async.AwaitTask
         let fileName = Path.GetFileName(path)
@@ -56,14 +56,14 @@ let countFileBytes path =
 
 [<EntryPoint>]
 let main argv =
-    countFileBytes "path-to-file.txt"
+    printTotalFileBytes "path-to-file.txt"
     |> Async.RunSynchronously
 
     Console.Read() |> ignore
     0
 ```
 
-In the example, the `countFileBytes` function is of type `string -> Async<unit>`. Calling the function does not actually execute the asynchronous computation. Instead, it returns an `Async<unit>` that acts as a *specification* of the work that is to execute asynchronously. It will call `Async.AwaitTask` in its body, which will convert the result of <xref:System.IO.File.WriteAllBytesAsync%2A> to an appropriate type when it is called.
+In the example, the `printTotalFileBytes` function is of type `string -> Async<unit>`. Calling the function does not actually execute the asynchronous computation. Instead, it returns an `Async<unit>` that acts as a *specification- of the work that is to execute asynchronously. It will call `Async.AwaitTask` in its body, which will convert the result of <xref:System.IO.File.WriteAllBytesAsync%2A> to an appropriate type when it is called.
 
 Another important line is the call to `Async.RunSynchronously`. This is one of the Async module starting functions that you'll need to call if you want to actually execute an F# asynchronous computation.
 
@@ -77,7 +77,7 @@ Here is an example that builds upon the previous one by combining computations:
 open System
 open System.IO
 
-let countFileBytes path =
+let printTotalFileBytes path =
     async {
         let! bytes = File.ReadAllBytesAsync(path) |> Async.AwaitTask
         let fileName = Path.GetFileName(path)
@@ -87,7 +87,7 @@ let countFileBytes path =
 [<EntryPoint>]
 let main argv =
     argv
-    |> Array.map countFileBytes
+    |> Array.map printTotalFileBytes
     |> Async.Parallel
     |> Async.Ignore
     |> Async.RunSynchronously
@@ -98,18 +98,18 @@ let main argv =
 As you can see, the `main` function has quite a few more calls made. Conceptually, it does the following:
 
 1. Transform the command-line arguments into `Async<unit>` computations with `Array.map`.
-2. Create an `Async<'T[]>` that schedules and runs the `countFileBytes` computations in parallel when it runs.
+2. Create an `Async<'T[]>` that schedules and runs the `printTotalFileBytes` computations in parallel when it runs.
 3. Create an `Async<unit>` that will run the parallel computation and ignore its result.
 4. Explicitly run the last computation with `Async.RunSynchronously` and block until it is completes.
 
-When this program runs, `countFileBytes` runs in parallel for each command-line argument. Because asynchronous computations execute independently of program flow, there is no order in which they print their information and finish executing. The computations will be scheduled in parallel, but their order of execution is not guaranteed.
+When this program runs, `printTotalFileBytes` runs in parallel for each command-line argument. Because asynchronous computations execute independently of program flow, there is no order in which they print their information and finish executing. The computations will be scheduled in parallel, but their order of execution is not guaranteed.
 
 ## Sequencing asynchronous computations
 
 Because `Async<'T>` is a specification of work rather than an already-running task, you can perform more intricate transformations easily. Here is an example that sequences a set of Async computations so they execute one after another.
 
 ```fsharp
-let countFileBytes path =
+let printTotalFileBytes path =
     async {
         let! bytes = File.ReadAllBytesAsync(path) |> Async.AwaitTask
         let fileName = Path.GetFileName(path)
@@ -119,13 +119,13 @@ let countFileBytes path =
 [<EntryPoint>]
 let main argv =
     argv
-    |> Array.map countFileBytes
+    |> Array.map printTotalFileBytes
     |> Async.Sequential
     |> Async.RunSynchronously
     |> ignore
 ```
 
-This will schedule `countFileBytes` to execute in the order of the elements of `argv` rather than scheduling them in parallel. Because the next item will not be scheduled until after the last computation has finished executing, the computations are sequenced such that there is no overlap in their execution.
+This will schedule `printTotalFileBytes` to execute in the order of the elements of `argv` rather than scheduling them in parallel. Because the next item will not be scheduled until after the last computation has finished executing, the computations are sequenced such that there is no overlap in their execution.
 
 ## Important Async module functions
 
@@ -140,18 +140,18 @@ Starts a child computation within an asynchronous computation. This allows multi
 Signature:
 
 ```fsharp
-computation: Async<'T> * timeout: ?int -> Async<Async<'T>>
+computation: Async<'T> - timeout: ?int -> Async<Async<'T>>
 ```
 
 When to use:
 
-* When you want to execute multiple asynchronous computations concurrently rather than one at a time, but not have them scheduled in parallel.
-* When you wish to tie the lifetime of a child computation to that of a parent computation.
+- When you want to execute multiple asynchronous computations concurrently rather than one at a time, but not have them scheduled in parallel.
+- When you wish to tie the lifetime of a child computation to that of a parent computation.
 
 What to watch out for:
 
-* Starting multiple computations with `Async.StartChild` isn't the same as scheduling them in parallel. If you wish to schedule computations in parallel, use `Async.Parallel`.
-* Canceling a parent computation will trigger cancellation of all child computations it started.
+- Starting multiple computations with `Async.StartChild` isn't the same as scheduling them in parallel. If you wish to schedule computations in parallel, use `Async.Parallel`.
+- Canceling a parent computation will trigger cancellation of all child computations it started.
 
 ### Async.StartImmediate
 
@@ -160,16 +160,16 @@ Runs an asynchronous computation, starting immediately on the current operating 
 Signature:
 
 ```fsharp
-computation: Async<unit> * cancellationToken: ?CancellationToken -> unit
+computation: Async<unit> - cancellationToken: ?CancellationToken -> unit
 ```
 
 When to use:
 
-* When you need to update something on the calling thread in the middle of an asynchronous computation.
+- When you need to update something on the calling thread in the middle of an asynchronous computation.
 
 What to watch out for:
 
-* Code in the asynchronous computation will run on whatever thread one happens to be scheduled on. This can be problematic if that thread is in some way sensitive, such as a UI thread. In such cases, `Async.StartImmediate` is likely inappropriate to use.
+- Code in the asynchronous computation will run on whatever thread one happens to be scheduled on. This can be problematic if that thread is in some way sensitive, such as a UI thread. In such cases, `Async.StartImmediate` is likely inappropriate to use.
 
 ### Async.StartAsTask
 
@@ -178,16 +178,16 @@ Executes a computation in the thread pool. Returns a <xref:System.Threading.Task
 Signature:
 
 ```fsharp
-computation: Async<'T> * taskCreationOptions: ?TaskCreationOptions * cancellationToken: ?CancellationToken -> Task<'T>
+computation: Async<'T> - taskCreationOptions: ?TaskCreationOptions - cancellationToken: ?CancellationToken -> Task<'T>
 ```
 
 When to use:
 
-* When you need to call into a .NET API that expects a <xref:System.Threading.Tasks.Task%601> to represent the result of an asynchronous computation.
+- When you need to call into a .NET API that expects a <xref:System.Threading.Tasks.Task%601> to represent the result of an asynchronous computation.
 
 What to watch out for:
 
-* This call will allocate an additional `Task` object, which can increase overhead if it is used often.
+- This call will allocate an additional `Task` object, which can increase overhead if it is used often.
 
 ### Async.Parallel
 
@@ -196,18 +196,18 @@ Schedules a sequence of asynchronous computations to be executed in parallel. Th
 Signature:
 
 ```fsharp
-computations: seq<Async<'T>> * ?maxDegreesOfParallelism: int -> Async<'T[]>
+computations: seq<Async<'T>> - ?maxDegreesOfParallelism: int -> Async<'T[]>
 ```
 
 When to use it:
 
-* If you need to run a set of computations at the same time and have no reliance on their order of execution.
-* If you don't require results from computations scheduled in parallel until they have all completed.
+- If you need to run a set of computations at the same time and have no reliance on their order of execution.
+- If you don't require results from computations scheduled in parallel until they have all completed.
 
 What to watch out for:
 
-* You can only access the resulting array of values once all computations have finished.
-* Computations will be run however they end up getting scheduled. This means you cannot rely on their order of their execution.
+- You can only access the resulting array of values once all computations have finished.
+- Computations will be run however they end up getting scheduled. This means you cannot rely on their order of their execution.
 
 ### Async.Sequential
 
@@ -221,12 +221,12 @@ computations: seq<Async<'T>> -> Async<'T[]>
 
 When to use it:
 
-* If you need to execute multiple computations in order.
+- If you need to execute multiple computations in order.
 
 What to watch out for:
 
-* You can only access the resulting array of values once all computations have finished.
-* Computations will be run in the order that they are passed to this function, which can mean that more time will elapse before the results are returned.
+- You can only access the resulting array of values once all computations have finished.
+- Computations will be run in the order that they are passed to this function, which can mean that more time will elapse before the results are returned.
 
 ### Async.AwaitTask
 
@@ -240,11 +240,11 @@ task: Task<'T>  -> Async<'T>
 
 When to use:
 
-* When you are consuming a .NET API that returns a <xref:System.Threading.Tasks.Task%601> within an F# asynchronous computation.
+- When you are consuming a .NET API that returns a <xref:System.Threading.Tasks.Task%601> within an F# asynchronous computation.
 
 What to watch out for:
 
-* Exceptions are wrapped in <xref:AggregateException> following the convention of the Task Parallel Library, and this is different from how F# async generally surfaces exceptions.
+- Exceptions are wrapped in <xref:AggregateException> following the convention of the Task Parallel Library, and this is different from how F# async generally surfaces exceptions.
 
 ### Async.Catch
 
@@ -258,11 +258,11 @@ computation: Async<'T> -> Async<Choice<'T, exn>>
 
 When to use:
 
-* When you are performing asynchronous work that may fail with an exception and you want to handle that exception in the caller.
+- When you are performing asynchronous work that may fail with an exception and you want to handle that exception in the caller.
 
 What to watch out for:
 
-* When using combined or sequenced asynchronous computations, the encompassing computation will fully stop if one of its "internal" computations throws an exception.
+- When using combined or sequenced asynchronous computations, the encompassing computation will fully stop if one of its "internal" computations throws an exception.
 
 ### Async.Ignore
 
@@ -276,11 +276,11 @@ computation: Async<'T> -> Async<unit>
 
 When to use:
 
-* When you have an asynchronous computation whose result is not needed. This is analogous to the `ignore` code for non-asynchronous code.
+- When you have an asynchronous computation whose result is not needed. This is analogous to the `ignore` code for non-asynchronous code.
 
 What to watch out for:
 
-* If you must use this because you wish to use `Async.Start` or another function that requires `Async<unit>`, consider if discarding the result is okay to do. Discarding results just to fit a type signature should not generally be done.
+- If you must use this because you wish to use `Async.Start` or another function that requires `Async<unit>`, consider if discarding the result is okay to do. Discarding results just to fit a type signature should not generally be done.
 
 ### Async.RunSynchronously
 
@@ -289,17 +289,17 @@ Runs an asynchronous computation and awaits its result on the calling thread. Th
 Signature:
 
 ```fsharp
-computation: Async<'T> * timeout: ?int * cancellationToken: ?CancellationToken -> 'T
+computation: Async<'T> - timeout: ?int - cancellationToken: ?CancellationToken -> 'T
 ```
 
 When to use it:
 
-* If you need it, use it only once in an application - at the entry point for an executable.
-* When you don't care about performance and want to execute a set of other asynchronous operations at once.
+- If you need it, use it only once in an application - at the entry point for an executable.
+- When you don't care about performance and want to execute a set of other asynchronous operations at once.
 
 What to watch out for:
 
-* Calling `Async.RunSynchronously` blocks the calling thread until the execution completes.
+- Calling `Async.RunSynchronously` blocks the calling thread until the execution completes.
 
 ### Async.Start
 
@@ -308,20 +308,20 @@ Starts an asynchronous computation in the thread pool that returns `unit`. Doesn
 Signature:
 
 ```fsharp
-computation: Async<unit> * cancellationToken: ?CancellationToken -> unit
+computation: Async<unit> - cancellationToken: ?CancellationToken -> unit
 ```
 
 Use only when:
 
-* You have an asynchronous computation that doesn't yield a result and/or require processing of one.
-* You don't need to know when an asynchronous computation completes.
-* You don't care which thread an asynchronous computation runs on.
-* You don't have any need to be aware of or report exceptions resulting from the task.
+- You have an asynchronous computation that doesn't yield a result and/or require processing of one.
+- You don't need to know when an asynchronous computation completes.
+- You don't care which thread an asynchronous computation runs on.
+- You don't have any need to be aware of or report exceptions resulting from the task.
 
 What to watch out for:
 
-* Exceptions raised by computations started with `Async.Start` aren't propagated to the caller. The call stack will be completely unwound.
-* Any effectful work (such as calling `printfn`) started with `Async.Start` won't cause the effect to happen on the main thread of a program's execution.
+- Exceptions raised by computations started with `Async.Start` aren't propagated to the caller. The call stack will be completely unwound.
+- Any effectful work (such as calling `printfn`) started with `Async.Start` won't cause the effect to happen on the main thread of a program's execution.
 
 ## Interoperating with .NET
 
@@ -377,7 +377,7 @@ Although F# provides some abilities to start an asynchronous computation on the 
 
 ## See also
 
-* [The F# Asynchronous Programming Model](https://www.microsoft.com/research/publication/the-f-asynchronous-programming-model)
-* [Jet.com's F# Async Guide](https://medium.com/jettech/f-async-guide-eb3c8a2d180a)
-* [F# for fun and profit's Asynchronous Programming guide](https://fsharpforfunandprofit.com/posts/concurrency-async-and-parallel/)
-* [Async in C# and F#: Asynchronous gotchas in C#](http://tomasp.net/blog/csharp-async-gotchas.aspx/)
+- [The F# Asynchronous Programming Model](https://www.microsoft.com/research/publication/the-f-asynchronous-programming-model)
+- [Jet.com's F# Async Guide](https://medium.com/jettech/f-async-guide-eb3c8a2d180a)
+- [F# for fun and profit's Asynchronous Programming guide](https://fsharpforfunandprofit.com/posts/concurrency-async-and-parallel/)
+- [Async in C# and F#: Asynchronous gotchas in C#](http://tomasp.net/blog/csharp-async-gotchas.aspx/)
--- a/docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
+++ b/docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
@@ -23,7 +23,7 @@ Put together, "asynchronous" simply means, "not at the same time". This holds tr
 
 In other words, asynchronous computations execute independently of the main program flow. This does not imply any particular threading model, nor does it imply that a computation happens in the background. In fact, asynchronous computations could even execute synchronously, depending on the nature of the computation and the environment the computation is executing in.
 
-The main takeaway you should have is that because asynchronous computations are simply independent of the main program flow, there are few guarantees about when or how a computation will execute. This necessitates a means to control their execution. In F#, there are multiple ways to do this.
+The main takeaway you should have is that because asynchronous computations are simply independent of the main program flow, there are few guarantees about when or how a computation executes. This necessitates a means to control their execution. In F#, there are multiple ways to do this.
 
 ## Core concepts
 
@@ -59,7 +59,7 @@ In the example, the `countFileBytes` function is of type `string -> Async<unit>`
 
 The `Async.Start` function is used to start the computation.
 
-Additionally, `Async.AwaitTask` is called in the `countFileBytes` function. This creates an `Async<'T>` that runs <xref:File.ReadAllBytesAsync> (which returns a `Task<byte[]>`) and returns its result.
+Additionally, `Async.AwaitTask` is called in the `countFileBytes` function. This creates an `Async<'T>` that runs <xref:System.IO.File.ReadAllBytesAsync> (which returns a `Task<byte[]>`) and returns its result.
 
 This is a key difference with the C#/VB style of `async` programming. In F#, asynchronous computations can be thought of as **Cold tasks**. They must be explicitly started to actually execute. This has some advantages, as it allows you to combine and sequence asynchronous work very easily.
 
@@ -97,7 +97,7 @@ As you can see, the `main` function has quite a few more calls made. Conceptuall
 3. Create an `Async<unit>` that will run the parallel computation and ignore its result
 4. Explicitly run the last computation with `Async.Start`
 
-When this program is ran, `countFileBytes` will be ran in parallel for each command line argument. Because asynchronous computations execute independently of program flow, there is no order in which they will print their information and finish executing.
+When this program is ran, `countFileBytes`is ran in parallel for each command line argument. Because asynchronous computations execute independently of program flow, there is no order in which they print their information and finish executing.
 
 ## Sequencing asynchronous computations
 
@@ -132,15 +132,15 @@ let main argv =
     |> ignore
 ```
 
-This will execute `countFileBytes` in order rather than executing them in parallel. This may be preferable if you wish to preserve the order of computations.
+This will execute `countFileBytes` for each item in `argv` in the order of the elements of `argv` rather than executing them in parallel. This may be preferable if you wish to preserve the order of computations.
 
 ## Common async starting functions
 
 Because F# asynchronous computations are a _specification_ of work rather than a representation of work that is already executing, they must be explicitly started with a starting function. There are many [Async starting functions](https://msdn.microsoft.com/library/ee370232.aspx) that are helpful in different contexts. You should take care in which to use, making sure that it is appropriate for your particular environment. The following section describes some of the more common starting functions.
 
-### `Async.Start`
+### Async.Start
 
-Starts an asynchronous computation with no return type in the thread pool. Does not wait for its result. Nested computations ran with `Async.Start` are independent of the parent computation that called them. If the parent computation is cancelled, the nested computations are not cancelled.
+Starts an asynchronous computation with no return type in the thread pool. Doesn't wait for its result. Nested computations ran with `Async.Start` are independent of the parent computation that called them. If the parent computation is cancelled, the nested computations aren't cancelled.
 
 Signature:
 
@@ -156,10 +156,10 @@ When to use:
 
 What to watch out for:
 
-* Exceptions raised by computations ran with `Async.Start` are not propagated to the caller
-* Any effectful work (such as calling `printfn`) ran with `Async.Start` will not cause the effect to happen on the main thread of a program's execution
+* Exceptions raised by computations ran with `Async.Start` aren't propagated to the caller
+* Any effectful work (such as calling `printfn`) ran with `Async.Start` won't cause the effect to happen on the main thread of a program's execution
 
-### `Async.StartChild`
+### Async.StartChild
 
 Starts a child computation within an asynchronous computation. This allows multiple asynchronous computations to be executed simultaneously. The child computation shares a cancellation token with the parent computation. When the parent is cancelled, the child is also cancelled.
 
@@ -176,10 +176,10 @@ When to use:
 
 What to watch out for:
 
-* Starting multiple computations with `Async.StartChild` is not the same as scheduling them to run in parallel
+* Starting multiple computations with `Async.StartChild` isn't the same as scheduling them to run in parallel
 * Cancelling a parent computation will cancel all children it started
 
-### `Async.StartImmediate`
+### Async.StartImmediate
 
 Runs an asynchronous computation, starting immediately on the current operating system thread. This is helpful if you need to update something on the calling thread during the computation. For example, if an asynchronous computation must update a UI (such as updating a progress bar), then `Async.StartImmediate` should be used.
 
@@ -195,11 +195,11 @@ When to use:
 
 What to watch out for:
 
-* Code in the asynchronous computation will run on the caller's thread, so if you need this thread to perform other work, `Async.StartImmediate` is likely not an appropriate starting function
+* Code in the asynchronous computation will run on the caller's thread, so if you need this thread to perform other work, `Async.StartImmediate` isn't likely an appropriate starting function
 
-### `Async.StartAsTask`
+### Async.StartAsTask
 
-Executes a computation in the thread pool. Returns a <xref:System.Threading.Task`1> that will be completed in the corresponding state once the computation terminates (produces the result, throws exception or gets canceled). If no cancellation token is provided, then the default cancellation token is used.
+Executes a computation in the thread pool. Returns a <xref:System.Threading.Task%601> that will be completed in the corresponding state once the computation terminates (produces the result, throws exception or gets canceled). If no cancellation token is provided, then the default cancellation token is used.
 
 Signature:
 
@@ -209,13 +209,13 @@ computation: Async<'T> * taskCreationOptions: ?TaskCreationOptions * cancellatio
 
 When to use:
 
-* When you need to call into a .NET API that expects a <xref:System.Threading.Task`1> to represent the result of an asynchronous computation
+* When you need to call into a .NET API that expects a <xref:System.Threading.Task%601> to represent the result of an asynchronous computation
 
 What to watch out for:
 
 * This call will allocate an additional `Task` object, which can increase overhead if it is used often
 
-### `Async.RunSynchronously`
+### Async.RunSynchronously
 
 Runs an asynchronous computation and awaits its result on the calling thread. This call is blocking.
 
@@ -228,7 +228,7 @@ computation: Async<'T> * timeout: ?int * cancellationToken: ?CancellationToken -
 When to use:
 
 * Only once at the entry point for an executable
-* When you do not care about performance and want to execute a set of other asynchronous operations, such as the previous example using `Async.Parallel`
+* When you don't care about performance and want to execute a set of other asynchronous operations, such as the previous example using `Async.Parallel`
 
 What to watch out for:
 
@@ -236,7 +236,14 @@ What to watch out for:
 
 ## A Note on Threads
 
-The phrase "on another thread" is mentioned above, but it is important to know that **this does not mean that async computations are a facade for multithreading**. The computation can actually "jump" between threads, borrowing them for a small amount of time to do useful work. When an async computation is effectively "waiting" (for example, waiting for a network call to return something), any thread it was borrowing at the time is freed up to go do useful work on something else. This allows async computations to utilize the system they run on as effectively as possible.
+Although threading is mentioned throughout this article, there are two important things to remember:
+
+1. There is no affinity between an asynchronous computation and a thread, unless explicitly started on the current thread
+1. Asynchronous programming in F# is not an abstraction for multithreading
+
+
+
+but it is important to know that **this doesn't mean that async computations are a facade for multithreading**. A computation can actually "jump" between threads, borrowing them for a small amount of time to do useful work. When an async computation is effectively "waiting" (for example, waiting for a network call to return something), any thread it was borrowing at the time is freed up to go do useful work on something else. This allows async computations to utilize the system they run on as effectively as possible.
 
 ## Further reading
 
--- a/docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
+++ b/docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
@@ -319,24 +319,16 @@ let computationForCaller param =
 
 ### How to work with .NET async and Task
 
-To work with <xref:System.Threading.Tasks.Task> types (i.e., .NET async computations that have no return value) requires a bit more work. These helper functions are up to the task (pun _definitely_ intended):
+To work with APIs that use <xref:System.Threading.Tasks.Task> (i.e., .NET async computations that have no return value), you may need to add an additional function that will convert an `Async<'T>` to a <xref:System.Threading.Tasks.Task>:
 
 ```fsharp
 module Async =
     // Async<unit> -> Task
     let startTaskFromAsyncUnit (comp: Async<unit>) =
         Async.StartAsTask comp :> Task
-
-    // Task -> Async<unit>
-    let awaitTaskToAsyncUnit (task: Task) =
-        task.ContinueWith(fun t -> ()) |> Async.AwaitTask
 ```
 
-You can then use them like so:
-
-```fsharp
-// TODO examples
-```
+There is already an `Async.AwaitTask` that accepts a <xref:System.Threading.Tasks.Task> as input. With this and the previously-defined `startTaskFromAsyncUnit` function, you can start and await <xref:System.Threading.Tasks.Task> types from an F# async computation.
 
 ## Relationship to multithreading
 
--- a/docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
+++ b/docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
@@ -1,11 +1,11 @@
 ---
 title: Async Programming in F#
-description: Learn how F# async programming is accomplished via a language-level programming model that utilizes core functional programming concepts to allow for easy use of asynchrony in F# programming.
+description: Learn how F# provides clean support for asynchrony based on a language-level programming model derived from core functional programming concepts.
 ms.date: 12/17/2018
 ---
 # Async Programming in F\#
 
-Asynchronous programming is essential to modern applications. The applications for asynchronous programming are large in number. There are two primary use cases that most developers will encounter:
+Asynchronous programming is essential to modern applications. The applications for asynchronous programming are large in number. There are two primary use cases that most developers end up having:
 
 * Having a non-blocking server process that can service incoming requests while awaiting results for other requests
 * Having a responsive UI or main thread while background work is happening
@@ -19,11 +19,11 @@ If you consider the etymology of the word "asynchronous", there are two pieces:
 * "a", meaning "not"
 * "synchronous", meaning "at the same time"
 
-Put together, "asynchronous" simply means, "not at the same time". This holds true for asynchronous programming in F#. Code that executes asynchronously is code that does not necessarily execute at the same time.
+Put together, "asynchronous" simply means, "not at the same time". This holds true for asynchronous programming in F#. Code that executes asynchronously is code that doesn't necessarily execute at the same time.
 
-In other words, asynchronous computations execute independently of the main program flow. This does not imply any particular threading model, nor does it imply that a computation happens in the background. In fact, asynchronous computations could even execute synchronously, depending on the nature of the computation and the environment the computation is executing in.
+Asynchronous computations execute independently of the main program flow. This doesn't imply any particular threading model, nor does it imply that a computation happens in the background. In fact, asynchronous computations could even execute synchronously, depending on the nature of the computation and the environment the computation is executing in.
 
-The main takeaway you should have is that because asynchronous computations are simply independent of the main program flow, there are few guarantees about when or how a computation executes. This necessitates a means to control their execution. In F#, there are multiple ways to do this.
+The main takeaway you should have is that because asynchronous computations are independent of the main program flow. There are few guarantees about when or how a computation executes, which necessitates a means to control their execution. In F#, there are multiple ways to do this.
 
 ## Core concepts
 
@@ -61,7 +61,7 @@ The `Async.Start` function is used to start the computation.
 
 Additionally, `Async.AwaitTask` is called in the `countFileBytes` function. This creates an `Async<'T>` that runs <xref:System.IO.File.WriteAllBytesAsync%2A> (which returns a `Task<byte[]>`) and returns its result.
 
-This is a key difference with the C#/VB style of `async` programming. In F#, asynchronous computations can be thought of as **Cold tasks**. They must be explicitly started to actually execute. This has some advantages, as it allows you to combine and sequence asynchronous work very easily.
+This is a key difference with the C#/VB style of `async` programming. In F#, asynchronous computations can be thought of as **Cold tasks**. They must be explicitly started to actually execute. This has some advantages, as it allows you to combine and sequence asynchronous work easily.
 
 ## Combining asynchronous computations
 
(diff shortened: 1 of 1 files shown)


cartermp:
Hey @bartelink and @eiriktsarpalis, I'd love if you could tear this one up a bit before I consider merging it 🙂

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
 ---
-title: Async Programming
-description: Learn how F# async programming is accomplished via a language-level programming model that is easy to use and natural to the language.
-ms.date: 06/20/2016
+title: Async Programming in F#
+description: Learn how F# async programming is accomplished via a language-level programming model that utilizes core functional programming concepts to allow for easy use of asynchrony in F# programming.
```
bartelink: Learn how F# provides clean support for asynchrony based on a language-level programming model derived from core functional programming concepts.

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+
+* This function blocks the calling thread, which undoes the use of asynchrony if it is used to only execute a single `Async<'T>`
+
+## Interoperating with .NET
+
+You may be working with a library or C# codebase that uses [async/await](../../../standard/async.md)-style asynchronous programming. Because these use the <xref:System.Threading.Tasks.Task%601> and <xref:System.Threading.Tasks.Task> types as their core abstractions rather than `Async<'T>`, you must cross a boundary between these two approaches to asynchrony.
+
+### How to work with .NET async and Task<T>
+
+Working with .NET async libraries and codebases that use <xref:System.Threading.Tasks.Task%601> (i.e., async computations that have return values) is straightforward and has built-in support with F#.
```
mairaw:
```suggestion
Working with .NET async libraries and codebases that use <xref:System.Threading.Tasks.Task%601> (that is, async computations that have return values) is straightforward and has built-in support with F#.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
 ---
 # Async Programming in F# #
 
-> [!NOTE]
-> Some inaccuracies have been discovered in this article.  It is being rewritten.  See [Issue #666](https://github.com/dotnet/docs/issues/666) to learn about the changes.
+Asynchronous programming is essential to modern applications. The applications for asynchronous programming are large in number. There are two primary use cases that most developers will encounter:
 
-Async programming in F# can be accomplished through a language-level programming model designed to be easy to use and natural to the language.
+* Having a non-blocking server process that can service incoming requests while awaiting results for other requests
+* Having a responsive UI or main thread while background work is happening
```
bartelink: Maintaining a responsive UI or main thread while concurrently progressing background work

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+
+1. There is no affinity between an asynchronous computation and a thread, unless explicitly started on the current thread
+1. Asynchronous programming in F# is not an abstraction for multithreading
+
+What this means is that **asynchronous computations are not an abstraction for multithreading**. For example, a computation may actually run on the its caller's thread, depending on the nature of the work. A computation could also "jump" between threads, borrowing them for a small amount of time to do useful work in between periods of "waiting" (such as when a network call is in transit).
+
+Although F# provides some abilities to start an asynchronous computation on the current thread (or explicitly not on the current thread), asynchrony generally is not associated with a particular threading strategy.
 
-## Further resources:
+## Further reading
```
mairaw: More common:

```suggestion
## See also
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+
+As you can see, the `main` function has quite a few more calls made. Conceptually, it does the following:
+
+1. Transform the command line arguments into `Async<unit>` computations with `Array.map`
+2. Create an `Async<'T[]>` that will schedule and run the `countFileBytes` computations in parallel when it is ran
+3. Create an `Async<unit>` that will run the parallel computation and ignore its result
+4. Explicitly run the last computation with `Async.Start`
 
-2.  `Async.Start` will start an async workflow on another thread, and will **not** await its result.
+When this program is ran, `countFileBytes`is ran in parallel for each command line argument. Because asynchronous computations execute independently of program flow, there is no order in which they print their information and finish executing.
```
mairaw:
```suggestion
When this program runs, `countFileBytes` runs in parallel for each command line argument. Because asynchronous computations execute independently of program flow, there is no order in which they print their information and finish executing.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+ms.date: 12/17/2018
 ---
 # Async Programming in F# #
 
-> [!NOTE]
-> Some inaccuracies have been discovered in this article.  It is being rewritten.  See [Issue #666](https://github.com/dotnet/docs/issues/666) to learn about the changes.
+Asynchronous programming is essential to modern applications. The applications for asynchronous programming are large in number. There are two primary use cases that most developers will encounter:
 
-Async programming in F# can be accomplished through a language-level programming model designed to be easy to use and natural to the language.
+* Having a non-blocking server process that can service incoming requests while awaiting results for other requests
```
bartelink: Presenting a server process that can service a significant number of parallel incoming requests, while minimizing the system resources locked (reserved/goto thesaurus) while request processing awaits inputs from systems or services external to that process.

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
- Supporting cancellation of a task midway through its execution in C#/VB requires checking the `IsCancellationRequested` property or calling `ThrowIfCancellationRequested()` on a `CancellationToken` object that’s passed into the async method.
+* This call will allocate an additional `Async<'T>` object, which can increase overhead if it is used often
 
-In contrast, F# async workflows are more naturally cancellable. Cancellation is a simple three-step process.
+### Async.Catch
 
-1.  Create a new `CancellationTokenSource`.
-2.  Pass it into a starting function.
-3.  Call `Cancel` on the token.
+Creates an asychronous computation that executes a given `Async<'T>`, returning a `Async<Choice<'T, exn>>`. If the given `Async<'T>` completes successfully, then a `Choice1Of2` is returned with the resultant value. If it raises an exception before it completes, then a `Choice2of2` is returned with the raised exception. If it is used on an asynchronous computation that is itself composed of many computations, and one of those computations throws an exception, then encompassing computation will be stopped entirely.
```
mairaw:
```suggestion
Creates an asychronous computation that executes a given `Async<'T>`, returning an `Async<Choice<'T, exn>>`. If the given `Async<'T>` completes successfully, then a `Choice1Of2` is returned with the resultant value. If it raises an exception before it completes, then a `Choice2of2` is returned with the raised exception. If it is used on an asynchronous computation that is itself composed of many computations, and one of those computations throws an exception, then encompassing computation will be stopped entirely.
```

mairaw (commented):
Some quick comments. I'd also run this through Acrolinx.

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
-> [!NOTE]
-> Some inaccuracies have been discovered in this article.  It is being rewritten.  See [Issue #666](https://github.com/dotnet/docs/issues/666) to learn about the changes.
+Asynchronous programming is essential to modern applications. The applications for asynchronous programming are large in number. There are two primary use cases that most developers will encounter:
 
-Async programming in F# can be accomplished through a language-level programming model designed to be easy to use and natural to the language.
+* Having a non-blocking server process that can service incoming requests while awaiting results for other requests
+* Having a responsive UI or main thread while background work is happening
 
-The core of async programming in F# is `Async<'T>`, a representation of work that can be triggered to run in the background, where `'T` is either the type returned via the special `return` keyword or `unit` if the async workflow has no result to return.
+Although background work may involve multiple threads, there is no affinity between threading and asynchrony.
```
bartelink: Although background work often does involve the utilization of multiple threads, it's important to consider the concepts of asynchrony and multi-threading separately

(they are really quite separate concerns)
(one should not be assumed to imply the other)
(without conflating)
(or in Rich Hickey style: to not complect the concerns)

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
 
-The key concept to understand is that an async expression’s type is `Async<'T>`, which is merely a _specification_ of work to be done in an asynchronous context. It is not executed until you explicitly start it with one of the starting functions (such as `Async.RunSynchronously`). Although this is a different way of thinking about doing work, it ends up being quite simple in practice.
+## Asynchrony defined
```
bartelink: I think it's important to separate concurrency for parallelism in the terms of refernce early on - mayeb we'll get there and/or maybe it'll turn out that the narrative works (I'm deliberately not doing a read-through at this point)

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
 
-For example, say you wanted to download the HTML from dotnetfoundation.org without blocking the main thread. You can accomplish it like this:
+If you look at the etymology of the word "asynchronous", there are two pieces:
+
+* "a", meaning "not"
+* "synchronous", meaning "at the same time"
+
+Put together, "asynchronous" simply means, "not at the same time". This holds true for asynchronous programming in F#. Code that executes asynchronously is code that does not necessarily execute at the same time.
```
bartelink: >that does not necessarily execute at the same time.

that can be scheduled to be executed independently ?

contemporaneously?

mairaw:
You batch the suggestions on the file tab 😉

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+
+In other words, asynchronous computations execute independently of the main program flow. This does not imply any particular threading model, nor does it imply that a computation happens in the background. In fact, asynchronous computations could even execute synchronously, depending on the nature of the computation and the environment the computation is executing in.
```
bartelink: >execute independently of the main program flow
this is far too concrete for this point in the explanation - I'm thinking folks have tied it down to there being exactly two threads operating in some very closely controlled interlocked dance now.

The next sentence untangles it a bit, but the strawman has already done the damage IMO

>could even execute synchronously

may even be executed synchronously inline ?

Maybe extend the concurrency etymological exploration by explaining that concurrency can be abstract - it affords one the ability to defer the exact time and manner in which to actually run things a) on a thread b) at a particular point in time c) with a particular priority relative to other work ?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+
+The main takeaway you should have is that because asynchronous computations are simply independent of the main program flow, there are few guarantees about when or how a computation executes. This necessitates a means to control their execution. In F#, there are multiple ways to do this.
```
bartelink: Instead of this being a negative "you cant control it", inidcate that it's arbitrary and you get to defer when and how it's made concrete. Then say "lets explore some of the ways in which you can drive the execution / compose the flow / orchestrate the rendezvous"

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+
+## Core concepts
+
+In F#, asynchronous programming is centered around three core components:
```
bartelink: components -> elements ?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+
+* The `Async<'T>` type, which represents an asynchronous computation
```
bartelink: I like how this is succinct but would be very tempted to load on the fact that it's a builder affording one the ability to run a computation

   The `Async<'T>` type, which represents a Builder for a given computation from which one can compose a computation flow

i.e. allude to composition and workflows. `Task<T>` also "represents an asynchronous computation"

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+* The `async { }` [computation expression](../../language-reference/computation-expressions.md), which provides a convenient syntax for describing and controlling asynchronous computations
+* The `Async` functions, which let you start asynchronous work, coordinate multiple asynchronous computations, and transform asynchronous results
```
bartelink: Needs to mention `module`. Should probably mention that the `async` computation expression is expressed in terms of these? Maybe bring this sentence up before it ?
Should probably mention compose (as should the async comptutaiton expression) (though its nice and succicnt as it is and we don't want to sell the golden goos either)

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
     }
 
- // Execution will pause until fetchHtmlAsync finishes
- let html = "https://dotnetfoundation.org" |> fetchHtmlAsync |> Async.RunSynchronously
+[<EntryPoint>]
+let main argv =
+    argv
+    |> Array.map countFileBytes
+    |> Async.Parallel
+    |> Async.Ignore
```
bartelink: This Ignore irks me - Async.Start is wrong, and it triggers the need for this confusion - if you instead did a startAsChild and then Async.Ignore at the end (but maybe find a way to run Array.sum without showing concepts too early?)

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
- // you actually have the result from fetchHtmlAsync now!
- printfn "%s" html
- ```
+    Console.Read() |> ignore
+    0 // return an integer exit code
+```
+
+As you can see, the `main` function has quite a few more calls made. Conceptually, it does the following:
+
+1. Transform the command-line arguments into `Async<unit>` computations with `Array.map`
```
bartelink: computaiton builders? workflows ? I don't like the term computations for Async<T>, though in fairness the term is applied consistently

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+2. Create an `Async<'T[]>` that schedules and runs the `countFileBytes` computations in parallel when it runs.
+3. Create an `Async<unit>` that will run the parallel computation and ignore its result
+4. Explicitly run the last computation with `Async.Start`
```
bartelink: last -> overall composed ?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
 
-2.  `Async.Start` will start an async workflow on another thread, and will **not** await its result.
+When this program runs, `countFileBytes` runs in parallel for each command-line argument. Because asynchronous computations execute independently of program flow, there is no order in which they print their information and finish executing.
```
bartelink: it _may_ run in parallel - I think having the intro mention that parallelism is when there is an opportunity for stuff to really run at the same time - perhaps mention if oyu have one processor that two concurrent processes can be running concurrently, in parallel, but there's never a point where the system is contemporaneously advancing both threads of computation ?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+
+## Sequencing asynchronous computations
+
+Because `Async<'T>` is a specification of work rather than a computation that is currently running, you can perform more intricate transformations easily. For example, here is an implementation of the Sequence monad that converts a list of `Async<'T>` into an `Async<'T>` that contains an array of results. It uses the previous `countFileBytes` function:
```
bartelink: specification of work is a good term
"computation  that is currently running" is problematic for me - suggest "running task"

>you can perform more intricate transformations easily.

you can use the fact that you have a set of specifications / builders to build up a quite different overall computation from those same building blocks thanks to the fact that the actual scheduling is always deferred ?

>into an `Async<'T>` that contains an array of results

into an `Async<'T[]>` that will yield the aggregated results from the individual tasks at such a point as the full set becomes available.

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
 
 ```fsharp
-open System
-open System.Net
-  
-let uploadDataAsync url data = 
+module Async =
+    let sequence computations =
```
bartelink: now the computations stuff makes sense - perhaps leading with a glossary mentiong what a word means and doesnt mean contrasting it to Task and Async might work ?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+        async {
+            let results = ResizeArray()
```
bartelink: should this say `// where ResizeArray is the F# alias for System.Collections.LIst<T>`

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+        printfn "File %s has %d bytes" fileName bytes.Length
     }
 
-let workflow = uploadDataAsync "https://url-to-upload-to.com" "hello, world!"
+[<EntryPoint>]
+let main argv =
+    argv
+    |> Array.map countFileBytes
+    |> Async.sequence
+    |> Async.RunSynchronously
```
bartelink: why no start here (take the start out in other places)

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+    |> ignore
+```
 
-// Execution will continue after calling this!
-Async.Start(workflow)
+This will execute `countFileBytes` for each item in `argv` in the order of the elements of `argv` rather than executing them in parallel. This may be preferable if you wish to preserve the order of computations.
```
bartelink: Probably should mention side effects or somethng as people following too fast might think this implies that Async.Parallel might deliver the _results_ out of order ?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
+        Async.StartAsTask comp :> Task
+```
+
+There is already an `Async.AwaitTask` that accepts a <xref:System.Threading.Tasks.Task> as input. With this and the previously defined `startTaskFromAsyncUnit` function, you can start and await <xref:System.Threading.Tasks.Task> types from an F# async computation.
+
+## Relationship to multithreading
+
+Although threading is mentioned throughout this article, there are two important things to remember:
+
+1. There is no affinity between an asynchronous computation and a thread, unless explicitly started on the current thread
```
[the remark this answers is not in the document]
cartermp: I decided to remove it. The intro covers this well enough now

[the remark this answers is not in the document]
bartelink: 👍

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
cartermp: I'm going with computation so that there is consistent terminology with CEs in our documentation. But I'll have a "terminology" section at the top that mentions other, often equivalent terms.

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,360 @@
...
+    argv
+    |> Array.map countFileBytes
+    |> Async.sequence
+    |> Async.RunSynchronously
+    |> ignore
+```
+
+This will execute `countFileBytes` for each item in `argv` in the order of the elements of `argv` rather than executing them in parallel. This may be preferable if you wish to preserve the order of computations.
+
+## Common async functions
```
mcintyre94: Would it make sense to include `Async.Parallel` here too? It's used in one of the code examples above, but unlike the other Async functions used there it's not in this section.

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,360 @@
...
+
+What to watch out for:
+
+* This function blocks the calling thread, which undoes the use of asynchrony if it is used to only execute a single `Async<'T>`
+
+## Interoperating with .NET
+
+You may be working with a .NET library or C# codebase that uses [async/await](../../../standard/async.md)-style asynchronous programming. Because C# and the majority of .NET libraries use the <xref:System.Threading.Tasks.Task%601> and <xref:System.Threading.Tasks.Task> types as their core abstractions rather than `Async<'T>`, you must cross a boundary between these two approaches to asynchrony.
+
+### How to work with .NET async and Task<T>
```
BillWagner: One question you don't answer here: In F#, if I call a .NET API that returns `Task`, or `Task<T>`, does that return a "hot" task?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,360 @@
...
+    |> ignore
+```
+
+This will execute `countFileBytes` for each item in `argv` in the order of the elements of `argv` rather than executing them in parallel. This may be preferable if you wish to preserve the order of computations.
+
+## Common async functions
+
+Because F# asynchronous computations are a _specification_ of work rather than a representation of work that is already executing, they must be explicitly started with a starting function. There are many [Async starting functions](https://msdn.microsoft.com/library/ee370232.aspx) that are helpful in different contexts. The following section describes some of the more common starting functions.
+
+### Async.Start
```
BillWagner: Starting here, based on @bartelink's comments, here's my recommendation:

1. Reorder these different methods from most common use to least common. You can then structure the guidance as "use this, unless <some conditions in your scenario>".
1. I like the "when to use it" and "what to watch for" sections. What will help readers is to point to a "default" choice. (see above item). And, for those choices that aren't the "default", when should you use it?
1. You've done a great job above distinguishing multi-threaded, parallel, and asynchronous code. Where the samples and APIs below use parallel and multi-threaded algorithms, it would help to reinforce those distinctions.

This section is the key for readers. (I learned quite a bit from reading it and the comments). Getting this part crystal clear will pay huge dividends. Mistakes in understanding are mine, but I hope the feedback helps polish it.

BillWagner (changes requested):
This is an important and useful addition. I had some comments that related to my lack of F# knowledge. It's well on its way to being a great resource.

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
bartelink: that's a good question which encapsulates it all - if there's a FAQ/intro section it can be asked as a rhetorical device, that might work (I've not done a re-read since the initial drop - feel free to ping me when it reaches a point where that'd be appreciated).

(Essentially that's dictated by the method you call - they'll typically be hot as things returning `Task`s will be. When one uses an `|> Async.AwaitTask` in the body of an `async {}` CE, you're just adding a step into the builder to call the method in question (to kick off the work) and wrapping the result wrt handling the completion continuations; each time one instantiates/starts the `Async<T>` it'll traverse and call the method to start the task at the relevant part in the flow)

(Just typing al this in case it happens to overlap with what you mean ;) )

mairaw:
@cartermp this PR has been sitting here for a few months. Any plans on finishing this one?

cartermp:
Yep, when I'm boarding and/or on the plane for OpenFSharp next week

mairaw:
Ok, no worries! We were just going over some stale PRs.

mairaw:
Ping us when this is ready for review again, ok? Thanks!

cartermp:
I think this one is ready. I'd like to mention auto-propagation of cancellation tokens, but that's not essential to getting this one merged. It's currently a huge improvement over the existing doc

bartelink:
Drop me a line a day before you intend to press the merge button and I'll give it a final read thru

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+    0 // return an integer exit code
 ```
 
-And that’s it! Aside from the use of `async`, `let!`, and `return`, this is just normal F# code.
+As you can see, the `main` function has quite a few more calls made. Conceptually, it does the following:
 
-There are a few syntactical constructs which are worth noting:
+1. Transform the command-line arguments into `Async<unit>` computations with `Array.map`
+2. Create an `Async<'T[]>` that schedules and runs the `countFileBytes` computations in parallel when it runs.
+3. Create an `Async<unit>` that will run the parallel computation and ignore its result
```
Youssef1313:
```suggestion
3. Create an `Async<unit>` that will run the parallel computation and ignore its result.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+4. Explicitly run the last computation with `Async.Start`
```
Youssef1313:
```suggestion
4. Explicitly run the last computation with `Async.Start`.
```

cartermp:
After folks like @mairaw and/or @BillWagner make sure I didn't commit any cardinal sins in my recent changes 🙂

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+When you write async code in F# you'll usually interact with a framework that handles scheduling of computations for you. However, this is not always the case, so it is good to learn the various starting functions to schedule asynchronous work.
 
-    let fetchHtmlAsync url =
-        async {
-            let uri = Uri(url)
-            use webClient = new WebClient()
-            let! html = webClient.AsyncDownloadString(uri)
-            return html
-        }
+Because F# asynchronous computations are a _specification_ of work rather than a representation of work that is already executing, they must be explicitly started with a starting function. There are many [Async starting functions](https://msdn.microsoft.com/library/ee370232.aspx) that are helpful in different contexts.The following section describes some of the more common starting functions.
```
Youssef1313:
```suggestion
Because F# asynchronous computations are a *specification* of work rather than a representation of work that is already executing, they must be explicitly started with a starting function. There are many [Async starting functions](https://msdn.microsoft.com/library/ee370232.aspx) that are helpful in different contexts.The following section describes some of the more common starting functions.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
-    open System.Net
+When to use:
 
-    let uploadDataAsync url data =
-        async {
-            let uri = Uri(url)
-            use webClient = new WebClient()
-            webClient.UploadStringAsync(uri, data)
-        }
+* When you want to execute multiple asynchronous computations concurrently rather than one at a time, but not necessarily in parallel.
```
Youssef1313:
```suggestion
- When you want to execute multiple asynchronous computations concurrently rather than one at a time, but not necessarily in parallel.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+* When you wish to tie the lifetime of a child computation with a parent computation.
```
Youssef1313:
```suggestion
- When you wish to tie the lifetime of a child computation with a parent computation.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
 
-    let workflow = uploadDataAsync "https://url-to-upload-to.com" "hello, world!"
+What to watch out for:
 
-    // Execution will continue after calling this!
-    Async.Start(workflow)
+* Starting multiple computations with `Async.StartChild` isn't the same as scheduling them to run in parallel.
```
Youssef1313:
```suggestion
- Starting multiple computations with `Async.StartChild` isn't the same as scheduling them to run in parallel.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+* Canceling a parent computation will cancel all children it started.
```
Youssef1313:
```suggestion
- Canceling a parent computation will cancel all children it started.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+```fsharp
+computation: Async<unit> * cancellationToken: ?CancellationToken -> unit
+```
 
-    printfn "%s" "uploadDataAsync is running in the background..."
-    ```
+When to use:
 
-There are other ways to start an async workflow available for more specific scenarios. They are detailed [in the Async reference](https://msdn.microsoft.com/library/ee370232.aspx).
+* When you need to update something on the calling thread in the middle of an asynchronous computation.
```
Youssef1313:
```suggestion
- When you need to update something on the calling thread in the middle of an asynchronous computation.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
 
-### A Note on Threads
+What to watch out for:
 
-The phrase "on another thread" is mentioned above, but it is important to know that **this does not mean that async workflows are a facade for multithreading**. The workflow actually "jumps" between threads, borrowing them for a small amount of time to do useful work. When an async workflow is effectively "waiting" (for example, waiting for a network call to return something), any thread it was borrowing at the time is freed up to go do useful work on something else. This allows async workflows to utilize the system they run on as effectively as possible, and makes them especially strong for high-volume I/O scenarios.
+* Code in the asynchronous computation will run on the caller's thread, so if you need this thread to perform other work, `Async.StartImmediate` isn't likely an appropriate starting function.
```
Youssef1313:
```suggestion
- Code in the asynchronous computation will run on the caller's thread, so if you need this thread to perform other work, `Async.StartImmediate` isn't likely an appropriate starting function.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+When to use:
 
-let fetchHtmlAsync url =
-    async {
-        let uri = Uri(url)
-        use webClient = new WebClient()
-        let! html = webClient.AsyncDownloadString(uri)
-        return html
-    }
+* When you need to call into a .NET API that expects a <xref:System.Threading.Tasks.Task%601> to represent the result of an asynchronous computation.
```
Youssef1313:
```suggestion
- When you need to call into a .NET API that expects a <xref:System.Threading.Tasks.Task%601> to represent the result of an asynchronous computation.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+
+What to watch out for:
+
+* This call will allocate an additional `Task` object, which can increase overhead if it is used often.
```
Youssef1313:
```suggestion
- This call will allocate an additional `Task` object, which can increase overhead if it is used often.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
-    printfn "%s" html
+```fsharp
+computations: seq<Async<'T>> * maxDegreesOfParallelism: ?int -> Async<'T[]>
 ```
 
-## Important Info and Advice
+When to use it:
 
-* Append "Async" to the end of any functions you’ll consume
+* If you need to run a set of computations at the same time and have no reliance on their order of execution.
```
Youssef1313:
```suggestion
- If you need to run a set of computations at the same time and have no reliance on their order of execution.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
 
- Although this is just a naming convention, it does make things like API discoverability easier. Particularly if there are synchronous and asynchronous versions of the same routine, it’s a good idea to explicitly state which is asynchronous via the name.
+What to watch out for:
 
-* Listen to the compiler!
+* Computations will be run however they end up getting scheduled. This means you cannot rely on their order of their execution.
```
Youssef1313:
```suggestion
- Computations will be run however they end up getting scheduled. This means you cannot rely on their order of their execution.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
-There is a fundamental difference between the C#/VB async model and the F# async model.
+```fsharp
+computations: seq<Async<'T>> -> Async<'T[]>
+```
 
-When you call a function which returns a `Task` or `Task<'T>`, that job has already begun execution. The handle returned represents an already-running asynchronous job. In contrast, when you call an async function in F#, the `Async<'a>` returned represents a job which will be **generated** at some point. Understanding this model is powerful, because it allows for asynchronous jobs in F# to be chained together easier, performed conditionally, and be started with a finer grain of control.
+When to use it:
 
-There are a few other similarities and differences worth noting.
+* If you need to execute multiple computations in order and/or you rely on their order of execution.
```
Youssef1313:
```suggestion
- If you need to execute multiple computations in order and/or you rely on their order of execution.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
 
-### Similarities
+What to watch out for:
 
-* `let!`, `use!`, and `do!` are analogous to `await` when calling an async job from within an `async{ }` block.
+* Computations will be run in the order that they are passed to this function, which can mean significantly more time spent to get a result when compared with `Async.Parallel`.
```
Youssef1313:
```suggestion
- Computations will be run in the order that they are passed to this function, which can mean significantly more time spent to get a result when compared with `Async.Parallel`.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
-### Differences
+```fsharp
+task: Task<'T>  -> Async<'T>
+```
 
-* Nested `let!` is not allowed, unlike nested `await`
+When to use:
 
-  Unlike `await`, which can be nested indefinitely, `let!` cannot and must have its result bound before using it inside of another `let!`, `do!`, or `use!`.
+* When you are consuming a .NET API that returns a <xref:System.Threading.Tasks.Task%601> within an F# asynchronous computation.
```
Youssef1313:
```suggestion
- When you are consuming a .NET API that returns a <xref:System.Threading.Tasks.Task%601> within an F# asynchronous computation.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
 
-* Cancellation support is simpler in F# than in C#/VB.
+What to watch out for:
 
-  Supporting cancellation of a task midway through its execution in C#/VB requires checking the `IsCancellationRequested` property or calling `ThrowIfCancellationRequested()` on a `CancellationToken` object that’s passed into the async method.
+* This call will allocate an additional `Async<'T>` object, which can increase overhead if it is used often.
```
Youssef1313:
```suggestion
- This call will allocate an additional `Async<'T>` object, which can increase overhead if it is used often.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
 ```fsharp
-open System.Threading
+computation: Async<'T> -> Async<Choice<'T, exn>>
+```
 
-// Create a workflow which will loop forever.
-let workflow =
+When to use:
+
+* When you are performing asynchronous work that may fail with an exception and you want to handle that exception in the caller.
```
Youssef1313:
```suggestion
- When you are performing asynchronous work that may fail with an exception and you want to handle that exception in the caller.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+
+What to watch out for:
+
+* When using combined or sequenced asynchronous computations, the encompassing computation will fully stop if one of its "internal" computations throws an exception.
```
Youssef1313:
```suggestion
- When using combined or sequenced asynchronous computations, the encompassing computation will fully stop if one of its "internal" computations throws an exception.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+
+Signature:
+
+```fsharp
+computation: Async<'T> -> Async<unit>
+```
+
+When to use:
+
+* When you have an asynchronous computation whose result can be discarded.
```
Youssef1313:
```suggestion
- When you have an asynchronous computation whose result can be discarded.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+* In conjunction with `Async.Start` or other functions that accept an `Async<unit>`.
```
Youssef1313:
```suggestion
- In conjunction with `Async.Start` or other functions that accept an `Async<unit>`.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+
+What to watch out for:
+
+* If you must use this because you wish to use `Async.Start` or another function that requires `Async<unit>`, consider if discarding the result is okay to do. Discarding results just to fit a type signature should not generally be done
```
Youssef1313:
```suggestion
- If you must use this because you wish to use `Async.Start` or another function that requires `Async<unit>`, consider if discarding the result is okay to do. Discarding results just to fit a type signature should not generally be done.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+
+Signature:
+
+```fsharp
+computation: Async<'T> * timeout: ?int * cancellationToken: ?CancellationToken -> 'T
+```
+
+When to use it:
+
+* If you need it, use it only once in an application - at the entry point for an executable.
```
Youssef1313:
```suggestion
- If you need it, use it only once in an application - at the entry point for an executable.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+* When you don't care about performance and want to execute a set of other asynchronous operations at once.
```
Youssef1313:
```suggestion
- When you don't care about performance and want to execute a set of other asynchronous operations at once.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+
+What to watch out for:
+
+* This function blocks the calling thread, which nullifies the point of asynchrony if it is used to only execute a single `Async<'T>`
```
Youssef1313:
```suggestion
- This function blocks the calling thread, which nullifies the point of asynchrony if it is used to only execute a single `Async<'T>`.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+
+Signature:
+
+```fsharp
+computation: Async<unit> * cancellationToken: ?CancellationToken -> unit
+```
+
+Use only when:
+
+* You have an asynchronous computation but don't care about the result.
```
Youssef1313:
```suggestion
- You have an asynchronous computation but don't care about the result.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+* You don't care about when an asynchronous computation runs.
```
Youssef1313:
```suggestion
- You don't care about when an asynchronous computation runs.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+* You don't care which thread an asynchronous computation runs on.
```
Youssef1313:
```suggestion
- You don't care which thread an asynchronous computation runs on.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+* You don't have any need to be aware of or report exceptions resulting from the task.
```
Youssef1313:
```suggestion
- You don't have any need to be aware of or report exceptions resulting from the task.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+
+What to watch out for:
+
+* Exceptions raised by computations started with `Async.Start` aren't propagated to the caller.
```
Youssef1313:
```suggestion
- Exceptions raised by computations started with `Async.Start` aren't propagated to the caller.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+* Any effectual work (such as calling `printfn`) started with `Async.Start` won't cause the effect to happen on the main thread of a program's execution.
```
Youssef1313:
```suggestion
- Any effectual work (such as calling `printfn`) started with `Async.Start` won't cause the effect to happen on the main thread of a program's execution.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
 ```
 
-And that’s it!
+There is already an `Async.AwaitTask` that accepts a <xref:System.Threading.Tasks.Task> as input. With this and the previously defined `startTaskFromAsyncUnit` function, you can start and await <xref:System.Threading.Tasks.Task> types from an F# async computation.
+
+## Relationship to multithreading
+
+Although threading is mentioned throughout this article, there are two important things to remember:
+
+1. There is no affinity between an asynchronous computation and a thread, unless explicitly started on the current thread
```
Youssef1313:
```suggestion
1. There is no affinity between an asynchronous computation and a thread, unless explicitly started on the current thread.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+1. Asynchronous programming in F# is not an abstraction for multithreading
```
Youssef1313:
```suggestion
1. Asynchronous programming in F# is not an abstraction for multi-threading.
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+
+What this means is that **asynchronous computations are not an abstraction for multithreading**. For example, a computation may actually run on its caller's thread, depending on the nature of the work. A computation could also "jump" between threads, borrowing them for a small amount of time to do useful work in between periods of "waiting" (such as when a network call is in transit).
```
Youssef1313:
```suggestion
What this means is that **asynchronous computations are not an abstraction for multi-threading**. For example, a computation may actually run on its caller's thread, depending on the nature of the work. A computation could also "jump" between threads, borrowing them for a small amount of time to do useful work in between periods of "waiting" (such as when a network call is in transit).
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+
+Although F# provides some abilities to start an asynchronous computation on the current thread (or explicitly not on the current thread), asynchrony generally is not associated with a particular threading strategy.
 
-## Further resources:
+## See also
 
-* [Async Workflows on MSDN](https://msdn.microsoft.com/library/dd233250.aspx)
-* [Asynchronous Sequences for F#](https://fsprojects.github.io/FSharp.Control.AsyncSeq/library/AsyncSeq.html)
-* [F# Data HTTP Utilities](https://fsharp.github.io/FSharp.Data/library/Http.html)
+* [The F# Asynchronous Programming Model](https://www.microsoft.com/en-us/research/publication/the-f-asynchronous-programming-model)
```
Youssef1313:
```suggestion
- [The F# Asynchronous Programming Model](https://www.microsoft.com/research/publication/the-f-asynchronous-programming-model)
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+* [Jet.com's F# Async Guide](https://medium.com/jettech/f-async-guide-eb3c8a2d180a)
```
Youssef1313:
```suggestion
- [Jet.com's F# Async Guide](https://medium.com/jettech/f-async-guide-eb3c8a2d180a)
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+* [F# for fun and profit's Asynchronous Programming guide](https://fsharpforfunandprofit.com/posts/concurrency-async-and-parallel/)
```
Youssef1313:
```suggestion
- [F# for fun and profit's Asynchronous Programming guide](https://fsharpforfunandprofit.com/posts/concurrency-async-and-parallel/)
```

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+* [Async in C# and F#: Asynchronous gotchas in C#](http://tomasp.net/blog/csharp-async-gotchas.aspx/)
```
Youssef1313:
```suggestion
- [Async in C# and F#: Asynchronous gotchas in C#](http://tomasp.net/blog/csharp-async-gotchas.aspx/)
```

Youssef1313 (changes requested):

BillWagner (approved):

BillWagner:
I really like how this turned out @cartermp 

You can :shippit: when you are ready.

cartermp:
cc @bartelink

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
 
-let html = "https://dotnetfoundation.org" |> fetchHtmlAsync |> Async.RunSynchronously
-printfn "%s" html
+[<EntryPoint>]
+let main argv =
+    argv
+    |> Array.map countFileBytes
+    |> Async.Parallel
+    |> Async.Ignore
+    |> Async.Start
```
eiriktsarpalis: Async.Start should be avoided by beginners. Async.StartAsTask is better.

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+Schedules a sequence of asynchronous computations to be executed in parallel. The degree of parallelism can be optionally tuned/throttled by specifying the `maxDegreesOfParallelism` parameter.
 
-let htmlList = getHtmlList urlList
+Signature:
 
-// We now have the downloaded HTML for each site!
-for html in htmlList do
-    printfn "%s" html
+```fsharp
+computations: seq<Async<'T>> * maxDegreesOfParallelism: ?int -> Async<'T[]>
```
eiriktsarpalis: --> `?maxDegreesOfParallelism: int`

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+When to use:
 
-  Unlike `await`, which can be nested indefinitely, `let!` cannot and must have its result bound before using it inside of another `let!`, `do!`, or `use!`.
+- When you are consuming a .NET API that returns a <xref:System.Threading.Tasks.Task%601> within an F# asynchronous computation.
 
-* Cancellation support is simpler in F# than in C#/VB.
+What to watch out for:
 
-  Supporting cancellation of a task midway through its execution in C#/VB requires checking the `IsCancellationRequested` property or calling `ThrowIfCancellationRequested()` on a `CancellationToken` object that’s passed into the async method.
+- This call will allocate an additional `Async<'T>` object, which can increase overhead if it is used often.
```
eiriktsarpalis: this is probably insignificant in the context of async and computation expressions in general. I would probably remove.

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+```
 
-* Nested `let!` is not allowed, unlike nested `await`
+When to use:
 
-  Unlike `await`, which can be nested indefinitely, `let!` cannot and must have its result bound before using it inside of another `let!`, `do!`, or `use!`.
+- When you are consuming a .NET API that returns a <xref:System.Threading.Tasks.Task%601> within an F# asynchronous computation.
 
-* Cancellation support is simpler in F# than in C#/VB.
+What to watch out for:
```
eiriktsarpalis: - Exceptions are wrapped in `TaskCanceledException` following TPL convention, however this is different compared to how the rest of F# async generally surfaces exceptions.

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,349 @@
...
-let workflow =
+* When you are performing asynchronous work that may fail with an exception and you want to handle that exception in the caller.
+
+What to watch out for:
+
+* When using combined or sequenced asynchronous computations, the encompassing computation will fully stop if one of its "internal" computations throws an exception.
+
+### Async.Ignore
+
+Creates an asynchronous computation that runs the given computation and ignores its result.
```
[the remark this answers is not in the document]
eiriktsarpalis: +1 need to mention exception behaviour here.

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+- When you have an asynchronous computation whose result can be discarded.
+- In conjunction with `Async.Start` or other functions that accept an `Async<unit>`.
+
+What to watch out for:
+
+- If you must use this because you wish to use `Async.Start` or another function that requires `Async<unit>`, consider if discarding the result is okay to do. Discarding results just to fit a type signature should not generally be done.
+
+### Async.RunSynchronously
+
+Runs an asynchronous computation and awaits its result on the calling thread. This call is blocking.
```
eiriktsarpalis: "This call blocking the calling thread"?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+Use only when:
+
+- You have an asynchronous computation but don't care about the result.
+- You don't care about when an asynchronous computation runs.
+- You don't care which thread an asynchronous computation runs on.
+- You don't have any need to be aware of or report exceptions resulting from the task.
+
+What to watch out for:
+
+- Exceptions raised by computations started with `Async.Start` aren't propagated to the caller.
```
eiriktsarpalis: They will unwind the stack completely resulting in the process terminating.

eiriktsarpalis: Unless there's a global exception handler installed in full framework. I believe classic fsi does this.

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+All three are orthogonal concepts, but can be easily conflated, especially when they are used together. For example, you may need to execute multiple asynchronous computations in parallel. This does not mean that parallelism or asynchrony imply one another in this situation.
+
+If you consider the etymology of the word "asynchronous", there are two pieces involved:
+
+- "a", meaning "not".
+- "synchronous", meaning "at the same time".
+
+When you put these two terms together, you'll see that "asynchronous" means "not at the same time". That's it! There is no implication of concurrency or parallelism in this definition. This is also true in practice.
+
+In practical terms, asynchronous computations in F# are scheduled to execute independently of the main program flow. This doesn't imply concurrency or parallelism, nor does it imply that a computation always happens in the background. In fact, asynchronous computations could even execute synchronously, depending on the nature of the computation and the environment the computation is executing in.
```
bartelink: could even execute -> can even execute

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+
+The main takeaway you should have is that asynchronous computations are independent of the main program flow. Although there are few guarantees about when or how an asynchronous computation executes, there are some approaches to orchestrating and scheduling them. The rest of this article explores core concepts for F# asynchrony and how to use the types, functions, and expressions built into F#.
+
+## Core concepts
+
+In F#, asynchronous programming is centered around three core concepts:
+
+- The `Async<'T>` type, which represents a composable asynchronous computation.
+- The `Async` module functions, which let you schedule asynchronous work, compose multiple asynchronous computations, and transform asynchronous results.
```
bartelink: lose multiple?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+- The `async { }` [computation expression](../../language-reference/computation-expressions.md), which provides a convenient syntax for describing and controlling asynchronous computations.
```
bartelink: >describing and controlling 

describing feels very meta. can't this just be 'building' ?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+let main argv =
+    countFileBytes "path-to-file.txt"
+    |> Async.RunSynchronously
 
-        // Execution of fetchHtmlAsync won't continue until the result
-        // of AsyncDownloadString is bound.
-        let! html = webClient.AsyncDownloadString(uri)
-        return html
+    Console.Read() |> ignore
+    0 // return an integer exit code
```
bartelink: >// return an integer exit code

is distracting

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+```
+
+In the example, the `countFileBytes` function is of type `string -> Async<unit>`. Calling the function does not actually execute the asynchronous computation. Instead, it returns an `Async<unit>` that acts as a *specification* of the work that is to execute asynchronously. It calls `Async.AwaitTask` in its body, which will convert the result of <xref:System.IO.File.WriteAllBytesAsync%2A> to an appropriate type when it is called.
```
bartelink: >It calls 

it will call, but it doesnt yet - 'it uses' ?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+
+Another important line is the call to `Async.RunSynchronously`. This is one of the async starting functions that you'll need to call if you want to actually execute an F# asynchronous computation.
```
bartelink: >async starting 

Async starting - capital letter might mentally tie Async module to Async<T> in readers mind?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+
+This is a key difference with the C#/VB style of `async` programming. In F#, asynchronous computations can be thought of as **Cold tasks**. They must be explicitly started to actually execute. This has some advantages, as it allows you to combine and sequence asynchronous work much more easily than in C#/VB.
```
bartelink: >key difference

fundamental ?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+
+    Console.Read() |> ignore
+    0 // return an integer exit code
 ```
 
-And that’s it! Aside from the use of `async`, `let!`, and `return`, this is just normal F# code.
+As you can see, the `main` function has quite a few more calls made. Conceptually, it does the following:
 
-There are a few syntactical constructs which are worth noting:
+1. Transform the command-line arguments into `Async<unit>` computations with `Array.map`
```
bartelink: it transforms it into `Async<int>` computations, no ?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+2. Create an `Async<'T[]>` that schedules and runs the `countFileBytes` computations in parallel when it runs.
+3. Create an `Async<unit>` that will run the parallel computation and ignore its result.
+4. Explicitly run the last computation with `Async.Start`.
```
bartelink: worth mentioning that it kicks them off but does not await and that maybe what will take place if you press a key. (Or consider using StartAsChild for more easy to talk about semantics?)

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
bartelink: to restate the above: there is an Async.Parallel, but there is no guarantee of parallelism - only concurrency is guaranteed>

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
bartelink: I still wish the Ignore was not there, but am less convinced there is a better way to express it

bartelink (approved):
This is looking well - some suggestions within. I'll post some more comments shortly.

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+        return result
+    } |> Async.StartAsTask
+```
 
-let tokenSource = new CancellationTokenSource()
+### How to work with .NET async and Task
 
-// Start the workflow in the background
-Async.Start (workflow, tokenSource.Token)
+To work with APIs that use <xref:System.Threading.Tasks.Task> (that is, .NET async computations that have no return value), you may need to add an additional function that will convert an `Async<'T>` to a <xref:System.Threading.Tasks.Task>:
```
bartelink: >have no return value

do not return a value

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+There is already an `Async.AwaitTask` that accepts a <xref:System.Threading.Tasks.Task> as input. With this and the previously defined `startTaskFromAsyncUnit` function, you can start and await <xref:System.Threading.Tasks.Task> types from an F# async computation.
+
+## Relationship to multithreading
+
+Although threading is mentioned throughout this article, there are two important things to remember:
+
+1. There is no affinity between an asynchronous computation and a thread, unless explicitly started on the current thread.
+1. Asynchronous programming in F# is not an abstraction for multi-threading.
+
+What this means is that **asynchronous computations are not an abstraction for multi-threading**. For example, a computation may actually run on its caller's thread, depending on the nature of the work. A computation could also "jump" between threads, borrowing them for a small amount of time to do useful work in between periods of "waiting" (such as when a network call is in transit).
```
bartelink: what this means ... verbatim repeats note 2 and does not read well

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
+```fsharp
+let computationForCaller param =
+    async {
+        let! result = getAsyncResult param
+        return result
+    } |> Async.StartAsTask
+```
 
-let tokenSource = new CancellationTokenSource()
+### How to work with .NET async and Task
```
bartelink: Task -> `Task`

bartelink (approved):

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
 ---
-# Async Programming in F\#
+# Async programming in F\#
 
-> [!NOTE]
-> Some inaccuracies have been discovered in this article.  It is being rewritten.  See [Issue #666](https://github.com/dotnet/docs/issues/666) to learn about the changes.
+Asynchronous programming is a mechanism that is essential to modern applications for diverse reasons. There are two primary use cases that most developers will encounter:
 
-Async programming in F# can be accomplished through a language-level programming model designed to be easy to use and natural to the language.
+- Presenting a server process that can service a significant number of concurrent incoming requests, while minimizing the system resources locked while request processing awaits inputs from systems or services external to that process.
```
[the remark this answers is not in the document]
chrnola: Or perhaps "utilized"? 🤔

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
 
-The key concept to understand is that an async expression’s type is `Async<'T>`, which is merely a _specification_ of work to be done in an asynchronous context. It is not executed until you explicitly start it with one of the starting functions (such as `Async.RunSynchronously`). Although this is a different way of thinking about doing work, it ends up being quite simple in practice.
+## Asynchrony defined
 
-For example, say you wanted to download the HTML from dotnetfoundation.org without blocking the main thread. You can accomplish it like this:
+The previous point - that asynchrony is independent of the utilization of multiple threads - is worth explaining a bit further. There are three concepts that are sometimes related, but strictly independent of one another:
+
+- Concurrency, which is when multiple computations execute in overlapping time periods.
+- Parallelism, which is when multiple computations or several parts of a unique computation run at exactly the same time.
```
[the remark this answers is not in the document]
Thraka: @bartelink  Just FYI, these suggestions are easier to implement if you use the suggestion button at the top of the comment:

image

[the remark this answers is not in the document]
bartelink: Thanks @Thraka - I have used that in the past when it was in beta [and evidently easier to locate!] (I looked for and failed to find the icon with this batch!)
Not sure how much has been lost though - the bulk of my suggestions are not simple word substitution suggestions, and I also fully expect Phillip to be selective in applying them given the bulk of his word choices are far from arbitrary

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
cartermp: Replaced with "Async module"

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
cartermp: Nope, in this case it's `Async<unit>`:

```
val countFileBytes : path:string -> Async<unit>
```

via FSI

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
```diff
@@ -1,213 +1,382 @@
...
-    open System.Net
+When to use:
 
-    let uploadDataAsync url data =
-        async {
-            let uri = Uri(url)
-            use webClient = new WebClient()
-            webClient.UploadStringAsync(uri, data)
-        }
+- When you want to execute multiple asynchronous computations concurrently rather than one at a time, but not necessarily in parallel.
```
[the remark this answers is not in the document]
cartermp: Changed to say scheduling

cartermp:
```
docs/core/porting/project-structure.md: 66: MD004/ul-style Unordered list style [Expected: asterisk; Actual: dash]
```

These make no sense

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
bartelink: right, I guess what I was angling at is that a `countFileBytes` function could reasonably be expected to return an int?

docs/fsharp/tutorials/asynchronous-and-concurrent-programming/async.md
bartelink: thanks for the rename, I finally get it now, after all this time !  (I am still wondering whether returning the counts per file and then totalling might make a better motivating example - with the bonus of not having to involve `Async.Ignore` (this is not really a big deal))

cartermp:
@BillWagner is this still good to merge with the markdownlint complaints? Lots of the complaints aren't this PR at all

BillWagner:
@cartermp Yes, you can :shipit:  We're trying this tool, and we'll be adjusting the warning level.

PR #9652, opened 2018-12-18
C54009: Nested asterisks

v-makoud:
Hello, @rpetrusha,
Translator has reported possible source content issue.
Please review and merge the suggested proposed file change to avoid this error. If you make related fix in another PR, then share your PR number, so we can confirm and close this PR.
Many thanks in advance.

## Summary

Describe your changes here.


mairaw (approved):
I don't see any difference for the en-us content, so it might be a transformer issue not being able to process those well. But I'll approve these changes. /cc @JasonCard

PR #9659, opened 2018-12-18
Improve Getting Started Guide by removing inconsistencies in service names 

### LINKED ISSUE #6484: host example needlessly/confusingly duplicates service
Hopefully someone else will benefit from this note. In this example, the given Main() used for the console host app is pointless. All of this code could be deleted except for the Console.WriteLine/ReadLine lines. This code creates a **second** service in addition to the first one shown in the previous pages of the tutorial. And it creates it on a different port. The earlier app.config defined one is on 8733 and this one is on 8000.

This is a good example of creating stuff in code instead of the app.config, but it's not in any way marked like that and only confuses people learning this. I was trying to switch the example from http to named pipes and I wound up getting "already in use" issues. Only now does it make sense as to why, since there were basically two servers on the same named pipe.

Edit: And to help anyone who might be getting confused by another aspect of this, by default when you have both the host and the lib in the same solution, running the host will ALSO cause VS to run a service host for the lib. This can be disabled in the project properties for the lib project, under WCF Options. Uncheck "Start WCF Service Host when debugging another project in the same solution." Once you do that, you get ONLY the service endpoint you create in the host.

⚠ Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.

ID: 0c8cbb16-05b1-a5f2-9031-1d7beb52efea
Version Independent ID: 98092a9f-c19a-cca9-75d6-75b9c3d4cc0b
Content: How to: Host and Run a Basic Windows Communication Foundation Service
Content Source: docs/framework/wcf/how-to-host-and-run-a-basic-wcf-service.md
Product: dotnet-framework
GitHub Login: @dotnet-bot
Microsoft Alias: dotnetcontent

fredrikhaglund:
## Summary

- Fixed formatting of Important section.
- Fixed inconsistency with different baseAddress in csharp and vb example (also aligned to what is used in other steps)
- Updated instruction how to verify that your service is running.

Fixes #6484


BillWagner (approved):
Thanks for putting this together @fredrikhaglund 

We appreciate it.  I've reviewed the changes, and I'll :shipit: now.  You should see the change published on our live site in the next few days, on our regular publishing cycle.

Thanks again for making your first contribution to docs. 🎆

PR #9666, opened 2018-12-19
Fix broken markdown table

Happypig375:
## Summary

Fixes a broken markdown table.


mairaw (approved):
Thank you for fixing that @Happypig375! Your contribution should be live shortly.

PR #9672, opened 2018-12-19
Correct comparison and minor cleanup in example

mikkelbu:
Fixes #9670


BillWagner (approved):
Thanks for making this update @mikkelbu We appreciate it.

I'll :shipit: now.

PR #9674, opened 2018-12-19
Typo

Fyers:


BillWagner (approved):
Thanks for spotting this and fixing it @Fyers We appreciate it.

I'll :shipit: now.  You should see the changes on the live site in the next day or two.

Thanks again.

PR #9711, opened 2018-12-21
Fix double space

pascalberger:
## Summary

Fixes typo with double space


mairaw (approved):
Thank you for removing the extra space.

PR #9719, opened 2018-12-22
fix missing period in global.json

gengle:
## Summary
global.json was spelled globaljson


gengle:
@mairaw - you are right I confused filename with template moniker.

PR #9740, opened 2018-12-25
Methods: Return Values - Removing conditional and return null

### LINKED ISSUE #8959: Bad example, doesn't compile, needs to be nullable
Hi.
This code from this docs' page doesn't compile as the ValueTuple is not nullable, please fix it:

```
public (string FName, string MName, string LName, int Age) GetPersonalInfo(string id)
{
    PersonInfo per = PersonInfo.RetrieveInfoById(id);
    if (per != null)
       return (per.FirstName, per.MiddleName, per.LastName, per.Age);
    else
       return null;    // here's the problem
}
```

---
#### Document Details

* ID: 158425f1-86b1-101a-9172-14c944867784
* Version Independent ID: 6386f636-99b3-93c3-ca9e-0134f3e3be10
* Content: [Methods - C# Guide](https://docs.microsoft.com/en-us/dotnet/csharp/methods)
* Content Source: [docs/csharp/methods.md](https://github.com/dotnet/docs/blob/master/docs/csharp/methods.md)
* Product: **dotnet-csharp**
* GitHub Login: @rpetrusha
* Microsoft Alias: **ronpet**

lucasnobredev:
## Summary

This PR came out from this [issue](https://github.com/dotnet/docs/issues/8959).

I gave two ideas, but I believe the first one is more important and objective. I remove conditional and return nulls from examples of return values with tuples.

I'm not sure if an example with nullable return it's so important to the docs, that's why I didn't do it.

Fixes #8959

### DIFF
--- a/docs/csharp/methods.md
+++ b/docs/csharp/methods.md
@@ -191,10 +191,7 @@ Sometimes, you want your method to return more than a single value. Starting wit
 public (string, string, string, int) GetPersonalInfo(string id)
 {
     PersonInfo per = PersonInfo.RetrieveInfoById(id);
-    if (per != null)
-       return (per.FirstName, per.MiddleName, per.LastName, per.Age);
-    else
-       return null;
+    return (per.FirstName, per.MiddleName, per.LastName, per.Age);
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Thraka (approved):
Awesome! This makes sense. Thanks a bunch for fixing it.

Issue #8364, opened 2018-10-15
Change absolute https://docs.microsoft.com/dotnet links to relative links


rpetrusha:
## Change absolute https://docs.microsoft.com/dotnet links to relative links
 
There are a number of fully qualified URLs in the dotnet/docs repo that point to files in the dotnet/docs repo itself. These can be changed to relative links. For example, `https://docs.microsoft.com/dotnet/csharp/somearticle.md` can be changed a relative link.
 
There may also be some links in the dotnet/dotnet-api-docs repo, although most of the ones that surface in a search are in the XML portion of the managed reference and should be left as-is.

msschl:
There are still some absolute links pointing to `https://docs.microsoft.com/dotnet/api/...`.

I don't know whether these links could be changed to relative links as well.

NextTurn:
@rpetrusha 

I would suggest a supplement to the issue description:

Prefer `../` to `~/` in links for files in the same repo, as described in [styleguide/template.md](https://github.com/dotnet/docs/blob/master/styleguide/template.md#internal-links).

PR #9741, opened 2018-12-25
Change links to relative links 

lucasnobredev:
## Summary

This PR is from #8364 . These are some fixes to this case.


Thraka (changes requested):
Hi! Thanks a bunch for this PR. There are 2 links that need to be fixed. And for reference, I think you can use something like `~/api/index.md` instead of `../../api/index.md` where `~` signals the root of the repo.

Thraka:
I checked the other PR and your links that were using ~ were just wrong in themselves, it's not that the ~ doesn't work.

Thraka (approved):

Thraka:
Looks good, thanks!

PR #9745, opened 2018-12-26
Fix XML code blocks

nxtn:
Fix em dashes `'\u2014'` in XML comments.

See #9744


Thraka (approved):
@NextTurn  Awesome! Thanks for the fixes!

PR #9747, opened 2018-12-26
C57267: Adding spaces before code to avoid broken display

v-maudel:
Hello, @JRAlexander ,
Description: add to spaces after end of line so that code snippet can be seen.

Many thanks in advance.

### DIFF
--- a/docs/machine-learning/tutorials/taxi-fare.md
+++ b/docs/machine-learning/tutorials/taxi-fare.md
@@ -94,7 +94,8 @@ The `TaxiTripFarePrediction` class represents predicted results. It has a single
 
 ## Define data and model paths
 
-Add the following additional `using` statements to the top of the *Program.cs* file:  
+Add the following additional `using` statements to the top of the *Program.cs* file:
+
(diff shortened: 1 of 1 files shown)


Thraka:
@v-maudel 

I think the better fix is to insert a blank line between the two lines. A double space at the end of a line is a specific markdown format to indicate that the line should be continued with a soft break on the next line instead of a paragraph. 

In this case, the author is trying to insert a code example, which is a block element. A blank line would be best.

Thraka (changes requested):
@v-maudel 

I think the better fix is to insert a blank line between the two lines. A double space at the end of a line is a specific markdown format to indicate that the line should be continued with a soft break on the next line instead of a paragraph. 

In this case, the author is trying to insert a code example, which is a block element. A blank line would be best.

v-maudel:
Thanks so much @Thraka ,
I've added a break-line between the title and the code now.

@JRAlexander

Thraka (approved):

v-maudel:
Thanks @Thraka

PR #9748, opened 2018-12-26
C57274: Changing wrong closing bracket in link structure

v-maudel:
Hello, @CESARDELATORRE,
Description: There is "\}" bracket instead of "\]" in link that is breaking formatting on EN and some LOC pages.
 
Many thanks in advance.


Thraka (approved):

v-maudel:
Thanks @Thraka

PR #9828, opened 2019-01-04
C57279: Correcting bold format and escaping Square brackets

v-maudel:
Hello, @rpetrusha , @drewnoakes,
Description: There is problem with strings that are written for example like this one - "**/a**[**ssembly**]**:**". Based on the content in source it doesn't make sense to write it like this, because for example there is also string "**/c[ompiler]:**", which has brackets "[]" in bold as well. We suggest to keep it consistent and write all strings like the second one, so it works on localized languages too.

Many thanks in advance.

## Summary

Describe your changes here.

### DIFF
--- a/docs/standard/serialization/xml-serializer-generator-tool-sgen-exe.md
+++ b/docs/standard/serialization/xml-serializer-generator-tool-sgen-exe.md
@@ -16,19 +16,19 @@ sgen [options]
   
 |Option|Description|  
 |------------|-----------------|  
-|**/a**[**ssembly**]**:**_filename_|Generates serialization code for all the types contained in the assembly or executable specified by *filename*. Only one file name can be provided. If this argument is repeated, the last file name is used.|  
-|**/c[ompiler]:** *options*|Specifies the options to pass to the C# compiler. All csc.exe options are supported as they are passed to the compiler. This can be used to specify that the assembly should be signed and to specify the key file.|  
-|**/d**[**ebug**]|Generates an image that can be used with a debugger.|  
-|**/f[orce]**|Forces the overwriting of an existing assembly of the same name. The default is **false**.|  
+|**/a\[ssembly\]:**_filename_|Generates serialization code for all the types contained in the assembly or executable specified by *filename*. Only one file name can be provided. If this argument is repeated, the last file name is used.|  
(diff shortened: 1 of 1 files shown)


docs/standard/serialization/xml-serializer-generator-tool-sgen-exe.md
```diff
@@ -16,19 +16,19 @@ sgen [options]
...
 |------------|-----------------|  
-|**/a**[**ssembly**]**:**_filename_|Generates serialization code for all the types contained in the assembly or executable specified by *filename*. Only one file name can be provided. If this argument is repeated, the last file name is used.|  
-|**/c[ompiler]:** *options*|Specifies the options to pass to the C# compiler. All csc.exe options are supported as they are passed to the compiler. This can be used to specify that the assembly should be signed and to specify the key file.|  
-|**/d**[**ebug**]|Generates an image that can be used with a debugger.|  
-|**/f[orce]**|Forces the overwriting of an existing assembly of the same name. The default is **false**.|  
+|**/a\[ssembly\]:**_filename_|Generates serialization code for all the types contained in the assembly or executable specified by *filename*. Only one file name can be provided. If this argument is repeated, the last file name is used.|  
+|**/c\[ompiler\]:** *options*|Specifies the options to pass to the C# compiler. All csc.exe options are supported as they are passed to the compiler. This can be used to specify that the assembly should be signed and to specify the key file.|
```
drewnoakes: Not related to this PR per-se, but is the space here invalid? It's at least inconsistent with other rows.

```suggestion
|**/c\[ompiler\]:**_options_|Specifies the options to pass to the C# compiler. All csc.exe options are supported as they are passed to the compiler. This can be used to specify that the assembly should be signed and to specify the key file.|
```

drewnoakes (approved):
I'm not familiar with the localisation issue you're referencing, but the content looks correct when rendered.

If this is GitHub flavored markdown then the `[` and `]` characters have to be escaped with `\`, which may be simpler. i.e. `**/c:[ompile]**` renders as **/c:[ompile]**.

I've suggested one other change that fixes an unrelated issue.

rpetrusha (approved):
This looks good, @v-maudel. I've made one change suggested by @drewnoakes, and I'll merge when the build completes.

v-maudel:
Thanks for your help @rpetrusha, @drewnoakes @mairaw

PR #9829, opened 2019-01-04
C57279: Correcting bold format and escaping square brackets

v-maudel:
Hello, @rpetrusha , @drewnoakes,
Description: There is problem with strings that are written for example like this one - "/a[ssembly]:". Based on the content in source it doesn't make sense to write it like this, because for example there is also string "/c[ompiler]:", which has brackets "[]" in bold as well. We suggest to keep it consistent and write all strings like the second one, so it works on localized languages too.

Many thanks in advance.

## Summary

Describe your changes here.


rpetrusha (approved):
I've made some additional formatting changes, so I'll merge once the build completes, @v-maudel..

mairaw (changes requested):
Please create a new PR targeting the master branch @v-maudel.

v-maudel:
Thanks @mairaw  for pointing out this was created in the Live Branch and sorryu for the inconveniences, I didn't notice it when I created the PR. 
I've created a new [PR](https://github.com/dotnet/docs/pull/9871) and also added the changes made by @rpetrusha.

regards.

rpetrusha:
Thanks for noticing that this targeted live, @mairaw. 
I'll close this PR now, since it's replaced by #9871.

PR #9830, opened 2019-01-04
Clarify relationship between project and directory names

drewnoakes:
The project name does not have to match the directory name, although they are usually created in this way. I feel this change makes it clearer where the default `<AssemblyName>` and `<PackageId>` properties come from.

A subject matter expert should confirm this however.


BillWagner (approved):
Thanks for making this change @drewnoakes We appreciate it.

Thanks for contributing again to docs!

PR #9853, opened 2019-01-05
Change to simplify doc

ghost:
## Summary

Describe your changes here.

### DIFF
--- a/docs/csharp/language-reference/keywords/virtual.md
+++ b/docs/csharp/language-reference/keywords/virtual.md
@@ -24,7 +24,7 @@ The implementation of a virtual member can be changed by an [overriding member](
 
 ## Remarks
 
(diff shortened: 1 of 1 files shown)


BillWagner:
Thanks for continuing to participate in docs @volkan-yazilim 

I reviewed this, and the original wording is more accurate than your change. I'm going to close this PR.

If you want, you can reopen it, and add comments about what you wanted to achieve. I'll help suggest updated wording.

volkan-yazilim:
Methods may be overriden many times via derived classes. But maybe the last one would not override the method and use the way last overriden state before itself . 

Base class B -> base method -> version 0
C:B -> override the method -> version 1
D:C -> override the method -> version 2
E:D -> not override the method -> version 2

In this case, this sentence is wrong. "The overriding member in the most derived class is called, which might be the original member, if no derived class has overridden the member."

I wanted to clear this by saying that "The very last overridden member is called, which might be the original member, if no derived class has overridden the member."

Thanks for your time.

Issue #9532, opened 2018-12-11
Performance recommendations for .NET Core applications


BillWagner:
Increased performance is a major part of the .NET Core story for developers.

We need to expand on that story in the .NET Core section of our docs. This should answer the following questions:

- What performance benefits are gained simply by migrating from .NET Framework to .NET Core?
- What techniques could produce further benefits?
- Of those, which (likely) have the best payback for common scenarios?
- How should customers measure the impact of adopting new .NET Core based performance techniques?

BillWagner:
I shared this with internal folks, and here are thoughts for what content should be created. 

Performance guidance and raising visibility on the changes made are two pretty different topics. In some cases, they coincide like ValueTask<T>, Pipelines and IAsyncEnumerable<T>. These things should be in different sections not try to put them in the same doc.

> What performance benefits are gained simply by migrating from .NET Framework to .NET Core?

This comes down to fundamental improvements in core technology, places that we’ve removed overhead, made fundamental operations go faster, etc., the kinds of things I talked about in https://blogs.msdn.microsoft.com/dotnet/2017/06/07/performance-improvements-in-net-core/ and https://blogs.msdn.microsoft.com/dotnet/2018/04/18/performance-improvements-in-net-core-2-1/, and that Mukul highlighted for Bing in https://blogs.msdn.microsoft.com/dotnet/2018/08/20/bing-com-runs-on-net-core-2-1/.  That’s all about coreclr/corefx, but I agree with Wade that ASP.NET is a key factor there as well, and there are many significant improvements at the ASP.NET layer.

> What techniques could produce further benefits?

This comes down to mainly:

- Understanding ways to rearchitect an app to take advantage of the new platform.  This would be more at the ASP.NET layer, taking advantage of the recommended patterns for structuring one’s app to best adapt to the patterns ASP.NET optimizes for.
- Taking advantage of new features / overloads / etc. exposed in the new platform.  This is things like Span<T>, all of the UTF8-related work in .NET Core, additional overloads on things like ThreadPool to let you optimize further, etc.

> How should customers measure the impact of adopting new .NET Core based performance techniques?
Measure, measure, measure.  And ensure the right things are being measured.  What are the key metrics that drive success in the app, create benchmarks for those, and tune to those.

brianrob:
Adding to the comment above - we should definitely have guidance on how to measure.  In the end, this is likely to be one of the most important pieces of performance guidance that we can provide.  I'm happy to work with you to get this done.

PR #9877, opened 2019-01-07
Extend String interpolation doc

OlehPetrenko:
## Extend String interpolation doc according to a new Roslyn's behaviour.

Added a new paragraph according to changes made in the Roslyn. 
Interpolated strings are not transformed into a String.Format method call all time. There is a common case when a String.Concat is used instead.

Please, see the issue #26612 https://github.com/dotnet/roslyn/pull/26612/commits/301eda3fe3618c5e39dd264d84fee1412a944f41 to additional information.

### DIFF
--- a/docs/csharp/tutorials/string-interpolation.md
+++ b/docs/csharp/tutorials/string-interpolation.md
@@ -26,7 +26,7 @@ As the example shows, you include an expression in an interpolated string by enc
 
 At compile time, an interpolated string is typically transformed into a <xref:System.String.Format%2A?displayProperty=nameWithType> method call. That makes all the capabilities of the [string composite formatting](../../standard/base-types/composite-formatting.md) feature available to you to use with interpolated strings as well.
 
-If all expressions are strings, and none of them have alignment or format specifiers, an interpolated string will be transformed into a more efficient string concatenation <xref:System.String.Concat%2A?displayProperty=nameWithType> at compile time.
+The compiler may substitute a <xref:System.String.Format%2A?displayProperty=nameWithType> for <xref:System.String.Concat%2A?displayProperty=nameWithType> if the analyzed behavior would be equivalent to concatenation.
 
 ## How to specify a format string for an interpolated expression
 
--- a/docs/csharp/tutorials/string-interpolation.md
+++ b/docs/csharp/tutorials/string-interpolation.md
@@ -26,6 +26,8 @@ As the example shows, you include an expression in an interpolated string by enc
 
 At compile time, an interpolated string is typically transformed into a <xref:System.String.Format%2A?displayProperty=nameWithType> method call. That makes all the capabilities of the [string composite formatting](../../standard/base-types/composite-formatting.md) feature available to you to use with interpolated strings as well.
 
+If all expressions are strings, and none of them have alignment or format specifiers, an interpolated string will be transformed into a more efficient string concatenation <xref:System.String.Concat%2A?displayProperty=nameWithType> at compile time.
+
 ## How to specify a format string for an interpolated expression
 
 You specify a format string that is supported by the type of the expression result by following the interpolated expression with a colon (":") and the format string:


docs/csharp/tutorials/string-interpolation.md
```diff
@@ -26,6 +26,8 @@ As the example shows, you include an expression in an interpolated string by enc
 
 At compile time, an interpolated string is typically transformed into a <xref:System.String.Format%2A?displayProperty=nameWithType> method call. That makes all the capabilities of the [string composite formatting](../../standard/base-types/composite-formatting.md) feature available to you to use with interpolated strings as well.
 
+If all expressions are strings, and none of them have alignment or format specifiers, an interpolated string will be transformed into a more efficient string concatenation <xref:System.String.Concat%2A?displayProperty=nameWithType> at compile time.
```
pkulikov: I wonder how much of the implementation details should be exposed in the docs. How does knowing it help me to *use* interpolated strings? Note, the paragraph above uses the "typically" word, so it already allows the implementation to use other method instead of `string.Format`. The main motivation behind mentioning of the `string.Format` method was to make a link to the composite formatting feature. The message is "hey, folks! you can easily convert your existing `string.Format` calls into the interpolated string; not a big deal, the core formatting features are the same". Looking at it now, I think that paragraph can be rephrased the following way:
> Interpolated strings support all the capabilities of the [string composite formatting](../../standard/base-types/composite-formatting.md) feature. That makes them a more readable alternative to the use of the `<xref:System.String.Format%2A?displayProperty=nameWithType>` method.

That way, the main idea is the same and implementation details are even more covered. And then, most of the next sections show how to use those "capabilities".

@OlehPetrenko @BillWagner thoughts?
@OlehPetrenko anyway, thanks for the PR; that way I've learnt about that optimization 👍

OlehPetrenko: Thanks for your review and good comment.

I thought about the "typically" word that you mentioned, but as for me this change in the Roslyn is too important to hide it under the "typically" word.
A lot of people think that string interpolation is just a syntactic sugar for a String.Format, but after the commit it is not a true, and it is not a easy to find right commit in the Roslyn to understand what happens.

> How does knowing it help me to use interpolated strings?

Just for for example, if you need to write a performance sensitive code and you want to use integer value as part of interpolated string. You can write 5.ToString() instead of 5 to avoid boxing and call String.Contac that is more efficient than String.Format. 

My point is, this paragraph can save time to understand why sometimes you see String.Concat instead of String.Format in the IL.

pkulikov: @OlehPetrenko thanks for the additional context.

Do I understand it correctly, that knowledge (compilation to the `string.Concat` call) is important, because one (one, working on the performance sensitive code, at least) might decide to use string interpolation instead of manually crafting calls to the `string.Concat`? Then, I think that knowledge is more valuable to answer WHETHER one should use string interpolation, rather than HOW to use string interpolation. The tutorial is about *how*; it might be a little bit late to provide the implementation detail.

If that info is to be included into the docs, then I think it should be mentioned in the main article about the feature. If I want to decide if to use the feature, I would start with the main article about it, not tutorial. The main article is:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated

Maybe, the separate section just before the "Additional resources" section should be added to discuss `string.Format`/`string.Concat`.

Also, the language version, from which this optimization is available, should be mentioned. If that knowledge is important to make any decisions from it, one also needs to know WHEN this information is relevant/correct. I guess, the implemented optimization is not for C#6 and C# 7.x, but available starting with C# 8.0? I mean, if I write code using C# 7.0, I still must write the calls to `string.Concat` myself?

OlehPetrenko: @pkulikov yes, you are right about performance sensivice code, otherwise this optimization would not be added to the Roslyn, I guess. Also, this knowledge can be useful for everyone who wants to know what happens under the hood.

I've decided to add this paragraph here, because here is the place where mentioned how it works at compile time. A lot of information exists at the StackOverFlow and other resources where it says that interpolation is the same thing with String.Format. It was true up to a #26612 issue, but now it may confuse. The official documentation is the best way to help people.

I think, the paragraph that I suggested is just the easiest way to provide people with this new behavior, but separate section it is also a good idea, but maybe it could be redundant.

Anyway, if my paragraph is not be approved, it would be great if existed paragraph about String.Format would be rephrased.

It is not depends on language version. Only depends on Roslyn's version. I've checked on 2.9.0.63208 and this oprimization works.

BillWagner:
Thanks for writing this @OlehPetrenko 

Adding @jcouv and @agocke for  spec related question. 

My preference is to document only the behavior that is guaranteed by the spec in the C# language docset. I don't this this optimization is guaranteed in the spec. Andy or Julien can comment on that. The team makes optimizations often, and we can't keep those updated appropriately.

That said, we don't want to ignore those optimizations entirely. I want to find the right location for this kind of guidance based on the upcoming work for #9532 

I'll keep this open at this time, pending notes from the LDM team on what the spec will say.

docs/csharp/tutorials/string-interpolation.md
pkulikov: @OlehPetrenko 

> I've decided to add this paragraph here, because here is the place where mentioned how it works at compile time.

>Anyway, if my paragraph is not be approved, it would be great if existed paragraph about String.Format would be rephrased.

That makes sense. I think the existing paragraph should be rephrased in any case. I've made an attempt in my first comment, but maybe it's better to address it in another PR.

Implementation details, both about `string.Format` and `string.Concat`, fit better another article. At least, the tutorial is not the best place for them, I feel.

>It doesn't depend on language version. Only depends on Roslyn's version. I've checked on 2.9.0.63208 and this oprimization works.

Thanks for the check and clarification. It's then about the tools installed (.NET Core SDK or Visual Studio).

By the way, have you noticed the follow-up PR dotnet/roslyn#27411 that limits the optimization scope?

agocke:
I think we actually have to spec something here. The draft spec for C# 6 says: "If the type is string, the meaning of the expression is a call to string.Format"

Moreover, this isn't really an optimization because it's trivially observable if you write your own `string.Format`. I'm not sure what language we'll choose, but something vague but precise like, "the compiler may substitute a string.Format for string.Concat if the analyzed behavior would be equivalent to concatenation" is probably safe.

agocke:
cc @gafter for the above: I think we took a language change as an optimization and we should run this through LDM quickly.

gafter:
The LDM explicitly discussed and permitted such optimizations.  See https://github.com/dotnet/csharplang/blob/98043cdc889303d956d540d7ab3bc4f5044a9d3b/meetings/2014/LDM-2014-10-15.md#semantics-1 and the words "or equivalent code".

However, no such optimizations are guaranteed by the specification and they therefore should not be documented as part of the semantics of the language feature. At least, not in their particulars. At most the documents should say that the compiler is permitted to produce equivalent code without actually calling `string.Format`.

docs/csharp/tutorials/string-interpolation.md
OlehPetrenko: Yes, I've seen it. 

Thanks for your comments.

OlehPetrenko:
@gafter thanks a lot for your comment.

> Having said that, the compiler is permitted to produce equivalent code for just about everything.

That's a good way to keep documentation clear and actual.

However, the main idea of this PR is to help people to understand what happens under the hood of string interpolation. There are a lot of articles and answers in the network where they are said that string interpolation is just a syntactic sugar for `string.Format`. From my point of view, the article I want to change can become a kind of core of this.

> At compile time, an interpolated string is typically transformed into a `String.Format`.

This sentence can confuse a bit. I have spent a couple of hours to find a right commit with optimization we discuss to understand why sometimes IL gives me a `string.Format`  and sometimes `string.Concat`.

When you google something about string interpolation and `string.Format` you will take this article in most cases. I guess, it would be great to have here some actual information about optimization to help people fix unclear information got from tons of articles and answers at the StackOverFlow and others.

What do you think about that?

BillWagner:
@OlehPetrenko 

> When you google something about string interpolation and string.Format you will take this article in most cases. I guess, it would be great to have here some actual information about optimization to help people fix unclear information got from tons of articles and answers at the StackOverFlow and others.

> What do you think about that?

That idea has merit. I'm still of the mind that the detailed performance optimizations should be covered in the work for #9532 (which is why I keep tagging it). Someone starting to do serious performance tuning would want to find this information, and other idioms that could be affecting perf at that level.

I like @agocke 's suggested change to this PR. If you update it to match, I'll approve and merge it. 

Thanks for starting this conversation, making the fix, and engaging with us to make the experience better.

BillWagner (approved):
Thanks for all the discussion @OlehPetrenko and @pkulikov 

I'll :shipit: now.

Thanks again.

PR #9852, opened 2019-01-05
Fix xml Blocks

msschl:
## Summary

Contributes to #8125

Fixes #8125

PR #9886, opened 2019-01-08
Fix XML code blocks

nxtn:
Fix XML attributes.

See #9744


Thraka:
@NextTurn I reviewed this and it looks good. But something else happened that is causing a conflict.

docs/framework/configure-apps/file-schema/network/defaultproxy-element-network-settings.md
```diff
@@ -19,13 +19,12 @@ Configures the Hypertext Transfer Protocol (HTTP) proxy server.
...
   
 ```xml  
-      <defaultProxy  
-        enabled="true|false"  
-        useDefaultCredentials="true|false">  
-           <bypasslist> … </bypasslist>  
-           <proxy> … </proxy>  
-           <module> … </module>  
-      </defaultProxy>
+<defaultProxy enabled="true|false"
```
karelz: Can you please use original indentation (4 spaces for subelements and 2 spaces for attributes) and line wrapping? (multi-line attributes start on new line) I assume it is consistent with other docs.

(applies to multiple occurrences later on as well)

docs/framework/network-programming/enabling-and-disabling-ipv6.md
```diff
@@ -23,10 +23,10 @@ To use the IPv6 protocol, ensure that you are running a version of the operating
  The .NET Framework has a configuration switch set in a configuration file as follows  
   
 ```xml  
-<system.net>…  
-    <settings>…  
-        <ipv6 enabled="true"/>…  
-    </settings>…  
+<system.net>...
```
karelz: Nit: I would put each `...` on new line, unless it is in the form `<elem>...</elem>`.

docs/framework/wcf/samples/message-security-sample.md
```diff
@@ -23,7 +23,7 @@ This sample demonstrates how to implement an application that uses the `basicHtt
                bindingConfiguration="Binding1"   
                contract="Microsoft.ServiceModel.Samples.ICalculator" />  
     </service>  
-  </services>   …  
+  </services>...
```
karelz: `...` on new line, it can be easily misses behind the closing elem tag.

karelz (commented):
Nice changes! You found quitre a few mistakes, thanks!
Couple of minor comments above.

karelz:
BTW: I would suggest to stop adding additional changes which are not part of feedback. Otherwise the review will become messy ...
You can submit another PR with additional mistakes you find ...

docs/framework/configure-apps/file-schema/network/defaultproxy-element-network-settings.md
NextTurn: On the contrary, I think my changed styles are consistent with other docs.

See also #9852

NextTurn:
I'm sorry that I haven't seen your comments at that moment ...

docs/framework/configure-apps/file-schema/network/defaultproxy-element-network-settings.md
NextTurn: According to the default behaviors of `XmlWriterSettings.NewLineOnAttributes`, it could be better to use 2 spaces indentation for both subelements and attributes, and start a new line for all attributes if there are any multi-line ones.

docs/framework/configure-apps/file-schema/wcf/httpstransport.md
```diff
@@ -24,12 +24,11 @@ Specifies an HTTP transport for transmitting SOAP messages for a custom binding.
                 maxBufferSize="Integer"
                 maxReceivedMessageSize="Integer"
                 proxyAddress="Uri"
-                proxyAuthenticationScheme="None/Digest/Negotiate/Ntlm/Basic/Anonymous"
+                proxyAuthenticationScheme="None/Digest/Negotiate/Ntlm/Basic/Anonymous/IntegratedWindowsAuthentication"
```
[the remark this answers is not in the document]
NextTurn: https://github.com/dotnet/docs/blob/master/docs/framework/configure-apps/file-schema/wcf/httpstransport.md#attributes-and-elements

docs/framework/configure-apps/file-schema/wcf/endpoint-element.md
```diff
@@ -72,8 +72,8 @@ Specifies binding, contract, and address properties for a service endpoint, whic
           binding="customBinding"
           contract="HelloWorld">
   <headers>
-    <region xmlns="http://tempuri.org/">EastCoast</region>
-    <member xmlns="http://tempuri.org/">Gold</member>
+    <Region xmlns="http://tempuri.org/">EastCoast</Region>
```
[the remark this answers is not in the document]
karelz: Did you confirm it is not a bug in the table that it is capitalized there?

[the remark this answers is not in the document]
NextTurn: They were originally capitalized and incorrectly changed in #9852. I've reviewed this PR and reverted these changes.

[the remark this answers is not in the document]
karelz: Just the fact they were capitalized before does not mean it was the right value ... we should look at how they are defined in code / XML schema IMO.

docs/framework/configure-apps/file-schema/network/defaultproxy-element-network-settings.md
NextTurn: I would expect a consensus since there are unformatted blocks elsewhere that need to be fixed.

karelz: That should be separate discussion IMO.

docs/framework/configure-apps/file-schema/wcf/endpoint-element.md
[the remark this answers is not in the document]
mairaw: @Lxiamail @zhenlan where can we check if this is correct?
In VS, I see totally different elements under \<headers>.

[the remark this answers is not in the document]
Thraka: I think headers is an open node, you define what you want to be included in the http request/response headers?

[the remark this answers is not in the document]
mconnew: The problem here is the Child Elements table shouldn't be there. From the documentation text it says:

>  Entries in the endpoint header collection are user-defined XML elements. Each element has to be well-formed XML.

The member and region elements were just examples of how you can add a custom header, the names can be arbitrary as long as it's valid xml. This specific change should be reverted and instead remove the child elements table.

docs/framework/configure-apps/file-schema/wcf/headers.md
```diff
@@ -14,8 +14,8 @@ An endpoint can be addressed by one or more SOAP headers in addition to its basi
   
 ```xml  
 <headers>
-  <region xmlns="Uri">"String"</region>
-  <member xmlns="Uri">"String"</member>
+  <Region xmlns="Uri">"String"</Region>
+  <Member xmlns="Uri">"String"</Member>
```
mconnew: Same fix here. Revert this change and instead remove the Child Elements table.

docs/framework/configure-apps/file-schema/wcf/httpstransport.md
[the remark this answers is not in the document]
mconnew: I'm pretty certain this is incorrect. If you look at [this code ](https://referencesource.microsoft.com/#System.ServiceModel/System/ServiceModel/HttpProxyCredentialType.cs,29)you can see that it will throw an exception if anything other than Anonymous, Basic, Digest, Ntlm or Negotiate is specified.

docs/framework/wcf/feature-details/transaction-protocols.md
```diff
@@ -282,8 +282,8 @@ Windows Communication Foundation (WCF) implements WS-Atomic Transaction and WS-C
 <wst:RequestSecurityTokenResponse   
 xmlns:wsse="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-secext-1.0.xsd"   
 xmlns:wssu="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-     wssecurity-utility-1.0.xsd"
```
mconnew: The namespace definition here is also wrong. There's whitespace in the middle of the uri

docs/framework/configure-apps/file-schema/wcf/httpstransport.md
mconnew: Commenting again as previous conversation was marked as resolved.

I'm pretty certain this is incorrect. If you look at [this code](https://referencesource.microsoft.com/#System.ServiceModel/System/ServiceModel/HttpProxyCredentialType.cs,29) you can see that it will throw an exception if anything other than Anonymous, Basic, Digest, Ntlm or Negotiate is specified.

NextTurn: Is the table also incorrect?

docs/framework/configure-apps/file-schema/wcf/headers.md
NextTurn: Is it OK to remove the Child Elements section?

docs/framework/configure-apps/file-schema/wcf/httpstransport.md
mconnew: I believe so. I'll need to write some test code to verify in case I've missed something. I'll update this conversation once I have done so.

mconnew: I just verified and we throw an exception if `IntegratedWindowsAuthentication` is specified. This is because it's value is actually `Negotiate | Ntlm` and we only allow a single authentication scheme to be specified. So it does need to be removed from the table. It might be worth adding a note after the where it says this is of type AuthenticationSchemes that this value isn't supported.

mconnew (approved):
All the WCF and serialization changes look good. Thanks for cleaning these up.

docs/framework/wcf/feature-details/using-the-nethttpbinding.md
```diff
@@ -39,7 +39,7 @@ ms.assetid: fe134acf-ceca-49de-84a9-05a37e3841f1
         </binding>  
       </netHttpBinding>  
     </bindings>  
-    <!- ... -->   
+    <!-- ... -->
```
Thraka: You're using ... in other places without the XML comment, so I would do the same here.

Thraka:
Final two things to fix and then I think this is good!

Thraka (changes requested):
Final two things to fix and then I think this is good!

Thraka (approved):

PR #9888, opened 2019-01-08
Added missing function in example

cacamille3:
Just added the missing function


cacamille3:
Actually not needed...

PR #9916, opened 2019-01-09
fix delegate subtraction has unpredictable warning

Kevinlinpr:
Avoid the delegate subtraction has unpredictable result warning.

### DIFF
--- a/docs/csharp/programming-guide/events/codesnippet/CSharp/how-to-use-a-dictionary-to-store-event-instances_1.cs
+++ b/docs/csharp/programming-guide/events/codesnippet/CSharp/how-to-use-a-dictionary-to-store-event-instances_1.cs
@@ -2,33 +2,44 @@
     using System.Collections.Generic;    
 
     public delegate void EventHandler1(int i);
+
     public delegate void EventHandler2(string s);
 
     public class PropertyEventsSample
     {
-        private Dictionary<string, Delegate> eventTable;
-
+        private readonly Dictionary<string, Delegate> _eventTable;
+        private readonly List<EventHandler1> _event1List = new List<EventHandler1>();
+        private readonly List<EventHandler2> _event2List = new List<EventHandler2>();
         public PropertyEventsSample()
         {
-            eventTable = new Dictionary<string, Delegate>();
-            eventTable.Add("Event1", null);
-            eventTable.Add("Event2", null);
+            _eventTable = new Dictionary<string, Delegate>
+            {
+                {"Event1", null},
+                {"Event2", null}
+            };
         }
 
         public event EventHandler1 Event1
         {
             add
             {
-                lock (eventTable)
+                _event1List.Add(value);
+                lock (_eventTable)
                 {
-                    eventTable["Event1"] = (EventHandler1)eventTable["Event1"] + value;
+                    _eventTable["Event1"] = (EventHandler1) _eventTable["Event1"] + value;
                 }
             }
             remove
             {
-                lock (eventTable)
+                if (!_event1List.Contains(value)) return;
+                _event1List.Remove(value);
+                lock (_eventTable)
                 {
-                    eventTable["Event1"] = (EventHandler1)eventTable["Event1"] - value;
+                    _eventTable["Event1"] = null;
+                    foreach (var event1 in _event1List)
+                    {
+                        _eventTable["Event1"] = (EventHandler1) _eventTable["Event1"] + event1;
+                    }
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks for spotting this and making the update @Kevinlinpr We appreciate it.

I've reviewed the changes, and I've approved this PR. I'll :shipit: now.

@rpetrusha I looked and I did not see a corresponding VB sample. If you know of one, post the link and I'll update that sample.

docs/csharp/programming-guide/events/codesnippet/CSharp/how-to-use-a-dictionary-to-store-event-instances_1.cs
```diff
@@ -2,33 +2,44 @@
...
+                {"Event2", null}
+            };
         }
 
         public event EventHandler1 Event1
         {
             add
             {
-                lock (eventTable)
+                _event1List.Add(value);
```
pkulikov: One more remark: by the same reasoning as access to `_eventTable` is synchronized with the `lock` statement, shouldn't the access to `_event1List` (and `_event2List`) be synchronized as well?

rpetrusha:
There is no corresponding VB example, @BillWagner.

The comments from @pkulikov should be considered, though.

docs/csharp/programming-guide/events/codesnippet/CSharp/how-to-use-a-dictionary-to-store-event-instances_1.cs
```diff
@@ -37,63 +48,70 @@ public PropertyEventsSample()
...
 
         internal void RaiseEvent1(int i)
         {
-            EventHandler1 handler1;
-            if (null != (handler1 = (EventHandler1)eventTable["Event1"]))
+            lock (_eventTable)
             {
-                handler1(i);
+                var handler1 = (EventHandler1) _eventTable["Event1"];
+                handler1?.Invoke(i);
```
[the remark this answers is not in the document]
BillWagner: Yes, really good catch @pkulikov 

I'll dismiss my review, because this should be fixed.

docs/csharp/programming-guide/events/codesnippet/CSharp/how-to-use-a-dictionary-to-store-event-instances_1.cs
BillWagner: Also correct. This makes it more complicated because it will have to acquire multiple structures in a single locked region.

docs/csharp/programming-guide/events/codesnippet/CSharp/how-to-use-a-dictionary-to-store-event-instances_1.cs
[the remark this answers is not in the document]
pkulikov: @BillWagner I'll do it.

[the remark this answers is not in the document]
BillWagner: thank you.

docs/csharp/programming-guide/events/codesnippet/CSharp/how-to-use-a-dictionary-to-store-event-instances_1.cs
pkulikov: @BillWagner can we go here with only one lock object? I mean: move the `lock (_eventTable)` line one line up?

BillWagner: Yes.

I think the lock on Line 36 (in the corresponding remove) should also move up one line to cover the `eventList.Remove()` call.

Also, a similar change to the `EventHandler2` `add` and `remove` accessors. Do you agree?

pkulikov: Yes. And not to confuse readers, I'll just introduce a dedicated `handlerStorageLock` object.

pkulikov:
@BillWagner @Kevinlinpr I think this PR should be reverted.

> Avoid the delegate subtraction has unpredictable result warning

I guess, it's not C# compiler warning, but ReSharper warning:
http://www.jetbrains.com/help/resharper/2018.3/DelegateSubtraction.html

The following StackOverflow answer explains that this warning might be ignored in case of implementing events with `add`/`remove` keywords:
https://stackoverflow.com/questions/11180068/delegate-subtraction-has-unpredictable-result-in-resharper-c

Moreover, in #9937 I have remark that the updated sample doesn't fit the containing article anymore.

Kevinlinpr:
> Thanks to @pkulikov your solution of lock keyword problem. 

This "delegate subtraction has unpredictable" warning is far more serious than we think, and it will cause operational warning.Please see my latest pull request[Solution of delegate subtraction has unpredictable.#9953](https://github.com/dotnet/docs/pull/9953) without allocating a field for each event.

PR #9932, opened 2019-01-10
Update values to match code snippet

emagers:
The text says '9129' did not match the regular expression, but that was not one of the values being evaluated. The correct value in the code snippet is '9219'.


rpetrusha (approved):
Thank you, @emagers, for contributing to the dotnet/docs repo and correcting the incorrect value. We'll merge your PR now. The change should be live on docs.microsoft.com in the next day or two.

PR #9937, opened 2019-01-10
WIP - Refactored sample to fix locking

pkulikov:
Follow up of #9916

@BillWagner this PR fixes locking issues in the updated sample. However, I have a doubt that the new sample fits the purpose of its article.

The [How to: Use a Dictionary to Store Event Instances](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/events/how-to-use-a-dictionary-to-store-event-instances) says:
> One use for accessor-declarations is to expose many events without allocating a field for each event, but instead using a Dictionary to store the event instances.

In an updated sample, we allocate a field for each event (for every list that contains event handlers). Is my understanding correct?

### DIFF
--- a/docs/csharp/programming-guide/events/codesnippet/CSharp/how-to-use-a-dictionary-to-store-event-instances_1.cs
+++ b/docs/csharp/programming-guide/events/codesnippet/CSharp/how-to-use-a-dictionary-to-store-event-instances_1.cs
@@ -1,125 +1,123 @@
-    using System;    
-    using System.Collections.Generic;    
+using System;
+using System.Collections.Generic;
 
-    public delegate void EventHandler1(int i);
+public delegate void EventHandler1(int i);
+public delegate void EventHandler2(string s);
 
-    public delegate void EventHandler2(string s);
+public class PropertyEventsSample
+{
+    private readonly object handlerStorageLock = new object();
 
-    public class PropertyEventsSample
+    private readonly Dictionary<string, Delegate> handlers;
+    private readonly List<EventHandler1> event1Handlers = new List<EventHandler1>();
+    private readonly List<EventHandler2> event2Handlers = new List<EventHandler2>();
+
+    public PropertyEventsSample()
     {
-        private readonly Dictionary<string, Delegate> _eventTable;
-        private readonly List<EventHandler1> _event1List = new List<EventHandler1>();
-        private readonly List<EventHandler2> _event2List = new List<EventHandler2>();
-        public PropertyEventsSample()
+        handlers = new Dictionary<string, Delegate>
         {
-            _eventTable = new Dictionary<string, Delegate>
-            {
-                {"Event1", null},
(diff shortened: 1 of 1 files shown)


pkulikov:
Set it to WIP, until discussion in  #9916 is finished.

BillWagner:
@pkulikov I think we've finished the discussion on #9916 

Do you still have questions to resolve? Should I give this a final review?

pkulikov:
@BillWagner I don't think we finished discussion. Or I might have missed the answer on [this comment](https://github.com/dotnet/docs/pull/9916#issuecomment-453289762).

Moreover, in [this comment](https://github.com/dotnet/docs/pull/9953#issuecomment-453684403) I observe that the current live version is different from the original one. That is one more rationale for reverting #9916 in order to come back to the original point and maybe start discussion on how and if the original example should be modified.

As for this PR, I think it might be closed without merging. Also #9953 partially duplicates this one.

What do you think?
In general, the story has become complicatedly spread among three PRs.

BillWagner:
Thanks @pkulikov Adding @Kevinlinpr 

I went through the current content and the open PRs: this one, #9937, and #9953

The final fix should do the following:

1. Use a `Dictionary<string, Delegate>` or `Dictionary<string, MulticastDelegate>` to store the multiple event handlers. 
1. The example that calls this implementation should keep a reference to the delegates used to subscribe to events so they correctly unsubscribe.
1. Add the locking logic that @pkulikov put in place.
1. Implement the `Add` and `Remove` functionality using the `Dictionary` above. Ideally, this should remove the two `List<eventHandler>` collections, which stores duplicated information that must be kept in sync.  (I'm not sure the lists can be removed safely, but it looks like a bad design).

First, do you both agree? (Up vote or down vote). Offer any suggestions for improvement.

Then, let's figure out which PR(s) to update and merge, and which to close.

Thanks for the patience as we sorted through this. You've both raised important issues on this article.

pkulikov:
@BillWagner thanks for going through the PRs. I agree with your suggestion for the final fix.

Kevinlinpr:
@BillWagner I agree with your solution.

BillWagner:
Thanks @pkulikov @Kevinlinpr 

Should we close both PRs and make a new, clean PR? Or is either this PR or #9953 closer to what we want?

pkulikov:
@BillWagner @Kevinlinpr two PRs in question must be closed anyway, because they edit deleted source code files. All *codesnippet* folders have been removed from this repo recently.

Instead, a new PR should be open in the **dotnet/samples** repo to update the following lines:
https://github.com/dotnet/samples/blob/master/snippets/csharp/VS_Snippets_VBCSharp/csProgGuideEvents/CS/Events.cs#L455-L581

I can do the necessary changes, If both of you are fine with that.

BillWagner:
closing based on https://github.com/dotnet/docs/pull/9937#issuecomment-469331672

PR #9953, opened 2019-01-11
WIP - Solution of delegate subtraction has unpredictable.

Kevinlinpr:
Solution of delegate subtraction has unpredictable.

## Summary

> First of all, thanks to @pkulikov [#9937](https://github.com/dotnet/docs/pull/9937) fix the lock keyword problem. 

This sample solves the following problem:

- Completely avoid "delegate subtraction has unpredictable" warning using the Delegate.Remove() function instead of directly subtracting by the "-" operator.

- If you subtract the delegate list directly using the "-" operator, "delegate subtraction has unpredictable" warning will come out, because of the following situation 
```
static void Main()
{
  Action a = () => Console.Write("A");
  Action b = () => Console.Write("B");
  Action c = () => Console.Write("C");
  Action s = a + b + c + Console.WriteLine;
  s();                  //ABC
  (s - a)();            //BC
  (s - b)();            //AC
  (s - c)();            //AB
  (s - (a + b))();      //C
  (s - (b + c))();      //A
  (s - (a + c))();      //ABC
}
```

> In the code below, removing the A and C delegates from the ABC invocation list via the expression s - (a + c) has no effect whatsoever, because AC is not a sublist of ABC.
> So I create functions to retrieve step by step via the Delegate.Equals() function and consider whether the delegate item adding through the accessor-declarations is necessary for the event list avoiding repeated additions.

- Can add or delete delegate event lists regardless of order.

- The same delegate event can only be added once, and the method will gradually find out if there is already an item to be added.

- It will also delete all the same delegate events in the list.
```
            p.Event1 += handler1Method1;                        //Method1: 2
            p.Event1 += combinationHandler1List;                //Method1: 2/Method3: 2/Method2: 2
            p.Event1 += (handler1Method3 + handler1Method2);    //Method1: 2/Method3: 2/Method2: 2
            p.Event1 -= (handler1Method2 + handler1Method3);    //Method1: 2
            p.RaiseEvent1(2);
            p.Event2 -= handler2Method3;                        //
            p.Event2 += (handler2Method3 + handler2Method1);    //Method3: TestString/Method1: TestString
            p.Event2 -= combinationHandler2List;                //
```
Method:

- Add addList to eventList
```
        /// <summary>
        /// Add addList to eventList
        /// </summary>
        /// <param name="eventList">the delegate list which you want to add into</param>
        /// <param name="addList">the delegate list which you want to add</param>
        /// <returns>Return the latest delegate list after the adding operation</returns>
        private static Delegate AddOperator(Delegate eventList, Delegate addList)
        {
            return addList.GetInvocationList().Aggregate(eventList, AddEvent);
        }
```

- Add each addTarget item into the eventList
```
        /// <summary>
        /// Add each addTarget item into the eventList
        /// </summary>
        /// <param name="eventList">the delegate list which you want to add into</param>
        /// <param name="addTarget">the delegate item which you want to add</param>
        /// <returns>Return the latest delegate list after the adding operation</returns>
        private static Delegate AddEvent(Delegate eventList, Delegate addTarget)
        {
            var resultList = eventList;
            if (eventList != null)
            {
                if (!eventList.GetInvocationList().Contains(addTarget))
                    resultList = Delegate.Combine(eventList, addTarget);
            }
            else
                resultList = addTarget;
            return resultList;
        }
```

- Remove removeList from eventList

```
        /// <summary>
        /// Remove removeList from eventList
        /// </summary>
        /// <param name="eventList">the delegate list which you want to remove its item from</param>
        /// <param name="removeList">the delegate list which you want to remove</param>
        /// <returns>Return the latest delegate list after the removing operation</returns>
        private static Delegate RemoveOperator(Delegate eventList, Delegate removeList)
        {
            return removeList.GetInvocationList().Aggregate(eventList, RemoveEvent);
        }
```

- Remove each removeTarget item from the eventList

```
        /// <summary>
        /// Remove each removeTarget item from the eventList
        /// </summary>
        /// <param name="eventList">the delegate list which you want to remove its item from</param>
        /// <param name="removeTarget">the delegate item which you want to remove</param>
        /// <returns>Return the latest delegate list after the removing operation</returns>
        private static Delegate RemoveEvent(Delegate eventList, Delegate removeTarget)
        {
            var returnList = eventList;
            if (eventList == null) return returnList;
            while (true)
            {
                var resultList = eventList;
                foreach (var eventItem in eventList.GetInvocationList())
                {
                    if (!eventItem.Equals(removeTarget)) continue;
                    resultList = Delegate.Remove(eventList, eventItem);
                    break;
                }

                var continueRetrieve = resultList != null && resultList.GetInvocationList().Contains(removeTarget);
                if (continueRetrieve)
                {
                    eventList = resultList;
                    continue;
                }
                returnList = resultList;
                break;
            }
            return returnList;
        }
```

### DIFF
--- a/docs/csharp/programming-guide/events/codesnippet/CSharp/how-to-use-a-dictionary-to-store-event-instances_1.cs
+++ b/docs/csharp/programming-guide/events/codesnippet/CSharp/how-to-use-a-dictionary-to-store-event-instances_1.cs
@@ -1,21 +1,24 @@
-    using System;    
-    using System.Collections.Generic;    
+using System;
+using System.Collections.Generic;
+using System.Linq;
 
+namespace ConsoleApplication1
+{
     public delegate void EventHandler1(int i);
-
     public delegate void EventHandler2(string s);
 
     public class PropertyEventsSample
     {
-        private readonly Dictionary<string, Delegate> _eventTable;
-        private readonly List<EventHandler1> _event1List = new List<EventHandler1>();
-        private readonly List<EventHandler2> _event2List = new List<EventHandler2>();
+        private readonly object _handlerStorageLock = new object();
+
+        private readonly Dictionary<string, Delegate> _handlers;
+
         public PropertyEventsSample()
         {
-            _eventTable = new Dictionary<string, Delegate>
+            _handlers = new Dictionary<string, Delegate>
             {
-                {"Event1", null},
-                {"Event2", null}
+                ["Event1"] = null,
+                ["Event2"] = null
             };
         }
 
@@ -23,23 +26,18 @@
         {
             add
             {
-                _event1List.Add(value);
-                lock (_eventTable)
+                lock (_handlerStorageLock)
                 {
-                    _eventTable["Event1"] = (EventHandler1) _eventTable["Event1"] + value;
+                    var addList = value;
+                    _handlers["Event1"] = AddOperator(_handlers["Event1"], addList);
                 }
             }
             remove
             {
-                if (!_event1List.Contains(value)) return;
-                _event1List.Remove(value);
-                lock (_eventTable)
+                lock (_handlerStorageLock)
                 {
-                    _eventTable["Event1"] = null;
-                    foreach (var event1 in _event1List)
-                    {
-                        _eventTable["Event1"] = (EventHandler1) _eventTable["Event1"] + event1;
-                    }
+                    var removeList = value;
+                    _handlers["Event1"] = RemoveOperator(_handlers["Event1"], removeList);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


pkulikov:
@Kevinlinpr I'm afraid it's not refactoring.

The code being updated serves as an example to the [How to: Use a Dictionary to Store Event Instances](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/events/how-to-use-a-dictionary-to-store-event-instances) article, which says:
> One use for accessor-declarations is to expose many events without allocating a field for each event, but instead using a Dictionary to store the event instances. This is only useful if you have many events, but you expect most of the events will not be implemented.

In particular, that means that the implementation with `Dictionary` must behave exactly as the implementation without `Dictionary` (at the end, using `Dictionary` is the implementation detail). The implementation without dictionary is as follows:
```csharp
using System;

public delegate void EventHandler1(int i);
public delegate void EventHandler2(string s);

public class PropertyEventsSample
{
    public event EventHandler1 Event1;
    public event EventHandler2 Event2;

    public PropertyEventsSample() { }

    internal void RaiseEvent1(int i) => Event1?.Invoke(i);
    internal void RaiseEvent2(string s) => Event2?.Invoke(s);
}
```

When the above defined `PropertyEventsSample` class is used in the `TestClass.Main` call from this PR, the output is as follows
```text
Method1: 2/Method1: 2/Method3: 2/Method2: 2/Method3: TestString/Method1: TestString/
```

It doesn't behave as the new code does:
```text
Method1: 2
```

By the way, the current code (the one merged in #9916) outputs:
```text
Method1: 2/Method1: 2/Method3: 2/Method2: 2/Method3: 2/Method2: 2/Method3: TestString/Method1: TestString/
```

That means it was not refactoring as well.

The original code (with a dictionary only) produces the following output:
```text
Method1: 2/Method1: 2/Method3: 2/Method2: 2/Method3: TestString/Method1: TestString/
```

which corresponds to the solution without a dictionary, with a field per event.

Kevinlinpr:
Set it to WIP, until the problem is completely solved.Thank you so much @pkulikov.

BillWagner:
This and #9937 both get close to but don't discuss a common issue with adding and removing delegates. It came up in a different context in a question I answered [on my blog](http://thebillwagner.com/Blog/Item/2015-10-13-LambdaSyntaxandPerformance) a few years ago.

Thanks for setting both of these to WIP status. I'm going to chat internally with a few folks and then be back with a proposal. Hopefully by Wednesday.

pkulikov:
@BillWagner 

> ...don't discuss a common issue with adding and removing delegates.

Do you mean the issue the following section discusses:
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/events/how-to-subscribe-to-and-unsubscribe-from-events#to-subscribe-to-events-by-using-an-anonymous-method

>It is important to notice that you cannot easily unsubscribe from an event if you used an anonymous function to subscribe to it. To unsubscribe in this scenario, it is necessary to go back to the code where you subscribe to the event, store the anonymous method in a delegate variable, and then add the delegate to the event. In general, we recommend that you do not use anonymous functions to subscribe to events if you will have to unsubscribe from the event at some later point in your code.

BillWagner:
closing based on https://github.com/dotnet/docs/pull/9937#issuecomment-469331672

PR #10008, opened 2019-01-16
 Description is not very precise.

asthman666:
Static class contains only static members.


jzeferino (approved):

BillWagner (approved):
Thanks for making this addition @asthman666 We appreciate it.

PR #10064, opened 2019-01-19
Standardize See also sections

nxtn:
- Add `- `
- Replace `* ` with `- `
- Trim spaces

Contributes to #6923 

Hide white space changes for easier diff.

If there is no problem, can I resolve the issue within a huge PR?


mairaw (approved):
That looks great @NextTurn! Thanks so much.

As for your question, how large are we talking about? For straightforward changes like this, we could perhaps do in one PR.

NextTurn:
There are about 5K files.

AvaGlobalgithub:
excellent!

PR #10068, opened 2019-01-19
Update instance-constructors.md

fredeil:
## Summary

#9763 changed the class/struct name in the snippets but not in the docs

PR #9763, opened 2018-12-27
Fix the CoOrds class

Thraka:
## Summary

Fix the CoOrds class name.

Fixes #9558

PR #10069, opened 2019-01-19
Update using-structs.md

fredeil:
## Summary

#9763 changed the class/struct name in the snippets but not in the docs

also see #10068


BillWagner (approved):
Thanks again @fredeil 

I'll :shipit: this PR as well.

Thanks for making docs better.

PR #10115, opened 2019-01-23
CCI6636: Replacing HTML entities by characters

v-maudel:
Hello, @mairaw ,
Description: HTML entities are not rendering properties in Metadata. Also we suggest to change the entities in Title for consistency.

Many thanks in advance.

Link to page: [\<value> (Visual Basic)](https://review.docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/xmldoc/value?branch=pr-en-us-10115)


rpetrusha (approved):
Your fix didn't work, @v-maudel; the `<` has to be escaped. I've made the change, and I'll merge once the build completes successfully.

PR #10131, opened 2019-01-24
Small clarification about the selected ML Task for tutorial

CESARDELATORRE:
IT's not clear that the ML Task to use is "Multiclass classification".
The text just exposes "classification Task" and its possible types.
Clearly, a "Binary classification" doesn't match this particular scenario but a "multiclass classification"

### DIFF
--- a/docs/machine-learning/tutorials/github-issue-classification.md
+++ b/docs/machine-learning/tutorials/github-issue-classification.md
@@ -102,6 +102,8 @@ Classification tasks are frequently one of the following types:
 * Binary: either A or B.
 * Multiclass: multiple categories that can be predicted by using a single model.
 
+For this particular Github issues classification, you clearly need to use a Multiclass classification task, since the possible issue categories to choose from are multiple categories rather than just two (binary).
+
 ## Create a console application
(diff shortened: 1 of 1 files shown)


docs/machine-learning/tutorials/github-issue-classification.md
```diff
@@ -102,6 +102,8 @@ Classification tasks are frequently one of the following types:
 * Binary: either A or B.
 * Multiclass: multiple categories that can be predicted by using a single model.
 
+For this particular Github issues classification, you clearly need to use a Multiclass classification task, since the possible issue categories to choose from are multiple categories rather than just two (binary).
```
JRAlexander: How about "For this type of problem, use a Multiclass classification task, since your issue category prediction can be one of multiple categories (multiclass) rather than just two (binary)."

JRAlexander (changes requested):
Hi Cesar, 
Great addition! I've edited an example a bit for readability. If you'd make the change, we'll be good to go. Thanks again!

JRAlexander (approved):

Issue #10143, opened 2019-01-25
Complete code sample link is broken.


RayMMond:
"Animation Timing Behavior Sample" link is broken.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e6cf772b-8d51-a207-cc23-6471e7881a41
* Version Independent ID: 2db3af0a-1f53-066a-1779-ad4e9346d62c
* Content: [How to: Accelerate or Decelerate an Animation](https://docs.microsoft.com/en-us/dotnet/framework/wpf/graphics-multimedia/how-to-accelerate-or-decelerate-an-animation)
* Content Source: [docs/framework/wpf/graphics-multimedia/how-to-accelerate-or-decelerate-an-animation.md](https://github.com/dotnet/docs/blob/master/docs/framework/wpf/graphics-multimedia/how-to-accelerate-or-decelerate-an-animation.md)
* Product: **dotnet-framework**
* GitHub Login: @dotnet-bot
* Microsoft Alias: **dotnetcontent**

PR #10144, opened 2019-01-25
Fix sample link

imba-tjd:
Fix https://github.com/dotnet/docs/issues/10143


mairaw (approved):
Thanks @imba-tjd. I just changed that fwlink to go to https://github.com/Microsoft/WPF-Samples/tree/master/Animation/AnimationTiming, which is where the original WPF sample was moved to and I'm not sure how they're different from each other. But your changes look good to me in this context!

PR #10167, opened 2019-01-26
Update introduction-to-classes.md

dogafincan:
Fix two small inconsistencies in the examples.


jzeferino (approved):

rpetrusha (approved):
Thanks, @dogafincan, for making these two changes. Our convention is to end comments with a period rather than a colon; I've made that change in your PR. I'll merge when the internal build completes.

PR #10179, opened 2019-01-28
Update nullable-reference-types.md

dogafincan:
Fix class name to SurveyResponse, and move a comma to the left.


rpetrusha (approved):
Thanks, @dogafincan, for making these corrections. I've made one addition to your PR -- I've updated the GitHub sample folder name, which I noticed was also incorrect. I'll merge when our internal build completes, and your updates should be live on docs.microsoft.com in the next day or two.

PR #10186, opened 2019-01-28
Allow elmish style tuples

forki:
Since Fsharplint is using this styleguide as source and ionide is using fsharplint, it seems this guide is going to be the canonical styleguide. In general I think this is a good thing and I try to be not too opionated here (which is hard).
Anyway, I'd like to suggest a small change to the very restrictive tuple instantiation. The new fsharplint is complaining on thousands of lines in our elmish update functions. Stuff that we as community suggested to use for everyone in safe stack. Basically all samples and all existing safe stack code is now considered as bad.
I'm not sure if my wording here is good, but I want to open up the discussion.

### DIFF
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -274,7 +274,7 @@ It is also commonly accepted to omit parentheses if the tuple is the return valu
 
 ```fsharp
 // OK
-let f model msg =
+let update model msg =
     match msg with
     | 1 -> model + 1, []
     | _ -> model, [ msg ]
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -270,7 +270,7 @@ match x, y with
 | x, y -> 1
 ```
 
-It is qlso accepted to omit parentheses if the tuple is the return value of a function:
+It is also commonly accepted to omit parentheses if the tuple is the return value of a function:
 
 ```fsharp
 // OK
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -279,7 +279,7 @@ let f model msg =
     | 1 -> model + 1, []
     | _ -> model, [ msg ]
 ```
-
+In summary, prefer parenthesized tuple instantiations, but when using tuples for pattern matching or a return value, it is considered fine to avoid parentheses.
 
 ## Formatting discriminated union declarations
 
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -270,6 +270,17 @@ match x, y with
 | x, y -> 1
 ```
 
+It is qlso accepted to omit parentheses if the tuple is the return value of a function:
+
+```fsharp
+// OK
+let f model msg =
+    match msg with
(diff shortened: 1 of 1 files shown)


smoothdeveloper:
A few times, I've been facing ambiguous issues around tuple instanciation and pattern matching, and those issues have been solved using explicit parens (https://github.com/Microsoft/visualfsharp/issues/5259 is example of me tripping on this), like it is the case in several areas of F#.

* Do I like to wrap all the expressions disregarding the fact there is no ambiguity: no
* Is it a better default to have tuple always wrapped: probably if we don't have better solution
* Is it better to have the warning show only for tuples that are ambiguous: yes

I feel code is more readable when parens are only added when necessary or when it clarifies possible ambiguity (despite not being necessary in all cases, the guideline to wrap would apply when some of the elements are not mere identifers or literals, but more complex expressions).

I'd like to relax the guidelines along those lines.

Ideally, in the tradition of Resharper in C#, I'd like the tooling to propose me to wrap / unwrap expressions which has extraneous parens or lacking parens, and have some place to define my prefered styling.

MangelMaxime:
I think the addition requested by @forki make sense when we consider Elmish applications.

forki:
There is another argument for lifting the restriction: symmetry. If we don't require it in destructering/pattern matching, then we should not do it in construction.

docs/fsharp/style-guide/formatting.md
```diff
@@ -270,6 +270,17 @@ match x, y with
 | x, y -> 1
 ```
 
+It is qlso accepted to omit parentheses if the tuple is the return value of a function:
```
cartermp:
```suggestion
It is also commonly accepted to omit parentheses if the tuple is the return value of a function:
```

docs/fsharp/style-guide/formatting.md
```diff
@@ -270,6 +270,17 @@ match x, y with
...
+
+```fsharp
+// OK
+let f model msg =
+    match msg with
+    | 1 -> model + 1, []
+    | _ -> model, [ msg ]
+```
+
```
cartermp:
```suggestion
In summary, prefer parenthesized tuple instantiations, but when using tuples for pattern matching or a return value, it is considered fine to avoid parentheses.
```

docs/fsharp/style-guide/formatting.md
```diff
@@ -270,6 +270,17 @@ match x, y with
 | x, y -> 1
 ```
 
+It is qlso accepted to omit parentheses if the tuple is the return value of a function:
+
+```fsharp
+// OK
+let f model msg =
```
cartermp:
```suggestion
let update model msg =
```

Just so that folks coming here from Fable know it immediately

cartermp (changes requested):
Some quick changes, but I otherwise think that this is good.

Can you also update line 258 along the following lines?

"A tuple instantiation should generally have the right-hand side parenthesized, and the [...]"

forki:
Applied your suggestions. Thx. (btw that new github feature is rad)

cartermp (approved):

cartermp:
@forki Yep, whoever built that out deserves a raise

cartermp:
Here's what your change looks like:
image

Merging, thanks @forki

PR #10232, opened 2019-01-30
Fix typos in example code

gyuwon:
`addOneIfOdd` function is misspelled as `addOneIffOdd`. So fix it.


cartermp (approved):
Thanks @gyuwon!

PR #10240, opened 2019-01-30
Redirect Managed Thread States to the API reference

### LINKED ISSUE #9896: Different views on the effect of the Thread.Start call on a thread state
The dotnet/dotnet-api-docs#1527 updated the [ThreadState enum](https://docs.microsoft.com/en-us/dotnet/api/system.threading.threadstate?view=netframework-4.7.2) reference such that the conceptual [Managed Thread States](https://docs.microsoft.com/en-us/dotnet/standard/threading/managed-thread-states) article now duplicates the API reference page and can be removed.

Almost duplicates, though.

There is one difference between the articles that should be resolved. This is the description of the behavior of the `Thread.Start` method.

The [conceptual docs](https://docs.microsoft.com/en-us/dotnet/standard/threading/managed-thread-states) say:
> The thread remains in the `Unstarted` state until it is moved into the started state by the operating system. Calling `Start` lets the operating system know that the thread can be started; it does not change the state of the thread.

And the table (at right: the *resulting* thread state of the action mentioned at left):
image

The [API reference](https://docs.microsoft.com/en-us/dotnet/api/system.threading.threadstate?view=netframework-4.7.2#remarks) describes the result of the `Thread.Start` call differently:
image

The [API doc on `Thread.Start`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.thread.start?view=netframework-4.7.2#System_Threading_Thread_Start) claims that `Thread.Start`
>Causes the operating system to change the state of the current instance to `Running`.

To my understanding, the conceptual article does not match the API docs. What docs describe the effect of the `Thread.Start` call correctly? If the API reference is correct, then this issue will be fixed with the removal of the conceptual docs article.

/cc: @rpetrusha

pkulikov:
As agreed in dotnet/dotnet-api-docs#1527
Fixes #9896


rpetrusha (approved):
This looks good, @pkulikov. Thanks. Unrelated to your PR, I've removed an unnecessary comma. I'll merge when the  build completes.

PR #10256, opened 2019-01-31
Add unloadability howto document

janvorli:
## Summary

This document describes how to use unloadability in .NET Core 3.0 and
how to debug issues with unloading.

It also contains a link to a complete sample in dotnet/samples repo.

### DIFF
--- a/docs/standard/assembly/unloadability-howto.md
+++ b/docs/standard/assembly/unloadability-howto.md
@@ -234,7 +234,7 @@ As you can see, the last thread has `test.Program.ThreadProc()`. This is a funct
 
 ## Example source with unloadability issues
 
-This example is used in the debugging above.
+The following code is used in the previous debugging example.
 
 ### Main testing program
 
--- a/docs/standard/assembly/unloadability-howto.md
+++ b/docs/standard/assembly/unloadability-howto.md
@@ -181,7 +181,7 @@ Another case that can prevent unloading of an `AssemblyLoadContext` is when a th
 ```
 
 The command means "apply to all threads the `!clrstack` command". Here is the output of that command for our example. Unfortunately, LLDB on Unix doesn't have any way to apply a command to all threads, so you'll need to resort to manually switching threads and repeating the `clrstack` command.
-You should ignore all threads where the debugger says "Unable to walk the managed stack."
+Ignore all threads where the debugger says "Unable to walk the managed stack."
 
 ```console
 OS Thread Id: 0x6ba8 (0)
--- a/docs/standard/assembly/unloadability-howto.md
+++ b/docs/standard/assembly/unloadability-howto.md
@@ -242,6 +242,6 @@ This example is used in the debugging above.
 
 ## Program loaded into the TestAssemblyLoadContext
 
-This is the `test.dll` passed to the `ExecuteAndUnload` method in the main testing program.
+The following code represents the `test.dll` passed to the `ExecuteAndUnload` method in the main testing program.
 
 [!code-csharp[Program loaded into the TestAssemblyLoadContext](~/samples/snippets/standard/assembly/unloading/unloadability_issues_example_test.cs)]
--- a/docs/standard/assembly/unloadability-howto.md
+++ b/docs/standard/assembly/unloadability-howto.md
@@ -1,0 +1,247 @@
+---
+title: "How to use and debug assembly unloadability in .NET Core"
+description: "Learn how to use collectible AssemblyLoadContext for loading and unloading managed assemblies and how to debug issues preventing the unloading success."
+author: "janvorli"
+ms.author: "janvorli"
+ms.date: "02/05/2019"
+---
+# How to use and debug assembly unloadability in .NET Core
+
+Starting with .NET Core 3.0, the ability to load and later unload a set of assemblies is supported. In .NET Framework, custom app domains were used for this purpose, but .NET Core only supports a single default app domain.
+
+.NET Core 3.0 and later versions support unloadability through <xref:System.Runtime.Loader.AssemblyLoadContext>. You can load a set of assemblies into a collectible `AssemblyLoadContext`, execute methods in them or just inspect them using reflection, and finally unload the `AssemblyLoadContext`. That unloads the assemblies loaded into that `AssemblyLoadContext`.
+
+There's one noteworthy difference between the unloading using `AssemblyLoadContext` and using AppDomains. With AppDomains, the unloading is forced. At the unload time, all threads running in the target AppDomain are aborted, managed COM objects created in the target AppDomain are destroyed, etc. With `AssemblyLoadContext`, the unload is "cooperative". Calling the <xref:System.Runtime.Loader.AssemblyLoadContext.Unload%2A?displayProperty=nameWithType> method just initiates the unloading. The unloading finishes after:
+
+- No threads have methods from the assemblies loaded into the `AssemblyLoadContext` on their call stacks.
+- None of the types from the assemblies loaded into the `AssemblyLoadContext`, instances of those types and the assemblies themselves outside of the `AssemblyLoadContext` are referenced by:
+  - References outside of the `AssemblyLoadContext`, except of weak references (<xref:System.WeakReference> or <xref:System.WeakReference%601>).
+  - Strong GC handles (<xref:System.Runtime.InteropServices.GCHandleType>.<xref:System.Runtime.InteropServices.GCHandleType.Normal> or <xref:System.Runtime.InteropServices.GCHandleType>.<xref:System.Runtime.InteropServices.GCHandleType.Pinned>) from both inside and outside of the `AssemblyLoadContext`.
+
+## Using collectible AssemblyLoadContext
+
+This section contains a detailed step-by-step tutorial that shows a simple way to load a .NET Core application into a collectible `AssemblyLoadContext`, execute its entry point, and then unload it. You can find a complete sample at <https://github.com/dotnet/samples/tree/master/core/tutorials/Unloading>.
+
+### Create a collectible AssemblyLoadContext
+
+You need to derive your class from the <xref:System.Runtime.Loader.AssemblyLoadContext> and overload its <xref:System.Runtime.Loader.AssemblyLoadContext.Load%2A?displayProperty=nameWithType> method. That method resolves references to all assemblies that are dependencies of assemblies loaded into that `AssemblyLoadContext`.
+The following code is an example of the simplest custom `AssemblyLoadContext`:
+
+[!code-csharp[Simple custom AssemblyLoadContext](~/samples/snippets/standard/assembly/unloading/simple_example.cs#1)]
+
+As you can see, the `Load` method returns `null`. That means that all the dependency assemblies are loaded into the default context, and the new context contains only the assemblies explicitly loaded into it.
+
+If you want to load some or all of the dependencies into the `AssemblyLoadContext` too, you can use the `AssemblyDependencyResolver` in the `Load` method. The `AssemblyDependencyResolver` resolves the assembly names to absolute assembly file paths using the `*.deps.json` file contained in the directory of the main assembly loaded into the context and using assembly files in that directory.
+
+[!code-csharp[Advanced custom AssemblyLoadContext](~/samples/snippets/standard/assembly/unloading/complex_assemblyloadcontext.cs)]
+
+### Use a custom collectible AssemblyLoadContext
+
+This section assumes the simpler version of the `TestAssemblyLoadContext` is being used.
+
+You can create an instance of the custom `AssemblyLoadContext` and load an assembly into it as follows:
+
+[!code-csharp[Part 1](~/samples/snippets/standard/assembly/unloading/simple_example.cs#3)]
+
+For each of the assemblies referenced by the loaded assembly, the `TestAssemblyLoadContext.Load` method is called so that the `TestAssemblyLoadContext` can decide where to get the assembly from. In our case, it returns `null` to indicate that it should be loaded into the default context from locations that the runtime uses to load assemblies by default.
+
+Now that an assembly was loaded, you can execute a method from it. Run the `Main` method:
+
+[!code-csharp[Part 2](~/samples/snippets/standard/assembly/unloading/simple_example.cs#4)]
+
+After the `Main` method returns, you can initiate unloading by either calling the `Unload` method on the custom `AssemblyLoadContext` or getting rid of the reference you have to the `AssemblyLoadContext`:
+
+[!code-csharp[Part 3](~/samples/snippets/standard/assembly/unloading/simple_example.cs#5)]
+
+This is sufficient to unload the test assembly. Let's actually put all of this into a separate non-inlineable method to ensure that the `TestAssemblyLoadContext`, `Assembly`, and `MethodInfo` (the `Assembly.EntryPoint`) can't be kept alive by stack slot references (real- or JIT-introduced locals). That could keep the `TestAssemblyLoadContext` alive and prevent the unload.
+
+Also, return a weak reference to the `AssemblyLoadContext` so that you can use it later to detect unload completion.
+
+[!code-csharp[Part 4](~/samples/snippets/standard/assembly/unloading/simple_example.cs#2)]
+
+Now you can run this function to load, execute, and unload the assembly.
+
+[!code-csharp[Part 5](~/samples/snippets/standard/assembly/unloading/simple_example.cs#6)]
+
+However, the unload doesn't complete immediately. As previously mentioned, it relies on the GC to collect all the objects from the test assembly. In many cases, it isn't necessary to wait for the unload completion. However, there are cases where it's useful to know that the unload has finished. For example, you may want to delete the assembly file that was loaded into the custom `AssemblyLoadContext` from disk. In such a case, the following code snippet can be used. It triggers a GC and waits for pending finalizers in a loop until the weak reference to the custom `AssemblyLoadContext` is set to `null`, indicating the target object was collected. Note that in most cases, just one pass through the loop is required. However, for more complex cases where objects created by the code running in the `AssemblyLoadContext` have finalizers, more passes may be needed.
+
+[!code-csharp[Part 6](~/samples/snippets/standard/assembly/unloading/simple_example.cs#7)]
+
+### The Unloading event
+
+In some cases, it may be necessary for the code loaded into a custom `AssemblyLoadContext` to perform some cleanup when the unloading is initiated. For example, it may need to stop threads, clean up some strong GC handles, etc. The `Unloading` event can be used in such cases. A handler that performs the necessary cleanup can be hooked to this event.
+
+### Troubleshoot unloadability issues
+
+Due to the cooperative nature of the unloading, it's easy to forget about references keeping the stuff in a collectible `AssemblyLoadContext` alive and preventing unload. Here is a summary of things (some of them non-obvious) that can hold the references:
+
+- Regular references held from outside of the collectible `AssemblyLoadContext`, stored in a stack slot or a processor register (method locals, either explicitly created by the user code or implicitly by the JIT), a static variable or a strong / pinning GC handle, and transitively pointing to:
+  - An assembly loaded into the collectible `AssemblyLoadContext`.
+  - A type from such an assembly.
+  - An instance of a type from such an assembly.
+- Threads running code from an assembly loaded into the collectible `AssemblyLoadContext`.
+- Instances of custom non-collectible `AssemblyLoadContext` types created inside of the collectible `AssemblyLoadContext`
+- Pending <xref:System.Threading.RegisteredWaitHandle> instances with callbacks set to methods in the custom `AssemblyLoadContext`
+
+Hints to find stack slot / processor register rooting an object:
+
+- Passing function call results directly to another function may create a root even though there is no user-created local variable.
+- If a reference to an object was available at any point in a method, the JIT might have decided to keep the reference in a stack slot / processor register for as long as it wants in the current function.
+
+## Debug unloading issues
+
+Debugging issues with unloading can be tedious. You can get into situations where you don't know what can be holding an `AssemblyLoadContext` alive, but the unload fails.
+The best weapon to help with that is WinDbg (LLDB on Unix) with the SOS plugin. You need to find what's keeping a `LoaderAllocator` belonging to the specific `AssemblyLoadContext` alive.
+This plugin allows you to look at GC heap objects, their hierarchies, and roots.
+To load the plugin into the debugger, enter the following command in the debugger command line:
+
+In WinDbg (it seems WinDbg does that automatically when breaking into .NET Core application):
+
+```console
+.loadby sos coreclr
+```
+
+In LLDB:
+
+```console
+plugin load /path/to/libsosplugin.so
+```
+
+Let's try to debug an example program that has problems with unloading. Source code is included below. When you run it under WinDbg, the program breaks into the debugger right after attempting to check for the unload success. You can then start looking for the culprits.
+
+Note that if you debug using LLDB on Unix, the SOS commands in the following examples don't have the `!` in front of them.
+
+```console
+!dumpheap -type LoaderAllocator
+```
+
+This command dumps all objects with a type name containing `LoaderAllocator` that are in the GC heap. Here is an example:
+
+```console
+         Address               MT     Size
+000002b78000ce40 00007ffadc93a288       48     
+000002b78000ceb0 00007ffadc93a218       24     
+
+Statistics:
+              MT    Count    TotalSize Class Name
+00007ffadc93a218        1           24 System.Reflection.LoaderAllocatorScout
+00007ffadc93a288        1           48 System.Reflection.LoaderAllocator
+Total 2 objects
+```
+
+In the "Statistics:" part below, check the `MT` (`MethodTable`) belonging to the `System.Reflection.LoaderAllocator`, which is the object we care about. Then in the list at the beginning, find the entry with `MT` matching that one and get the address of the object itself. In our case, it is "000002b78000ce40"
+
+Now that we know the address of the `LoaderAllocator` object, we can use another command to find its GC roots
+
+```console
+!gcroot -all 0x000002b78000ce40 
+```
+
+This command dumps the chain of object references that lead to the `LoaderAllocator` instance. The list starts with the root, which is the entity that keeps our `LoaderAllocator` alive and thus is the core of the problem you're debugging. The root can be a stack slot, a processor register, a GC handle, or a static variable.
+
+Here is an example of the output of the `gcroot` command:
+
+```console
+Thread 4ac:
+    000000cf9499dd20 00007ffa7d0236bc example.Program.Main(System.String[]) [E:\unloadability\example\Program.cs @ 70]
+        rbp-20: 000000cf9499dd90
+            ->  000002b78000d328 System.Reflection.RuntimeMethodInfo
+            ->  000002b78000d1f8 System.RuntimeType+RuntimeTypeCache
+            ->  000002b78000d1d0 System.RuntimeType
+            ->  000002b78000ce40 System.Reflection.LoaderAllocator
+
+HandleTable:
+    000002b7f8a81198 (strong handle)
+    -> 000002b78000d948 test.Test
+    -> 000002b78000ce40 System.Reflection.LoaderAllocator
+
+    000002b7f8a815f8 (pinned handle)
+    -> 000002b790001038 System.Object[]
+    -> 000002b78000d390 example.TestInfo
+    -> 000002b78000d328 System.Reflection.RuntimeMethodInfo
+    -> 000002b78000d1f8 System.RuntimeType+RuntimeTypeCache
+    -> 000002b78000d1d0 System.RuntimeType
+    -> 000002b78000ce40 System.Reflection.LoaderAllocator
+
+Found 3 roots.
+```
+
+Now you need to figure out where the root is located so you can fix it. The easiest case is when the root is a stack slot or a processor register. In that case, the `gcroot` shows you the name of the function whose frame contains the root and the thread executing that function. The difficult case is when the root is a static variable or a GC handle. 
+
+In the previous example, the first root is a local of type `System.Reflection.RuntimeMethodInfo` stored in the frame of the function `example.Program.Main(System.String[])` at address `rbp-20` (`rbp` is the processor register `rbp` and -20 is a hexadecimal offset from that register).
+
+The second root is a normal (strong) `GCHandle` that holds a reference to an instance of the `test.Test` class. 
+
+The third root is a pinned `GCHandle`. This one is actually a static variable. Unfortunately, there is no way to tell. Statics for reference types are stored in a managed object array in internal runtime structures.
+
+Another case that can prevent unloading of an `AssemblyLoadContext` is when a thread has a frame of a method from an assembly loaded into the `AssemblyLoadContext` on its stack. You can check that by dumping managed call stacks of all threads:
+
+```console
+~*e !clrstack
+```
+
+The command means "apply to all threads the `!clrstack` command". Here is the output of that command for our example. Unfortunately, LLDB on Unix doesn't have any way to apply a command to all threads, so you'll need to resort to manually switching threads and repeating the `clrstack` command.
+You should ignore all threads where the debugger says "Unable to walk the managed stack."
+
+```console
+OS Thread Id: 0x6ba8 (0)
+        Child SP               IP Call Site
+0000001fc697d5c8 00007ffb50d9de12 [HelperMethodFrame: 0000001fc697d5c8] System.Diagnostics.Debugger.BreakInternal()
+0000001fc697d6d0 00007ffa864765fa System.Diagnostics.Debugger.Break()
+0000001fc697d700 00007ffa864736bc example.Program.Main(System.String[]) [E:\unloadability\example\Program.cs @ 70]
+0000001fc697d998 00007ffae5fdc1e3 [GCFrame: 0000001fc697d998] 
+0000001fc697df28 00007ffae5fdc1e3 [GCFrame: 0000001fc697df28] 
+OS Thread Id: 0x2ae4 (1)
+Unable to walk the managed stack. The current thread is likely not a 
+managed thread. You can run !threads to get a list of managed threads in
+the process
+Failed to start stack walk: 80070057
+OS Thread Id: 0x61a4 (2)
+Unable to walk the managed stack. The current thread is likely not a 
+managed thread. You can run !threads to get a list of managed threads in
+the process
+Failed to start stack walk: 80070057
+OS Thread Id: 0x7fdc (3)
+Unable to walk the managed stack. The current thread is likely not a 
+managed thread. You can run !threads to get a list of managed threads in
+the process
+Failed to start stack walk: 80070057
+OS Thread Id: 0x5390 (4)
+Unable to walk the managed stack. The current thread is likely not a 
+managed thread. You can run !threads to get a list of managed threads in
+the process
+Failed to start stack walk: 80070057
+OS Thread Id: 0x5ec8 (5)
+        Child SP               IP Call Site
+0000001fc70ff6e0 00007ffb5437f6e4 [DebuggerU2MCatchHandlerFrame: 0000001fc70ff6e0] 
+OS Thread Id: 0x4624 (6)
+        Child SP               IP Call Site
+GetFrameContext failed: 1
+0000000000000000 0000000000000000 
+OS Thread Id: 0x60bc (7)
+        Child SP               IP Call Site
+0000001fc727f158 00007ffb5437fce4 [HelperMethodFrame: 0000001fc727f158] System.Threading.Thread.SleepInternal(Int32)
+0000001fc727f260 00007ffb37ea7c2b System.Threading.Thread.Sleep(Int32)
+0000001fc727f290 00007ffa865005b3 test.Program.ThreadProc() [E:\unloadability\test\Program.cs @ 17]
+0000001fc727f2c0 00007ffb37ea6a5b System.Threading.Thread.ThreadMain_ThreadStart()
+0000001fc727f2f0 00007ffadbc4cbe3 System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
+0000001fc727f568 00007ffae5fdc1e3 [GCFrame: 0000001fc727f568] 
+0000001fc727f7f0 00007ffae5fdc1e3 [DebuggerU2MCatchHandlerFrame: 0000001fc727f7f0] 
+
+```
+
+As you can see, the last thread has `test.Program.ThreadProc()`. This is a function from the assembly loaded into the `AssemblyLoadContext`, and so it keeps the `AssemblyLoadContext` alive.
+
+## Example source with unloadability issues
+
+This example is used in the debugging above.
+
+### Main testing program
+
+[!code-csharp[Main testing program](~/samples/snippets/standard/assembly/unloading/unloadability_issues_example_main.cs)]
+
+## Program loaded into the TestAssemblyLoadContext
+
+This is the `test.dll` passed to the `ExecuteAndUnload` method in the main testing program.
+
+[!code-csharp[Program loaded into the TestAssemblyLoadContext](~/samples/snippets/standard/assembly/unloading/unloadability_issues_example_test.cs)]
--- a/docs/standard/assembly/unloadability-howto.md
+++ b/docs/standard/assembly/unloadability-howto.md
@@ -180,7 +180,7 @@ Another case that can prevent unloading of an `AssemblyLoadContext` is when a th
 ~*e !clrstack
 ```
 
-The command means "apply to all threads the `!clrstack` command". Here is the output of that command for our example. Unfortunately, LLDB on Unix doesn't have any way to apply a command to all threads, so you'll need to resort to manually switching threads and repeating the `clrstack` command.
+The command means "apply to all threads the `!clrstack` command". The following is the output of that command for the example. Unfortunately, LLDB on Unix doesn't have any way to apply a command to all threads, so you'll need to resort to manually switching threads and repeating the `clrstack` command.
 Ignore all threads where the debugger says "Unable to walk the managed stack."
 
 ```console
--- a/docs/standard/assembly/index.md
+++ b/docs/standard/assembly/index.md
@@ -48,11 +48,12 @@ Assemblies form the fundamental unit of deployment, version control, reuse, acti
 
 ## See also
 
- - [.NET assembly file format](file-format.md)  
+ - [.NET assembly file format](file-format.md)
  - [Assemblies in the Common Language Runtime](../../framework/app-domains/assemblies-in-the-common-language-runtime.md)  
  - [Friend Assemblies (C#)](../../csharp/programming-guide/concepts/assemblies-gac/friend-assemblies.md)  
  - [Friend Assemblies (Visual Basic)](../../visual-basic/programming-guide/concepts/assemblies-gac/friend-assemblies.md)  
  - [How to: Load and Unload Assemblies (C#)](../../csharp/programming-guide/concepts/assemblies-gac/how-to-load-and-unload-assemblies.md)  
  - [How to: Load and Unload Assemblies (Visual Basic)](../../visual-basic/programming-guide/concepts/assemblies-gac/how-to-load-and-unload-assemblies.md)  
+ - [How to: Use and Debug Assembly Unloadability in .NET Core](unloadability-howto.md)
  - [How to: Determine If a File Is an Assembly (C#)](../../csharp/programming-guide/concepts/assemblies-gac/how-to-determine-if-a-file-is-an-assembly.md)  
  - [How to: Determine If a File Is an Assembly (Visual Basic)](../../visual-basic/programming-guide/concepts/assemblies-gac/how-to-determine-if-a-file-is-an-assembly.md)  
--- a/docs/toc.md
+++ b/docs/toc.md
@@ -33,6 +33,7 @@
 ## [Handling and throwing exceptions](standard/exceptions/)
 ## [Assemblies in .NET](standard/assembly/index.md)
 ### [.NET Assembly File Format](standard/assembly/file-format.md)
+### [How to: Use and Debug Assembly Unloadability in .NET Core](standard/assembly/unloadability-howto.md)
 ## [Garbage Collection](standard/garbage-collection/)
 ## [Generic types](standard/generics.md)
 ## [Delegates and lambdas](standard/delegates-lambdas.md)


janvorli:
cc: @jkotas, @jkoritzinsky, @AaronRobinsonMSFT

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+
+    protected override Assembly Load(AssemblyName name)
+    {
+        return null;
+    }
+}
+```
+As you can see, the `Load` method returns `null`. That means that all the dependency assemblies are loaded into the default context and only the assemblies explicitly loaded into the new context are in this context. 
+
+In case you want to load some or all of the dependencies into the `AssemblyLoadContext` too, you can use the `AssemblyDependencyResolver` in the `Load` method. The `AssemblyDependencyResolver` resolves the assembly names to absolute assembly file paths using the *.deps.json file contained in the directory of the main assembly loaded into the context and using assembly files in that directory.
```
jkotas:
```suggestion
In case you want to load some or all of the dependencies into the `AssemblyLoadContext` too, you can use the `AssemblyDependencyResolver` in the `Load` method. The `AssemblyDependencyResolver` resolves the assembly names to absolute assembly file paths using the `*.deps.json` file contained in the directory of the main assembly loaded into the context and using assembly files in that directory.
```

AaronRobinsonMSFT: @jkoritzinsky added a plugin sample for this. Reference it?

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+
+    return result;
+}
+```
+Now we can run this function to Load, execute and Unload the assembly.
+```C#
+WeakReference testAlcWeakRef;
+int result = ExecuteAndUnload("absolute/path/to/your/assembly", out testAlcWeakRef);
+```
+However, the unloading doesn't complete immediately. As I've already mentioned, it relies on GC to collect all the objects from the test assembly (etc.). In many cases, it is not necessary to wait for the unload completion. However there are cases where it is useful to know that the unload has finished. For example, you may want to delete the assembly file that was loaded into the custom context from disk. In such case, the following code snippet can be used. It triggers a GC and waits for pending finalizers in a loop until the weak reference to the custom AssemblyLoadContext is set to null, indicating the target object was collected. Please note that in most cases, just one pass through the loop is required. However for more complex cases where objects created by the code running in the AssemblyLoadContext have finalizers, more passes may be needed.
```
AaronRobinsonMSFT: Please remove all uses of "I". In this case, ~~As I've already mentioned,~~ -> "As previously mentioned,"

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+```C#
+for (int i = 0; testAlcWeakRef.IsAlive && (i < 10); i++)
+{
+    GC.Collect();
+    GC.WaitForPendingFinalizers();
+}
+```
+### The Unloading event
+In some cases, it may be necessary for the code loaded into a custom AssemblyLoadContext to perform some cleanup when the unloading is initiated. For example, it may need to stop threads, cleanup some strong GC handles, etc. The `Unloading` event can be used in such cases. A handler that performs the necessary cleanup can be hooked to this event.
```
AaronRobinsonMSFT: Unloading should be quoted not back-ticked.

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+### Troubleshooting unloadability issues
+While the previous paragraphs may sound like it is all easy and simple, it is often not the case. Due to the cooperative nature of the unloading, it is easy to forget about references keeping the stuff in the custom AssemblyLoadContext alive and preventing unload. Here is a summary of things (some of them non-obvious) that can hold the references:
+* Regular reference held from outside of the custom AssemblyLoadContext  or a strong / pinning GC handle. Such reference may point to:
+  * An assembly loaded into the custom `AssemblyLoadContext`
+  * Type from such an assembly
+  * Instance of a type  from such an assembly
+* Threads running code from an assembly loaded into the custom AssemblyLoadContext
+* Instances of non-collectible AssemblyLoadContext
```
AaronRobinsonMSFT: We are inconsistent on back-ticks around `AssemblyLoadContext` and types in general.

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+* Pending RegisteredWaitHandle instances
+The above mentioned reference can come from any object with root in a stack slot or processor register (method locals, either explicitly created by the user code or implicitly by the JIT), a static variable or strong / pinning GC handle. TODO: explain what root means
```
AaronRobinsonMSFT: >TODO: explain what roo means

Is there a nice definition in the BOTR we can reference? If not, I would add one there.

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+This plugin allows us to look at GC heap objects, their hierarchies and roots. 
+To load the plugin into the debugger, enter the following command in the debugger command line:
+```
+In WinDbg (it seems WinDbg does that automatically when breaking into .NET Core application):
+.loadby sos coreclr
+
+In LLDB:
+plugin load /path/to/libsosplugin.so
+```
+Let's try to debug an example program that has problems with unloading. I have included its source code below. When you run it under WinDbg, the program breaks into the debugger right after attempting to check for the unload success. We can then start looking for the culprits.
```
AaronRobinsonMSFT: ~~I have included its source code below~~ -> "Source code is included below."

docs/core/tutorials/unloadability-howto.md
janvorli: Do you mean a sample that uses *.deps.json in some interesting way or a just a sample that uses AssemblyDependencyResolver? If it is the latter, I already have a link above to my own unloadability sample that uses it.

janvorli: My sample is also in that official sample repo. I've merged it in two days ago. It covers unloadability and since I wanted to make it a boiler plate that people can use, I've used the AssemblyDependencyResolver there too. If you've missed the link in the text above, it is here:
https://github.com/dotnet/samples/tree/master/core/tutorials/Unloading

janvorli:
@AaronRobinsonMSFT I've updated the doc based on your feedback. The only place where I've left the type names without backticks is in the section titles, as it looks awkward in those.

docs/core/tutorials/unloadability-howto.md
jkoritzinsky: In that case you should use a code reference tag to automatically import the code from the sample like the following

`[!code-csharp[Assembly-load-context-with-assembly-dependency-resolver](~/samples/blob/master/core/tutorials/Unloading/Host/Program.cs)]`

You might also want to seperate out the ALC into a different file to make the code reference work a little cleaner.

janvorli: I didn't want to include the code from my sample here. I see this doc and the sample as separate entities. This doc focuses on a simple version where the Load method always returns NULL and just mentions the more advanced way briefly so that people know it exist. 
The sample is a separate full blown end to end example of using unloadable ALC that complements this article. I didn't actually know you were working on your sample and frankly even didn't know about the AssemblyDependencyResolver existence when I've completed the first version of my sample and I was originally using just the simple ALC returning NULL from the Load overried. @jkotas then pointed out during review of that sample that it might be nice to make it more complete by adding the usage of the AssemblyDependencyResolver. Thus I've added it there and mentioned it in this doc.

jkoritzinsky: That tag is a specially handled by the markdown engine that docs.microsoft.com uses. Github doesn't understand it fully.

janvorli:
Oh, I've forgotten about the TODO about explaining gc root. I am going to add that. I have not found it in any doc.

janvorli:
I've actually reworded the text that was mentioning the roots before and got rid of the "roots" name there as it doesn't seem to be necessary for the explanation.

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+    -> 000002b790001038 System.Object[]
+    -> 000002b78000d390 example.TestInfo
+    -> 000002b78000d328 System.Reflection.RuntimeMethodInfo
+    -> 000002b78000d1f8 System.RuntimeType+RuntimeTypeCache
+    -> 000002b78000d1d0 System.RuntimeType
+    -> 000002b78000ce40 System.Reflection.LoaderAllocator
+
+Found 3 roots.
+```
+Now you need to figure out where is the root located so that you can fix it. The easiest case is when the root is a stack slot or a processor register. In that case, the `gcroot` shows you the name of the function whose frame contains the root and the thread executing that function. The difficult case is when the root is a static variable or a GC handle.
```
rpetrusha:
```suggestion
Now you need to figure out where the root is located so that you can fix it. The easiest case is when the root is a stack slot or a processor register. In that case, the `gcroot` shows you the name of the function whose frame contains the root and the thread executing that function. The difficult case is when the root is a static variable or a GC handle. 
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+    -> 000002b78000d1d0 System.RuntimeType
+    -> 000002b78000ce40 System.Reflection.LoaderAllocator
+
+Found 3 roots.
+```
+Now you need to figure out where is the root located so that you can fix it. The easiest case is when the root is a stack slot or a processor register. In that case, the `gcroot` shows you the name of the function whose frame contains the root and the thread executing that function. The difficult case is when the root is a static variable or a GC handle. 
+
+In our example above, the first root is a local of type `System.Reflection.RuntimeMethodInfo` stored in the frame of the function `example.Program.Main(System.String[])` at address `rbp-20` (`rbp` is the processor register `rbp` and -20 is a hexadecimal offset from that register).
+
+The second root is a normal (strong) `GCHandle` that holds reference to an instance of the `test.Test` class.
```
rpetrusha:
```suggestion
The second root is a normal (strong) `GCHandle` that holds a reference to an instance of the `test.Test` class. 
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+
+In our example above, the first root is a local of type `System.Reflection.RuntimeMethodInfo` stored in the frame of the function `example.Program.Main(System.String[])` at address `rbp-20` (`rbp` is the processor register `rbp` and -20 is a hexadecimal offset from that register).
+
+The second root is a normal (strong) `GCHandle` that holds reference to an instance of the `test.Test` class. 
+
+The third root is a pinned `GCHandle`. This one is actually a static variable. Unfortunately, there is no way to tell. Statics for reference types are stored in a managed object array in internal runtime structures.
+
+Another case that can prevent unloading of an `AssemblyLoadContext` is when a thread has a frame of a method from assembly loaded into the `AssemblyLoadContext` on its stack. You can check that by dumping managed call stacks of all threads:
```
rpetrusha:
```suggestion
Another case that can prevent unloading of an `AssemblyLoadContext` is when a thread has a frame of a method from an assembly loaded into the `AssemblyLoadContext` on its stack. You can check that by dumping managed call stacks of all threads:
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+```
```
rpetrusha:
```suggestion
```console
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+~*e !clrstack
+```
+The command means "apply to all threads the `!clrstack` command". Here is an example of an output of that command for our example. Unfortunately, LLDB on Unix doesn't have any way to apply a command to all threads, so you'll need to resort to manual switching threads and repeating the `clrstack` command.
```
rpetrusha:
```suggestion
The command means "apply to all threads the `!clrstack` command". Here is the output of that command for our example. Unfortunately, LLDB on Unix doesn't have any way to apply a command to all threads, so you'll need to resort to manually switching threads and repeating the `clrstack` command.
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+You should ignore all threads where the debugger says "Unable to walk the managed stack."
+```
```
rpetrusha:
```suggestion
```console
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+0000001fc727f158 00007ffb5437fce4 [HelperMethodFrame: 0000001fc727f158] System.Threading.Thread.SleepInternal(Int32)
+0000001fc727f260 00007ffb37ea7c2b System.Threading.Thread.Sleep(Int32)
+0000001fc727f290 00007ffa865005b3 test.Program.ThreadProc() [E:\unloadability\test\Program.cs @ 17]
+0000001fc727f2c0 00007ffb37ea6a5b System.Threading.Thread.ThreadMain_ThreadStart()
+0000001fc727f2f0 00007ffadbc4cbe3 System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
+0000001fc727f568 00007ffae5fdc1e3 [GCFrame: 0000001fc727f568] 
+0000001fc727f7f0 00007ffae5fdc1e3 [DebuggerU2MCatchHandlerFrame: 0000001fc727f7f0] 
+
+```
+As you can see the last thread has `test.Program.ThreadProc()`. This is a function from the assembly loaded into the `AssemblyLoadContext` and so it keeps the `AssemblyLoadContext` alive.
```
rpetrusha:
```suggestion
As you can see, the last thread has `test.Program.ThreadProc()`. This is a function from the assembly loaded into the `AssemblyLoadContext`, and so it keeps the `AssemblyLoadContext` alive.
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+## Example source with unloadability issues
+This example is used in the debugging above.
+### Main testing program
+```C#
```
rpetrusha:
```suggestion
```csharp
```
But here again, it's best to put this in the samples repo.

rpetrusha (changes requested):
This looks good, @janvorli. I've left a number of questions, comments, and suggestions for you to consider. You can bulk-review my suggested changes by reviewing from the **Files changes** tab.

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+
+The .NET Core 3.0 brings support of unloadability using `AssemblyLoadContext`. You can now load a set of assemblies into an `AssemblyLoadContext`, execute methods in them or just inspect them using reflection and finally unload the `AssemblyLoadContext`. That results in unloading of the assemblies loaded into that `AssemblyLoadContext`.
+
+There is one noteworthy difference between the unloading using `AssemblyLoadContext` and using AppDomains. With AppDomains, the unloading is forced. At the unload time, all threads running in the target AppDomain are aborted, managed COM objects created in the AppDomain that's being unloaded are destroyed, etc. 
+
+With the new way that uses `AssemblyLoadContext`, the unload is "cooperative". Calling the `Unload` method on the `AssemblyLoadContext` just initiates the unloading. But the unload will not complete until there are no threads having code from the assemblies loaded into the `AssemblyLoadContext` on their call stacks and until there are no strong references to types from the assemblies loaded into the `AssemblyLoadContext`, their instances and the assemblies themselves. And that means that GC needs to collect those first.
```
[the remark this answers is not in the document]
janvorli: "their instances" means instances of the types from the assemblies  ... I can see it is a bit unclear, I'll reword it.

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,359 @@
...
+## Using unloadable AssemblyLoadContext
```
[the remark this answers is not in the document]
janvorli: Collectible and unloadable AssemblyLoadContext means the same thing. So it is an AssemblyLoadContext created with the isCollectible argument of the constructor set to true. While the underlying features that it uses are collectible Assemblies (which means they can be collected by the GC after the AssemblyLoadContext is unloaded / collected), the term "unloadability" is what people use in connection to the AssemblyLoadContext as in the desktop .NET framework, the equivalent functionality was called AppDomain unloading (it doesn't use collectibility).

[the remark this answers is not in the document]
janvorli: I am not sure I understand the suggested change, it seems to change the meaning of it. It means "How to use unloadable AssemblyLoadContext", but in your wording, it sounds like an imperative.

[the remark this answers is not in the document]
janvorli: After giving it more thought, I've changed all references to "unloadable AssemblyLoadContext" to "collectible AssemblyLoadContext". It seems it is really better that way.

janvorli:
@rpetrusha thank you so much for the detailed review and all the suggested corrections of the grammar. I can see I should get better especially in the comma usage.
I've applied all of your suggestions except for changing one of the titles and also changed "unloadable AssemblyLoadContext" at few places to "collectible AssemblyLoadContext".
I have also created a PR containing all the code snippets from this document here: https://github.com/dotnet/samples/pull/617. I will update this doc once that PR goes in to reference the snippets instead of having them inline.

AaronRobinsonMSFT:
LGTM.

mairaw:
There's a broken link in the build report and the article needs to be added to the TOC file as well (https://github.com/dotnet/docs/blob/master/docs/toc.md)

rpetrusha:
The broken link in the build report is to a .NET Core 3.0 API, @mairaw. I thought that it would be better to leave the xref than fence it and have to change it later.

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+        }
+
+        return null;
+    }
+}
+```
+### Use a custom collectible AssemblyLoadContext
+This section assumes the simpler version of the `TestAssemblyLoadContext` is being used.
+
+You can create an instance of the custom `AssemblyLoadContext` and load an assembly into it as follows:
```
mairaw:
```suggestion
You can create an instance of the custom `AssemblyLoadContext` and load an assembly into it as follows:

```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+```csharp
+var alc = new TestAssemblyLoadContext();
+Assembly a = alc.LoadFromAssemblyPath("absolute/path/to/your/assembly");
+```
```
mairaw:
```suggestion
```

```

File: docs/core/tutorials/unloadability-howto.md [hunk 19]
```diff
@@ -0,0 +1,362 @@
...
+For each of the assemblies referenced by the loaded assembly, the `TestAssemblyLoadContext.Load` method is called so that the `TestAssemblyLoadContext` can decide where to get the assembly from. In our case, it returns `null` to indicate that it should be loaded into the default context from locations that the runtime uses to load assemblies by default.
+
+Now that we have loaded an assembly, we can execute a method from it. Let's run the `Main` method:
```
mairaw: We don't use we in our content, unless it's a Microsoft recommendation. Suggestion:

```suggestion
Now that an assembly was loaded, you can execute a method from it. Run the `Main` method:
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+```csharp
+var args = new object[1] {new string[] {"Hello"}};
+int result = (int)a.EntryPoint.Invoke(null, args);
+```
+After the `Main` method returns, we can initiate unloading by either calling the `Unload` method on the custom `AssemblyLoadContext` or getting rid of the reference we have to the `AssemblyLoadContext`:
```
mairaw:
```suggestion

After the `Main` method returns, you can initiate unloading by either calling the `Unload` method on the custom `AssemblyLoadContext` or getting rid of the reference you have to the `AssemblyLoadContext`:
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+```csharp
+alc.Unload();
+```
+This is sufficient to unload the test assembly. Let's actually put all of this into a separate non-inlineable method to ensure that the `TestAssemblyLoadContext`, `Assembly` and `MethodInfo` (the `Assembly.EntryPoint`) can't be kept alive by stack slot references (real- or JIT-introduced locals). That could keep the `TestAssemblyLoadContext` alive and prevent the unload.
```
mairaw:
```suggestion

This is sufficient to unload the test assembly. Let's actually put all of this into a separate non-inlineable method to ensure that the `TestAssemblyLoadContext`, `Assembly`, and `MethodInfo` (the `Assembly.EntryPoint`) can't be kept alive by stack slot references (real- or JIT-introduced locals). That could keep the `TestAssemblyLoadContext` alive and prevent the unload.
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+
+Let's also return a weak reference to the `AssemblyLoadContext` so that we can use it later to detect unload completion.
```
mairaw:
```suggestion
Also, return a weak reference to the `AssemblyLoadContext` so that you can use it later to detect unload completion.
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+    Assembly a = alc.LoadFromAssemblyPath(assemblyPath);
+
+    var args = new object[1] { new string[] {"Hello"}};
+    int result = (int)a.EntryPoint.Invoke(null, args);
+    alc.Unload();
+
+    return result;
+}
+```
+Now we can run this function to load, execute, and unload the assembly.
```
mairaw:
```suggestion

Now you can run this function to load, execute, and unload the assembly.
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+```csharp
+WeakReference testAlcWeakRef;
+int result = ExecuteAndUnload("absolute/path/to/your/assembly", out testAlcWeakRef);
+```
+However, the unload doesn't complete immediately. As previously mentioned, it relies on the GC to collect all the objects from the test assembly. In many cases, it is not necessary to wait for the unload completion. However there are cases where it is useful to know that the unload has finished. For example, you may want to delete the assembly file that was loaded into the custom `AssemblyLoadContext` from disk. In such a case, the following code snippet can be used. It triggers a GC and waits for pending finalizers in a loop until the weak reference to the custom `AssemblyLoadContext` is set to `null`, indicating the target object was collected. Note that in most cases, just one pass through the loop is required. However for more complex cases where objects created by the code running in the `AssemblyLoadContext` have finalizers, more passes may be needed.
```
mairaw:
```suggestion

However, the unload doesn't complete immediately. As previously mentioned, it relies on the GC to collect all the objects from the test assembly. In many cases, it isn't necessary to wait for the unload completion. However, there are cases where it's useful to know that the unload has finished. For example, you may want to delete the assembly file that was loaded into the custom `AssemblyLoadContext` from disk. In such a case, the following code snippet can be used. It triggers a GC and waits for pending finalizers in a loop until the weak reference to the custom `AssemblyLoadContext` is set to `null`, indicating the target object was collected. Note that in most cases, just one pass through the loop is required. However, for more complex cases where objects created by the code running in the `AssemblyLoadContext` have finalizers, more passes may be needed.

```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+```csharp
+for (int i = 0; testAlcWeakRef.IsAlive && (i < 10); i++)
+{
+    GC.Collect();
+    GC.WaitForPendingFinalizers();
+}
+```
+### The Unloading event
```
mairaw:
```suggestion

### The Unloading event

```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+In some cases, it may be necessary for the code loaded into a custom `AssemblyLoadContext` to perform some cleanup when the unloading is initiated. For example, it may need to stop threads, clean up some strong GC handles, etc. The `Unloading` event can be used in such cases. A handler that performs the necessary cleanup can be hooked to this event.
+### Troubleshooting unloadability issues
```
mairaw:
```suggestion

### Troubleshoot unloadability issues

```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+Due to the cooperative nature of the unloading, it is easy to forget about references keeping the stuff in a collectible `AssemblyLoadContext` alive and preventing unload. Here is a summary of things (some of them non-obvious) that can hold the references:
```
mairaw:
```suggestion
Due to the cooperative nature of the unloading, it's easy to forget about references keeping the stuff in a collectible `AssemblyLoadContext` alive and preventing unload. Here is a summary of things (some of them non-obvious) that can hold the references:

```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+* Regular references held from outside of the collectible `AssemblyLoadContext`, stored in a stack slot or a processor register (method locals, either explicitly created by the user code or implicitly by the JIT), a static variable or a strong / pinning GC handle, and transitively pointing to:
+  * An assembly loaded into the collectible `AssemblyLoadContext`.
+  * A type from such an assembly.
+  * An instance of a type from such an assembly.
+* Threads running code from an assembly loaded into the collectible `AssemblyLoadContext`.
+* Instances of custom non-collectible `AssemblyLoadContext` types created inside of the collectible `AssemblyLoadContext`
+* Pending <xref:System.Threading.RegisteredWaitHandle> instances with callbacks set to methods in the custom `AssemblyLoadContext`
+
+Hints to find stack slot / processor register rooting an object:
```
mairaw:
```suggestion
Hints to find stack slot / processor register rooting an object:

```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+
+The .NET Core 3.0 supports unloadability through <xref:System.Runtime.Loader.AssemblyLoadContext>. You can load a set of assemblies into a collectible `AssemblyLoadContext`, execute methods in them or just inspect them using reflection, and finally unload the `AssemblyLoadContext`. That unloads the assemblies loaded into that `AssemblyLoadContext`.
+
+There is one noteworthy difference between the unloading using `AssemblyLoadContext` and using AppDomains. With AppDomains, the unloading is forced. At the unload time, all threads running in the target AppDomain are aborted, managed COM objects created in the target AppDomain are destroyed, etc. With `AssemblyLoadContext`, the unload is "cooperative". Calling the <xref:System.Runtime.Loader.AssemblyLoadContext.Unload%2A?displayProperty=nameWithType> method just initiates the unloading. It does not complete until:
+- No threads have code from the assemblies loaded into the `AssemblyLoadContext` on their call stacks.
```
mairaw: Following the sentence as a whole reads a bit weird...

It doesn't complete until no threads have code from the assemblies loaded into the `AssemblyLoadContext` on their call stacks.

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
+# Using and debugging unloadability in .NET Core
+An ability to load and later unload a set of assemblies is one of the features that was missing until now in .NET Core. In .NET Framework, custom app domains are used for this purpose. But since .NET Core supports only a single default app domain, how can we achieve that goal in .NET Core?
```
mairaw: These first few paragraphs read a lot like a blog post to me. Also, we need to think of topics as future proof. So saying things like now, etc. should be avoided.

A suggestion on how to get this started:

```suggestion
Starting with .NET Core 3.0, the ability to load and later unload a set of assemblies is supported. In .NET Framework, custom app domains were used for this purpose, but .NET Core only supports a single default app domain.
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+
+The .NET Core 3.0 supports unloadability through <xref:System.Runtime.Loader.AssemblyLoadContext>. You can load a set of assemblies into a collectible `AssemblyLoadContext`, execute methods in them or just inspect them using reflection, and finally unload the `AssemblyLoadContext`. That unloads the assemblies loaded into that `AssemblyLoadContext`.
```
mairaw:
```suggestion
.NET Core 3.0 and later versions support unloadability through <xref:System.Runtime.Loader.AssemblyLoadContext>. You can load a set of assemblies into a collectible `AssemblyLoadContext`, execute methods in them or just inspect them using reflection, and finally unload the `AssemblyLoadContext`. That unloads the assemblies loaded into that `AssemblyLoadContext`.
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+* Threads running code from an assembly loaded into the collectible `AssemblyLoadContext`.
+* Instances of custom non-collectible `AssemblyLoadContext` types created inside of the collectible `AssemblyLoadContext`
+* Pending <xref:System.Threading.RegisteredWaitHandle> instances with callbacks set to methods in the custom `AssemblyLoadContext`
+
+Hints to find stack slot / processor register rooting an object:
+* Passing function call results directly to another function may create a root even though there is no user-created local variable.
+* If a reference to an object was available at any point in a method, the JIT might have decided to keep the reference in a stack slot / processor register for as long as it wants in the current function.
+## Debugging unloading issues
+Debugging issues with unloading can be tedious. You can get into situations where you don't know what can be holding an `AssemblyLoadContext` alive, but the unload fails.
+The best weapon to help with that is Windbg (LLDB on Unix) with the SOS plugin. We need to find what's keeping a `LoaderAllocator` belonging to the specific `AssemblyLoadContext` alive.
```
mairaw:
```suggestion
The best weapon to help with that is Windbg (LLDB on Unix) with the SOS plugin. You need to find what's keeping a `LoaderAllocator` belonging to the specific `AssemblyLoadContext` alive.
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+
+In WinDbg (it seems WinDbg does that automatically when breaking into .NET Core application):
+```console
+.loadby sos coreclr
+```
+In LLDB:
+```console
+plugin load /path/to/libsosplugin.so
+```
+Let's try to debug an example program that has problems with unloading. Source code is included below. When you run it under WinDbg, the program breaks into the debugger right after attempting to check for the unload success. We can then start looking for the culprits.
```
mairaw:
```suggestion
Let's try to debug an example program that has problems with unloading. Source code is included below. When you run it under WinDbg, the program breaks into the debugger right after attempting to check for the unload success. You can then start looking for the culprits.
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+
+Note that if you debug using LLDB on Unix, the SOS commands in the examples below don't have the `!` in front of them.
```
mairaw: don't use below

```suggestion
Note that if you debug using LLDB on Unix, the SOS commands in the following examples don't have the `!` in front of them.

```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+```console
+!dumpheap -type LoaderAllocator
+```
+It will dump all objects with a type name containing `LoaderAllocator` that are in the GC heap. Here is an example:
```
mairaw:
```suggestion
This command dumps all objects with a type name containing `LoaderAllocator` that are in the GC heap. Here is an example:
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+    -> 000002b790001038 System.Object[]
+    -> 000002b78000d390 example.TestInfo
+    -> 000002b78000d328 System.Reflection.RuntimeMethodInfo
+    -> 000002b78000d1f8 System.RuntimeType+RuntimeTypeCache
+    -> 000002b78000d1d0 System.RuntimeType
+    -> 000002b78000ce40 System.Reflection.LoaderAllocator
+
+Found 3 roots.
+```
+Now you need to figure out where the root is located so that you can fix it. The easiest case is when the root is a stack slot or a processor register. In that case, the `gcroot` shows you the name of the function whose frame contains the root and the thread executing that function. The difficult case is when the root is a static variable or a GC handle.
```
mairaw:
```suggestion
Now you need to figure out where the root is located so you can fix it. The easiest case is when the root is a stack slot or a processor register. In that case, the `gcroot` shows you the name of the function whose frame contains the root and the thread executing that function. The difficult case is when the root is a static variable or a GC handle. 
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+    -> 000002b78000d328 System.Reflection.RuntimeMethodInfo
+    -> 000002b78000d1f8 System.RuntimeType+RuntimeTypeCache
+    -> 000002b78000d1d0 System.RuntimeType
+    -> 000002b78000ce40 System.Reflection.LoaderAllocator
+
+Found 3 roots.
+```
+Now you need to figure out where the root is located so that you can fix it. The easiest case is when the root is a stack slot or a processor register. In that case, the `gcroot` shows you the name of the function whose frame contains the root and the thread executing that function. The difficult case is when the root is a static variable or a GC handle. 
+
+In our example above, the first root is a local of type `System.Reflection.RuntimeMethodInfo` stored in the frame of the function `example.Program.Main(System.String[])` at address `rbp-20` (`rbp` is the processor register `rbp` and -20 is a hexadecimal offset from that register).
```
mairaw:
```suggestion
In the previous example, the first root is a local of type `System.Reflection.RuntimeMethodInfo` stored in the frame of the function `example.Program.Main(System.String[])` at address `rbp-20` (`rbp` is the processor register `rbp` and -20 is a hexadecimal offset from that register).
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+* Instances of custom non-collectible `AssemblyLoadContext` types created inside of the collectible `AssemblyLoadContext`
+* Pending <xref:System.Threading.RegisteredWaitHandle> instances with callbacks set to methods in the custom `AssemblyLoadContext`
+
+Hints to find stack slot / processor register rooting an object:
+* Passing function call results directly to another function may create a root even though there is no user-created local variable.
+* If a reference to an object was available at any point in a method, the JIT might have decided to keep the reference in a stack slot / processor register for as long as it wants in the current function.
+## Debugging unloading issues
+Debugging issues with unloading can be tedious. You can get into situations where you don't know what can be holding an `AssemblyLoadContext` alive, but the unload fails.
+The best weapon to help with that is Windbg (LLDB on Unix) with the SOS plugin. We need to find what's keeping a `LoaderAllocator` belonging to the specific `AssemblyLoadContext` alive.
+This plugin allows us to look at GC heap objects, their hierarchies and roots.
```
mairaw:
```suggestion
This plugin allows you to look at GC heap objects, their hierarchies, and roots.
```

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+00007ffadc93a288        1           48 System.Reflection.LoaderAllocator
+Total 2 objects
+```
+In the "Statistics:" part below, check the `MT` (`MethodTable`) belonging to the `System.Reflection.LoaderAllocator`, which is the object we care about. Then in the list at the beginning, find the entry with `MT` matching that one and get the address of the object itself. In our case, it is "000002b78000ce40"
+
+Now that we know the address of the `LoaderAllocator` object, we can use another command to find its GC roots
+```console
+!gcroot -all 0x000002b78000ce40 
+```
+This will dump the chain of object references that lead to the `LoaderAllocator` instance. The list starts with the root, which is the entity that keeps our `LoaderAllocator` alive and thus is the core of the problem we are debugging. The root can be a stack slot, a processor register, a GC handle, or a static variable.
```
mairaw:
```suggestion
This command dumps the chain of object references that lead to the `LoaderAllocator` instance. The list starts with the root, which is the entity that keeps our `LoaderAllocator` alive and thus is the core of the problem you're debugging. The root can be a stack slot, a processor register, a GC handle, or a static variable.
```

mairaw (changes requested):
Thanks for doing this @janvorli. I've added some initial comments to get this going.

mairaw:
Sounds good @rpetrusha about the link to the API!

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+0000001fc727f2f0 00007ffadbc4cbe3 System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
+0000001fc727f568 00007ffae5fdc1e3 [GCFrame: 0000001fc727f568] 
+0000001fc727f7f0 00007ffae5fdc1e3 [DebuggerU2MCatchHandlerFrame: 0000001fc727f7f0] 
+
+```
+As you can see, the last thread has `test.Program.ThreadProc()`. This is a function from the assembly loaded into the `AssemblyLoadContext`, and so it keeps the `AssemblyLoadContext` alive.
+## Example source with unloadability issues
+This example is used in the debugging above.
+### Main testing program
+```csharp
```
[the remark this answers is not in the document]
janvorli: I have a pending PR to put all the snippets into that repo. After it is approved and merged, I'll update this doc to not to have any inline code.

janvorli:
@mairaw thank you so much for the review and the info on the docs authoring pack. I didn't know such a thing exists.

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+        string assemblyPath = _resolver.ResolveAssemblyToPath(name);
+        if (assemblyPath != null)
+        {
+            return LoadFromAssemblyPath(assemblyPath);
+        }
+
+        return null;
+    }
+}
+```
```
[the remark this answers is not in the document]
janvorli: Ah, got it, after installing the doc authoring pack, it complains that there should be an empty line after that.

docs/core/tutorials/unloadability-howto.md
```diff
@@ -0,0 +1,362 @@
...
+
+There is one noteworthy difference between the unloading using `AssemblyLoadContext` and using AppDomains. With AppDomains, the unloading is forced. At the unload time, all threads running in the target AppDomain are aborted, managed COM objects created in the target AppDomain are destroyed, etc. With `AssemblyLoadContext`, the unload is "cooperative". Calling the <xref:System.Runtime.Loader.AssemblyLoadContext.Unload%2A?displayProperty=nameWithType> method just initiates the unloading. It does not complete until:
+- No threads have code from the assemblies loaded into the `AssemblyLoadContext` on their call stacks.
+- There are no strong references to types from the assemblies loaded into the `AssemblyLoadContext`, instances of those types and the assemblies themselves. The GC needs to collect those first.
+## Using collectible AssemblyLoadContext
```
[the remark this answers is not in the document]
janvorli: I am not sure what would you like to see here. You can use Visual Studio or just the SDK package with any editor you want. This seems general to any .NET Core development.

As for the section on debugging issues, it already mentions that you'll need WinDbg on Windows or LLDB on Unix.

docs/core/tutorials/unloadability-howto.md
janvorli: I've reworded this part to make it easier to understand.

janvorli:
@rpetrusha, regarding the broken link - I wonder if the link is really correct, as it contains `%2A` in the middle. Or does it have a special meaning?
`<xref:System.Runtime.Loader.AssemblyLoadContext.Unload%2A?displayProperty=nameWithType>`

rpetrusha:
@janvorli, ordinarily, links have to be fully qualified with parameter types. %2A indicates that the link should be to the overload page rather than to a specific overload.

rpetrusha:
Closing and reopening to begin new build after dotnet/samples#617 was merged.

janvorli:
@rpetrusha, @mairaw  I've updated the doc so that the last two code blocks are not inline anymore and point to the samples repo. As for the rest of those, I am actually wondering whether to leave them all inline or whether to update the doc to replace the ones that are compilable with a reference to the samples and only keep the snippets that cannot be compiled on their own inline.
Do you have a preference w.r.t. that?

janvorli:
@rpetrusha, @mairaw can you please reply to my previous comment? I would like to merge the doc in soon.

rpetrusha:
I apologize for the long delay in responding, @janvorli. The best option is to place all compilable samples in the samples repo, with small snippets inline.  In addition, though, some of the inline snippets can be referenced from the samples repo. For example, the currently inline code 
```
for (int i = 0; testAlcWeakRef.IsAlive && (i < 10); i++)
{
    GC.Collect();
    GC.WaitForPendingFinalizers();
}
```
is included in your main testing program. So we'd want to pull that snippet into the build, instead of having two versions existing independently. The way to do that is to add a sniippet tag with a label (the snippet tags are removed from the documentation when it's built):
```
// <Snippet1>
for (int i = 0; testAlcWeakRef.IsAlive && (i < 10); i++)
{
    GC.Collect();
    GC.WaitForPendingFinalizers();
}
// </Snippet1>
```
Then add the reference to the snippet to the topic by including the snippet label without the string "snippet":

\[!code-csharp[Main testing program]\(~/samples/snippets/core/tutorials/unloading/unloadability_issues_example_main.cs#1)]

But it's OK to leave most the remaining one- or two-line snippets in the "Use a custom collectible AssemblyLoadContext" section inline, since they don't exactly correspond to code that would go in the repo.

mairaw:
Also closing and reopening this to see if the broken link would go away - it should since I loaded the 3.0 APIs.

docs/core/tutorials/unloadability-howto.md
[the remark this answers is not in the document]
mairaw: Yep, but any tutorial that we do we list what are the prereqs.

Here are some examples:
https://docs.microsoft.com/en-us/dotnet/core/tutorials/with-visual-studio-code#prerequisites
https://docs.microsoft.com/en-us/dotnet/core/tutorials/with-visual-studio#prerequisites
https://docs.microsoft.com/en-us/dotnet/core/tutorials/using-with-xplat-cli#prerequisites

rpetrusha:
Thanks, @mairaw; I'd meant to close and reopen, but I forgot. The broken link should definitely go away.

There are a couple of additional things to do, @janvorli:
- Maira and I thought that the best location for this topic would be a how-to directory (dotnet/docs/core/how-to) rather than the tutorials directory. Could you create the directory and move the topic to it? (That also means that you should change the path in the samples-and-tutorials/index.md document in your PR.)
- That also requires the creation of the an index.md page whose title is ".NET Core how-to articles"  for the directory, which at this point would just have a single entry (your topic).
- It also requires that both the top-level how-to-page and your article be added to the TOC (found in dotnet/docs/docs/toc.md).
- Finally, it would be good to add a link to your article to the managed reference documentation. Once your PR is merged, I can take care of doing that.

janvorli:
@rpetrusha, @mairaw  thank you for your feedback. Could you please give me an advise on the description  / title in the new `dotnet/docs/core/how-to/index.md`? I have hard time coming up with wording that would differentiate it from the tutorials and make it general enough for possible future articles in this area.

janvorli:
And I have additional questions:
Can snippets be embedded in other snippets? For example, can I do this?
```csharp
class Test
{    
    // <Snippet2>
    [MethodImpl(MethodImplOptions.NoInlining)]
    static int ExecuteAndUnload(string assemblyPath, out WeakReference alcWeakRef)
    {
    // <Snippet3>
        var alc = new TestAssemblyLoadContext();
        Assembly a = alc.LoadFromAssemblyPath(assemblyPath);
    // </Snippet3>

        alcWeakRef = new WeakReference(alc, trackResurrection: true);

    // <Snippet4>
        var args = new object[1] { new string[] {"Hello"}};
        int result = (int)a.EntryPoint.Invoke(null, args);
    // </Snippet4>

    // </Snippet5>
        alc.Unload();
    // </Snippet5>

        return result;
    }
    // </Snippet2>
}
```

And when you say the sample should be compilable, what does it really mean? Should it contain all the "using xxxx" and have methods placed inside of classes so that it can be pasted into a .cs file and compiled without any additional touches?

rpetrusha:
Yes, @janvorli,  snippets can be embedded in other snippets.
"Compilable" means that the source code in dotnet/samples should be able to be compiled from the command line (for .NET Framework source) or with the dotnet CLI. We've also usually been including a *.csproj or *.vbproj file for source code that targets .NET Core. You can choose whether to present a all of the source code file in the docs (in that case, it should be compilable, with using statements, etc., if it is copied and pasted) or just a code fragment.

rpetrusha:
I also suspect that I gave you some bad guidance, @janvorli. In thinking about it, rather than creating a separate .NET Core how to node in the TOC, it's better to locate this topic in the Assemblies in .NET section of the .NET Guide. The section is sparse (it only has a top-level overview topic and a child topic on the assembly file format) largely because a while ago assembly-related topics were treated as language topics and duplicated in the C# and Visual Basic documentation. We're planning to consolidate the documentation by moving it to the .NET Guide.

So the file can be moved to ~/docs/docs/standard/assembly/. Then the index.md file in that directory can be updated and the article can be added to the TOC at ~/docs/docs/toc.md. 

I'm also wondering if we should rename it to "How to: Load, unload, and debug assembly loading in .NET Core".

janvorli:
@rpetrusha I've updated the snippets in the samples repo so that they use the <snippetXX> tags and  thus reduced their number. I've also moved them to a new location that corresponds to the new suggested location of the doc. Can you please take a look so that I can update my doc accordingly?

rpetrusha:
I've merged dotnet/samples#677, @janvorli.

janvorli:
I've somehow screwed the commits. Will fix it soon.

janvorli:
@karelz, @JRAlexander, @BillWagner, @cartermp, @Lxiamail - please ignore the request to review. I have not sent them, they somehow got added automatically based on a bad merge commit that I've fixed.

janvorli:
@rpetrusha I've tried to reference the snippets, but it works only for the snippet #1, for the rest it complains that the links are invalid. I have no idea what could be causing that. Do you have any idea what can be wrong?

docs/standard/assembly/unloadability-howto.md
```diff
@@ -0,0 +1,247 @@
...
+
+[!code-csharp[Advanced custom AssemblyLoadContext](~/samples/snippets/standard/assembly/unloading/complex_assemblyloadcontext.cs)]
+
+### Use a custom collectible AssemblyLoadContext
+
+This section assumes the simpler version of the `TestAssemblyLoadContext` is being used.
+
+You can create an instance of the custom `AssemblyLoadContext` and load an assembly into it as follows:
+
+!code-csharp[Part 1](~/samples/snippets/standard/assembly/unloading/simple_example.cs#3)]
```
mairaw: I think you're just missing part of the syntax 

```suggestion
[!code-csharp[Part 1](~/samples/snippets/standard/assembly/unloading/simple_example.cs#3)]
```

janvorli: Ah, it was right under my nose! Thank you!

mairaw:
Also, feel free to just push new commits, since we squash all of them when we merge. We like to have each PR as a single commit in our history.

docs/standard/assembly/unloadability-howto.md
```diff
@@ -0,0 +1,247 @@
...
+The third root is a pinned `GCHandle`. This one is actually a static variable. Unfortunately, there is no way to tell. Statics for reference types are stored in a managed object array in internal runtime structures.
+
+Another case that can prevent unloading of an `AssemblyLoadContext` is when a thread has a frame of a method from an assembly loaded into the `AssemblyLoadContext` on its stack. You can check that by dumping managed call stacks of all threads:
+
+```console
+~*e !clrstack
+```
+
+The command means "apply to all threads the `!clrstack` command". Here is the output of that command for our example. Unfortunately, LLDB on Unix doesn't have any way to apply a command to all threads, so you'll need to resort to manually switching threads and repeating the `clrstack` command.
+You should ignore all threads where the debugger says "Unable to walk the managed stack."
```
mairaw:
```suggestion
Ignore all threads where the debugger says "Unable to walk the managed stack."
```

docs/standard/assembly/unloadability-howto.md
```diff
@@ -0,0 +1,247 @@
...
+
+The third root is a pinned `GCHandle`. This one is actually a static variable. Unfortunately, there is no way to tell. Statics for reference types are stored in a managed object array in internal runtime structures.
+
+Another case that can prevent unloading of an `AssemblyLoadContext` is when a thread has a frame of a method from an assembly loaded into the `AssemblyLoadContext` on its stack. You can check that by dumping managed call stacks of all threads:
+
+```console
+~*e !clrstack
+```
+
+The command means "apply to all threads the `!clrstack` command". Here is the output of that command for our example. Unfortunately, LLDB on Unix doesn't have any way to apply a command to all threads, so you'll need to resort to manually switching threads and repeating the `clrstack` command.
```
mairaw:
```suggestion
The command means "apply to all threads the `!clrstack` command". The following is the output of that command for the example. Unfortunately, LLDB on Unix doesn't have any way to apply a command to all threads, so you'll need to resort to manually switching threads and repeating the `clrstack` command.
```

docs/standard/assembly/unloadability-howto.md
```diff
@@ -0,0 +1,247 @@
...
+
+This example is used in the debugging above.
+
+### Main testing program
+
+[!code-csharp[Main testing program](~/samples/snippets/standard/assembly/unloading/unloadability_issues_example_main.cs)]
+
+## Program loaded into the TestAssemblyLoadContext
+
+This is the `test.dll` passed to the `ExecuteAndUnload` method in the main testing program.
```
mairaw:
```suggestion
The following code represents the `test.dll` passed to the `ExecuteAndUnload` method in the main testing program.
```

docs/standard/assembly/unloadability-howto.md
```diff
@@ -0,0 +1,247 @@
...
+0000001fc727f568 00007ffae5fdc1e3 [GCFrame: 0000001fc727f568] 
+0000001fc727f7f0 00007ffae5fdc1e3 [DebuggerU2MCatchHandlerFrame: 0000001fc727f7f0] 
+
+```
+
+As you can see, the last thread has `test.Program.ThreadProc()`. This is a function from the assembly loaded into the `AssemblyLoadContext`, and so it keeps the `AssemblyLoadContext` alive.
+
+## Example source with unloadability issues
+
+This example is used in the debugging above.
```
mairaw:
```suggestion
The following code is used in the previous debugging example.
```

mairaw (approved):
I'm approving the PR but left a few suggestions for you to consider. @rpetrusha will have to approve it as well since he requested some changes earlier.

janvorli:
Closing and reopeing so that it picks a typo fix in a snippet tag in the samples repo

rpetrusha (approved):

rpetrusha:
This looks good, @janvorli. I'll merge your PR now.

PR #10269, opened 2019-02-01
Remove referential transparecy from the intro to FP

### LINKED ISSUE #10253: Referential transparency and side effects
This document says that a referentially transparent expression can be replaced with its output without affecting a program's behavior and also says that it does not have to be pure. I think that two sentences are contradictory. If an expression does not always return same value for same arguments it can't be replaced with its output. It's so clear. And if an expression has side effects it can't be replaced with its output too. Let me explain.

Let's assume some function that returns the length of the output stream.

```fs
let readOutputStreamLength = ...
```

This document defines `addOneToValue` function and we know that it has a side effect because it uses `printfn` function.

```fs
let addOneToValue x = 
    printfn "x is %d" x
    x + 1
```

Consider following program code.

```fs
let x = addOneToValue 1
readOutputStreamLength
```

Given that the output stream is empty before execute the program then the result will be `12`. But if replace `addOneToValue 1` with `2` the result of the program will be `0`. The program's behavior has been changed. That means `addOneToValue` function is referentially opaque. The reason why is that `addOneToValue` function has a side effect. So an expression should be pure to be referentially transparent. 

[The definition of referential transparency in Wikipedia](https://en.wikipedia.org/wiki/Referential_transparency) backs up my opinion.

>  An expression is called referentially transparent if it can be replaced with its corresponding value without changing the program's behavior. **This requires that the expression is pure**, that is to say the expression value must be the same for the same inputs and its evaluation must have no side effects.

---
#### Document Details

* ID: 1df55ba4-e543-c31c-ba44-e2ae6e78fd33
* Version Independent ID: c88d5609-718e-3df6-6ae5-613438fd8295
* Content: [Introduction to Functional Programming in F#](https://docs.microsoft.com/en-us/dotnet/fsharp/introduction-to-functional-programming/)
* Content Source: [docs/fsharp/introduction-to-functional-programming/index.md](https://github.com/dotnet/docs/blob/master/docs/fsharp/introduction-to-functional-programming/index.md)
* Product: **dotnet-fsharp**
* GitHub Login: @cartermp
* Microsoft Alias: **phcart**

cartermp:
Fixes #10253

After discussion there, it's become clear that this section isn't the most valuable in the context of the whole article.

### DIFF
--- a/docs/fsharp/introduction-to-functional-programming/index.md
+++ b/docs/fsharp/introduction-to-functional-programming/index.md
@@ -156,7 +156,7 @@ let addOneToValue x =
     x + 1
 ```
 
-Although this function does not depend on a global value, it writes the value of `x` to the output of the program. Although there is nothing inherently wrong with doing this, it does mean that the function is not pure. If another part of your program depends on something such as the output buffer, then calling this function can affect that other part of your program.
+Although this function does not depend on a global value, it writes the value of `x` to the output of the program. Although there is nothing inherently wrong with doing this, it does mean that the function is not pure. If another part of your program depends on something external to the program, such as the output buffer, then calling this function can affect that other part of your program.
 
 Removing the `printfn` statement makes the function pure:
 
--- a/docs/fsharp/introduction-to-functional-programming/index.md
+++ b/docs/fsharp/introduction-to-functional-programming/index.md
@@ -164,7 +164,7 @@ Removing the `printfn` statement makes the function pure:
 let addOneToValue x = x + 1
 ```
 
-Although this function is not inherently _better_ than the previous version with the `printfn` statement, it does guarantee that all this function does is return a value. Calling this function once or 1 billion times will still result in the same thing: just producing a value. The predictability given by purity is something many functional programmers strive for.
+Although this function is not inherently _better_ than the previous version with the `printfn` statement, it does guarantee that all this function does is return a value. Calling this function any number of times produces the same result: it just produces a value. The predictability given by purity is something many functional programmers strive for.
 
 ### Immutability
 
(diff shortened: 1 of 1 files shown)


docs/fsharp/introduction-to-functional-programming/index.md
```diff
@@ -156,90 +156,15 @@ let addOneToValue x =
     x + 1
 ```
 
-Although this function does not depend on a global value, it writes the value of `x` to the output of the program. Although there is nothing inherently wrong with doing this, it does mean that the function is not pure.
+Although this function does not depend on a global value, it writes the value of `x` to the output of the program. Although there is nothing inherently wrong with doing this, it does mean that the function is not pure. If another part of your program depends on something such as the output buffer, then calling this function can affect that other part of your program.
```
rpetrusha: - "depends on something such as the output buffer" seems vague. Can there be some other dependency. Perhaps just "depends on the output buffer".

docs/fsharp/introduction-to-functional-programming/index.md
```diff
@@ -156,90 +156,15 @@ let addOneToValue x =
...
-    x + 1
-```
-
-Any call to this function can also be replaced by its body and the same things will happen each time:
-
-* The value, prior to being added to, is printed to the output
-* The value has 1 added to it
-
-When programming in F#, it is often referential transparency that is the goal, rather than purity. However, it is still good practice to write pure functions when you can.
+Although this function is not inherently _better_ than the previous version with the `printfn` statement, it does guarantee that all this function does is return a value. Calling this function once or 1 billion times will still result in the same thing: just producing a value. The predictability given by purity is something many functional programmers strive for.
```
rpetrusha:
```suggestion
Although this function is not inherently _better_ than the previous version with the `printfn` statement, it does guarantee that all this function does is return a value. Calling this function any number of times produces the same result: it just produces a value. The predictability given by purity is something many functional programmers strive for.
```

rpetrusha (approved):
I've left a couple of suggestions, @cartermp. You can merge when you're ready.

PR #10287, opened 2019-02-02
Update how-to-create-temporary-certificates-for-use-during-developmen…

roger-cruz:
…t.md

I believe the filename TempCA.cer is incorrect here.  Given the examples above where the self-signed certificate was exported as RootCA.pfx, the instructions here should be changed to use that file.

## Summary

Describe your changes here.


rpetrusha (approved):
Thank you, @roger-cruz, for changing the file name of the certificate to correspond to the earlier part of the example. We'll merge your PR now.

PR #10294, opened 2019-02-02
Byte can be simplified to byte

hassanhabib:
## Simplify Byte

Using ``` byte ``` instead of ``` Byte ```.

### DIFF
--- a/docs/csharp/programming-guide/types/index.md
+++ b/docs/csharp/programming-guide/types/index.md
@@ -83,7 +83,7 @@ Value types and reference types in the CTS
   
 ```csharp  
 // Static method on type Byte.  
-byte b = Byte.MaxValue;  
+byte b = byte.MaxValue;
 ```  
   
  But you declare and assign values to them as if they were simple non-aggregate types:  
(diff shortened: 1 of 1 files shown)


docs/csharp/programming-guide/types/index.md
```diff
@@ -83,7 +83,7 @@ Value types and reference types in the CTS
   
 ```csharp  
 // Static method on type Byte.
```
rpetrusha:
```suggestion
// Static method on type byte.  
```

rpetrusha (changes requested):
Thanks for contributing to the dotnet/docs repo, @hassanhabib. If Byte is changed to byte in the variable declaration, it should also be changed in the comment.

docs/csharp/programming-guide/types/index.md
hassanhabib: Good catch! thanks. fixed.

rpetrusha (approved):

rpetrusha:
Thank you for making the additional change, @hassanhabib. We'll merge your PR now.

PR #10356, opened 2019-02-07
Fix broken table in Obsolete Types article

reflectronic:


mairaw (approved):
Thank you @reflectronic for fixing the broken table. Your changes should be live within a day. We really appreciate your contribution!

PR #10392, opened 2019-02-08
Update object expression formatting

cartermp:
It was kind of weird


mairaw (approved):
LGTM. worth updating the date?

also, there are two issues for this article. Worth tackling in this PR?

PR #10403, opened 2019-02-09
Fix header sizing for consistency in Extensibility doc

zach27:
## Summary

There are three main headings in this doc that should be at the same level:

 - Per-project based extensibility
 - Custom targets
 - PATH-based extensibility

However, the last two were incorrectly a level lower than the first, making the structure confusing.


rpetrusha (approved):
Thank you, @zach27, for correcting the level of these two headers. We'll merge your PR now. Your changes should be live on docs.microsoft.com in the next day or two.

PR #10409, opened 2019-02-10
Implicit numeric conversions: fix remarks

pkulikov:
There are implicit conversions from `char`: the third row of the [table](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/implicit-numeric-conversions-table) is about that.


BillWagner (approved):
Thanks for spotting this and supplying the fix @pkulikov 

Thanks again for all your contributions! 🌠

PR #10534, opened 2019-02-18
GetMemberProps is a helper

JackTrapper:
Pointing out that GetMemberProps is a helper, and it will call GetMethodProps or GetFieldProps based on the type of token being passed.

## Summary

From the 2001 Unamanged Metadata API documentation:

> Gets the information stored in metadata for a specified member definition. This is a
simple helper method: if _md_ is a MethodDef, then we call _GetMethodProps_; if _md_ is a
FieldDef, then we call _GetFieldProps_. See these other methods for details.

### DIFF
--- a/docs/framework/unmanaged-api/metadata/imetadataimport-getmemberprops-method.md
+++ b/docs/framework/unmanaged-api/metadata/imetadataimport-getmemberprops-method.md
@@ -19,7 +19,7 @@ author: "mairaw"
 ms.author: "mairaw"
 ---
 # IMetaDataImport::GetMemberProps Method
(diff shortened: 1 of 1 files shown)


rpetrusha:
@hoyosjs and @noahfalk, could you take a look at this and the other PRs from @JackTrapper (#10535, #10547, #10548, #10587)?

noahfalk (approved):

mairaw (approved):
Thank you for your contribution @JackTrapper and for the review @noahfalk. Merging this. Your fix should be live within a day or so.

PR #10628, opened 2019-02-21
Add "What's new" for C# 8 preview 2

### LINKED ISSUE #10266: Create Preview 2 version of What's new in C#
Now that Preview 2 has shipped, there are a few new features that have been implemented and shipped.  This include:

- Nullable reference types
- ranges
- async streams
- pattern matching expressions
- static local functions

The new article should give a brief overview of those features and refer to the csharplang proposals for the details.  Where we have (or are) adding tutorials for the preview features we will link to those tutorials.

Finally, the redirection for /whats-new should go to this new page.

BillWagner:
Fixes #10266

### DIFF
--- a/docs/csharp/whats-new/csharp-8.md
+++ b/docs/csharp/whats-new/csharp-8.md
@@ -264,10 +264,10 @@ You can learn more about the feature in the overview of [nullable reference type
 Starting with C# 8.0, you can create and consume streams asynchronously. A method that returns an asynchronous stream has three properties:
 
 1. It was declared with the `async` modifier.
-1. It returns an <xref:System.Collections.Generic,IAsyncEnumerable%601>.
+1. It returns an <xref:System.Collections.Generic.IAsyncEnumerable%601>.
 1. The method contains `yield return` statements to return successive elements in the asynchronous stream.
 
-Consuming an asynchronous stream requires you to add the `await` keyword before the `foreach` keyword when you enumerate the elements of the stream. Adding the `await` keyword requires the method that enumerates the asynchronous stream to be declared with the `async` modifier and to return a type allowed for an `async` method. Typically that means returning a <xref:System.Threading.Tasks.Task> or <xref:System.Threading.Tasks.Task%601>. It can also be a <xref:System.Threading.Tasks.ValueTask> or <xref:System.Threading.Tasks.ValueTask%601>. A method can both consume and produce an asynchronous stream, which means it would return an <xref:System.Collections.Generic,IAsyncEnumerable%601>. The following code generates a sequence from 1 to 20, waiting 100 ms between generating each number:
+Consuming an asynchronous stream requires you to add the `await` keyword before the `foreach` keyword when you enumerate the elements of the stream. Adding the `await` keyword requires the method that enumerates the asynchronous stream to be declared with the `async` modifier and to return a type allowed for an `async` method. Typically that means returning a <xref:System.Threading.Tasks.Task> or <xref:System.Threading.Tasks.Task%601>. It can also be a <xref:System.Threading.Tasks.ValueTask> or <xref:System.Threading.Tasks.ValueTask%601>. A method can both consume and produce an asynchronous stream, which means it would return an <xref:System.Collections.Generic.IAsyncEnumerable%601>. The following code generates a sequence from 1 to 20, waiting 100 ms between generating each number:
 
 ```csharp
 public static async System.Collections.Generic.IAsyncEnumerable<int> GenerateSequence()
--- a/docs/csharp/whats-new/csharp-8.md
+++ b/docs/csharp/whats-new/csharp-8.md
@@ -23,13 +23,13 @@ The following language features first appeared in C# 8.0 preview 1:
 
 - [Nullable reference types](#nullable-reference-types)
 - [Asynchronous streams](#asynchronous-streams)
-- [Ranges and indices](#ranges-and-indices)
+- [Indices and ranges](#indices-and-ranges)
 
 The remainder of this article briefly describes these features. Where in-depth articles are available, links to those tutorials and overviews are provided.
 
 ## More patterns in more places
 
-**Pattern matching** gives tools to provide shape-dependent functionality across related but different kinds of data. C# 7.0 introduced syntax for type patterns and constant patterns by using the `is` expression and the `switch` statement. These features represented the first tentative steps toward supporting programming paradigms where data and functionality live apart. As the industry moves toward more microservices and other cloud-based architectures, other language tools are needed.
+**Pattern matching** gives tools to provide shape-dependent functionality across related but different kinds of data. C# 7.0 introduced syntax for type patterns and constant patterns by using the [`is`](../language-reference/keywords/is.md) expression and the [`switch`](../language-reference/keywords/switch.md) statement. These features represented the first tentative steps toward supporting programming paradigms where data and functionality live apart. As the industry moves toward more microservices and other cloud-based architectures, other language tools are needed.
 
 C# 8.0 expands this vocabulary so you can use more pattern expressions in more places in your code. Consider these features when your data and functionality are separate. Consider pattern matching when your algorithms depend on a fact other than the runtime type of an object. These techniques provide another way to express designs.
 
@@ -69,7 +69,7 @@ public static RGBColor fromRainbow(Rainbow colorBand) =>
 
 There are several syntax improvements here:
 
-- The variable comes before the `switch` keyword. That enables switch expression to compose using recursive patterns. The different order makes it visually easy to distinguish the switch expression from the switch statement.
+- The variable comes before the `switch` keyword. The different order makes it visually easy to distinguish the switch expression from the switch statement.
 - The `case` and `:` elements are replaced with `=>`. It's more concise.
 - The `default` case is replaced with a `_` discard.
 - The bodies are expressions, not statements.
@@ -190,6 +190,8 @@ static void WriteLinesToFile(IEnumerable<string> lines)
 }
 ```
 
+In the preceding example, the file is disposed when the closing brace associated with the `using` statement is reached.
+
 A few braces are removed by replacing the `using` statement with a `using` declaration:
 
 ```csharp
@@ -207,6 +209,8 @@ static void WriteLinesToFile(IEnumerable<string> lines)
 // file is disposed here
 }
 ```
+
+In the preceding example, the file is disposed when the closing brace for the method is reached. That's the end of the scope in which `file` is declared.
 
 Using declarations provide the same behavior as [using statements](../language-reference/keywords/using-statement.md) when exceptions are thrown: the `Dispose` call is in a `finally` block.
 
@@ -227,6 +231,19 @@ int M()
 }
 ```
 
+The following code contains a static local function. It can be static because it doesn't access any variables in the enclosing scope:
+
+```csharp
+int M()
+{
+    int y = 5;
+    int x = 7;
+    return Add(x, y);
+
+    int Add(int left, int right) => left + right;
+}
+```
+
 ## Disposable ref structs
 
 A `struct` declared with the `ref` modifier may not implement any interfaces and so cannot implement <xref:System.IDisposable>. Therefore, to enable a `ref struct` to be disposed, it must have an accessible `void Dispose()` method.
@@ -235,7 +252,7 @@ A `struct` declared with the `ref` modifier may not implement any interfaces and
 
 Inside a nullable annotation context, any variable of a reference type is considered to be a **nonnullable reference type**. If you want to indicate that a variable may be null, you must append the type name with the `?` to declare the variable as a **nullable reference type**.
 
-For nonnullable reference types, the compiler uses flow analysis to ensure that local variables are initialized to a non-null value when declared. Fields must be initialized to a non-null value using either an initializer or all constructors. Furthermore, nonnullable reference types can't be assigned a value that could be null.
+For nonnullable reference types, the compiler uses flow analysis to ensure that local variables are initialized to a non-null value when declared. Fields must be initialized during construction. The compiler generates a warning if the variable is not set in a constructor, or an initializer. Furthermore, nonnullable reference types can't be assigned a value that could be null.
 
 Nullable reference types aren't checked to ensure they aren't assigned or initialized to null. However, the compiler uses flow analysis to ensure that any variable of a nullable reference type is checked against null before it's accessed or assigned to a nonnullable reference type.
 
@@ -249,13 +266,39 @@ Starting with C# 8.0, you can create and consume streams asynchronously. A metho
 1. It returns an <xref:System.Collections.Generic,IAsyncEnumerable%601>.
 1. The method contains `yield return` statements to return successive elements in the asynchronous stream.
 
-Consuming an asynchronous stream requires you to add the `await` keyword before the `foreach` keyword when you enumerate the elements of the stream. Adding the `await` keyword requires the method that enumerates the asynchronous stream to be declared with the `async` modifier and to return a type allowed for an `async` method. Typically that means returning a <xref:System.Threading.Tasks.Task> or <xref:System.Threading.Tasks.Task%601>. It can also be a <xref:System.Threading.Tasks.ValueTask> or <xref:System.Threading.Tasks.ValueTask%601>. A method can both consume and produce an asynchronous stream, which means it would return an <xref:System.Collections.Generic,IAsyncEnumerable%601>.
+Consuming an asynchronous stream requires you to add the `await` keyword before the `foreach` keyword when you enumerate the elements of the stream. Adding the `await` keyword requires the method that enumerates the asynchronous stream to be declared with the `async` modifier and to return a type allowed for an `async` method. Typically that means returning a <xref:System.Threading.Tasks.Task> or <xref:System.Threading.Tasks.Task%601>. It can also be a <xref:System.Threading.Tasks.ValueTask> or <xref:System.Threading.Tasks.ValueTask%601>. A method can both consume and produce an asynchronous stream, which means it would return an <xref:System.Collections.Generic,IAsyncEnumerable%601>. The following code generates a sequence from 1 to 20, waiting 100 ms between generating each number:
+
+```csharp
+public static async System.Collections.Generic.IAsyncEnumerable<int> GenerateSequence()
+{
+    for (int i = 0; i < 20; i++)
+    {
+        await Task.Delay(100);
+        yield return i;
+    }
+}
+```
+
+You would enumerate the sequence using the `await foreach` statement:
+
+```csharp
+await foreach (var number in GenerateSequence())
+{
+    Console.WriteLine(number);
+}
+```
 
 You can try asynchronous streams yourself in our tutorial on [creating and consuming async streams](../tutorials/generate-consume-asynchronous-stream.md).
 
-## Ranges and indices
+## Indices and ranges
 
-Ranges and indices provide a succinct syntax for specifying subranges in an array, <xref:System.Span%601>, or <xref:System.ReadOnlySpan%601>. Consider the following array:
+Ranges and indices provide a succinct syntax for specifying subranges in an array, <xref:System.Span%601>, or <xref:System.ReadOnlySpan%601>.
+
+You can specify an index **from the end**. You specify **from the end** using the `^` operator. You are familiar with `array[2]` meaning the element "2 from the start". Now, `array[^2]` means the element "2 from the end". The index `^0` means "the end", or one past that last element.
+
+You can specify a **range** with the **range operator**: `..`. For example, `0..^0` specifies the entire range of the array: 0 from the start up to, but not including 0 from the end. Either operand may use "from the start" or "from the end". Furthermore, either operand may be omitted. The defaults are `0` for the start index, and `^0` for the end index.
+
+Let's look at a few examples. Consider the following array:
 
 ```csharp
 var words = new string[]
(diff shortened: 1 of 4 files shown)


docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,314 @@
...
+You can also specify ranges of an array. The following code creates a subrange with the words "quick", "brown", and "fox":
+
+```csharp
+var brownFox = words[1..4];
+```
+
+Ranges can be specified from the end as well as the beginning. The following code creates a subrange with "lazy" and "dog":
+
+```csharp
+var lazyDog = words[^2..^0];
```
rpetrusha: Also confusing. If indexing from the last element is one-based, why isn't this `^2..^1`?

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,314 @@
...
+```csharp
+Console.WriteLine($"The last word is {words[^1]}");
+```
+
+When you take only a single element and specify from the end, indexing starts at 1.
+
+You can also specify ranges of an array. The following code creates a subrange with the words "quick", "brown", and "fox":
+
+```csharp
+var brownFox = words[1..4];
```
rpetrusha: It seems confusing that this returns only 3 words, when you'd expect it to return the elements at indices 1 through 4, or 4 words An explanation of the syntax would help.

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,314 @@
...
+
+1. It was declared with the `async` modifier.
+1. It returns an `IAsyncEnumerable<T>`.
+1. The method contains `yield return` statements to return successive elements in the asynchronous stream.
+
+Consuming an asynchronous stream requires you to add the `await` keyword before the `foreach` keyword when you enumerate the elements of the stream. Adding the `await` keyword requires the method that enumerates the asynchronous stream was declared with the `async` modifier and returns a type allowed for an `async` method. Typically that means returning a <xref:System.Threading.Tasks.Task> or <xref:System.Threading.Tasks.Task%601>. It can also be a <xref:System.Threading.Tasks.ValueTask> or <xref:System.Threading.Tasks.ValueTask%601>. A method can both consume and produce an asynchronous stream, which means it would return an `IAsyncEnumerable<T>`.
+
+You can try asynchronous streams yourself in our tutorial on [creating and consuming async streams](../tutorials/generate-consume-asynchronous-stream.md).
+
+## Ranges and indices
```
rpetrusha: This is a really useful feature, but I just don't get the syntax. It would be helpful to briefly describe the syntax of ranges and indices rather than just provide a set of examples. (Unless I'm just being obtuse...)

rpetrusha (approved):
This is a really useful overview with some great examples, @BillWagner. I've left a number of comments and suggestions. Let me know if you want an additional review before you merge.

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,314 @@
...
+
+Ranges and indices provide a succinct syntax for specifying subranges in an array, <xref:System.Span%601>, or <xref:System.ReadonlySpan%601>. Consider the following array:
```
rpetrusha:
```suggestion
Ranges and indices provide a succinct syntax for specifying subranges in an array, <xref:System.Span%601>, or <xref:System.ReadOnlySpan%601>. Consider the following array:
```

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,314 @@
...
+        Rainbow.Blue   => new RGBColor(0x00, 0x00, 0xFF),
+        Rainbow.Indigo => new RGBColor(0x4B, 0x00, 0x82),
+        Rainbow.Violet => new RGBColor(0x94, 0x00, 0xD3),
+        _              => throw new ArgumentException(message: "invalid enum value", paramName: nameof(colorBand)),
+    };
+```
+
+There are several syntax improvements here.
+
+- The variable comes before the `switch` keyword. That enables switch expression to compose using recursive patterns. The different order makes it visually easy to distinguish from the switch statement.
```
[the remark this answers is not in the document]
BillWagner: I removed that sentence here. I added a recursive example later on.

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,314 @@
...
+    (var x, var y) => "on a border",
+    _ => "unknown"
+};
+```
+
+## using declarations
+
+A **using declaration** is a variable declaration preceded by the `using` keyword. It tells the compiler that the variable being declared should be disposed at the end of the enclosing scope. For example, consider the following code that writes a text file:
+
+```csharp
```
Thraka: In the previous pattern matching section you used the format:
```
NEW

compare to

OLD
```

but in this one you are starting with the old way of doing it and showing the new way. I would rather see the new way upfront because most likely everyone is familiar with how to write a `using` statement and would rather just see the change first.

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,314 @@
...
+}
+```
+
+Using declarations provide the same behavior when exceptions are thrown: the `Dispose()` call is in a `finally` block.s
+
+## Static local functions
+
+You can now add the `static` modifier to local functions to ensure that local function doesn't capture (reference) any variables from the enclosing scope. Doing so generates `CS8421`, "A static local function can't contain a reference to <variable>." 
+
+Consider the following code. The local function `LocalFunction` accesses the variable `y`, declared in the enclosing scope (the method `M`). Therefore, `LocalFunction` can't be declared with the `static` modifier:
```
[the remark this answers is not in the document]
Thraka: I agree with this comment. One place I used a local function was to do a comparison on a backing `List`. I had a `AddObject` type of method that looked at the object and added it to one of four lists and then called `List.Sort` passing in the local function to do the compare. In C#8 that will now be a static function since it was a compare method

```csharp
int CompareComponent(IComponent left, IComponent right)
{
    if (left.SortOrder > right.SortOrder)
        return 1;

    if (left.SortOrder < right.SortOrder)
        return -1;

    return 0;
}
```

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,314 @@
...
+
+## Disposable ref structs
+
+A `struct` declared with the `ref` modifier may not implement any interfaces. Therefore, to enable a `ref struct` to be disposed, it need not implement <xref:System.IDisposable>, but it must have an accessible `void Dispose()` method.
+
+## Nullable reference types
+
+Inside a nullable annotation context, any variable of a reference type is considered to be a **nonnullable reference type**. If you want to indicate that a variable may be null, you must append the type name with the `?` to declare the variable as a **nullable reference type**.
+
+For nonnullable reference types, the compiler uses flow analysis to ensure that local variables are initialized to a non-null value when declared. Fields must be initialized to a non-null value using either an initializer or all constructors. Furthermore, nonnullable reference types can't be assigned a value that could be null.
```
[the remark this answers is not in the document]
BillWagner: I reworked the sentence.

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,314 @@
...
+    LocalFunction();
+    return y;
+
+    void LocalFunction() => y = 0;
+}
+```
+
+## Disposable ref structs
+
+A `struct` declared with the `ref` modifier may not implement any interfaces and so cannot implement <xref:System.IDisposable>. Therefore, to enable a `ref struct` to be disposed, it must have an accessible `void Dispose()` method.
```
Thraka: I would clarify how `void Dispose()` gets called. Is this something that has to be done by hand? Will the runtime do it for you? How does it interact with using statements? In C# 7 you get an error if you try to use an object that doesn't implement IDisposable: `using (var a = new ObjNotImplementIDisposable)`: did that change with C# 8?

docs/csharp/whats-new/csharp-8.md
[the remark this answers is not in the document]
Thraka: I think it should be `in all constructors`

docs/csharp/whats-new/csharp-8.md
BillWagner: I rewrote this entire section.  I do think I made it much harder to understand than it is.

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,314 @@
...
+};
+```
+
+You can retrieve the last word by prefixing the index `1` with `^`:
+
+```csharp
+Console.WriteLine($"The last word is {words[^1]}");
+```
+
+When you take only a single element and specify from the end, indexing starts at 1.
```
Thraka: I would state this above the code with the instructions on retrieving the last word.

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,314 @@
...
+var brownFox = words[1..4];
+```
+
+Ranges can be specified from the end as well as the beginning. The following code creates a subrange with "lazy" and "dog":
+
+```csharp
+var lazyDog = words[^2..^0];
+```
+
+You can also create ranges that are open ended for the start, end, or both, as in the following code:
```
Thraka: It may be easier to start with this stuff because this is a little more familiar to those used to 0-based indexing. You may want to start with using the `Range` type directly and then say that there is some sort of implicit conversion going on with the range syntax? (if that is true, I just assume it is)

Thraka (changes requested):

Thraka (approved):

Thraka:
I put a bunch of comments, though it happened in the middle of you responding to rons feedback :) I set it to REQUEST CHANGES but changed that to APPROVE as I didn't want to block you. So my feedback may be collapsed and hidden, not sure

docs/csharp/whats-new/csharp-8.md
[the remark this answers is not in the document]
BillWagner: I went with a simpler example for this topic. When I write the "explore" topic, with interactive samples, this is a good one.

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,314 @@
...
+
+- [Nullable reference types](#nullable-reference-types)
+- [Asynchronous streams](#asynchronous-streams)
+- [Ranges and indices](#ranges-and-indices)
+
+The remainder of this article briefly describes these features. Where in-depth articles are available, links to those tutorials and overviews are provided.
+
+## More patterns in more places
+
+**Pattern matching** gives tools to provide shape-dependent functionality across related but different kinds of data. C# 7.0 introduced syntax for type patterns and constant patterns by using the `is` expression and the `switch` statement. These features represented the first tentative steps toward supporting programming paradigms where data and functionality live apart. As the industry moves toward more microservices and other cloud-based architectures, other language tools are needed.
```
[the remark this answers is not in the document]
BillWagner: I don't think perf work has been completed. So I won't address this yet. I think it will be similar perf, but I'll verify as we get closer to release.

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,314 @@
...
+```csharp
+static string Quadrant(Point p) => p switch
+{
+    (0, 0) => "origin",
+    (var x, var y) when x > 0 && y > 0 => "Quadrant 1",
+    (var x, var y) when x < 0 && y > 0 => "Quadrant 2",
+    (var x, var y) when x < 0 && y < 0 => "Quadrant 3",
+    (var x, var y) when x > 0 && y < 0 => "Quadrant 4",
+    (var x, var y) => "on a border",
+    _ => "unknown"
```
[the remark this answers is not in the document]
BillWagner: It's triggered when either X or Y (but not both) is 0.

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,356 @@
...
+
+C# 8.0 expands this vocabulary so you can use more pattern expressions in more places in your code. Consider these features when your data and functionality are separate. Consider pattern matching when your algorithms depend on a fact other than the runtime type of an object. These techniques provide another way to express designs.
+
+In addition to new patterns in new places, C# 8.0 adds **recursive patterns**. The result of any pattern expression is an expression. A recursive pattern is simply a pattern expression applied to the output of another pattern expression.
+
+### switch expressions
+
+Often, a switch statement produces a value in each of its `case` blocks. **Switch expressions** enable you to use more concise expression syntax. There are fewer repetitive `case`, and `break` keywords, and fewer curly braces.  As an example, consider the following enum that lists the colors of the rainbow:
```
rpetrusha:
```suggestion
Often, a switch statement produces a value in each of its `case` blocks. **Switch expressions** enable you to use more concise expression syntax. There are fewer repetitive `case` and `break` keywords, and fewer curly braces.  As an example, consider the following enum that lists the colors of the rainbow:
```

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,356 @@
...
+        Rainbow.Indigo => new RGBColor(0x4B, 0x00, 0x82),
+        Rainbow.Violet => new RGBColor(0x94, 0x00, 0xD3),
+        _              => throw new ArgumentException(message: "invalid enum value", paramName: nameof(colorBand)),
+    };
+```
+
+There are several syntax improvements here:
+
+- The variable comes before the `switch` keyword. The different order makes it visually easy to distinguish the switch expression from the switch statement.
+- The `case` and `:` elements are replaced with `=>`. It's more concise.
```
rpetrusha:
```suggestion
- The `case` and `:` elements are replaced with `=>`. It's more concise and intuitive.
```

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,356 @@
...
+}
+```
+
+In the preceding example, the file is disposed when the closing brace associated with the `using` statement is reached.
+
+In both cases, the compiler generates the call to `Dispose()`. The compiler generates an error if the expression in the using statement is not disposable.
+
+## Static local functions
+
+You can now add the `static` modifier to local functions to ensure that local function doesn't capture (reference) any variables from the enclosing scope. Doing so generates `CS8421`, "A static local function can't contain a reference to <variable>."
```
rpetrusha:
```suggestion
You can now add the `static` modifier to local functions to ensure that a local function doesn't capture (reference) any variables from the enclosing scope. Doing so generates `CS8421`, "A static local function can't contain a reference to <variable>." 
```

rpetrusha (commented):
I've left a few small nits, @BillWagner. You've made some great changes, and I really like the additional examples you've added, and also your clarification of indices and ranges.

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,356 @@
...
+You can specify an index **from the end**. You specify **from the end** using the `^` operator. You are familiar with `array[2]` meaning the element "2 from the start". Now, `array[^2]` means the element "2 from the end". The index `^0` means "the end", or one past that last element.
+
+You can specify a **range** with the **range operator**: `..`. For example, `0..^0` specifies the entire range of the array: 0 from the start up to, but not including 0 from the end. Either operand may use "from the start" or "from the end". Furthermore, either operand may be omitted. The defaults are `0` for the start index, and `^0` for the end index.
+
+Let's look at a few examples. Consider the following array:
+
+```csharp
+var words = new string[]
+{
+    "The",
```
Thraka: would it be useful to put in the element numbers here for index vs ending index?

```csharp
var words = new string[]
{
                // index    ending index
    "The",      // 0        ^9
    "quick",    // 1        ^8
    "brown",    // 2        ^7
    "fox",      // 3        ^6
    "jumped",   // 4        ^5
    "over",     // 5        ^4
    "the",      // 6        ^3
    "lazy",     // 7        ^2
    "dog"       // 8        ^1
};
```

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,356 @@
...
+
+You can try asynchronous streams yourself in our tutorial on [creating and consuming async streams](../tutorials/generate-consume-asynchronous-stream.md).
+
+## Indices and ranges
+
+Ranges and indices provide a succinct syntax for specifying subranges in an array, <xref:System.Span%601>, or <xref:System.ReadOnlySpan%601>.
+
+You can specify an index **from the end**. You specify **from the end** using the `^` operator. You are familiar with `array[2]` meaning the element "2 from the start". Now, `array[^2]` means the element "2 from the end". The index `^0` means "the end", or one past that last element.
+
+You can specify a **range** with the **range operator**: `..`. For example, `0..^0` specifies the entire range of the array: 0 from the start up to, but not including 0 from the end. Either operand may use "from the start" or "from the end". Furthermore, either operand may be omitted. The defaults are `0` for the start index, and `^0` for the end index.
```
Thraka: It might be worth pointing out that the end of a range is exclusive of the index. Much like how random works. The start is inclusive lower bound of the value, but the end is exclusive upper bound of the value. (I'm assuming this is how range works?)

And additionally, it seem the `^` (from end) indicator is also exclusive upper bound-based.

The "2 from start" doesn't make sense to me either. I would make sure to use **element-index** or something to clarify that. When you say `element "2 from the start"` it sounds like you're pulling the element value of 2 out, versus `element index 2 from the start` or `3rd element from the start` would be clearer. Or compare it more directly and indicate the difference with something like:

> `array[2]` means the 3rd element from the start of the array and `array[^2]` means the 2nd element from the end of the array.

And how can you go passed the end of an array with `^0` it seems like that should be invalid? I've not really investigated this stuff so these are my "first look" thoughts :)

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,359 @@
...
+The following code contains a static local function. It can be static because it doesn't access any variables in the enclosing scope:
+
+```csharp
+int M()
+{
+    int y = 5;
+    int x = 7;
+    return Add(x, y);
+
+    int Add(int left, int right) => left + right;
```
pkulikov: @BillWagner is the local function intentionally not `static` in this example, though it can be such? For the sake of example, adding `static` modifier is reasonable.

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,359 @@
...
+}
+```
+
+## Disposable ref structs
+
+A `struct` declared with the `ref` modifier may not implement any interfaces and so cannot implement <xref:System.IDisposable>. Therefore, to enable a `ref struct` to be disposed, it must have an accessible `void Dispose()` method.
```
pkulikov: I guess `readonly ref struct` can also be disposable as a particular case of `ref struct`?

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,359 @@
...
+
+## Nullable reference types
+
+Inside a nullable annotation context, any variable of a reference type is considered to be a **nonnullable reference type**. If you want to indicate that a variable may be null, you must append the type name with the `?` to declare the variable as a **nullable reference type**.
+
+For nonnullable reference types, the compiler uses flow analysis to ensure that local variables are initialized to a non-null value when declared. Fields must be initialized during construction. The compiler generates a warning if the variable is not set in all constructors, or an initializer. Furthermore, nonnullable reference types can't be assigned a value that could be null.
```
pkulikov:
```suggestion
For nonnullable reference types, the compiler uses flow analysis to ensure that local variables are initialized to a non-null value when declared. Fields must be initialized during construction. The compiler generates a warning if the variable is not set in any of available constructors, or an initializer. Furthermore, nonnullable reference types can't be assigned a value that could be null.
```

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,359 @@
...
+
+Nullable reference types aren't checked to ensure they aren't assigned or initialized to null. However, the compiler uses flow analysis to ensure that any variable of a nullable reference type is checked against null before it's accessed or assigned to a nonnullable reference type.
+
+You can learn more about the feature in the overview of [nullable reference types](../nullable-references.md). Try it yourself in a new application in this [nullable reference types tutorial](../tutorials/nullable-reference-types.md). Learn about the steps to migrate an existing codebase to make use of nullable reference types in the [migrating an application to use nullable reference types tutorial]../tutorials/upgrade-to-nullable-references.md).
```
pkulikov: Broken link:
```suggestion
You can learn more about the feature in the overview of [nullable reference types](../nullable-references.md). Try it yourself in a new application in this [nullable reference types tutorial](../tutorials/nullable-reference-types.md). Learn about the steps to migrate an existing codebase to make use of nullable reference types in the [migrating an application to use nullable reference types tutorial](../tutorials/upgrade-to-nullable-references.md).
```

docs/csharp/whats-new/csharp-8.md
```diff
@@ -0,0 +1,359 @@
...
+}
+```
+
+You can try asynchronous streams yourself in our tutorial on [creating and consuming async streams](../tutorials/generate-consume-asynchronous-stream.md).
+
+## Indices and ranges
+
+Ranges and indices provide a succinct syntax for specifying subranges in an array, <xref:System.Span%601>, or <xref:System.ReadOnlySpan%601>.
+
+You can specify an index **from the end**. You specify **from the end** using the `^` operator. You are familiar with `array[2]` meaning the element "2 from the start". Now, `array[^2]` means the element "2 from the end". The index `^0` means "the end", or one past that last element.
```
pkulikov: Would be also useful to specify what index means "the last element". `^1`? 
And "...one past that last element" reads with difficulty: there is no one after the last element. However, I don't know how to rephrase it.

pkulikov:
Oh, @BillWagner I was a little bit late...

BillWagner:
@pkulikov No worries, I'll open a new PR for your feedback.

docs/csharp/whats-new/csharp-8.md
BillWagner: I tried "the index following the last element".

PR #10637, opened 2019-02-21
Add ML.NET 0.10.0 parameter order change warning

JRAlexander:
Add ML.NET 0.10.0 parameter order change warning


luisquintanilla (changes requested):
@JRAlexander Looks good to me. Added a minor change specifying that the error happens at runtime when building the model as opposed to at build time of the app. If you agree with it, it's good to go.

luisquintanilla (approved):

JRAlexander:
Great, Thanks @luisquintanilla!

PR #10647, opened 2019-02-22
Replace some links to docs/standard with relative links

mikkelbu:
Relates to #9968


mairaw (approved):
Thanks @mikkelbu for your changes!

PR #10658, opened 2019-02-23
Fixed closing of code samples

damirarh:
## Summary

Two code samples weren't closed correctly which broke the formatting for a part of the article.


BillWagner (approved):
Thanks for spotting this and submitting the fix @damirarh 

I'll :shipit: now.

You should see the changes on the live site in the next day or so, on our regular publishing cycle.

PR #10668, opened 2019-02-25
Update blog url to dev domain

jpreese:
## Summary

Updates the blog url to point to the newly acquired .dev domain.


BillWagner (approved):
Thanks for making this change @jpreese 

I've verified the destination and I'll :shipit: now.

PR #10672, opened 2019-02-25
Fix code blocks in C# 8 features docs

Renerick:
## Summary

Fixed small indentation problems in C# 8 features documentation, which were causing incorrect text display


BillWagner (approved):
Thanks for spotting this, @Renerick 

I'll :shipit: now.

PR #10702, opened 2019-02-26
chore: Add MSDocs authoring extension suggestion

nschonni:
Adds suggestion when authors open in VS Code to install Authoring Pack


BillWagner (approved):
This is a great addition @nschonni 

Ping @meganbradley Is this something we should publicize for all the docs repos?

PR #10712, opened 2019-02-26
fix: MD035/hr-style

nschonni:
Use the triple dash consitently for HR blocks


mairaw (approved):
Changes look good @nschonni. Thank you!

PR #10713, opened 2019-02-26
fix: MD011/no-reversed-links

nschonni:
Likely a false positive, but replacing code with backticks results in same rendering


mairaw (approved):
Cool! Thanks for changing this @nschonni. Now that we're using the CommonMark spec, we can review the content to replace the pipes inside code blocks. We probably have more throughout the content.

PR #10734, opened 2019-02-26
Clarify when methods should end in Async

kendrahavens:
Using corrected text suggested in [bug on the in-product recommendation](https://github.com/dotnet/roslyn/issues/33082#issuecomment-467536910).

## Summary

"Async" shouldn't always be appended to a method name in cases of an overloaded methods or for methods that start an asynchronous operation, but do not actually return an awaitable type. Updated guidance to reflect these cases.


rpetrusha (approved):
Thanks, @kendrahavens. Your changes look good. The .NET Guide also includes a section on naming conventions in [Task-based asynchronous pattern (TAP)](https://docs.microsoft.com/dotnet/standard/asynchronous-programming-patterns/task-based-asynchronous-pattern-tap). I'll open a PR to modify it and request you review it. I'll also merge your PR now.

PR #10836, opened 2019-03-01
fix: MD014/commands-show-output

nschonni:
Dollar signs used before commands without showing output


mairaw (approved):
I like these changes @nschonni. Thanks!

PR #10841, opened 2019-03-01
C67708: Issue with Bookmark

v-makoud:
Hello @rpetrusha,
Description of the source issue:  Spotted string does not match with bookmark reference, therefore anchor to heading is not rendered. Consider writing heading like this:  Clickonce-1
Many thanks in advance.

## Summary

Describe your changes here.


rpetrusha:
Thanks, @v-makoud. Your change creates a header that has no meaning, though. I've modified your PR to add an anchor tag but leave the header unchanged.

rpetrusha (approved):

PR #10867, opened 2019-03-02
fix: MD019/no-multiple-space-atx

nschonni:
Multiple spaces after hash on atx style heading


BillWagner (approved):
Thanks for spotting and fixing all these @nschonni 

I'll :shipit: now.

PR #10914, opened 2019-03-02
typo: Protecution -> Protection

nschonni:


BillWagner (approved):
LGTM. :shipit:

BillWagner:
@meganbradley Can you look at this change? It failed validation, but it looks like a false positive. I don't see any markdown content before the H1.

nschonni:
@BillWagner I believe it was because it was one of the BOM affected files. Since that was done afterwards, this failed validation. the file in master has the BOM fix, so it shouldn't pop again.

PR #10993, opened 2019-03-03
fix: docs/framework/unmanaged-api/wmi/compareto.md

nschonni:
- pcompare -> pCompare
- instanes -> instances
- Add code fence languages


mairaw (changes requested):
Great changes again @nschonni but the parameters sections lost its soft return formatting for the param names

mairaw (approved):

PR #10999, opened 2019-03-03
fix: docs/framework/unmanaged-api/wmi/getmethod.md

nschonni:
- paramteers -> parameters
- Insignature -> InSignature
- Add code fence language


mairaw (changes requested):
Changes look great but this one also has the issue with the parameters section

mairaw (approved):

PR #11004, opened 2019-03-03
fix: docs/framework/unmanaged-api/wmi/next.md

nschonni:
- retireves -> retrieves
- betweeen -> between
- Add code fence language


mairaw (changes requested):
Again - looks great but same issue with the parameter formatting

mairaw (approved):

PR #11009, opened 2019-03-03
typo: GLAG -> FLAG

nschonni:
Assuming this isn't a product bug


docs/framework/unmanaged-api/wmi/qualifierset-put.md
```diff
@@ -2,47 +2,49 @@
...
    [in] LONG                 lFlavor
-); 
-```  
+);
+```
 
 ## Parameters
 
-`vFunc`   
+`vFunc`
```
mairaw: The two spaces at the end create a soft return in these cases and removing them causes that to go away. 
image

Perhaps escape like the following?
```suggestion
`vFunc`\
```

mairaw (changes requested):
Good catch on the typo @nschonni! left a comment re. the style change though.

mairaw (approved):
I'll fix my suggestion in PR #11012 instead. Thanks!

PR #11015, opened 2019-03-04
What's new in C# 8.0: fix minor inaccuracies

pkulikov:
Two changes:
- use the same tense in the list
- corrected the description of the code output; according to the `for` loop in the code, values from 0 to 19 are expected, not from 1 to 20.


rpetrusha (approved):
Thanks, @pkulikov, for making these corrections. I'll merge your PR now.

PR #11040, opened 2019-03-05
typo: Extensibliity -> Extensibility

nschonni:
- Added redirect record
- Removed trailing commas from redirect.json


mairaw (approved):
Good catch with those commas. Weird that it didn't cause any build issues. Thanks for fixing these.

PR #11047, opened 2019-03-05
typo: Visisbility -> Visibility

nschonni:


mairaw (changes requested):
Thanks @nschonni. Can you just apply a suggestion I left to escape a * char?

mairaw (approved):
Thank you!

PR #11053, opened 2019-03-05
[winforms] Refactoring redundant path

Potapy4:
Contributes to #10653

cc @mairaw


rpetrusha (approved):
Thanks, @Potapy4, for refactoring these redundant file paths. I'll merge your PR now.

For future large PRs, it would be helpful if you could limit the number of files changed to about 100. That makes it significantly easier for us to review them.

PR #11076, opened 2019-03-06
typo: Progess -> Progress

nschonni:
- Make H4s H3s to not skip level


mairaw (changes requested):
Changes look good. Found a few more still skipping.

mairaw (approved):
LGTM

PR #11137, opened 2019-03-07
typo: Withtype -> WithType

nschonni:


rpetrusha (approved):
Thanks for correcting the typo. I also noticed an incomplete sentence in the article, so I corrected it. We'll merge when the build completes successfully.

PR #11141, opened 2019-03-07
Tiny bug in .gitignore, should be just .vscode/

mvelosop:
## Summary

Change .gitignore from `.vscode/*` to `.vscode/` b/c some local vscode settings where not being ignored


mairaw (approved):
Thanks @mvelosop!

PR #11198, opened 2019-03-08
fix: delagate -> delegate

nschonni:
This might be a product rather than doc bug


BillWagner (approved):
LGTM as well.
:shipit:

PR #11239, opened 2019-03-09
Follow naming conventions for public members

miroslavpopovic:
Fix camel cased fromRainbowClassic method name to pascal cased FromRainbowClassic


BillWagner (approved):
Thanks for spotting this and updating the sample @miroslavpopovic 

We appreciate it. 

Thanks again.

PR #11259, opened 2019-03-11
Mention that ionide generates net461, not dotnet core

ChrisPritchard:
@cartermp This has caused confusion with a number of starters, who install core then follow these instructions only to get a host of exceptions when they try to run it.

I'd almost suggest the command line tools be given higher priority, as there is a lot to say for telling newcomers to: `dotnet new console -lang F#`, then `dotnet run` and their finished :)

## Summary

Describe your changes here.


cartermp (approved):
Thanks @ChrisPritchard  - I'll probably not mention the FORGE templates and just use dotnet tooling in the article.

PR #11277, opened 2019-03-12
Fix invalid alert box

WiredUK:
`!INFORMATION` isn't a valid alert box, use `!TIP` instead


mairaw (changes requested):
Thank you @WiredUK for fixing that. I also noticed there's a second one on line 17. Would you like to fix that as well?

WiredUK:
> Thank you @WiredUK for fixing that. I also noticed there's a second one on line 17. Would you like to fix that as well?

Oops, good spot! I swear it wasn't there before :)
And I also swear I've contributed to docs previously, maybe that was EF docs or something...

mairaw (approved):
Thank you for making the additional change as well @WiredUK.

That is probably just your first contribution to our repo, not docs.microsoft.com in general. I believe you 😉

PR #11290, opened 2019-03-13
Create pattern tutorial

### LINKED ISSUE #8205: Improve pattern matching tutorial (C# 7 feature)
The existing tutorials are correct, but don't provide a great motivator for the feature.

Instead, the tutorial should be one that helps readers understand the kinds of problems that are solved with Pattern Matching in C#.

BillWagner:
Fixes #8205 

Our customers have found challenges using pattern matching effectively. This tutorial provides a stronger motivation for the scenarios where pattern matching can be most effectively used.

/cc @MadsTorgersen @KathleenDollard


BillWagner:
closing and opening for a new build.

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,325 @@
...
+In this tutorial, you'll learn how to:
+
+> [!div class="checklist"]
+> * How to recognize situations where pattern matching should be used.
+> * How to use pattern matching expressions to implement behavior based on types and property values.
+> * How to combine pattern matching with other techniques to create complete algorithms.
+
+## Prerequisites
+
+You'll need to set up your machine to run .NET Core, including the C# 8.0 preview compiler. The C# 8 preview compiler is available with [Visual Studio 2019 preview 4](https://visualstudio.microsoft.com/vs/preview/?utm_medium=microsoft&utm_source=docs.microsoft.com&utm_campaign=inline+link&utm_content=download+vs2019+preview), or [.NET Core 3.0 preview 3](https://dotnet.microsoft.com/download/dotnet-core/3.0).
```
Thraka: I would change the **preview 4** and **preview 3** to **latest preview**. Using **preview #** will become out of date very fast as the next preview or RC gets released.

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,325 @@
...
+
+Consider a major metro area that is using tolls and peak time pricing to manage traffic. You write an application that calculates tolls for a vehicle based on its type. Later enhancements incorporate pricing based on the number of occupants in the vehicle. Further enhancements add pricing based on the time and the day of the week.
+
+From that brief description, you may have quickly sketched out an object hierarchy to model this system. If you're building the entire system, and these were the only requirements, that would be a good solution. That's not the situation. This metro area had already been using software to manage vehicle registrations for years. They bought different systems from different suppliers for personal cars, taxis, buses, and delivery trucks. Different systems all provide classes to represent the data they export. But these different systems don't provide a single object model for you to use. In this tutorial, you'll use these simplified classes for the vehicle classes from these external systems, as shown in the following code:
```
Thraka: I had to read this a few times just to make sure I understood what you were trying to say. I think this paragraph could be simplified. there is a lot of text to really just state what is stated a few paragraphs below: 

>You can see that the vehicle classes are from different systems, and are in different namespaces. No common base class, other than `System.Object` can be leveraged.

Also, it is said above this paragraph:

>In this tutorial, you'll create and explore an application that takes incoming data from several external sources for a single scenario. 

So I think that this paragraph is just explaining too much. How about the following or something like it?

```suggestion
From that brief description, you may have quickly sketched out an object hierarchy to model this system. However, your data is coming from multiple sources like other vehicle registration management systems. These systems provide different classes to model that data and you don't have a single object model you can use. In this tutorial, you'll use these simplified classes to model for the vehicle data from these external systems, as shown in the following code:
```

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,325 @@
...
+You can download the starter code from the [dotnet/samples](https://github.com/dotnet/samples/tree/master/csharp/tutorials/patterns/start) GitHub repository. You can see that the vehicle classes are from different systems, and are in different namespaces. No common base class, other than `System.Object` can be leveraged.
+
+## Pattern matching designs
+
+The scenario used in this tutorial highlights the kinds of problems that are well suited to use pattern matching to solve: 
+
+- The objects you need to work with aren't in an object hierarchy that matches your goals. You may be working with classes that are part of unrelated systems.
+- The functionality you're adding isn't part of the core abstraction for these classes. The toll paid by a vehicle *changes* for different types of vehicles, but the toll isn't a core function of the vehicle.
+
+The pattern matching features in C# make it easier to build software when the *shape* of the data and the *operations* on that data are not described together.
```
Thraka: This makes it sound like pattern matching is **only** useful "when the *shape* of the data..." I suspect this is because of the word **when**. Maybe this should be reversed?

```suggestion
When the *shape* of the data and the *operations* on that data are not described together, the pattern matching features in C# make it easier to work with. 
```

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,325 @@
...
+## Implement the basic toll calculations
+
+The most basic toll calculation relies only on the vehicle type:
+
+- A `Car` is $2.00.
+- A `Taxi` is $3.50.
+- A `Bus` is $5.00.
+- A `DeliveryTruck` is $10.00
+
+Create a new `TollCalculator` class, and implement the You can implement that using this version of a `TollCalculator` class:
```
Thraka:
```suggestion
Create a new `TollCalculator` class, and implement pattern matching on the vehicle type to get the toll amount.
```

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,325 @@
...
+            Bus b => 5.00m,
+            DeliveryTruck t => 10.00m,
+            { } => throw new ArgumentException(message: "Not a known vehicle type", paramName: nameof(vehicle)),
+            null => throw new ArgumentNullException(nameof(vehicle))
+        };
+    }
+}
+```
+
+The preceding code uses a **switch expression** that tests the **type pattern**. A **switch expression** begins with the variable, `vehicle` in the preceding code, followed by the `switch` keyword. Next comes all the switch arms inside curly braces. The `switch` expression makes other refinements to the ceremony that surrounded the `switch` statement. The `case` keyword is omitted, and the result of each arm is an expression. The last two arms show a new language feature. The `{ }` case matches any non-null object that didn't match an earlier arm. This arm catches any incorrect types passed to this method. Finally, the `null` pattern catches when `null` is passed to this method. The `null` pattern can be last because the other type patterns match only a non-null object of the correct type.
```
Thraka: I think it's OK to point out this is not the same as a switch **statement**. If someone is new to this, they may not realize they are two different things. 

"A **switch expression** (not the same as a switch **statement**) begins with the variable..." 

Maybe a link in this first use of **switch expression** and **switch statement** to the keyword docs?

OOOOOR maybe I'm totally wrong in that thought assumption and they really are the same just expressed differently? With the rule difference of not having `case` or `default` it makes it feel like it's a different concept.

----

Also, the use of the word **ceremony** feels odd... "syntax" ??

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,325 @@
...
+- Cars and taxis with no passengers pay an extra $0.50.
+- Cars and taxis with two passengers get a 0.50 discount.
+- Cars and taxis with three or more passengers get a $1.00 discount.
+- Buses that are less than 50% full pay an extra $2.00.
+- Buses that are more than 90% full get a $1.00 discount.
+
+These rules can be implemented using the **property pattern** in the same switch expression. The property pattern examines properties of the object once the type has been determined.  The single case for a `Car` expands to four different cases:
+
+```csharp
+Car { Passengers: 0} => 2.00m + 0.50m,
```
Thraka: I would put the whole switch statement in these code blocks (at least defining switch stuff) since this is teaching people how to use it, it would help them to see the context of these examples at all times.

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,325 @@
...
+```
+
+## Recursive patterns
+
+You can make this code less repetitive by using **recursive patterns**. The `Car` and `Taxi` both have four different arms in the preceding examples. In both cases, you can create a type pattern that feeds into a property pattern. This technique is shown in the following code:
+
+```csharp
+public decimal CalculateToll(object vehicle) =>
+    vehicle switch
+{
```
Thraka: Indentation doesn't line up on this and the next few code examples

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,325 @@
...
+
+```csharp
+DeliveryTruck t when (t.GrossWeightClass > 5000) => 10.00m + 5.00m,
+DeliveryTruck t when (t.GrossWeightClass < 3000) => 10.00m - 2.00m,
+DeliveryTruck t => 10.00m,
+```
+
+## Recursive patterns
+
+You can make this code less repetitive by using **recursive patterns**. The `Car` and `Taxi` both have four different arms in the preceding examples. In both cases, you can create a type pattern that feeds into a property pattern. This technique is shown in the following code:
```
Thraka: nested switch expressions?

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,325 @@
...
+    DayOfWeek.Thursday => true,
+    DayOfWeek.Friday => true,
+    DayOfWeek.Saturday => false,
+    DayOfWeek.Sunday => false
+};
+```
+
+That method works, but it's repetitious. You can simplify it, as shown in the following code:
+
+[!code-csharp[IsWeekDay](../../../samples/csharp/tutorials/patterns/finished/toll-calculator/TollCalculator.cs#IsWeekDay)]
```
Thraka: Code is blank

Thraka (changes requested):
This is an AMAZING tutorial. I really love it.

docs/csharp/tutorials/pattern-matching.md
BillWagner: fixed in dotnet/samples#715

docs/csharp/tutorials/pattern-matching.md
BillWagner: All the docs from the product team refer to this as "recursive patterns".

We should be consistent.

docs/csharp/tutorials/pattern-matching.md
BillWagner: I added context on each of the snippets, then included the full method at the end of this section.

docs/csharp/tutorials/pattern-matching.md
BillWagner: I did add the clarifying remarks.

We haven't written the reference docs for the switch expression yet, so I only added the link to the switch statement. 😭

docs/csharp/tutorials/pattern-matching.md
BillWagner: Some of this is also fixed in dotnet/samples#715

Thraka (approved):

BillWagner:
closing and reopening to force a new build

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,386 @@
...
+    Car { Passengers: 0} => 2.00m + 0.50m,
+    Car { Passengers: 1 } => 2.0m,
+    Car { Passengers: 2} => 2.0m - 0.50m,
+    Car c when c.Passengers > 2 => 2.00m - 1.0m,
+
+    // ...
+};
+```
+
+The first three cases test the type as a `Car`, then check the value of the `Passengers` property. If both match, that expression is evaluated and returned. The final clause shows the `when` clause for a property pattern. You use the `when` clause to test conditions other than equality on a property. In the preceding example, the `when` clause tests to see that there are more than 2 passengers in the car. Strictly speaking, it's not necessary in this example.
```
KathleenDollard: Last sentence: 

The when clause illustrates the technique, but it's not necessary in this case because all other expected cases are accounted for.

(Its not just not needed, I think it doesn't help the clarity).

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,386 @@
...
+using LiveryRegistration;
+
+namespace toll_calculator
+{
+    public class TollCalculator
+    {
+        public decimal CalculateToll(object vehicle) =>
+            vehicle switch
+        {
+            Car c => 2.00m,
```
KathleenDollard: Thank you, thank you, thank you for just quietly using decimal.

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,386 @@
...
+        },
+    
+        Bus b when ((double)b.Riders / (double)b.Capacity) < 0.50 => 5.00m + 2.00m,
+        Bus b when ((double)b.Riders / (double)b.Capacity) > 0.90 => 5.00m - 1.00m, 
+        Bus b => 5.00m,
+    
+        DeliveryTruck t when (t.GrossWeightClass > 5000) => 10.00m + 5.00m,
+        DeliveryTruck t when (t.GrossWeightClass < 3000) => 10.00m - 2.00m,
+        DeliveryTruck t => 10.00m,
+        { } => throw new ArgumentException(message: "Not a known vehicle type", paramName: nameof(vehicle)),
```
KathleenDollard: nit: For consistency, add a blank line above this

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,386 @@
...
+        (true, TimeBand.Overnight, _) => 0.75m,
+        (false, _, _) => 1.00m,
+    };
+```
+
+Finally, you can remove the two rush hour times that pay the regular price. Once you remove those arms, you can replace the `false` with a discard (`_`) in the final switch arm. You'll have the following finished method:
+
+[!code-csharp[SimplifiedTuplePattern](../../../samples/csharp/tutorials/patterns/finished/toll-calculator/TollCalculator.cs#FinalTuplePattern)]
+
+This example highlights one of the advantages of pattern matching. The pattern branches are evaluated in order. If you rearrange them so that an earlier branch handles one of your later cases, the compiler warns you. Those language rules made it easier to do the preceding simplifications with confidence that the code didn't change.
```
KathleenDollard: There are many ways to screw up where the compiler won't warn you. I think that things are evaluated in order should be early in the article. Possibly mention that unreachable code is still a thing for switch expressions. But overall this paragraph doesn't have the same weight as the rest of the article, imo.

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,386 @@
...
+
+Pattern matching provides a natural syntax to implement different solutions than you'd create if you used object-oriented techniques. The cloud is causing data and functionality to live apart. The *shape* of the data and the *operations* on it aren't necessarily described together. In this tutorial, you consumed existing data in entirely different ways from its original function. Pattern matching gave you the ability to write functionality that over those types, even though you couldn't extend them.
```
KathleenDollard: First sentence here could be more clear. Maybe something like:

Pattern matching makes some types of code more readable and offers an alternative to object-oriented techniques when you can't add code to your classes.

KathleenDollard (commented):
Absolutely superb.

And I have a new pattern matching demo domain to "borrow". Excellent choice of domain.

docs/csharp/tutorials/pattern-matching.md
```diff
@@ -0,0 +1,386 @@
...
+            Console.WriteLine($"The toll for a bus is {tollCalc.CalculateToll(bus)}");
+            Console.WriteLine($"The toll for a truck is {tollCalc.CalculateToll(truck)}");
+
+            try
+            {
+                tollCalc.CalculateToll("this will fail");
+            }
+            catch (ArgumentException e)
+            {
+                Console.WriteLine("Caught an argument exception when using the wrong type", DayOfWeek.Friday);
```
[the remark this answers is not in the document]
BillWagner: I have no idea. Fixed

docs/csharp/tutorials/pattern-matching.md
BillWagner: Fixed, and moved the discussion of `when` to the bus switch arms, where it is used in a real condition.

docs/csharp/tutorials/pattern-matching.md
BillWagner: I did introduce the ordering concept earlier. Other than that, I left this unchanged

PR #11300, opened 2019-03-14
Fix a typo

laurentkempe:
## Summary

Add missing "to"


BillWagner (approved):
Thanks for making this change @laurentkempe 
We appreciate it.

I'll :shipit: it now. You should see the changes on the live site in the next day or two, on our regular publishing cycle.

PR #11310, opened 2019-03-14
Moved resources documentation

### LINKED ISSUE #8882: Move Resources documentation to .NET Guide

### LINKED ISSUE #10031: EmbeddedResource logical name is not set properly for .NET Core desktop applications
@diverdan92 commented on [Thu Dec 20 2018](https://github.com/dotnet/corefx/issues/34190)

When porting existing WPF or WinForms applications to target .NET Core that contain EmbeddedResources, the logical name is not fully defined and results in missing resources. 

@ericstj @dsplaisted fyi from our debugging session last week. 

---

@danmosemsft commented on [Thu Dec 20 2018](https://github.com/dotnet/corefx/issues/34190#issuecomment-449136579)

In what cases -- do you have a repro case? There is a bit convoluted mechanism for picking these names.

This is all done in the build though -- any reason I shouldn't move this to MSBuild?

---

@nguerrera commented on [Thu Dec 20 2018](https://github.com/dotnet/corefx/issues/34190#issuecomment-449159316)

RootNamespace is defaulted to project name in sdk. If you had an empty one or a different one before the port, then you need to keep it in your csproj.

---

@nguerrera commented on [Thu Dec 20 2018](https://github.com/dotnet/corefx/issues/34190#issuecomment-449159658)

Please supply a repro showing expected and actual behavior. If it's the root namespace default as I suspect, I don't think we can do anything about it. 

---

@ericstj commented on [Thu Dec 20 2018](https://github.com/dotnet/corefx/issues/34190#issuecomment-449171593)

I think this bug belongs in the SDK or MSBuild repo, it's all about the build tooling for this stuff.  I think the gist here is the globbing and how that interacts with folks when they are porting their stuff.  If they manually organized a project into folders without using those folders to represent a namespace then the generated .resources will have the wrong name.

---

@nguerrera commented on [Thu Dec 20 2018](https://github.com/dotnet/corefx/issues/34190#issuecomment-449181313)

If that's what the issue is, then I think this should go to docs, not sdk or msbuild. 

We can't say that it "is not set properly", or that it is "not fully defined". I will grant that it is certainly a pain point in migration.

The algorithm for picking the names is the same as always. If you are using default RootNamespace + default globs, then you get the corresponding names from those defaults. If you go from File -> New classic project, add a folder, add a resource to it,  then do the same for an sdk project, you will get the same names in both projects: (RootNamespace=ProjectNameByDefault).(FolderName).(Filename).resources.

If you were not using the folder names in your resource names before, then you had one of LogicalName, Link, or MainfestResourceName metadata (possibly others) to override that behavior. In the new world, you can still do that with item updates or by disabling globs and copying over your old explicit includes wholesale. We cannot guess these overrides, you have to carry them over.

---

@danmosemsft commented on [Thu Dec 20 2018](https://github.com/dotnet/corefx/issues/34190#issuecomment-449219500)

Are there any exception messages commonly seen in working through these difficulties, that we can improve? That at least is within BCL control. Eg., "did not find stream named X, only streams named Y and Z" might be pretty helpful?

---

@nguerrera commented on [Thu Dec 20 2018](https://github.com/dotnet/corefx/issues/34190#issuecomment-449220537)

Yes, we don't have the Y and Z part, but it would be helpful. A link to a document describing how names are assigned and how to control them would be a nice touch as well.

---

@nguerrera commented on [Thu Dec 20 2018](https://github.com/dotnet/corefx/issues/34190#issuecomment-449228714)

I _think_ this is the common failure case: https://github.com/dotnet/coreclr/blob/791332b101524034ca9a69ab7669d5a87713447b/src/System.Private.CoreLib/src/System/Resources/ManifestBasedResourceGroveler.cs#L512

---

@tarekgh commented on [Thu Jan 17 2019](https://github.com/dotnet/corefx/issues/34190#issuecomment-455310378)

Will move this to the doc repo.

rpetrusha:
## Moved resources documentation

This PR:

- Moves the resources documentation from the .NET Framework Guide to the .NET Guide.
- Redirects requests for the .NET Framework Guide articles to the .NET Guide articles.
- Removes the private TOC for .NET resources and incorporates it into the master TOC.
- Renames some of the resources files (removes references to the desktop from the filename).

Fixes #8882


Thraka (approved):
I don't see any problems. However, you have merge conflicts :)

Also, with the conversation we had today about accessibility, should you remove the `index.md` and change it to overview? Also, I think that you should just promote the entry `#### Work with .resx files programmatically` to `###` instead of nesting it.

rpetrusha:
Closing. I'll reopen as a new PR once the TOC is more stable.

PR #11352, opened 2019-03-17
Fixed variable name in printString code snippet

roketworks:
## Summary

The `printString` code snippet function has a parameter name of `str` but the snippet used the non-existent variable `s`. Changed `s` to match parameter name `str`.


cartermp (approved):
Thanks @roketworks!

Issue #11396, opened 2019-03-19
Reorganize interop articles


nxtn:
## Status

Several problems with interop topics in .NET docs.

- Scattered and repetitive. There are identical/similar contents in [.NET Guide](https://github.com/dotnet/docs/tree/master/docs/standard/native-interop) and [.NET Framework Guide](https://github.com/dotnet/docs/tree/master/docs/framework/interop).
- Incomplete. For more information, see #4649.

## Rationale

- P/Invoke is still widely adopted.
- CoreCLR has enabled us to embed .NET Core into native code.
- The behaviors apply to both .NET Framework and .NET Core.
- The behaviors apply to both P/Invoke method and `calli` instruction.
- The behaviors apply to both calling from managed code to native code and the reverse.

## Proposal

- Migrate archived interop topics back to docs. #4649
- Integrate and reorganize topics in [.NET Guide](https://github.com/dotnet/docs/tree/master/docs/standard/native-interop) and [.NET Framework Guide](https://github.com/dotnet/docs/tree/master/docs/framework/interop). dotnet/dotnet-api-docs#1199 and #6629.
- New topics about calling from native code to managed code. [.NET Core hosting](https://github.com/dotnet/docs/blob/master/docs/core/tutorials/netcore-hosting.md)
- Keep consistent with [C++ docs](https://github.com/MicrosoftDocs/cpp-docs/blob/master/docs/dotnet/native-and-dotnet-interoperability.md) if possible.

rpetrusha:
@NextTurn, thanks for opening this issue. I also added a link to #6629 to the **Proposal** section. We'd welcome your contribution in improving the interop documentation, and your proposal looks really good. Before actually migrating archived content, it's a good idea to agree on the topics that are to be migrated.

NextTurn:
Change mentions of Unicode to UTF-16. https://github.com/dotnet/dotnet-api-docs/pull/4238#discussion_r443747196

PR #11414, opened 2019-03-20
Fix platform invoke data types

nxtn:
Contributes to #11396

See the rich diff for visible changes.


rpetrusha (approved):
Thank you, @NextTurn, for beginning revision of the interop documentation. Your changes look good, except that the date in the ms.date metadata field should also have been changed. I've made that change, so I'll merge after the internal build completes.

Issue #10089, opened 2019-01-22
Missed UnmanagedType.LPUTF8Str


qaqz111:
The document missed `UnmanagedType.LPUTF8Str` in all cases. And, what is the default mashaling in platform invoke?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 25d350b8-a514-6c13-a091-0952a8fc0610
* Version Independent ID: 3ee9bbfd-2060-d97f-c2f1-12867899034a
* Content: [Default Marshaling for Strings](https://docs.microsoft.com/en-us/dotnet/framework/interop/default-marshaling-for-strings)
* Content Source: [docs/framework/interop/default-marshaling-for-strings.md](https://github.com/dotnet/docs/blob/master/docs/framework/interop/default-marshaling-for-strings.md)
* Product: **dotnet-framework**
* GitHub Login: @rpetrusha
* Microsoft Alias: **ronpet**

Issue #9894, opened 2019-01-08
Make default values more obvious


Const-me:
The article has a title "Default Marshaling for Strings".

In the content, it lists options, but fails to say which one is the default, for all cases except COM interfaces. What's the default for Platform Invoke? What's the default for structures? Does the default depend on OS? No answers...

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 25d350b8-a514-6c13-a091-0952a8fc0610
* Version Independent ID: 3ee9bbfd-2060-d97f-c2f1-12867899034a
* Content: [Default Marshaling for Strings](https://docs.microsoft.com/en-us/dotnet/framework/interop/default-marshaling-for-strings#cpcondefaultmarshalingforstringsanchor5)
* Content Source: [docs/framework/interop/default-marshaling-for-strings.md](https://github.com/dotnet/docs/blob/master/docs/framework/interop/default-marshaling-for-strings.md)
* Product: **dotnet-framework**
* GitHub Login: @rpetrusha
* Microsoft Alias: **ronpet**

NextTurn:
@rpetrusha 

What about *Customizing string marshaling* (like [Customizing parameter marshaling](https://docs.microsoft.com/dotnet/standard/native-interop/customize-parameter-marshalling))? These articles not just focus on the default behaviors.

Const-me:
@NextTurn Thanks, at least title is more relevant now.

But I would really like to have the default marshalling documented somewhere. Don’t care in which article, but the exact behavior is quite important.

From experiments, I know the defaults are different based on everything: OS and platform (tested on windows .NET 4.7+ and ARM Linux .NET core 2.0+), and also the location (dll/so imported functions, IUnknown interfaces, structures, unmanaged function pointers).

These experiments are time consuming and not particularly reliable. E.g. impossible to distinguish BSTR from LPCWSTR, they point to the same data the only difference is memory management.

While it’s possible to always specify the format manually, for larger API surface, it makes the C# part of the interop harder to read and support.

NextTurn:
Well... This article is in .NET Framework Guide so it only applies to .NET Framework on Windows :(

Const-me:
@NextTurn The defaults are different in every case, and undocumented, even on Windows.

NextTurn:
Do you want to take a look @jkoritzinsky ?

jkoritzinsky:
[...]

PR #11415, opened 2019-03-20
Fix default marshaling for strings

nxtn:
Contributes to #11396, #9894, #10089

Indicate the default marshaling for strings used in platform invoke.


docs/framework/interop/default-marshaling-for-strings.md
```diff
@@ -1,264 +1,267 @@
...
+    [MarshalAs(UnmanagedType.LPTStr)] public string f1;
+    [MarshalAs(UnmanagedType.ByValTStr, SizeConst = 256)] public string f2;
+}
+```
+
+## Fixed-Length String Buffers
+
+In some circumstances, a fixed-length character buffer must be passed into unmanaged code to be manipulated. Simply passing a string does not work in this case because the callee cannot modify the contents of the passed buffer. Even if the string is passed by reference, there is no way to initialize the buffer to a given size.
+
+The solution is to pass a <xref:System.Text.StringBuilder> buffer as the argument instead of a string. A `StringBuilder` can be dereferenced and modified by the callee, provided it does not exceed the capacity of the `StringBuilder`. It can also be initialized to a fixed length. For example, if you initialize a `StringBuilder` buffer to a capacity of `N`, the marshaler provides a buffer of size (`N`+1) characters. The +1 accounts for the fact that the unmanaged string has a null terminator while `StringBuilder` does not.
```
rpetrusha:
```suggestion
The solution is to pass a <xref:System.Text.StringBuilder> buffer as the argument instead of a <xref:System.String>. A `StringBuilder` can be dereferenced and modified by the callee, provided it does not exceed the capacity of the `StringBuilder`. It can also be initialized to a fixed length. For example, if you initialize a `StringBuilder` buffer to a capacity of `N`, the marshaler provides a buffer of size (`N`+1) characters. The +1 accounts for the fact that the unmanaged string has a null terminator while `StringBuilder` does not.
```

docs/framework/interop/default-marshaling-for-strings.md
```diff
@@ -1,264 +1,267 @@
...
-Public Declare Auto Sub PassBStr Lib "StringLib.Dll" _  
-(<MarshalAs(UnmanagedType.BStr)> s As String)      
-Public Declare Auto Sub PassAnsiBStr Lib "StringLib.Dll" _  
-(<MarshalAs(UnmanagedType.AnsiBStr)> s As String)      
-Public Declare Auto Sub PassTBStr Lib "StringLib.Dll" _  
-(<MarshalAs(UnmanagedType.TBStr)> s As String)  
-End Class  
+A `StringBuilder` can be dereferenced and modified by the callee, provided it does not exceed the capacity of the `StringBuilder`. The following code example demonstrates how `StringBuilder` can be initialized to a fixed length.
+
+```vb
```
rpetrusha: - There's no good reason here and in the C# example to define GetWindowText in a separate class. That just adds overhead to the example.

docs/framework/interop/default-marshaling-for-strings.md
```diff
@@ -1,264 +1,267 @@
 ---
 title: "Default Marshaling for Strings"
 ms.date: "03/30/2017"
```
rpetrusha:
```suggestion
ms.date: "03/20/2019"
```

rpetrusha (changes requested):
Thanks for revising this article, @NextTurn. I've left a number of comments and suggestions, many unrelated to your PR.

I was considering whether the C# and VB code should be moved to the samples repo, but since it's not actually executable as meaningful examples, I thought that the examples could be left inline. Do you agree?

docs/framework/interop/default-marshaling-for-strings.md
```diff
@@ -1,264 +1,267 @@
...
+
+The solution is to pass a <xref:System.Text.StringBuilder> buffer as the argument instead of a string. A `StringBuilder` can be dereferenced and modified by the callee, provided it does not exceed the capacity of the `StringBuilder`. It can also be initialized to a fixed length. For example, if you initialize a `StringBuilder` buffer to a capacity of `N`, the marshaler provides a buffer of size (`N`+1) characters. The +1 accounts for the fact that the unmanaged string has a null terminator while `StringBuilder` does not.
+
+For example, the Microsoft Win32 API `GetWindowText` function (defined in Windows.h) is a fixed-length character buffer that must be passed into unmanaged code to be manipulated. `LpString` points to a caller-allocated buffer of size `nMaxCount`. The caller is expected to allocate the buffer and set the `nMaxCount` argument to the size of the allocated buffer. The following code shows the `GetWindowText` function declaration as defined in Windows.h.
```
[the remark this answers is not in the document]
NextTurn: `GetWindowText` is not a function. But P/Invoke will probe for function names following the A/W suffix on Windows. Is this covered in docs?

NextTurn:
I've no idea whether the snippets should be moved to *samples* repo, but actually there are a number of even smaller files.

https://github.com/dotnet/samples/blob/master/snippets/csharp/concepts/linq/query-expression-basics_1.cs

docs/framework/interop/default-marshaling-for-strings.md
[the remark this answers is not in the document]
NextTurn: Moreover, this function is unavailable in console apps. Could you think of a better candidate?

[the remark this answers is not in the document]
rpetrusha: I linked to GetWindowTextW, since that's most likely the function to be called. Probing for function names, and even the use of an ASCII and a wide-character function, used to be covered in the VB 6 interop docs, but I don't think that it's covered in our doc set.

Finally, I can't think of a better candidate for GetWindowText. The problem with most code in console apps that I've seen, as I recall, is that they retrieve a handle to the console window based on its text, then call GetWindowText to get the text, most of which was already used to retrieve the window handle.

rpetrusha:
Although we have a lot of non-executable code in the sample repo, the basic goal is to have all code in the samples repro compilable and (optionally) executable. The LINQ fragment falls well short of that.

Given that PInvoke calls often give developers difficulty, it's very useful to have some immediately compilable and executable examples. If you'd like to move the examples to the Samples repo and make them executable, we'd certainly welcome the PR. If you'd prefer to leave them inline, that's fine too, @NextTurn.

docs/framework/interop/default-marshaling-for-strings.md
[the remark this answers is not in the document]
NextTurn: P/Invoke probing is mentioned in [DllImportAttribute.CharSet Field](https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.dllimportattribute.charset#remarks) and covered in .NET Framework [Specifying a Character Set](https://docs.microsoft.com/dotnet/framework/interop/specifying-a-character-set) article. It's another issue whether we should port them to .NET Guide [Charsets and marshalling](https://docs.microsoft.com/dotnet/standard/native-interop/charset) article in a "Windows Specific" section, or add a See also reference.

docs/framework/interop/default-marshaling-for-strings.md
NextTurn: The compiler asked to [Move P/Invokes to NativeMethods class](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1060-move-p-invokes-to-nativemethods-class) before, but I don't know when this warning disappeared.

NextTurn:
Got that. Compilable examples would be in separate PRs.

docs/framework/interop/default-marshaling-for-strings.md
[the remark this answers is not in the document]
rpetrusha: Thanks for finding information on probing. We should consider moving all of interop to the .NET Guide. That can be done later, once the revisions and migrations have been completed.

docs/framework/interop/default-marshaling-for-strings.md
rpetrusha: I didn't realize that. Thanks for pointing it out. That suggests that perhaps we should rename the class names to NativeMethods. If not, we should rename Win32API to WindowsAPI.

docs/framework/interop/default-marshaling-for-strings.md
```diff
@@ -1,264 +1,281 @@
...
+    HRESULT PassString2([in] BSTR s);
+    HRESULT PassString3([in] LPStr s);
+    HRESULT PassString4([in] LPWStr s);
+    HRESULT PassStringRef1([in, out] BSTR *s);
+    HRESULT PassStringRef2([in, out] BSTR *s);
+    HRESULT PassStringRef3([in, out] LPStr *s);
+    HRESULT PassStringRef4([in, out] LPWStr *s);
+};
 ```
-[…]
```
NextTurn: Is this useful?

rpetrusha: No, I don't think it's useful, @NextTurn. I have no idea of why it's there.

docs/framework/interop/default-marshaling-for-strings.md
```diff
@@ -1,264 +1,280 @@
...
+
+Platform invoke copies string arguments, converting from the .NET Framework format (Unicode) to the platform unmanaged format. Strings are immutable and are not copied back from unmanaged memory to managed memory when the call returns.
+
+The following table lists the marshaling options for strings when marshaled as a method argument of a platform invoke call. The <xref:System.Runtime.InteropServices.MarshalAsAttribute> attribute provides several <xref:System.Runtime.InteropServices.UnmanagedType> enumeration values to marshal strings.
+
+|Enumeration type|Description of unmanaged format|
+|----------------------|-------------------------------------|
+|`UnmanagedType.AnsiBStr`|A COM-style `BSTR` with a prefixed length and ANSI characters.|
+|`UnmanagedType.BStr`|A COM-style `BSTR` with a prefixed length and Unicode characters.|
+|`UnmanagedType.LPStr` (default)|A pointer to a null-terminated array of ANSI characters.|
```
NextTurn: In [CharSet Enum](https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.charset):

> Although the common language runtime default is `Auto`, languages may override this default. For example, by default C# marks all methods and types as `Ansi`.

In [DllImportAttribute.CharSet Field](https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.dllimportattribute.charset):

> The default enumeration member for C# and Visual Basic is `CharSet.Ansi` and the default enumeration member for C++ is `CharSet.None`, which is equivalent to `CharSet.Ansi`.

`Auto` is the CLR's default but `Ansi` is the languages' default, which actually takes effect 😕. What are your thoughts?

mairaw:
There are some merge conflicts now. Do you want to take a look @NextTurn?

docs/framework/interop/default-marshaling-for-strings.md
```diff
@@ -1,264 +1,280 @@
...
+
+The solution is to pass a <xref:System.Text.StringBuilder> buffer as the argument instead of a <xref:System.String>. A `StringBuilder` can be dereferenced and modified by the callee, provided it does not exceed the capacity of the `StringBuilder`. It can also be initialized to a fixed length. For example, if you initialize a `StringBuilder` buffer to a capacity of `N`, the marshaler provides a buffer of size (`N`+1) characters. The +1 accounts for the fact that the unmanaged string has a null terminator while `StringBuilder` does not.
+
+For example, the Windows [`GetWindowText`](/windows/desktop/api/winuser/nf-winuser-getwindowtextw) API function (defined in *Windows.h*) requires that the caller pass a fixed-length character buffer to which the function writes the window's text. `LpString` points to a caller-allocated buffer of size `nMaxCount`. The caller is expected to allocate the buffer and set the `nMaxCount` argument to the size of the allocated buffer. The following example shows the `GetWindowText` function declaration as defined in *Windows.h*.
```
NextTurn: `GetWindowText` is defined in *winuser.h*, but one should include *Windows.h* to use it.

https://docs.microsoft.com/windows/desktop/api/winuser/nf-winuser-getwindowtexta#requirements

NextTurn: @rpetrusha

NextTurn:
@rpetrusha I'd like this PR to be merged now so that further changes won't make it messy.

AaronRobinsonMSFT:
This looks fine to me overall. There are some minor nits with respect to how some of this applies to CoreCLR, but I will let @jkoritzinsky make the call on what he wants to do here. Based on the file name, I am assuming this doc is only for framework? @NextTurn is that true?

jkoritzinsky:
@AaronRobinsonMSFT these docs are under the .NET Framework guide on docs.microsoft.com, so I'm fine with this PR since the page does show that it's under the .NET Framework section.

NextTurn:
The corresponding .NET (Core) guide is in [Customizing parameter marshalling](https://docs.microsoft.com/dotnet/standard/native-interop/customize-parameter-marshalling). I guess they'll eventually be merged into a single article (https://github.com/dotnet/docs/pull/11415#discussion_r268227113 @rpetrusha). Please feel free to leave your suggestions with respect to CoreCLR.

rpetrusha:
I'll approve and merge the PR now, @AaronRobinsonMSFT, @jkoritzinsky, and @NextTurn. 

Integrating all of the interop content into a single location in the .NET Guide is a longer-term goal -- at the point content was ported from MSDN, a lot of content was assigned to the .NET Framework Guide that applied equally or equally with modifications to .NET Core and .NET Standard. 

.

rpetrusha (approved):

docs/framework/interop/default-marshaling-for-strings.md
NextTurn: Opened #12249

PR #11425, opened 2019-03-20
Addressed language differences for null conditional operator

rpetrusha:
## Addressed language differences for null conditional operator

The title of the C# article indicated that it applied to Visual Basic as well, reflecting the time that there was a combined C#/VB language reference. This PR corrects the title.

The VB article used `PropertyChanged` as its delegate invocation example. `PropertyChanged` is most commonly an event (though that wasn't actually clear from the VB code, which lacked context). However, Visual Basic (unlike C#) doesn't throw an exception if no invocation list is present when an event is raised. On the other hand, a non-event delegate with an empty invocation list does throw a NullReferenceException. To eliminate confusion, I've replaced `PropertyChanged` with `SendNews`, a delegate of type `Action<String>`.


BillWagner (approved):
This LGTM @rpetrusha 

You can :shipit: when ready.

PR #11457, opened 2019-03-22
Remove mini-TOCs

nxtn:
As well as remove meaningless anchors.

Since there is an **In this article** column on the right.


rpetrusha (changes requested):
Thanks, @NextTurn, for removing these legacy mini-TOCs. I've left some suggestions for the remaining mini-TOC.

rpetrusha (approved):

PR #11491, opened 2019-03-25
Update xaml overview.md

chanthu1993:
## Summary

Describe your changes here.


rpetrusha:
Thank you for contributing to the dotnet/docs repo, @chanthu1993. However, the original content is correct. The first XAML block does not include \<Border.Child>, while the second one does use \<Border.Child> explicitly. I'll close your PR.

PR #11499, opened 2019-03-26
fix: MD003/heading-style/header-style

nschonni:


BillWagner (approved):
Thanks for making this fix @nschonni

I'll :shipit: now.

PR #11500, opened 2019-03-26
Use slash instead of backslash

imba-tjd:
Fix https://github.com/dotnet/docs.zh-cn/issues/357


BillWagner (approved):
Thanks for making this change @imba-tjd  We appreciate it.

Thanks again.

PR #11512, opened 2019-03-26
Use case for AllowExplicitVersion

jzabroski:
Partial Fixes for https://github.com/dotnet/sdk/issues/2879

### DIFF
--- a/docs/core/tools/csproj.md
+++ b/docs/core/tools/csproj.md
@@ -24,6 +24,10 @@ Metapackages are implicitly referenced based on the target framework(s) specifie
  </PropertyGroup>
  ```
 
+### Disabling Implicit Package References
+
+In some scenarios, it might make sense to disable implicit package references.  For example, if you have project `Api.csproj` that targets netcoreapp2.2, and another project `Api.Tests.csproj` that targets `netcoreapp2.2`, then `Api.Tests.csproj` will not know the correct implicit package reference to use (it may choose netcoreapp2.1 dependencies).
+
 ### Recommendations
 
 Since `Microsoft.NETCore.App` or `NetStandard.Library` metapackages are implicitly referenced, the following are our recommended best practices:
--- a/docs/core/tools/csproj.md
+++ b/docs/core/tools/csproj.md
@@ -26,7 +26,13 @@ Metapackages are implicitly referenced based on the target framework(s) specifie
 
 ### Disabling Implicit Package References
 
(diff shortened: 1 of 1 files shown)


jzabroski:
I'm trying to rework this pull request based on @dsplaisted feedback, but am struggling with how to concisely phrase a recommendation.  Suffice to say, I agree with @dsplaisted : AllowExplicitVersion should never be encouraged in the documentation on implicit references, even as a workaround.

Thraka:
If it shouldn't be encouraged, why would we document it? If we're going to document it, then we should point out why it's discouraged.

jzabroski:
@Thraka  I think `AllowExplicitVersion` is basically a "measure of last resort".  There's a separate discussion between @dsplaisted and I in https://github.com/dotnet/sdk/issues/3046#issuecomment-477797052 - he is still trying to understand my issue, so I think this PR is in a holding pattern until we can come up with a permanent solution.

So, while I agree with you documentation is the end goal, I'd rather get it right once than rush something out the door.

Thraka:
OK I've labeled this issue. :) Just follow up when you have the right wording.

jzabroski:
@Thraka  I have sent @dsplaisted a full repro sln that I've uploaded to my GitHub, as part of https://github.com/dotnet/sdk/issues/3046#issuecomment-478624315 - this repro puts an interesting twist as it seems to indicate in some cases AllowExplicitVersion is absolutely required.  What I don't understand is why, and I am hoping Daniel can shed light on it, as in a more complicated solution, his advice to include the metapackage on Tests project works.

Thraka:
@jzabroski any update?

jzabroski:
@dsplaisted  and I closed the related issue - I think that we don't have a good use case for AllowExplicitVersion, other than perhaps the same scenarios solved by `TargetFrameworkVersion` here: https://docs.microsoft.com/en-us/dotnet/core/versions/selection#self-contained-deployments-include-the-selected-runtime - except, instead of framework version, there might be a case where you need to load a specific version of a third-party library.  Either because a bugfix breaks your program (e.g., surfaces a race condition), or because you're trying to reduce the scope of regression testing (human resource constraints) in a mechanical way.

Thraka:
@jzabroski I closed this, you could rollback your delete and keep the content, but as it stands right now, the PR doesn't contain any changes.

jzabroski:
@Thraka Sorry Andy, I was trying to think of where AllowExplicitVersion should go.  I think csproj.md is the right place, but was fumbling with where to put it.

Thraka:
I think it's a good place for it right now. We're going to rework this article eventually. I think that after this section would be good: https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#implicit-version-for-some-package-references

PR #11515, opened 2019-03-26
C76692: Adding break-lines (two spaces) at the end of rows

v-maudel:
Hello, @BillWagner,
Description: Missing break lines at the end of the spotted lines are preventing to show the content in one line instead of one line for every string

Many thanks in advance.

## Summary

Describe your changes here.

### DIFF
--- a/docs/framework/configure-apps/file-schema/wcf/message-of-netmsmqbinding.md
+++ b/docs/framework/configure-apps/file-schema/wcf/message-of-netmsmqbinding.md
@@ -8,12 +8,12 @@ ms.assetid: 6ebf0240-d7be-4493-b0fe-f00fd5989d77
 
 Defines the SOAP message security settings on this `netMsmqBinding` binding.
 
-\<system.ServiceModel>\
-\<bindings>\
-\<netMsmqBinding>\
-\<binding>\
-\<security>\
-\<message>
+\<system.ServiceModel>\  
+\<bindings>\  
+\<netMsmqBinding>\  
+\<binding>\  
+\<security>\  
+\<message>  
 
 ## Syntax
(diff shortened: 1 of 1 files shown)


mairaw:
@v-maudel this makes the \ appear in the docs which is not our intention here. Does using \ at the end of sentences cause any issues for loc? Because we don't want to add the two spaces at the end back.

BillWagner:
Should these just be code-fenced?  Would this be the correct presentation:

```xml
<system.ServiceModel>
<bindings>
<netMsmqBinding>
<binding>
<security>
<message>
```

v-maudel:
Hi @BillWagner @mairaw , the proposed fix from Bill should work. Also I want to mention that in the localized page the last \ is visible. 
image
Maybe, if the \ shouldn't be visible we can remove it and that's perhaps fix the issue but I need to check it with the Bugfixing team first. I'm checking with the team, will let you know asap. 

Regards.

mairaw:
For now, the code fence works. One day though I'd like to recover what we had on [MSDN](https://msdn.microsoft.com/en-us/library/6ebf0240-d7be-4493-b0fe-f00fd5989d77(v=vs.90)?toc=xxx) that we lost through the migrations.

mairaw (changes requested):
Again, let's code fence this instead of adding spaces. And let's match the indentation we had in MSDN (node names are fine):

image

mairaw (approved):

nemrism:
Almost all my changes in PRs pushed to solve #2037 & #2038 use the \ at end of lines...
Do you know if it still break the localization?

/cc @BillWagner / @v-maudel

PR #11524, opened 2019-03-27
Operators overview: link sections from the introduction

pkulikov:
So, the article went live and I've made an experiment:
- In Visual Studio, navigate with the mouse cursor to the `/` operator usage
- Press `F1`
- End up in the browser on the "Arithmetic operators" page a little bit puzzled: introduction mentions the `/` operator but no way to navigate to its definition fast; the navigation at right doesn't jump into eyes immediately. 

And then, I've noticed related feedback on Twitter.
So, adding links to the individual operator sections to the introduction.

### DIFF
--- a/docs/csharp/language-reference/operators/arithmetic-operators.md
+++ b/docs/csharp/language-reference/operators/arithmetic-operators.md
@@ -32,8 +32,8 @@ helpviewer_keywords:
 
 The following operators perform arithmetic operations with numeric types:
 
-- Unary `++` (increment), `--` (decrement), `+` (plus), and `-` (minus) operators.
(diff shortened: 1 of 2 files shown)


BillWagner (approved):
This is a good addition. It will help as we continue making these changes.

I'll :shipit: after the build finishes.

For my reference, can you point to the twitter feedback? I hadn't seen that comment.

Thanks again for being so responsive.

pkulikov:
> For my reference, can you point to the twitter feedback? I hadn't seen that comment.

@BillWagner that was the reply to your tweet :)
https://twitter.com/wim_hoek/status/1110908186150780929

PR #11529, opened 2019-03-27
Added links to sections about floating-point representation

pkulikov:
Provides escapes to the minds that are curious about the origin of round-off errors.


BillWagner (approved):
This is a great fix @pkulikov 
I'll :shipit: now.

Thanks again.

PR #11532, opened 2019-03-28
Expanded upon the positional pattern example

IEvangelist:
## Summary

In preparation for a talk that I'm giving I was reading through the positional pattern example. I felt inclined to fix a sentence that was difficult to read, as part of that clean up I'm also including some additions to the example itself.

Fixes N/A


BillWagner (approved):
Thanks for making these improvements @IEvangelist I like it.

PR #11570, opened 2019-03-29
update samples for scaffolding

BillWagner:
While the previous code worked on the internal review site, it doesn't work on the public site due to some onfig changes.

The short version: the public site always uses Blazor, and always uses scaffolding. That's not how the review site works.


mairaw (approved):
I can't test this even on an incognito window but I'll take your word on this that it works 😉 🙈

BillWagner:
It's Friday afternoon.  I'll :shipit:

🙈 🙉 🙊

PR #11574, opened 2019-03-29
fix: MD010/no-hard-tabs

nschonni:


mairaw (approved):
Thanks for the fixes @nschonni. Looks great!

PR #11583, opened 2019-03-30
typo

KoditkarVedant:
Code snippet gets the "information of `System.Int32` Assembly" and not the "information from an Assembly."

## Summary

change `from` to `of`


BillWagner (approved):
Thanks for making this change @KoditkarVedant  We appreciate it.

Thanks again.

PR #11592, opened 2019-03-31
Replace .NET Framework usage

JamesNK:
Article applies to both .NET Framework and .NET Core. Have replaced ".NET Framework" with just ".NET"


rpetrusha (approved):
Thanks, @JamesNK. I'll merge your PR now.

PR #11603, opened 2019-04-01
Update example code, prevent naming with numbers.

Anduin2017:
## Summary

Update the example code and rename `Class1` to `MyClass`. 

For some programmers may learn from that and add `Class2` and `Class3` which makes their code less maintainable.


jzeferino (approved):

mairaw (approved):
Thank you @Anduin2017 for your contribution. I like your proposed changes! Your contribution should be live within a day or so.

PR #11605, opened 2019-04-01
Update PinvokeLib.dll links

nxtn:
Contributes to #11396


rpetrusha (approved):
Thanks, @NextTurn, for updating these links to point away from archived versions to content that we're currently maintaining. I'll merge your PR now.

PR #11645, opened 2019-04-03
Fixed inconsistent list format

TimoSalomaki:
## Summary

There are two steps when specifying a certificate as the client credential type. Because of that, the current styling of that section of the document is confusing (the first part is an unordered list, the second part is outside the list). I think there are three options to fix it:

1. Numbered list (1 & 2)
2. Unordered list (dots)
3. No list at all

As far as I've seen in other documents, the first option seems to be the most fitting.

Also, the second code sample had a small problem with the indentation of the closing tags.


rpetrusha (approved):
Thank you, @HankiDesign, for contributing to the dotnet/docs repo and for correcting the formatting of this section. A numbered list is indeed the best choice. We'll merge your PR now. Your changes should be live on docs.microsoft.com in the next day or two.

PR #11656, opened 2019-04-04
Added overview of C# Boolean logical operators

### LINKED ISSUE #10740: Abridged Reference: Operators /  logical operators
The logical operators are:

- `&` (logical AND for `bool` operands)
- `|` (logical OR for `bool` operands)
- `^` (logical exclusive-OR for `bool` operands)
- `!` (logical negation)
- `&&` and `||` (conditional logical AND and OR operators)
- `true` and `false` operators (**note:** it's still might be a separate article on its own and only referenced from the overview) 
- corresponding compound assignment operators

Sub-tasks are:

- [ ] Create the new articles
- [ ] Remove old articles.

pkulikov:
Fixes #10740
Supported by dotnet/samples#787

@BillWagner this is ready for review. The reason for the WIP label and remaining part to do — setup redirects. However, before I do that, I would like to confirm several points. Please check my review comments below.

[internal review link](https://review.docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators?branch=pr-en-us-11656)

- edit by @billwagner to add internal review link


docs/csharp/language-reference/operators/boolean-logical-operators.md
```diff
@@ -0,0 +1,159 @@
...
+  - "short-circuiting logical OR operator [C#]"
+  - "|| operator [C#]"
+---
+# Boolean logical operators (C# Reference)
+
+The following operators perform logical operations with [bool](../keywords/bool.md) operands:
+
+- Unary [`!` (logical negation)](#logical-negation-operator-) operator.
+- Binary [`&` (logical AND)](#logical-and-operator-), [`|` (logical OR)](#logical-or-operator-), and [`^` (logical exclusive OR)](#logical-exclusive-or-operator-) operators. Those operators always evaluate both operands.
+- Binary [`&&` (conditional logical AND)](#conditional-logical-and-operator-) and [`||` (conditional logical OR)](#conditional-logical-or-operator-) operators. Those operators evaluate the second operand only if it's necessary.
```
pkulikov: Given that listed operators are for `bool` operands, I haven't included `true` and `false` operators here; also, I assume their usage is very rare with comparison to the operators in this article. Anyway, the `true` and `false` operators are mentioned in the "Operator overloadability" section.

Please let me know, if it's fine.

docs/csharp/language-reference/operators/boolean-logical-operators.md
```diff
@@ -0,0 +1,159 @@
...
+
+The `&` and `|` operators are also defined for `bool?` operands. For more information, see [The bool? type](../../programming-guide/nullable-types/using-nullable-types.md#the-bool-type) section of the [Using nullable types](../../programming-guide/nullable-types/using-nullable-types.md) article.
+
+For the operands of [integral](../keywords/integral-types-table.md) types, the `&`, `|`, and `^` operators perform bitwise logical operations.
```
pkulikov: No links to operators reference by design (though individual sections below have links): I'll add the link to the overview of bitwise operators here. To create such an overview is my next PR.

docs/csharp/language-reference/operators/boolean-logical-operators.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Logical negation operator !
```
pkulikov: Should I mention the null-forgiving operator and add the link to the nullable reference types overview? Or that is to be done after the official release of C# 8.0?

docs/csharp/language-reference/operators/boolean-logical-operators.md
```diff
@@ -0,0 +1,159 @@
...
+
+The unary `!` operator computes logical negation of its operand. That is, it produces `true`, if the operand evaluates to `false`, and `false`, if the operand evaluates to `true`:
+
+[!code-csharp-interactive[logical negation](~/samples/snippets/csharp/language-reference/operators/LogicalOperators.cs#Negation)]
+
+## Logical AND operator &amp;
```
pkulikov: @BillWagner could you please tell me what anchor is generated for this header on the review site. I use `logical-and-operator-` in this article (does it work?); however, Visual Studio Code doesn't recognize such a link and doesn't navigate me to the section.

docs/csharp/language-reference/operators/boolean-logical-operators.md
```diff
@@ -0,0 +1,159 @@
...
+
+The `&` operator evaluates both operands even if the first operand evaluates to `false`, so that the result must be `false` regardless of the value of the second operand. The following example demonstrates that behavior:
+
+[!code-csharp-interactive[logical AND](~/samples/snippets/csharp/language-reference/operators/LogicalOperators.cs#And)]
+
+The [conditional logical AND operator](#conditional-logical-and-operator-) `&&` also computes the logical AND of its operands, but evaluates the second operand only if the first operand evaluates to `true`.
+
+For the operands of integral types, the `&` operator computes [bitwise logical AND](and-operator.md#integer-logical-bitwise-and-operator) of its operands. The unary `&` operator is the [address-of operator](and-operator.md#unary-address-of-operator).
+
+## Logical exclusive OR operator ^
```
pkulikov: I order operator sections in each overview by the operator precedence.

docs/csharp/language-reference/operators/boolean-logical-operators.md
```diff
@@ -0,0 +1,159 @@
...
+
+The `|` operator evaluates both operands even if the first operand evaluates to `true`, so that the result must be `true` regardless of the value of the second operand. The following example demonstrates that behavior:
+
+[!code-csharp-interactive[logical OR](~/samples/snippets/csharp/language-reference/operators/LogicalOperators.cs#Or)]
+
+The [conditional logical OR operator](#conditional-logical-or-operator-) `||` also computes the logical OR of its operands, but evaluates the second operand only if the first operand evaluates to `false`.
+
+For the operands of integral types, the `|` operator computes [bitwise logical OR](or-operator.md) of its operands.
+
+## Conditional logical AND operator &amp;&amp;
```
pkulikov: Also, what is the generated anchor for this header? Does `conditional-logical-and-operator-` work?

docs/csharp/language-reference/operators/toc.yml
```diff
@@ -4,6 +4,9 @@
   - name: Arithmetic operators
     href: arithmetic-operators.md
     displayName: ++, --, +, -, *, /, %
+  - name: Boolean logical operators
+    href: boolean-logical-operators.md
+    displayName: "&, |, ^, !"
```
pkulikov: @BillWagner question about the `displayName` field (as far as I understand it's *one* string that is checked if it contains the search entry): do we want to use commas to separate symbols? Or white spaces are more preferred? Now, if I search for "+ -" in the TOC I don't get "Arithmetic operators" entry; the search for "+, -" picks up the overview article. Is there any data about what search strings are more popular?

docs/csharp/language-reference/operators/boolean-logical-operators.md
BillWagner: I agree. The `true` and `false` operators should be a separate topic.

docs/csharp/language-reference/operators/boolean-logical-operators.md
BillWagner: I think that's a good plan.

I am watching traffic patterns to determine when readers land in the wrong article, and which article they are trying to find.

docs/csharp/language-reference/operators/boolean-logical-operators.md
BillWagner: Let's wait on that. I'd rather add a link to a page on the null forgiveness operator (which doesn't exist yet).

A nice side-effect of using the displayName in the TOCs is that it's easy to find which pages discuss an operator and a different usage. 😸

docs/csharp/language-reference/operators/boolean-logical-operators.md
BillWagner: The anchor in the review site is `logical-and-operator-` .  Good guess!

docs/csharp/language-reference/operators/boolean-logical-operators.md
BillWagner: This one is `conditional-logical-and-operator-`.  Good guess again.

docs/csharp/language-reference/operators/toc.yml
BillWagner: Here's what the docs say on the `displayName` feature:

> `displayName` (optional) - Alternate search terms for TOC filtering. An additional string value that doesn’t get displayed (yes, it's a poor name) but is searched, as well as `name`, during TOC filtering. For multiple values, use a comma separated list such as `displayName: batch, asynchonous, off-line, offline`

We don't have data on what symbols people search for (yet). My hypothesis is that most will search for one operator token.

docs/csharp/language-reference/operators/boolean-logical-operators.md
```diff
@@ -0,0 +1,159 @@
...
+
+The unary `!` operator computes logical negation of its operand. That is, it produces `true`, if the operand evaluates to `false`, and `false`, if the operand evaluates to `true`:
+
+[!code-csharp-interactive[logical negation](~/samples/snippets/csharp/language-reference/operators/LogicalOperators.cs#Negation)]
+
+## Logical AND operator &amp;
+
+The `&` operator computes the logical AND of its operands. The result of `x & y` is `true` if both `x` and `y` evaluate to `true`. Otherwise, the result is `false`.
+
+The `&` operator evaluates both operands even if the first operand evaluates to `false`, so that the result must be `false` regardless of the value of the second operand. The following example demonstrates that behavior:
```
BillWagner: As you introduce the concept of short-circuiting (or not), I think it's important to explain that `x` and `y` could be Boolean expressions (including method calls). In other words, the operands are not limited to variables.

BillWagner: This is a super-nit, but it's more accurate to say it as follows:

> ...even if the first operand doesn't evaluate to `true`, so that …

That difference only matters for types that override operator `true` and operator `false` *and* could be in a state where both operators return `false`.

The pertinent clause in the spec is [here](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#conditional-logical-operators). See the two bullets following the sample.

docs/csharp/language-reference/operators/boolean-logical-operators.md
```diff
@@ -0,0 +1,159 @@
...
+
+[!code-csharp-interactive[logical exclusive-OR](~/samples/snippets/csharp/language-reference/operators/LogicalOperators.cs#Xor)]
+
+For the operands of integral types, the `^` operator computes [bitwise logical exclusive OR](xor-operator.md) of its operands.
+
+## Logical OR operator |
+
+The `|` operator computes the logical OR of its operands. The result of `x | y` is `true` if either `x` or `y` evaluates to `true`. Otherwise, the result is `false`.
+
+The `|` operator evaluates both operands even if the first operand evaluates to `true`, so that the result must be `true` regardless of the value of the second operand. The following example demonstrates that behavior:
```
BillWagner: This should be updated similarly to my previous comment, except that the first operand doesn't evaluate to `false`.

docs/csharp/language-reference/operators/boolean-logical-operators.md
```diff
@@ -0,0 +1,159 @@
...
+
+[!code-csharp-interactive[logical OR](~/samples/snippets/csharp/language-reference/operators/LogicalOperators.cs#Or)]
+
+The [conditional logical OR operator](#conditional-logical-or-operator-) `||` also computes the logical OR of its operands, but evaluates the second operand only if the first operand evaluates to `false`.
```
BillWagner: Again:  

> only if the first operand doesn't evaluate to `true`.

docs/csharp/language-reference/operators/boolean-logical-operators.md
```diff
@@ -0,0 +1,159 @@
...
+
+For the operands of integral types, the `|` operator computes [bitwise logical OR](or-operator.md) of its operands.
+
+## Conditional logical AND operator &amp;&amp;
+
+The conditional logical AND operator `&&`, also known as the "short-circuiting" logical AND operator, computes the logical AND of its operands. The result of `x && y` is `true` if both `x` and `y` evaluate to `true`. Otherwise, the result is `false`. If the first operand evaluates to `false`, the second operand is not evaluated and the result of operation is `false`. The following example demonstrates that behavior:
```
BillWagner:
```suggestion
The conditional logical AND operator `&&`, also known as the "short-circuiting" logical AND operator, computes the logical AND of its operands. The result of `x && y` is `true` if both `x` and `y` evaluate to `true`. Otherwise, the result is `false`. The second operand is evaluated only when the first operand is not `false`. The following example demonstrates that behavior:
```

docs/csharp/language-reference/operators/boolean-logical-operators.md
```diff
@@ -0,0 +1,159 @@
...
+
+[!code-csharp-interactive[conditional logical AND](~/samples/snippets/csharp/language-reference/operators/LogicalOperators.cs#ConditionalAnd)]
+
+The [logical AND operator](#logical-and-operator-) `&` also computes the logical AND of its operands, but always evaluates both operands.
+
+## Conditional logical OR operator ||
+
+The conditional logical OR operator `||`, also known as the "short-circuiting" logical OR operator, computes the logical OR of its operands. The result of `x || y` is `true` if either `x` or `y` evaluates to `true`. Otherwise, the result is `false`. If the first operand evaluates to `true`, the second operand is not evaluated and the result of operation is `true`. The following example demonstrates that behavior:
```
BillWagner:
```suggestion
The conditional logical OR operator `||`, also known as the "short-circuiting" logical OR operator, computes the logical OR of its operands. The result of `x || y` is `true` if either `x` or `y` evaluates to `true`. Otherwise, the result is `false`. The second operand is evaluated only when the first operand is not `true`. The following example demonstrates that behavior:
```

docs/csharp/language-reference/operators/boolean-logical-operators.md
```diff
@@ -0,0 +1,159 @@
...
+
+[!code-csharp-interactive[operator precedence](~/samples/snippets/csharp/language-reference/operators/LogicalOperators.cs#Precedence)]
+
+For the complete list of C# operators ordered by precedence level, see [C# operators](index.md).
+
+## Operator overloadability
+
+A user-defined type can [overload](../keywords/operator.md) the `!`, `&`, `|`, and `^` operators. When a binary operator is overloaded, the corresponding compound assignment operator is also implicitly overloaded. A user-defined type cannot explicitly overload a compound assignment operator.
+
+A user-defined type cannot overload the conditional logical operators `&&` and `||`. However, if a user-defined type overloads the `&` operator (the `|` operator) and the [true and false operators](../keywords/true-false-operators.md) in a certain way, the `&&` operation (the `||` operation) can be evaluated for the operands of that type. For more information, see the [User-defined conditional logical operators](~/_csharplang/spec/expressions.md#user-defined-conditional-logical-operators) section of the [C# language specification](~/_csharplang/spec/introduction.md).
```
BillWagner: The parenthetical expressions "(the `|` operator)" and "(the `||` operator) look odd to me. I don't have a good suggestion right now. I'll keep noodling. But, if you come up with a better wording make an edit here.

docs/csharp/language-reference/operators/toc.yml
```diff
@@ -4,6 +4,9 @@
   - name: Arithmetic operators
     href: arithmetic-operators.md
     displayName: ++, --, +, -, *, /, %
+  - name: Boolean logical operators
```
BillWagner: Not related to this PR, but the H2 anchors in the TOC are not officially supported. We should remove them.

(Yes, they do work now, but that's not guaranteed behavior).

BillWagner (commented):
Adding comments on this review.

It's a great start.  I did do a thorough read of the text, so hopefully once the samples are merged and I see a clean build, this should be close to ready. I had a small number of comments (most on one bit of spec trivia)

docs/csharp/language-reference/operators/boolean-logical-operators.md
pkulikov: Cool, thanks!

docs/csharp/language-reference/operators/toc.yml
pkulikov: > My hypothesis is that most will search for one operator token

I think that's the case too. So, I was nit-picking about low traffic cases. Then, I will follow the docs guidance to "use a comma separated list". Thank you for the docs quote!

docs/csharp/language-reference/operators/boolean-logical-operators.md
pkulikov: @BillWagner hm, I had a feeling I've explained it: the "evaluates" word means that an operand could be an expression (or not?). Also, at the example one of operands is a local function call, clearly not a variable. Do you suggest explicitly state in the article that operands might be expressions? I feel a bit odd here, because operands typically could be expressions and only exceptions to that rule are usually mentioned in the docs (for example, the `++` and `--` operators).

pkulikov: @BillWagner I like this nit-picking; also your comment confused me a bit.

According to the spec `x && y` is evaluated as `T.false(x) ? x : T.&(x, y)`. 

First, the `true` operator doesn't participate at all in definition and we cannot refer to it then. I assume when you say
>`x` doesn't evaluate to `true`

you talk about the `true` operator and mean that "`x` is not definitely true", or `T.true(x)` returns `false`. Do I understand you correctly?

Second, back to `x && y` as `T.false(x) ? x : T.&(x, y)`. I read that definition as 
> If the first operand is definitely false, do short-circuit. Otherwise, call the `&` operator (which means the second operator is evaluated as well).

So, when we say that the second operand is evaluated always, we emphasize that "always" means even if the first operand is definitely false (so even if there is condition for short-circuiting). I think that "the first operand is definitely false" can be rephrased as "the first operand evaluates to `false`".

If there is a flaw in my reasoning, please let me know. I remember I've looked at that spec and tried to build a sentence with keeping user-defined conditional operators in mind, but might have missed or misinterpreted something.

docs/csharp/language-reference/operators/boolean-logical-operators.md
pkulikov: Please check my comment to the `&` operator section.

docs/csharp/language-reference/operators/boolean-logical-operators.md
pkulikov: @BillWagner is it fine to replace in your suggestion "first operand is not `false`" with "first operand doesn't evaluate to `false`" I understand that it cannot be replaced with "first operand evaluates to `true`"

docs/csharp/language-reference/operators/boolean-logical-operators.md
pkulikov: I like you've mentioned it. I'm also not fully satisfied with parentheses. Wander more about it, then.

docs/csharp/language-reference/operators/toc.yml
pkulikov: @BillWagner does that mean that the following:
```
- name: / Operator
  href: arithmetic-operators.md#division-operator-
```

should be replaced with:
```
- name: / Operator
  href: arithmetic-operators.md
```

??

docs/csharp/language-reference/operators/boolean-logical-operators.md
pkulikov: @BillWagner while your suggestion is correct, I'm afraid readers would translate it to "...evaluates to `false`" anyway ("doesn't evaluate to `true`" might be hard to comprehend), with most proactive of them submitting PR to "fix" wording.

I suggest another fix. I've noticed that I'm not consistent when talking about short-circuiting behavior. Sometimes I mention condition for short-circuiting ("...the first operand evaluates to `true`" in case of `||`), in other cases — condition for evaluation of the second operand ("...the first operand doesn't evaluate to `true`"). I think only one condition should be mentioned. I suggest to describe condition for short-circuit; in particular, because the [spec](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#user-defined-conditional-logical-operators) definition of user-defined conditional operators says when to short-circuit.

Then, the paragraph in question would read:
>The conditional logical OR operator `||` also computes the logical OR of its operands, but doesn't evaluate the second operand if the first operand evaluates to `true`.

docs/csharp/language-reference/operators/boolean-logical-operators.md
BillWagner: You're analysis is correct. I'll resolve this conversation.

I had started editing and suggesting, then pulled out my copy of the spec as I got farther along. I was trying to make small edits without checking. That's always a mistake.

docs/csharp/language-reference/operators/boolean-logical-operators.md
BillWagner: Same answer... Resolving.

docs/csharp/language-reference/operators/boolean-logical-operators.md
BillWagner: I like your wording better. Let's go with it.

docs/csharp/language-reference/operators/boolean-logical-operators.md
BillWagner: Resolved again.

docs/csharp/language-reference/operators/toc.yml
BillWagner: Sorry, I should have been more clear.

I want to try removing those items from the TOC. Ideally, readers find the individual operators as needed. If not, we'll address (and I'll likely keep pushing for a better experience for H2s in the TOC.

BillWagner:
samples are merged. Closing and re-opening to get a fresh build.

docs/csharp/language-reference/operators/toc.yml
pkulikov: > I want to try removing those items from the TOC.

Ah! OK, let's do that.

By the way, with the new TOC feature that filters by `displayName`, maybe it's time to revise the UX. The current one:
image


It says "Filter by title". If most of the individual operator entries are removed and it's obvious that there are no titles in TOC with "+" in it, I wouldn't even try to type "+" into this field.

BillWagner: I asked for a change See MicrosoftDocs/feedback#1359

docs/csharp/language-reference/operators/boolean-logical-operators.md
pkulikov: @BillWagner I've updated the introductions to the examples for the `&`, `|`, `&&`, and `||` operators. Also, examples are updated: dotnet/samples#797

What do you think of a new text?

docs/csharp/language-reference/operators/boolean-logical-operators.md
pkulikov: Rephrased.

pkulikov:
@BillWagner this is ready for the final review. Please take a look at
- the new section about nullable Boolean logical operators
- rephrased parts about short-circuiting behavior (I only describe the condition to short-circuit, which is not evaluate the second operand and also try to make clearer that operands are not limited to variables)

BillWagner:
closing and re-opening to get a fresh build with new snippets.

BillWagner (approved):
This is a great set of work @pkulikov 

I'll :shipit: now.  

Thanks for all this work, and the ongoing conversations to make this section of the reference better.  💯

PR #11709, opened 2019-04-06
Indicate default charset in VB

nxtn:
Contributes to #11396


BillWagner (approved):
Thanks for making these changes @NextTurn 
I'll :shipit: now.

PR #11716, opened 2019-04-07
make caller info sample work with C#

CaringDev:
## Summary
Make this sample work when called from C#

### TL;DR;
As it is, C# (respectively the C# compiler) does not understand how to handle e.g. parameters of type `option<string>` decorated with some `[CallerInfoX]` attribute. Instead we would need to supply some `FSharpOption<string>`. Marking the `option`al parameter `[<Optional>]` leads to an optional option, which is OK from C# perspective<sup>1</sup> but always `null`, i.e. `None`. I believe, we should have examples that work for F# and C# out-of-the-box. I'm happy to provide two samples including an explanation, if you think different.

<sup>1</sup> maybe the compiler should emit a warning if it encounters a [CallerInfo] attribute on a type it cannot supply?

### DIFF
--- a/docs/fsharp/language-reference/caller-information.md
+++ b/docs/fsharp/language-reference/caller-information.md
@@ -22,24 +22,22 @@ The following example shows how you might use these attributes to trace a caller
 ```fsharp
 open System.Diagnostics
 open System.Runtime.CompilerServices
+open System.Runtime.InteropServices
 
 type Tracer() =
     member __.DoTrace(message: string,
-                      [<CallerMemberName>] ?memberName: string,
-                      [<CallerFilePath>] ?path: string,
-                      [<CallerLineNumber>] ?line: int) =
+                      [<CallerMemberName; Optional; DefaultParameterValue("")>] memberName: string,
+                      [<CallerFilePath; Optional; DefaultParameterValue("")>] path: string,
+                      [<CallerLineNumber; Optional; DefaultParameterValue(0)>] line: int) =
         Trace.WriteLine(sprintf "Message: %s" message)
-        match (memberName, path, line) with
-        | Some m, Some p, Some l ->
-            Trace.WriteLine(sprintf "Member name: %s" m)
-            Trace.WriteLine(sprintf "Source file path: %s" p)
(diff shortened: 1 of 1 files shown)


docs/fsharp/language-reference/caller-information.md
```diff
@@ -22,24 +22,22 @@ The following example shows how you might use these attributes to trace a caller
...
 open System.Diagnostics
 open System.Runtime.CompilerServices
+open System.Runtime.InteropServices
 
 type Tracer() =
     member __.DoTrace(message: string,
-                      [<CallerMemberName>] ?memberName: string,
-                      [<CallerFilePath>] ?path: string,
-                      [<CallerLineNumber>] ?line: int) =
+                      [<CallerMemberName; Optional>] memberName: string,
```
cartermp:
```suggestion
                      [<CallerMemberName; Optional; DefaultParameterValue("")>] memberName: string,
```

docs/fsharp/language-reference/caller-information.md
```diff
@@ -22,24 +22,22 @@ The following example shows how you might use these attributes to trace a caller
...
+                      [<CallerFilePath; Optional>] path: string,
```
cartermp:
```suggestion
                      [<CallerFilePath; Optional; DefaultParameterValue("")>] path: string,
```

docs/fsharp/language-reference/caller-information.md
```diff
@@ -22,24 +22,22 @@ The following example shows how you might use these attributes to trace a caller
...
+                      [<CallerLineNumber; Optional>] line: int) =
```
cartermp:
```suggestion
                      [<CallerLineNumber; Optional; DefaultParameterValue(0)>] line: int) =
```

cartermp (changes requested):
Looks good modulo these changes (this would make the sample equivalent to the C# docs)

CaringDev:
@cartermp ah, sorry, I should have batch-merged your suggestions instead of fixing them locally... now Github does not recognise them as resolved. Could you check again? Thank you.

cartermp (approved):
Thanks @CaringDev

PR #11723, opened 2019-04-08
Escape \ in table

svick:
Correctly escapes the `\` character in a table. (It seems `\\` doesn't work, at least in GitHub preview.)

To verify that this is actually supposed to be `\` (and not `|`, as is currently shown), I looked at [the relevant source code](https://github.com/dotnet/coreclr/blob/9a465e4/src/System.Private.CoreLib/shared/System/Globalization/DateTimeFormat.cs#L449) and it indeed handles `\` in a special way, but does not have any code for `|`.


rpetrusha (approved):
Thanks, @svick, for noticing and making this correction. I'll merge your PR now.

PR #11741, opened 2019-04-09
Clarify mentioned difference between lifted operators

pkulikov:
I've mentioned that the behavior of `&` and `|` with `bool?` differs from the typical one of lifted operators, but didn't describe how. Explicitly stating that by this PR.


BillWagner (approved):
This is a great clarification @pkulikov 

I'll :shipit: now.

PR #11774, opened 2019-04-10
Update MessageBox PInvoke example

### LINKED ISSUE #11772: Interop MessageBox example should be updated.
The initial example should  be updated to map the "fully" correct way. While the example partially works, it doesn't work well and could cause grief as developers try and build upon or explore the example.

As the example stands it will invoke the legacy ANSI version of `GetMessageBox`, or `GetMessageBoxA`. Not only is it slower, but characters that don't fit into the ANSI code page will be turned into "garbage".

Additionally it isn't setting the last error, which the API is documented to do. If you don't check for the error this doesn't matter, but if you do you will get some random value when calling `Marshal.GetLastWin32Error()`.

https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-messagebox

Here is an updated example:

``` C#
using System;
using System.Runtime.InteropServices;

public class Program
{

    // Import user32.dll (containing the function we need) and define
    // the method corresponding to the native function.
    [DllImport("user32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
    public static extern int MessageBox(IntPtr hWnd, string text, string caption, int options);

    public static void Main(string[] args)
    {
        // Invoke the function as a regular managed method.
        MessageBox(IntPtr.Zero, "Command-line message box", "Attention!", 0);
    }
}
```

cc: @jkoritzinsky 

---
#### Document Details

* ID: cf08b819-7446-dc6c-8040-d55ecccd317d
* Version Independent ID: 931f4809-51f0-5b61-30ec-760bcfe44cb8
* Content: [Platform Invoke (P/Invoke)](https://docs.microsoft.com/en-us/dotnet/standard/native-interop/pinvoke#feedback)
* Content Source: [docs/standard/native-interop/pinvoke.md](https://github.com/dotnet/docs/blob/master/docs/standard/native-interop/pinvoke.md)
* Product: **dotnet**
* GitHub Login: @jkoritzinsky

jkoritzinsky:
Fixes #11772


docs/standard/native-interop/pinvoke.md
```diff
@@ -32,7 +32,7 @@ public class Program {
 The previous example is simple, but it does show off what's needed to invoke unmanaged functions from managed code. Let’s step through the example:
 
 *   Line #1 shows the using statement for the `System.Runtime.InteropServices` namespace that holds all the items needed.
-*   Line #7 introduces the `DllImport` attribute. This attribute is crucial, as it tells the runtime that it should load the unmanaged DLL. The string passed in is the DLL our target function is in.
+*   Line #7 introduces the `DllImport` attribute. This attribute is crucial, as it tells the runtime that it should load the unmanaged DLL. The string passed in is the DLL our target function is in. Additionally, it specifies which [character set](./charset.md) to use for marshalling the strings. Finally, it specifies that this function calls `SetLastError` and that the runtime should capture that error code so the user can retrieve it via <xref:System.Runtime.InteropServices.Marshal.GetLastWin32Error>.
```
rpetrusha:
```suggestion
*   Line #7 introduces the `DllImport` attribute. This attribute is crucial, as it tells the runtime that it should load the unmanaged DLL. The string passed in is the DLL our target function is in. Additionally, it specifies which [character set](./charset.md) to use for marshalling the strings. Finally, it specifies that this function calls [SetLastError](~/windows/desktop/api/errhandlingapi/nf-errhandlingapi-setlasterror) and that the runtime should capture that error code so the user can retrieve it via <xref:System.Runtime.InteropServices.Marshal.GetLastWin32Error?displayProperty=nameWithType>.
```

rpetrusha (changes requested):

rpetrusha:
Thanks, @jkoritzinsky. I thought that SetLastError would be most helpful as a link, and that Marshal.GetLastError should be fully qualified.

rpetrusha (approved):

rpetrusha:
Thanks for making the additional changes, @jkoritzinsky. I'll merge your PR now.

PR #11805, opened 2019-04-12
Invalid dockerfile

stevejgordon:
When reading a part of this document I noticed that the dockerfile example appears to be invalid.

I believe -0 is meant to be -o to define the output directory for the build and publish steps.

Also, the final stage copy should include the destination for the copy.

## Summary
This PR includes fixes for these points.

NOTE: I incorrectly stated dockercompose in the commit, it should be dockerfile!


rpetrusha (approved):
Thank you, @stevejgordon, for contributing to the dotnet/docs repo and making these corrections. We'll merge your PR now. Your changes should be visible on docs.microsoft.com in the next day or two.

PR #11831, opened 2019-04-14
Added overview of C# bitwise and shift operators

### LINKED ISSUE #10731: Abridged Reference: operators / bitwise operators
The bitwise operators are:

- `&` (bitwise AND for integral operands)
- `|` (bitwise OR for integral operands)
- `^` (bitwise exclusive-OR for integral operands)
- `~` (bitwise negation)
- `<<` and `>>` (shift operators)
- corresponding compound assignment operators

Sub-tasks are:

- [ ] Create the new articles
- [ ] Remove old articles.

pkulikov:
Fixes #10731
Supported by dotnet/samples#816


docs/csharp/language-reference/operators/arithmetic-operators.md
```diff
@@ -161,10 +161,10 @@ You also can use the `-` operator for delegate removal. For more information, se
...
 The following list orders arithmetic operators starting from the highest precedence to the lowest:
 
-- Postfix increment `x++` and decrement `x--` operators.
-- Prefix increment `++x` and decrement `--x` and unary `+` and `-` operators.
-- Multiplicative `*`, `/`, and `%` operators.
-- Additive `+` and `-` operators.
+- Postfix increment `x++` and decrement `x--` operators
+- Prefix increment `++x` and decrement `--x` and unary `+` and `-` operators
+- Multiplicative `*`, `/`, and `%` operators
+- Additive `+` and `-` operators
```
pkulikov: Updated punctuation in this file based on the [style guide](https://docs.microsoft.com/en-us/style-guide/scannable-content/lists#punctuation)

docs/csharp/language-reference/operators/bitwise-and-shift-operators.md
```diff
@@ -0,0 +1,182 @@
...
+
+For information about how the second operand of the `>>` operator defines the shift count, see the [Shift count of the shift operators](#shift-count-of-the-shift-operators) section.
+
+## Logical AND operator &amp;
+
+The `&` operator computes the bitwise logical AND of its operands:
+
+[!code-csharp-interactive[bitwise AND](~/samples/snippets/csharp/language-reference/operators/BitwiseAndShiftOperators.cs#BitwiseAnd)]
+
+For the operands of the `bool` type, the `&` operator computes the [logical AND](boolean-logical-operators.md#logical-and-operator-) of its operands. The unary `&` operator is the [address-of operator](and-operator.md#unary-address-of-operator).
```
pkulikov: Alternative phrasing to start the sentence: "For the `bool` operands,..." Should we switch to this one? If yes, then how to talk about other types:
- "For the numeric (integral) type operands..."
- "For the numeric (integral) operands..." (sounds best to me)
- "For numeric (integral) operands..."
- Or just use the existing approach: "For the operands of the numeric (integral) types..."

docs/csharp/language-reference/operators/bitwise-and-shift-operators.md
```diff
@@ -0,0 +1,182 @@
...
+
+is equivalent to
+
+```csharp
+x = x op y
+```
+
+except that `x` is only evaluated once.
+
+The following example demonstrates the usage of compound assignment with the bitwise and shift operators:
```
pkulikov: Is the article necessary here: "..with **the** bitwise and shift operators:"?
I've used it as the text is about defined five operators. If current version is correct, then I'll need to update other overviews with "the arithmetic operators", "the logical operators".

pkulikov (commented):
@BillWagner I have a number of questions about the usage of "the". Any feedback will be appreciated. However, I'll address it not in this PR, but in a separate one (to update other overviews and for simpler review).

rpetrusha:
Closing and reopening to begin new build.

docs/csharp/language-reference/operators/bitwise-and-shift-operators.md
```diff
@@ -0,0 +1,182 @@
+---
+title: "Bitwise and shift operators - C# Reference"
+description: "Learn about C# operators that perform bitwise logical or shift operations with the operands of the integral types."
```
BillWagner:
```suggestion
description: "Learn about C# operators that perform bitwise logical or shift operations with operands of integral types."
```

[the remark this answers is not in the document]
BillWagner: I think without the article is preferred.

docs/csharp/language-reference/operators/bitwise-and-shift-operators.md
```diff
@@ -0,0 +1,182 @@
...
+  - "& operator [C#]"
+  - "bitwise logical exclusive OR operator [C#]"
+  - "bitwise logical XOR operator [C#]"
+  - "^ operator [C#]"
+  - "bitwise logical OR operator [C#]"
+  - "| operator [C#]"
+---
+# Bitwise and shift operators (C# Reference)
+
+The following operators perform bitwise or shift operations with the operands of the [integral types](../keywords/integral-types-table.md):
```
BillWagner: The shift operators are only defined on `int`, `uint`, `long`, and `ulong`. They are not defined on other integral types (Note that I can't say why they aren't defined on `ushort`, or `short`, but it's what the [spec says](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#shift-operators).)

The same is true for the bitwise complement (`~`) operator. See [spec link](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#bitwise-complement-operator)

And also true for the [logical operators](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#logical-operators).

Yes, I had to check the spec to make sure on this one. I always enjoy those days.

docs/csharp/language-reference/operators/bitwise-and-shift-operators.md
BillWagner: I like the current phrasing: This article is about the bitwise operators, so pointing out that `bool` is different, with a link, feels natural.

Also, it avoids the fact that it's not all *integral types*. This article applies to `int`, `uint`, `long`, and `ulong`.

docs/csharp/language-reference/operators/bitwise-and-shift-operators.md
BillWagner: I'd prefer removing it.

docs/csharp/language-reference/operators/boolean-logical-operators.md
```diff
@@ -41,7 +41,7 @@ The following operators perform logical operations with the [bool](../keywords/b
 - Binary [`&` (logical AND)](#logical-and-operator-), [`|` (logical OR)](#logical-or-operator-), and [`^` (logical exclusive OR)](#logical-exclusive-or-operator-) operators. Those operators always evaluate both operands.
 - Binary [`&&` (conditional logical AND)](#conditional-logical-and-operator-) and [`||` (conditional logical OR)](#conditional-logical-or-operator-) operators. Those operators evaluate the second operand only if it's necessary.
 
-For the operands of [integral](../keywords/integral-types-table.md) types, the `&`, `|`, and `^` operators perform bitwise logical operations.
+For the operands of the [integral](../keywords/integral-types-table.md) types, the `&`, `|`, and `^` operators perform bitwise logical operations. For more information, see [Bitwise and shift operators](bitwise-and-shift-operators.md).
```
BillWagner: I think this is fine here, despite my pedantic comments on the other article.

BillWagner (changes requested):
Overall, this is excellent @pkulikov 

I had a few comments on this one, based on some text where this text differs from the spec.

docs/csharp/language-reference/operators/bitwise-and-shift-operators.md
pkulikov: @BillWagner yes, they are defined for only those four, but another four types are also *supported*. That is, operands might be of `byte`, but they would be converted to `int` before the operation. The same takes place for the arithmetic operators. However, we don't stress that there. That's why, in the text, I don't say that operators are *defined*, but perform operations. However I silence the fact that less-than-32-bit operands are promoted to `int` in the process of such an operation. I assumed that knowledge  is for the spec. Do you think *this* overview should be more explicit about integer operand promotion? Then, I will think how to update the introduction.

BillWagner: 🤔 It might be more important for these operators than the arithmetic operators. My reasoning is that the result is the promoted type. 

```csharp
byte b = 0x88;
var shifted = b >> 3;
Console.WriteLine(shifted.GetType()); // "System.Int32"
```

In the above example, it doesn't matter, but it matter when shifting left. The result might be truncated if stored in its original type:

```csharp
    byte b = 0x88;
    var shifted = b << 3;
    Console.WriteLine(shifted.GetType());
    Console.WriteLine(shifted); // 1088
    byte b2 = (byte) shifted;
    Console.WriteLine(b2);  // 64
```

What are your thoughts?

pkulikov: @BillWagner 

> It might be more important for these operators than the arithmetic operators.

> ...it matter when shifting left.

Agree, in the end, those bit operations work not with the numbers but rather with the sets of bits. "Exploding" a bit set in size is indeed quite important to be mentioned, especially, for the shift operations. 

I'll update the text and add an example for the left shift. Thanks for pointing out to the matter.

pkulikov:
@BillWagner I've added paragraphs about numeric promotion; please take a look. Supporting snippets are dotnet/samples#820

BillWagner:
closing & reopening to get a fresh build.

BillWagner (approved):

BillWagner:
The latest changes look great @pkulikov 

I've approved this, and I'll :shipit: after checking the build status.

PR #11838, opened 2019-04-15
Update LoadAndCompleteInstance method

tarikkirsanli:
Activity contents should be the same.

## Summary

WorkflowApplication has to be loaded with the same activity so that changed the line of creating a new instance of WorkflowApplication.

##Fixes 
"Pausing and Resuming a Workflow" example


rpetrusha (approved):
Thank you, @KoadNa, for contributing to the dotnet/docs repo and correcting the example. We'll merge your PR now. The change should be live on docs.microsoft.com in the next day or two.

PR #11860, opened 2019-04-16
Added true and false operators

paulomorgado:
## Summary

The `true` and `false` operators were missing.

### DIFF
--- a/docs/csharp/language-reference/operators/index.md
+++ b/docs/csharp/language-reference/operators/index.md
@@ -153,6 +153,14 @@ This operator has higher precedence than the next section and lower precedence t
 
 [x &#124; y](or-operator.md) – logical or bitwise OR. You can generally use this with integer types and `enum` types.
 
+## True operator
+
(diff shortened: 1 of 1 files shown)


rpetrusha (approved):
Thanks, @paulomorgado, for contributing to the dotnet/docs repo and adding the missing operators. We'll merge your PR now. Your changes should be live on docs.microsoft.com in the next day or two.

pkulikov:
@paulomorgado @rpetrusha the sections are ordered by operator precedence, so added sections should be merged into one and moved. But to which location? Even the language spec says nothing about precedence of the `true` and `false` operators:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#operator-precedence-and-associativity

In general, those two operators are exceptional: they are not predefined and a user must overload them to use. So, not sure if they should be listed at all.

paulomorgado:
Maybe @BillWagner can shed some light into this.

PR #11870, opened 2019-04-16
correct var is pattern

### LINKED ISSUE #11804: Incorrect description of `is var` behavior
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/is#-var-pattern- states the following:

>  A pattern match with the var pattern always succeeds for non-null expressions; if expr is null, the is expression is false. The non-null value of expr is always assigned to a local variable the same type as the runtime time type of expr. Its syntax is:
> 
>  ``` cs
>  expr is var varname
>  ```
> 
>  The following example uses the var pattern to assign an expression to a variable named obj. It then displays the value and the type of obj.

It then provides an example, of which the key component is the following:

``` cs
object[] items = { new Book("The Tempest"), new Person("John") };
foreach (var item in items) {
if (item is var obj)
  Console.WriteLine($"Type: {obj.GetType().Name}, Value: {obj}"); 
}
```

This very clearly and unambiguously indicates that `item is var obj` will return `false` if `item` is null.  This assertion is incorrect, and there are multiple references, including on https://github.com/dotnet/csharplang, stating that `is var` will always return true, regardless of nullability.

This error seems to be the cause of some confusion:

* https://stackoverflow.com/questions/45844586/usage-of-var-pattern-in-c-sharp-7/55646164
* https://github.com/dotnet/csharplang/issues/981

There was a proposal to add this behavior, but it was declined: https://github.com/dotnet/csharplang/issues/792

---
#### Document Details

* ID: 3126aaa9-8f5f-04f3-2d75-96477b6273cd
* Version Independent ID: 3f4b3eac-11eb-2dd6-4f27-b24537367288
* Content: [is - C# Reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/is#feedback)
* Content Source: [docs/csharp/language-reference/keywords/is.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/keywords/is.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixes #11804

### DIFF
--- a/docs/csharp/language-reference/keywords/is.md
+++ b/docs/csharp/language-reference/keywords/is.md
@@ -139,7 +139,7 @@ The following example shows a comparison of `null` checks:
  
 ### <a name="var" /> var pattern </a>
 
-A pattern match with the var pattern always succeeds for non-null expressions; if *expr* is `null`, the `is` expression is `false`. The non-null value of *expr* is always assigned to a local variable the same type as the runtime time type of *expr*.  Its syntax is:
+The `var` pattern is a catch-all for any type or value. The value of *expr* is always assigned to a local variable the same type as the compile time type of *expr*. The result of the `is` expression is always `true`. Its syntax is:
 
 ```csharp 
    expr is var varname


Thraka (approved):
What is the point of this pattern? I saw the benefit as being the return of true/false based on if it was assignable. 😄 so now I have no idea why this pattern exists.

BillWagner:
@Thraka It gets a lot more useful in C# 8 when you have the property patterns: you can assign a variable to any property of an object.

Zenexer:
It’s also useful in cases like this:

``` cs
bool Test(SomeType x)
    => x.CalculateSomething() is var y && y.Value1 < y.Value2;
```

PR #11886, opened 2019-04-17
Moved true and false operators to the unary operators section

pkulikov:
Continuation of #11860

My guess is that, when grouped by precedence, the `true` and `false` operators should be in the group of unary operators.

@BillWagner is it a correct guess?

The spec doesn't say anything about precedence of those two operators:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#operator-precedence-and-associativity

Moreover, that silence is [intentional](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#operator-overloading) (emphasis mine):

> Although `true` and `false` are not used explicitly in expressions **(and therefore are not included in the precedence table in Operator precedence and associativity)**, they are considered operators because they are invoked in several expression contexts: boolean expressions (Boolean expressions) and expressions involving the conditional (Conditional operator), and conditional logical operators (Conditional logical operators).

/cc: @paulomorgado

### DIFF
--- a/docs/csharp/language-reference/operators/index.md
+++ b/docs/csharp/language-reference/operators/index.md
@@ -85,6 +85,10 @@ These operators have higher precedence than the next section and lower precedenc
 
 [*x](multiplication-operator.md) – dereferencing.
 
+[true operator](../keywords/true-false-operators.md) - returns the [bool](../keywords/bool.md) value `true` to indicate that an operand is definitely true.
+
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks for doing the research and making this change @pkulikov 

I cross-checked the C# 6.0 draft with the ECMA 5.0 standard, and the clauses you cite are the same in both versions. (We haven't updated the 6.0 draft with the work the standards committee did for 5.0 yet, so there are some differences.)

This is ready to :shipit:

PR #11898, opened 2019-04-18
Arithmetic operators: mention possible casting with compound assignment

pkulikov:
As an operation performed as it is and the same operation performed as a part of compound assignment might end up with "inconsistent" results, I think it's worth to mention possible casting involved in compound assignment with arithmetic operators.

Supported by dotnet/samples#825

Also, added `checked` and `unchecked` to the `displayName` field in the TOC entry for arithmetic operators


BillWagner:
closing and reopening for a new build while I review.

BillWagner (approved):
Thanks for these additions @pkulikov It's an important clarification.

I'll :shipit: now.

PR #11775, opened 2019-04-10
Update charset.md for changes in .NET Core 3.0

jkoritzinsky:
In .NET Core 3.0, we have changed the meaning of `CharSet.Auto` to match Mono's behavior cross-platform. This PR updates the docs to reflect that behavior change.

PR #11934, opened 2019-04-19
Update charset.md

nxtn:
Contributes to #11396 
Continued from #11775 

It should be **.NET Core on Unix** instead of **Unix on .NET Core**.

### DIFF
--- a/docs/standard/native-interop/charset.md
+++ b/docs/standard/native-interop/charset.md
@@ -12,10 +12,10 @@ The way `char` values, `string` objects, and `System.Text.StringBuilder` objects
 
 The following table shows a mapping between each charset and how a character or string is represented when marshalled with that charset:
 
-| CharSet | Windows            | Unix on .NET Core 2.2 and earlier | Mono and Unix on .NET Core 3.0 and later |
-|---------|--------------------|-----------------------------|------------------------------------------|
-| Ansi    | `char` (ANSI)      | `char` (UTF-8)              | `char` (UTF-8)                           |
-| Unicode | `wchar_t` (UTF-16) | `char16_t` (UTF-16)         | `char16_t` (UTF-16)                      |
-| Auto    | `wchar_t` (UTF-16) | `char16_t` (UTF-16)         | `char` (UTF-8)                           |
+| `CharSet` value | Windows            | .NET Core 2.2 and earlier on Unix | .NET Core 3.0 and later and Mono on Unix |
+|-----------------|--------------------|-----------------------------------|------------------------------------------|
+| Ansi            | `char` (ANSI)      | `char` (UTF-8)                    | `char` (UTF-8)                           |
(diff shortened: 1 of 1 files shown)


docs/standard/native-interop/charset.md
```diff
@@ -12,10 +12,10 @@ The way `char` values, `string` objects, and `System.Text.StringBuilder` objects
 
 The following table shows a mapping between each charset and how a character or string is represented when marshalled with that charset:
 
-| CharSet | Windows            | Unix on .NET Core 2.2 and earlier | Mono and Unix on .NET Core 3.0 and later |
-|---------|--------------------|-----------------------------|------------------------------------------|
-| Ansi    | `char` (ANSI)      | `char` (UTF-8)              | `char` (UTF-8)                           |
-| Unicode | `wchar_t` (UTF-16) | `char16_t` (UTF-16)         | `char16_t` (UTF-16)                      |
-| Auto    | `wchar_t` (UTF-16) | `char16_t` (UTF-16)         | `char` (UTF-8)                           |
+| `CharSet` | Windows            | .NET Core 2.2 and earlier on Unix | .NET Core 3.0 and later and Mono on Unix |
```
rpetrusha: I think that the second column should read "Mono and .NET Core 3.0 and later on Unix"; Mono doesn't exclusively run on Unix systems.

rpetrusha (changes requested):
Thanks for correcting these, @NextTurn. I've left one comment for you to consider.

docs/standard/native-interop/charset.md
NextTurn: I don't think Mono on Windows behaves the same as Mono on Unix.

/cc @jkoritzinsky

jkoritzinsky: Mono follows .NET Framework and .NET Core on Windows. That’s why I named that column “Windows”. If you have a clearer way of phrasing it, I’m all for it.

NextTurn: So the last column applies to .NET Core 3.0 and later on Unix, and Mono on Unix. Mono should be put after .NET Core 3.0, otherwise it seems to be Mono 3.0 and .NET Core 3.0?

jkoritzinsky: Sounds good to me!

NextTurn:
BTW does `MarshalAs.LPStr/LPWStr/LPTStr` behaves the same as `Charset.Ansi/Unicode/Auto`?

jkoritzinsky:
LPTStr is still Unicode on all platforms. Main reasoning there is because it’s not tied to the CharSet setting and the native LPTSTR type is a Windows-only concept. And for all interop scenarios that it’s supported on ( PInvoke only) the default behavior is marshaling as LPSTR or LPWSTR based on CharSet setting.

rpetrusha (approved):

rpetrusha:
I apologize for my delay in approving this PR, @NextTurn. I'll merge it now. The changes should be live on docs.microsoft.com next Monday.

PR #11935, opened 2019-04-19
Fixed FWLinks for 4.8 devpack and offline installer

rpetrusha:
## Fixed FWLinks for 4.8 devpack and offline installer

//cc @NKarnam15


mairaw (approved):
I don't know if these links are correct or not though 😝

PR #11969, opened 2019-04-22
Add missing "Add the following code ..."

tdykstra:
In the tuples section, the first code block that says "Add the following code to your browser to try it" depends on the previous one, but the previous one didn't say to add it to the browser.

Added a couple of typo fixes.


BillWagner (approved):
Thanks for making this change @tdykstra 

PR #11983, opened 2019-04-23
Add note about nullable value types

### LINKED ISSUE #11834: What about Nullable struct?
Should there be some information here on whether to use or not the in-modifier with Nullable struct?
For example should we have
public void Method(int? arg) {}
or
public void Method(in int? arg) {}
?
ErrorProne.NET for example flags this as a special case: https://github.com/SergeyTeplyakov/ErrorProne.NET/blob/5cc962abfaf0005890dd0a9fba46a9e304a8a450/src/ErrorProne.NET.StructAnalyzers/TypeExtensions.cs#L28

---
#### Document Details

* Content: [Write safe and efficient C# code](https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#feedback)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixes #11834

### DIFF
--- a/docs/csharp/write-safe-efficient-code.md
+++ b/docs/csharp/write-safe-efficient-code.md
@@ -233,6 +233,8 @@ The compiler generates more efficient code when you call members of a
 is always an `in` parameter passed by reference to the member method. This optimization
 saves copying when you use a `readonly struct` as an `in` argument.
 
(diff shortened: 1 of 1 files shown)


rpetrusha (approved):
LGTM, @BillWagner. I've left a couple of nit-picking suggestions.

PR #11985, opened 2019-04-23
Move P/Invokes to NativeMethods class

nxtn:
Contributes to #11396 

In .NET Framework Guide.

[CA1060: Move P/Invokes to NativeMethods class](https://docs.microsoft.com/visualstudio/code-quality/ca1060-move-p-invokes-to-nativemethods-class)

### DIFF
--- a/docs/framework/interop/creating-prototypes-in-managed-code.md
+++ b/docs/framework/interop/creating-prototypes-in-managed-code.md
@@ -43,7 +43,7 @@ Friend Class NativeMethods
 End Class
 ```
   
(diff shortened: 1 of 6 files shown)


rpetrusha (approved):
I apologize that I've allowed this PR to languish, @NextTurn. I've made a few additional changes in the VB code -- I've eliminated unnecessary Import statements, and I found that the Static keyword is not valid in a Declare statement, so I've removed them. I'll merge now.

NextTurn:
`System` namespace is for `IntPtr`.

rpetrusha:
The System namespace only needs to be imported in VB code in early versions of the .NET Framework; it's imported automatically starting in .NET Framework 2.0.

NextTurn:
It seems that `Microsoft.VisualBasic` is also automatically imported? I can't find the information in docs. Shall we remove these Imports statements in other snippets?

rpetrusha:
Yes, @NextTurn, both System and Microsoft.VisualBasic are automatically imported, so the `Imports` statements for them (and actually should be) removed.

PR #11995, opened 2019-04-24
BOM required to override computer codepage

breyed:
This page said that **-codepage** is not needed for source code files created with UNICODE or UTF-8. If that's true, then if I have a .cs file encoded in UTF-8 with no Byte Order Mark, csc.exe will interpret it as UTF-8 regardless of the codepage in effect on the computer. That doesn't seem possible, however, since csc.exe can't know whether the file is encoded as UTF-8 or the computer's codepage.

This change corrects the text to indicate that if the source file indicates that uses a Unicode encoding by including a byte order mark, then the computer's codepage is ignored.

### DIFF
--- a/docs/csharp/language-reference/compiler-options/codepage-compiler-option.md
+++ b/docs/csharp/language-reference/compiler-options/codepage-compiler-option.md
@@ -23,10 +23,8 @@ This option specifies which codepage to use during compilation if the required p
  The id of the code page to use for all source code files in the compilation.  
   
 ## Remarks  
- If you compile one or more source code files that were not created to use the default code page on your computer, you can use the **-codepage** option to specify which code page should be used. **-codepage** applies to all source code files in your compilation.  
-  
- If the source code files were created with the same codepage that is in effect on your computer or if the source code files are encoded as Unicode (e.g. UTF-8) with a Byte Order Mark (BOM), you need not use **-codepage**.  
-  
+ The compiler will first attempt to intrept all source files as UTF-8. If your source code files are in an encoding other than UTF-8 and use characters other than 7-bit ASCII characters, use the **-codepage** option to specify which code page should be used. **-codepage** applies to all source code files in your compilation.  
+    
  See [GetCPInfo](/windows/desktop/api/winnls/nf-winnls-getcpinfo) for information on how to find which code pages are supported on your system.  
   
  This compiler option is unavailable in Visual Studio and cannot be changed programmatically.  
--- a/docs/csharp/language-reference/compiler-options/codepage-compiler-option.md
+++ b/docs/csharp/language-reference/compiler-options/codepage-compiler-option.md
@@ -25,7 +25,7 @@ This option specifies which codepage to use during compilation if the required p
 ## Remarks  
  If you compile one or more source code files that were not created to use the default code page on your computer, you can use the **-codepage** option to specify which code page should be used. **-codepage** applies to all source code files in your compilation.  
   
- If the source code files were created with the same codepage that is in effect on your computer or if the source code files were created with UNICODE or UTF-8, you need not use **-codepage**.  
+ If the source code files were created with the same codepage that is in effect on your computer or if the source code files are encoded as Unicode (e.g. UTF-8) with a Byte Order Mark (BOM), you need not use **-codepage**.  
   
  See [GetCPInfo](/windows/desktop/api/winnls/nf-winnls-getcpinfo) for information on how to find which code pages are supported on your system.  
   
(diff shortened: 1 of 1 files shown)


BillWagner:
Ping @jcouv 

This looks correct to me. I'd like someone on the compiler team to verify the technical accuracy before I merge.

jcouv:
Tagging @gafter who may be able to confirm on top of his head. 
Otherwise, I'll dig in the code to confirm.

gafter:
I don't know.

BillWagner:
@jaredpar Can you route this to the right person to check for accuracy?

jaredpar:
The summary of the method used by the compiler here is: it's complicated.

Here is the basic order in which we use enocdings:

1. If an encoding is specified on the command line then use that and no other. 
1. Attempt to decode with the UTF8 encoding. If that fails the use the fallback enocding

The fallback encoding is determined by looking for the following code pages, via `Enocding.GetEncoding`, in the following order:

- 0: the default ANSI code page for the OS
- 1252: historical reasons
- Latin1

BillWagner:
Based on @jaredpar 's comment, I'll close this. The chance is not accurate for how the compiler determines the encoding.

breyed:
The docs still need to be updated. Based on how the compiler works per @jaredpar's description, there are two errors in the docs as they stand:

1. The "UNICODE or" part of "if the source code files were created with UNICODE or UTF-8" is incorrect. The meaning of "UNICODE" is ambiguous, but given the dated capitalization and usage as something that's in addition to UTF-8, it comes across as meaning UCS-2. However, the C# compiler doesn't check for UTF-8 and UCS-2 (or any other 2-byte character encoding) before using the fallback encoding; it only checks for UTF-8. "UNICODE or" should be removed.

2. The docs say that you don't need **-codepage** if "the source code files were created with the same codepage that is in effect on your computer". That's not correct. For example, if your source file was created in code page 1252 and your running the compiler with that same code page in effect, and if the file contains the string "Â¢", the C# compiler would try to first interpret the file as UTF-8, which it would be able to do, but when it sees the byes C2 A2 for the "Â¢", it will interpret it as valid a UTF-8 "¢". The result is that the "Â" will be lost.

I updated the PR with guidance that fixes the issues.

Thanks also for answering my question that got me started on this whole thing with the clarification that BOMs are not required for UTF-8 files.

tiesmaster:
@jaredpar Awesome write-up, and investigation. Where does this happen in the compiler?

BillWagner:
@breyed I don't see any updates on this PR.  Am I missing something?

breyed:
@BillWagner I had added another [commit](https://github.com/breyed/core-docs/commit/7e4c78c500c2aab1a823ed661cbef63d5167e22e) to the branch that was created for the PR. It looks like the PR didn't pick up the new commit, but instead is pinned to the first commit. After a quick web search, it wasn't clear how to get the PR to float to the latest commit. Please take a look and let me know if you have any suggestions.

BillWagner:
@breyed Aha!  I had to reopen the PR to see the new commit. (I didn't know that).  

Reviewing now.

docs/csharp/language-reference/compiler-options/codepage-compiler-option.md
```diff
@@ -23,10 +23,8 @@ This option specifies which codepage to use during compilation if the required p
  The id of the code page to use for all source code files in the compilation.  
   
 ## Remarks  
- If you compile one or more source code files that were not created to use the default code page on your computer, you can use the **-codepage** option to specify which code page should be used. **-codepage** applies to all source code files in your compilation.  
-  
- If the source code files were created with the same codepage that is in effect on your computer or if the source code files were created with UNICODE or UTF-8, you need not use **-codepage**.  
-  
+ The compiler will first attempt to intrept all source files as UTF-8. If your source code files are in an encoding other than UTF-8 and use characters other than 7-bit ASCII characters, use the **-codepage** option to specify which code page should be used. **-codepage** applies to all source code files in your compilation.
```
BillWagner:
```suggestion
 The compiler will first attempt to interpret all source files as UTF-8. If your source code files are in an encoding other than UTF-8 and use characters other than 7-bit ASCII characters, use the **-codepage** option to specify which code page should be used. **-codepage** applies to all source code files in your compilation.  
```

BillWagner (changes requested):
Thanks for making this update @breyed 

I noticed once spelling mistake, and have a suggestion to fix it.  Once that's added, I'll :shipit:

BillWagner (approved):

PR #12009, opened 2019-04-24
fix: MD007/ul-indent

nschonni:
- Unordered list indentation
- Trim leading/trailing space
- Removed "To Top links"
- Removed Manual TOC
- Use unorded lists for some implied lists
- Convert inline to code fences


mairaw (approved):
Thanks @nschonni for all this cleanup. Looks good!

PR #12028, opened 2019-04-25
fix: MD030/list-marker-space

nschonni:
Split from #12024


BillWagner (approved):
A spot check, followed by checking the diffs with hide whitespace, and this LGTM @nschonni 

I'll :shipit: now.

PR #12032, opened 2019-04-25
Removing .NET Framework docker content

### LINKED ISSUE #8905: [rewrite] .NET Framework docker article to focus more on docker than making an app
SkyRay reports this article has a decent amount of hits, but it's the only Docker article for .NET Framework. According to feedback and some negative verbatim, it's too verbose and doesn't get to the point of the article fast enough. Put a .NET Framework console app in a container and run it.

~I suggest rewriting this article to get it down in size and more to the point.~
https://docs.microsoft.com/en-us/dotnet/framework/docker/console

After some discussions we've decided to retire this content. .NET Framework console app docker content is not an area we want to invest in. There are tutorials hosted on other sites that cover this scenario. We suggest that people move to using .NET Core with Docker, which is a scenario we cover.

### SkyRay

|||
|-|-|
| Page views (30 days) | 1,777 |
| Ratings | 0 |

### Related issues

#8540 #8619

Thraka:
## Summary

After some discussions we've decided to retire this content. .NET Framework console app docker content is not an area we want to invest in. There are tutorials hosted on other sites that cover this scenario. We suggest that people move to using .NET Core with Docker, which is a scenario we cover.

Fixes #8905


.openpublishing.redirection.json
```diff
@@ -997,6 +997,14 @@
             "redirect_url": "/aspnet/mvc/overview/deployment/docker-aspnetmvc",
             "redirect_document_id": true
         },
+        {
+            "source_path": "docs/framework/docker/console.md",
```
BillWagner: Instead of sending redirects to the framework guide, maybe redirect to our .NET Core Docker content: `/dotnet/core/docker/intro-net-docker`

.openpublishing.redirection.json
```diff
@@ -997,6 +997,14 @@
...
+            "redirect_url": "/dotnet/framework"
+        },
+        {
+            "source_path": "docs/framework/docker/index.md",
+            "redirect_url": "/dotnet/framework"
```
BillWagner: same comment as above.

BillWagner (approved):
This LGTM @Thraka 

I had a possible alternative destination for the redirects. Choose what you think is best.

.openpublishing.redirection.json
Thraka: Yeah, I was thinking about that too. I wasn't sure if that would just be more confusing because they would think something was wrong, they wanted framework not core... It's too bad we really do not have a "Sorry this content has been removed" type of link we can send people too besides leaving a 404.

PR #12039, opened 2019-04-25
fix: MD030/list-marker-space (docs/csharp)

nschonni:
Spaces after list markers for unordered lists
Find: `^( *)\-   `
Replace: `$1- `
Filter: *.md


BillWagner (approved):
This also LGTM after a whitespace diff and a spot check on the review site and the regular diff.

I'll :shipit: now.

PR #12045, opened 2019-04-26
fix: MD029/ol-prefix

nschonni:
Ordered list item prefix had a gap, swapped back to 1.s


BillWagner (approved):
This LGTM @nschonni 

I'll :shipit: now.

Thanks again for all the work to help us write cleaner markdown.

PR #12072, opened 2019-04-27
Added spec information about string interpolation

pkulikov:
- Added the link to the spec
- Moved the description of the compile-time behavior from the tutorial to the feature article
- After that, realized that not all cases are covered, so added the paragraph about the behavior in case of `IFormattable` or `FormattableString`


BillWagner (approved):
Thanks for creating this @pkulikov This looks great.

I'll :shipit: now.

These will go live next week, during Build.

PR #12092, opened 2019-04-29
Typo fix

pkulikov:
Based on the Address class next entry, I conclude that only class name should be bolded


mairaw (approved):
Thanks @pkulikov! Publishing is on hold until next week. So changes will go live just then.

PR #12137, opened 2019-05-01
Fix missing opening angle bracket

jo-ninja:
## Summary

Fix missing opening angle bracket


rpetrusha (approved):
Thanks, @jo-ninja, for correcting the badly formed xref. We'll merge your PR now. The change should be live early next week.

PR #12138, opened 2019-05-01
Add missing word "of" in sentence

vanillajonathan:
Add missing word "of" between "instead" and "exceptions".


rpetrusha (approved):
Thank you, @vanillajonathan, for adding the missing word. We'll merge your PR now. The change will be live on docs.microsoft.com early next week.

PR #12182, opened 2019-05-04
Fix spelling

deisterhold:
## Summary

Change pubish to publish.


rpetrusha (approved):
Thanks, @deisterhold, for correcting this typo. We'll merge your PR a bit later, and the correction should be live on docs.microsoft.com in the next day or two.

PR #12203, opened 2019-05-06
Fixed grammar and simplified language

pkulikov:
Original version was: 
> substitute `string.Format` for `string.Concat`

That was incorrect, because it is `string.Format` that is replaced with `string.Concat`. So, the correct version would be
>substitute `string.Concat` for `string.Format`

However, I think using "replace" verb makes sentence easier to understand.


rpetrusha (approved):
Thank you, @pkulikov, for making this change. I'll merge your PR now.

PR #12220, opened 2019-05-07
Improvements to definitions of metrics.md

justinormont:
## Summary

Few inaccuracies cleaned up, for instance NDCG is 0.00 to 1.00; was 0 to 100 previously and was changed in https://github.com/dotnet/machinelearning/pull/2697 (and discussed in https://github.com/dotnet/machinelearning/issues/2637). Expanded some guidelines, like when accuracy is non-useful.

Could still use an additional pass at some later point. Perhaps when we add in the missing definitions of ranking/clustering/etc metrics.

Also partially addresses: https://github.com/dotnet/machinelearning/issues/3671


docs/machine-learning/resources/metrics.md
```diff
@@ -10,10 +10,10 @@ author:
...
+| **Accuracy** |  [Accuracy](https://en.wikipedia.org/wiki/Accuracy_and_precision#In_binary_classification) is the proportion of correct predictions with a test data set. It is the ratio of number of correct predictions to the total number of input samples. It works well only if there are similar number of samples belonging to each class.| **The closer to 1.00, the better**. But exactly 1.00 indicates an issue (commonly: label/target leakage, over-fitting, or testing with training data). When the test data is unbalanced (where most of the instances belong to one of the classes), the dataset is very small, or scores approach 0.00 or 1.00, then accuracy doesn’t really capture the effectiveness of a classifier and you need to check additional metrics. |
+| **AUC** |    [aucROC](https://en.wikipedia.org/wiki/Receiver_operating_characteristic) or *Area under the ROC curve*: This is measuring the area under the curve created as you sweep the true positive rate vs. the false positive rate.  |   **The closer to 1.00, the better**. It should be greater than 0.50 for a model to be acceptable; a model with AUC of 0.50 or less is worthless. |
```
JRAlexander:
```suggestion
| **AUC** |    [aucROC](https://en.wikipedia.org/wiki/Receiver_operating_characteristic) or *Area under the 
ROC curve*: This is measuring the area under the curve created by sweeping the true positive rate vs. the false positive rate.  |   **The closer to 1.00, the better**. It should be greater than 0.50 for a model to be acceptable; a model with AUC of 0.50 or less is worthless. |
```

JRAlexander: Would this work?

docs/machine-learning/resources/metrics.md
```diff
@@ -10,10 +10,10 @@ author:
...
-| **F1-score** | [F1 score](https://en.wikipedia.org/wiki/F1_score) also known as *balanced F-score or F-measure*. It can be interpreted as a weighted average of the precision and recall. F1 Score is needed when you want to seek a balance between Precision and Recall.| **The closer to 1.00, the better**.  An F1 score reaches its best value at 1 and worst score at 0. It tells you how precise your classifier is. |
+| **F1-score** | [F1 score](https://en.wikipedia.org/wiki/F1_score) also known as *balanced F-score or F-measure*. It is the harmonic mean of the precision and recall. F1 Score is helpful when you want to seek a balance between Precision and Recall.| **The closer to 1.00, the better**.  An F1 score reaches its best value at 1.00 and worst score at 0.00. It tells you how precise your classifier is. |
```
JRAlexander:
```suggestion
| **F1-score** | [F1 score](https://en.wikipedia.org/wiki/F1_score) also known as *balanced F-score or F-measure*. It's the harmonic mean of the precision and recall. F1 Score is helpful when you want to seek a balance between Precision and Recall.| **The closer to 1.00, the better**.  An F1 score reaches its best value at 1.00 and worst score at 0.00. It tells you how precise your classifier is. |
```

JRAlexander (changes requested):
Looks good. Just a couple small things.

docs/machine-learning/resources/metrics.md
justinormont: Yep. Thanks.

JRAlexander (approved):

JRAlexander:
Nice work, @justinormont!

PR #12232, opened 2019-05-07
Clarified introduction about the abstract keyword

svick:
1. I think it makes sense to explain what `abstract` classes can't do, because otherwise the sentence could be interpreted as implying that e.g. you can't use abstract classes as variable types.
2. Members included in an `abstract` class that are not marked as `abstract` do not have to be overriden in derived classes.


BillWagner (approved):
Thanks for these clarifying remarks @svick 

I've reviewed this and I'll :shipit: now.

Thanks again for your continued contributions.

PR #12240, opened 2019-05-07
Added overview of C# member access operators

### LINKED ISSUE #11118: Abridged Reference: Operators / Member access
The member access operators are: 

- `.` (member access)
- `[]` (array or indexer access)
- `?.` and `?[]` (null-conditional access)
- `()` (method call)

Sub-tasks are:

- [ ] Create the new articles
- [ ] Remove old articles.

pkulikov:
Fixes #11118
Supported by dotnet/samples#886


BillWagner (changes requested):
Super great work @pkulikov 

I had one item for discussion.  I'm marking as "request changes" so it stays in my review queue. Checking the build status and review site next.

BillWagner:
Here's the [internal review link](https://review.docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators?branch=pr-en-us-12240)

docs/csharp/language-reference/operators/member-access-operators.md
```diff
@@ -0,0 +1,163 @@
...
+```csharp
+[System.Diagnostics.Conditional("DEBUG")]
+void TraceMethod() {}
+```
+
+## Null-conditional operators ?. and ?[]
+
+Available in C# 6 and later, a null-conditional operator applies a member access, `?.`, or element access, `?[]`, operation to its operand only if that operand evaluates to non-null. If the operand evaluates to `null`, the result of applying the operator is `null`.
+
+The null-conditional operators are short-circuiting. That is, if one operation in a chain of conditional member or element access operations returns `null`, the rest of the chain doesn't execute. In the following example, `E` doesn't execute if `A`, `B`, or `C` evaluates to `null`:
```
pkulikov: @BillWagner agree with you; I've made the chain shorter and updated the wording. Please take a look at the updated text.

pkulikov:
@BillWagner thanks for review! I've addressed your feedback.

(Strangely, your review comments are timestamped into the future (at least on my side))

BillWagner (approved):

BillWagner:
Thanks again @pkulikov 

I'll :shipit: now.

docs/csharp/language-reference/operators/member-access-operators.md
BillWagner: I think it is worth adding that if `B` is `null`, `C` is not evaluated. (as an example of the short circuiting).

docs/csharp/language-reference/operators/member-access-operators.md
```diff
@@ -0,0 +1,163 @@
...
+A?.B?.C?[E];
+```
+
+The following example demonstrates the usage of the `?.` and `?[]` operators:
+
+[!code-csharp-interactive[null-conditional operators](~/samples/snippets/csharp/language-reference/operators/MemberAccessOperators.cs#NullConditional)]
+
+The preceding example also shows the usage of the [null-coalescing operator](null-coalescing-operator.md). You might use the null-coalescing operator to provide an alternative expression to evaluate in case the result of the null-conditional operation is `null`.
+
+### Thread-safe delegate invocation
```
BillWagner: I'm so glad you added this example.

PR #12241, opened 2019-05-07
Create initial tutorial for Default Interface Members

### LINKED ISSUE #11604: Tutorial: Default Interface members
This tutorial will explore scenarios where interfaces can be versioned and updated using default interface members to provide compatibility with existing implementors.

BillWagner:
Fixes #11604 

This relies on the samples in dotnet/samples#888

According to one of our site engineers, highlighting code using named snippets is currently supported. I'm testing that in this PR.

[internal review link](https://review.docs.microsoft.com/en-us/dotnet/csharp/tutorials/default-interface-members-versions?branch=pr-en-us-12241)

### DIFF
--- a/docs/csharp/tutorials/default-interface-members-versions.md
+++ b/docs/csharp/tutorials/default-interface-members-versions.md
@@ -21,7 +21,7 @@ You’ll need to set up your machine to run .NET Core, including the C# 8.0 prev
 
 ## Scenario overview
 
-This tutorial starts with version 1 of a customer relationship library. The company that built this library intended customers with existing applications to adopt their library. They provided minimal interface definitions for users of their library to implement. They defined an interface with methods for a customer:
+This tutorial starts with version 1 of a customer relationship library. The company that built this library intended customers with existing applications to adopt their library. They provided minimal interface definitions for users of their library to implement. Here's the interface definition for a customer:
 
 [!code-csharp[InitialCustomerInterface](~/samples/csharp/tutorials/default-interface-members-versions/starter/customer-relationship/ICustomer.cs?name=SnippetICustomerVersion1)]
 
@@ -33,7 +33,7 @@ From those interfaces, the team could build a library for their users to create 
 
 Now, it's time to upgrade the library for the next release. One of the requested features enables a loyalty discount for customers that have lots of orders. This new loyalty discount gets applied whenever a customer makes an order. The specific discount is a property of each individual customer. Each implementation of ICustomer can set different rules for the loyalty discount. 
 
-The most natural way to add this functionality is to enhance the `ICustomer` interface with a method to apply any loyalty discount. This design suggestion caused concern among experienced developers: "Interfaces are immutable once they've been released! This is a breaking change!" C# 8.0 adds *default interface implementations* for scenarios like these. The library authors can add new members to the interface and provide a default implementation for those members.
+The most natural way to add this functionality is to enhance the `ICustomer` interface with a method to apply any loyalty discount. This design suggestion caused concern among experienced developers: "Interfaces are immutable once they've been released! This is a breaking change!" C# 8.0 adds *default interface implementations* for upgrading interfaces. The library authors can add new members to the interface and provide a default implementation for those members.
 
 Default interface implementations enable developers to upgrade an interface while still enabling any implementors to override that implementation. Users of the library can accept the default implementation as a non-breaking change. If their business rules are different, they can override.
 
@@ -51,7 +51,7 @@ The library author wrote a first test to check the implementation:
 
 [!code-csharp[TestDefaultImplementation](~/samples/csharp/tutorials/default-interface-members-versions/finished/customer-relationship/Program.cs?name=SnippetTestDefaultImplementation&highlight=SnippetHighlightCast)]
 
-Notice the highlighted portion of the test. That cast from `SampleCustomer` to `ICustomer` is necessary. The `SampleCustomer` class does not need to provide an implementation for `ComputeLoyaltyDiscount`; that's provided by the `ICustomer` interface. However, the `SampleCustomer` class does not inherit members from its interfaces. That hasn't changed. In order to call any method declared and implemented in the interface, the variable must be the type of the interface, `ICustomer` in this example.
+Notice the highlighted portion of the test. That cast from `SampleCustomer` to `ICustomer` is necessary. The `SampleCustomer` class doesn't need to provide an implementation for `ComputeLoyaltyDiscount`; that's provided by the `ICustomer` interface. However, the `SampleCustomer` class doesn't inherit members from its interfaces. That rule hasn't changed. In order to call any method declared and implemented in the interface, the variable must be the type of the interface, `ICustomer` in this example.
 
 ## Provide parameterization
 
@@ -59,7 +59,7 @@ That's a good start. But, the default implementation is too restrictive. Many co
 
 [!code-csharp[VersionTwoImplementation](~/samples/csharp/tutorials/default-interface-members-versions/finished/customer-relationship/ICustomer.cs?name=SnippetLoyaltyDiscountVersionTwo)]
 
-There's a lot of new capabilities shown in that small code fragment. Interfaces can now include static members, including fields and methods. Different access modifiers are also enabled. In the previous example, the additional fields are private, and the new method is public. Any of the modifiers are allowed on interface members.
+There's many new capabilities shown in that small code fragment. Interfaces can now include static members, including fields and methods. Different access modifiers are also enabled. In the previous example, the additional fields are private, and the new method is public. Any of the modifiers are allowed on interface members.
 
 Applications that use the general formula for computing the loyalty discount, but different parameters don't need to provide a custom implementation; they can set the arguments through a static method. For example, the following code sets a "customer appreciation" that rewards any customer with more than one month's membership:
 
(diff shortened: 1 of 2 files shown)


BillWagner:
closing and opening to get a new build.

BillWagner:
*sigh* one more fresh build.

docs/csharp/tutorials/default-interface-members-versions.md
```diff
@@ -0,0 +1,80 @@
+---
+title: Safely update interfaces using default interface members in C#
+description: This advanced tutorial explores how you can safely add new capabilities to existing interface definitions without breaking all classes and structs that implement that interface.
+ms.date: 05/06/2019
+ms.custom: mvc
+---
+# Tutorial: Learn how to use default interface members in C# 8 to safely update interface definitions to add functionality to all implementations
```
Thraka: This renders HUGE. I think it's way too long hehehe. Something like this perhaps?

```suggestion
# Tutorial: Update interfaces with default interface members in C# 8
```

docs/csharp/tutorials/default-interface-members-versions.md
```diff
@@ -0,0 +1,80 @@
...
+
+The most natural way to add this functionality is to enhance the `ICustomer` interface with a method to apply any loyalty discount. This design suggestion caused concern among experienced developers: "Interfaces are immutable once they've been released! This is a breaking change!" C# 8.0 adds *default interface implementations* for upgrading interfaces. The library authors can add new members to the interface and provide a default implementation for those members.
+
+Default interface implementations enable developers to upgrade an interface while still enabling any implementors to override that implementation. Users of the library can accept the default implementation as a non-breaking change. If their business rules are different, they can override.
+
+## Upgrade with default interface members
+
+The team agreed on the most likely default implementation: a loyalty discount for customers
```
Thraka:
```suggestion
The team agreed on the most likely default implementation: a loyalty discount for customers.
```

docs/csharp/tutorials/default-interface-members-versions.md
```diff
@@ -0,0 +1,80 @@
...
+
+Default interface implementations enable developers to upgrade an interface while still enabling any implementors to override that implementation. Users of the library can accept the default implementation as a non-breaking change. If their business rules are different, they can override.
+
+## Upgrade with default interface members
+
+The team agreed on the most likely default implementation: a loyalty discount for customers 
+
+The upgrade should provide the functionality to set two properties: the number of orders needed to be eligible for the discount, and the percentage of the discount. This makes it a perfect scenario for default interface members. You can add a method to the ICustomer interface, and provide the most likely implementation. All existing, and any new implementations can use the default implementation, or provide their own.
```
Thraka: I feel like in this scenario the default implementation would be to provide 0 discount. This feature is described as an opt-in feature where users can provide loyalty discounts. I can see where the creators of the library provide the default implementation that does nothing, so as to not break compatibility and enable users to provide their own discount logic. So the first code example of the implementation would end up being `public decimal ComputeLoyaltyDiscount() => 0;`

docs/csharp/tutorials/default-interface-members-versions.md
```diff
@@ -0,0 +1,80 @@
...
+
+First, add the new method to the implementation:
+
+[!code-csharp[InitialOrderInterface](~/samples/csharp/tutorials/default-interface-members-versions/finished/customer-relationship/ICustomer.cs?name=SnippetLoyaltyDiscountVersionOne)]
+
+The library author wrote a first test to check the implementation:
+
+[!code-csharp[TestDefaultImplementation](~/samples/csharp/tutorials/default-interface-members-versions/finished/customer-relationship/Program.cs?name=SnippetTestDefaultImplementation&highlight=SnippetHighlightCast)]
+
+Notice the highlighted portion of the test. That cast from `SampleCustomer` to `ICustomer` is necessary. The `SampleCustomer` class doesn't need to provide an implementation for `ComputeLoyaltyDiscount`; that's provided by the `ICustomer` interface. However, the `SampleCustomer` class doesn't inherit members from its interfaces. That rule hasn't changed. In order to call any method declared and implemented in the interface, the variable must be the type of the interface, `ICustomer` in this example.
```
Thraka: Perhaps that feature isn't working? I don't see the highlighted portion. I think the ASP.NET guys use this a lot, maybe they can troubleshoot?

docs/csharp/tutorials/default-interface-members-versions.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Provide parameterization
+
+That's a good start. But, the default implementation is too restrictive. Many consumers of this system may choose different thresholds for number of purchases, a different length of membership, or a different percentage discount. You can provide a better upgrade experience for more customers by providing a way to set those parameters. Let's add a static method that sets those three parameters controlling the default implementation:
+
+[!code-csharp[VersionTwoImplementation](~/samples/csharp/tutorials/default-interface-members-versions/finished/customer-relationship/ICustomer.cs?name=SnippetLoyaltyDiscountVersionTwo)]
+
+There's many new capabilities shown in that small code fragment. Interfaces can now include static members, including fields and methods. Different access modifiers are also enabled. In the previous example, the additional fields are private, and the new method is public. Any of the modifiers are allowed on interface members.
```
Thraka: This was a little confusing. Just for the simple fact that the first sentence is talking in the present "this example" style (mentioning the small code fragment). But it goes on to state that "In the previous example" which made me think it was a different "code fragment" not this one. My first reaction was "we didn't create any fields in the previous example?"

docs/csharp/tutorials/default-interface-members-versions.md
```diff
@@ -0,0 +1,80 @@
...
+
+Applications that use the general formula for computing the loyalty discount, but different parameters don't need to provide a custom implementation; they can set the arguments through a static method. For example, the following code sets a "customer appreciation" that rewards any customer with more than one month's membership:
```
Thraka:
```suggestion
Applications that use the general formula for computing the loyalty discount, but different parameters, don't need to provide a custom implementation; they can set the arguments through a static method. For example, the following code sets a "customer appreciation" that rewards any customer with more than one month's membership:
```

docs/csharp/tutorials/default-interface-members-versions.md
```diff
@@ -0,0 +1,80 @@
...
+
+[!code-csharp[SetLoyaltyThresholds](~/samples/csharp/tutorials/default-interface-members-versions/finished/customer-relationship/Program.cs?name=SnippetSetLoyaltyThresholds)]
+
+## Extend the default implementation
+
+The code you've added so far has provided a convenient implementation for those scenarios where users want something like the default implementation, or to provide an unrelated set of rules. For a final feature, let's refactor the code a bit to enable scenarios where users may want to build on the default implementation. 
+
+Consider a startup that wants to attract new customers. They offer a 50% discount off a new customer's first order. Otherwise, existing customers get the standard discount. A bit of refactoring in the `ICustomer` interface makes this scenerio easy:
+
+[!code-csharp[VersionTwoImplementation](~/samples/csharp/tutorials/default-interface-members-versions/finished/customer-relationship/ICustomer.cs?name=SnippetFinalVersion)]
```
Thraka: I think this adds complexity without any benefit. I expected to see the `ComputeLoyaltyDiscount` method fully implemented here but it instead calls out to the shared static member. Why is that? It wasn't implied in this scenario that the user is creating multiple ICustomer implementations and would want to share functionality between them all (`DefaultLoyaltyDiscount`).

After reading this again and the next paragraph, I think that this code is being written by the library author? However, the introduction of this area talks about things from the users perspective, so I expected to see the user implement some overriding logic here.

docs/csharp/tutorials/default-interface-members-versions.md
```diff
@@ -0,0 +1,80 @@
...
+
+This final version uses a `protected` static method to implement the default algorithm. The public interface method uses that version.  A derived implementation can now extend that default implementation:
+
+[!code-csharp[VersionTwoImplementation](~/samples/csharp/tutorials/default-interface-members-versions/finished/customer-relationship/SampleCustomer.cs?name=SnippetOverrideAndExtend)]
```
Thraka: Maybe it should be pointed out that once you provide your implementation of the interface, the default implementation can no longer be referenced. By moving the default implementation to a static member, you allow the consumer to reference if they need to. I think that concept pointed out in this section makes more sense.

Also, this code, from the users side, is them implementing the "override" for the interface, but it's not totally clear. Maybe show the final class implementing the whole interface?

Thraka (approved):
Great article. Some suggestions 😄

docs/csharp/tutorials/default-interface-members-versions.md
BillWagner: I think I'll leave this as is. It flows easier into the next sections. You are right, that a reasonable default is "don't make any changes". That would make the scenario less interesting.

docs/csharp/tutorials/default-interface-members-versions.md
BillWagner: I reworked the section to explain the scenario better.  I do want to show a utility method added to the interface.

docs/csharp/tutorials/default-interface-members-versions.md
BillWagner: I do want to leave the scenario the same. It flows better into the additional features. (But you do have a point)

docs/csharp/tutorials/default-interface-members-versions.md
BillWagner: I re-wrote this paragraph, and the later sections you commented on. I now make it more clear when I'm discussing the library vs. the code using the library. Good point.

docs/csharp/tutorials/default-interface-members-versions.md
BillWagner: Same comment as above.

PR #11439, opened 2019-03-21
Fix platform invoke data types

nxtn:
Contributes to #11396

*wtypes.h* is for Win16 only. These types are defined in either *winnt.h* or *minwindef.h*.

PR #12251, opened 2019-05-08
Replace Visual Basic tokens

nxtn:
As did in https://github.com/dotnet/docs/pull/11439#discussion_r267941794


rpetrusha (approved):
Thanks for removing these tokens, @NextTurn. In reviewing your PR, I noticed a few other tokens that could be removed, as well as a few inaccuracies, so I've made those changes as well. I'll merge as soon as the internal build completes successfully.

PR #12273, opened 2019-05-09
fix: MD031/blanks-around-fences

nschonni:
Indented so the code fences were under the list items rather than separate blocks


BillWagner (approved):
Thanks again @nschonni We appreciate all the work you continue to do.

Thanks again!

PR #12292, opened 2019-05-10
Replace .NET Framework tokens

nxtn:


BillWagner (approved):
Thanks for making these updates @NextTurn 

Thanks again.

NextTurn:
@BillWagner Does it mean that most of the **[!INCLUDE...]** tokens should be replaced by the target content?

BillWagner:
Ping @rpetrusha for details on which includes are being removed.

I think we want to remove most of them, but Ron would have the best answer.

PR #12299, opened 2019-05-10
Fix header depth

FranklinYu:
“Reliable Sessions” and “Queues” should be parallel to each other, not nested.


rpetrusha (approved):
Thank you, @FranklinYu, for contributing to the dotnet/docs repo and correcting the header level. We'll merge your PR now. Your change should be live on docs.microsoft.com in the next day or two.

PR #12309, opened 2019-05-11
Remove vbprvblong-md.md

nxtn:
Based on #12251


BillWagner (approved):
Thanks for providing the follow-up and cleaning the old sample @NextTurn 

I'll :shipit: now.

PR #12340, opened 2019-05-14
fix: MD047/single-trailing-newline

nschonni:
Files should end with a single newline character


nschonni:
This is a new rule, so the VS Code extension doesn't flag it yet https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md047---files-should-end-with-a-single-newline-character
I used ECLint to fix most of the files https://github.com/jedmao/eclint#insert_final_newline, although there were 3 that didn't seem to pick up and needed manual change

BillWagner (approved):
This all LGTM @nschonni (Hide Whitespace made this one easy)

I'll :shipit: now.

PR #12284, opened 2019-05-10
Replace Visual C# tokens

nxtn:

PR #12350, opened 2019-05-14
Remove csprcslong-md.md

nxtn:
Based on #12284


mairaw (approved):
Thanks @NextTurn for removing this include file.

PR #12359, opened 2019-05-14
Fixed typo

cartacioS:


JRAlexander:
Hi, @cartacioS! This isn't a typo. It's the name of the datafile and since it's downloaded we can't change it. Therefore, I'm closing this. Thanks!

cartacioS:
Oh okay! So just to double check, the files being referred to on step 2 and 3 are two different files? Since one is yelp_labelled and the other is yelp_labeled?

2. Copy the `yelp_labelled.txt` file into the *Data* directory you created. 

3. In Solution Explorer, right-click the `yelp_labeled.txt` file and select **Properties**. Under

PR #12361, opened 2019-05-14
Static constructor behavior with static methods assigned to delegates

rpetrusha:
## Static constructor behavior with static methods assigned to delegates

See Microsoft/dotnet#1034

//cc @jinek


BillWagner (approved):
This LGTM @rpetrusha 

You can :shipit: when ready.

PR #12367, opened 2019-05-15
fix: MD046/code-block-style

nschonni:
Code block style turn on and enforcing fences


BillWagner (approved):
This LGTM @nschonni 

Thanks for all your continued work!

I'll :shipit: now.

PR #12368, opened 2019-05-15
Removed possibly duplicate content: foreach with arrays

pkulikov:
The [How to: Access Command-Line Arguments Using foreach](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/main-and-command-args/how-to-access-command-line-arguments-using-foreach) article begins as follows:

> Another approach to iterating over the array is to use the `foreach` statement as shown in this example. 

So, it's more about iterating through an array, rather than about the command line arguments.

However, the docs already have an article about the `foreach` statement with arrays:
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/arrays/using-foreach-with-arrays


BillWagner (approved):
Thanks for doing this work @pkulikov 

I agree that this content is quite outdated.

In today's world, I'd recommend something like [DragonFruit](https://github.com/dotnet/command-line-api/) for working with command line arguments.

I'll :shipit: now.

PR #12308, opened 2019-05-11
Replace .NET Framework tokens

nxtn:

PR #12307, opened 2019-05-11
Replace .NET Framework tokens

nxtn:

PR #12285, opened 2019-05-10
Replace .NET Framework tokens

nxtn:

PR #12375, opened 2019-05-15
Remove dnprdnshort-md.md

nxtn:
Based on #12285, #12292, #12307, #12308


rpetrusha (approved):
Thanks, @NextTurn, for eliminating all uses of the token and removing the token file. I'll merge your PR now.

PR #12376, opened 2019-05-15
Replace ASP.NET tokens

nxtn:


rpetrusha (approved):
Thank you, @NextTurn, for continuing to remove tokens. I've made a few additional changes -- I've removed several ADO.NET tokens in one topic, and also changed one sentence to use active voice rather than passive. I'll merge your PR as soon as the internal build completes.

PR #12386, opened 2019-05-16
Nit-picking: use dashes for compiler options

pkulikov:


BillWagner (approved):
These are great changes @pkulikov 

I'll :shipit: now.

PR #12404, opened 2019-05-17
fixed keyword: minor edits

pkulikov:
Two changes:
- Remove the duplicate link to the fixed size buffers article from the middle of the article. First, it's mentioned in the first paragraph. Second, in the middle, it was just in-between the example intro and the example itself.
- Remove details about `stackalloc`. Now, with `Span<T>`, `stackalloc` can be used in safe blocks, `unsafe` not always required. The `stackalloc` article describes that, so I've just removed not precise info, rather that describe a new behavior.


BillWagner (approved):
Thanks for updating these @pkulikov 

These are great changes, and I'll :shipit: now.

PR #12428, opened 2019-05-18
Replace WinFX tokens

nxtn:


BillWagner (approved):
Thanks for making these changes @NextTurn  We appreciate it.

Thanks again.

PR #12429, opened 2019-05-18
Replace .NET Core tokens

nxtn:


BillWagner (approved):
This change also looks great @NextTurn 
I'll :shipit: now

Thanks again.

PR #12436, opened 2019-05-19
Replace Silverlight tokens

nxtn:


BillWagner (approved):
These changes LGTM @NextTurn 

I'll :shipit: now.

Thanks again for helping to make the source more readable.

PR #12443, opened 2019-05-20
Null-conditional operator: add Elvis to search keywords

pkulikov:
@BillWagner is it fine for the official docs to use not official name (at least to support search in TOC and help viewer)?
- If not, let's just close this PR then.
- If yes, should I also update the _article text_ to mention "Elvis"?


BillWagner:
I do want to try this @pkulikov 

I want to cache some data on our side so I can see what effect adding this term has on search results, and if the searches result in the correct traffic.

pkulikov:
@BillWagner I've updated the article text as well, then. Please take a look.

BillWagner (approved):
This LGTM @pkulikov 

I'll :shipit: now, and I'll start measuring its impact on search results.

PR #12435, opened 2019-05-19
Replace JScript tokens

nxtn:

PR #12467, opened 2019-05-21
Delete jsprjscript-md.md

nxtn:
Based on #12435


mairaw (approved):
Thanks again @NextTurn!

PR #12492, opened 2019-05-22
Replace .NET Native tokens

nxtn:


rpetrusha (approved):
Your changes look good, @NextTurn. Thanks for replacing more tokens with their token text. I'll merge your PR now.

PR #12494, opened 2019-05-22
 The term  'default ' should be used with or instead of 'null'

MichaelKappel:
A developer I was talking with felt this article was pushing for the use of only types that are nullable. In order to clarify 'default' should also be included so no one sees this as a reason to forbid structs, primitive types, ect...

## Summary

Describe your changes here.

### DIFF
--- a/docs/standard/exceptions/best-practices-for-exceptions.md
+++ b/docs/standard/exceptions/best-practices-for-exceptions.md
@@ -49,7 +49,7 @@ A class can provide methods or properties that enable you to avoid making a call
 [!code-csharp[Conceptual.Exception.Handling#5](../../../samples/snippets/csharp/VS_Snippets_CLR/conceptual.exception.handling/cs/source.cs#5)]
 [!code-vb[Conceptual.Exception.Handling#5](../../../samples/snippets/visualbasic/VS_Snippets_CLR/conceptual.exception.handling/vb/source.vb#5)]
 
-Another way to avoid exceptions is to return `null` for extremely common error cases instead of throwing an exception. An extremely common error case can be considered normal flow of control. By returning `null` in these cases, you minimize the performance impact to an app.
+Another way to avoid exceptions is to return `null` or `default` for extremely common error cases instead of throwing an exception. An extremely common error case can be considered normal flow of control. By returning `null` or `default` in these cases, you minimize the performance impact to an app.
 
 ## Throw exceptions instead of returning an error code
 
--- a/docs/standard/exceptions/best-practices-for-exceptions.md
+++ b/docs/standard/exceptions/best-practices-for-exceptions.md
@@ -49,7 +49,9 @@ A class can provide methods or properties that enable you to avoid making a call
 [!code-csharp[Conceptual.Exception.Handling#5](../../../samples/snippets/csharp/VS_Snippets_CLR/conceptual.exception.handling/cs/source.cs#5)]
 [!code-vb[Conceptual.Exception.Handling#5](../../../samples/snippets/visualbasic/VS_Snippets_CLR/conceptual.exception.handling/vb/source.vb#5)]
 
-Another way to avoid exceptions is to return `null` or `default` for extremely common error cases instead of throwing an exception. An extremely common error case can be considered normal flow of control. By returning `null` or `default` in these cases, you minimize the performance impact to an app.
+Another way to avoid exceptions is to return null (or default) for extremely common error cases instead of throwing an exception. An extremely common error case can be considered normal flow of control. By returning null (or default) in these cases, you minimize the performance impact to an app.
+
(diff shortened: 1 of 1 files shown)


MichaelKappel:
Changed wording  'default ' to 'default' or 'null'

BillWagner:
I like this change @MichaelKappel 

With the addition, I think we need to discuss `Nullable<T>` types.  That would provide a way to distinguish success from error conditions with a `struct`.  I think that was the original intent for recommending `null`. It wasn't expressed well.  

Can you add something like the following:

> To distinguish between the `default` value of a `struct` and a missing value, you can use nullable value types, where the `default` value indicates a missing value.

What do you think?

MichaelKappel:
That's a great idea, I will add that over lunch.

mikedn:
This seems really wrong to me. Of course the article pushing for the use of only types that are nullable because `null` is the only thing that can be reasonably returned if you want to indicate an error condition in this manner. `default` is most of the time a bad idea in the case of primitive types and structs and having to explain

> To distinguish between the default value of a struct and a missing value, you can use nullable value types, where the default value indicates a missing value.

seems like chasing your own tail. You've been trying to avoid nulls and ended up returning a null.

MichaelKappel:
> This seems really wrong to me. Of course the article pushing for the use of only types that are nullable because `null` is the only thing that can be reasonably returned if you want to indicate an error condition in this manner. `default` is most of the time a bad idea in the case of primitive types and structs and having to explain
> 
> > To distinguish between the default value of a struct and a missing value, you can use nullable value types, where the default value indicates a missing value.
> 
> seems like chasing your own tail. You've been trying to avoid nulls and ended up returning a null.

I think this updated change addresses you concerns. In cased default does indicate an error and has no further need for nullables. For example  database keys 0 is invalid, default datetime is invalid, empty guid is invalid, so we have no need to make them nullable.

mikedn:
> I think this updated change addresses your concerns. In many cases default does indicate an error and has no further need for nullables. For example database keys 0 is invalid, default datetime is invalid, empty guid is invalid, so we have no need to make them nullable.

I don't know, this looks like a slippery slope to me. While default `DateTime`/`Guid` may typically be used as special values, the same is not obviously true for numbers. 0 may be invalid as a database id but in other cases it is a completely valid value and then other values are used to indicate errors (most commonly -1, sometimes any negative value, sometimes `int.Min/MaxValue` etc.). As for arbitrary structs that's just a can of worms, you've no idea what the default of such struct represents.

The main problem is that this documentation is mainly targeted at less experienced developers. The experienced ones do not need to read this to know what to do. As for the less experienced ones, you can be sure that someone will read this and sooner or later will end up writing the equivalent of:
```C#
int parse(string s) => int.TryParse(s, out var i) ? i : default;
```

MichaelKappel:
> > I think this updated change addresses your concerns. In many cases default does indicate an error and has no further need for nullables. For example database keys 0 is invalid, default datetime is invalid, empty guid is invalid, so we have no need to make them nullable.
> 
> I don't know, this looks like a slippery slope to me. While default `DateTime`/`Guid` may typically be used as special values, the same is not obviously true for numbers. 0 may be invalid as a database id but in other cases it is a completely valid value and then other values are used to indicate errors (most commonly -1, sometimes any negative value, sometimes `int.Min/MaxValue` etc.). As for arbitrary structs that's just a can of worms, you've no idea what the default of such struct represents.
> 
> The main problem is that this documentation is mainly targeted at less experienced developers. The experienced ones do not need to read this to know what to do. As for the less experienced ones, you can be sure that someone will read this and sooner or later will end up writing the equivalent of:
> 
> ```cs
> int parse(string s) => int.TryParse(s, out var i) ? i : default;
> ```

This is not not a 'Hello World' article, when I show a junior programmer a "best practice" article on software architecture it should not suggest a return type of 'null' without mention the 'default'.

I feel that just by saying 'Null' (or 'Default') we better illustrate the best practice of considering both, I am trying to improve a code base with overuse of nullable types and bad exception handling and the Jr Programmers saying the this article vindicates every property, parameter, and return type being a nullable... well that's not a best practice

mikedn:
>  I am trying to improve a code base with overuse of nullable types and bad exception handling and the Jr Programmers saying the this article vindicates every property, parameter, and return type being a nullable... well that's not a best practice

I don't understand how adding `default` into the mix solves anything. The default for a reference type is null so it's the same thing. The default for int 0, a value that may very well be a valid value. The default for a struct is who knows what.

As far as I can tell you have a problem with overuse of error returns in place of exceptions. That's understandable in a language that doesn't have any special support for this (no option type, no ability to enforce the checking of returned values etc.). In this the best course of action might be to just delete this paragraph from documentation.

BillWagner (approved):
Thanks for adding the additional language @MichaelKappel 

I do think you've captured the tradeoffs when `default` can indicate *the return value is missing* and `default` means *the default return value is a different success case*. 

I've approved the changes, and I'll :shipit: now.

PR #12496, opened 2019-05-23
Fixing a typo

Jiffrope:
"your" looks like it should be "you're"

## Summary

Describe your changes here.


rpetrusha:
Thank you for contributing to the dotnet/docs repo, @Jiffrope. Unfortunately, "your" is correct here **Your using directives** refers to the **using** directives in your code; **you are using directives** is unclear, since using becomes a verb and directives becomes an indefinite noun that does not clearly refer to using directives. As a result, I'll close this PR.

PR #12499, opened 2019-05-23
Fix line wrapping

flcdrg:
Ensure each line starts on new line

## Summary

Suffix with two spaces so next line starts on new line rather than being appended to current


BillWagner (approved):
Thanks for spotting this and fixing it @flcdrg 

We appreciate it.

Thanks again.

PR #12525, opened 2019-05-24
Try to fix Loc issue

v-thepet:
Per User Story https://mseng.visualstudio.com/TechnicalContent/_queries/edit/1537022/?triage=true.
The `$`` coding is causing rendering issues in MT languages.


v-thepet:
@rpetrusha - Per Loc bug https://mseng.visualstudio.com/TechnicalContent/_queries/edit/1537022/?triage=true, the code formatting around the $' entry in the table worked ok in the en-us and some other languages, but caused rendering issues in MT languages. The only fix seems to be to remove the code formatting, so to make it look better I removed the code formatting from all the entries. Please approve/merge if you agree. Thanks.

rpetrusha (approved):
I think that this is the best solution, @v-thepet. I'll merge your PR now.

PR #12529, opened 2019-05-25
Move the article about true and false operators

pkulikov:
As far as I understand, in the consolidated C# reference *all* operators are in one TOC. That TOC and part of the doc set is [C# operators](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/). That means that all [operator keywords](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/operator-keywords) pages should move to that location. This PR contributes to that move by moving the article about `true` and `false` operators. Also, I've used that opportunity to remove [true](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/true) and [false](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/false) "link farms" and simplify the keywords TOC a bit.


docs/csharp/language-reference/operators/true-false-operators.md
```diff
@@ -0,0 +1,38 @@
+---
```
pkulikov: Because of the move, I had to update all the relative links in this file. No content changes were made.

BillWagner (approved):

I'll :shipit: now.

PR #12523, opened 2019-05-24
Replace .NET Framework 4.6.1 tokens

nxtn:

PR #12531, opened 2019-05-25
Delete net-v461-md.md

nxtn:
Based on #12523


BillWagner (approved):
Thanks @NextTurn 

This builds with no warnings, verifying that we've removed all the necessary instances of the includes.

I'll :shipit: now.

PR #12548, opened 2019-05-28
C91896: Deleting extra backslahes

v-makoud:


v-makoud:
#sign-off

rpetrusha:
The `\` character was used to introduce a line break, @v-makoud. I've added two spaces instead, which should have the same effect. I'll merge when the build completes successfully.

rpetrusha (approved):
The build looks good, so I'll merge, @v-makoud.

PR #12573, opened 2019-05-29
remove duplicate gac-assemblies content

BillWagner:
Extracted from #12244 

This removes the assemblies and GAC section from the C# guide.

It is duplicated in the .NET guide.


rpetrusha:
I'm somewhat concerned about the deletions and the removals in this PR, @BillWagner. I'm wondering I f it requires a more careful approach.
As I recall (and I may be wrong on many of the details), the basic history of this content was that: 
- There was a single assemblies section, with some topics identified as focusing on VB and C# (together).
- Topics that seemed VB or C#-related were indiscriminately duplicated in the VB and C# doc set, even though nothing language-specific is really covered here.
- Many of these topics also continued to exist in the .NET Framework documentation.
- The result when we migrated to docs was that there's a duplication of the C#, Visual Basic, and .NET documentation.
My concerns are: 
1. Since we have two or three identical topics and all have potentially been maintained independently of one another, I think it's necessary to try to reconcile differences before proceeding. In particular, since once these topics are deleted, the only remaining copy is in the VB doc set, I'm wondering if we first need to diff the C# and the VB versions.
2. Some of the removed topics are now silly or unnecessary (like how-to-determine-if-a-file-is-an-assembly). but others remain important (the signed friend assembly one, for example).
3. Something needs to be done with the VB topics as well, which don't belong in the VB doc set.

What do you think?

BillWagner:
@rpetrusha Thanks for providing the history. I wasn't sure how we got to where we are now.  That's part of why I made this PR confined to this section.

There were some other interesting issues I found while looking at this:

- This section (assemblies & gac) is not linked anywhere in the C# Guide TOC. 
- I don't find the corresponding VB files in the VB TOC either.

I agree with all your concerns.

I'll close this, and open a new issue where we can decide on the plan to retire the unnecessary topics and move the relevant ones to a better location with more visibility, while merging duplicate content.

PR #12578, opened 2019-05-29
Update the LINQ to Objects TOC section

BillWagner:
Extracted from #12244 

This one is rather small, but LINQ to XML starts to explode quickly


docs/csharp/programming-guide/concepts/linq/toc.yml
```diff
@@ -89,12 +91,12 @@
         href: how-to-compute-column-values-in-a-csv-text-file-linq.md
     - name: LINQ and Reflection
       href: linq-and-reflection.md
-      items:
-      - name: "How to: Query An Assembly's Metadata with Reflection (LINQ)"
-        href: how-to-query-an-assembly-s-metadata-with-reflection-linq.md
+    - name: "How to: Query An Assembly's Metadata with Reflection (LINQ)"
```
rpetrusha: The TOC here seems problematic. The "How to: Query..." is really a child of the LINQ and Reflection topic. I think that I would have LINQ and Reflection as the TOC node with a single child, the "How to Query..." topic.  The how-to isn't really a how-to, though. So I'd just move the existing content of the LINQ and Reflection topic (which is really just a portal page with one child) to the How to... topic, and rename the How to topic something like: "Query an assembly's metadata with LINQ".

rpetrusha (approved):
LGTM. Left a couple of comments for you to consider, @BillWagner. You can merge when you're ready, or let me know if you think this needs another review.

docs/csharp/programming-guide/concepts/linq/toc.yml
BillWagner: Good idea. I merged these two topics into the "How to" article.

PR #12601, opened 2019-05-31
Remove net-v45-md.md

nxtn:


BillWagner:
@NextTurn 

This build failed, https://github.com/NextTurn/docs/blob/v45/docs/standard/collections/thread-safe/index.md uses the include you removed.

I notice that the include you reference in the title is different than the include you removed.  Once you make the correct updates, I'll look again.

NextTurn:
What a mistake.

NextTurn:
I've made the update. Good to go?

rpetrusha (approved):
Yes. It's good to go. Thanks for removing the token, @NextTurn.

PR #12643, opened 2019-06-03
Replace File Explorer tokens

nxtn:


rpetrusha (approved):
Thanks for replacing the token references with token text, @NextTurn, as well as for removing the token. I'll merge your PR now.

PR #12644, opened 2019-06-03
Replace Transact-SQL tokens

nxtn:


rpetrusha (approved):
Thank you, @NextTurn, for replacing  the Transact-SQL token with its token text and for deleting the token. I'll merge your PR now.

PR #12673, opened 2019-06-04
Combine selected TOC nodes in the C# Guide

### LINKED ISSUE #12649: Simplify C# Guide TOC structure
The C# Guide made use of many child TOCs as part of the migration from MSDN.

A number of these should be combined into a single TOC, especially in the C# Language reference.

That would enable the DisplayName feature to be more useful to find reference nodes with operator symbols (e.g. `??`, `+=`, etc)

BillWagner:
Fixes #12649 

After merging from MSDN, every folder was given its own child TOC. This has proven somewhat unfriendly.

This task combines many of those into a set of top-level TOC nodes.

[Language reference home page](https://review.docs.microsoft.com/en-us/dotnet/csharp/language-reference/?branch=pr-en-us-12673) internal review link.

A good test is to use the "displayName" feature in the TOC for different operators.


rpetrusha (approved):
This looks good, @BillWagner. And it certainly creates a much better experience! You can merge when you're ready.

PR #12678, opened 2019-06-05
Fix definition of false operator

peterszekeli:
False operator stated to return the `true` value.

### DIFF
--- a/docs/csharp/language-reference/operators/index.md
+++ b/docs/csharp/language-reference/operators/index.md
@@ -91,7 +91,7 @@ These operators have higher precedence than the next section and lower precedenc
 
 [true operator](true-false-operators.md) - returns the [bool](../keywords/bool.md) value `true` to indicate that an operand is definitely true.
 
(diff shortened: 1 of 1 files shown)


docs/csharp/language-reference/operators/index.md
```diff
@@ -91,7 +91,7 @@ These operators have higher precedence than the next section and lower precedenc
 
 [true operator](true-false-operators.md) - returns the [bool](../keywords/bool.md) value `true` to indicate that an operand is definitely true.
 
-[false operator](true-false-operators.md) - returns the [bool](../keywords/bool.md) value `true` to indicate that an operand is definitely false.
+[false operator](true-false-operators.md) - returns the [bool](../keywords/bool.md) value `false` to indicate that an operand is definitely false.
```
pkulikov: The current definition is correct. The operator replies the question: is the operand definitely false. Answer: yes (returns `true`) or no (returns `false`).

peterszekeli: Okay, clear now. Thanks.

BillWagner: See #12679

PR #12715, opened 2019-06-07
Fix contravariant/covariant markings for IReadOnlyList and IReadOnlyCollection

DillonN:
## Summary

`IReadOnlyList<T>` and `IReadOnlyCollection<T>` are actually covariant in `T`, not contravariant (as evident by the fact they both declare `<out T>` and inherit from the covariant `IEnumerable<T>`).


BillWagner (approved):
Thank you for spotting these and providing a fix @DillonN We appreciate it.

You should see the changes on the live site in the next day or so.

DillonN:
Awesome, thanks!

PR #12733, opened 2019-06-07
Expand Static Constructor C# documentation

### LINKED ISSUE #12639: C# Static Constructor Exception handling documentation
For the [Static Constructors - C# Programming Guide](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/static-constructors?view=netframework-4.8), a section on exception handling should be added:

1.  Unhandled exceptions thrown in the static class are wrapped in a [TypeInitializationException exception](https://docs.microsoft.com/en-us/dotnet/api/system.typeinitializationexception?view=netframework-4.8#remarks).

1.  A remark that though not explicitly accessible, the presence of a static constructor should be documented to make debugging an exception in the static constructor possible. ([Undocumented static constructors may require disassembly to identify](https://docs.microsoft.com/en-us/dotnet/api/system.typeinitializationexception?view=netframework-4.8#remarks).)

---
#### Document Details

* ID: aa3ed1c5-1f79-9234-beec-b5771c3fb3c0
* Version Independent ID: ac4f6361-70a7-fb43-101e-c12044f60896
* Content: [Static Constructors - C# Programming Guide](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/static-constructors?view=netframework-4.8)
* Content Source: [docs/csharp/programming-guide/classes-and-structs/static-constructors.md](https://github.com/dotnet/docs/blob/master/docs/csharp/programming-guide/classes-and-structs/static-constructors.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

EugeneKramer:
## Summary

Expands documentation on C# static constructor.

Added:
- A static constructor will run before an instance constructor
- Static field initializer will run before static constructor
- Presence of implicit static constructor
- Note of System.TypeInitializationException exception for unhandled exceptions in static class
- Is only meant to be called by the common language runtime (CLR).
- Added link to CA2121: Static constructors should be private
- Note that presence of an explicit static constructor should be documented.
-  C# language spec link for static constructor
- Constructor Design link
- Cannot be inherited/overloaded. Only appears once.
- System.Reflection.TypeAttributes.BeforeFieldInit
- Static Readonly coverage
- <s>RuntimeHelpers.RunClassConstructor - Explicit initialization method</s>

Formatted:
- Moved usage example to separate section. Usage is not a property.

Fixes https://github.com/dotnet/docs/issues/12639


docs/csharp/programming-guide/classes-and-structs/static-constructors.md
```diff
@@ -16,22 +16,30 @@ A static constructor is used to initialize any [static](../../../csharp/language
...
-- If a static constructor throws an exception, the runtime will not invoke it a second time, and the type will remain uninitialized for the lifetime of the application domain in which your program is running.  
   
 ## Example  
  In this example, class `Bus` has a static constructor. When the first instance of `Bus` is created (`bus1`), the static constructor is invoked to initialize the class. The sample output verifies that the static constructor runs only one time, even though two instances of `Bus` are created, and that it runs before the instance constructor runs.  
   
- [!code-csharp[csProgGuideObjects#15](~/samples/snippets/csharp/VS_Snippets_VBCSharp/csProgGuideObjects/CS/Objects.cs#15)]  
+ [!code-csharp[csProgGuideObjects#15](~/samples/snippets/csharp/VS_Snippets_VBCSharp/csProgGuideObjects/CS/Objects.cs#15)]
+ 
+## C# language specification  
+For more information, see the [Static constructors](~/_csharplang/spec/classes.md#static-constructors) section of the [C# language specification](../../language-reference/language-specification/index.md).
```
pkulikov: There is no *language-specification* folder in this repo anymore and the referenced index page is redirected to the one in this suggestion from the csharplang repo

```suggestion
For more information, see the [Static constructors](~/_csharplang/spec/classes.md#static-constructors) section of the [C# language specification](~/_csharplang/spec/introduction.md).
```

docs/csharp/programming-guide/classes-and-structs/static-constructors.md
```diff
@@ -11,27 +11,46 @@ ms.assetid: 151ec95e-3c4d-4ed7-885d-95b7a3be2e7d
...
-  
-- A static constructor cannot be called directly.  
-  
-- The user has no control on when the static constructor is executed in the program.  
+- A static constructor is called automatically to initialize the [class](../../../csharp/language-reference/keywords/class.md) before the first instance is created or any static members are referenced. A static constructor will run before an instance constructor. Note that a type's static constructor is called when a static method assigned to an event or a delegate is invoked and not when it is assigned. If static field variable initializers are present in the class of the static constructor, they will be executed in the textual order in which they appear in the class declaration immediately prior to the execution of the static constructor.
+
+- If you don't provide a static constructor to initialize static fields, the C# compiler will supply a default implicit static constructor that initializes static fields to their default value as listed in the [Default Values Table](../../../csharp/language-reference/keywords/default-values-table.md).
```
BillWagner: I'd like to rephrase this. The end result (initializing values to their default) is correct, but the compiler is not required to supply a static constructor. See the [last paragraph on default values in the C# language spec](https://docs.microsoft.com/dotnet/csharp/language-reference/language-specification/variables#default-values).

```suggestion
- If you don't provide a static constructor to initialize static fields, all static fields are initialized to their default value as listed in the [Default Values Table](../../../csharp/language-reference/keywords/default-values-table.md). 
```

BillWagner (changes requested):
Thanks for making these updates @EugeneKramer These are great additions that provide needed information about defining static constructors.

I have one suggestion for an update where the phrasing could be mis-interpreted. Let me know what you think of the suggestion, and then this is ready to :shipit:

docs/csharp/programming-guide/classes-and-structs/static-constructors.md
EugeneKramer: Appreciate the input.

That line was incorporated from [Constructors (C# Programming Guide)](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/constructors#static-constructors) :
>...If you don't provide a static constructor to initialize static fields, the C# compiler will supply a default static constructor that initializes static fields to their default value as listed in the Default Values Table.

Also, similar phrasing is used in [Static constructors and the TypeInitializationException](https://docs.microsoft.com/en-us/dotnet/api/system.typeinitializationexception?view=netframework-4.8#static-constructors-and-the-typeinitializationexception):
> ...Static constructors can be explicitly defined by a developer. If a static constructor is not explicitly defined, compilers automatically create one to initialize any static (in C#) or Shared (in Visual Basic) members of the type.

Should the text on these other pages be revised?

EugeneKramer:
I've added a note for static readonly initialization and using RuntimeHelpers.RunClassConstructor for explicitly calling the static constructor.

If they're ok, feel free to :shipit:

docs/csharp/programming-guide/classes-and-structs/static-constructors.md
BillWagner: 😢 Yes, those pages should be updated as well. We're moving toward ensuring that the articles in these sections use language that's consistent with the spec (although often more informal). These should agree with the spec, and not imply different behavior.

Do you want to make those additional changes, or should I create an issue to update those docs and reference this conversation?

docs/csharp/programming-guide/classes-and-structs/static-constructors.md
```diff
@@ -11,27 +11,55 @@ ms.assetid: 151ec95e-3c4d-4ed7-885d-95b7a3be2e7d
...
-  
-- If a static constructor throws an exception, the runtime will not invoke it a second time, and the type will remain uninitialized for the lifetime of the application domain in which your program is running.  
-  
-## Example  
+
+- Static constructors are also a convenient place to enforce run-time checks on the type parameter that cannot be checked at compile-time via constraints (Type parameter constraints).
+
+### Explicit initialization
+
+> [!CAUTION]
```
BillWagner: I think that leaving this section out would be better. Readers that already know about `RunClassConstructor` would (hopefully) know they should not call the method. Other readers can remain blissfully unaware of its existence.

What were your thoughts regarding adding this?

EugeneKramer: The `RunClassConstructor` was mainly added for completion's sake . It has been removed as its presence would promote misuse.

It may be worthwhile to include an example that an empty static void method would serve the same purpose.

```csharp
using System; 

public class B 
{      
    public static void TriggerStaticConstructor() {} 
	
    static B(){
        Console.WriteLine("Static constructor invoked.");
    }
} 
public class A 
{
    public static void Main(string[] args){
        B.TriggerStaticConstructor();
    }
} 
```

docs/csharp/programming-guide/classes-and-structs/static-constructors.md
EugeneKramer: I'm not sure how to best rephrase the text, so it'd be better if you were to file the issue.

Also, having looked at the IL output on [SharpLab](https://sharplab.io), I'm not sure the phrasing is incorrect.  When a static field is initialized at declaration, a `.cctor` is generated within which the static fields are initialized.

```csharp
//no .cctor
public class A{
    public static int x;
}
```
```csharp
//has .cctor 
//initializiation takes place in .cctor
public class B{
    public static int x=1;
}
```
```csharp
//has .cctor 
//initializiation takes place in .cctor
//x initialized before  y
public class C{
    public static int x=1;
    static C(){int y=2;}
}
```

docs/csharp/programming-guide/classes-and-structs/static-constructors.md
BillWagner: Thanks for the update. I think this addition isn't necessary, so I'll :shipit: now.

BillWagner (approved):

PR #12743, opened 2019-06-08
Remove > character from console commands

natke:
The character causes an error if you copy the text directly.


rpetrusha (approved):
Thanks for making this change, @natke. I'll merge your PR now.

PR #12745, opened 2019-06-09
Correct C# language specification section location for sealed

EugeneKramer:
## Summary

Move C# language specification section to above "See also" section.


rpetrusha (approved):
Thank you, @EugeneKramer, for standardizing the C# specification and see also sections. We'll merge your PR now. The change should be live on docs.microsoft.com in the next day or two.

PR #12749, opened 2019-06-09
Remove reference to F# 4.1 in records doc

cartermp:
F# 4.6 is now out, so there's been more than enough time to have the structness spread such that it's a new normal


rpetrusha (approved):
LGTM. You can merge when you're ready, @cartermp.

PR #12752, opened 2019-06-10
Value Options use postfix form

Happypig375:
## Summary

Added Value Options which use postfix form


cartermp (approved):
Thanks @Happypig375

PR #12766, opened 2019-06-11
Corrected that datatype is optional in for each

Youssef1313:
The docs said the the as datatype part in for each loop is required if element isn't already declared.
I've tested for each loop without the datatype part and it works for me

[testing witout datatype part]
image

[testing with datatype part]
image

### DIFF
--- a/docs/visual-basic/language-reference/statements/for-each-next-statement.md
+++ b/docs/visual-basic/language-reference/statements/for-each-next-statement.md
@@ -107,8 +107,8 @@ Next [ element ]
  When a `For Each`…`Next` statement runs, Visual Basic evaluates the collection only one time, before the loop starts. If your statement block changes `element` or `group`, these changes don't affect the iteration of the loop.  
   
  When all the elements in the collection have been successively assigned to `element`, the `For Each` loop stops and control passes to the statement following the `Next` statement.  
-  
- If `element` hasn't been declared outside this loop, you must declare it in the `For Each` statement. You can declare the type of `element` explicitly by using an `As` statement, or you can rely on type inference to assign the type. In either case, the scope of `element` is the body of the loop. However, you cannot declare `element` both outside and inside the loop.  
+ 
+If [Option Infer](option-infer-statement.md) is on (its default setting), the Visual Basic compiler can infer the data type of `element`. If it is off and `element` hasn't been declared outside the loop, you must declare it in the `For Each` statement. To declare the data type of `element` explicitly, use an `As` clause. Unless the data type of element is defined outside the `For Each`...`Next` construct, its scope is the body of the loop. Note that you cannot declare `element` both outside and inside the loop.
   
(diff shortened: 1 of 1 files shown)


BillWagner:
@rpetrusha Can you look at this? I think the original text is correct, but I'm not certain. The screenshots show an array of `Int`, so the type of the `num` variable is declared as an `Int`.  Is that correct?

rpetrusha:
Thanks for submitting this PR, @Youssef1313. You've certainly identified an issue in this article, which is in serious need of revision, but your suggested change isn't quite correct. Whether or not declaration of the iterator variable is optional depends on the value `Option Infer`. If on (the default value), the compiler uses type inference to determine the type of the iterator variable. If off, type inference is disallowed, and compiler emits error BC30451. I'll suggest an alternate phrasing, as well as a few other changes that it would be nice to make to modernize the article.

docs/visual-basic/language-reference/statements/for-each-next-statement.md
```diff
@@ -43,7 +43,7 @@ Next [ element ]
 |Term|Definition|  
 |---|---|  
 |`element`|Required in the `For Each` statement. Optional in the `Next` statement. Variable. Used to iterate through the elements of the collection.|  
-|`datatype`|Required if `element` isn't already declared. Data type of `element`.|  
+|`datatype`|Optional. Data type of `element`.|
```
rpetrusha:
```suggestion
|`datatype`|Optional if [`Option Infer`](option-infer-statement.md) is on (the default) or `element` is already declared; required if `Option Infer` is off and `element` isn't already declared. The data type of `element`.|  
```

rpetrusha (changes requested):
I've left a suggested change to the text that you've modified, @Youssef1313. In addition, could you make the following changes?
- line 111: 
   If \[Option Infer](option-infer-statement.md) is on (its default setting), the Visual Basic compiler can infer the data type of \`element\`. If it is off and \`element\` hasn't been declared outside the loop, you must declare it in the \`For Each\` statement. To declare the data type of \`element\` explicitly, use an \`As\` clause.  

   Unless the data type of `element` is defined outside the \`For Each\`...\`Next\` construct, its scope is the body of the loop. Note that you cannot declare \`element\` both outside and inside the loop.

- line 121:
   The runtime must be able to convert the elements in \group\` to \`element\`. The \[\`Option Strict\`] statement controls whether both widening and narrowing conversions are allowed (\`Option Strict\` is off, its default value), or whether only widening conversions are allowed (\`Option Strict\` is on). For more information, see \[Narrowing conversions](#narrowing-conversions).

Youssef1313:
@rpetrusha , Thanks for the correction. I've committed your changes.

rpetrusha (approved):

rpetrusha:
Thanks for making the additional changes, @Youssef1313, and for doing it so quickly! I'll merge your PR now. The updated article should be live on docs.microsoft.com in the next day or two.

PR #12767, opened 2019-06-11
Replace Windows Runtime tokens

nxtn:


rpetrusha (approved):
Thank you so much for continuing to replace tokens with their token text and remove the tokens, @NextTurn.

One issue that occurred to me is that token removal has localization costs -- each article that has tokens removed must be re-translated even though the text has not changed, which can incur significant cost. In general, removing these tokens has an enormous benefit -- it makes the text significantly more accessible to community members. What we would ask is that if token removal touches an especially large number of files -- somewhere in the neighborhood of 75 seems like a good limit -- that the tokens be left in place until some later date.

I'll merge your PR now.

NextTurn:
In addition to token replacement, are there others actions that lead to retranslation?

NextTurn:
Can't we suppress the localization process for such minor changes?

rpetrusha:
Unfortunately, any change triggers re-localization, @NextTurn. There's no way to avoid it without completely redesigning the way our documentation is localized.

NextTurn:
What about those style changes (blank line updates)?

Does it mean that we should avoid touching large articles if not necessary?

PR #12768, opened 2019-06-11
Replace ASP.NET 2.0 tokens

nxtn:


mairaw (approved):
Thank you @NextTurn!

NeonFireAngel:
Thank you for me involving me here.  I have a concern here is there by chance I could interfere or reconfigure some way as to make any concerns to another by way of lacking knowledge on my part. this is in the back of my mind.

PR #12769, opened 2019-06-11
Fix xref links

pkulikov:
I've mixed `displayProperty` with `displayName` in TOC
Also added the link to `unsafe` on its first occurrence (missed that when swapped two list items during review)


rpetrusha (approved):
Thanks for making these corrections, @pkulikov. I'll merge your PR now.

PR #12789, opened 2019-06-12
fix explanation of range start and end

adamralph:
## Summary

The current explanation is wrong, since ranges are not exclusive. The range start is _inclusive_.


BillWagner:
Ping @agocke 

Is this change consistent with how we want to describe ranges? I think it's more clear than the original, but wanted to make sure we use the same language in docs as we plan for the specs.

agocke (approved):
LGTM

BillWagner (approved):
Thanks for making these clarifying remarks @adamralph 

Thanks for reviewing @agocke 

I'll :shipit: now.  

Congrats on your first contribution to docs! 🎆

PR #12795, opened 2019-06-12
Replace Release key values note tokens

nxtn:


rpetrusha:
Thanks for opening this PR, @NextTurn. It exposes an error of omission on my part -- I originally created the token because I intended to use identical token text in multiple places. I then forgot to add the token to the other cases in which the release DWORDs appeared. For that purpose, the token remains valuable. So I'll close this PR and open a PR that adds the token to the other articles that should have used it in the first place.

PR #12810, opened 2019-06-13
Fixed incorrect Predict() method reference

StevenJDH:
## Summary

The Predict() method does not exist and is meant to be the UseModelWithSingleItem() method.


JRAlexander (approved):
LGTM. Good catch, @StevenJDH, thanks for your contribution! I'll merge this straightaway and you should see the changes on live within the next day or so.

Issue #11793, opened 2019-04-11
Change link text, like destination for Go Global Developer Center


rpetrusha:
## Change link text, like destination for Go Global Developer Center

In numerous places in the documentation (in fact, in almost every place that we mention a culture name as a parameter or the return value of a property), we link to a list of language/region names in the Go Global Developer Center. The Go Global Developer Center has been retired; the content has been moved to <docs.microsoft.com/openspecs/windows-protocols/>. 

As a result of this change, the link description and link target should be changed. Currently, it generally reads "see the National Language Support (NLS) API Reference at the Go Global Developer Center." The general description should be changed to:

> see the **Language tag** column in the \[list of language/region names supported by Windows](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-lcid/a9eac961-e77d-41a6-90a5-ce1a8b0cdb9c). Culture names follow the standard defined by \[BCP 47](https://tools.ietf.org/html/bcp47).

//cc @tarekgh @breyed

PR #12813, opened 2019-06-13
Modified Go Global Developer Center Url

florinciubotariu:
Relates to #11793.

This is the only file where I could find "Go Global Developer Center" text.


rpetrusha (approved):
It is indeed the only mention of Go Global Developer Center, @florinciubotariu. Thank you for updating it. I'll merge your PR now. The change should be live on docs.microsoft.com in the next day or two.

PR #12833, opened 2019-06-14
Minor version change

nishanil:
## Summary
Removed minor version numbering in the **EDITION**. Versions are in sync with the .NET Core release. Since there are frequent changes in the online docs by means of Community PRs and such, we will have to frequently bump this number which is not required. We will curate the fixes here and bump the minor versions in the downloadable formats with changelogs as and when necessary.


BillWagner (approved):
This LGTM @nishanil 

You can :shipit: when you are ready

PR #12837, opened 2019-06-14
Minor updates

nxtn:
Continued from #12789


mairaw (approved):
Thanks @NextTurn! I like your change.

PR #12852, opened 2019-06-16
Replace desktop tokens

nxtn:


rpetrusha (approved):
Thanks for replacing the token with its token text and removing the token, @NextTurn. I'll merge your PR now.

PR #12854, opened 2019-06-16
C# XML serialization example- text clarification

EugeneKramer:
## Summary

Add in which class requires a a public, parameterless constructor:
The class **_being serialized_** must have a public constructor without parameters.


BillWagner (approved):
Thanks for making this clarifying change @EugeneKramer. 

I've reviewed the change, and I'll :shipit: now. Thanks again for contributing to docs!

PR #12858, opened 2019-06-17
Sync the H1 title with the title metadata

pkulikov:


BillWagner (approved):
Thanks for making this change @pkulikov 

This is a good way to voice these keywords in the title.

PR #12876, opened 2019-06-18
Added additional COR_PRF_HIGH* flags

rpetrusha:
## Added additional COR_PRF_HIGH* flags

@maoni0 and @mjsabby, could you please review this? Also, @Maoni0, what are the APIs for the  updating generational bounds callbacks?


Maoni0:
https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/icorprofilerinfo2-getgenerationbounds-method is the API for getting the generation bounds. the doc says "That is, it can be called from any callback except those that occur between ICorProfilerCallback2::GarbageCollectionStarted and ICorProfilerCallback2::GarbageCollectionFinished.". I don't actually see a reason why it can't be called during those callbacks.

Maoni0 (approved):
:shipit:

rpetrusha:
Thanks for the review, @Maoni0!

BillWagner (approved):
This LGTM @rpetrusha 

You can :shipit: when ready.

PR #12887, opened 2019-06-18
disable some interactive samples

BillWagner:
The samples being disabled don't work under the current implementation. There are a few different reasons:

- some define classes or structs
- some use Span or ReadonlySpan
- some use unsafe code.


rpetrusha (approved):
LGTM, @BillWagner. You can merge when you're ready.

PR #12906, opened 2019-06-19
Replace ADO.NET 2.0 tokens

nxtn:


rpetrusha (changes requested):
Thanks for removing these tokens, @NextTurn. I think that in both cases, though, the references to a specific version of ADO.NET are no longer useful, and that ADO.NET 2.0 can be replaced with ADO.NET.

rpetrusha (approved):

rpetrusha:
Thank for making the additional changes, @NextTurn. I'll merge your PR now.

PR #12907, opened 2019-06-19
Replace .NET Compact Framework tokens

nxtn:


rpetrusha (approved):
Thank you, @NextTurn, for replacing the .NET Compact Framework token with its token text and removingt the token. I'll merge your PR now.

PR #12912, opened 2019-06-20
typo: Double word "was"

nschonni:


mairaw (approved):
Thanks for removing the double word!

PR #12926, opened 2019-06-20
Replace Entity Data Model tokens

nxtn:


rpetrusha (approved):
Thank you, @NextTurn, for replacing another token with its text and removing the token. I'll merge your PR now.

PR #12949, opened 2019-06-21
Make is and as more discoverable in TOC

pkulikov:
I've tried to filter TOC by "is" and "as": not the best experience with too many not relevant results. Thus, this update.


BillWagner (approved):
I tried this on the review site, and this is a great solution @pkulikov 

I'll :shipit: now.

PR #12950, opened 2019-06-21
Delete local tokens TOC

pkulikov:
Looks like this TOC was forgotten to be removed. The *keywords* and *operators* folders don't contain local TOC files.


BillWagner (approved):
Good catch.  Thanks @pkulikov 

I'll :shipit: now

PR #12958, opened 2019-06-21
Replace Microsoft Visual C# 2005 tokens

nxtn:

PR #12965, opened 2019-06-22
Replace Visual C# 2005 with C# 2.0

nxtn:
Continued from #12958


BillWagner (approved):
Thanks for making this change @NextTurn 

Thanks again!

PR #12957, opened 2019-06-21
Replace Visual C# 2008 tokens

nxtn:

PR #12966, opened 2019-06-22
Replace Visual C# 2008 with C# 3.0

nxtn:
Continued from #12957


BillWagner (approved):
Thanks again @NextTurn 

PR #12967, opened 2019-06-22
Fixed list formatting

pkulikov:
The first list item about the `readonly` field contains a sublist of two items.


mairaw (approved):
Great catch @pkulikov! Thanks for making these changes.

PR #12953, opened 2019-06-21
Use left/right terms for operands

pkulikov:
Fixes #12952

PR #12973, opened 2019-06-24
Use left-/right-hand terms for operands

pkulikov:
The consistency follow-up of #12953 

@BillWagner I've checked and updated all operator-related articles, including ones about shift operators and delegate operations.


BillWagner (approved):
Thanks for making these update @pkulikov 

It will add to the consistency of our docs.

I'll :shipit: now.  Thanks!

PR #12975, opened 2019-06-24
Direct readers to dotnet try tutorial instead of online exploration

BillWagner:
The interactive explorations are in the dotnet/try-samples repo.

Update all the "waht's new' articles for C# 7.x and C# 8 to point to the new `dotnet try` based tutorials.

In addition, retire the C# 7 interactive exploration. The dotnet try experience ofers better fidelity with the latest language features.

PR #12984, opened 2019-06-24
Update TOC

rpetrusha:
## Update TOC

This PR removes the C# 7 interactive exploration, which was removed in #12975.


BillWagner (approved):
Thanks @rpetrusha 

I missed this in my early morning edits on the plan.

This is a great catch, and it LGTM.

PR #12988, opened 2019-06-25
Add missing quote

mskelton:
## Summary

Adds a missing quote to the code comment.


mairaw (approved):
Thank you @markypython for adding the missing quote. Your contribution should be live within a day or so.

PR #12995, opened 2019-06-25
True and false operators article: split in sections

pkulikov:
Moved around paragraphs and added H2 titles. Otherwise, the TOC at the right-hand side of the page contains only one item.


BillWagner (approved):
Thanks @pkulikov 

This is a nice reorganization.  I'll :shipit: now.

PR #13003, opened 2019-06-25
chore: Add cSpell words from docs/core

nschonni:
Not everything caught but the CLI, but ones that should be safe to ignore


rpetrusha (approved):
LGTM. I'll merge your PR now, @nschonni.

PR #13007, opened 2019-06-25
Fix name in XML from Tags to PackageTags

Thraka:


rpetrusha (approved):
LGTM. You can merge when you're ready, @Thraka.

PR #13011, opened 2019-06-25
C98648: Fixing bookmark to avoid issues on localized pages

v-maudel:
Description: <Provide further explanation here as needed>  
@pkulikov

### DIFF
--- a/docs/csharp/language-reference/operators/pointer-related-operators.md
+++ b/docs/csharp/language-reference/operators/pointer-related-operators.md
@@ -36,7 +36,7 @@ For information about pointer types, see [Pointer types](../../programming-guide
 > [!NOTE]
 > Any operation with pointers requires an [unsafe](../keywords/unsafe.md) context. The code that contains unsafe blocks must be compiled with the [`-unsafe`](../compiler-options/unsafe-compiler-option.md) compiler option.
 
-## Address-of operator &amp;
+## <a name="address-of-operator-amp"></a> Address-of operator &amp;
 
 The unary `&` operator returns the address of its operand:
 
--- a/docs/csharp/language-reference/operators/pointer-related-operators.md
+++ b/docs/csharp/language-reference/operators/pointer-related-operators.md
@@ -36,7 +36,7 @@ For information about pointer types, see [Pointer types](../../programming-guide
 > [!NOTE]
 > Any operation with pointers requires an [unsafe](../keywords/unsafe.md) context. The code that contains unsafe blocks must be compiled with the [`-unsafe`](../compiler-options/unsafe-compiler-option.md) compiler option.
 
-## <a name="address-of-operator-amp"></a> Address-of operator &amp;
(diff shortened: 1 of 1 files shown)


docs/csharp/language-reference/operators/pointer-related-operators.md
```diff
@@ -36,7 +36,7 @@ For information about pointer types, see [Pointer types](../../programming-guide
 > [!NOTE]
 > Any operation with pointers requires an [unsafe](../keywords/unsafe.md) context. The code that contains unsafe blocks must be compiled with the [`-unsafe`](../compiler-options/unsafe-compiler-option.md) compiler option.
 
-## Address-of operator &amp;
+## <a name="address-of-operator-amp"></a> Address-of operator &amp;
```
pkulikov: (1) Shouldn't it be on its own line?
(2) Updated the bookmark name to match one used throughout the docs (see line 28 of this file)
```suggestion
<a name="address-of-operator-"></a>
## Address-of operator &amp;
```

v-maudel:
Hi @pkulikov , 
Actually, it should be all in the same line. When the transformer tool to convert it to HTML and localized languages process the file, detect the \#\# already as a bookmark, so if we split the reference and the name of the subtitle, it will create two links. Also, the issue with the bookmark itself is the transformer tool, create in the html format the bookmark (with the \<a> tags), the thing is it utilizes the subtitle name (Address-of operator \&amp\;) as name attribute and it can't contain spaces. 

Your second change is fine, we didn't notice that

pkulikov:
> Actually, it should be all in the same line. When the transformer tool to convert it to HTML and localized languages process the file, detect the ## already as a bookmark, so if we split the reference and the name of the subtitle, it will create two links.

@v-maudel thanks for clarification; it's clear; I've updated suggestion.

docs/csharp/language-reference/operators/pointer-related-operators.md
pkulikov: Another option is to try out the following H2 title:
```text
## Address-of operator &
```

Would it work? I guess it would generate the same bookmark text in both original and translated versions.

v-maudel:
Thansk to you @pkulikov , 
I've also submitted another similar (matched the previous references) request here: https://github.com/dotnet/docs/pull/13012

Regards.

docs/csharp/language-reference/operators/pointer-related-operators.md
rpetrusha: I think that the solution proposed by @pkulikov should work, @v-maudel. That's the bookmark that's automatically generated by our builds, and also the bookmark that's referenced in three additional files.

v-maudel:
Hi @rpetrusha 
Could you please approve this PR as well if agree?
Regards.

rpetrusha (approved):
This looks good, so I'll merge, @v-maudel. Thanks for your help, @pkulikov.

PR #13043, opened 2019-06-26
fix: docs/standard/serialization/system-xml-serialization-element.md

nschonni:
- neuutral -> neutral
- sxml -> xml
- Trim leading/trailing spaces


rpetrusha (approved):
Thanks for these changes, @nschonni. I've made some additional changes to what is supposed to be the assembly qualified name of the type. I'll merge when the build completes.

Issue #12924, opened 2019-06-20
Abridged Reference: Operators / operator overloading


pkulikov:
There are a number of articles in the current doc set that can be consolidated in the two following articles.

**Operator overloading**

This article should introduce the `operator` keyword and list the overloadable operators. The following articles are to be substituted with the new article:

- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/operator
- https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/overloadable-operators

**User-defined conversion operators**

This article should show how to define custom conversion operators with the `implicit` and `explicit` keywords. The following articles (they share a lot now) are to be substituted with the new article:

- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/implicit
- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/explicit
- https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/conversion-operators
- https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/using-conversion-operators
- https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-implement-user-defined-conversions-between-structs

@BillWagner what do you think? If that sounds like a fine plan, please assign this issue to me.

BillWagner:
@pkulikov This is a great plan.  Thank you.

I've assigned it to you, and added it to our next sprint. (I added myself as an additional assignee so this stays in my daily list to 👀 at.

PR #13049, opened 2019-06-26
Equality operators: added section about delegate equality

pkulikov:
The [language spec](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#delegate-equality-operators) contains a good explanation of the delegate equality. So, the new section mostly provides an example and references further the spec.

Language spec is clear, except one point (emphasis mine):
> Invocation list entries produced from evaluation of semantically identical anonymous_method_expressions or lambda_expressions with the same (possibly empty) set of captured outer variable instances **are permitted (but not required)** to be equal.

That's why one more example to show that identical lambdas result in not equal delegate instances.

Supported by dotnet/samples#1018

@BillWagner, on a side note, have you had a chance to take a look at #12924: it suggests further possible consolidations of the operators reference.

### DIFF
--- a/docs/csharp/language-reference/operators/equality-operators.md
+++ b/docs/csharp/language-reference/operators/equality-operators.md
@@ -1,7 +1,7 @@
 ---
 title: "Equality operators - C# reference"
-description: "Learn about C# equality comparison operators."
-ms.date: 03/28/2019
+description: "Learn about C# equality comparison operators and C# type equality."
+ms.date: 06/26/2019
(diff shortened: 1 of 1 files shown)


rpetrusha:
Closing and reopening to begin new build after dotnet/samples#1018.

BillWagner (approved):
These changes LGTM @pkulikov 

I've reviewed everything, and I'll :shipit: now.

Thanks again.

PR #13099, opened 2019-06-27
typo: encorder -> encoder

nschonni:


rpetrusha (approved):
I've added an article before the now correctly spelled encoder. I'll merge when the build completes, @nschonni.

PR #13100, opened 2019-06-27
typo: Condtion -> Condition

nschonni:


rpetrusha:
@HongGit, do you know whether SAMLUnableToLoadConditions and SAMLUnableToLoadCondition are actually the same exception, or whether they are separate exceptions?

rpetrusha:
Thanks for catching these, @nschonni. I'm wondering whether this is an accidental duplication with a typo, or whether there actually are two similar exceptions with the same exception description. For the moment, I mark this as waiting on feedback.

HongGit:
Unfortunately, in IdentityModel\System\IdentityModel\Tokens\SamlAssertion.cs, it is referencing "SAMLUnableToLoadCondtions".  Here is the line of code:

_throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new SecurityTokenException(SR.GetString(SR.**SAMLUnableToLoadCondtions**)));_

jzeferino (approved):

jzeferino (changes requested):
`SAMLUnableToLoadCondtions` should be fixed in separated PR.

rpetrusha:
Thanks, @HongGit. I'll merge your PR now, @nschonni.

rpetrusha (approved):

HongGit:
@rpetrusha sorry, I was not explicitly clear.  Could we revert this change?  The source code was referencing SAMLUnableToLoad**Condtion**s.  Update the spelling will break the reference, so this change need to be reverted.

rpetrusha:
Ah. Sorry about that, @HongGit. I'll open a PR to revert the change right now.

rpetrusha:
Should the other line be fixed, @HongGit, or does that also need to be reverted, or should it be deleted:
image
I don't see it in SAMLAssertion.cs on index.

HongGit:
@rpetrusha I think both should be fixed.  you could find the file at https://github.com/microsoft/referencesource/tree/master/System.IdentityModel/System/IdentityModel/Tokens.

From the source code resource file, there is only one entry for **SAMLUnableToLoadCondtion**, not sure why there are two entries in this file?.

HongGit:
@rpetrusha there should be one **SAMLUnableToLoadCondtion** and one **SAMLUnableToLoadCondtions**.

rpetrusha:
OK. I'll revert the second change as well, @HongGit.

PR #13143, opened 2019-06-27
fix: docs/framework/windows-workflow-foundation/how-to-query-for-non-persisted-instances.md

nschonni:
- Suspsend -> Suspend
- add code fence language
- trim trailng/leading space


mairaw (approved):
👍

PR #13144, opened 2019-06-27
fix: docs/framework/winforms/controls/how-to-change-styles-on-an-element-in-the-managed-html-document-object-model.md

nschonni:
- hander -> handler
- convert to code fence language
- trim trailing/leading space


mairaw (approved):
Great changes. Thanks!

PR #13164, opened 2019-06-28
Replace IIS 7.0 tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for continuing to replace tokens with token text. I've changed the link text that included "Beta" in the title to the current title of those articles, so I'll merge your PR as soon as the internal build completes successfully.

PR #13165, opened 2019-06-28
Replace CardSpace tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for removing another token. I've change "an CardSpace" to "a CardSpace" in a few places and added some missing language identifiers, so I'll merge when the internal build completes.

PR #13166, opened 2019-06-28
Replace LINQ to ADO.NET tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for replacing the token with its token text and removing the token. I'll merge your PR now.

PR #13170, opened 2019-06-28
Revert #13100

rpetrusha:
## Revert #13100

This PR reverts a spelling correction that causes source and documentation to not correspond. (The misspelling is found in source.) I've added a comment (which I don't think will appear in the build) so that we don't inadvertently correct the misspelling again.

@HongGit @nschonni


BillWagner (approved):
This LGTM @rpetrusha 

You can :shipit: when ready.

nschonni:
PS: The other way might be to add the suppression here https://github.com/dotnet/docs/blob/master/cSpell.json#L997. I find that doesn't always supress it in VS Code, but it doesn't pick up from the commandline then

HongGit:
@rpetrusha I would suggest to update both **SAMLUnableToLoadCondtion** and **SAMLUnableToLoadCondtions**.

rpetrusha:
Doesn't putting it in the global file affect all misspellings of condition, @nschonni? It might be best to just add a comment to each item rather than to suppress what could be a common typo.

HongGit (approved):

nschonni:
> Doesn't putting it in the global file affect all misspellings of condition

The "overrides" section allows you to apply/ignore a word for a file or file glob

PR #13178, opened 2019-06-29
Replace Microsoft Word tokens

nxtn:


BillWagner (approved):
Thanks again @NextTurn 

I’ll :shipit: now.

PR #13195, opened 2019-07-01
Equality operators: reordered sections about reference type equality

pkulikov:
Moved the "Reference types equality" section before the "String equality" section. Also, reworded the "Reference types equality" section in order not to give impression that only strings are compared differently.


BillWagner (approved):
This is a well-thought out update. It does improve clarity in our explanation on equality.

I’ll :shipit: now.

Thanks again @pkulikov

PR #13198, opened 2019-07-01
Replace GDI+ tokens

nxtn:


rpetrusha (approved):
Thank you, @NextTurn, for replacing this token with its token text and removing the token. I'll merge your PR now.

PR #13215, opened 2019-07-02
Replace LINQ to Entities tokens

nxtn:


rpetrusha (approved):
Thank you, @NextTurn, for replacing another token with its token text and removing the token. I'll merge your PR now.

PR #13244, opened 2019-07-04
Replace SQL Server Compact 3.5 tokens

nxtn:


BillWagner (approved):
Thanks again @NextTurn 

I’ll :shipit: now.

PR #13251, opened 2019-07-05
Replace 2-D, 3-D tokens

nxtn:


mairaw (dismissed):
Do you also want to remove another include called tlasharptla-3d-md.md at the same time? It seems a repeat and is used at the same topics. That way, we don't have to resend the same change over to localization multiple times.

NextTurn:
Of course @mairaw.

BillWagner (approved):
Reviewed and approved. 

This PR now includes the changes that @mairaw had requested in her review.

Thanks @NextTurn I'll :shipit: now.

PR #13231, opened 2019-07-03
Incorrect image referenced in Dockerfile

Dev-Mom:
See this issue:  https://github.com/dotnet/dotnet-docker/issues/1112.
And this issue:  https://github.com/aspnet/aspnet-docker/issues/322.

## Summary

Describe your changes here.

Fixes #Issue_Number (if available)

PR #13278, opened 2019-07-08
Revert "Incorrect image referenced in Dockerfile"

Thraka:
Reverts dotnet/docs#13231

@TracyGH Can you clarify why you got the error in the original issue? This tutorial doesn't use ASP.NET at all and the original command works as intended if you're following the tutorial.

### DIFF
--- a/docs/core/docker/build-container.md
+++ b/docs/core/docker/build-container.md
@@ -173,7 +173,7 @@ The *Dockerfile* file is used by the `docker build` command to create a containe
 In your terminal, navigate to up a directory to the working folder you created at the start. Create a file named *Dockerfile* in your working folder and open it in a text editor. Add the following command as the first line of the file:
 
 ```dockerfile
-FROM mcr.microsoft.com/dotnet/core/aspnet:2.2
+FROM mcr.microsoft.com/dotnet/core/runtime:2.2
 ```
 
 The `FROM` command tells Docker to pull down the image tagged **2.2** from the **mcr.microsoft.com/dotnet/core/runtime** repository. Make sure that you pull the .NET Core runtime that matches the runtime targeted by your SDK. For example, the app created in the previous section used the .NET Core 2.2 SDK and created an app that targeted .NET Core 2.2. So the base image referred to in the *Dockerfile* is tagged with **2.2**.


TracyGH:
@Thraka - When I tried to complete the tutorial, I received the error shown in the GitHub issues referenced in the original PR.  Both threads suggested updating 'runtime' to 'aspnet' and that fixed it for me.  Below is the project I used to complete the tutorial.
image

Thraka:
Hi! I can see how you got that error, because you created an ASP.NET Core app, but the tutorial is built on using a .NET Core app.

Perhaps if I can understand why you used the tutorial but didn't follow it, I can update the article to help solve the problem you ran into. Did you perhaps arrive at this tutorial because you had an existing ASP.NET Core website you want to dockerize?

TracyGH:
@Thraka - Also, [Docker's version of the same tutorial](https://docs.docker.com/engine/examples/dotnetcore/) appears to use 'aspnet' rather than 'runtime'.

image

mairaw (approved):
Approving while you figure out what's wrong 😄

Thraka:
@TracyGH Did you see the ASP.NET Core tutorial on docker? https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/docker/building-net-docker-images?view=aspnetcore-2.2

This tutorial is for .NET Core. I would think that the article I just linked to and docker's article are the same. My article here deals with .NET Core, specifically a console app and not a web app. For a console app you would use the `runtime` image as stated in this tutorial.

TracyGH:
Hi @Thraka - Thank-you for the clarification!  Yes, that seems to be the issue.  I was using the 2 (.NET Core & ASP.NET Core) interchangeably, and Docker appears to do the same thing on their tutorial (see image below).  
For me, I think it may be a keyword search issue.  When Googling for things like 'asp.net docker container', the .NET Core article shows up but not the asp.net relevant one you referenced above.  

image

Thraka:
OK. I thought that was probably the issue but I wanted to make sure. It's a problem that won't be fixed until [.NET 5 is released](https://devblogs.microsoft.com/dotnet/introducing-net-5/) and the Core designation is dropped :)

Thanks for working with me on making sure I understood what happened!

PR #13296, opened 2019-07-09
Adjust section headers levels

pkulikov:
@BillWagner I don't know how much it was by design, but adjusted headers seem to be the part of a containing section about integral literals.


BillWagner (approved):
Good catch @pkulikov 

I think I’d had a different outline in mind, then didn’t update the headers as I filled in the outline.

I’ll :shipit: now.

PR #13299, opened 2019-07-09
fix: MD001/heading-increment/header-increment

nschonni:
Heading levels should only increment by one level at a time


BillWagner (approved):
Thanks again @nschonni 

I’ll :shipit: now.

PR #13308, opened 2019-07-09
Replace Object Relational Designer tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for replacing two tokens with their token text. I'll merge your PR now.

PR #13318, opened 2019-07-09
C# versioning update

### LINKED ISSUE #8192: List of language features and framework versions
Hi,
Is there anywhere a list of language features, and the corresponding versions of .NET / .NET Core / Netstandard as target platform which are needed to use the feature?
Also, the required SDK versions would be useful.
This list could also include extension packages like the Async Targeting Pack, which allow to circumvent some of those restrictions.
I think a list of this kind (or a pointer to it, if it already exists elsewhere) would be very useful in this article.

Thanks!

---
#### Document Details

* ID: 7d0be2e3-8f90-8843-f98c-aadba19bec6d
* Version Independent ID: 012620d5-fa79-cc15-341b-3da8e5748890
* Content: [The relationship between language features and library types](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/relationships-between-language-and-library)
* Content Source: [docs/csharp/whats-new/relationships-between-language-and-library.md](https://github.com/dotnet/docs/blob/master/docs/csharp/whats-new/relationships-between-language-and-library.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

cartermp:
Aligning with changes for 16.3 preview 1/.NET Core 3.0 feature complete

Fixes #8192 

Edit by @BillWagner to refer to issue being addressed.


docs/csharp/language-reference/configure-language-version.md
```diff
@@ -1,64 +1,69 @@
...
 
-This capability decouples the decision to install new versions of the SDK and tools in your development environment from the decision to incorporate new language features in a project. You can install the latest SDK and tools on your build machine. Each project can be configured to use a specific version of the language for its build. The default behavior means that any language features that rely on new types or new CLR behavior are enabled only when projects target those frameworks.
+The compiler determines a default based on these rules:
 
-You can override the default behavior by specifying a language version. There are several ways to set the language version:
+|Target framework|version|C# language version default|
+|----------------|-------|---------------------------|
+|.NET Core|3.x|C# 8.0|
+|.NET Core|2.x|C# 7.3|
+|.NET Standard|all|C# 7.3|
```
BillWagner: Why wouldn’t .NET Standard 2.1 select C# 8.0?

docs/csharp/language-reference/configure-language-version.md
```diff
@@ -1,64 +1,69 @@
...
+
+```xml
+<Project>
+ <PropertyGroup>
+   <LangVersion>preview</LangVersion>
+ </PropertyGroup>
+</Project>
+```
+
+Now, builds in every subdirectory of the directory containing that file will use C# version 7.3 syntax. For more information, see the article on [Customize your build](/visualstudio/msbuild/customize-your-build).
```
BillWagner: Why would 7.3 be selected from “preview”? Is there a mis-match between the example and the text?

docs/csharp/language-reference/configure-language-version.md
```diff
@@ -1,64 +1,69 @@
...
+
+### Set the langversion compiler option
+
+You can use the `-langversion` command-line option. For more information, see the article on the [-langversion](../language-reference/compiler-options/langversion-compiler-option.md) compiler option. You can see a list of the valid values by typing  `csc -langversion:?`.
```
BillWagner: Hmm. `csc` won’t be available on a Mac or Linux system with only the .NET Core SDK installed. Should we have an equivalent `dotnet build` option?

BillWagner (changes requested):
Overall, this is great @cartermp 

I had a couple questions, and found one issue that should be fixed. 

Then, it’s ready to :shipit:

docs/csharp/language-reference/configure-language-version.md
cartermp: Rather complicated reasons, but the long and short of it is there's no way to actually enforce any kind of default for any arbitrary target that implements a .NET Standard. Hence it is 7.3.

docs/csharp/language-reference/configure-language-version.md
cartermp: I wonder if it's worth having this section given this, actually. Generally we'll be messaging LangVersion.

BillWagner: @cartermp I think that’s the best resolution: I’d delete this paragraph.

BillWagner (approved):

PR #13321, opened 2019-07-10
Correct usage of async void.

Youssef1313:
The edit I made was also said in a NDC Sydney conference by a Microsoft developer and I think it's correct.
https://youtu.be/av5YNd4X3dY?t=899
(from time 15:00 to 15:12)

### DIFF
--- a/docs/csharp/language-reference/keywords/async.md
+++ b/docs/csharp/language-reference/keywords/async.md
@@ -65,7 +65,7 @@ An async method can have the following return types:
 
 - <xref:System.Threading.Tasks.Task>
 - <xref:System.Threading.Tasks.Task%601>
-- [void](../../../csharp/language-reference/keywords/void.md), which should only be used for event handlers or only call it from the main thread.
+- [void](../../../csharp/language-reference/keywords/void.md). `async void` methods are generally discouraged for code other than event handlers because callers cannot `await` those methods and must implement a different mechanism to report successful completion or error conditions.
 - Starting with C# 7.0, any type that has an accessible `GetAwaiter` method. The `System.Threading.Tasks.ValueTask<TResult>` type is one such implementation. It is available by adding the NuGet package `System.Threading.Tasks.Extensions`. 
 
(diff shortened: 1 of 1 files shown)


stephentoub:
Both the existing guidance and the update could easily be construed as law rather than guidance (I also don't really understand the update).  `async void` can absolutely be misused, and its use should be limited, but in docs like this we should avoid absolutes, as there are always gray areas.

Youssef1313:
> 
> 
> Both the existing guidance and the update could easily be construed as law rather than guidance (I also don't really understand the update). `async void` can absolutely be misused, and its use should be limited, but in docs like this we should avoid absolutes, as there are always gray areas.

As far as I know, The `misuse` of `async void` is when it's called from a thread different from the main thread. and `main thread` isn't limited to event handlers only. That's what the update is.

stephentoub:
> The misuse of async void is when it's called from a thread different from the main thread.

No, it's valid to use `async void` from any thread.

docs/csharp/language-reference/keywords/async.md
```diff
@@ -65,7 +65,7 @@ An async method can have the following return types:
 
 - <xref:System.Threading.Tasks.Task>
 - <xref:System.Threading.Tasks.Task%601>
-- [void](../../../csharp/language-reference/keywords/void.md), which should only be used for event handlers.
+- [void](../../../csharp/language-reference/keywords/void.md), which should only be used for event handlers or only call it from the main thread.
```
BillWagner: How about this as a different update:

```suggestion
- [void](../../../csharp/language-reference/keywords/void.md). `async void` methods are generally discouraged because callers cannot `await` those methods and must implement a different mechanism to report successful completion or error conditions.
```

stephentoub: That's fine.  I'd suggest changing "are generally discouraged" to "are generally discouraged for code other than event handlers"

BillWagner: updated. Thanks @stephentoub. 

@Youssef1313 If you commit this suggestion, I’ll :shipit:

Youssef1313:
@BillWagner, Thanks for your update, I see it's better.
@stephentoub, Thanks for your review and explanation. (btw, Your pause implementation for async helped me!)

BillWagner (approved):
Thanks for updating this @Youssef1313 and participating in the discussion on this update.

I’ll :shipit: now.

PR #13328, opened 2019-07-10
Replace Active Accessibility tokens

nxtn:


rpetrusha (approved):
Thank you, @NextTurn, for continuing to replace tokens with their token text and remove the unused tokens. I'll merge your PR now.

PR #13054, opened 2019-06-26
Reference consolidation numeric types

BillWagner:
This contributes to #10728 (The next PR will do the same with the floating point types.)

Notes for reviewers: 

docs/csharp/language-reference/builtin-types/integral-numeric-types.md is the only new file.  The redirection file and the Language Reference TOC are the only other files with major changes. The other changed files were updated for links to files that have been removed.

[Internal review link](https://review.docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types?branch=pr-en-us-13054)

Other changes to test:
- The Language reference TOC should have "displayName" attributes for all integral types. 
- Build warnings for redirects and updated links.

PR #13333, opened 2019-07-10
Update value-types-table.md

hillmarcus:
Removed a typo causing an extra row to be in the type table

### DIFF
--- a/docs/csharp/language-reference/keywords/value-types-table.md
+++ b/docs/csharp/language-reference/keywords/value-types-table.md
@@ -18,7 +18,6 @@ The following table shows the C# value types:
 |[bool](bool.md)|Boolean||
 |`byte`|Unsigned, numeric, [integral](../builtin-types/integral-numeric-types.md)||
 |[char](char.md)|Unsigned, numeric, [integral](../builtin-types/integral-numeric-types.md)
-)||
 |`decimal`|Numeric, [floating-point](../builtin-types/floating-point-numeric-types.md)|M or m|
 |`double`|Numeric, [floating-point](../builtin-types/floating-point-numeric-types.md)|D or d|
 |[enum](enum.md)|Enumeration||


Youssef1313 (changes requested):
Some of the table records has hyperlinks to redirect to its corresponding docs page, but others don't.
I suggest you add hyperlinks to those who don't already have.

hillmarcus:
The ones that have links in the first column have pages in the `language-reference/keywords` folder, while those that don't have links are missing those pages. The links in the second column contain similar information only those link to the `language-reference/builtin-types` folder.

I'm not going to make any more changes for this pull request. Adding additional keyword pages should be another task for someone more involved in the project.

Youssef1313:
I think it won't require adding any additional pages, I didn't search, but there should be a page somewhere in the docs about `byte` for example which don't have hyperlink, even if not in the `language-reference/keywords`. Anyway, I may add that change later if you're interested in it.

pkulikov:
> there should be a page somewhere in the docs about `byte`

@Youssef1313 not anymore; PR #13054 has combined information about individual integral numeric types into one article, which is referenced in the table under consideration in the third column. So, I think missing hyperlinks are by design.

BillWagner:
@hillmarcus @Youssef1313 Thanks for creating this, and the discussion.  As @pkulikov points out, you’re seeing the content during a time when some reorganization is happening.  [This project](https://github.com/dotnet/docs/projects/71) contains issues that maps out the proposed reorganization of the C# reference. 

There’s also more details in [this gist](https://gist.github.com/BillWagner/047cb97f6c0fe542d3a1299e994b2acd) for the entire project plan for an updated C# Guide.

As a result of some of these changes, the individual pages for many of the built in types are now consolidated (or in progress for that). 

Thanks for helping us improve!

BillWagner (approved):
Thanks for making this change @hillmarcus 
I'll :shipit: now.

PR #13342, opened 2019-07-11
Clarify "abstract"

### LINKED ISSUE #11840: clarification on virtual and abstract as applied to methods
“The virtual keyword is used to modify a method, property, indexer, or event declaration and allow for it to be overridden in a derived class.”
Very top of the page: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/virtual

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/abstract
	“Members marked as abstract, or included in an abstract class, must be implemented by classes that derive from the abstract class.”

A couple of paragraphs down that page: “An abstract method is implicitly a virtual method.”

So, an abstract method MUST be implemented by an subclass.  But an abstract method is implicitly virtual.  And a virtual method can be overridden by the subclass, but doesn’t necessarily have to be.

What am I missing?

---
#### Document Details

* ID: 89022fe7-b2e3-5fb5-4930-d6bc5284081a
* Version Independent ID: 6a4f3ec1-e2eb-e046-d840-1c442276b3fb
* Content: [virtual - C# Reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/virtual#feedback)
* Content Source: [docs/csharp/language-reference/keywords/virtual.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/keywords/virtual.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

geoff-m:
Clarify that only non-abstract classes must implement abstract members.

## Summary

Reduces observed confusion about the meaning of "abstract".

Fixes #11840

### DIFF
--- a/docs/csharp/language-reference/keywords/abstract.md
+++ b/docs/csharp/language-reference/keywords/abstract.md
@@ -11,7 +11,7 @@ helpviewer_keywords:
 ms.assetid: b0797770-c1f3-4b4d-9441-b9122602a6bb
 ---
 # abstract (C# Reference)
-The `abstract` modifier indicates that the thing being modified has a missing or incomplete implementation. The abstract modifier can be used with classes, methods, properties, indexers, and events. Use the `abstract` modifier in a class declaration to indicate that a class is intended only to be a base class of other classes, not instantiated on its own. Members marked as abstract must be implemented by classes that derive from the abstract class.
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks for spotting this and adding this clarifying statement @geoff-m We appreciate it.

You should see the changes on our live site in the next day or two on our regular publishing cycle. Thanks again for making your first contribution to docs! 🎆

PR #13349, opened 2019-07-11
Remove outdated sample

### LINKED ISSUE #11770: Adding null delegates to dictionary costs more memory than field-like events would
The sample initializes the dictionary with null delegates as values:

```C#
    public PropertyEventsSample()
    {
        handlers = new Dictionary<string, Delegate>
        {
            [Event1Key] = null,
            [Event2Key] = null
        };
    }
```

This means the dictionary will always have an entry for each event, even if no event handlers are ever added. These entries cost even more memory than field-like events would, because each entry also keeps track of the key, hash code, and next entry in the chain. It is not useful in the situation claimed here:

> This is only useful if a type has many events, but you expect that most of the events will not be subscribed to.

To be actually useful, the entries would have to be allocated on demand, like `System.ComponentModel.EventHandlerList` does.

---
#### Document Details

* ID: 0e2cb228-430f-43a7-4447-85d21d2de420
* Version Independent ID: 2af9e215-97f2-fb66-1090-591e83fbda92
* Content: [How to: use a dictionary to store event instances - C# Programming Guide](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/events/how-to-use-a-dictionary-to-store-event-instances#feedback)
* Content Source: [docs/csharp/programming-guide/events/how-to-use-a-dictionary-to-store-event-instances.md](https://github.com/dotnet/docs/blob/master/docs/csharp/programming-guide/events/how-to-use-a-dictionary-to-store-event-instances.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
This sample demonstrates a practice that should not be used today. Instead, users should follow the practice in the article being redirected to on using the `EventHandlerList`

Fixes #11770


rpetrusha (commented):
This looks good, @BillWagner. You can merge when you're ready.

rpetrusha (approved):

PR #13351, opened 2019-07-11
Replace APIs tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for removing the two APIs tokens. (It's interesting that a token would be used at all.) I'll merge your PR now.

PR #13358, opened 2019-07-11
CC140377: Correcting comment syntax inside code-block

v-maudel:
Description: comment syntax inside code-block doesn't work when the example is used.


v-maudel:
@BillWagner , can you please, review this change and approve it if agree?
Thanks in advance.

BillWagner (approved):
This LGTM @v-maudel 

I’ll :shipit: now.

v-maudel:
Thanks  @BillWagner !

PR #13359, opened 2019-07-11
updating vsmac getting started

timeyoutakeit:


BillWagner (approved):
Everything LGTM @asb3993 

I'll :shipit: now.

We publish daily, so you should see these updates on the live site by Monday.

Thanks for making these updates.

Issue #10741, opened 2019-02-26
Abridged Reference: Operators / other operators and tokens


BillWagner:
These don't fit any group, and should have a single page for each operator.

- `?` `:` (conditional operator). Somehow, it feels out of tune with the logical operators.
- `??` (null coalescing). It also feels not in the logical operations group. Most probably, it's because I limit logical operations to those on `bool` operands.
- `+`, `+=` (string concatenation and delegate combination)
- `-`, `-=` (delegate removal)
- `::` (namespace alias qualifier)
- `nameof`
- `sizeof`
- `await`
- `new` operator
- `stackalloc` operator
- `=>` (lambda declaration)

Sub-tasks are:

- [ ] Create the new articles
- [ ] Search for existing issues on content being migrated and consolidated. This is a good opportunity to fix those.
- [ ] Redirect existing content to the new articles.
- [ ] Remove old articles.
- [ ] Update all links in the .NET Content

BillWagner:
Operators that were missing in the first review that should be added:

- `new` operator
- `stackalloc` operator

pkulikov:
@BillWagner to confirm: the listed operators should have a separate page for each of them.

Given that, I  suggest (they are already merged in the list in the issue text) to merge pages for the `+` and `+=` operators (`-` and `-=` pair as well) and put focus at these pages on the non-arithmetic behavior.

BillWagner:
@pkulikov That would be perfect.

pkulikov:
@BillWagner could you please assign this issue to me, as the half of it is already done.

The remaining operators to revise and move to the operators' TOC:

- [ ] `::` (namespace alias qualifier)
- [ ] `nameof`
- [ ] `sizeof`
- [ ] `await`
- [ ] `new` operator

Not in the original list, but should the following two also be included?

- [ ] `default`
- [ ] `delegate`

PR #13364, opened 2019-07-12
Moved and revised the nameof operator article

pkulikov:
Contributes to #10741
Supported by dotnet/samples#1084
@BillWagner this revision makes the article much shorter and more to the point.


docs/visual-basic/getting-started/whats-new.md
```diff
@@ -180,7 +180,7 @@ For more information, see [Reference Return Values](../programming-guide/languag
 
 ## Visual Basic 14
 
-[Nameof](../../csharp/language-reference/keywords/nameof.md)
+[Nameof](../../csharp/language-reference/operators/nameof.md)
```
pkulikov: It seems there is no Visual Basic reference article for the `nameof` operator.

BillWagner: /cc @rpetrusha Do you have a better link for this?

docs/csharp/language-reference/operators/nameof.md
```diff
@@ -0,0 +1,29 @@
...
+ms.custom: seodec18
+ms.date: 07/12/2019
+f1_keywords:
+  - "nameof_CSharpKeyword"
+  - "nameof"
+ms.assetid: 33601bf3-cc2c-4496-846d-f9679bccf2a7
+---
+# nameof operator (C# reference)
+
+The `nameof` operator obtains the simple name of a variable, type, or member as the string constant:
```
BillWagner: We should define "simple name". Suggestion:

```suggestion
The `nameof` operator obtains the simple name of a variable, type, or member as the string constant. A [**simple name**](~/_csharplang/spec/expressions#simple-names) is the identifier name. In the case of a generic definition, it is the identifier name followed by any type parameters.
```

BillWagner (changes requested):
This looks great @pkulikov 

I had one suggestion for defining "simple name". I'm open to other suggestions there, then this is ready to :shipit:

docs/csharp/language-reference/operators/nameof.md
pkulikov: @BillWagner I hoped the example would substitute the definition, which is quite difficult to craft.

For example, from the suggestion:
>In the case of a generic definition, it is the identifier name followed by any type parameters.

I would assume that type parameters are included, but the following code
```csharp
Console.WriteLine(nameof(List<int>));
```

produces 
```console
List
```

I've looked at the spec now. It doesn't actually say "simple name", but just a "name". So i suggest to remove the word "simple":
>The `nameof` operator obtains the name of a variable, type, or member as the string constant:

pkulikov: @BillWagner I think I was wrong with the word choice in the first version of the article. Could you please take a look at the updated version.

Additionally, I've added the language version, in which the operator was added, and the link from the "What's new..." section.

docs/csharp/language-reference/operators/nameof.md
```diff
@@ -0,0 +1,33 @@
...
+  - "nameof"
+ms.assetid: 33601bf3-cc2c-4496-846d-f9679bccf2a7
+---
+# nameof operator (C# reference)
+
+The `nameof` operator obtains the name of a variable, type, or member as the string constant:
+
+[!code-csharp-interactive[nameof operator](~/samples/csharp/language-reference/operators/NameOfOperator.cs#Examples)]
+
+As the preceding example shows, in the case of a type and a namespace, the produced name is not [fully qualified](~/_csharplang/spec/basic-concepts.md#fully-qualified-names).
```
pkulikov: @BillWagner I'm not technically correct here. The sentence should finish with "...unless a type or a namespace is the member of the global namespace." However I didn't want to over-complicate things. Is the trimmed version fine?

BillWagner: I like the way you've worded this. I was looking through the spec this morning to find good language as well. I think a trimmed version is fine, but I'll suggest a bit of weasel words:

```suggestion
As the preceding example shows, in the case of a type and a namespace, the produced name is usually not [fully qualified](~/_csharplang/spec/basic-concepts.md#fully-qualified-names).
```

pkulikov: @BillWagner perfect! I've looked for that, but overlooked:) Thanks!

BillWagner (approved):

BillWagner:
I'm happy with how this turned out @pkulikov 

I'm ready to :shipit:, but I'm waiting to see if either @KathleenDollard or @rpetrusha have a better suggestion for the one link on the VB content.

docs/visual-basic/getting-started/whats-new.md
rpetrusha: There are two choices for the link, both of which should be temporary:
- The mention of `nameof` in https://docs.microsoft.com/en-us/dotnet/visual-basic/getting-started/whats-new#visual-basic-14.
- The design notes, at <https://github.com/dotnet/vblang/blob/189d2c15703b91a9241ce9ade46e6bc24b193cd9/meetings/2014/README.md>.
There's an outstanding issue to add the nameof operator for VB that's assigned to me for this sprint (though it was also assigned to me in the last sprint, and perhaps the one before that). I'll note in the issue that the what's new should be modified to link to the new nameof article when it's ready.

pkulikov: @rpetrusha the first suggested link would be a self-reference; the design notes, I'm afraid, would be more confusing than the existing link to the C# `nameof` article (design notes mix C# and VB examples). Thus, I suggest to keep the existing text as it is until the dedicated VB article is created. Please let me know if you prefer the link to the design notes, I'll update the text then.

BillWagner:
Since any link would be temporary, I'll :shipit: now, and we can make a better link choice when the VB article is available.

PR #13378, opened 2019-07-13
Replace ADO.NET tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for removing the ADO.NET token and replacing it with its token text. I'll merge your PR now.

PR #13379, opened 2019-07-13
Replace API tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for removing the API token. (I wonder why an API token was needed in the first place.) I'll merge your PR now.

PR #13392, opened 2019-07-15
C# numeric types: explicitly mention corresponding .NET types

pkulikov:
Integrated .NET types into the overview tables of the numeric types. That allows: (1) unload the **See also** sections, (2) mention alias/.NET type relation, which might be useful to the newcomers to C#.

Also, removed the sentence about `decimal` constants, as, based on this [Stack Overflow answer](https://stackoverflow.com/a/745638), it doesn't seem to be worth mentioning.


BillWagner (approved):
These are great changes @pkulikov I like the additions and the edits.

I'll :shipit: now.

PR #13408, opened 2019-07-16
Fix wording about C# aliases for .NET types

pkulikov:
As far as I understand, a C# type keyword IS an alias, so it should be that ".NET types and their aliases are interchangeable".


BillWagner (approved):
That is better wording @pkulikov 
I'll :shipit: now.

Thanks again for all your contributions.

PR #13478, opened 2019-07-19
Typo fix and minor edits

pkulikov:
The comments in code don't help to understand the subject of the section, so they can be removed.


BillWagner (approved):
Thanks @pkulikov 

These are good, clarifying changes.  I'll :shipit: now.

PR #13484, opened 2019-07-19
Replace Annotations Framework tokens

nxtn:


rpetrusha (approved):
Thank you, @NextTurn, for replacing these tokens with their token text and deleting the tokens. I'll merge your PR now.

PR #13488, opened 2019-07-19
Change incorrect verbiage

MadsTorgersen:
Change incorrect description of the difference between `Span<T>` and `Memory<T>`.

## Summary

Make clear that a `Span<T>` can _only_ be allocated on the stack, and that a `Memory<T>` can _also_ be allocated on the heap.


rpetrusha (approved):
Thanks for making this correction, @MadsTorgersen. I'll merge your PR now.

PR #13499, opened 2019-07-21
Replace ClickOnce tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for replacing the ClickOnce tokens with their token text. I'll merge your PR now.

PR #13504, opened 2019-07-22
Removed duplicative article about lambda expressions

pkulikov:
[This article](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-use-lambda-expressions-outside-linq) adds nothing new in comparison with the [main article about lambda expressions](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/lambda-expressions). So, I think it can safely removed.


BillWagner (approved):
Thanks for updating this @pkulikov 

I'll :shipit: now.

PR #4967, opened 2018-04-19
Coding convention: use string interpolation for short strings

pkulikov:
Snippet update: dotnet/samples#20

PR #13520, opened 2019-07-23
Replace ampersand with string interpolation

Youssef1313:
I think this change has been missed in https://github.com/dotnet/docs/pull/4967.


rpetrusha (approved):
Thanks for updating the guidelines to use string interpolation, @Youssef1313. I'll merge your PR now.

PR #13534, opened 2019-07-24
Added example

### LINKED ISSUE #13525: No info
This page gives no example or extra information, it simply rehashes the error text verbatim.

---
#### Document Details

* ID: ad02bbc1-1404-9b54-622a-3da00f2da70f
* Version Independent ID: 05492892-56f7-833c-06be-f7b1553416ff
* Content: [Handles clause requires a WithEvents variable defined in the containing type or one of its base types](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/error-messages/handles-clause-requires-a-withevents-variable-defined#feedback)
* Content Source: [docs/visual-basic/language-reference/error-messages/handles-clause-requires-a-withevents-variable-defined.md](https://github.com/dotnet/docs/blob/master/docs/visual-basic/language-reference/error-messages/handles-clause-requires-a-withevents-variable-defined.md)
* Technology: **vb-diagnostics**

Youssef1313:
Fixes #13525

### DIFF
--- a/docs/visual-basic/language-reference/error-messages/handles-clause-requires-a-withevents-variable-defined.md
+++ b/docs/visual-basic/language-reference/error-messages/handles-clause-requires-a-withevents-variable-defined.md
@@ -42,7 +42,6 @@ End Module
 
 The following example compiles successfully because the `_timer1` variable is defined with the `WithEvents` keyword:
 
-
 ```vb
 Imports System.Timers
 
--- a/docs/visual-basic/language-reference/error-messages/handles-clause-requires-a-withevents-variable-defined.md
+++ b/docs/visual-basic/language-reference/error-messages/handles-clause-requires-a-withevents-variable-defined.md
@@ -9,14 +9,31 @@ helpviewer_keywords:
 ms.assetid: 5b66f6a8-f050-4e03-a57f-a64e85f80cb5
 ---
 # Handles clause requires a WithEvents variable defined in the containing type or one of its base types
-You did not supply a `WithEvents` variable in your `Handles` clause. The `Handles` keyword at the end of a procedure declaration causes it to handle events raised by an object variable declared using the `WithEvents` keyword.  
+You did not supply a `WithEvents` variable in your `Handles` clause. The `Handles` keyword at the end of a procedure declaration causes it to handle events raised by an object variable declared using the `WithEvents` keyword.
   
- **Error ID:** BC30506  
+ **Error ID:** BC30506
+
+## To correct this error
   
-## To correct this error  
+- Supply the necessary `WithEvents` variable.
   
-- Supply the necessary `WithEvents` variable.  
-  
+## Example
+In the following example, we create a timer and use its `Tick` event to repeat some action. The error `BC30506` is produced when `WithEvents` isn't supplied to the timer variable.
+
+```vb
+Public Class Form1
+    ' This line will cause an error in the Handles clause.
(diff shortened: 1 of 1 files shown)


rpetrusha (changes requested):
Thanks, @Youssef1313, for adding this code example. I think that adding a code example here adds real value. However, I think that the example itself requires modification:
- Any necessary imports should be included.
- The example should be compilable and executable.
- If this were a Windows Forms application that were executable, the use of MessageBox in an event handler that's fired every second produces message boxes faster than the user is likely to be able to close them.

I'd recommend doing this as a console app instead with a Main method that just waits for any key to be pressed, and that uses the System.Threading.Timers.Timer class and its Elapsed event. The handler can then display a message to the console, and the example terminates when the user presses a key.

In addition, comments are not localized, which makes the fact that a lines of code need to be switched clear only to users who read English. That should be noticed in the example description.

Youssef1313:
Thanks, @rpetrusha for the suggested changes. I've updated the example. May you review the updates ?
Also, Unfortunately, I didn't understand this point well and how to deal with it:

> In addition, comments are not localized, which makes the fact that a lines of code need to be switched clear only to users who read English. That should be noticed in the example description.

docs/visual-basic/language-reference/error-messages/handles-clause-requires-a-withevents-variable-defined.md
```diff
@@ -9,14 +9,42 @@ helpviewer_keywords:
...
+- Supply the necessary `WithEvents` variable.
   
+## Example
+In the following example, we create a [System.Timers.Timer](https://docs.microsoft.com/dotnet/api/system.timers.timer) and use its `Elapsed` event to repeat some action. The error `BC30506` is produced when `WithEvents` isn't supplied to `_timer1` variable.
+
+```vb
+Imports System.Timers
+
+Module Module1
+    ' This line will cause an error in the Handles clause.
```
KathleenDollard: I think @rpetrusha comment was that this line, which is critical to understanding the code, will not be localized in the docs (code isn't) 

I would suggest that we leave in this code sample the correct version (the declaration that includes "WithEvents" and have a separate code fragment with a paragraph (not code) that says "This would not compile in the code below." or something to that effect. 

@rpetrusha  can clarify as well.

KathleenDollard (commented):
@Youssef1313 Thank you for your help with this! One comment.

rpetrusha:
I like your suggestion, @KathleenDollard. So we can have two code blocks for the example, @Youssef1313. The first should be the one that fails to compile, with the introductory text:

> In the following example, Visual Basic generates compiler error \`BC30506\` because the \[WithEvents](../modifiers/withevents.md) keyword is not used in the definition of the \<xref:System.Timers.Timer?displayProperty=nameWithType> instance.

The introduction to the second code block can then read:

> The following example compiles successfully because the \`_timer1\` variable is defined with the \`WithEvents` keyword:

Youssef1313:
Thanks, @rpetrusha & @KathleenDollard. I've updated the article with two separate examples.
May you review it now ?
And before merging, I'm not sure if `dev_langs` is missing and should be added ?
May you check that ?

rpetrusha (approved):
Thanks for making the additional changes, @Youssef1313. The dev_langs metadata element isn't needed here, since vb is the default. I'll merge your PR now. It should be live on docs.microsoft.com in the next day or two.

PR #13538, opened 2019-07-24
Replace C++ tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for replacing the C++ token with its text and deleting the two token files. I'll merge your PR now.

PR #13544, opened 2019-07-24
C103800: Changing italic asterisks by italic underscore

v-maudel:
Description: Reporting source issues as a result of nested asterisks (i.e. *** instead of **_) that are breaking code in localized sites.


rpetrusha (approved):
This looks good, @v-maudel. I'll merge your PR now.

v-maudel:
Thanks @rpetrusha

PR #13556, opened 2019-07-25
Eliminated fragmentary code for VB Char data type

rpetrusha:
## Eliminated fragmentary code for VB Char data type

I found the structureless code fragment to be jarring, so I felt compelled to fix it.


mairaw (approved):
Nice improvement! It makes the code much better to read.

PR #13560, opened 2019-07-25
[fsharp] Refactoring redundant path

Potapy4:
Contributes to #10653


mairaw (approved):
I like this a lot. Thanks @Potapy4!

PR #13579, opened 2019-07-26
Removed extra spaces and requirements section

Youssef1313:
I think that requirements section contain wrong information. Why I would need `Microsoft.VisualBasic.dll` to use the `Throw` keyword ?


rpetrusha (approved):
Thank you for making this correction, @Youssef1313. The Visual Basic Throw statement directly corresponds to the throw IL opcode and so does not require the use of the Visual Basic runtime library. I'll merge your PR now.

PR #13582, opened 2019-07-26
Replace Microsoft Direct3D tokens

nxtn:


rpetrusha (changes requested):
Thanks, @NextTurn. For continuing to replace tokens with their token text. I'll merge your PR now.

rpetrusha (approved):

PR #13587, opened 2019-07-26
Replace device-independent units tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for replacing this token with its text and removing the token. I'll merge your PR now.

PR #13592, opened 2019-07-26
Fix snippet code tags

pkulikov:
All other snippets in those files are tagged as F#


rpetrusha (approved):
Thanks for correcting the language identifiers, @pkulikov. I'll merge your PR now.

PR #13614, opened 2019-07-29
Replace DirectX tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for replacing the DirectX tokens with their token text and removing the tokens. I'll merge your PR now.

PR #13616, opened 2019-07-29
Extend and revise the default values table article

pkulikov:
The article has already contained information about the default values of reference and nullable value types. I've moved that into the table. Also, consolidate numerics-related rows.

### DIFF
--- a/docs/csharp/language-reference/keywords/default-values-table.md
+++ b/docs/csharp/language-reference/keywords/default-values-table.md
@@ -20,7 +20,7 @@ The following table shows the default values of C# types:
 |[char](char.md)|`'\0'` (U+0000)|
 |[enum](enum.md)|The value produced by the expression `(E)0`, where `E` is the enum identifier.|
 |[struct](struct.md)|The value produced by setting all value-type fields to their default values and all reference-type fields to `null`.|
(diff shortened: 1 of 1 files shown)


docs/csharp/language-reference/keywords/default-values-table.md
```diff
@@ -52,19 +34,23 @@ Beginning with C# 7.1, you can use the [`default` literal](../../programming-gui
 int a = default;
 ```
 
-You also can use the parameterless constructor or the implicit parameterless constructor to produce the default value of a value type, as the following example shows. For more information about constructors, see the [Constructors](../../programming-guide/classes-and-structs/constructors.md) article.
+For a value type, the implicit parameterless constructor also produces the default value of the type, as the following example shows:
```
pkulikov: @BillWagner Do I understand that correctly: a value type cannot have a user-defined, that is, explicit, parameterless constructor, so any parameterless constructor is implicit?

rpetrusha: That's correct, @pkulikov.

rpetrusha (approved):
These revisions look really good, @pkulikov. I'll merge your PR now.

PR #13618, opened 2019-07-29
Revised the introduction of the lambda expressions article

pkulikov:
Mostly reshuffled sentences to the following structure:
- What is a lambda expression and how to make it?
- What does a lambda expression produces/compiles to?
- Where can it be used?


BillWagner (approved):
I like this @pkulikov It's a better introduction to the topic.

Thanks again.

PR #13622, opened 2019-07-29
Replace GAC tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for removing these two tokens. I'll merge your PR now.

PR #13639, opened 2019-07-30
Replace GDI(+) tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for continuing to replace tokens with token text and for removing these four token files. I'll merge your PR now.

PR #13661, opened 2019-07-31
Replace GUI(s) tokens

nxtn:


rpetrusha (approved):
Thanks for replacing these three tokens with their token text and removing the tokens, @NextTurn. I'll merge your PR now.

PR #13672, opened 2019-07-31
Remove empty top-level node and shift indentation

ghogen:
This fix would "fix" the offline book issue - this TOC is not read in by the current version of the offline book tool, resulting in missing content.

## Summary

Remove the top line of the TOC and indent everything else.


ghogen:
outdated - opened new one

PR #13691, opened 2019-08-01
Update VectorType usage explanation

luisquintanilla:
Clarifying VectorType usage training speed impact based on feedback received.


natke (approved):
That was my bad.

PR #13706, opened 2019-08-02
Clarify that memory doesn't have to be freed MANUALLY.

ericmutta:
Saying that the garbage collector...

> Enables you to develop your application without having to free memory.

...may give the wrong impression that applications can allocate memory indefinitely and keep working even if they never free it...which is obviously not possible. I think the intention here was to say that objects will be freed automatically by the garbage collector, so you don't have to worry about it. I have edited the line to that effect.


rpetrusha (approved):
Thanks for adding this clarification, @ericmutta. I'll merge your PR now. The change should be live on docs.microsoft.com in the next day or two.

ericmutta:
@rpetrusha thanks for the fast follow up!

PR #13714, opened 2019-08-03
Fix delegate type

pkulikov:
The current text
> Note that the `Where` method in this example has an input parameter of the delegate type `Func<TResult>`...

doesn't specify the correct delegate type; fixed that into:
> Note that the `Where` method in this example has an input parameter of the delegate type `Func<T, TResult>`...

@BillWagner should [the how-to article from this PR](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-use-lambda-expressions-in-a-query) be redirected to the [Lambdas with the standard query operators](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/lambda-expressions#lambdas-with-the-standard-query-operators) section of the [Lambda expressions](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/lambda-expressions) article?


rpetrusha (approved):
Thanks for making this correction, @pkulikov. I'll merge your PR now.

In terms of the redirection, I don't think that it's necessary. The content is sufficiently different that this how-to can stand on its own.

pkulikov:
> In terms of the redirection, I don't think that it's necessary. The content is sufficiently different that this how-to can stand on its own.

@rpetrusha that's clear, thanks!

PR #13717, opened 2019-08-04
Replace IFD tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for replacing the IFD token with its token text. I'll merge your PR now.

PR #13765, opened 2019-08-08
Expression-bodied members: clarified logic

### LINKED ISSUE #13764: Article Contradicts Itself
"Note that expression can be a statement expression only if the member's return type is void, or if the member is a constructor, a finalizer, or a property set accessor."

And yet, in the above example, an expression bodied statement is used to define a member method that returns a string? Clarification is needed.

---
#### Document Details

* ID: a695fde2-da48-835f-423d-d34fb048d6ab
* Version Independent ID: e8eab8da-72ba-991a-c011-a41d2ed9939d
* Content: [=&gt; operator - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-operator#expression-body-definition)
* Content Source: [docs/csharp/language-reference/operators/lambda-operator.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/operators/lambda-operator.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

pkulikov:
Fixes #13764

The text can be made clearer. Now, it might be read (if one puts comma before "only") as if statement expressions can be used only with some kinds of members, with which other expressions also can be used. However, it's about members, not expressions. Some kinds of members allow *only* statement expressions; that said, statement expressions (any method call or `await` expression) still can be used with other kinds of members.

For example, from [the spec](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/classes#method-body):
>When a method has a void result and an expression body, the expression E must be a statement_expression

### DIFF
--- a/docs/csharp/language-reference/operators/lambda-operator.md
+++ b/docs/csharp/language-reference/operators/lambda-operator.md
@@ -39,7 +39,7 @@ An expression body definition has the following general syntax:
 member => expression;
 ```
 
-where *expression* is a valid expression. Note that *expression* can be a *statement expression* only if the member's return type is `void`, or if the member is a constructor, a finalizer, or a property `set` accessor.
+where *expression* is a valid expression. If the member's return type is `void` or if the member is a constructor, a finalizer, or a property `set` accessor, the *expression* must be a [*statement expression*](~/_csharplang/spec/statements.md#expression-statements).
 
 The following example shows an expression body definition for a `Person.ToString` method:
 
--- a/docs/csharp/language-reference/operators/lambda-operator.md
+++ b/docs/csharp/language-reference/operators/lambda-operator.md
@@ -39,7 +39,7 @@ An expression body definition has the following general syntax:
 member => expression;
 ```
 
-where *expression* is a valid expression. If the member's return type is `void` or if the member is a constructor, a finalizer, or a property `set` accessor, the *expression* must be a [*statement expression*](~/_csharplang/spec/statements.md#expression-statements).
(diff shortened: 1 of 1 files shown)


JustAnotherG:
I've being trying to understand this statement for a couple of hours now. One the one hand, it seems like what your saying is so obvious that one questions why you would say it. But then again on the other hand, I just may still not completely understand what is being said. Could you add at least 1 example that contrasts the right way and wrong way so that your statements are more concrete?

pkulikov:
@JustAnotherG 

I think more context would help. Typically, the type of the expression result must be the same (or implicitly convertible) as the member's return type. Seems obvious. Then, what about the type of the expression in case the member returns nothing, for example, `void` method? Obviously, the expression might be of any type, as its result would be discarded.

However, the following definition doesn't compile:

```csharp
public void Method() => 1 + 1;
```

The compiler realizes that the result of the expression would be discarded, so why even to allow performing an operation? It makes sense to perform an operation if it might have side effects. I think of statement expressions as the expressions that might have side effects. That's why only such expressions are allowed for nothing-returning members. The following version would compile:

```csharp
public void Method() => Do();
public int Do() => 1 + 1;
```

And yes, the expression result can be of any type then:

```csharp
public void Method() => Do();
public Func<int, Action<double>> Do() => i => s => Console.WriteLine(s);
```

I've updated the article text with more context about the type of the expression result.

BillWagner (approved):
This is a great clarification @pkulikov 

I'll :shipit: now.

PR #13766, opened 2019-08-08
Revised the namespace alias qualifier article

pkulikov:
Contributes to #10741


docs/csharp/language-reference/operators/namespace-alias-qualifier.md
```diff
@@ -1,45 +1,64 @@
...
+  {
+      public static forwpf::Point Convert(forwinforms::Point point) => new forwpf::Point(point.X, point.Y);
+  }
+  ```
 
-The `::` operator can also be used with a *using alias directive*:
+- An [extern alias](../keywords/extern-alias.md).
+- The `global` alias, which is the global namespace alias. The global namespace is the namespace that contains namespaces and types that are not declared inside a named namespace. When used with the `::` qualifier, the `global` alias always references the global namespace, even if there is the user-defined `global` namespace alias.
+  
+  The following example uses the `global` alias to access the .NET <xref:System> namespace, which is a member of the global namespace. Without the `global` alias, the user-defined `System` namespace, which is a member of the `MyCompany.MyProduct` namespace, would be accessed:
```
pkulikov: With that example in this article, I suggest to remove and redirect the following programming guide article:
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/namespaces/how-to-use-the-global-namespace-alias

BillWagner: I agree. Do you want to add it to this PR, or make a separate PR?

BillWagner (commented):
This LGTM @pkulikov 

Do you want to add the removal of the programming guide article to this PR, or make a separate PR? If you prefer a separate PR, this is ready to :shipit:

docs/csharp/language-reference/operators/namespace-alias-qualifier.md
pkulikov: @BillWagner I prefer to make it as a separate PR. I'll resolve the merge conflict with TOC and then it's ready to ship.

BillWagner (approved):
Thanks again @pkulikov 

Once the build finishes, I'll :shipit:

pkulikov:
@BillWagner thanks for review; I've resolved the merge conflict. Now, it should be ready.

PR #13803, opened 2019-08-10
Replace Windows SDK tokens

nxtn:


rpetrusha (approved):
Thank you, @NextTurn, for removing these two Windows SDK tokens and replacing them with their token text. I'll merge your PR now.

PR #13814, opened 2019-08-12
Format operator precedence as table

pkulikov:
Reformat sections into the table:
- more compact view (might even fit one screen)
- no more duplicative descriptions of operators


BillWagner (approved):
Thanks again @pkulikov 

This looks great (and it is close to one page).  I'll :shipit: now.

PR #13816, opened 2019-08-12
fix: MD001/heading-increment/header-increment

nschonni:
- Heading levels should only increment by one level at a time
- Trim traling/leading space
- Space around headers
- Add code fence langauges


rpetrusha (approved):
Your changes look good, @nschonni. I'll merge your PR now.

PR #13823, opened 2019-08-12
Pointing "What is ASP.NET?" to Core

CamSoper:


mairaw (changes requested):
So it should be What is ASP.NET Core, no?

CamSoper:
@mairaw Good call.

mairaw (approved):
Looks good now! Thanks @CamSoper.

PR #13839, opened 2019-08-13
WinForms improvements

gewarren:
- Heading improvements (e.g., "To add..." -> "Add..." and "Compiling" -> "Compile")
- Add "In Visual Studio," to beginning of some procedures
- Remove Visual Studio settings note
- Remove trailing whitespace (recommend hiding whitespace changes in diff)

### DIFF
--- a/docs/framework/winforms/advanced/effects-of-modifying-base-form-appearance.md
+++ b/docs/framework/winforms/advanced/effects-of-modifying-base-form-appearance.md
@@ -1,7 +1,7 @@
 ---
 title: "Effects of Modifying a Base Form's Appearance"
 ms.date: "03/30/2017"
-helpviewer_keywords: 
+helpviewer_keywords:
   - "parent forms [Windows Forms]"
   - "inherited forms [Windows Forms], modifications to base form"
   - "Windows Forms, base form appearance"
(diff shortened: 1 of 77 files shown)


docs/framework/winforms/advanced/effects-of-modifying-base-form-appearance.md
```diff
@@ -1,24 +1,22 @@
...
   - "base forms"
   - "inheritance [Windows Forms], forms"
 ms.assetid: 1c3f2b29-a05c-4c6f-aa1a-4e66b94f343a
 ---
-# Effects of Modifying a Base Form's Appearance
-During application development, you may often need to change the appearance of the base form from which other forms in the project (or in other projects) are inheriting.  
-  
- At design time, changes to the base form's appearance (be it the setting of properties or the addition and subtraction of controls) are reflected on inherited forms when the project containing the base form is built. It is not sufficient for you to simply save the changes to the base form. To build a project, choose **Build** from the **Build** menu.  
-  
-> [!NOTE]
```
Thraka: I'm curious, why are these notes removed?

gewarren: I removed a lot of these notes from the VS docs too. IMO they take up valuable real estate and appear prominently, but are actually not even relevant in most cases. The thing that different development settings affects the most is keyboard shortcuts, and I think that most of these topics don't even use keyboard shortcuts. I'm guessing these notes were originally added as a sort of future-proofing or cover-our-bases thing. If you feel that they are important, perhaps we put a note in the top-level topic of each section instead of each and every topic.

Thraka (changes requested):
Thanks for the cleanup 😄 Here are a few more fixes I could mostly do via suggestions. One fix you have to do back in the original file.

Thraka (changes requested):
Thanks for the cleanup 😄 Here are a few more fixes I could mostly do via suggestions. One fix you have to do back in the original file.

gewarren:
@Thraka I think I got all your suggested changes. And the ellipsis button issue is fixed too: https://review.docs.microsoft.com/en-us/dotnet/framework/winforms/controls/how-to-set-the-image-displayed-by-a-windows-forms-control-using-the-designer?branch=pr-en-us-13839.

#sign-off

Thraka (approved):

PR #13848, opened 2019-08-14
Fix header level so they show up in the navigation

jawn:
## Summary

Make header level consistent


Thraka (approved):
Looks good. Thanks!!

PR #13849, opened 2019-08-14
C110186: Change italic asterisk by italic underscoe

v-maudel:
Description: asterisks needs to be replaced by underscores, because MT engine breaks formatting on some languages.


v-maudel:
@rpetrusha , can you please review this change and merge if agree?

rpetrusha (approved):
The underscore was rendered in doc builds, @v-maudel. I've changed the text to use two underscores in place of two asterisks, and to keep the asterisk for italicization.

v-maudel:
Thanks @rpetrusha , I will report it to the localization team. We didn't though it will be rendered as underscores.

PR #13859, opened 2019-08-15
Replace OPC tokens

nxtn:


rpetrusha (approved):
Thanks, @NextTurn, for replacing the OPC token. I'll merge your PR now.

PR #13867, opened 2019-08-16
Fix the article name

pkulikov:
The article to rename:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-conversion-operators

According to the [spec](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/classes#conversion-operators), "conversion operator" is the operator that defines a user-defined conversion. While the article describes the cast operator `()` that performs/facilitates either a built-in or user-defined conversion.


BillWagner (approved):
Thanks for making this important fix @pkulikov 

Thanks again.

PR #13892, opened 2019-08-17
Replace IIS tokens

nxtn:
Support for IIS 5.0/6.0/7.0 has ended.


rpetrusha (approved):
Thanks, @NextTurn, for removing these four tokens. I'll merge your PR now.

NextTurn:
@rpetrusha Shall we remove references to those outdated versions?

rpetrusha:
I'm not sure, @NextTurn. whether they should be removed or not, and what it is that should be removed. I just don't know enough about IIS. The problem with removing version references is that it implies that the content applies to all versions, and the problem with removing entire blocks of versioned content is that it still may apply to current versions. If you have some sense of what should and shouldn't be changed, it's fine to submit a PR.

PR #13894, opened 2019-08-17
Specify camelCase usage for local variables

Youssef1313:
- Specified that camelCase is used for both parameters and local variables (it was mentioned for parameters only).
- Removed extra spaces.


rpetrusha:
Thank you for submitting this PR, @Youssef1313. However, this material is reprinted with the permission of Pearson Education, Inc., which retains its publishing rights. Because of that, we are not free to modify it. Consequently, I'll close this PR.

PR #13897, opened 2019-08-17

nemrism:
## Summary

I tried to get hands on some refactoring work again with this one.

Contributes to #10653


nemrism:
Maybe I'll have to separate PR by subfolders if I don't want to impact to many files at the same time.
Please let me know if you prefer that I break PR in even smaller ones!

nemrism:
This one actually concerns only the /csharp/misc folder at the moment

BillWagner (approved):
Thanks for making these updates @nemrism We appreciate it.

This was a large PR, but it was a search / replace, so it wasn't that hard to review.

Thanks for scoping it to one folder. I  do think this folder is one of the largest.

I've reviewed this, and checked the build, Everything is good, so I'll :shipit: now.

Thanks again for making this change.

PR #13900, opened 2019-08-18

nemrism:
Contributes to #10653

PR contains all files concerned in /csharp/language-reference/**


BillWagner (approved):
This one is also good to :shipit: @nemrism 

Thank you for doing all this work!

PR #13925, opened 2019-08-19
[framework] Remove redundant path to other sections

nemrism:
Remove redundant path in links to other sections in /framework/** files

Contributes to #10653


mairaw (approved):
Looks good @nemrism! Thanks!

Issue #13912, opened 2019-08-19
ByVal in VB samples


Youssef1313:
I think specifying `ByVal` for method arguments makes the code look noisy. It will be much cleaner to write:

```vb
Private Function FunctionName(arg1 As Type1, arg2 As Type2)
End function
```
instead of:

```vb
Private Function FunctionName(ByVal arg1 As Type1, ByVal arg2 As Type2)
End function
```

There are many code samples I've seen so far that has `ByVal` specified for method arguments.
Should you consider updating these samples ?

rpetrusha:
Since `ByVal` is the default, it doesn't have to be explicitly specified. I agree with you, @Youssef1313, that is noisy. It also has the unfortunate effect, in my experience, of causing the occasional `ByRef` keyword to be overlooked. So it would be useful to remove it.
I'll add this item to our backlog and the Community Contributor's project. If you're interested in removing ByVal from some or all of the examples, try to submit PRs that touch a small subset (approximately 100 or so) of files at a time.

paul1956:
There should be a CodeFix for removing ByVal and there may be third party tools that can do it for a whole project.

Youssef1313:
@paul1956 , I thought of that. but I'm not sure if unintended removal could be done in an automated solution. For example, the ByVal article itself has a code example, `ByVal` should NOT be removed there.
I'm not sure if there could be other places where `ByVal` shouldn't be removed.

paul1956:
A CodeFix only operates on "Code" and is smart enough to do the right thing, it can also be interactive.

Youssef1313:
@paul1956 , I'll try to do that soon when I get free time and see if I could get good results.

paul1956:
I would be happy to post/contribute a standalone remove ByVal analyzer and refactoring if you think its interesting. It is very small.

Youssef1313:
Well, you can start that anytime and submit a **small** PRs ( as suggested by @rpetrusha ) if you want and they will review it, @paul1956 .

paul1956:
I would like to start with correctly documenting line continuation in VB 16 I looked at the special character $ and has a line [!CODE-vb[interpolated-string-symbol](../../../../samples/snippets/visualbasic/language-reference/special-characters/dollar-sign1.vb)] which seems to do no where in the Doc GitHub repo. does [!Code-vb have some special meaning?

Youssef1313:
[!code-lang....] is used in .md files to include a code sample.
[...]

PR #13935, opened 2019-08-20
Added NOTE in ByVal

Youssef1313:
Related to https://github.com/dotnet/docs/issues/13912

Not sure if the NOTE can be re-phrased in a better way than mine. Can you check ?, @rpetrusha .

### DIFF
--- a/docs/visual-basic/language-reference/modifiers/byval.md
+++ b/docs/visual-basic/language-reference/modifiers/byval.md
@@ -11,6 +11,9 @@ ms.assetid: 1eaf4e58-b305-4785-9e3d-e416b9c75598
 ---
 # ByVal (Visual Basic)
 Specifies that an argument is passed [by value](../../programming-guide/language-features/procedures/passing-arguments-by-value-and-by-reference.md), so that the called procedure or property cannot change the value of a variable underlying the argument in the calling code. If no modifier is specified, ByVal is the default.
+
+> [!NOTE]
+> We don't recommend to explicitly specify `ByVal` for argument as it's the default. Explicitly specifying it makes the code looks noisy.
   
 ## Remarks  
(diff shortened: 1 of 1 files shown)


docs/visual-basic/language-reference/modifiers/byval.md
```diff
@@ -11,25 +11,28 @@ ms.assetid: 1eaf4e58-b305-4785-9e3d-e416b9c75598
 ---
 # ByVal (Visual Basic)
 Specifies that an argument is passed [by value](../../programming-guide/language-features/procedures/passing-arguments-by-value-and-by-reference.md), so that the called procedure or property cannot change the value of a variable underlying the argument in the calling code. If no modifier is specified, ByVal is the default.
+
+> [!NOTE]
+> We don't recommend to explicitly specify `ByVal` for argument as it's the default. Explicitly specifying it makes the code looks noisy.
```
rpetrusha:
```suggestion
> Because it is the default, you do not have to explicitly specify the `ByVal` keyword in method signatures. It tends to produce noisy code and often leads to the non-default `ByRef` keyword being overlooked.
```

rpetrusha (changes requested):
Thanks for adding the note, @Youssef1313. I thought that, given the overall context, saying that "we don't recommend" is much too strong. So I've suggested an alternate wording.

docs/visual-basic/language-reference/modifiers/byval.md
Youssef1313: Thanks! Looks much better now :)

rpetrusha (approved):

rpetrusha:
Thanks for making the additional change, @Youssef1313. I'll merge your PR now.

PR #13992, opened 2019-08-22
Simplified the note

pkulikov:
The paragraph before the note mentions already *four* types. The note talks about the fifth type, then. However, I think that arithmetic can be removed from the note at all.


rpetrusha (approved):
I agree that it's best to discard arithmetic, @pkulikov, particularly when that arithmetic is incorrect! Thanks for making this change. I'll merge your PR now.

PR #14050, opened 2019-08-27
Replace Windows Vista tokens

nxtn:


rpetrusha (changes requested):
Thanks, @NextTurn, for replacing these Windows Vista tokens. I've suggested a number of revisions for cases in which mention of Windows Vista is inappropriate. When I've been uncertain, I've left the mention of Windows Vista in place.

rpetrusha (approved):

PR #14055, opened 2019-08-27
remove extra "tuples" subdirectory

BillWagner:
Companion to dotnet/samples#1349

When I updated the tuples sample, I removed the extra subdirectory created by the framework templates.
This PR updates the code includes.


rpetrusha (approved):
From looking at the failed code references in the build report, @BillWagner, I think that the build should succeed without issue once the code changes are accessible to the build. I'll approve this PR so that you can merge when you're ready.

PR #14058, opened 2019-08-27
Mention concept of "empty tuple" in `unit` doc

mcgwiz:
## Summary

As simple as it seems, coming from a C#/JS background, the `unit` type was one of the most mysterious and challenging parts of learning F# for me. And because it has such fundamental role in the language, my learning was impeded and lacked confidence.

However, when I stumbled upon an explanation that noted that the value of the `unit` type is also thought of as the empty tuple, it all immediately clicked. I think a casual mention of this can trigger this same connection for many readers coming from a similar background, who have arrived at this page seeking a comprehensive description of the `unit` concept.

### DIFF
--- a/docs/fsharp/language-reference/unit-type.md
+++ b/docs/fsharp/language-reference/unit-type.md
@@ -18,7 +18,7 @@ The `unit` type is a type that indicates the absence of a specific value; the `u
 
 Every F# expression must evaluate to a value. For expressions that do not generate a value that is of interest, the value of type `unit` is used. The `unit` type resembles the `void` type in languages such as C# and C++.
 
-The `unit` type has a single value, and that value is indicated by the token `()`.
+The `unit` type has a single value, and that value is indicated by the token `()`. As its literal appearance suggests, `unit` can essentially be thought of as the empty tuple.
 
 The value of the `unit` type is often used in F# programming to hold the place where a value is required by the language syntax, but when no value is needed or desired. An example might be the return value of a `printf` function. Because the important actions of the `printf` operation occur in the function, the function does not have to return an actual value. Therefore, the return value is of type `unit`.
 


BillWagner:
@cartermp Can you take a look at this one?

cartermp:
I'm not sure if I agree with this change. F# `unit` and tuples are fundamentally different things, both in terms of how the language treats them and how they are emitted as .NET IL. The section directly below explains the concept more:

> The value of the `unit` type is often used in F# programming to hold the place where a value is required by the language syntax, but when no value is needed or desired. [...]

Tuples in F# are ad-hoc, small groupings of data used to aggregate data when you need it. The concept of an empty tuple or one-element tuple doesn't exist in the language, so I don't think we'd use that concept to draw an analogue to bottom types like `unit` here.

mcgwiz:
Hi, thanks for the review!

Is the concern that a high-level comparison which isn't technically/completely accurate at a low level is harmful? I've found conceptual, historical, or related information quite appropriate and helpful when properly presented in reference documentation, and that's the category in which I'd put the empty- / 0-tuple comparison. If this is the concern, would it help to soften the language or add a caveat, e.g.

> As its appearance suggests, `unit` behaves in some ways like the 0-tuple (which exists in some functional languages but not in F#).

Or is the concern that the comparison is simply inaccurate even at a conceptual level? The Wikipedia article on `unit` indicates that several languages do not have a bespoke `unit` type and instead use the 0-tuple, written `()`. When I originally saw the mysterious sequence of characters "()" in the context of a 0-tuple, F#'s `unit` type suddenly and deeply made sense to me. `unit`'s outward behavior is like a tuple (which are easy to understand): it's a value type; it only has a single value since it's "tuple size" could be thought of as 0, its conceptual meaning of "no interesting value" is then readily apparent, syntax for defining functions with no parameters makes sense (`() = /* body */`) , and likewise syntax for invoking functions with no parameters makes sense (`func ()`).

I might also add that the line preceding my proposed addition states that `unit` resembles `void` in C#/C++. I've read this in several places and always found it misleading and confusing. For one, `void` cannot be instantiated in C#; it effectively has no value. (Not actually proposing this, but it's almost more sensible to say `unit` is like a special form of `null` that can't be assigned to reference bindings; you pass `null` in for `unit` when calling F# from C#, after all.) As far as the CLR is concerned, this all becomes a little muddled to the typical developer, but I doubt the typical developer is overly concerned about the CLR behavior. If I'm mistaken, we might add an explicit explanation of that behavior to this article.

I'm happy to take suggestions for improvement because overall, the conceptual connection between `unit` and a 0-tuple was critical link in my understanding of it, especially coming from a C#/OOP background.

cartermp:
Yes, the chief concern is that this comparison is not accurate or reflective of how the F# compiler treats `unit` or how it's emitted as IL. There's also no analogue in C#, because tuples must have at least two elements - creating a 1-element or 0-element tuple is a compile error.

The `unit` type in F# is actually a reference type, see its definition [here](https://github.com/dotnet/fsharp/blob/c18e1780b3f3f345364cb1ad8e510ea9f4590d3a/src/fsharp/FSharp.Core/prim-types.fs#L23-L30). Furthermore, tuples in F# are reference types by default, though you can construct value type tuples via `struct (x, y)` syntax.

While I appreciate that this particular comparison could have been helpful for you if it were documented, `unit` is not comparable to the concept of an empty tuple in a way that could be rationalized in a .NET environment.

I think there's definitely some improvement to make here by removing the preceding sentence:

> type resembles the `void` type in languages such as C# and C++.

This is also not quite accurate, since as you point out, `void` is not a "real" type in C#. So perhaps it's worth expounding upon how there is a special need to represent the "empty value", or "no value" in a language where everything needs a value. And that this special value is called unit, and has the `()` notation in large part because this is a similar notation to other languages where declaring and invoking functions/methods with no parameters uses `()`.

mcgwiz:
Thanks for your clarifications; I should have been more clear. `()`
technically represents references, but like records, they behave with value
semantics w.r.t. equality. E.g. `printf "%b" (() = ())` prints "true". But
I concede that using 0-tuple as a conceptual foothold is just a personal
anecdote that has no substantiated benefit for a general audience, so I'll
close.

I agree that any explanation of the concepts behind the notation would be
helpful.

Additionally, also at a conceptual level and to benefit newcomers, it would
be helpful to be as explicit as possible about `unit` existing to support
side-effects.

> The value of the `unit` type is often used in F# programming to hold the
place where a value is required by the language syntax but when no value is
needed or desired, such as with expressions that produce side effects.

The discussions about the `printf` example discussion and about `do`
bindings should include the term "side-effect" as well. However, if
newcomers aren't the primary audience, then I could see this being too
prescriptive / conceptual.

mcgwiz:
Thanks for your clarifications; I should have been more clear. `()` technically represents references, but like records, they behave with value semantics w.r.t. equality. E.g. `printf "%b" (() = ())` prints "true". But I concede that using 0-tuple as a conceptual foothold is just a personal anecdote that has no substantiated benefit for a general audience, so I'll close.

I agree that any explanation of the concepts behind the notation would be helpful.

Additionally, also at a conceptual level and to benefit newcomers, it would be helpful to be as explicit as possible about `unit` existing to support side-effects.

> The value of the `unit` type is often used in F# programming to hold the place where a value is required by the language syntax but when no value is needed or desired, such as with expressions that produce side effects.

The discussions about the `printf` example discussion and about `do` bindings should include the term "side-effect" as well. However, if newcomers aren't the primary audience, then I could see this being too prescriptive / conceptual.

PR #14067, opened 2019-08-28
Replace .NET tokens

nxtn:


mairaw (changes requested):
Thank you @NextTurn. I don't think we need to qualify .NET with Microsoft, so made some suggestions.

mairaw (approved):
Thanks again!

PR #14068, opened 2019-08-28
Replace .NET Framework attribute tokens

nxtn:


mairaw (changes requested):
Thank you @NextTurn. I don't think that term is really needed, so I added some suggestions.

mairaw (approved):
Thanks for making the changes!

PR #14088, opened 2019-08-29
Restore config file hierarchy and links

nemrism:
Used the formatting used in root folder and applied it to the `/framework/configure-apps/file-schema/compiler/` one

Contributes to #2038


mairaw (approved):
Thank you so much for jumping into these issues @nemrism! Looks great!

PR #14092, opened 2019-08-29
Updating using specificity

RookTKO:
Added specificity to using syntax.

## Summary

Added proper technical naming conventions for others to be able to reference directives and namespaces.


rpetrusha (changes requested):
Thank you, @RookTKO, for contributing to the dotnet/docs repo. I've suggested some changes, since the `using` keyword in this context is documented as a statement and not a directive; see [Using (C# Reference)](https://docs.microsoft.com/dotnet/csharp/language-reference/keywords/using). The `using` statement is also unrelated to the `using` directive, which imports a namespace or the members of a type.

rpetrusha (approved):

PR #14107, opened 2019-08-30
Replace XML namespace tokens

nxtn:


rpetrusha (approved):
Thank you, @NextTurn, for removing another token and replacing it with its text. I'll merge your PR now.

PR #14112, opened 2019-08-30
US 1583733 Add missing language IDs - 03

v-thpra:

- Used "console" for command syntax definitions.
- Used "output" for examples of command output. (This is not a language - obviously - but it is an ID that has been used before.

Contributes to #2192


mairaw (approved):
Looks great. Thanks @v-thpra!

PR #14118, opened 2019-08-30
US 1583733 Add missing language IDs - 05

v-thpra:

- Used the tag "output" for code blocks with program output examples. 
- Used the tag "text" for code blocks with text content.

Contributes to #2192


mairaw (approved):
Looks great @v-thpra! Thanks.

PR #14120, opened 2019-08-30
Changed "Exposing .NET Framework Components to COM" title

rpetrusha:
## Changed "Exposing .NET Framework Components to COM" title

The text had been revised, but not the title.


mairaw (approved):
LGTM

PR #14154, opened 2019-09-02
US 1583733 Add missing language IDs - 06

v-thpra:
- Used "output" for program output examples: Some of these could be XML but they include some plain text output also, so they are not strictly XML.

Contributes to #2192


docs/csharp/language-reference/tokens/interpolated.md
```diff
@@ -32,7 +32,7 @@ The structure of an item with an interpolation expression is as follows:
 {<interpolationExpression>[,<alignment>][:<formatString>]}
 ```
 
-Elements in square brackets are optional. The following table describes each element:
+Elements in square brackets are optional. The following table describes each element:
```
pkulikov: I guess that was a typo to add unnecessary white spaces:
```suggestion
Elements in square brackets are optional. The following table describes each element:
```

v-thpra:
@mairaw Some of these could be XML but they include some plain text output also, so they are not strictly XML. If you want them changed just let me know. Thanks!

docs/csharp/language-reference/tokens/interpolated.md
v-thpra: Good catch Petr. I deleted the extra spaces.

BillWagner (approved):
Thanks for making these updates @v-thpra 

I'll :shipit: now.

PR #14159, opened 2019-09-02
Update template.md

Youssef1313:
Updated ms.date value to be `MM/dd/yyyy` instead of `mm/dd/yyyy`.
Usually, `mm` refers to the minutes and `MM` refers to the month. It's surely obvious that the month is meant here, but to be accurate.

Also updated the `ms.date` explanation to be `MM/dd/yyyy` instead of ` MM/DD/YYYY`.


mairaw (approved):
Thank you @Youssef1313. Your changes look good!

PR #14198, opened 2019-09-04
C# 8.0: unmanaged constructed types

### LINKED ISSUE #14101: C# 8.0 - unmanaged constructed types
dotnet/csharplang#1937

@BillWagner this is for the [C# 8 docs project](https://github.com/dotnet/docs/projects/58).
Could you please assign this issue to me?

---
#### Document Details

* ID: 93e5d75f-07a3-e7ee-2842-ac754bdd2e41
* Version Independent ID: 4a3622fb-7d09-78ec-ba38-b88001475018
* Content: [Unmanaged types - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/unmanaged-types)
* Content Source: [docs/csharp/language-reference/builtin-types/unmanaged-types.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/builtin-types/unmanaged-types.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

pkulikov:
Fixes #14101
Supported by dotnet/samples#1411

### DIFF
--- a/docs/csharp/language-reference/builtin-types/unmanaged-types.md
+++ b/docs/csharp/language-reference/builtin-types/unmanaged-types.md
@@ -19,7 +19,7 @@ Beginning with C# 8.0, a *constructed* struct type that contains fields of unman
 
 [!code-csharp[unmanaged constructed types](~/samples/csharp/language-reference/builtin-types/UnmanagedTypes.cs#ProgramExample)]
 
-A generic struct may be the source of both unmanaged and not unmanaged constructed types. The preceding example defines a generic struct `Coords<T>` and presents the examples of unmanaged constructed types. The example of not an unmanaged type is `Coords<object>`. It's not unmanaged because it has the fields of the `object` type, which is not unmanaged. If you want *any* constructed type to be an unmanaged type, use the `unmanaged` constraint in the definition of a generic struct:
+A generic struct may be the source of both unmanaged and not unmanaged constructed types. The preceding example defines a generic struct `Coords<T>` and presents the examples of unmanaged constructed types. The example of not an unmanaged type is `Coords<object>`. It's not unmanaged because it has the fields of the `object` type, which is not unmanaged. If you want *all* constructed types to be unmanaged types, use the `unmanaged` constraint in the definition of a generic struct:
 
 [!code-csharp[unmanaged constraint in type definition](~/samples/csharp/language-reference/builtin-types/UnmanagedTypes.cs#AlwaysUnmanaged)]
 
--- a/docs/csharp/language-reference/builtin-types/unmanaged-types.md
+++ b/docs/csharp/language-reference/builtin-types/unmanaged-types.md
@@ -1,19 +1,23 @@
 ---
 title: "Unmanaged types - C# reference"
-ms.date: 07/23/2019
+ms.date: 09/04/2019
 helpviewer_keywords: 
   - "unmanaged type [C#]"
 ---
 # Unmanaged types (C# reference)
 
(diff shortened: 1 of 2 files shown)


BillWagner:
This is a great start @pkulikov There's a couple scenarios that we should cover, but it may require a slightly different sample.  For discussion, I'll use the `Coords<T>` type you've defined.  

We should explain that if the type argument is not an unmanaged type, the constructed type is not unmanaged:

```csharp
var objCoords = new Coords<object>();
```

That's a contrived example, but you see the point.

It is also interesting that adding the `unmanaged` constraint would mean that *any* constructed type would be an unmanaged type:

```csharp
public struct Coords<T> where T : unmanaged
~
    public T X;
    public T Y;
}
```

Thoughts on how to update to make those two points?

pkulikov:
@BillWagner 

> We should explain that if the type argument is not an unmanaged type...

I understand what you mean, but wording should be different. Type argument *can* be a reference type. Not practical example:
```csharp
public struct Coords<T>
{
    public int X;
    public int Y;
}
```

With that definition, `Coords<T>` is unmanaged for *any* type argument. All fields are of an unmanaged type, regardless of the type argument.

I'll think how to update the article.

pkulikov:
@BillWagner please take a look at the latest commit; I've also added one more snippet to the sample PR.

pkulikov:
ok, I think the current edition is final. What do you think @BillWagner?

docs/csharp/language-reference/builtin-types/unmanaged-types.md
```diff
@@ -1,20 +1,28 @@
...
+- Any user-defined [struct](../keywords/struct.md) type that contains fields of unmanaged types only and, in C# 7.3 and earlier, is not a constructed type (a type that includes at least one type argument)
 
 Beginning with C# 7.3, you can use the [`unmanaged` constraint](../../programming-guide/generics/constraints-on-type-parameters.md#unmanaged-constraint) to specify that a type parameter is a non-pointer unmanaged type.
 
+Beginning with C# 8.0, a *constructed* struct type that contains fields of unmanaged types only is also unmanaged, as the following example shows:
+
+[!code-csharp[unmanaged constructed types](~/samples/csharp/language-reference/builtin-types/UnmanagedTypes.cs#ProgramExample)]
+
+A generic struct may be the source of both unmanaged and not unmanaged constructed types. The preceding example defines a generic struct `Coords<T>` and presents the examples of unmanaged constructed types. The example of not an unmanaged type is `Coords<object>`. It's not unmanaged because it has the fields of the `object` type, which is not unmanaged. If you want *any* constructed type to be an unmanaged type, use the `unmanaged` constraint in the definition of a generic struct:
```
BillWagner: nit:

```suggestion
A generic struct may be the source of both unmanaged and not unmanaged constructed types. The preceding example defines a generic struct `Coords<T>` and presents the examples of unmanaged constructed types. The example of not an unmanaged type is `Coords<object>`. It's not unmanaged because it has the fields of the `object` type, which is not unmanaged. If you want *all* constructed types to be unmanaged types, use the `unmanaged` constraint in the definition of a generic struct:
```

BillWagner (changes requested):
This is great work again @pkulikov 

I had one small grammar nit to consider, and then I'll :shipit:

docs/csharp/language-reference/builtin-types/unmanaged-types.md
pkulikov: Thank you @BillWagner

BillWagner (approved):

PR #14258, opened 2019-09-06
Update cs0219.md

### LINKED ISSUE #14231: Please explain when this warning gets issued more clearly
From reading this documentation, I would expect the following code to trigger this warning, but it doesn’t:

```csharp
public class C {
    public void M(IThing thing) {
        var needUpdateVariable = thing.DoSomething();
    }
    public interface IThing {
        object DoSomething();
    }
}
```

Please add documentation explain what is different about that code and why it doesn’t throw this warning.

Thanks.

---
#### Document Details

* ID: 34fc8443-8e5e-2e87-6ee5-fbab61709e1e
* Version Independent ID: 41a54122-d4f3-c233-7e67-6ee521af4338
* Content: [Compiler Warning (level 3) CS0219](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0219)
* Content Source: [docs/csharp/misc/cs0219.md](https://github.com/dotnet/docs/blob/master/docs/csharp/misc/cs0219.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

Youssef1313:
Fixes #14231

### DIFF
--- a/docs/csharp/misc/cs0219.md
+++ b/docs/csharp/misc/cs0219.md
@@ -24,11 +24,12 @@ public class MyClass
 {
    public static void Main()
    {
+      var interpolated = "Interpolated";
       var a = 0; // CS0219
       int b = GetZero(); // Doesn't generate a warning.
       var c = "Regular string"; // CS0219
       var d = $"Constant interpolated string"; // Doesn't generate a warning.
-      var e = $"Interpolated string, a = {a}"; // Doesn't generate a warning.
+      var e = $"{interpolated} string"; // Doesn't generate a warning.
    }
    
    private static int GetZero()
--- a/docs/csharp/misc/cs0219.md
+++ b/docs/csharp/misc/cs0219.md
@@ -27,7 +27,8 @@ public class MyClass
       var a = 0; // CS0219
       int b = GetZero(); // Doesn't generate a warning.
       var c = "Regular string"; // CS0219
-      var d = $"Interpolated string"; // Doesn't generate a warning.
+      var d = $"Constant interpolated string"; // Doesn't generate a warning.
+      var e = $"Interpolated string, a = {a}"; // Doesn't generate a warning.
    }
    
    private static int GetZero()
--- a/docs/csharp/misc/cs0219.md
+++ b/docs/csharp/misc/cs0219.md
@@ -24,10 +24,10 @@ public class MyClass
 {
    public static void Main()
    {
-      int a = 0; // CS0219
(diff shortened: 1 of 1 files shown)


binki (changes requested):
Should probably change `string c = "Interpolated string";` to `string c = "Regular string";`.

Also, maybe be consistent and use `var` everywhere?

Youssef1313:
@binki , That's correct. I'll update it now.

docs/csharp/misc/cs0219.md
```diff
@@ -1,27 +1,40 @@
...
-   public static void Main()  
-   {  
-      int a = 0;   // CS0219  
-   }  
+The variable 'variable' is assigned but its value is never used
+
+ The compiler issues a level-three warning, when you declare and assign a variable, but do not use it.
+
+ > [!NOTE]
+ > The compiler generates this warning only when the variable value is a compile-time constant. Because interpolated strings are transformed into a <xref:System.String.Format%2A?displayProperty=nameWithType> method call, it's not considered a compile-time constant.
```
BillWagner: I think a better note would reference the justification in dotnet/roslyn#15695, namely that assigning the result of some expression to an (otherwise) unused local variable keeps that in memory and enables viewing with a debugger.

```suggestion
 > The compiler generates this warning only when the variable value is a compile-time constant. Assigning an expression or method result to a local variable makes it easier to observe those expressions in the debugger. It also makes the result reachable, preventing garbage collection while that variable is reachable.
```

BillWagner (changes requested):
Thanks for making these changes @Youssef1313 

While what you wrote is correct, a better explanation would explain why the compiler behaves this way. I've provided a suggestion.

docs/csharp/misc/cs0219.md
Youssef1313: I just noticed that `Assigning an expression or method result` seems to be in-accurate I think.
because regular string literal or number like `"string"` or `0` in:

```csharp
var x = "string";
var y = 0;
```

can be considered as expressions. Actually, there are the simplest expressions. But... assigning them to an unused variable generates the warning, while assigning a method result doesn't.

BillWagner: Good point. I updated the suggestion.

Youssef1313: @BillWagner , The updated suggestion looks good. I've applied it.

BillWagner (approved):

PR #14283, opened 2019-09-07
Replace Entity Framework tokens

Youssef1313:


rpetrusha (approved):
Thanks, @Youssef1313, for removing the Entity Framework token and replacing it with its text. I'll merge your PR now.

PR #14287, opened 2019-09-07
Fix snippet language

Youssef1313:
This code snippet isn't highlighted at [live docs](https://docs.microsoft.com/en-us/dotnet/core/dependency-loading/understanding-assemblyloadcontext#debugging-type-conversion-issues). Changing identifier from `C#` to `csharp` should fix it.


rpetrusha (approved):
Thank you for fixing the language identifier, @Youssef1313. I'll merge your PR now.

PR #14291, opened 2019-09-08
Add missing cell in table

Youssef1313:
Found the label1 text in the [designer file](https://github.com/dotnet/samples/blob/master/snippets/visualbasic/VS_Snippets_CLR/CryptoWalkThru/vb/Form1.Designer.vb#L132)


rpetrusha (changes requested):
Changing the cell value to "Key not set" implies that that's the value of the label's text property. It's best to leave it blank, as it was, or change it to something that differs in appearance from the text properties that should be set, such as **\<not set>**.

Youssef1313:
@rpetrusha, The value is **indeed** the literal string `"Key not set"`.

image

rpetrusha:
Sorry, @Youssef1313. I'd completely overlooked line 132. I'll approve and merge your PR now. Thanks for making this correction.

rpetrusha (approved):

PR #14295, opened 2019-09-08
fix: Replace bold Note with alert style [!NOTE]

nschonni:
## Summary

Describe your changes here.


docs/csharp/misc/cs0414.md
```diff
@@ -1,43 +1,45 @@
...
+{
+   private int i = 1;  // CS0414
+
+   public static void Main()
+   { }
+}
+```
+
+> [!NOTE]
+> If the variable `i` is declared as `protected or public`, no error will be generated because the compiler cannot know whether a derived class might use it or some other client code might instantiate the class and reference the variable
```
pkulikov:
```suggestion
> If the variable `i` is declared as `protected` or `public`, no error will be generated because the compiler cannot know whether a derived class might use it or some other client code might instantiate the class and reference the variable
```

BillWagner (approved):
These all LGTM @nschonni 

Thanks for helping us improve the quality of our markdown. I'm happy to :shipit: now.

Thanks again!

PR #14296, opened 2019-09-09
C# 8.0: null-coalescing assignment

### LINKED ISSUE #13686: Update page
It seems the page needs an update, as the "null coalescing assignment" operator is already available in C# 8 and not just proposed anymore.

---
#### Document Details

* ID: d273731d-0f9a-1870-1275-bde4f4b3f69d
* Version Independent ID: 7d8b872e-6d86-77c1-2fad-afa110ab8f4c
* Content: [null coalescing assignment - C# 8.0 language proposals](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/null-coalescing-assignment#feedback)
* Content Source: [proposals/csharp-8.0/null-coalescing-assignment.md](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/null-coalescing-assignment.md)
* Product: **dotnet-csharp**
* Technology: **csharp-spec**
* GitHub Login: @dotnet-bot
* Microsoft Alias: **wiwagn**

pkulikov:
I've assumed the following:

- Like the `??` operator, `??=` is not overloadable (also mentioned in the [feature proposal](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/null-coalescing-assignment))
- Like assignment operators, `??=` has the lowest precedence. 

Fixes #13686
Supported by dotnet/samples#1445


docs/csharp/language-reference/operators/null-coalescing-operator.md
```diff
@@ -1,15 +1,15 @@
...
-title: "?? operator - C# reference"
+title: "?? and ??= operators - C# reference"
 ms.custom: seodec18
-ms.date: 06/07/2019
+ms.date: 09/09/2019
 f1_keywords: 
   - "??_CSharpKeyword"
 helpviewer_keywords: 
   - "null-coalescing operator [C#]"
   - "?? operator [C#]"
```
Youssef1313: Isn't it better to include `??=` in the `f1_keywords` and `helpviewer_keywords` metadata ?

docs/csharp/language-reference/operators/null-coalescing-operator.md
```diff
@@ -1,15 +1,15 @@
...
 ms.assetid: 088b1f0d-c1af-4fe1-b4b8-196fd5ea9132
 ---
-# ?? operator (C# reference)
+# ?? and ??= operators (C# reference)
 
 The null-coalescing operator `??` returns the value of its left-hand operand if it isn't `null`; otherwise, it evaluates the right-hand operand and returns its result. The `??` operator doesn't evaluate its right-hand operand if the left-hand operand evaluates to non-null.
```
Youssef1313: This paragraph is under `# ?? and ??= operators (C# reference)` heading, but it also talks about `??` but not `??=`

There is a `## Null-coalescing assignment` that talks about `??=`, so I suggest adding `## Null-coalescing operator` just under `# ?? and ??= operators (C# reference)`.

docs/csharp/language-reference/operators/null-coalescing-operator.md
pkulikov: That's a good catch, thanks!

docs/csharp/language-reference/operators/null-coalescing-operator.md
BillWagner: For the reference, I prefer the single H2, as this article has now.

I agree that in the standard, we would have separate headings for each sub clause.

docs/csharp/language-reference/operators/null-coalescing-operator.md
```diff
@@ -25,7 +28,25 @@ is evaluated as
...
+
+[!code-csharp[unconstrained type parameter](~/samples/csharp/language-reference/operators/NullCoalescingOperator.cs#UnconstrainedType)]
+
+## Null-coalescing assignment
+
+Available in C# 8.0 and later, the null-coalescing assignment operator `??=` assigns the value of its right-hand operand to its left-hand operand only if the left-hand operand is null. The `??=` operator doesn't evaluate its right-hand operand if the left-hand operand evaluates to non-null.
+
+[!code-csharp[null-coalescing assignment](~/samples/csharp/language-reference/operators/NullCoalescingOperator.cs#Assignment)]
+
+Unlike with [compound assignment operators](assignment-operator.md#compound-assignment), the type of the result of the `??=` operator may differ from the type of its left-hand operand. For expression `a ??= b`, if `a` is of a nullable value type `T?` and `b` is implicitly convertible to the corresponding underlying type `T`, the type of the result of `a ??= b` is `T`. The null-coalescing operator `??` also behaves that way.
```
BillWagner: This wording troubles me. But, I haven't come up with a better way to explain it. I'll discussmy concerns, so we can discuss together:

The result type *is* the same as the left hand side argument. However, it is guaranteed to be non-null (other than the pathological case of `a ??= null;`).  So, introducing the concept of *type* might be confusing. For example:

```csharp
int? a = null;
int b = 5;
a ??= b; // a is still an int?, but it's HasValue property is true.
```

By saying "the result of the `??=` operator may differ..." that could lead readers to assume `a` is an `int`, not an `int?`.

BillWagner (changes requested):
I really like what you've written here @pkulikov 

I had one comment for discussion. As worded, I'm afraid it could mislead readers. I don't have a better solution yet. Let's see what we come up with.

docs/csharp/language-reference/operators/null-coalescing-operator.md
pkulikov: @BillWagner 

> The result type _is_ the same as the left hand side argument.

What do you mean by the "result type"? If the type of the result returned by the `??=` operator, then it isn't always the same as the left-hand argument. The following code shows that:

```csharp
var xs = new List<int>();
int? a = 5;
xs.Add(a ??= 10);  \\ that compiles because ??= returns an `int` value
xs.Add(a);  \\ that doesn't compile because a is `int?`
```

That's why I mentioned, it's "unlike with compound assignment operators".

> By saying "the result of the `??=` operator may differ..." that could lead readers to assume `a` is an `int`, not an `int?`.

I agree with that possible confusion. Is that the "dangerous" confusion though? That would mean that a variable *has changed* its type as a result of the operation.

Though, what about the following wording:

> If the left-hand operand is of the nullable value type `T?`, expression `a ??= b` can return the value of the corresponding underlying type `T`. That happens when the type of `b` is implicitly convertible to type `T`.

pkulikov: @BillWagner another option is to remove that paragraph at all. Technically interested readers would follow the [feature proposal link](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/null-coalescing-assignment) where all the details are mentioned.

I think that would simplify the article.

pkulikov:
@BillWagner I've removed the confusing paragraph. That allowed me to remove one of the H2 headings and restructure the article as a lot of properties are shared by both `??` and `??=`. What do you think of the latest edition? If that looks go to you, then I'll update the samples PR and relative links in this PR.

BillWagner:
@pkulikov I really like the direction you've taken. Removing that paragraph is the right choice. The explanations now help readers learn how to use the feature, and those that want deep details can find the spec proposals.

When you're ready, I'll give this one and the samples PR a final review and hopefully merge them.

pkulikov:
@BillWagner thanks; this and the samples PR are ready for your final review.

docs/csharp/language-reference/operators/null-coalescing-operator.md
pkulikov: @Youssef1313 thanks for your remark. It contributed to the final version of the article.

docs/csharp/language-reference/operators/null-coalescing-operator.md
```diff
@@ -1,31 +1,48 @@
...
-is evaluated as
+are evaluated as
 
 ```csharp
 a ?? (b ?? c)
+d ??= (e ??= f)
 ```
 
-The `??` operator can be useful in the following scenarios:
+## Use cases
```
BillWagner: nit (for consistency with other articles)

```suggestion
## Examples
```

BillWagner (changes requested):
Overall, this is great. I like the examples, and how it leads readers to practices they should change.

I had one suggestion on an H2, and then it's ready to merge.

Thanks again @pkulikov

docs/csharp/language-reference/operators/null-coalescing-operator.md
pkulikov: Applied, thanks @BillWagner

BillWagner (approved):

PR #14303, opened 2019-09-09
gRPC e-book Chapter 2

ThatRendle:
General overview of gRPC and comparison to key WCF features and models


docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
```diff
@@ -1,11 +1,26 @@
...
 title: Interface Definition Language - gRPC for WCF Developers
-description: TO BE WRITTEN
+description: Introducing the Protocol Buffers IDL
 author: markrendle
 ms.date: 09/02/2019
 ---
 
 # Interface Definition Language
 
+With WCF, services can expose description metadata using the Web Service Definition Language (WSDL), which is generated dynamically using .NET Reflection at runtime. Developers can use this metadata to generate clients for those services, potentially in other languages if a platform-neutral binding such as SOAP over HTTP is used.
```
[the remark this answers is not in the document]
markrendle: Added note and link on server reflection at end of page.

docs/architecture/grpc-for-wcf-developers/grpc-overview.md
```diff
@@ -1,11 +1,27 @@
...
 
 # gRPC overview
 
+Chapter 1 looked at the genesis of both WCF and gRPC. This chapter will consider some of the key features of gRPC and how they compare to WCF.
+
+## Key principles
+
+As discussed in Chapter one, Google wanted to utilize the introduction of HTTP/2 to rework Stubby, their internal, general purpose RPC infrastructure. Stubby, renamed gRPC, would now be able to take advantage of standardization and would extend its applicability to mobile technology, Cloud and the Internet of Things.
+
+In order to achieve this, Google established a set of principles which would govern gRPC. The most relevant ones are covered below and are mainly concerned with maximizing access and usability:
```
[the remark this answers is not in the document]
markrendle: Changed.

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
```diff
@@ -1,11 +1,35 @@
...
 title: WCF bindings and transports - gRPC for WCF Developers
-description: TO BE WRITTEN
+description: How the various WCF bindings and transports compare to gRPC
 author: markrendle
 ms.date: 09/02/2019
 ---
 
 # WCF bindings and transports
 
+WCF has lots of different built in *bindings* specifying different network protocols, wire formats and other implementation details. gRPC effectively has just one network protocol and one wire format. However, you are still likely to discover that gRPC offers the best solution in most cases. What follows is a short discussion about the most relevant WCF Bindings and how they compare to their equivalent in gRPC.
```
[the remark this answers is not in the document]
markrendle: I thought about mentioning the alternative wire formats but I think it would overcomplicate the guide in the context of its intended purpose and readership. It will be easy to add a new section on custom transports when 5.0 launches.

markrendle:

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
```diff
@@ -1,11 +1,29 @@
...
+
+One of the advantages of the Protobuf IDL is that as a custom language it enables gRPC to be completely language and platform agnostic, not favoring any technology over another.
+
+The Protobuf IDL is also much easier for humans to both read and write than WSDL. Changing the WSDL of a WCF service typically requires making the changes to the service code itself, running the service and regenerating the WSDL file from the server. By contrast, with a `.proto` file, changes are simple to apply and automatically flow through the generated code immediately.
+
+When compared with XML, messages encoded using Protocol buffers have numerous advantages:
```
JamesNK: Disadvantage is messages are not human readable. Additional tooling and the proto file is required to debug message content.

docs/architecture/grpc-for-wcf-developers/why-grpc.md
```diff
@@ -1,11 +1,41 @@
...
+
+### Performance
+
+As already discussed, using HTTP/2 rather than HTTP/1.1 removes the requirement for human-readability code and instead use the smaller faster binary protocol. This is more efficient for computers to parse. HTTP/2 also supports multiplexing requests over a single connection enabling responses to be sent as soon as they are ready without the need to wait in a queue.  Fewer resources are needed when using gRPC which makes it a good solution to use for mobile devices and over slower networks. [Mark to check]
+
+### Interoperability
+
+There are gRPC tools and libraries for all major programming languages and platforms, including .NET, Java, Python, Go, C++, Node.js, Swift, Dart, Ruby, and PHP. Thanks to the Protocol Buffers binary wire format and the efficient code generation for each platform, developers can achieve performance better than WCF's NetTCP bindings while still enjoying full cross-platform support.
+
+### Simplicity and productivity
```
JamesNK: I recall that the docs team doesn't like the term simplicity - it is broad and can be meaningless. You should be more specific // @mairaw

I described gRPC as being prescriptive instead here - https://docs.microsoft.com/en-us/aspnet/core/grpc/comparison?view=aspnetcore-3.0#strict-specification

docs/architecture/grpc-for-wcf-developers/why-grpc.md
```diff
@@ -1,11 +1,41 @@
...
+gRPC is a comprehensive RPC solution. It has thoroughly addressed the full range of connectivity issues. gRPC works consistently across multiple languages. gRPC is simple with much of the boilerplate auto-generated. Therefore more developer time is freed up to focus on business Logic.
+
+### Streaming
+
+gRPC has full bi-directional streaming, which is very similar to WCF's Full Duplex services but without the dependency on complicated network bindings and transports. gRPC streaming can operate over regular internet connections, load balancers and service meshes.
+
+### Security
+
+gRPC is implicitly secure when using HTTP/2 over an SSL/TLS end-to-end encrypted connection. Support for Client Certificate authentication (see chapter 6) further increases security and trust between client and server.
+
```
JamesNK: Add timeouts/cancellation - https://docs.microsoft.com/en-us/aspnet/core/grpc/comparison?view=aspnetcore-3.0#deadlinetimeouts-and-cancellation

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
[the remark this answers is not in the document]
JamesNK: > My fear is that is content won't age well

Updating the doc in the future won't be that difficult...

docs/architecture/grpc-for-wcf-developers/network-protocols.md
```diff
@@ -1,11 +1,35 @@
...
+
+HTTP/2, released by Internet Engineering Task Force in 2015 was derived from the experimental SPDY protocol, which was already being used by Google. It was specifically designed to be simpler, faster and more secure than HTTP/1.1.
+
+## Key features of HTTP/2
+
+Some of the key features and advantages of HTTP/2 are discussed below.
+
+### Binary protocol
+
+Request/response cycles no longer need text commands. This simplifies and speeds up implementation of commands. Specifically, parsing data is faster and uses less memory, network latency is reduced with obvious related improvements to speed and there's an overall better use of network resources.
```
mairaw:
```suggestion
Request/response cycles no longer need text commands. This simplifies and speeds up implementation of commands. Specifically, parsing data is faster and uses less memory, network latency is reduced with obvious related improvements to speed, and there's an overall better use of network resources.
```

docs/architecture/grpc-for-wcf-developers/network-protocols.md
```diff
@@ -1,11 +1,35 @@
...
+
+### Server push
+
+Servers are now able to send data to the client without receiving a request. Another efficiency saving. This further increases the efficient use of network resources by anticipating additional information that hasn't been requested yet and pushing it at the same time as the original request, freeing up future TCP connections by doing that.
+
+### Request multiplexing over a single TCP connection
+
+This is one of the most important features of HTTP/2. By allowing multiple parallel requests for data, it's now possible to download web files concurrently from a single server. Websites load faster and the need for optimization is reduced.
```
mairaw:
```suggestion
This feature is one of the most important features of HTTP/2. By allowing multiple parallel requests for data, it's now possible to download web files concurrently from a single server. Websites load faster and the need for optimization is reduced.
```

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
```diff
@@ -1,11 +1,35 @@
 ---
 title: WCF bindings and transports - gRPC for WCF Developers
-description: TO BE WRITTEN
+description: Learn how the various WCF bindings and transports compare to gRPC.
```
mairaw:
```suggestion
description: Learn how the different WCF bindings and transports compare to gRPC.
```

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
```diff
@@ -1,11 +1,35 @@
...
 author: markrendle
 ms.date: 09/02/2019
 ---
 
 # WCF bindings and transports
 
+WCF has lots of different built-in *bindings* specifying different network protocols, wire formats, and other implementation details. gRPC effectively has just one network protocol and one wire format. However, you're still likely to discover that gRPC offers the best solution in most cases. What follows is a short discussion about the most relevant WCF bindings and how they compare to their equivalent in gRPC.
```
mairaw:
```suggestion
WCF has lots of different built-in *bindings* that specify different network protocols, wire formats, and other implementation details. gRPC effectively has just one network protocol and one wire format. However, you're still likely to discover that gRPC offers the best solution in most cases. What follows is a short discussion about the most relevant WCF bindings and how they compare to their equivalent in gRPC.
```

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
```diff
@@ -1,11 +1,35 @@
...
+
+## Named Pipes
+
+WCF provided a Named Pipes binding for communication between processes on the same physical machine. Named Pipes is a proprietary Windows feature and doesn’t work like TCP, so HTTP/2 doesn’t work over Named Pipes, meaning gRPC can’t work over Named Pipes.
+
+Outside Windows, the functionality provided by Named Pipes is instead generally provided by Unix Domain Sockets. These sockets are regular TCP-like sockets represented with file-system addresses, such as `/var/run/docker.sock`, which gRPC can work with as both client and server. If you need to use Named Pipes style functionality on Windows, the next update to Windows 10 and Windows Server, in Fall 2019, adds Domain Sockets as a fully supported native feature within Windows. So, gRPC services running on these and later versions of Windows (or on Linux) can use Domain Sockets instead of Named Pipes. However, if your team is unable to update to the latest version of Windows, then you'll need to use localhost TCP addresses. Security concerns about using local TCP sockets can be alleviated with the use of certificate authentication between client and server.
```
mairaw: Fall 2019 has already finished in the southern hemisphere. Can you be more specific about the date? Maybe talk about in terms of quarters?

alleviate is considered a formal word. Cam you use a less formal synonym?

docs/architecture/grpc-for-wcf-developers/why-grpc.md
```diff
@@ -1,11 +1,41 @@
...
 title: Why gRPC is recommended for WCF developers - gRPC for WCF Developers
-description: TO BE WRITTEN
+description: A discussion of why gRPC is a good fit for WCF developers looking to migrate to modern architectures and platforms.
 author: markrendle
 ms.date: 09/02/2019
 ---
 
 # Why gRPC is recommended for WCF developers
 
+Before diving deeply into the language and techniques of gRPC, it's worth pulling together some ideas about why gRPC is the right solution for WCF developers who want to migrate to .NET Core, given there are alternatives available.
```
mairaw: worth pulling together some ideas about > maybe saying "it's worth discussing" would be better here? more direct?

docs/architecture/grpc-for-wcf-developers/why-grpc.md
```diff
@@ -1,11 +1,41 @@
...
+
+## Benefits of gRPC
+
+Additional reasons why gRPC stands above other solutions are:
+
+### Performance
+
+As already discussed, using HTTP/2 rather than HTTP/1.1 removes the requirement for human-readability code and instead use the smaller faster binary protocol. This is more efficient for computers to parse. HTTP/2 also supports multiplexing requests over a single connection enabling responses to be sent as soon as they're ready without the need to wait in a queue. Fewer resources are needed when using gRPC, which makes it a good solution to use for mobile devices and over slower networks. [Mark to check]
```
mairaw:
```suggestion
As already discussed, using HTTP/2 rather than HTTP/1.1 removes the requirement for human-readable code and instead uses the smaller faster binary protocol. This is more efficient for computers to parse. HTTP/2 also supports multiplexing requests over a single connection enabling responses to be sent as soon as they're ready without the need to wait in a queue. Fewer resources are needed when using gRPC, which makes it a good solution to use for mobile devices and over slower networks. [Mark to check]
```

mairaw: Don't forget to remove [Mark to check]

docs/architecture/grpc-for-wcf-developers/why-grpc.md
```diff
@@ -1,11 +1,41 @@
...
+
+There are gRPC tools and libraries for all major programming languages and platforms, including .NET, Java, Python, Go, C++, Node.js, Swift, Dart, Ruby, and PHP. Thanks to the Protocol Buffers binary wire format and the efficient code generation for each platform, developers can achieve better performance than WCF's NetTCP bindings while still enjoying full cross-platform support.
+
+### Simplicity and productivity
+
+gRPC is a comprehensive RPC solution. It has thoroughly addressed the full range of connectivity issues. gRPC works consistently across multiple languages. gRPC is simple with much of the boilerplate auto-generated. So, more developer time is freed up to focus on business logic.
+
+### Streaming
+
+gRPC has full bi-directional streaming, which is very similar to WCF's Full Duplex services but without the dependency on complicated network bindings and transports. gRPC streaming can operate over regular internet connections, load balancers, and service meshes.
```
mairaw:
```suggestion
gRPC has full bidirectional streaming, which is very similar to WCF's Full Duplex services but without the dependency on complicated network bindings and transports. gRPC streaming can operate over regular internet connections, load balancers, and service meshes.
```

mairaw (commented):
I hope you don't mind I made the edits directly @markrendle. I thought that would speed up the review process, so please review my changes. But I'm also leaving some comments for additional things that I missed while editing or places I wasn't sure what to do. There are other ways we can go about this. Just let me know what you prefer.

docs/architecture/grpc-for-wcf-developers/why-grpc.md
Rick-Anderson: https://microsoft-ce-csi.acrolinx.cloud/htmldata/en/rules/help/avoid_simply.html
**Avoid "simple/simply"**
It's generally a good idea to avoid describing instructions with simple or simply as the reader might not find it so simple.

docs/architecture/grpc-for-wcf-developers/grpc-overview.md
```diff
@@ -1,11 +1,29 @@
...
+description: Learn about the set of principles guiding the development of gRPC.
 author: markrendle
 ms.date: 09/02/2019
 ---
 
 # gRPC overview
 
+Chapter 1 looked at the genesis of both WCF and gRPC. This chapter will consider some of the key features of gRPC and how they compare to WCF.
+
+ASP.NET Core 3.0 is the first release of ASP.NET that natively supports gRPC as a first-class citizen, with the implementation contributed to the gRPC core project by Microsoft teams. It's recommended as the best approach for building distributed applications with .NET that can interoperate with all other major programming languages and frameworks.
```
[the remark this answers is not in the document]
markrendle: Yes. Changed to `ASP.NET Core 3.0 is the first release of ASP.NET that natively supports gRPC as a first-class citizen, with Microsoft teams contributing to the official .NET implementation of gRPC.`

markrendle:
@mairaw 
> I hope you don't mind I made the edits directly @markrendle.

I don't mind at all, it makes life much easier for the copy-editing changes.

docs/architecture/grpc-for-wcf-developers/why-grpc.md
```diff
@@ -1,11 +1,45 @@
...
+
+### Interoperability
+
+There are gRPC tools and libraries for all major programming languages and platforms, including .NET, Java, Python, Go, C++, Node.js, Swift, Dart, Ruby, and PHP. Thanks to the Protocol Buffers binary wire format and the efficient code generation for each platform, developers can achieve better performance than WCF's NetTCP bindings while still enjoying full cross-platform support.
+
+### Usability and productivity
+
+gRPC is a comprehensive RPC solution. It has thoroughly addressed the full range of connectivity issues. gRPC works consistently across multiple languages. gRPC provides excellent tooling with much of the boilerplate auto-generated. So, more developer time is freed up to focus on business logic.
```
JunTaoLuo: Curious what kind of connectivity issues are you referring to here?

docs/architecture/grpc-for-wcf-developers/why-grpc.md
```diff
@@ -1,11 +1,45 @@
...
+
+## Benefits of gRPC
+
+Additional reasons why gRPC stands above other solutions are:
+
+### Performance
+
+As already discussed, using HTTP/2 rather than HTTP/1.1 removes the requirement for human-readable messages and instead uses the smaller faster binary protocol. This is more efficient for computers to parse. HTTP/2 also supports multiplexing requests over a single connection enabling responses to be sent as soon as they're ready without the need to wait in a queue. Fewer resources are needed when using gRPC, which makes it a good solution to use for mobile devices and over slower networks.
```
JunTaoLuo: > enabling responses to be sent as soon as they're ready without the need to wait in a queue

Is this referring to HOL blocking?

docs/architecture/grpc-for-wcf-developers/why-grpc.md
```diff
@@ -1,11 +1,45 @@
...
+
+gRPC is a comprehensive RPC solution. It has thoroughly addressed the full range of connectivity issues. gRPC works consistently across multiple languages. gRPC provides excellent tooling with much of the boilerplate auto-generated. So, more developer time is freed up to focus on business logic.
+
+### Streaming
+
+gRPC has full bidirectional streaming, which is very similar to WCF's Full Duplex services but without the dependency on complicated network bindings and transports. gRPC streaming can operate over regular internet connections, load balancers, and service meshes.
+
+### Deadline/timeouts and cancellation
+
+gRPC allows clients to specify a maximum time for an RPC to complete. If the deadline is exceeded the server can cancel the operation. Deadlines and cancellations can be propagated through further gRPC calls to help enforce resource usage limits.
```
JunTaoLuo: The client can also abort the operation due to deadline being exceeded.

JunTaoLuo (commented):
A few thoughts, not all require an explicit reaction though.

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
```diff
@@ -1,11 +1,31 @@
...
+
+One of the advantages of the Protobuf IDL is that as a custom language it enables gRPC to be completely language and platform agnostic, not favoring any technology over another.
+
+The Protobuf IDL is also much easier for humans to both read and write than WSDL. Changing the WSDL of a WCF service typically requires making the changes to the service code itself, running the service and regenerating the WSDL file from the server. By contrast, with a `.proto` file, changes are simple to apply and automatically flow through the generated code immediately.
+
+When compared with XML, messages encoded using Protobuf have many advantages:
+
+- They can be up to 10x smaller.
```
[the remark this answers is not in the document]
markrendle: I've got no specific reference, but the 10x is an estimate based on a test I did using a dataset with lists of records, where SOAP repeated opening and closing XML tags all over the place while Protobuf used its packed array format.

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
```diff
@@ -1,11 +1,31 @@
...
+
+gRPC uses the Interface Definition Language (IDL) from Protocol Buffers. The Protocol Buffers IDL is a custom, platform-neutral language with an open specification. Developers hand-code `.proto` files to describe services along with their inputs and outputs. These `.proto` files can then be used to generate language- or platform-specific stubs for clients and servers, allowing multiple different platforms to communicate. By sharing `.proto` files, teams can generate code to use each others' services without needing to take a code dependency.
+
+One of the advantages of the Protobuf IDL is that as a custom language it enables gRPC to be completely language and platform agnostic, not favoring any technology over another.
+
+The Protobuf IDL is also much easier for humans to both read and write than WSDL. Changing the WSDL of a WCF service typically requires making the changes to the service code itself, running the service and regenerating the WSDL file from the server. By contrast, with a `.proto` file, changes are simple to apply and automatically flow through the generated code immediately.
```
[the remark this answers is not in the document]
markrendle: Clarified.

docs/architecture/grpc-for-wcf-developers/why-grpc.md
markrendle: Yes

docs/architecture/grpc-for-wcf-developers/why-grpc.md
markrendle: Issues is the wrong word.

markrendle: Changes to "networking and concerns". Basically just saying it does networking properly.

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
```diff
@@ -1,11 +1,35 @@
...
 title: WCF bindings and transports - gRPC for WCF Developers
-description: TO BE WRITTEN
+description: Learn how the different WCF bindings and transports compare to gRPC.
 author: markrendle
 ms.date: 09/02/2019
 ---
 
 # WCF bindings and transports
 
+WCF has lots of different built-in *bindings* that specify different network protocols, wire formats, and other implementation details. gRPC effectively has just one network protocol and one wire format. However, you're still likely to discover that gRPC offers the best solution in most cases. What follows is a short discussion about the most relevant WCF bindings and how they compare to their equivalent in gRPC.
```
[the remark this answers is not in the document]
markrendle: Added a note that it's customizable but that's beyond the scope of this book.

docs/architecture/grpc-for-wcf-developers/network-protocols.md
```diff
@@ -1,11 +1,35 @@
...
+
+## Key features of HTTP/2
+
+The following list shows some of the key features and advantages of HTTP/2:
+
+### Binary protocol
+
+Request/response cycles no longer need text commands. This simplifies and speeds up implementation of commands. Specifically, parsing data is faster and uses less memory, network latency is reduced with obvious related improvements to speed, and there's an overall better use of network resources.
+
+### Server push
```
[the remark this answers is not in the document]
markrendle: Which feature? Server push?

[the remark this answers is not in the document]
markrendle: Replaced with description of HTTP/2 Streams.

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
```diff
@@ -1,11 +1,31 @@
...
+
+One of the advantages of the Protobuf IDL is that as a custom language it enables gRPC to be completely language and platform agnostic, not favoring any technology over another.
+
+The Protobuf IDL is also much easier for humans to both read and write than WSDL. Changing the WSDL of a WCF service typically requires making the changes to the service code itself, running the service and regenerating the WSDL file from the server. By contrast, with a `.proto` file, changes are simple to apply and automatically flow through the generated code. Visual Studio 2019 builds `.proto` files in the background when they are saved; when using other editors such as VS Code the changes will be applied when the project is built.
```
mconnew: Would it be better to describe this as Protobug IDL is intended for humans to read and write whereas WSDL was intended to be read and written programatically?

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
```diff
@@ -1,11 +1,31 @@
...
+
+When compared with XML, messages encoded using Protobuf have many advantages:
+
+- They can be up to 10x smaller.
```
mconnew: This isn't true when you use WCF's binary message encoding. The difference is only about 10% in size.

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
```diff
@@ -1,11 +1,31 @@
...
+- Encoding, decoding and transmitting can be up to 100x faster.
```
mconnew: This also isn't true. When using binary message encoding, WCF's encoding is about 10% slower than protobuf. Also any benchmarks comparing NetTcp with protobuf are showing WCF to be significantly faster.

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
```diff
@@ -1,11 +1,31 @@
...
+- Simpler and less ambiguous.
+- The generated data access classes are easier to use programmatically.
```
mconnew: Are they? In what way? They are just classes. WCF has the added capability of being able to handle object graphs with cycles which gRPC can't do. You need to turn on the feature with WCF, but it can do it whereas gRPC can't. So WCF can handle application OM classes better than gRPC where you will sometimes need to create a specific Data OM to handle the protobuf limitations.

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
```diff
@@ -1,11 +1,31 @@
...
+
+The potential disadvantage of Protobuf is that, because the messages are not human readable, additional tooling is required to debug message content.
```
mconnew: The primary use of WCF when used for RPC uses the binary message encoder so this is only true when using the text message encoder. Otherwise you have the same problems with WCF. Actually sometimes it can be worse as WCF has a session dictionary which requires you to have a copy of every message sent in the session prior to the one you are interested in to be able to guarantee you can debug the message content. So in some ways, even with tooling a protobuf message can be easier to deal with.

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
```diff
@@ -1,11 +1,35 @@
...
 ms.date: 09/02/2019
 ---
 
 # WCF bindings and transports
 
+WCF has lots of different built-in *bindings* that specify different network protocols, wire formats, and other implementation details. gRPC effectively has just one network protocol and one wire format (technically the wire format *may* be customized, but that is beyond the scope of this book). You are likely to discover that gRPC offers the best solution in most cases. What follows is a short discussion about the most relevant WCF bindings and how they compare to their equivalent in gRPC.
+
+## NetTCP
+
+WCF's NetTCP binding allows for persistent connections, small messages, and two-way messaging but only works between .NET clients and servers. gRPC allows the same functionality but based on an open standard. gRPC has almost all the features of WCF NetTCP binding, although not always in the same way. For example, in WCF, encryption is controlled through configuration and handled in the framework. In gRPC, encryption is achieved at the connection level using HTTP/2 over TLS.
```
mconnew: I'm not sure about semantics here, but the NetTcp framing protocols and message framing protocols have been published by Microsoft for the purposes of others creating an inter-operable implementation. So the standards are open, but not community developed.

mconnew: gRPC does not have almost all the features of WCF NetTCP binding. WCF NetTcp binding supports distributed transactions, reliable sessions, message level encryption, NTAuth negotiation encryption, impersonation, delegation and that's just off the top of my head. If I spend the time to look at all the properties on the binding I'm sure there's some I missed.

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
JamesNK: Just above this (line 18) the doc mentions that these comparisons are to WCF with XML.

I think it is worth discussing either there or below this list that WCF also supports more efficient binary messaging.

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
```diff
@@ -1,11 +1,35 @@
...
+
+## HTTP
+
+WCF is only cross-platform when using SOAP over HTTP/1.1. WCF HTTP binding is text based, uses SOAP as the wire format, and is very slow by the standards of modern networked applications. It's only really used to provide cross-platform interoperability, or connection over internet infrastructure. The equivalent in gRPC—because it uses HTTP/2 as the underlying transport layer with the binary Protobuf wire format for messages—can offer Net TCP service level performance but with full cross-platform interoperability with all modern programming languages and frameworks.
```
mconnew: WCF is cross-platform using transports other than HTTP/1.1. You can also use SOAP over UDP which is WCF's UdpBinding: https://en.wikipedia.org/wiki/SOAP-over-UDP
There's also WS-Discovery which uses UDP and that's cross-platform too.
As for text based, you have to be specific, WCF's BasicHttpBinding with default properties uses text, but you can get it to use binary message encoding, or MTOM (which is open standards too). WCF also has NetHttpBinding which uses the binary message encoder by default. So you need to disambiguate that you are talking about the BasicHttpBinding what you say it's text based (by default). WCF also supports WebSockets so can run a service which would normally require NetTcp and have it work over the internet using WebSockets.

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
```diff
@@ -1,11 +1,35 @@
...
+
+## Named Pipes
+
+WCF provided a Named Pipes binding for communication between processes on the same physical machine. Named Pipes is a proprietary Windows feature and doesn’t work like TCP, so HTTP/2 doesn’t work over Named Pipes, meaning gRPC can’t work over Named Pipes.
```
mconnew: There is nothing stopping gRPC from working over named pipes. It's simply a different uri scheme. Once you have a named pipe connected, WCF uses the exact same code for named pipes and NetTcp with a basic send/receive abstraction layer. There's no reason why gRPC can't be run over named pipes.

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
```diff
@@ -1,11 +1,35 @@
...
+
+Outside Windows, the functionality provided by Named Pipes is instead generally provided by Unix Domain Sockets. These sockets are regular TCP-like sockets represented with file-system addresses, such as `/var/run/docker.sock`, which gRPC can work with as both client and server. If you need to use Named Pipes style functionality on Windows, the next update to Windows 10 and Windows Server, in 2019 Q4, adds Domain Sockets as a fully supported native feature within Windows. So, gRPC services running on these and later versions of Windows (or on Linux) can use Domain Sockets instead of Named Pipes. However, if your team is unable to update to the latest version of Windows, then you'll need to use localhost TCP sockets. Security concerns about using local TCP sockets can be addressed with the use of certificate authentication between client and server.
```
mconnew: @davidfowl, is AspNetCore going to support traditional Windows named pipes or will it only support unix domain sockets? CoreWCF will need Windows named pipes support and I thought this was already in the pipeline?

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
```diff
@@ -1,11 +1,35 @@
...
+
+## MSMQ
+
+MSMQ is a proprietary Windows message queue. WCF's binding to MSMQ enables "fire and forget" requests from clients that may be processed at any time in the future. gRPC doesn't natively provide any message queue functionality. The best alternative would be to directly use a messaging system like Azure Service Bus, RabbitMQ, or Kafka. This could be implemented with the client placing messages directly onto the queue, or a gRPC client streaming service that enqueues the messages.
+
+## WCF REST
+
+The WCF REST extensions, such as `WebGet` or `WebInvoke`, enabled you to develop RESTful APIs that could speak JSON at a time when this was less common than now. Therefore, if you have a RESTful API built with WCF REST, consider migrating it to a regular ASP.NET Core MVC Web API application, which would provide equivalent functionality, rather than converting to gRPC.
```
mconnew: The WCF REST extension can be summed up as the WebHttpBinding which supports the WebGet and WebInvoke attributes. You should at least mention this binding as someone reading this might be aware they use WebHttpBinding but not connect the dots to WebGet or WebInvoke. Whenever we're asked for this, WebHttpBinding is what's usually asked for, not WebGet or WebInvoke.

docs/architecture/grpc-for-wcf-developers/why-grpc.md
```diff
@@ -1,11 +1,45 @@
...
+
+Additional reasons why gRPC stands above other solutions are:
+
+### Performance
+
+As already discussed, using HTTP/2 rather than HTTP/1.1 removes the requirement for human-readable messages and instead uses the smaller faster binary protocol. This is more efficient for computers to parse. HTTP/2 also supports multiplexing requests over a single connection enabling responses to be sent as soon as they're ready without the need to wait in a queue (an issue in HTTP/1.1 known as "head-of-line (HOL) blocking"). Fewer resources are needed when using gRPC, which makes it a good solution to use for mobile devices and over slower networks.
+
+### Interoperability
+
+There are gRPC tools and libraries for all major programming languages and platforms, including .NET, Java, Python, Go, C++, Node.js, Swift, Dart, Ruby, and PHP. Thanks to the Protocol Buffers binary wire format and the efficient code generation for each platform, developers can achieve better performance than WCF's NetTCP bindings while still enjoying full cross-platform support.
```
mconnew: I'm yet to see any benchmarks showing gRPC out-performing NetTcp. All the benchmarks showing better performance are comparing BasicHttpBinding.

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
JamesNK: It has been a while since I touched WCF client APIs but I suspect configuration will be easier with gRPC since gRPC is more prescriptive. And I suspect streaming is much easier using gRPC .NET APIs than WCF APIs.

But yeah, being more specific would be good here.

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
JamesNK: Some other frameworks support gRPC over pipes (e.g. Java) but we can't do it in .NET yet - https://github.com/dotnet/core/issues/2312

docs/architecture/grpc-for-wcf-developers/why-grpc.md
```diff
@@ -1,11 +1,45 @@
...
+
+### Usability and productivity
+
+gRPC is a comprehensive RPC solution. It has thoroughly addressed the full range of networking and connectivity concerns. gRPC works consistently across multiple languages. gRPC provides excellent tooling, with much of the boilerplate auto-generated, so more developer time is freed up to focus on business logic.
```
mconnew: I think the phrase "thoroughly addressed the full range of networking and connectivity concerns" is going to result in some negative pushback because of how vague it is. What does "the full range of networking and connectivity concerns" include? Reliable session solves the problem of your connection dropping and retrying without the server processing the request twice. I would say this is included in "connectivity concerns" but gRPC doesn't solve that problem. It also doesn't solve the problem of ensuring in-order execution of requests. You should be able to send multiple request serially without waiting for the prior response to complete (think high latency network) but guarantee they execute in order serially on the service. gRPC can't do this out of the box.

docs/architecture/grpc-for-wcf-developers/why-grpc.md
```diff
@@ -1,11 +1,45 @@
...
+
+### Streaming
+
+gRPC has full bidirectional streaming, which is very similar to WCF's Full Duplex services but without the dependency on complicated network bindings and transports. gRPC streaming can operate over regular internet connections, load balancers, and service meshes.
```
mconnew: WCF doesn't need a complicated network binding to achieve this. It supports full duplex with WebSockets which can operate over regular internet connections and load balancers. Not sure what you mean by service meshes. Just use NetHttpBinding with default options and use a duplex contract and it just works, no complication.

docs/architecture/grpc-for-wcf-developers/why-grpc.md
```diff
@@ -1,11 +1,45 @@
...
+
+### Deadline/timeouts and cancellation
+
+gRPC allows clients to specify a maximum time for an RPC to complete. If the specified deadline is exceeded the server can cancel the operation independently of the client. Deadlines and cancellations can be propagated through further gRPC calls to help enforce resource usage limits. Clients may also abort operations when a deadline is exceeded, or earlier if necessary (e.g. due to a user interaction).
```
mconnew: This can be achieved with WCF using distributed transactions. When the deadline has been reached, the client aborts the transaction which will tell each server involved (including SQL server which gRPC can't propagate it's deadline to).

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
markrendle: I benchmarked NetTCP against gRPC and gRPC was faster and used less memory. https://visualrecode.com/blog/wcf-vs-grpc-round-2/

docs/architecture/grpc-for-wcf-developers/why-grpc.md
markrendle: We're not saying WCF can't do these things, just that gRPC can do them.

docs/architecture/grpc-for-wcf-developers/why-grpc.md
markrendle: https://visualrecode.com/blog/wcf-vs-grpc-round-2/

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
markrendle: Removed that sentence.

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
markrendle: Changed to `many of`.

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
markrendle: Removed that bullet point. Also the one above which wasn't saying anything.

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
markrendle: Added a note.

mconnew: Binary encoding is XML, it's just a binary encoded version of xml.

mconnew: Someone made a blog post and provided code comparing the different serializers and their formats including DataContractSerializer and protobuf among others. I forked their code and added a variant using the binary xml dictionary reader/writer. You can find my fork [here](https://github.com/mconnew/SerializationPerformanceTest_CSharp/tree/BinaryXml) on the BinaryXml branch.  

These are the results:  

**Simple Object**  

| Serializer | Object size | Serialization time | Deserialization time |
| ------------- | ------------- |------------- |------------- |
| DataContractSerializer (text) | 370 | 0.002255ms | 0.005327ms |
| DataContractSerializer (binary) | 83 | 0.000864ms | 0.002067ms |
| Protobuf | 62 | 0.000522ms | 0.001207 ms |

You can see with a simple object, even when using the text xml encoder (which NetTcp doesn't used by default) it's less than 6X bigger. When using the binary xml encoder, protobuf is only 1.33X smaller.

**Array of simple objects**  

| Serializer | Object size | Serialization time | Deserialization time |
| ------------- | ------------- |------------- |------------- |
| DataContractSerializer (text) | 364,299 | 2.431831ms | 4.770156ms |
| DataContractSerializer (binary) | 125,974 | 1.270527ms | 3.068528ms |
| Protobuf | 104,191 | 0.667872ms | 2.445094ms |

When serializing something significantly bigger, there's still nowhere near 10X. For an array of simply objects, when using the text xml encoder it's less than 3.5X bigger. When using the binary xml encoder, protobuf is only 1.21X smaller.

Remember, this is still using the DataContractSerializer so it's still XML, just in a binary form. But even if you are comparing to text, it's a lot less than 10X. I haven't even done anything to try and highlight the strengths of WCF. For example, if I were to enable references with the DataContractSerializer (removes duplicates objects from the output and replaces with a reference) and were to insert the exact same simple object in every position of the array, XML would come out significantly smaller than protobuf as protobuf can't do that.

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
mconnew: In your original [blog post](https://unwcf.com/posts/wcf-vs-grpc-round-2/) I listed some problems with your comparison where you were comparing apples to oranges. In your comment reply to me you said you believed neither scenario was using any encryption. WCF definitely was, so presuming gRPC wasn't using any encryption, that will make a massive difference. If gRPC is using encryption, it's using TLS whereas you are using Windows auth for WCF which translates to using NegotiateStream. This are different encryption mechanisms so there could be a difference in performance there. Also, you were running WCF using workstation GC and gRPC using server GC. This again can make a significant difference, especially for server scenarios. Workstation GC has a single heap that all processors have to contend with to get access to. It also means when you do need to do a GC pause, you have a single core doing the work instead of all your cores so pauses are longer.  
And if all those problems aren't enough, your own results with a hamstrung WCF only show grpc to be 1.11X faster, not 100X.

markrendle: Look, I'm not an expert at configuring WCF to get the best performance, but to be fair, I'm not an expert at configuring gRPC to get the best performance either. They were both running on whatever their default settings were, which for gRPC is obviously HTTP/2 over TLS, I don't know what it is for NetTCP.

_Anyway_, it's irrelevant here because the comparison is not gRPC versus WCF-with-NetTCP-binding. It's Protobuf versus XML.

> **When compared with XML**, messages encoded using Protobuf have many advantages:
> 
> - They can be up to 10x smaller.
> - Encoding, decoding and transmitting can be up to 100x faster.

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
```diff
@@ -1,11 +1,29 @@
...
+
+One of the advantages of the Protobuf IDL is that as a custom language it enables gRPC to be completely language and platform agnostic, not favoring any technology over another.
+
+The Protobuf IDL is also designed for humans to both read and write, whereas WSDL is intended as a machine-readable/writable format. Changing the WSDL of a WCF service typically requires making the changes to the service code itself, running the service and regenerating the WSDL file from the server. By contrast, with a `.proto` file, changes are simple to apply and automatically flow through the generated code. Visual Studio 2019 builds `.proto` files in the background when they are saved; when using other editors such as VS Code the changes will be applied when the project is built.
```
JamesNK:
```suggestion
The Protobuf IDL is also designed for humans to both read and write, whereas WSDL is intended as a machine-readable/writable format. Changing the WSDL of a WCF service typically requires making the changes to the service code itself, running the service and regenerating the WSDL file from the server. By contrast, with a `.proto` file, changes are simple to apply with a text editor, and automatically flow through the generated code. Visual Studio 2019 builds `.proto` files in the background when they are saved; when using other editors such as VS Code the changes will be applied when the project is built.
```

docs/architecture/grpc-for-wcf-developers/network-protocols.md
```diff
@@ -1,11 +1,35 @@
...
+description: An overview of the gRPC network protocols.
 author: markrendle
 ms.date: 09/02/2019
 ---
 
 # Network protocols
 
+Unlike WCF, gRPC uses HTTP/2 as a base for its networking. This offers significant advantages over WCF and SOAP, which only operate on HTTP/1.1. For developers wanting to use gRPC, given that there's no alternative to HTTP/2, it would seem to be the ideal moment to explore HTTP/2 in more detail and identify additional benefits to using gRPC.
+
+HTTP/2, released by Internet Engineering Task Force in 2015, was derived from the experimental SPDY protocol, which was already being used by Google. It was specifically designed to be simpler, faster, and more secure than HTTP/1.1.
```
JamesNK: Is HTTP/2 really simpler than HTTP/1? I now know a lot about HTTP/2 and it is pretty complicated 😋

docs/architecture/grpc-for-wcf-developers/why-grpc.md
```diff
@@ -1,11 +1,45 @@
...
+
+gRPC has full bidirectional streaming, which provides very similar functionality to WCF's full duplex services. gRPC streaming can operate over regular internet connections, load balancers, and service meshes.
+
+### Deadline/timeouts and cancellation
+
+gRPC allows clients to specify a maximum time for an RPC to complete. If the specified deadline is exceeded the server can cancel the operation independently of the client. Deadlines and cancellations can be propagated through further gRPC calls to help enforce resource usage limits. Clients may also abort operations when a deadline is exceeded, or earlier if necessary (e.g. due to a user interaction).
+
+### Security
+
+gRPC is implicitly secure when using HTTP/2 over an TLS end-to-end encrypted connection. Support for Client Certificate authentication (see chapter 6) further increases security and trust between client and server.
```
JamesNK: @blowdart This paragraph seems quite strong. Are you happy with its content?

JamesNK (approved):

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
davidfowl: There’s no wok item for it just prototypes that I’ve written

docs/architecture/grpc-for-wcf-developers/network-protocols.md
davidfowl: Yea not simpler but more efficient.

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
```diff
@@ -1,11 +1,29 @@
...
+
+One of the advantages of the Protobuf IDL is that as a custom language it enables gRPC to be completely language and platform agnostic, not favoring any technology over another.
+
+The Protobuf IDL is also designed for humans to both read and write, whereas WSDL is intended as a machine-readable/writable format. Changing the WSDL of a WCF service typically requires making the changes to the service code itself, running the service and regenerating the WSDL file from the server. By contrast, with a `.proto` file, changes are simple to apply with a text editor, and automatically flow through the generated code. Visual Studio 2019 builds `.proto` files in the background when they are saved; when using other editors such as VS Code the changes will be applied when the project is built.
+
+When compared with XML, and particularly SOAP, messages encoded using Protobuf have many advantages. Protobuf messages can be up to 10x smaller than the same data serialized as SOAP XML, and encoding, decoding and transmitting them over a network can be up to 100x faster.
```
shirhatti: @markrendle Can we remove references to actual numbers here? It's something we'd like to add back in later once we've measured it in a re-produce-able test. We're trying to avoid upsetting the applecart for folks already using WCF.

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
```diff
@@ -1,11 +1,35 @@
...
+
+## Named Pipes
+
+WCF provided a Named Pipes binding for communication between processes on the same physical machine. Named Pipes are not supported by the first release of ASP.NET Core gRPC.
+
+Outside Windows, the functionality provided by Named Pipes is instead generally provided by Unix Domain Sockets. These sockets are regular TCP-like sockets represented with file-system addresses, such as `/var/run/docker.sock`, which gRPC can work with as both client and server. If you need to use Named Pipes style functionality on Windows, the next update to Windows 10 and Windows Server, in 2019 Q4, adds Domain Sockets as a fully supported native feature within Windows. So, gRPC services running on these and later versions of Windows (or on Linux) can use Domain Sockets instead of Named Pipes. However, if your team is unable to update to the latest version of Windows, then you'll need to use localhost TCP sockets. Security concerns about using local TCP sockets can be addressed with the use of certificate authentication between client and server.
```
shirhatti: Support for AF_UNIX address family was added in Windows Server, version 1803 OR Windows 10 April 2018 Update

docs/architecture/grpc-for-wcf-developers/why-grpc.md
shirhatti:
```suggestion
There are gRPC tools and libraries for all major programming languages and platforms, including .NET, Java, Python, Go, C++, Node.js, Swift, Dart, Ruby, and PHP. Thanks to the Protocol Buffers binary wire format and the efficient code generation for each platform, developers can build performant apps while still enjoying full cross-platform support.
```

docs/architecture/grpc-for-wcf-developers/wcf-bindings.md
mconnew: I'll be opening an issue then as CoreWCF will need them.

docs/architecture/grpc-for-wcf-developers/why-grpc.md
blowdart: Oh, good call. "gRPC provides encryption if used over HTTP/2 and TLS. Support for certificate authentication further increases security if certificates are validated correctly."

docs/architecture/grpc-for-wcf-developers/interface-definition-language.md
shirhatti:
```suggestion
When compared with XML, and particularly SOAP, messages encoded using Protobuf have many advantages. Protobuf messages tend to be smaller than the same data serialized as SOAP XML, and encoding, decoding and transmitting them over a network can be faster.
```

mconnew (approved):

mconnew (approved):

mconnew (approved):

docs/architecture/grpc-for-wcf-developers/toc.yml
```diff
@@ -21,8 +21,6 @@
       href: interface-definition-language.md
     - name: Network protocols
       href: network-protocols.md
-    - name: WCF bindings and transports
```
mairaw: why is this article being removed from the TOC? The file is still in this PR.

shirhatti: I'm going to guess this this a mistake. We don't want to omit this file

mairaw: Ok reverting the change then

PR #14309, opened 2019-09-09
Update custom-token.md

Youssef1313:
The intend of this PR was to use `nameof` instead of the magic string. but I found some other slight changes that I included also in this PR.

# Summary:

- Use `nameof` instead of magic string.
- Leave blank line after headings.
- Use `var` when data type is clear from right side.
- Follow comments convention (space after `//`, first letter is capitalized and ends with a period).
- Remove redundant paths segment. (I know that @nemrism is currently doing a set of  PRs for that, but the current open PRs still doesn't address `custom-token.md`) (Contributes to: #10653)


rpetrusha (approved):
Thank you for making these changes, @Youssef1313. I'll merge your PR now.

PR #14313, opened 2019-09-09
Adding scenarios to Containers and Orchestrators section

ardalis:
## Summary

Adding scenarios to Containers and Orchestrators section
Fixing some images.


nishanil:
@ardalis  Some review comments:

- In the section, “Scaling Benefits” – You can also describe the declarative API model in Kubernetes that ensures high availability. It may be good to have an example of Kubernetes `config` file explaining the declarative approach. Also, mention that it provides an imperative approach for building and testing Kubernetes too. 

- In the “When should you avoid using containers..” section, we need to stress on the fact the complexity that it brings in so it is totally ok to avoid them for smaller apps.

nishanil (approved):
I am merging this. Please send another PR for adding scale-containers-serverless.md to toc and fix its nav links.

PR #14316, opened 2019-09-09
C117640Escaping backslah to avoid localization content

v-maudel:
Description: Backslash"\\" in section name is breaking formatting causing English strings to render on LOC pages.


v-maudel:
Hi @KathleenDollard, can you please merge this PR?
Thanks in advance.

mairaw (approved):
Thanks @v-maudel! I also escaped a * that I saw in a different heading.

v-maudel:
Thanks @mairaw

PR #14318, opened 2019-09-09
C117610: Closing square bracket

v-maudel:
Description: Missing hard break line and closing square bracket are creating the issue


mairaw (approved):
Thank you @v-maudel!

v-maudel:
Thanks @mairaw

PR #14323, opened 2019-09-09
Update how-to-know-the-difference-passing-a-struct-and-passing-a-clas…

### LINKED ISSUE #13768: This line is wrong grammatically and doesn't make any sense
"However, the called method can use the address to access the class members that both the original address and the copy reference."

FYI

URL: 
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/how-to-know-the-difference-passing-a-struct-and-passing-a-class-to-a-method 

---
#### Document Details

* ID: 4265e5ac-ef1c-7da2-fdf3-854494aa5a73
* Version Independent ID: 7c436200-ff45-5391-fefd-6c64f9ab91e6
* Content: [How to: Know the Difference Between Passing a Struct and Passing a Class Reference to a Method - C# Programming Guide](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/how-to-know-the-difference-passing-a-struct-and-passing-a-class-to-a-method#feedback)
* Content Source: [docs/csharp/programming-guide/classes-and-structs/how-to-know-the-difference-passing-a-struct-and-passing-a-class-to-a-method.md](https://github.com/dotnet/docs/blob/master/docs/csharp/programming-guide/classes-and-structs/how-to-know-the-difference-passing-a-struct-and-passing-a-class-to-a-method.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

Thraka:
Clearer paragraph.

Fixes #13768

From:

>However, the called method can use the address to access the class members that both the original address and the copy reference.

To (with `code` to signal new words)

>However, the called method can use the `copy` address to access the class members that both the original address and the copy `address` reference.


docs/csharp/programming-guide/classes-and-structs/how-to-know-the-difference-passing-a-struct-and-passing-a-class-to-a-method.md
```diff
@@ -13,7 +13,7 @@ The following example demonstrates how passing a [struct](../../language-referen
...
+ A class instance is a [reference type](../../language-reference/keywords/reference-types.md), not a value type. When [a reference type is passed by value](./passing-reference-type-parameters.md) to a method, the method receives a copy of the reference to the class instance. That is, the method receives a copy of the address of the instance, not a copy of the instance itself. The class instance in the calling method has an address, the parameter in the called method has a copy of the address, and both addresses refer to the same object. Because the parameter contains only a copy of the address, the called method cannot change the address of the class instance in the calling method. However, the called method can use the copy address to access the class members that both the original address and the copy address reference. If the called method changes a class member, the original class instance in the calling method also changes.
```
rpetrusha:
```suggestion
 A class instance is a [reference type](../../language-reference/keywords/reference-types.md), not a value type. When [a reference type is passed by value](./passing-reference-type-parameters.md) to a method, the method receives a copy of the reference to the class instance. That is, the called method receives a copy of the address of the instance, and the calling method retains the address of the instance. The class instance in the calling method has an address, the parameter in the called method has a copy of the address, and both addresses refer to the same object. Because the parameter contains only a copy of the address, the called method cannot change the address of the class instance in the calling method. However, the called method can use the copy of the address to access the class members that both the original address and the copy of the address reference. If the called method changes a class member, the original class instance in the calling method also changes.  
```

rpetrusha (approved):
I thought that "copy address" was less clear than "copy of the address", @Thraka. But I also thought that the confusion begins with the statement that the method does not receive a copy of the instance itself. While that's true, it muddies the issue, since receiving or not receiving a copy of an instance applies to value types only and is inapplicable here.

docs/csharp/programming-guide/classes-and-structs/how-to-know-the-difference-passing-a-struct-and-passing-a-class-to-a-method.md
Thraka: That is more clear. However, I don't know that I like all of changes suggested. I think it makes it more confusing talking about the calling method keeping the original address. I'm going to edit that change out of your suggestion :) Thanks!

docs/csharp/programming-guide/classes-and-structs/how-to-know-the-difference-passing-a-struct-and-passing-a-class-to-a-method.md
```diff
@@ -13,7 +13,7 @@ The following example demonstrates how passing a [struct](../../language-referen
...
+ A class instance is a [reference type](../../language-reference/keywords/reference-types.md), not a value type. When [a reference type is passed by value](./passing-reference-type-parameters.md) to a method, the method receives a copy of the reference to the class instance. That is, the called method receives a copy of the address of the instance, and the calling method retains the address of the instance. The class instance in the calling method has an address, the parameter in the called method has a copy of the address, and both addresses refer to the same object. Because the parameter contains only a copy of the address, the called method cannot change the address of the class instance in the calling method. However, the called method can use the copy of the address to access the class members that both the original address and the copy of the address reference. If the called method changes a class member, the original class instance in the calling method also changes.
```
Thraka:
```suggestion
 A class instance is a [reference type](../../language-reference/keywords/reference-types.md), not a value type. When [a reference type is passed by value](./passing-reference-type-parameters.md) to a method, the method receives a copy of the reference to the class instance. That is, the called method receives a copy of the address of the instance, and the calling method retains the original address of the instance. The class instance in the calling method has an address, the parameter in the called method has a copy of the address, and both addresses refer to the same object. Because the parameter contains only a copy of the address, the called method cannot change the address of the class instance in the calling method. However, the called method can use the copy of the address to access the class members that both the original address and the copy of the address reference. If the called method changes a class member, the original class instance in the calling method also changes.  
```

Thraka:
@rpetrusha 

Regarding this:

>But I also thought that the confusion begins with the statement that the method does not receive a copy of the instance itself. While that's true, it muddies the issue, since receiving or not receiving a copy of an instance applies to value types only and is inapplicable here.

I don't think that is correct. It is pointing out that while it is indeed a reference type, you still get a "copy" of the address. This means that you can (for example) initialize it to a new instance if you wanted to, and you wouldn't affect the original that was passed to the method. If you passed it `ref` then you could reassign the variable inside the method and indeed that reassignment would carry over to the callee.

PR #14347, opened 2019-09-10
Updated nuget.org's symbol server status

loic-sharma:
## Summary

Visual Studio 2019 version 16.1 added nuget.org's symbol server as a default symbol server. See https://docs.microsoft.com/en-us/visualstudio/releases/2019/release-notes-v16.1#debugger


mairaw (approved):
Thank you @loic-sharma for adding that info.

PR #14363, opened 2019-09-11
Updated .Net Core versions

gurmeetsinghdke:
Statement "either .NET Core 1.1 or .NET Core 2.0" is giving a reflection that user must use either version 1.1 or 2.0. 
So replaced text "either .NET Core 1.1 or .NET Core 2.0." with ".NET Core 1.1 or any higher version" as we have other versions available too.

## Summary

Describe your changes here.


Thraka:
Can you do this change in the other PR you have open? #14365 we've been talking in that one and it would be easier to review the changes as whole in the single PR. Thanks! Just keep modifying your Patch-2 branch.

PR #14369, opened 2019-09-11
Improve "virtual" topic

BillWagner:
See [AB#1591403](https://dev.azure.com/mseng/677da0fb-b067-4f77-b89b-f32c12bb8617/_workitems/edit/1591403)

There are two issues with this topic.

First, the click-through rate is low. Several of the links toward the bottom of the page are redundant, and unlikely to be used.

Secondly, the entire C# language reference was tagged as "conceptual" (the default). The language reference should be "language-reference" which removes the dwell time metric.

### DIFF
--- a/docfx.json
+++ b/docfx.json
@@ -166,6 +166,7 @@
       "ms.topic": {
         "_csharplang/**/*.md": "language-reference",
         "_vblang/spec/*.md": "language-reference",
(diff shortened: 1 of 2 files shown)


mcleblanc (commented):
Changes look fine to me.

rpetrusha (approved):
LGTM. You can merge when you're ready, @BillWagner.

PR #14392, opened 2019-09-12
Fix code errors

Youssef1313:
- Removed `ByVal` (To make code cleaner).
- Fixed the equal comparison operator. In vb, it's `=` not `==`. (That fixes a compile-error)
- When the body of `If` is in a new line, the compiler expects an `End If`. so I put the body in the same line. (That fixes a compile-error) (Another solution is to put `_` after `Then`. I see both are okay, but if you prefer one on the other I can edit it.
- Fixed `If` in C# code, it should be `if`. (That fixes a compile-error)


rpetrusha (changes requested):
Thank you for fixing this badly ported VB code and C# code, @Youssef1313. There are some further changes: throwing Exception is not recommended; and Date is a Visual Basic data type, not a C# data type.

rpetrusha (approved):

PR #14404, opened 2019-09-12
US 1583733 Add missing language IDs - 37

v-thpra:
- Used "console" for program output examples, including exception messages.

Contributes to #2192


mairaw (changes requested):
Thanks @v-thpra! I've left a few comments to be addressed.

mairaw (approved):
Thanks @v-thpra. Made just a small change caused by a bug in my suggestion and will merge once the build finishes.

PR #14433, opened 2019-09-14
Load Balancing

ardalis:
Adding load balancing section to chapter 3.


nishanil:
@ardalis A lot of this is already covered in the front-end communications section in Cloud Native Comm Patterns chapter. Probably, it's best to add the missing load balancer into that. Thoughts?

ardalis:
I'll merge this into ch4 content.

PR #14434, opened 2019-09-14
Monitoring/Logging Updates

ardalis:
More scenario based
Fixed descriptions
Fixed figure numbers


nishanil:
@ardalis Most of the content look good. Few suggestions here:
 - _Logging.md_ has to be renamed to _Logging with Elasitc Stack_ (as per its name) in both TOC and the file name
 - It's not clear why Azure Monitor is thrown in there. Adding relevance to cloud-native apps and how they can be consumed in the apps is missing there.
- Monitoring in AKS repeats Elastic Stack - shouldn't this be the chapter followed by logging as you have already introduced readers to elk stack there? If yes, move the _azure monitor_ to the last and add references to this chapter. Also, the identity chapter's previous has to be fixed to the corresponding end chapter.
-

ardalis:
- Renamed logging
- Azure Monitor covers a lot of the same functionality as the ELK stack and provides a native solution
- Reordered sections as requested
- ELK is mentioned in both sections, but not repeated. Logging doesn't talk about how to configure ELK for Kubernetes, which is all that's covered in the latter section.
- Fixed links (and dates)

nishanil (approved):

PR #14208, opened 2019-09-04
JSON Serialization docs

tdykstra:
Fixes #12207 

[Internal review URL - Overview](https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-overview?branch=pr-en-us-14208)
[Internal review URL - How to](https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to?branch=pr-en-us-14208)

PR #14439, opened 2019-09-15
Follow-up to #14208 

tdykstra:
Related to #14208 

Implements some review suggestions that I missed or didn't have time for earlier.


Thraka (approved):
LGTM I didn't check and review the preview site though.

PR #14453, opened 2019-09-16
move breadcrumb toc

mingwli:
## Summary

1. move `/toc.yml` to dedicated `/docs/breadcrumb/toc.yml`
2. remove route in docfx.json

## Reason

1. this makes it easier to identify that it's breadcrumb toc file
2. this has **NO** impact on existing breadcrumb and articles referencing it
3. docfx vNext may need to distinguish content-toc and breadcrumb-toc to create better experience for authors, and this is a prerequisite


mairaw (approved):
Thanks @mingwli. This changes makes sense and it seems the breadcrumb continues to work. 😄

PR #14460, opened 2019-09-16
Capitalized reference to "Dockerfile"

Auios:
There was a reference to the "Dockerfile" in the documents but it was shown as "dockerfile". This was confusing because every other reference to the "Dockerfile" was capitalized.

## Summary

Describe your changes here.

### DIFF
--- a/docs/core/docker/build-container.md
+++ b/docs/core/docker/build-container.md
@@ -173,7 +173,7 @@ The *Dockerfile* file is used by the `docker build` command to create a containe
 
 In your terminal, navigate up a directory to the working folder you created at the start. Create a file named *Dockerfile* in your working folder and open it in a text editor. Add the following command as the first line of the file:
 
-```dockerfile
+```Dockerfile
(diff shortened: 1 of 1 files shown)


Youssef1313:
@mairaw, Is it the correct usage to be capitalized ?
If so, then it should be updated in https://github.com/dotnet/docs/pull/14202/ ?

mairaw:
I'll have to check with engineering on this one as well. I'm not seeing the dockerfile identifier do anything special. Even when using dockerfile, the identifier here should be uppercase if the identifier was working as expected:
image

Youssef1313:
> the identifier here should be uppercase if the identifier was working as expected

It seems the identifier works for both uppercase and lowercase, but as most of the usages of it is uppercase, I agree to keep it uppercase and will update it in my PR for template.md

mairaw:
Closing this one. The engineering team has added the appropriate capitalization on the code block label now:
https://docs.microsoft.com/en-us/dotnet/core/docker/build-container#create-the-dockerfile

Thanks for bringing this to our attention @Auios!

PR #14502, opened 2019-09-19
Rename jpeg to png

ardalis:
Fixing broken image by saving it as PNG.


mairaw (approved):
LGTM. I think the problem is that it expected a jpg extension not jpeg:
https://github.com/dotnet/docs/blob/master/docfx.json#L76

PR #14569, opened 2019-09-21
Increase heading level

drewnoakes:
I don't think _Nullable annotation context_ was supposed to have a lower heading level than _Nullable warning context_. They seem like siblings, and it's useful to have them both in the TOC.


Youssef1313:
I think the correct change is to make `Nullable warning context` lower level, as both `Nullable warning context` and `Nullable annotation context` are under `Nullable context`

BillWagner (approved):
Thanks @drewnoakes These two headings should be at the same level.

I'll :shipit: now.

PR #14570, opened 2019-09-21
QueryDuplicates2 need to use LastWriteTime

VarunSaiTeja:
In QueryDuplicates2 method it using file property Creation time of file.

If same file copied in two directory's then the file creation time will be definitely different, so the file will not be treated as duplicate file.
I am sure no file having same creation time because it checking file copied or created time with milliseconds.

So if we use file property LastWriteTime (Last Modified Time) it will be same time for all same files if they were not modified.

## Summary

Describe your changes here.


BillWagner (approved):
Thanks @VarunSaiTeja 

PR #14044, opened 2019-08-27
[framework] Remove redundant path segments (part 12)

nemrism:
Files in this PR:
- /framework/data/wcf/*.md (x72)
- /framework/unmanaged-api/alink/*.md (x22)
- /framework/unmanaged-api/tlbexp/*.md

Contributes to #10653

PR #14575, opened 2019-09-22
Remove redundant path segments

Youssef1313:
These redundant path segments were already removed by @nemrism in #14044, but were back by mistake in #14214

Contributes to #10653


Thraka (approved):
Looks good. Thanks!!

PR #14601, opened 2019-09-23
Remove extraspaces and redundant path segments

Youssef1313:
Contributes to: #10653


mairaw (approved):
That looks good. Thanks @Youssef1313!

PR #14603, opened 2019-09-23
Adding ML.NET Series link

luisquintanilla:
Adding ML.NET Series link


natke (changes requested):
This is going to conflict with #14488. We should add it to the new landing page. Let's liaise with @cjgronlund

PR #14643, opened 2019-09-25
Add missing dev_langs

Youssef1313:
/cc @Thraka


Thraka (approved):
Thanks! I got a bit of cleanup to do here 😄

PR #14646, opened 2019-09-25
Add (non-null object ) & (null) arms to switch expression

amrbadawy:
Add (non-null object ) & (null) arms to switch expression

## Summary

Describe your changes here.


BillWagner (approved):
Thanks for making this addition @amrbadawy 
It does make sense to add these arms in the first version

PR #14637, opened 2019-09-25
Removed preview-related note

pkulikov:
I believe, with the release of .NET Core 3.0, the preview-related note can be removed. If not yet, please close this PR.

PR #14647, opened 2019-09-25
Remove preview-related note

Youssef1313:
This was missed in #14637


docs/csharp/tutorials/nullable-reference-types.md
```diff
@@ -38,9 +38,6 @@ Create a new console application either in Visual Studio or from the command lin
 <Nullable>enable</Nullable>
 ```
 
-> [!IMPORTANT]
-> The `Nullable` element was previously named `NullableContextOptions`. The rename ships with Visual Studio 2019, 16.2-p1. The .NET Core SDK 3.0.100-preview5-011568 does not have this change. If you are using the .NET Core CLI, you'll need to use `NullableContextOptions` until the next preview is available.
-
 > [!NOTE]
 > When C# 8 is released (not in preview mode), the `Nullable` element will be added by new project templates. Until then, you'll need to add it manually.
```
pkulikov: And what about this one? Can it go as well?

Youssef1313: I think yes, C# is already released and not in preview mode. I'm going to remove that note too.

BillWagner (approved):
Thanks for making these updates @Youssef1313 

Also thanks for the reviews @pkulikov Sharp 👀 again.

PR #14657, opened 2019-09-25
Add identifier to template.md

Youssef1313:
I noticed that `vbscript` is a valid identifier and is used [here](https://docs.microsoft.com/en-us/dotnet/framework/wcf/samples/using-the-wcf-moniker-with-com-clients), so added it to `template.md`.

### DIFF
--- a/styleguide/template.md
+++ b/styleguide/template.md
@@ -319,8 +319,8 @@ Use three backticks (\`\`\`) + a language ID to apply language-specific color co
 |SQL|sql|
 |Swift|swift|
 |TypeScript|typescript|
-|VB|vb|
-|vbscript|vbscript|
+|Visual Basic|vb|
+|VBScript|vbscript|
 |XAML|xaml|
 |XML|xml|
 
--- a/styleguide/template.md
+++ b/styleguide/template.md
@@ -320,6 +320,7 @@ Use three backticks (\`\`\`) + a language ID to apply language-specific color co
 |Swift|swift|
 |TypeScript|typescript|
 |VB|vb|
+|vbscript|vbscript|
 |XAML|xaml|
 |XML|xml|
 


styleguide/template.md
```diff
@@ -320,6 +320,7 @@ Use three backticks (\`\`\`) + a language ID to apply language-specific color co
 |Swift|swift|
 |TypeScript|typescript|
 |VB|vb|
```
rpetrusha:
```suggestion
|Visual Basic|vb|
```

styleguide/template.md
```diff
@@ -320,6 +320,7 @@ Use three backticks (\`\`\`) + a language ID to apply language-specific color co
...
+|vbscript|vbscript|
```
rpetrusha:
```suggestion
|VBScript|vbscript|
```

rpetrusha (approved):
Thanks for adding VBScript to the list of supported languages, @Youssef1313. I've suggested two changes.

styleguide/template.md
Youssef1313: This column says what would show in the live docs, the visual basic code in the live docs appears as `VB`, so I think it should be kept as `VB` not `Visual Basic`.

image

styleguide/template.md
Youssef1313: Same comment:

image

mairaw:
I've asked the engineering team to appropriately capitalize VBScript on docs too.

mairaw (changes requested):
Marking as request changes so Ron's comments can be addressed.

Youssef1313:
> 
> 
> Marking as request changes so Ron's comments can be addressed.

This is for both changes ?
- `VB` => `Visual Basic`
- `vbscript` => `VBScript`

rpetrusha:
I wouldn't interpret the left column to be the rendering the language name in the documentation as much as the language name itself. In the case of VBScript, @mairaw has opened a bug to have the rendering changed. In the case of Visual Basic versus VB, Visual Basic is a registered trademark and so should be used to identify the Visual Basic language here.

mairaw (approved):
LGTM

PR #14660, opened 2019-09-25
Remove duplicated dev_langs

Youssef1313:
The duplication was produced in #13839, I didn't check if the duplication was done to other files in the same PR.


mairaw (approved):
Thank you @Youssef1313!

PR #14661, opened 2019-09-25
Removed dev_langs

Youssef1313:
The article has only code written in vb.


mairaw (changes requested):
I'd actually keep the vb one. There's a search metadata that is set to C# by default.

mairaw (approved):
Thanks for making the change!

PR #14690, opened 2019-09-26
add missing toc item

mingwli:
## Summary
1. proposal -> proposals for matching patterns with other toc names
2. add `Index and Range Changes` to `C# 8.0 specification proposals`

image


BillWagner:
I'm going to close this @mingwli 

This information does need to get published, and I've opened dotnet/csharplang#2797 to incorporate this information into the original proposal. That should get merged later this week, or earlly next week. 

Then, the OPS build warning will go away because `index-range-changes.md` will be removed.

/cc @jaredpar

Issue #2363, opened 2017-06-07
Provide consistent handling of C# and VB compiler error/warning topics


rpetrusha:
# Provide consistent handling of C# and VB compiler error/warning topics

## General
Currently, there are several issues with C# and Visual Basic compiler error and warning message topics:

1. The topics are arbitrarily divided into two different directories. For C#, they are in a [language reference](https://github.com/dotnet/docs/tree/master/docs/csharp/language-reference/compiler-messages) subdirectory and a [miscellaneous](https://github.com/dotnet/docs/tree/master/docs/csharp/misc) directory. For Visual Basic, they are also in a [language reference subdirectory](https://github.com/dotnet/docs/tree/master/docs/visual-basic/language-reference/error-messages) and a [miscellaneous](https://github.com/dotnet/docs/tree/master/docs/visual-basic/misc) directory.
2. The TOC lists only the items in the language reference subdirectory. This means that, although all of the topics are published, they are not all easily accessible.
3. Most of the filenames of the Visual Basic error message topics consist of some variation of the error message text. This results in long filenames (and long TOC entries) that are difficult to parse and identify.
 
To address these issues, we should do the following:
1. Move all of a language's error message topics to a single directory, preferably the language reference subdirectory for [C#](https://github.com/dotnet/docs/tree/master/docs/csharp/language-reference/compiler-messages) and [Visual Basic](https://github.com/dotnet/docs/tree/master/docs/visual-basic/language-reference/error-messages).
2. Determine whether we need to define redirects for the moved files. (Possibly we won't need to, since the files in the miscellaneous directories are not easily discoverable.)
3. Make all compiler warning/error topics discoverable. Since there are too many of them for the TOC (there are currently 928 C# topics and 1,654 Visual Basic topics), index.md should list each of the compiler error/warning numbers.
4. Improve the discoverability of Visual Basic error topics in both GitHub and on docs.microsoft.com by renaming the files to have only the Visual Basic error/warning number rather than the message string. This also requires that we create redirects for the 203 existing files in the language reference directory.
 
//cc @billwagner @mairaw

BillWagner:
I'll add one more recommendation:

I don't think we have all the warning and errors messages documented. I believe some of the most recent additions are not included.

PR #14691, opened 2019-09-26
Update sorry-we-don-t-have-specifics-on-this-csharp-error.md

Youssef1313:
Removed **some** error codes from keywords that already has documentation. There are more need to be removed that I'll do in (a) later PR(s).


rpetrusha:
Thanks for submitting this PR, @Youssef1313. AFAIK, this file was originally intended to be used by Visual Studio to provide a default help page when information about a compiler error was not available. 

@mariaw and @BillWagner, do you know whether this file is still used, and if it is, whether help looks to the *misc* directory, which is the location of all of these "sorry we don't have specifics" compiler errors that actually are documented? As far as I can see, the current behavior of help in Visual Studio is to open the browser and search for the compiler error, with StackOverflow given priority. If that's the case, we can approve this PR, or possibly, if this file is no longer used, we can remove it.

There is, incidentally, an outstanding issue (#2363) about the organization of the compiler error message documentation.

BillWagner:
@rpetrusha Yes, this file is still used I can land here from Visual Studio from some of the newest warnings.

I don't know if that system looks at the misc directory as well.

MSDN-WhiteKnight:
If this file stays, it needs to be revised. "Visual C# Language" and "Visual C# IDE" MSDN forums were archived and these links now lead to MSDN forums homepage. "Visual C# General" forum is now called just "Visual C#". And Stack Overflow is QA site, not forum.

BillWagner (approved):
Since all these are covered in another folder, these items can be removed from this article.

I'll :shipit: now.

PR #14692, opened 2019-09-26
Update cs0161.md

Youssef1313:
- Markdown improvements.
- Fix indentation.
- Changed variable value to make the condition goes to the branch that HAS a return value. The reason is to make the reader understand that even if the method will always return a value by logic but a code path doesn't have a return statement, the error will be generated.


BillWagner (approved):
Thanks for making these changes @Youssef1313 

Thanks again.

PR #14703, opened 2019-09-26
Added detail to VB extensions article

rpetrusha:

### DIFF
--- a/docs/visual-basic/programming-guide/language-features/procedures/how-to-write-an-extension-method.md
+++ b/docs/visual-basic/programming-guide/language-features/procedures/how-to-write-an-extension-method.md
@@ -21,13 +21,13 @@ Extension methods enable you to add methods to an existing class. The extension 
     Imports System.Runtime.CompilerServices
     ```
 
-3. Within a module in your new or existing application, begin the method definition with the [Extension](xref:System.Runtime.CompilerServices.ExtensionAttribute) attribute:
+3. Within a module in your new or existing application, begin the method definition with the [`<Extension>`](xref:System.Runtime.CompilerServices.ExtensionAttribute) attribute:
 
     ```vb
     <Extension()>
     ```
  
-   Note that the `Extension` attribute can only be applied to a Visual Basic [Module](../../../language-reference/statements/module-statement.md). If you apply it to a `Class` or a `Structure`, the Visual Basic compiler generates error [BC36551](../../../misc/bc36551.md), "Extension methods can be defined only in modules."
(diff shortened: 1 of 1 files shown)


mairaw (approved):
LGTM

docs/visual-basic/programming-guide/language-features/procedures/how-to-write-an-extension-method.md
```diff
@@ -21,11 +21,13 @@ Extension methods enable you to add methods to an existing class. The extension
...
     ```
 
-3. Within a module in your new or existing application, begin the method definition with the extension attribute:
+3. Within a module in your new or existing application, begin the method definition with the [Extension](xref:System.Runtime.CompilerServices.ExtensionAttribute) attribute:
 
     ```vb
     <Extension()>
     ```
+ 
+   Note that the `Extension` attribute can only be applied to a Visual Basic [Module](../../../language-reference/statements/module-statement.md). If you apply it to a `Class` or a `Structure`, the Visual Basic compiler generates error [BC36551](../../../misc/bc36551.md), "Extension methods can be defined only in modules."
```
Youssef1313: I think you're mixing BC36550 with BC36551. The text written describes BC36550

docs/visual-basic/programming-guide/language-features/procedures/how-to-write-an-extension-method.md
```diff
@@ -21,11 +21,13 @@ Extension methods enable you to add methods to an existing class. The extension
     Imports System.Runtime.CompilerServices
     ```
 
-3. Within a module in your new or existing application, begin the method definition with the extension attribute:
+3. Within a module in your new or existing application, begin the method definition with the [Extension](xref:System.Runtime.CompilerServices.ExtensionAttribute) attribute:
```
Youssef1313: I'm not sure if xref links are translated when docs not in en-us or not. If it will be translated, I think it should be put between back ticks.

docs/visual-basic/programming-guide/language-features/procedures/how-to-write-an-extension-method.md
rpetrusha: Thanks for catching this, @Youssef1313. Actually, it's imprecise wording on my part. BC36550 results when the attribute is applied to a class. What I'd meant to say is that the attribute can only be applied to a method in a module, and not to a method in a class (which produces BC36550) or a structure (which produces BC30662). So the compiler error was correct, but the description wrong.

PR #14706, opened 2019-09-26
Remove orphaned clause (typographic error)

reisenberger:
## Summary

Remove orphaned text - is a duplication of text which appears in the paragraph above.


mairaw (approved):
Thanks @reisenberger!

PR #14737, opened 2019-09-27
Update some errors that are not used by roslyn

Youssef1313:
I'm not sure if the note addition is correct or these pages should be just deleted ?

* See [ErrorCode.cs Line 50](https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/Errors/ErrorCode.cs#L50), [ErrorCode.cs Line 1142](https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/Errors/ErrorCode.cs#L1142) and https://github.com/dotnet/roslyn/pull/22400

There are much more pages than I've changed that require an attention.

### DIFF
--- a/docs/csharp/misc/cs0043.md
+++ b/docs/csharp/misc/cs0043.md
@@ -1,13 +1,17 @@
 ---
 title: "Compiler Error CS0043"
 ms.date: 07/20/2015
-f1_keywords: 
+f1_keywords:
   - "CS0043"
-helpviewer_keywords: 
+helpviewer_keywords:
   - "CS0043"
 ms.assetid: 471e2cf8-4275-4618-ad20-408f829d2119
 ---
 # Compiler Error CS0043
-PDB file 'file' has an incorrect or out-of-date format. Delete it and rebuild.  
-  
+
+PDB file 'file' has an incorrect or out-of-date format. Delete it and rebuild.
+
  Delete the .pdb file for this compilation and recompile.
+
+> [!NOTE]
+> This compiler error is no longer used in Roslyn.
--- a/docs/csharp/misc/cs0459.md
+++ b/docs/csharp/misc/cs0459.md
@@ -1,40 +1,45 @@
 ---
 title: "Compiler Error CS0459"
 ms.date: 07/20/2015
-f1_keywords: 
+f1_keywords:
   - "CS0459"
-helpviewer_keywords: 
+helpviewer_keywords:
   - "CS0459"
 ms.assetid: 01b058dd-8d65-4e9d-9de1-d47f9488d22a
 ---
 # Compiler Error CS0459
-Cannot take the address of a read-only local variable  
-  
- There are three common scenarios in the C# language that generate read-only local variables: `foreach`, `using`, and `fixed`. In each of these cases, you are not allowed to write to the read-only local variable, or to take its address. This error is generated when the compiler realizes you are trying to take the address of a read-only local variable.  
-  
-## Example  
- The following example generates CS0459 when an attempt is made to take the address of a read-only local variable in a `foreach` loop and in a `fixed` statement block.  
-  
-```csharp  
-// CS0459.cs  
-// compile with: /unsafe  
-  
-class A  
-{  
-    public unsafe void M1()  
-    {  
-        int[] ints = new int[] { 1, 2, 3 };  
-        foreach (int i in ints)  
-        {  
-            int *j = &i;  // CS0459  
-        }  
-  
-        fixed (int *i = &_i)  
-        {  
-            int **j = &i;  // CS0459  
-        }  
-    }  
-  
-    private int _i = 0;  
-}  
+
+Cannot take the address of a read-only local variable
+
+ There are three common scenarios in the C# language that generate read-only local variables: `foreach`, `using`, and `fixed`. In each of these cases, you are not allowed to write to the read-only local variable, or to take its address. This error is generated when the compiler realizes you are trying to take the address of a read-only local variable.
+
+## Example
+
+ The following example generates CS0459 when an attempt is made to take the address of a read-only local variable in a `foreach` loop and in a `fixed` statement block:
+
+```csharp
+// CS0459.cs
+// compile with: /unsafe
+
+class Program
+{
+    public unsafe void M1()
+    {
+        int[] ints = new int[] { 1, 2, 3 };
+        foreach (int i in ints)
+        {
+            int *j = &i;  // CS0459
+        }
+
+        fixed (int *i = &_i)
+        {
+            int **j = &i;  // CS0459
+        }
+    }
+
+    private int _i = 0;
+}
 ```
+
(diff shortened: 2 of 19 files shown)


docs/csharp/misc/cs0459.md
```diff
@@ -1,40 +1,45 @@
...
+            int **j = &i;  // CS0459
+        }
+    }
+
+    private int _i = 0;
+}
 ```
+
+> [!NOTE]
+> Roslyn compiler was updated and this compiler error was removed starting with Visual Studio 2017 15.5, so the previous code would compile successfully with this version and later.
```
pkulikov: What if someone uses Visual Studio Code? In other words, is it possible to indicate the C# version, starting with which the example compiles?

docs/csharp/misc/cs0043.md
```diff
@@ -1,13 +1,17 @@
...
 # Compiler Error CS0043
-PDB file 'file' has an incorrect or out-of-date format. Delete it and rebuild.  
-  
+
+PDB file 'file' has an incorrect or out-of-date format. Delete it and rebuild.
+
  Delete the .pdb file for this compilation and recompile.
+
+> [!NOTE]
+> This compiler error is no longer used in Roslyn.
```
pkulikov: Nice to have (don't know if it's worth the effort): indicate the C# version, starting with which the error is not generated anymore.

docs/csharp/misc/cs0459.md
Youssef1313: I really was thinking on that while writing it, but have no idea.

docs/csharp/misc/cs0043.md
Youssef1313: I think it's not used in Roslyn from the beginning. It was used by the old compiler before Roslyn.

Youssef1313: https://github.com/dotnet/roslyn/releases

@pkulikov , Seems like Roslyn versioning is independent of C# language versioning, but it's being pushed in parallel with visual studio.

docs/csharp/misc/cs0459.md
pkulikov: It looks that Visual Studio 2017 15.5 was shipped with C# 7.2. However, if I use one of the latest .NET Core SDKs and set the language version to 7.0 or 6.0, the code still compiles. So, indeed, it's more about the Roslyn version and not the C# version.

docs/csharp/misc/cs0043.md
pkulikov: @Youssef1313 thanks, indeed, it looks that's more about Roslyn versions rather than the C# versions.

docs/csharp/misc/cs0459.md
svick: Based on [this list](https://github.com/dotnet/roslyn/wiki/NuGet-packages#versioning), VS 15.5 corresponds to Roslyn 2.6.  Though I think most people are not aware of the version of Roslyn they use, so that information is probably not useful.

Based on [this page](https://dotnet.microsoft.com/download/dotnet-core/2.0), C# 7.2/VS 15.5 corresponds to .Net Core SDK 2.1.100. Maybe that's a version that would be worth mentioning?

Youssef1313: @svick, The second page says that VS 15.5 supports C# 7.2, this doesn't mean that the error will be generated before C# 7.2, because I can be using VS 15.5 with some older C# version.

Not sure if Roslyn 2.6 should be mentioned or not.

BillWagner:
@Youssef1313 @pkulikov @svick 

I appreciate the work done here, but the best course of action may be to close this PR.  

These pages are typically visited by developers whose code exhibits this warning or error. For these pages, they'd be using the older compiler. These are also very low traffic pages. In the case of those warnings that aren't generated by the current compiler, the page views are in the 10s of pages per month (vs. 10s of thousands for other content).

(In time, these pages might be removed, but I want to check with the product team first).

Thoughts?

Youssef1313:
@BillWagner, A developer who doesn't have the error may come across a page of these and try to generate the error himself. so I think the minimal action is to include a note to the pages having examples. but I think it's better to take action for all whether it has example or not.

jaredpar:
> These pages are typically visited by developers whose code exhibits this warning or error. For these pages, they'd be using the older compiler

While uses of the older, native, compiler are dwindling there are still large customer basis that use it. I know of several large internal teams which still consume it and there are still many legacy ASP.NET web applications out there which use it for on demand compilation. As long as those customers exists, which is likely quite a long time, and the cost of maintaining these pages is low I would error on just keeping them as is.

That being said I don't have an intuition as to what it costs to keep these pages alive. I also don't have a good count for how many customers still use the native compiler but it's still enough that I have to answer a question from them a couple of times a month.

BillWagner:
Thanks for the comments @Youssef1313 @jaredpar 

Let's continue to address this, and merge this PR.  I have one global suggestion, because "roslyn" is an overloaded term.  Instead of:

> This compiler error is no longer used in Roslyn.

Let's change to:

> This compiler error is no longer used as of C# 6.0.

In those examples where you say "In the roslyn compiler, the code generates...." replace that with "Starting with C# 6.0, the code generates...."

Youssef1313:
@BillWagner, I'm not sure. but I think I can with old C# versions with roslyn. so I can for example use C# 7.0 (which is after C# 6.0), or use C# 5.0 (which is before C# 6.0) with Roslyn.

BillWagner:
@Youssef1313 You are correct that you can specify an older language version and still use the roslyn compiler. Even so, I think saying the language version is a reasonable proxy for when this message will and won't be generated.

Youssef1313:
> 
> 
> @Youssef1313 You are correct that you can specify an older language version and still use the roslyn compiler. Even so, I think saying the language version is a reasonable proxy for when this message will and won't be generated.

Hmmm... I don't really think it's a good idea to specify a C# language version for when the message will/won't be generated. Because the appearance of the message doesn't depend on the C# version, but depends on a specific Roslyn version.

svick:
Could different phrasing help?

> This compiler error is no longer used as of C# 6.0.

This is inaccurate, because you will not get the error if you compile using Roslyn with language version set to C# 5.0.

> This compiler error is no longer used as of the C# 6.0 compiler.

This is accurate. Roslyn is still "the C# 6.0 compiler", even if you set its language version to some older value.

Though the difference is fairly subtle, so this phrasing might still be confusing. But maybe it's the least bad option?

BillWagner:
I like @svick's suggested wording above. Let's go with that.

Youssef1313:
> 
> 
> I like @svick's suggested wording above. Let's go with that.

Some of the errors aren't used in Roslyn at all, and some were removed in a specific Roslyn version, so I think it's still inaccurate. Besides that, it could be confusing.

BillWagner (approved):
We've discussed this enough.

I'll :shipit: now.

PR #14738, opened 2019-09-27
Update cs0041.md

Youssef1313:
The old content of the article was about `CS0811`, not `CS0041`.

The following two files from roslyn repo are much helpful to determine the error codes and the corresponding error message:

https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/CSharpResources.resx
https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/Errors/ErrorCode.cs#L50


BillWagner (approved):
Thanks for updating this @Youssef1313 

I'll :shipit: now.

PR #14748, opened 2019-09-27
Update bc32051.md

Youssef1313:
I've improved the markdown and the code sample. However, I think the article content needs to be updated. The provided code sample does generate a compile error different from the `bc32051`


rpetrusha:
Thanks for noticing the disparity between the intended and actual compiler errors, @Youssef1313. I've opened #14765 to address it.

rpetrusha (approved):
Thanks for making these improvements. This document is in need of some major investigation and revision, though, so I've opened #14765. In the meantime, I'll merge your PR.

PR #14750, opened 2019-09-27
Update nullability warning description

gewarren:
Not 100% sure, but the current text seems to contradict itself.


BillWagner (approved):
Thanks for making this fix @gewarren 

I'll :shipit: now.

PR #14783, opened 2019-09-28
Fix localization issue

Youssef1313:
`"The student cannot be found."` this shouldn't be translated, so I put it between back ticks.

Another problem that I don't know how to solve it is the images appear in the English version only.

Screenshots:

English:
image

French:
image

### DIFF
--- a/docs/standard/exceptions/how-to-create-localized-exception-messages.md
+++ b/docs/standard/exceptions/how-to-create-localized-exception-messages.md
@@ -77,12 +77,16 @@ The problem with the previous line is that `"The student cannot be found."` is j
 To create the localized exception messages:
 
 1. Create a new folder named *Resources* to hold the resource files.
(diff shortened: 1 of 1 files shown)


mairaw (approved):
Thank you @Youssef1313. I made a change that I believe it would fix the localization problem.

@JasonCard can you check if these changes we made will make the images show up in the localized version? Also, can we find other articles that might have similar issues? Thanks.

JasonCard:
@mairaw Thank you for the heads up. Currently waiting for the fix to be merged live, but looks like this fix should work. Our loc testing would catch any broken images, or if same number of images are not displayed between English and loc, so we can assume there are no other similar issues.

JasonCard:
@mairaw @Youssef1313 Confirmed images are now correctly displayed for French:
[https://docs.microsoft.com/fr-fr/dotnet/standard/exceptions/how-to-create-localized-exception-messages](https://docs.microsoft.com/fr-fr/dotnet/standard/exceptions/how-to-create-localized-exception-messages)
All languages should be fixed by next week.

PR #14829, opened 2019-09-30
Replace SQL tokens

nxtn:


Youssef1313:
@NextTurn, All these files in `includes` directory should be replaced **ONLY** when its content is just a small word. but it should remain when it contains some bigger statements, is that right ?

NextTurn:
@Youssef1313 You're almost right. There's another exception that we don't want to replace the token if the replacement touches an especially large number of files.

rpetrusha:
That's somewhat true, @Youssef1313,  although some context is useful. Our previous authoring systems used tokens to substitute for product names. They are useful when a product name isn't finalized, or when the same string can be used for both preview and production releases (so that only the token text needs to change, rather than each occurrence of the string in the documentation). Unfortunately, though, they were overused, so they were applied to static technology or product names, and in some cases a single occurrence of a product or technology was represented by a token. Overuse of these tokens makes the documentation difficult to read and to edit. So @NextTurn has been submitting PRs to remove unnecessary tokens.

rpetrusha (approved):
Thanks, @NextTurn, for eliminating two more tokens! I'll merge your PR now.

PR #14868, opened 2019-10-01
Replace wrong with proper quotes in string.Join()

ognyandim:
## Summary
The example does not compile.

With single quotes, the example will not compile since single quotes are for char and not for string as the method [string.Join()](https://docs.microsoft.com/en-us/dotnet/api/system.string.join?view=netframework-4.8) requires in its documentation.

### DIFF
--- a/docs/csharp/whats-new/csharp-8.md
+++ b/docs/csharp/whats-new/csharp-8.md
@@ -455,7 +455,7 @@ numbers ??= new List<int>();
 numbers.Add(i ??= 17);
 numbers.Add(i ??= 20);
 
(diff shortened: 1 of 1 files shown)


pkulikov:
The example compiles. Though, only in NET Core, which has overloads with `char`:
https://docs.microsoft.com/en-us/dotnet/api/system.string.join?view=netcore-3.0#System_String_Join__1_System_Char_System_Collections_Generic_IEnumerable___0__

BillWagner (approved):
Thanks for making this update @ognyandim 

PR #14878, opened 2019-10-01
Replace two tokens

Youssef1313:


docs/framework/wcf/wcf-system-requirements.md
```diff
@@ -13,14 +13,14 @@ Windows Communication Foundation (WCF) is a communication infrastructure that is
 
 WCF is installed by default on [!INCLUDE[wv](../../../includes/wv-md.md)].
 
-WCF can also be installed on [!INCLUDE[wxpsp2](../../../includes/wxpsp2-md.md)], [!INCLUDE[ws2003r2](../../../includes/ws2003r2-md.md)], or [!INCLUDE[ws2003sp1](../../../includes/ws2003sp1-md.md)].
+WCF can also be installed on [!INCLUDE[wxpsp2](../../../includes/wxpsp2-md.md)], Windows Server 2003 R2, or [!INCLUDE[ws2003sp1](../../../includes/ws2003sp1-md.md)].
```
rpetrusha: This page is so dated, @Youssef1313, that the best thing to do is to delete it as well. That also requires redirecting requests for this page to the .NET Framework system requirements page at <https://docs.microsoft.com/en-us/dotnet/framework/get-started/system-requirements> and removing its entry from the TOC.

rpetrusha (changes requested):
Thanks for removing these two tokens, @Youssef1313. I also think that the article can also be deleted; I've added more detail in my comment.

rpetrusha:
@Youssef1313, I'll approve your PR so that the tokens are deleted, then open a separate PR to remove the topic and handle the redirection.

rpetrusha (approved):

PR #14918, opened 2019-10-03
Use PascalCase for methods + minor changes

Youssef1313:


mairaw (approved):
Thanks you @Youssef1313 for making these improvements. I've made a small change to fix the code block and it should be good to merge after it builds.

PR #14949, opened 2019-10-04
Add "Size" column in floating-point types table

Youssef1313:
image


BillWagner (approved):
Thanks for making this update @Youssef1313 

It is good information to add to this table.  I'll :shipit: now.

PR #15009, opened 2019-10-07
Update fundamentals.md

Crabzmatic:
## Summary

Some image descriptions are inline with image and they don't look good in browser. They were removed and tooltips for images were added instead, as those images already had a suitable description above them. Alt texts were left unchanged.


Thraka (approved):
Thanks a bunch for cleaning that up! It looks a lot better now.

PR #15066, opened 2019-10-09
Remove duplicated node in file name

tdykstra:
Related to #15062


mairaw (approved):
LGTM. One more question related to this, do we need to review the incoming links to see if they were intended to go to the binary or the xml ones or not needed?

tdykstra:
>do we need to review the incoming links to see if they were intended to go to the binary or the xml ones or not needed?

I didn't change any of the individual sample docs, so I don't think there's any need to review incoming links to them.

PR #15083, opened 2019-10-10
Rename default interface members -> default interface methods

### LINKED ISSUE #15072: Default interface methods, not default interface members
Some of our documents use "default interface members" instead of "default interface methods".

All occurrences of "members" in that phrase should be "methods".

Potapy4:
Rename members -> methods

## Summary

Fixes #15072

### DIFF
--- a/docs/csharp/tutorials/default-interface-methods-versions.md
+++ b/docs/csharp/tutorials/default-interface-methods-versions.md
@@ -1,10 +1,10 @@
 ---
-title: Safely update interfaces using default interface members in C#
+title: Safely update interfaces using default interface methods in C#
 description: This advanced tutorial explores how you can safely add new capabilities to existing interface definitions without breaking all classes and structs that implement that interface.
 ms.date: 05/06/2019
 ms.custom: mvc
 ---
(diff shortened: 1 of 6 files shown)


docs/csharp/tutorials/default-interface-methods-versions.md
```diff
@@ -1,10 +1,10 @@
 ---
-title: Safely update interfaces using default interface members in C#
```
pkulikov: The old filename should be redirected to the new one.

BillWagner: The redirections are configured in the `openpublishing.redirection.json` in the root of this repository.  We've sorted the entries by source, so this one should go [here](https://github.com/dotnet/docs/blob/master/.openpublishing.redirection.json#L1369)

BillWagner:
I should add that I made a quick review, and the redirection was the only concern I had as well.

Once that's done, I'll :shipit: it.

BillWagner (approved):
Thanks for taking this on and fixing this issue @Potapy4 

I've reviewed everything, and this is ready to :shipit:

PR #15084, opened 2019-10-10
Fix Microsoft.AspNetCore.App link

lkstz:
## Summary

In the _Metapackages_ section of the [_Packages, metapackages and frameworks_](https://docs.microsoft.com/en-us/dotnet/core/packages) page, the link to details on the `Microsoft.AspNetCore.App` metapackage incidentally points to `Microsoft.AspNetCore.All`. 

This PR fixes the link target to point to the correct location.


mairaw (approved):
Great catch @kolletzki. Thanks for fixing the link. Your contribution should be live within a day.

PR #15113, opened 2019-10-11
System.Text.Json - add scenarios and add converters doc

### LINKED ISSUE #14444: Add scenarios to the System.Text.Json how-to doc
Related to #12207

## From comments on the #14208 PR

* Add an example about converting enums to their label names (rather than numbers) using JsonStringEnumConverter. https://docs.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonstringenumconverter?view=netcore-3.0
* Add an example where a default value is not overridden by null on deserialize when IgnoreNullValues is set to true.
* Add a sample of using ArrayBufferWriter<byte> as the output-sink since that helps avoid the extra encoding/allocation when trying to write UTF-8 text to a stream (say FileStream). https://github.com/dotnet/docs/pull/14208/files#r324126902
* Add JsonWriter samples https://github.com/dotnet/docs/pull/14208/files#r324127488
  * Write really large JSON file - say 3 GB.
  * Write asynchronously with periodic flushing.
  * Customize the escaping/encoder behavior when writing non-ascii text.
  *  [advanced/lower priority] Writing to a System.IO.Pipeline.PipeWriter
  * For more info, review https://gist.github.com/ahsonkhan/f6f30656717548212693e5eaa49cece5
* Add JsonReader samples https://github.com/dotnet/docs/pull/14208/files#r324129542
  * Reading really large JSON file - say 3 GB.
  * Reading asynchronously (using the state for continuation).
  * [advanced/lower priority] Reading from a System.IO.Pipeline.PipeReader
  * The custom options (comment handling - skipping/allow, trailing commas, max depth)
  * For more info, review
    * https://github.com/dotnet/apireviews/blob/master/2018/System.Text.Json/Additional.Notes.md
    * https://gist.github.com/ahsonkhan/f6f30656717548212693e5eaa49cece5
* Add a section on JsonDocument/JsonElement between the serializer/deserializer and the reader/writer.
  * Samples worth looking through and adding (particularly scenario 4/5 that cover JsonDocument/Element): https://github.com/ahsonkhan/JsonUserStudy/tree/master/UxScenarios

## From the doc planning meeting

* Write to and read from a file, synchronous and async
* Write to and read from a Stream, synchronous and async
* Read from a Stream or file with multiple objects
* Change the encoding, for example, to allow a Unicode language character set. 

---
#### Document Details

* Content: [How to serialize JSON - .NET](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to)
* Product: **dotnet**
* GitHub Login: @tdykstra

### LINKED ISSUE #14445: Create doc on converters for System.Text.Json
Related to #12207

tdykstra:
Internal review URLs
* [How to serialize/deserialize](https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to?branch=pr-en-us-15113)
* [How to write custom converters](https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?branch=pr-en-us-15113)

Content added to the existing how-to doc;
* Serialize to a file, sync or async
* Deserialize from a file, sync or async
* Convert enums to strings
* Customize character encoding
  * Allow language character sets
  * Allow specific characters
  * Unsafe encoder @GrabYourPitchforks  - please check the caution note wording on this
* Ignore null when deserializing
* Intro to JsonDocument, reader, and writer
* Use JsonDocument to access data
* Use JsonDocument to write JSON
* Filter data using Utf8JsonReader

Fixes #14444 -- the scenarios list has been edited to exclude the ones not added in this PR.

Fixes #14445-- new converters doc

Note -- This PR still uses inline code for snippets; the next PR will convert them to references to snippets in the samples repo.

### DIFF
--- a/docs/standard/serialization/system-text-json-converters-how-to.md
+++ b/docs/standard/serialization/system-text-json-converters-how-to.md
@@ -213,18 +213,22 @@ The following sections provide converter samples that address some common scenar
 
 When deserializing to a property of type `Object`, a `JsonElement` object is created. The reason is that the deserializer doesn't know what CLR type to create, and it doesn't try to guess. For example, if a JSON element has "true", the deserializer doesn't infer that the value is a `Boolean`, and if an element has "01/01/2019", the deserializer doesn't infer that it's a `DateTime`. You can implement type inference in a custom converter.
 
-The following code shows a converter that converts `true` or `false` values in JSON to a Boolean value in an Object property. For other values, the code falls back to `JsonElement`.
+The following code shows a converter that converts `true` or `false` values in JSON to Boolean in an Object property, numbers to `long`, dates to `DateTime`, and strings to `string`. For other values, the code falls back to `JsonElement`.
 
 ```csharp
 using System;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
-namespace SystemTextJsonConverters
+namespace SystemTextJsonSamples
 {
-    public class ObjectToBoolConverter : JsonConverter<object>
+    public class ConverterInferredTypesToObject
+        : JsonConverter<object>
     {
-        public override object Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        public override object Read(
+            ref Utf8JsonReader reader,
+            Type typeToConvert,
+            JsonSerializerOptions options)
         {
             if (reader.TokenType == JsonTokenType.True)
             {
@@ -236,18 +240,40 @@ namespace SystemTextJsonConverters
                 return false;
             }
 
-            // Use JsonElement as fallback.
-            var converter = options.GetConverter(typeof(JsonElement)) as JsonConverter<JsonElement>;
-            if (converter != null)
+            if (reader.TokenType == JsonTokenType.Number)
             {
-                return converter.Read(ref reader, typeToConvert, options);
+                if (reader.TryGetInt64(out long l))
+                {
+                    return l;
+                }
+
+                return reader.GetDouble();
             }
-            throw new JsonException();
+
+            if (reader.TokenType == JsonTokenType.String)
+            {
+                if (reader.TryGetDateTime(out DateTime datetime))
+                {
+                    return datetime;
+                }
+
+                return reader.GetString();
+            }
+
+            // Use JsonElement as fallback.
+            // Newtonsoft uses JArray or JObject.
+            using (JsonDocument document = JsonDocument.ParseValue(ref reader))
+            {
+                return document.RootElement.Clone();
+            }
         }
 
-        public override void Write(Utf8JsonWriter writer, object value, JsonSerializerOptions options)
+        public override void Write(
+            Utf8JsonWriter writer,
+            object value,
+            JsonSerializerOptions options)
         {
-            throw new InvalidOperationException("Directly writing object not supported");
+            throw new InvalidOperationException("Should not get here.");
         }
     }
 }
@@ -257,39 +283,37 @@ The following code registers the converter:
 
 ```csharp
 options = new JsonSerializerOptions();
-options.Converters.Add(new ObjectToBoolConverter());
+options.Converters.Add(new ConverterInferredTypesToObject());
 ```
 
 Here's an example type with an Object property:
 
 ```csharp
-class WeatherForecast
+public class WeatherForecastWithObjectProperties
 {
-    public DateTimeOffset Date { get; set; }
-    public int TemperatureC { get; set; }
-    public string Summary { get; set; }
-    public object Rain { get; set; }
+    public object Date { get; set; }
+    public object TemperatureC { get; set; }
+    public object Summary { get; set; }
 }
 ```
 
-The following example of JSON to deserialize contains a Boolean value for the `Rain` property:
+The following example of JSON to deserialize contains values that will be deserialized as `DateTime`, `long`, and `string`:
 
 ```json
 {
   "Date": "2019-08-01T00:00:00-07:00",
   "TemperatureC": 25,
-  "Summary": "Hot and windy",
-  "Rain": true
+  "Summary": "Hot",
 }
 ```
 
-Without the custom converter, deserialization puts a `JsonElement` in the `Rain` property. With the custom converter the `Rain` property gets a Boolean value.
+Without the custom converter, deserialization puts a `JsonElement` in each property.
 
-The [unit tests folder](https://github.com/dotnet/corefx/blob/master/src/System.Text.Json/tests/Serialization/) in the `System.Text.Json.Serialization` has more examples of custom converters that handle deserialization to Object properties.
+The [unit tests folder](https://github.com/dotnet/corefx/blob/master/src/System.Text.Json/tests/Serialization/) in the `System.Text.Json.Serialization` namespace has more examples of custom converters that handle deserialization to Object properties.
 
 ### Support Dictionary with non-string key
 
-The built-in support for dictionary collections is for `Dictionary<string, TValue>`. That is, the key must be a string. If you need to support a dictionary with an integer or some other type as the key, a custom converter is required.
+The built-in support for dictionary collections is for `Dictionary<string, TValue>`. That is, the key must be a string. To support a dictionary with an integer or some other type as the key, a custom converter is required.
 
 The following code shows a custom converter that works with `Dictionary<int,string>`:
 
@@ -382,7 +406,7 @@ The JSON output from serialization looks like the following example:
 }
 ```
 
-The [unit tests folder](https://github.com/dotnet/corefx/blob/master/src/System.Text.Json/tests/Serialization/) in the `System.Text.Json.Serialization` has more examples of custom converters that handle non-string-key dictionaries.
+The [unit tests folder](https://github.com/dotnet/corefx/blob/master/src/System.Text.Json/tests/Serialization/) in the `System.Text.Json.Serialization` namespace has more examples of custom converters that handle non-string-key dictionaries.
 
 ### Support polymorphic deserialization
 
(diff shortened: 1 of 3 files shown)


BillWagner (approved):
Overall, this is great work @tdykstra I learned a lot.

I had a few questions and comments, which you can decide how you want to handle.

Then, :shipit: when ready.

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -114,26 +129,24 @@ The following example shows the same JSON, formatted (that is, pretty-printed wi
...
 ## Serialization behavior
 
-* By default, all public properties are serialized. You can [specify properties to exclude](#exclude-properties).
+* By default, all public properties are serialized. You can [specify properties to exclude](#exclude-properties-from-serialization).
 * The [default encoder](xref:System.Text.Encodings.Web.JavaScriptEncoder.Default) escapes non-ASCII characters, HTML-sensitive characters within the ASCII-range, and characters that must be escaped according to [the JSON spec](https://tools.ietf.org/html/rfc8259#section-7).
 * By default, JSON is minified. You can [pretty-print the JSON](#serialize-to-formatted-json).
-* By default, casing of JSON names matches the .NET names. You can [customize JSON name casing](#customize-json-names).
+* By default, casing of JSON names matches the .NET names. You can [customize JSON name casing](#customize-json-names-and-values).
 * Circular references are detected and exceptions thrown. For more information, see the [issue on circular references](https://github.com/dotnet/corefx/issues/38579) in the dotnet/corefx repository on GitHub.
 * Currently, fields are excluded.
```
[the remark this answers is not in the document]
tdykstra: It can't be configured in this first release, but you could handle fields in a custom converter. There's some discussion in this issue:  https://github.com/dotnet/corefx/issues/36505

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -412,7 +410,63 @@ Here's an example object to serialize and JSON output:
 
 The camel case naming policy applies to serialization only.
```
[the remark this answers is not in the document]
tdykstra: > This seems to imply that if I deserialize a dictionary, the keys match the JSON file, not the selected policy. I think that would be surprising. 

That is exactly what happens.  I'll add a note to that effect.

docs/standard/serialization/system-text-json-converters-how-to.md
```diff
@@ -0,0 +1,745 @@
...
+```
+
+## Samples for common scenarios
+
+The following sections provide converter samples that address some common scenarios that built-in functionality doesn't handle.
+
+### Deserialize inferred types to Object properties
+
+When deserializing to a property of type `Object`, a `JsonElement` object is created. The reason is that the deserializer doesn't know what CLR type to create, and it doesn't try to guess. For example, if a JSON element has "true", the deserializer doesn't infer that the value is a `Boolean`, and if an element has "01/01/2019", the deserializer doesn't infer that it's a `DateTime`. You can implement type inference in a custom converter.
+
```
steveharter: The above example about "true" and "datetime" are correct however a more ambiguous example is with a JSON number. If we try to parse as a long (no decimal points) then that may have range issues if the value was originally serialized as a ulong or BigInteger. If we try to parse as a double that will lose precision if the number was originally serialized as a decimal.

docs/standard/serialization/system-text-json-converters-how-to.md
```diff
@@ -0,0 +1,745 @@
...
+The following code shows a converter that converts `true` or `false` values in JSON to Boolean in an Object property, numbers to `long`, dates to `DateTime`, and strings to `string`. For other values, the code falls back to `JsonElement`.
+
+```csharp
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace SystemTextJsonSamples
+{
+    public class ConverterInferredTypesToObject
```
steveharter: The name seems a bit off. "ObjectToInferredTypesConverter"?

docs/standard/serialization/system-text-json-converters-how-to.md
```diff
@@ -0,0 +1,745 @@
...
+
+```json
+{
+  "Date": "2019-08-01T00:00:00-07:00",
+  "TemperatureC": 25,
+  "Summary": "Hot",
+}
+```
+
+Without the custom converter, deserialization puts a `JsonElement` in each property.
```
steveharter: However the TemperatureC property after deserialize will be a string (not the Temperature type).

docs/standard/serialization/system-text-json-converters-how-to.md
```diff
@@ -0,0 +1,745 @@
...
+    "Hot": 40
+  }
+}
+```
+
+The [unit tests folder](https://github.com/dotnet/corefx/blob/master/src/System.Text.Json/tests/Serialization/) in the `System.Text.Json.Serialization` namespace has more examples of custom converters that handle non-string-key dictionaries.
+
+### Support polymorphic deserialization
+
+[Polymorphic serialization](system-text-json-how-to.md#serialize-properties-of-derived-classes) is supported, but a custom converter is required for deserialization.
```
steveharter: Polymorphic serialization is not truly supported by default -- it is only supported if the property's type is System.Object. This is for secure-by-default and because we'd have no way to deserialize \ round-trip the type since the type information is not specified in the JSON in some way.

However we want to have an opt-in option for 5.0 to enable polymorphic serialization for any type but also want to ensure we have a pattern or support for polymorphic deserialization (such as including a type discriminator as metadata in the JSON in a secure way).

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -506,40 +560,107 @@ Here's an example object to serialize and JSON output:
...
+{
+    Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping
+};
+json = JsonSerializer.Serialize(weatherForecast, options);
+```
+
+> [!CAUTION]
+> Unlike the default encoder, the `UnsafeRelaxedJsonEscaping` encoder:
+>
+> * Doesn't escape HTML-sensitive characters such as `<`, `>`, and `&`.
```
GrabYourPitchforks: I'd probably change the bullet points to read as follows:

 * Doesn't escape HTML-sensitive characters such as `<`, `>`, `&`, and `'`.
 * Doesn't offer any additional defense-in-depth protections against XSS or information disclosure attacks, such as those which might result from the client and server disagreeing on the _charset_.
 * Is more permissive than the default encoder on which characters may pass through unescaped.

Use the unsafe encoder only when it is known that the client will be interpreting the resulting payload as UTF-8 encoded JSON, such as if it is known that the server is sending the response header `Content-Type: application/json; charset=utf-8`. Never allow the raw `UnsafeRelaxedJsonEscaping` output to be emitted into an HTML page or a `<script>` element.

GrabYourPitchforks (commented):
Left some feedback re: `UnsafeRelaxedJsonEscaping`.

docs/standard/serialization/system-text-json-converters-how-to.md
```diff
@@ -0,0 +1,745 @@
...
+  - "serialization"
+  - "objects, serializing"
+  - "converters"
+---
+
+# How to write custom converters for JSON serialization in .NET
+
+This article shows how to create custom converters for the JSON serialization classes that are provided in the <xref:System.Text.Json> namespace. For an introduction to `System.Text.Json`, see [How to serialize and deserialize JSON in .NET](system-text-json-how-to.md).
+
+A *converter* is a class that converts an object or a value to and from JSON. The `System.Text.Json` namespace has built-in converters for most primitive types that map to JavaScript primitives. You can write custom converters:
```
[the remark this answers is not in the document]
terrajobst: > Does "built-in" imply internal? 

In .NET Core, I'd treat our code base as "open", i.e. people should feel comfortable studying or copy & pasting from it. To me, "built-in" mostly means "automatically referenced/enabled".

docs/standard/serialization/system-text-json-converters-how-to.md
```diff
@@ -0,0 +1,755 @@
...
+
+* To override the default behavior of a built-in converter. For example, you might want `DateTime` values to be represented by mm/dd/yyyy format instead of the default  ISO 8601-1:2019 format.
+* To support a custom data type. For example, a `PhoneNumber` struct.
```
terrajobst: I believe we generally refer to `struct ` as value types in the documentation, no?

docs/standard/serialization/system-text-json-converters-how-to.md
```diff
@@ -0,0 +1,755 @@
...
+
+```text
+Unhandled exception. System.Text.Json.JsonException: 
+The JSON value could not be converted to System.Object. 
+Path: $.Date | LineNumber: 1 | BytePositionInLine: 37.
+```
+
+If you do provide a message (for example, `throw new JsonException("Error occurred)`, the exception still provides the path in the <xref:System.Text.Json.JsonException.Path> property.
+
+### Sample basic converter
```
terrajobst: It seems we should lead with this and then explain the factory pattern. It feels strange to "force" the customer to read a lot of conceptual text before they get to some sample code.

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -750,8 +1129,111 @@ while (reader.Read())
...
+    "alpha_two_code": "US",
+    "state-province": null,
+    "country": "United States",
+    "domains": [ "fabrikamuniversity.edu" ],
+    "name": "Fabrikam University"
+  }
+]
+```
+
 ## Additional resources
```
terrajobst: I love the links to GitHub in this section! We should do this for more stuff.

terrajobst (approved):
Left some comments, but looks great overall!

PR #15128, opened 2019-10-12
Nullable references: add link to the ! operator and minor edits

pkulikov:
- Reference the `!` operator article from the "Nullable reference types" article. Also, move the link to the spec draft to the `!` operator article.
- In the "Nullable reference types" article, do not use code formatting for "enabled" and "disabled" adjectives when talking about contexts. That way, those words would be localized and less confused with the values to specify in the code, which are `enable` and `disable`.
- Fix typos.


BillWagner (approved):

I'll :shipit: now.

PR #15134, opened 2019-10-12
Fix typo in "interfaced"

batkaevruslan:
Fixes "interfaced" to "interfaces"


mairaw (approved):
Good catch. Thanks for the fix @batkaevruslan!

PR #15139, opened 2019-10-12
Replace URI tokens

nxtn:


Thraka (approved):
Looks good. I added a minor change directly to your branch that spaced one of the tokens better. The error was in the original document.

PR #15140, opened 2019-10-12
Added the missing link to the Accessibility section

TimoSalomaki:
## Summary

As far as I understand, the Accessibility section should be completely removed from this list or linked to the closest match that we have in the .NET documentation. UI-Automation section includes accessibility best practices etc so I think it's suitable here.


mairaw (approved):
Thank you @HankiDesign. That's the right link to add.

PR #15148, opened 2019-10-13
Correcting a typo in CONTRIBUTING.md

MrSarno:
"for a variety of reason" → "for a variety of reasons"

If there is a 'variety' there must be more than one, hence changing the noun from singular to plural.


mairaw (approved):
Thank you @MrSarno for fixing this typo.

PR #15150, opened 2019-10-13
Nullable attributes: correct descriptions of NotNullWhen and MaybeNullWhen

pkulikov:
Both [NotNullWhen](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullwhenattribute?view=netcore-3.0) and [MaybeNullWhen](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.maybenullwhenattribute?view=netcore-3.0) attributes can be applied to any input argument, not only `out` or `ref`.

For the example, see the mentioned `string.IsNullOrEmpty` method:
https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/String.cs#L432-L441

Also, one cannot specify any condition for the method return value, but only to indicate if that return value is either `true` or `false`.


BillWagner (approved):

That's an important update. I've reviewed the changes, and I'll :shipit: now.

PR #15163, opened 2019-10-14
Update sentiment-analysis.md

LocalJoost:
_modelPath does not occur in the sample so it should not be mentioned in the text

Removed that from the text and changed the line above so it reflects only one field being added

Preview link: https://review.docs.microsoft.com/en-us/dotnet/machine-learning/tutorials/sentiment-analysis?branch=pr-en-us-15163


Youssef1313 (commented):
By removing `_modelPath`, the bullet list will now have one element only. I think making it just one statement is better than having the bullet list.

natke (approved):
Thank you for @LocalJoost! We appreciate your contribution.

PR #15171, opened 2019-10-14
C# operators: mention null-coalescing assignment

pkulikov:


BillWagner (commented):
This should be simple, but reading this posed a question in my brain.

This now mentions both `??` and `??=`. But, it doesn't mention any other compound assignment operators (like `+=` or `*=`). I would expect text to explain both, I think.

Thoughts?

BillWagner (approved):
I really like the final version @pkulikov 

I'll :shipit: now.

Thanks again!

PR #15188, opened 2019-10-15
Update virtual.md

MSDN-WhiteKnight:
## Summary

"Virtual properties behave like abstract methods" looks like an error, they are much different. Probably meant to be "Virtual properties behave like **virtual** methods"


BillWagner (approved):
This is a great change @MSDN-WhiteKnight 

Thanks for spotting and fixing this.

PR #15203, opened 2019-10-15
Update string-functions.md

Thraka:
@mairaw This seemed to work for me


mairaw (approved):
So odd. Yesterday it took two hours to build and it seemed broken. Thanks for checking again!

PR #15207, opened 2019-10-16
Misc updates to tasks.md

stevenjpr:
## Summary

- The "dog or fruit." binary classification example doesn't seem to fit with the rest of the examples. "dog or fruit" doesn't feel binary.
- "You can train a binary classification model using the following algorithms:". Suggest providing some guidance on where a developer should start when picking one of these algorithms.  After reading some more, I think this is covered in how-to-choose-an-ml-net-algorithm.md so maybe just provide a link?  I added this to the intro section.  Right spot?
-I was tripped up by the "one vs all upgrades" phrasing on line 69 several times.  "vs" is "versus" in this case, right.  But what is "one" and what is "all".  Are these properties I set in on the binary classification objects somewhere?
- Consider adding "PCA" to the glossary and provide a link (154)?  The rest of this page does a great job of linking terms such as this.  Maybe this one just got missed?
-I liked that the other sections gave good examples of scenarios for each of these algorithms.  This section is missing those scenarios.  Maybe they are obvious to more experienced readers.

### DIFF
--- a/docs/machine-learning/resources/tasks.md
+++ b/docs/machine-learning/resources/tasks.md
@@ -7,7 +7,7 @@ author: natke
 ---
 # Machine learning tasks in ML.NET
 
-When building a machine learning model, you first need to define what you are hoping to achieve with your data. This allows you to choose the right machine learning task for your situation. The following list describes the different machine learning tasks that you can choose from and some common use cases.
+When building a machine learning model, you first need to define what you are hoping to achieve with your data. This allows you to choose the right machine learning task for your situation. The following list describes the different machine learning tasks that you can choose from and some common use cases.  See [Algorithms](how-to-choose-an-ml-net-algorithm.md) for more information about choosing the task that is appropriate for your scenario.
 
 Once you have decided which task works for your scenario, then you need to choose the best algorithm to train your model. The available algorithms are listed in the section for each task.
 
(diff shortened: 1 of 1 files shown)


docs/machine-learning/resources/tasks.md
```diff
@@ -7,7 +7,7 @@ author: natke
 ---
 # Machine learning tasks in ML.NET
 
-When building a machine learning model, you first need to define what you are hoping to achieve with your data. This allows you to choose the right machine learning task for your situation. The following list describes the different machine learning tasks that you can choose from and some common use cases.
+When building a machine learning model, you first need to define what you are hoping to achieve with your data. This allows you to choose the right machine learning task for your situation. The following list describes the different machine learning tasks that you can choose from and some common use cases.  See [Algorithms](how-to-choose-an-ml-net-algorithm.md) for more information about choosing the task that is appropriate for your scenario.
```
cjgronlund:
```suggestion
When building a machine learning model, you first need to define what you are hoping to achieve with your data. This allows you to choose the right machine learning task for your situation. The following list describes the different machine learning tasks that you can choose from and some common use cases.  See [Algorithms](../how-to-choose-an-ml-net-algorithm.md) for more information about choosing the task that is appropriate for your scenario.
```

cjgronlund: @stevenjpr : Hi, Steven, More good edits. The link didn't work, but if you can accept my change we're good to go. Thank you!

Youssef1313:
```suggestion
When building a machine learning model, you first need to define what you are hoping to achieve with your data. This allows you to choose the right machine learning task for your situation. The following list describes the different machine learning tasks that you can choose from and some common use cases. For more information about choosing the task that is appropriate for your scenario, see [Algorithms](../how-to-choose-an-ml-net-algorithm.md).
```

Youssef1313 (changes requested):
This suggestion fixes the link like @cjgronlund's suggestion. In addition to that, I've modified the phrase to be consistent with the same wording used all over docs.

cjgronlund (approved):
Good fixes. Thanks, Steven. 

@stevenjpr: I know dog and fruit don't feel like binary values, it's a choice between two things. binary classification is also called "two-class classification." It's not common that the two choices would be dog and fruit, so point taken!

PR #15208, opened 2019-10-16
Rework C# Guide TOCs

BillWagner:
A second prerequisite to hub and landing pages.

This PR combines many sections of the C# Guide TOC.

Almost all of the sub-TOCs in the C# guide have been combined.  There are three exceptions:
- **C# compiler warnings messages**:  This section has a very large number of files. It's not intended for browsing, but for search only.
- **C# Language Specification**: This is its own section so that readers can generate a PDF of the draft language specification (rather than the entire C# Guide, or language reference)
- **C# Language Proposals**: If the spec wasn't in the main TOC, it seemed really strange to have the current proposals in the main TOC.

[Internal review link](https://review.docs.microsoft.com/en-us/dotnet/csharp/?branch=pr-en-us-15208)


mairaw (approved):
Great work @BillWagner! I think customers will like this change.

PR #15220, opened 2019-10-16
Programming concepts unified

lucasnobredev:
## Summary

This PR came out from issue #4494.

In [How to: Determine if a file is an assembly](https://docs.microsoft.com/en-us/dotnet/standard/assembly/identify) at "See Also" section, there is an item called Programming Concepts (Visual Basic), but not for C#. The both pages has the same idea, so I unified. The children pages also had the same problem, but I believe we can change this in another PR.


BillWagner:
@lucasnobredev Thanks for doing this work.

First, there are some merge conflicts to resolve.

Second, this is taking a bit longer to review. I've started an internal discussion. I'm not sure the "framework" section is the right final location for this overview. There are good reasons to consider either the "standard" or the "core" sections.

/cc @mairaw

BillWagner:
@lucasnobredev 

We appreciate the work you have put into this. However, at this time, we're going to close this PR.  Moving this content to the "Framework" section is not a good location for the content to reside.

We are starting some internal discussions regarding how we'll reorganize the .NET content as we take the journey to .NET 5. There are a number of decisions to make regarding the conceptual content as we get started.

We appreciate all your contributions, and we want to respect your time on this project. As you start on larger projects, it will help if you comment first on the related issue so we can discuss a plan of action before you invest too much time. Some issues may not be ready for work yet, and this is one of them.

PR #15227, opened 2019-10-16
Update predict-prices-with-model-builder.md

stevenjpr:
## Summary

Comments:
- In the "Train the Model" section, sentence: "Best accuracy displays the accuracy of the best performing model found by Model Builder so far. Higher accuracy means the model predicted more correctly on test data."
  -- The label in the wizard is "Best Quality(R squared)".  I was expecting it to be "Best accuracy" based on the above sentence.  Maybe the text is just out of date with the wizard?

- In the "Evaluate the Model" section, sentence: "Additionally, a summary table containing top five models and their metrics."
 -- This sentence feels incomplete to me.  Maybe something like" "A summary table that contains the top five models and their metrics is also provided" ??


cjgronlund (approved):
Good catch! Thank you, @stevenjpr!

PR #15242, opened 2019-10-17
Update "using declarations" examples to show variable scope changes

smack0007:
I've noticed on a few podcasts that I listen to during the discussion about the new using declarations syntax there is never a mention about the scoping advantages of the using declarations syntax and I would like to update the examples a bit to show that.

## Summary

I changed the 2 examples in the "using declarations" section to show that a variable can be declared after the using statement and it is in scope.


BillWagner (approved):
Thanks for adding these clarifying remarks @smack0007 

That is an important facet of the new syntax.

I'll :shipit: now. 

PR #15246, opened 2019-10-17
fix C# Guide label name

BillWagner:
This is causing new C# Guide issues to be placed under a new label, instead of using the existing label.

@Thraka Once this is merged, I'll cleanup the few existing labels with the wrong name.


mairaw (approved):

Thraka:
Looks like the copy-paste monster got me!

PR #15223, opened 2019-10-16
Inconsistent switch usage for compiler options on links to those articles

lucasnobredev:
## Summary

This PR came out from this issue #15222 .

I found a lot of results with VSCode, so I open this first PR to a better review.

PR #15276, opened 2019-10-18
Update bc40059.md

Youssef1313:
I think a change of `/` to `-` was missed in #15223. I've changed it.
Also removed extraspaces, left blank lines around headers, and fixed header level from h3 to h2.


Thraka (approved):
Thanks!

PR #15298, opened 2019-10-19
Fix header levels

Youssef1313:
The first `Parts` header is related to the `If operator called with three arguments ` header and should be under it. Same with the second `Parts` header. I've also removed some extra spaces and fixed header words case.


Thraka (approved):
Thanks!

PR #15308, opened 2019-10-19
Update docs/standard/security/index.md

Youssef1313:
- Remove redundant path segments. (Contributes to #10653)
- Remove links from `In this section`. ( The links in the index.md should be a list of the toc. Some links weren't in the toc, so deleted them. Correct me if I'm wrong )


mairaw (approved):
Looks good! You're right, the In this section should match what's on the TOC. I've encountered a few places where this was outdated before as well.

PR #15312, opened 2019-10-20
Update arrays/index.md

Youssef1313:
- Added a bit more detail in the article intro.
- Made the syntax as code fence.
- Added level-2 `Example` header.
- Changed `## Array Overview` to `## Array overview` (same with other headers).
- Removed unnecessary white-spaces.

### DIFF
--- a/docs/csharp/programming-guide/arrays/index.md
+++ b/docs/csharp/programming-guide/arrays/index.md
@@ -2,51 +2,47 @@
 title: "Arrays - C# Programming Guide"
 ms.custom: seodec18
 ms.date: 07/20/2015
-helpviewer_keywords: 
+helpviewer_keywords:
   - "arrays [C#]"
   - "C# language, arrays"
 ms.assetid: bb79bdde-e570-4c30-adb0-1dd5759ae041
 ---
 # Arrays (C# Programming Guide)
 
-You can store multiple variables of the same type in an array data structure. You declare an array by specifying the type of its elements.  
-  
- `type[] arrayName;`  
-  
- The following example creates single-dimensional, multidimensional, and jagged arrays:  
-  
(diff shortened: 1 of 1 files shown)


docs/csharp/programming-guide/arrays/index.md
```diff
@@ -2,51 +2,47 @@
...
-  
-- [Using foreach with Arrays](./using-foreach-with-arrays.md)  
-  
-- [Passing Arrays as Arguments](./passing-arrays-as-arguments.md)  
-  
-## C# Language Specification
-
- [!INCLUDE[CSharplangspec](~/includes/csharplangspec-md.md)]  
-  
+You can store multiple variables of the same type in an array data structure. You declare an array by specifying the type of its elements. If you want the array to store elements of different types, you can specify `object` as its type. In the unified type system of C#, all types, predefined and user-defined, reference types and value types, inherit directly or indirectly from <xref:System.Object>.
```
BillWagner: Nit: I changed "different types" to "any type" because it would generally be best practice to name the most specific type possible for a use case. (for example a common base class for a specific use.) But, this article isn't a good place to go into a lot of detail on those concerns.

```suggestion
You can store multiple variables of the same type in an array data structure. You declare an array by specifying the type of its elements. If you want the array to store elements of any type, you can specify `object` as its type. In the unified type system of C#, all types, predefined and user-defined, reference types and value types, inherit directly or indirectly from <xref:System.Object>.
```

BillWagner (changes requested):
I like the changes you've made @Youssef1313 

I had one small suggestion that makes the intent a bit more clear.

After that change, this is ready to merge.

BillWagner (approved):

PR #15315, opened 2019-10-20
Update cs1033.md as it's not used in Roslyn

Youssef1313:


BillWagner (approved):
Thanks for spotting this and fixing it.

I'll :shipit: now.

PR #15317, opened 2019-10-20
Update template.md with current docs style

Youssef1313:
I've used this regex to find headers ends in a `#`:

```regex
(\r\n|\r|\n)\#(.+)\#(\r\n|\r|\n)
```

**All** of the results found were escaping the `#`, not adding an extra one.


mairaw (approved):
Thank you @Youssef1313, you're right. We've been escaping those.

PR #15318, opened 2019-10-20
Refactor refout/refonly documentation

MSDN-WhiteKnight:
## Summary

- Because there's now an article about reference assemblies, no need to repeat these information in multiple places. Suggested to replace with a short note and see also link.
- Fix other minor error (`-refout` used in place of `-refonly`)

Related to #2638


BillWagner (approved):
Thanks for helping us consolidate this information @MSDN-WhiteKnight 

PR #15324, opened 2019-10-20
Removed Parenthetical Reference

### LINKED ISSUE #9288: Grammar Issue
The text advises, "Use past tense (as in this example) when the event reports something that has happened."  However, the example referred to, "Progress," is not a past tense form of "to progress." (It is present indefinite, IIRC.)  This could be improved by changing the name to a past tense verb, such as "MadeProgress" or "Progressed", or by removing the parenthetical reference to the example.  (The latter approach still leaves an example that conflicts with the advice given, so it's not ideal.)

---
#### Document Details

* ID: 0e2507bc-e567-23f9-aeb9-f7529baa7032
* Version Independent ID: c42718d0-15b4-a87f-4353-e15656686889
* Content: [Introduction to Events](https://docs.microsoft.com/en-us/dotnet/csharp/events-overview)
* Content Source: [docs/csharp/events-overview.md](https://github.com/dotnet/docs/blob/master/docs/csharp/events-overview.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

WretchedDade:
## Removed parenthetical reference from past tense event naming suggesting

Made change suggested by @BillWagner and @rpetrusha in the issue:

> Adding the "up for grabs" label. I think the correct fix is in the previous comment:
> 
> I do think, though, that the text becomes less confusing by remove the parenthetical reference to the example.

Fixes #9288

### DIFF
--- a/docs/csharp/events-overview.md
+++ b/docs/csharp/events-overview.md
@@ -65,7 +65,7 @@ The type of the event (`EventHandler<FileListArgs>` in this example) must be a
 delegate type. There are a number of conventions that you should follow
 when declaring an event. Typically, the event delegate type has a void return.
 Event declarations should be a verb, or a verb phrase.
-Use past tense (as in this example) when
+Use past tense when
 the event reports something that has happened. Use a present tense verb (for
 example, `Closing`) to report something that is about to happen. Often, using
 present tense indicates that your class supports some kind of customization


BillWagner (approved):
Thanks again @WretchedDade 

I'll :shipit: now.

PR #15320, opened 2019-10-20
C# reference: consolidated the numeric conversions articles

pkulikov:
While merging both implicit and explicit numeric conversions tables, I've moved all the `char`-related conversions to the article about the `char` keyword. With that, I've also cleaned up that article a bit.

PR #15326, opened 2019-10-21
Markdown fixes

Youssef1313:
- Fixed an [MD012](https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md014---dollar-signs-used-before-commands-without-showing-output) issue that showed in #15320.
- Used *italic* for folders.


mairaw (approved):
Looks good. Thanks @Youssef1313!

PR #15342, opened 2019-10-22
Fix lowercase at start of sentence

jiimaho:


mairaw (approved):
Thank you @jiimaho for improving our content.

PR #15350, opened 2019-10-22
Fix wrong period

Youssef1313:
The period in that place seems strange and incorrect. I suspect it's a comma.


BillWagner (approved):
Thanks for making these changes @Youssef1313 I've reviewed everything, and this LGTM.

I'll :shipit: now.

PR #15356, opened 2019-10-22
Replace redirected file link with new target

gewarren:
- replace link to data binding overview topic to avoid redirect hop
- remove trailing whitespace (diff with ignore WS changes)


gewarren:
Will fix this link as part of whole cloth effort to replace redirects.

PR #15358, opened 2019-10-22
Move XML dynamic properties files from VS repo

gewarren:
Moves WPF data binding with LINQ to XML files from VS repo to .NET repo under framework/wpf/data. Other than the sample, these are purely reference topics that aren't specific to Visual Studio. They belong here with the other WPF data binding topics.

Also adds a landing page for framework/wpf/data, makes the first link to Data binding overview contextual, sentence-cases the TOC, and deletes a top-level fluff topic.

Once this is merged, I'll delete [this section](https://docs.microsoft.com/en-us/visualstudio/designers/wpf-data-binding-with-linq-to-xml-overview?view=vs-2019) from the VS repo and redirect here.


docs/framework/wpf/data/data-binding-how-to-topics.md
```diff
@@ -1,63 +0,0 @@
----
```
Thraka: I actually refer to this article from the desktop guide. As the desktop guide isn't completed, there is a lot of information here with specific tutorials. I think we should keep this until I get the majority ported.

gewarren:
@Thraka I added the file back when you are ready to review this again. Thanks!

Thraka:
Thinking about it more....

While I do **love** these changes, especially the landing page, it makes it feel like we're actively working in this area, which we're not. We're working in the new desktop guide for wpf. This info is effectively archived as we migrate it into the new area piece-by-piece. Can we just move this linq-xml data binding stuff, add a toc link so it's nicely hooked up to the TOC, but that's all?

gewarren:
@Thraka I reverted to the old landing page. Thanks for reviewing!

Thraka (approved):
Awesome!

PR #15388, opened 2019-10-25
Update index.md

glennferrie:
Just switching 'can' to 'must' in the opening statement.  It implies that it is an imperative to implement all of the methods.  It also is representative of what is documented here: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/interface

## Summary

same explanation as above.  simply changing the wording. 

### DIFF
--- a/docs/csharp/programming-guide/interfaces/index.md
+++ b/docs/csharp/programming-guide/interfaces/index.md
@@ -9,7 +9,7 @@ ms.assetid: 2feda177-ce11-432d-81b4-d50f5f35fd37
 ---
 # Interfaces (C# Programming Guide)
 
-An interface contains definitions for a group of related functionalities that a [class](../../language-reference/keywords/class.md) or a [struct](../../language-reference/keywords/struct.md) can implement.
+An interface contains definitions for a group of related functionalities that a [class](../../language-reference/keywords/class.md) or a [struct](../../language-reference/keywords/struct.md) must implement.
   
 By using interfaces, you can, for example, include behavior from multiple sources in a class. That capability is important in C# because the language doesn't support multiple inheritance of classes. In addition, you must use an interface if you want to simulate inheritance for structs, because they can't actually inherit from another struct or class.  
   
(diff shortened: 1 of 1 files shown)


adisakshya (commented):
You can always partially implement an interface using abstract class, or throw ```NotImplementedException``` for the methods that you don't want to implement.

So I believe it is good to say that ```a class can implement the functionalities``` rather than saying ```a class must implement the functionalities```.

docs/csharp/programming-guide/interfaces/index.md
```diff
@@ -9,7 +9,7 @@ ms.assetid: 2feda177-ce11-432d-81b4-d50f5f35fd37
 ---
 # Interfaces (C# Programming Guide)
 
-An interface contains definitions for a group of related functionalities that a [class](../../language-reference/keywords/class.md) or a [struct](../../language-reference/keywords/struct.md) can implement.
+An interface contains definitions for a group of related functionalities that a [class](../../language-reference/keywords/class.md) or a [struct](../../language-reference/keywords/struct.md) must implement.
```
BillWagner: This wording gives a bit of freedom on the "must":

```suggestion
An interface contains definitions for a group of related functionalities that a non-abstract [class](../../language-reference/keywords/class.md) or a [struct](../../language-reference/keywords/struct.md) must implement.
```

BillWagner (changes requested):
Thanks for spotting this and suggesting better wording @glennferrie 

@adisakshya raises a valid point that abstract classes can declare support for an interface without defining those methods. 

See my suggestion to wiggle through those distinctions, which are covered more thoroughly in this section.

After accepting that suggestion, this is ready to merge.

docs/csharp/programming-guide/interfaces/index.md
adisakshya: This seems more accurate.

BillWagner (approved):

PR #15395, opened 2019-10-25
Update how-to-determine-which-versions-are-installed.md

nnpcYvIVl:
## Summary

Add **Release** DWORD value for Windows 10 November 2019 Update.

### DIFF
--- a/docs/framework/migration-guide/how-to-determine-which-versions-are-installed.md
+++ b/docs/framework/migration-guide/how-to-determine-which-versions-are-installed.md
@@ -14,7 +14,7 @@ ms.author: "ronpet"
 ---
 # How to: Determine which .NET Framework versions are installed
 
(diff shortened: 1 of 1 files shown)


gewarren:
@vivmishra Can you verify if these are the correct DWORD values for .NET Framework 4.8?

On Windows 10 May 2019 Update and Windows 10 November 2019 Update: 528040
On all other Windows operating systems (including other Windows 10 operating systems): 528049

Also, should we just say "On Windows 10 May 2019 Update and later" to be future proof?

vivmishra:
If you have 528040 or higher then you have 4.8 (or higher) on the box. That should be sufficient minimum check for 4.8. 

- 528049 is the key on all other platforms other than Windows 10 May 2019 Update and November 2019 Update. 
- When new version of Win 10 goes out early next year then it will have a key value greater than 528049.

CC: @jamshedd

gewarren (approved):

PR #15410, opened 2019-10-26
Fix O(N2) → O(N²)

ForNeVeR:
## Summary

There's a small formatting issue in the document that is fixed by this PR.

### DIFF
--- a/docs/standard/asynchronous-programming-patterns/consuming-the-task-based-asynchronous-pattern.md
+++ b/docs/standard/asynchronous-programming-patterns/consuming-the-task-based-asynchronous-pattern.md
@@ -627,7 +627,7 @@ double currentPrice = await NeedOnlyOne(
 ```
 
 ### Interleaved Operations
- There is a potential performance problem with using the <xref:System.Threading.Tasks.Task.WhenAny%2A> method to support an interleaving scenario when you're working with very large sets of tasks.  Every call to <xref:System.Threading.Tasks.Task.WhenAny%2A> results in a continuation being registered with each task. For N number of tasks, this results in O(N2) continuations created over the lifetime of the interleaving operation.  If you're working with a large set of tasks, you can use a combinator  (`Interleaved` in the following example) to address the performance issue:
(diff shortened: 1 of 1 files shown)


docs/standard/asynchronous-programming-patterns/consuming-the-task-based-asynchronous-pattern.md
```diff
@@ -627,7 +627,7 @@ double currentPrice = await NeedOnlyOne(
 ```
 
 ### Interleaved Operations
- There is a potential performance problem with using the <xref:System.Threading.Tasks.Task.WhenAny%2A> method to support an interleaving scenario when you're working with very large sets of tasks.  Every call to <xref:System.Threading.Tasks.Task.WhenAny%2A> results in a continuation being registered with each task. For N number of tasks, this results in O(N2) continuations created over the lifetime of the interleaving operation.  If you're working with a large set of tasks, you can use a combinator  (`Interleaved` in the following example) to address the performance issue:
+ There is a potential performance problem with using the <xref:System.Threading.Tasks.Task.WhenAny%2A> method to support an interleaving scenario when you're working with very large sets of tasks.  Every call to <xref:System.Threading.Tasks.Task.WhenAny%2A> results in a continuation being registered with each task. For N number of tasks, this results in O(N²) continuations created over the lifetime of the interleaving operation.  If you're working with a large set of tasks, you can use a combinator  (`Interleaved` in the following example) to address the performance issue:
```
Youssef1313:
```suggestion
 There is a potential performance problem with using the <xref:System.Threading.Tasks.Task.WhenAny%2A> method to support an interleaving scenario when you're working with very large sets of tasks. Every call to <xref:System.Threading.Tasks.Task.WhenAny%2A> results in a continuation being registered with each task. For N number of tasks, this results in O(N²) continuations created over the lifetime of the interleaving operation. If you're working with a large set of tasks, you can use a combinator (`Interleaved` in the following example) to address the performance issue:
```

Youssef1313 (changes requested):
Replaced double spaces with a single space and used `<sup>2</sup>`.

ForNeVeR:
Thanks for your suggestion, I've searched through the docs and found that we, indeed, use `O<sup>2</sup>` here. I'll apply it.

BillWagner (approved):
thanks for making these changes @ForNeVeR  

This will produce a better published document. I'll :shipit: now.

PR #15418, opened 2019-10-26
Removed the duplicate list of the article sections

pkulikov:
That list is presented at the right of the article or before (in the mobile version).
Also corrected the expression that invokes a constructor.


BillWagner (approved):

I like the changes, and I'll :shipit: now.

PR #15421, opened 2019-10-27
Update how-to-customize-a-time-based-cache-policy.md

Youssef1313:
- Fixed header levels from H3 to H2, that was causing the `In this article` side bar to contain only a link to `See also`.
- Used var in C# code.
- Removed bullet lists that has only one item.
- Removed some extra spaces.
- Left blank lines around headers.
- Followed the current convention in headers casing.


BillWagner (approved):
This LGTM. I'll :shipit: now.

PR #15439, opened 2019-10-27
Remove self-redirecting link from see also

Youssef1313:


mairaw (approved):
Looks good. Thanks @Youssef1313!

PR #15440, opened 2019-10-27
Fix XML code

Youssef1313:
Contributes to #9744


gewarren (approved):
These all look good to me. Thanks @Youssef1313 .

PR #15471, opened 2019-10-28
Add protobuf to template.md

Youssef1313:
I've seen it [here](https://docs.microsoft.com/en-us/dotnet/architecture/grpc-for-wcf-developers/protobuf-maps) and it seems to be working!


mairaw (approved):
Thanks @Youssef1313. I've confirmed that it's present in the repo where we have the highlighting logic. Thanks!

PR #15493, opened 2019-10-29
Fix some issues with Mono's description

nelsonjchen:
Looks like a word was dropped and an extra `)` was added.


mairaw (approved):
Thank you for making these improvements @nelsonjchen. Your contribution will be live within a day.

PR #15501, opened 2019-10-29
Fix typo abstrat to abstract

angellmethod:
## Summary

Describe your changes here.

I've fixed a typo in the first line that had "abstrat" instead "abstract."


mairaw (approved):
Thank you for fixing that typo @angellmethod. Your contribution should be live within a day.

Issue #14676, opened 2019-09-25
Edit pass on Desktop Guide - WPF articles


Thraka:
The WPF articles that shipped for Core 3 should have an edit pass conducted on them. The majority of them are from the old WPF content that had acrolinx run across them, but more can be done to clear them up.

PR #15549, opened 2019-10-30
Update xaml-resources-define.md

Kellylorenebaker:
@Thraka 

Edits complete per https://mseng.visualstudio.com/TechnicalContent/_workitems/edit/1608924. Per the style guidelines for Azure, either "app" or "application" are acceptable, though "app" is preferred, so I'm changing these instances where it seems to make sense. Feel free to accept or reject as you see fit. 

Thanks,
Kelly

Contributes to #14676


Thraka (approved):
LGTM Thanks!

PR #15556, opened 2019-10-30
Replace site-relative links where possible

gewarren:
Contributes to #15560


BillWagner (approved):
These all LGTM @gewarren 

You can :shipit: when ready

PR #15566, opened 2019-10-30
Fix links in repo

gewarren:
- Replace links with redirect target
- Replace site-relative links with file-relative links


gewarren:
Let me know if this PR is too large and I should do it by subfolder instead.

gewarren:
Oops, this includes deleting files. I'll create a new PR.

PR #15598, opened 2019-10-31
Added DO link

creativefisher:
## Summary

Describe your changes here.


nishanil (commented):
We don't need to explicitly link providers as the Terraform home is already linked. Since this is part of the book, we limit the links to avoid distractions. We will close this PR. Sorry :(

Issue #5169, opened 2018-05-04
Index Initializers section is inaccurate


drewjcooper:
In the Index initializers section, the initial text about collection initializers seems to imply that there was no way to initialize a Dictionary<,> prior to the new C# 6.0 syntax.  This is incorrect.  The following code is semantically equivalent to the index initializer example in the document.

```csharp
private Dictionary<int, string> webErrors = new Dictionary<int, string>
{
    { 404, "Page not Found" },
    { 302, "Page moved, but left a forwarding address." },
    { 500, "The web server can't come out to play today." }
};
```
---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 36fef60d-04e9-dc86-2e3f-134c10ea37bb
* Version Independent ID: 09b5aa7a-1b43-8bf9-9786-f10f457488dd
* Content: [What's New in C# 6 - C# Guide](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-6)
* Content Source: [docs/csharp/whats-new/csharp-6.md](https://github.com/dotnet/docs/blob/master/docs/csharp/whats-new/csharp-6.md)
* Service: **unspecified**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

pkulikov:
@BillWagner if my understanding is correct, this issue was fixed by #5577

DevonAleshire:
Drew, you are awesome. Great explanation as well. Let me find some time to clarify it better since I started working down the issue. I will focus on making sure there is a distinction made between the fact that the previous method is a collection initializer using the IEnumerable implementation and the new index initializer utilizes the indexer support as an object initializer for index assignment during initialization.

DevonAleshire:
@pkulikov and @BillWagner I will try to simplify the details a little for incorporation into the documentation as noted above by @drewjcooper . I think it is an important concept to hit the wave tops of without getting overly complex. Let me know if issue #5169 is something you want to keep open a little longer, otherwise I will hold off. 

Also off topic, @BillWagner I did not even realize who you were until I was looking through my Safari books queue yesterday and saw the effective and more effective books ha ha. I was like, I have seen that name recently. Pleasure to assist.

pkulikov:
[...]

PR #15600, opened 2019-10-31
Fix historical version info

lbragaglia:
## Summary

Index Initializers has been introduced in C# 6.

See [What's new](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-6#initialize-associative-collections-using-indexers) section.

See also #5169.


Youssef1313:
I think the original text is correct. The code sample uses a feature new in C# 6.0.
See https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-6#initialize-associative-collections-using-indexers

BillWagner (approved):
Thanks for spotting this and fixing it @lbragaglia 

I've reviewed your changes, and I'll :shipit: now.

PR #15617, opened 2019-11-01
Delete dev_langs in consuming-library-with-visual-studio.md

Youssef1313:
The problem with the dev_langs in this article is that it adds language selector at the top of article.

image

This selector is useless, because the article itself has a C# and VB tabs.

The other solution would be trying to remove the tabs from the article and combining the information together and keep the dev_langs.


mairaw (approved):
This change makes sense. Thanks for fixing it @Youssef1313.

PR #15629, opened 2019-11-01
Expression body -> expression-bodied

Youssef1313:


Youssef1313:
Closing, I think the original text is correct.

PR #15650, opened 2019-11-01
Swap a Decrypt for an Encrypt

nickdevore:
The wrong word was used in the opening paragraph describing what this example reflects

## Summary

I simply replaced the word `decrypts` (the wrong word) with `encrypts` in the opening paragraph


gewarren (approved):
@nickdevore Thanks for fixing this copy/paste error.

PR #15661, opened 2019-11-02
Fix punctuation and a header casing

Youssef1313:


BillWagner (approved):
Thanks for making these updates @Youssef1313 
I'll :shipit: now.

PR #15673, opened 2019-11-03
Increase SEO for nullable article

BillWagner:
This article was not as prominient as it should have been for key terms related to nullable reference types and the attributes to describe those API contracts. Update the title, description, and opening paragraph to increase its findability.

/cc @jaredpar @cartermp

### DIFF
--- a/docs/csharp/nullable-attributes.md
+++ b/docs/csharp/nullable-attributes.md
@@ -1,12 +1,12 @@
 ---
-title: Upgrade APIs with attributes to define null expectations
-description: This article explains the motivations and techniques for adding descriptive attributes to describe the null state of arguments and return values from APIs
+title: Upgrade APIs for nullable reference types with attributes that define null expectations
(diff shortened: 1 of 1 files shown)


docs/csharp/nullable-attributes.md
```diff
@@ -1,12 +1,12 @@
 ---
-title: Upgrade APIs with attributes to define null expectations
-description: This article explains the motivations and techniques for adding descriptive attributes to describe the null state of arguments and return values from APIs
+title: Upgrade APIs for nullable reference types with attributes that define null expectations
```
cartermp:
```suggestion
title: Upgrade APIs for nullable reference types with attributes that define expectations for null values
```

The previous one read kinda weird to me, but I'm not particularly intent on my specific suggestion getting in

cartermp (approved):

jaredpar (approved):

PR #15677, opened 2019-11-04
[WIP] Add visual basic snippets for system.text.json docs

Youssef1313:
Marking as WIP for two reasons:

- The added vb code isn't tested.
- I haven't adding vb for all the snippets.

Before continuing working on this, please let me know if you indeed want to add vb code for this article.


tdykstra:
@Youssef1313 Thanks for the offer, but as noted in the warning at the top of the page, the documentation is still under construction -- it would be better to wait until it's a little more stable.  I have a PR in review now that makes major changes to this article, and after that I will be making one to convert the inline snippets to snippet references, creating a samples project in the samples repo for that.  When that samples project is created would be a good time to create a VB project and begin adding VB snippets.

PR #15684, opened 2019-11-04
Update collection-type-dependency-properties.md

### LINKED ISSUE #15660: Collection-Type Dependency Properties presents incorrect code
As mentioned in @oldnewthing's [article](https://devblogs.microsoft.com/oldnewthing/20191004-00/?p=102964), the page [Collection-Type Dependency Properties](https://docs.microsoft.com/en-us/dotnet/framework/wpf/advanced/collection-type-dependency-properties?source=docs) shows incorrect code first, then explaining why it's wrong and then explaining how to fix it:

>  However, if you just left the code as shown, that single list default value is shared for all instances of `Aquarium`. If you ran the following test code, which is intended to show how you would instantiate two separate `Aquarium` instances and add a single different `Fish` to each of them, you would see a surprising result: […]
>
>  Instead of each collection having a count of one, each collection has a count of two! This is because each `Aquarium` added its `Fish` to the default value collection, which resulted from a single constructor call in the metadata and is therefore shared between all instances. This situation is almost never what you want.  
>
> To correct this problem, you must reset the collection dependency property value to a unique instance, as part of the class constructor call. Because the property is a read-only dependency property, you use the `SetValue` method to set it, using the `DependencyPropertyKey` that is only accessible within the class.

This is very confusing, because it's quite easy to miss this explanation when you're looking for code to use in your program. I think the only full example code in this article should show the correct code, and ideally, the correct code should go first, with an explanation of the common mistake after it.

Thraka:
Fixes #15660


Thraka:
@svick I added a bit more on the introduction of paragraph above the code. I think this gets people the right idea that there is something wrong with this so they won't just continue without reading subsequent paragraph.

Thoughts?

docs/framework/wpf/advanced/collection-type-dependency-properties.md
```diff
@@ -22,7 +22,7 @@ This topic provides guidance and suggested patterns for how to implement a depen
...
+ Consider the following example. The following section of the example shows the definition for a class `Aquarium`, which contains a flaw with the default value. The class defines the collection type dependency property `AquariumObjects`, which uses the generic <xref:System.Collections.Generic.List%601> type with a <xref:System.Windows.FrameworkElement> type constraint. In the <xref:System.Windows.DependencyProperty.Register%28System.String%2CSystem.Type%2CSystem.Type%2CSystem.Windows.PropertyMetadata%29> call for the dependency property, the metadata establishes the default value to be a new generic <xref:System.Collections.Generic.List%601>.
```
gewarren: I think it would be even more obvious with a callout Note or Warning block, if you want to consider that.

```
> [!WARNING]
> etc.
```

gewarren (approved):

docs/framework/wpf/advanced/collection-type-dependency-properties.md
Thraka: Good idea

PR #15686, opened 2019-11-04
Add ReadOnlySpan and adjust slice call in both

cartermp:
without `e-s`, we're quite prone to IndexOutOfRange exceptions


BillWagner (approved):
This LGTM @cartermp 

You can :shipit: when ready.

PR #15687, opened 2019-11-04
Missing break!

mvelosop:
## Summary

Spotted while browsing fast! going back to visible line breaks 😊

### DIFF
--- a/docs/architecture/microservices/net-core-net-framework-containers/net-framework-container-scenarios.md
+++ b/docs/architecture/microservices/net-core-net-framework-containers/net-framework-container-scenarios.md
@@ -45,16 +45,16 @@ In the meantime, if any platform or service in Azure still doesn't support .NET 
 
 ### Additional resources
 
-- **.NET Core Guide**  
+- **.NET Core Guide** \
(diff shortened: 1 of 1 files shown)


Youssef1313:
From the [Contributor Guide](https://docs.microsoft.com/en-us/contribute/how-to-write-links#link-text).

> The words that you include in link text should be friendly. In other words, they should be normal English words or the title of the page that you're linking to.

There is a bad example in the contributor guide which is similar to the links in this file:

```markdown
For more details, see [https://msdn.microsoft.com/library/ms173763.aspx](https://msdn.microsoft.com/library/ms173763.aspx).
```

mvelosop:
Hi @Youssef1313, thanks for your feedback.

The links in the additional resources sections in this guide are all formatted this way, with a title separated from the actual address, because the intent is for the links to be useful and informative even when printed.

That's why the link text is the actual absolute link that will be resolved from the docs-relative .md file link.

nishanil (approved):

nishanil:
Approving this based on the changes that are in this PR which looks fine to me.

PR #15689, opened 2019-11-04
Fix loc issue

Youssef1313:
This is how the article **currently** show in the localized versions:

image

I think the changes I made would fix that.

### DIFF
--- a/docs/framework/configure-apps/file-schema/add-element-for-custom-2.md
+++ b/docs/framework/configure-apps/file-schema/add-element-for-custom-2.md
@@ -14,8 +14,8 @@ ms.author: "mairaw"
 
 Adds custom application settings. Each **\<add>** tag contains a key/value pair.
 
-[**\<configuration>**](configuration-element.md)   
-&nbsp;&nbsp;[**\<sectionName>**](custom-element-2.md)   
+[**\<configuration>**](configuration-element.md)\
+&nbsp;&nbsp;[**\<sectionName>**](custom-element-2.md)\
 &nbsp;&nbsp;&nbsp;&nbsp;**\<add>**
 
 ## Syntax
--- a/docs/framework/configure-apps/file-schema/appsettings/add-element-for-appsettings.md
+++ b/docs/framework/configure-apps/file-schema/appsettings/add-element-for-appsettings.md
@@ -14,8 +14,8 @@ ms.author: "mairaw"
 
 Adds a custom application setting.
 
-[**\<configuration>**](../configuration-element.md)   
-&nbsp;&nbsp;[**\<appSettings>**](appsettings-element-for-configuration.md)   
+[**\<configuration>**](../configuration-element.md)\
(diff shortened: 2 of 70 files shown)


Youssef1313:
This PR doesn't cover all the articles with this problem, but covered most of them.

mairaw:
@JasonCard after we merge this PR, can you make sure to follow up and see if the issue is addressed for loc? Thanks.

mairaw (approved):
Thanks @Youssef1313!

JasonCard:
@mairaw @Youssef1313 Thank you for working on this fix. Yes, I'll follow up to check if this issue is resolved.

Youssef1313:
@mairaw, seems like this didn't fix the issue. :cry:
This is strange because <https://docs.microsoft.com/fr-fr/dotnet/framework/configure-apps/file-schema/runtime/> (and many others) renders correctly and the only difference I could find is the backslash thing.
<https://github.com/dotnet/docs/blob/master/docs/framework/configure-apps/file-schema/runtime/index.md>.

cc: @JasonCard

JasonCard:
@Youssef1313 @mairaw I'm sorry for the delayed update on this. I was monitoring for this fix to publish for loc, but after investigation found that all of our localized update commits to dotnet/docs loc repos were failing since 11/12. We've found the root cause, but having difficulty resolving the issue, so currently working on a workaround to have all affected files committed and published by Monday. Will update.

JasonCard:
@Youssef1313 @mairaw It's [fixed now](https://docs.microsoft.com/fr-fr/dotnet/framework/configure-apps/file-schema/network/specifiedpickupdirectory-element-network-settings), although MT engine is adding an extra space before closing brackets, so I will follow up to see if that can be fixed on MT engine side:
fr-fr_fixed

JasonCard:
@Youssef1313 @mairaw We have a question, are the escape/backslash chars (\) before the opening brackets (<) necessary? If those are removed, then those elements will be blocked from translation. If those are needed, then we can request for feature to update our transformer to block them even in those cases.

mairaw:
yes, they are; otherwise, they are not rendered @JasonCard 

Compare here on the preview:

Look like it will render:

with the escape:
\<configuration>

without the escape
<configuration>

mairaw:
This is how it was written:
image

JasonCard:
Thank you for confirming. I will request for a feature to update our transformer to block those even with the escape char, so no fix is needed on your end. Will update on this.

JasonCard:
@mairaw Would it be possible to use the inline no-loc md extensions like below?

\# :::no-loc text=”\<specifiedPickupDirectory>"::: Element (Network Settings) 

\[\**:::no-loc text=”\<configuration>":::**](../configuration-element.md)\

cc @mayast

mairaw:
Can you do one as a test @JasonCard?

JasonCard:
@mairaw Okay, I've opened this PR as test [https://github.com/dotnet/docs/pull/16079](https://github.com/dotnet/docs/pull/16079).

PR #15692, opened 2019-11-04
Remove outdated info from template.md

### LINKED ISSUE #15691: dotnet/docs/blob/master/styleguide/template.md is misleading in term of metadata rendering
**Issue description**

https://github.com/dotnet/docs/blob/master/styleguide/template.md is misleading in term of metadata rendering.
I found this document very useful for my own purpose but in fact it may be outdated on one point:

> Metadata and Markdown Template
>
> This dotnet/docs template contains examples of Markdown syntax, as well as guidance on setting the metadata. To get the most of it, you must view both the raw Markdown and the rendered view (for instance, the raw Markdown shows the metadata block, **while the rendered view does not**).

As you can notice by viewing this file, it seems github renderer saw some unfortunate changes since this document has been written, leading to metadata being rendered against the initial will.

Do you know if this change will be reverted or fixed?

Best regards

Youssef1313:
Fixes #15691

### DIFF
--- a/styleguide/template.md
+++ b/styleguide/template.md
@@ -21,7 +21,7 @@ helpviewer_keywords: [OFFLINE BOOK INDEX ENTRIES]
 ---
 # Metadata and Markdown Template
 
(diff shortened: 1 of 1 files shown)


Thraka (approved):

Thraka:
Thanks. I feel like that sentence didn't add any value, so I think it's appropriate to delete it.

PR #15693, opened 2019-11-04
Update hub page

gewarren:
build -> create (so as not to confuse with compile)
add a few periods
sentence capitalization
no -ing words


BillWagner (approved):
Thanks @gewarren 

PR #15697, opened 2019-11-05
Structs: remove a not-in-place section

pkulikov:
It's not in place, because literals are feature of the simple types, which is a small subset of all struct types. Moreover, with the latest updates to the docs, numeric literals are more fully documented in the articles about integral and floating-point numeric types. Thus, the deleted section doesn't seem to be useful anymore.


BillWagner (approved):
Thanks again @pkulikov 

This is a good catch.  I agree it's not as useful now, and can be removed.  I'll :shipit: now.

PR #15719, opened 2019-11-06
Value types table: removed duplicate info

pkulikov:
Removed the information about numeric suffixes. Those are better documented in the articles about integral and floating-point numeric types. Those two articles are referenced in the table. Ultimately, this whole page will be merged into the "Value types" article, but not by this PR.


BillWagner (approved):
This LGTM @pkulikov 

I'll :shipit: now.

Issue #15766, opened 2019-11-08
Document 'await foreach'


cremor:
Seems like `await foreach` is not yet documented. I'm not sure where the best place for it would be, but it should be documented _somewhere_.

Possible options that I see:
- [await operator (C# reference)](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/await)
- [foreach, in (C# reference)](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/foreach-in)
- A new page

Whatever it happens to be in the end, I think both the existing `await` and `foreach` pages should at least mention `await foreach` and link to its documentation.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 07164632-3cf8-864e-cd1e-485b73ed95a0
* Version Independent ID: 67bfb75a-8dd8-ecf3-ce15-2a531833e8b5
* Content: [await operator - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/await)
* Content Source: [docs/csharp/language-reference/operators/await.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/operators/await.md)
* Product: **dotnet-csharp**
* Technology: **csharp-language-reference**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

pkulikov:
@BillWagner given the trend to consolidate the C# reference docs, I think the `await foreach` statement is worth its own H2-section in the article about iteration statements, which is tracked by #10748

PR #15767, opened 2019-11-08
await operator: mention await foreach

pkulikov:
Contributes to #15766


BillWagner (approved):
This LGTM @pkulikov 

I'll :shipit: now.

Do you think the `await foreach` should be its own page in reference, or a section in the `foreach` reference page?

pkulikov:
> Do you think the `await foreach` should be its own page in reference, or a section in the `foreach` reference page?

@BillWagner I've replied in the issue thread: #15766

PR #15797, opened 2019-11-11
Add yml to supported languages in template.md

Youssef1313:
It seems that yml is a supported identifier. It's getting proper highlighting. Example from [Common web application architectures](https://docs.microsoft.com/en-us/dotnet/architecture/modern-web-apps-azure/common-web-application-architectures#docker-support):

image


mairaw (approved):
Thanks @Youssef1313!

PR #15832, opened 2019-11-13
Creating initial workflow for BC notifications

timheuer:
This adds a simple workflow for the .NET team to better monitor breaking change notifications.

## Summary
As a part of the triage and communication process for our partners working on .NET-related products we are automating some breaking-change notification.  This change enables us to use the public Issues but automatically send information to internal partners as well when this process is used.

## What it does
When an Issue is labeled, we inspect the label and if it is a match to a monitored one we care to notify about (e.g. breaking-change) then we will grab the content of the Issue (body) and create an email to the internal partner team notification list.  This only runs on when issues are labeled.

### DIFF
--- a/.github/workflows/bcnotify.yaml
+++ b/.github/workflows/bcnotify.yaml
@@ -1,0 +1,19 @@
+name: "bc-notification"
+on: 
+  issues:
+    types: [edited, labeled]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


mairaw (approved):

gewarren:
@timheuer Is this workflow still used and/or has it been working as expected? I think people have been sending manual notifications to the breaking change alias. cc @PriyaPurkayastha

PriyaPurkayastha:
This workflow is not being used and like you mentioned, breaking change doc issue creators send an email with link to the issue to dotnetbcn.

PR #15876, opened 2019-11-18
Unmanaged constraint allows only non-nullable types

pkulikov:
The `unmanaged` constraint implies the `struct` constraint and, thus, doesn't allow nullable value types. Beginning with C# 8.0, for an unmanaged value type `T`, the nullable type `T?` is also unmanaged but doesn't satisfy the `unmanaged` constraint because of the underlying `struct` constraint. This is how it's now, but may be changed in future. For more information, see dotnet/csharplang#2916


BillWagner (approved):
These changes LGTM @pkulikov 
I'll :shipit: now.

PR #15929, opened 2019-11-21
Add breaking change for removal of WinForms controls in 3.1

gewarren:
- adds new breaking change include file
- create new file for 3.0 -> 3.1
- reorder versions in selector and TOC from most recent to least recent

### DIFF
--- a/docs/core/compatibility/toc.yml
+++ b/docs/core/compatibility/toc.yml
@@ -10,7 +10,7 @@
     items:
     - name: 3.0 to 3.1
       href: 3.0-3.1.md
-    - name: 3.0 Preview 9 to 3.0 RC1
+    - name: 3.0 Preview 9 to 3.0
       href: 3.0.9-3.0rc1.md
     - name: 3.0 Preview 8 to 3.0 Preview 9
       href: 3.0.8-3.0.9.md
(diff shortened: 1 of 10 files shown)


docs/core/compatibility/toc.yml
```diff
@@ -8,18 +8,20 @@
...
-      href: 2.2-3.0.md
-    - name: 3.0 Preview 6 to 3.0 Preview 7
-      href: 3.0.6-3.0.7.md
-    - name: 3.0 Preview 7 to 3.0 Preview 8
-      href: 3.0.7-3.0.8.md
-    - name: 3.0 Preview 8 to 3.0 Preview 9
-      href: 3.0.8-3.0.9.md
+    - name: 3.0 to 3.1
+      href: 3.0-3.1.md
     - name: 3.0 Preview 9 to 3.0 RC1
```
tdykstra: Are there no breaking changes from 3.0 RC1 to 3.0 GA? If not maybe we should make this entry preview 9 to 3.0 so it won't look like we've left a gap in the list.

tdykstra (approved):
LGTM, one suggestion.

docs/core/compatibility/toc.yml
gewarren: I changed it to just 3.0, but have further questions about versions that I sent in an email. Thanks for the review, I'm going to merge this now.

Issue #10726, opened 2019-02-26
Abridged Reference: builtin value types


BillWagner:
This new article will include:

**Simple types**
- bool
- enum

Sub-tasks are:

- [ ] Create the new articles
- [ ] Search for existing issues on content being migrated and consolidated. This is a good opportunity to fix those.
- [ ] Redirect existing content to the new articles.
- [ ] Remove old articles.
- [ ] Update all links in the .NET Content

BillWagner:
There are new types that should be added to this task:

- `System.ValueTuple`
- `System.Index`
- `System.Range`

`ValueTuple` seems to fit nicely with the 'simple types', where "simple" means a value type that's not a single numeric value. `ValueTuple` is required by the 7.0 language and feels "simple".

Both `Index` and `Range` point to a possible new category: ValueTypes required by the language spec. Maybe it's enough to change the name from "simple" to "non-numeric value types". Or put it in the TOC following "numeric types" and label it "other value types".

pkulikov:
@BillWagner it looks that this issue *redefines* what "simple" means. The existing definition is according to the spec:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/value-types#simple-types

> Maybe it's enough to change the name from "simple" to "non-numeric value types". Or put it in the TOC following "numeric types" and label it "other value types".

I think those two approaches are better. First, no need to redefine the current "simple types" notion. Second, would it be fine even not to label it as "other value types", but have the TOC (and articles) as follows:

Built-in value types:
- Numeric types
- bool
- enum
- System.ValueTuple (or just "C# tuples")
- System.Index and System.Range (or just "Indices and ranges")

Another concern: `char`. I know it's inside the numeric types, but it feels weird. Is it worth its own TOC entry and article (with remarks about its relation to `short`)?

BillWagner:
Thanks for the thoughtful comments @pkulikov.  These discussions continue to be a key part of this project's success. Thank you. 🙇 

I agree that we should not use "Simple Types" to mean anything other than the meaning in the language specification.  That's a great point. 

And, one other type we missed in the lists earlier:  `System.Nullable<T>`. That should be in the list above between `enum` and `System.ValueTuple`.

Regarding `char`: Yes, it is unique. It should have its own page.

BillWagner:
The content from the following files should be moved into this new article:

- docs/csharp/language-reference/keywords/built-in-types-table.md
[...]

PR #15939, opened 2019-11-22
Moved and revised the char article

pkulikov:
Contributes to #10726
Supported by dotnet/samples#1785

@BillWagner could you also please take a look at https://github.com/dotnet/docs/issues/10726#issuecomment-550174448, which is about how to further address #10726


docs/csharp/language-reference/builtin-types/char.md
```diff
@@ -0,0 +1,53 @@
+---
+title: "char type- C# reference"
+ms.custom: seodec18
```
BillWagner: This should be removed. We were tracking the results of one targeted push for SEO. This new article shouldn't be tracked the same.

BillWagner (changes requested):
This is great @pkulikov 

There's one small suggestion, and then it's ready to :shipit:

Thanks for all your work in this area.

BillWagner (approved):

BillWagner:
Thanks @pkulikov 
I'll :shipit: when the build finishes.

docs/csharp/language-reference/builtin-types/char.md
pkulikov: @BillWagner that's clear: I've removed the line.

PR #15940, opened 2019-11-22
Try to resolve yaml duplicate keys

v-pegao:
## Summary

We created this pull request to fix yaml duplicate keys and unblock docfx v3 migration. **Please help review if this is expected fix or not**

If there is yaml duplicate key, only the value of the last key will be kept on published page by default, which can cause toc doesn't work as expected, It's better to avoid that. 

### Details
We find there are duplicate `href` for some toc items and there are some toc item without `href`, we try to resolve them by adjusting the line of duplicate `href` based on the `href` article name and toc name.


mairaw (approved):
Looks great @v-pegao. Thanks for fixing that!

PR #15943, opened 2019-11-22
updating links from fwlink to real link if exists

eddynaka:
updating links from fwlink to real link if exists - contributes to #3391


mairaw (approved):
LGTM :shipit:

PR #15965, opened 2019-11-24
Replace WCF Data Services tokens

nxtn:


gewarren (approved):
@NextTurn Thank you for doing this!

PR #15975, opened 2019-11-25
move installer area to win/mac only

Thraka:
Minor fix for the install guide. The "install with an installer" sections were appearing on linux which is not applicable.

Also, add myself as codeowner for this area.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -59,7 +59,7 @@
 # Deployment
 /docs/core/deploying/** @Thraka
 # Install
-/docs/core/install/**": @Thraka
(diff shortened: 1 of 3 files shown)


.github/CODEOWNERS
```diff
@@ -58,6 +58,8 @@
 /docs/core/whats-new/** @Thraka
 # Deployment
 /docs/core/deploying/** @Thraka
+# Install
+/docs/core/install/**": @Thraka
```
gewarren:
```suggestion
/docs/core/install/** @Thraka
```

gewarren (approved):
Approved with suggestion.

PR #16028, opened 2019-11-28
remove duplicate text in install instruction

pgroene:
This text duplicated in the instruction, removed on of the duplications

## Summary

Describe your changes here.


Youssef1313:
I see there is only one match of "`sudo apt-get install dotnet-runtime-3.0`", so I don't think there is a duplication.

pgroene:
You are correct, I did not see the difference between asp.net core and .net core.

PR #16036, opened 2019-11-30
Update reflection.md

Youssef1313:
- used an xref for GetType, and it was referred to as static. I removed the word "static" as I don't think it's a static method.
- Removed the noisy `System.` and `System.Reflection.` prefixes and added a note about add `using System;` and `using System.Reflection;`.
- Inlined outputs.
- Simplified `System.Int32` to just `int`.
- Updated sample output based on the output I'm getting using .NET Core 3.0 and VS 2019.
    
    image


BillWagner (approved):
This also LGTM @Youssef1313 
I'll :shipit: now.

PR #16092, opened 2019-12-04
Fix empty header for linux pivot

Youssef1313:
Since there is no pivot under `Next steps` for linux, the header itself renders without content. I've moved the header to the existing pivots.


Thraka:
Thanks for catching this. I think I did have next steps but I was working on a rendering bug and must have culled them without realizing. I'll add these back in with the updates I'm doing for 3.1 today. I'll close this but good catch!

PR #16107, opened 2019-12-04
removing en-us from link

eddynaka:


mairaw (approved):
LGTM. Thanks @eddynaka!

PR #16114, opened 2019-12-05
Update .NET Core get started articles: console app and class library

gewarren:
Combine C# and VB articles for first console app and class library. Also update them for VS 2019.

Contributes to #16046


gewarren:
Closing because wrong base branch.

PR #16153, opened 2019-12-07
Replace old Windows version tokens

nxtn:
Contributes to #5199 

Most of the next few tokens to be replaced (in this and following PRs) involve old Windows versions. This is a chance to review these outdated content based on #5199.

---

I assumed that the features "available only on Windows XP" should be also available on later versions of Windows?


Thraka (approved):
Looks good thanks!

PR #16166, opened 2019-12-09
Moved and revised the enum article

pkulikov:
Contributes to #10726
Supported by dotnet/samples#1830

@BillWagner while working on this PR I've realized that maybe the "built-in" word can be dropped from the TOC and paths:
image

Note that `enum`, `struct`, `class`, and `interface` are the constructs to create new types, rather than the builtin-types. So, they look a bit strange under the "Built-in types" node, but perfectly well under the "Types" one.

By the way, [your gist](https://gist.github.com/BillWagner/047cb97f6c0fe542d3a1299e994b2acd#-defining-types-members-and-variables) puts `class`, `struct`, and `interface` (but not `enum`) under the "Declaring types" node. Maybe, the whole "Defining types, members and variables" node should be moved under the new "Types" node as the "Defining types and its members" one? Or still let it be separate and just move `class`, `struct`, and `interface` out of it to the "Types" node? What do you think? As for the `class`, `struct`, `interface`, and `enum` entries, I would keep them as they are now in the TOC.


docs/csharp/language-reference/builtin-types/bool.md
```diff
@@ -23,16 +23,16 @@ You can use the `true` and `false` literals to initialize a `bool` variable or t
...
 ## Three-valued Boolean logic
 
 Use the nullable `bool?` type, if you need to support the three-valued logic, for example, when you work with databases that support a three-valued Boolean type. For the `bool?` operands, the predefined `&` and `|` operators support the three-valued logic. For more information, see the [Nullable Boolean logical operators](../operators/boolean-logical-operators.md#nullable-boolean-logical-operators) section of the [Boolean logical operators](../operators/boolean-logical-operators.md) article.
 
 For more information about nullable value types, see [Nullable value types](nullable-value-types.md).
 
+## Conversions
+
+C# provides only two conversions that involve the `bool` type. Those are an implicit conversion to the corresponding nullable `bool?` type and an explicit conversion from the `bool?` type. However, .NET provides additional methods that you can use to convert to or from the `bool` type. For more information, see the [Converting to and from Boolean values](/dotnet/api/system.boolean#converting-to-and-from-boolean-values) section of the <xref:System.Boolean?displayProperty=nameWithType> API reference page.
+
```
pkulikov: Minor bit not related to the enum work: swapped the sections to introduce the `bool?` type before it's talked about in the "Conversions" section.

docs/csharp/language-reference/builtin-types/enum.md
```diff
@@ -0,0 +1,92 @@
+---
+title: "Enumeration types - C# reference"
+description: "Learn about C# enumeration types that represent a set of named constants and how to define them with the `enum` keyword"
+ms.date: 09/12/2019
```
Youssef1313: I think you swapped the month with the day here ?

```suggestion
ms.date: 12/09/2019
```

docs/csharp/language-reference/builtin-types/enum.md
```diff
@@ -0,0 +1,92 @@
...
+
+```csharp
+enum Season
+{
+    Spring,
+    Summer,
+    Autumn,
+    Winter
+}
+```
```
Youssef1313: Why not reference these lines from the snippets repo ?

docs/csharp/language-reference/builtin-types/enum.md
```diff
@@ -0,0 +1,92 @@
...
+
+The <xref:System.Enum?displayProperty=nameWithType> type is the abstract base class of all enumeration types. It provides a number of methods to get information about an enumeration type and its values. For more information and examples, see the <xref:System.Enum?displayProperty=nameWithType> API reference page.
+
+Beginning with C# 7.3, you can use `System.Enum` in a base class constraint (that is known as the [enum constraint](../../programming-guide/generics/constraints-on-type-parameters.md#enum-constraints)) to specify that a type parameter is an enumeration type.
+
+You cannot define a method inside the definition of an enumeration type, but you can create an [extension method](../../programming-guide/classes-and-structs/extension-methods.md) to add functionality to an enumeration type.
+
+Like with any [constant](../../programming-guide/classes-and-structs/constants.md), at compile time all references to individual values of an enumeration type are replaced with the corresponding integral values.
```
Youssef1313:
```suggestion
Like with any [constant](../../programming-guide/classes-and-structs/constants.md), at compile time, all references to individual values of an enumeration type are replaced with the corresponding integral values.
```

docs/csharp/language-reference/builtin-types/enum.md
```diff
@@ -0,0 +1,92 @@
...
+
+For any enumeration type, there exist [boxing and unboxing](../../programming-guide/types/boxing-and-unboxing.md) conversions to and from the <xref:System.Enum?displayProperty=nameWithType> type, respectively.
+
+## Enumeration types as bit flags
+
+Typically you use an enumeration type to represent a set of mutually exclusive values. However, in some scenarios a combination of enumeration values can also be a meaningful enumeration value. Then, you can define only those enum members that represent a bit field in an enumeration value and use the [bitwise logical operators](../operators/bitwise-and-shift-operators.md#enumeration-logical-operators) to combine enumeration values. The associated values of the enum members that represent a bit field should be the powers of two. To indicate that an enumeration type declares bit fields, apply the [Flags](xref:System.FlagsAttribute) attribute to it. The following example defines the `Days` enumeration type and shows the behavior of the bitwise operators with its instances.
```
Youssef1313:
```suggestion
Typically, you use an enumeration type to represent a set of mutually exclusive values. However, in some scenarios, a combination of enumeration values can also be a meaningful enumeration value. Then, you can define only those enum members that represent a bit field in an enumeration value and use the [bitwise logical operators](../operators/bitwise-and-shift-operators.md#enumeration-logical-operators) to combine enumeration values. The associated values of the enum members that represent a bit field should be the powers of two. To indicate that an enumeration type declares bit fields, apply the [Flags](xref:System.FlagsAttribute) attribute to it. The following example defines the `Days` enumeration type and shows the behavior of the bitwise operators with its instances.
```

Youssef1313 (changes requested):

docs/csharp/language-reference/builtin-types/enum.md
pkulikov: I think such little examples easier to maintain inline as it is

docs/csharp/language-reference/builtin-types/enum.md
pkulikov: Where can I find the guidance for those comma suggestions? I wonder if those commas are must or can be dropped (that is, both variants would be possible).

Youssef1313: I don't know if there is a guidance on that. but I think it reads better with the comma.

pkulikov: @Youssef1313 I've tried to rephrase sentences to avoid such situations.

docs/csharp/language-reference/builtin-types/enum.md
BillWagner: We've noticed that, as a team, we've adopted slightly different strategies for managing snippets, full programs, tutorials and examples. We're starting discussions to see what we can share, and if there is a common set of practices we should use. 

We'll update the contributing articles once we've had those discussions. In the meantime, this is fine.

BillWagner (commented):
Overall, this is great.

I'm concerned that the enum article reads too much like the spec. I know this is not how you wrote it, but based on the prior article.  

To me, here are the key concepts I'd want developers to get from reading this article:
- `enum` types are a set of integral constants, represented by `int`, unless you specify a different integral type.
- You cannot specify any behavior in an `enum` type (although extension methods are allowed).
- The values of the enum fields are 0...n in text order, unless you specify them.
- 0 is the default value of any `enum` type, even if you do not specify a valid `0` value.
- Bitflags are useful for creating OR logic of enum values.

From those facts, I think we can state this recommendation for developers:

> Declare enums to represent a choice (one of many) or a combination of choices (many of many). 

What are your thoughts? Should this be a second PR?

pkulikov:
@BillWagner I think I understand your idea, thanks! I prefer to make related adjustments in a second PR, after this one is merged.

BillWagner:
@pkulikov That's a good plan.

I'll approve this PR, and :shipit: now.

Thanks again!

BillWagner (approved):
This is a great reorganization @pkulikov 

I'll :shipit: now.

BillWagner:
I realize I hadn't addressed this comment in your PR description:

> By the way, [your gist](https://gist.github.com/BillWagner/047cb97f6c0fe542d3a1299e994b2acd#-defining-types-members-and-variables) puts `class`, `struct`, and `interface` (but not `enum`) under the "Declaring types" node. Maybe, the whole "Defining types, members and variables" node should be moved under the new "Types" node as the "Defining types and its members" one? Or still let it be separate and just move `class`, `struct`, and `interface` out of it to the "Types" node? What do you think? As for the `class`, `struct`, `interface`, and `enum` entries, I would keep them as they are now in the TOC.

I like that idea (moving "defining types ..." under "Declaring types".  This also points to an ongoing task for me: I should update that gist once a month to reflect what's been done, and what we've discussed and agreed to in terms of changes.  I'll take a run at an update this week.

pkulikov:
> I like that idea (moving "defining types ..." under "Declaring types".

To confirm: type and member declaration nodes would be moved under the types nodes.

@BillWagner should I then (in a separate large and boring PR) rename the "builtin-types" folder into the "types" folder? That would result in the URLs of such kind:
> `https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/types/bool`

BillWagner:
> should I then (in a separate large and boring PR) rename the "builtin-types" folder into the "types" folder?

@pkulikov Let's wait on that. I'll update the TOC gist first. I'd like to avoid unnecessary churn. The exercise of reviewing and updating the TOC should help with that.

PR #16214, opened 2019-12-12
add nested stackalloc

BillWagner:
This is a small spec change, but should be documented.

This also fixes the warning on the missing title for this article.


gewarren (approved):
LG! :shipit:

PR #16217, opened 2019-12-12
Update how-to-compare-the-contents-of-two-folders-linq.md

AntiTcb:
## Summary

Change Count() to Any(). Improves performance for large file structures since it doesn't require a full enumeration.


BillWagner (approved):
That's a great change @AntiTcb We do recommend `Any()` over `Count()` for this check.

PR #15472, opened 2019-10-28
Data setup & load for model builder

cjgronlund:
## Summary
new article covering data set up and loading for Model Builder.

[AB#1622265](https://dev.azure.com/mseng/677da0fb-b067-4f77-b89b-f32c12bb8617/_workitems/edit/1622265)

PR #16227, opened 2019-12-12
Remove unwanted files

gewarren:
These were added by #15472, I assume accidentally.


gewarren:
@BillWagner or @maira Can you review please?

mairaw (approved):
Indeed, it looks like an accident. @cjgronlund could confirm it.

PR #16228, opened 2019-12-13
Delete orphan images

gewarren:
Also testing out new .svg image deletion functionality in cleanrepo.exe tool.


gewarren:
Close/reopen for build.

BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

Thraka:
Should we leave the machine learning content alone? Perhaps they are storing those svg files there for future reference? Considering it's a totally different team that does things separate from us (but is in the repo) we really should let them manage their assets, yes?

PR #16265, opened 2019-12-15
Removed the duplicate article about enums from the tour of C#

pkulikov:
As the tour is going to be rewritten, I think it's fine to remove the duplication of the revised content in the language reference. Also, there are already the links to the language reference articles in [Types and variables](https://docs.microsoft.com/en-us/dotnet/csharp/tour-of-csharp/types-and-variables). I would understand if that PR were not be merged. And fine with that as this PR was fast and small.


BillWagner (approved):
This LGTM @pkulikov 

I'll :shipit: now.

PR #16269, opened 2019-12-15
InvalidIndexException -> IndexOutOfRangeException

Youssef1313:


docs/standard/tour.md
```diff
@@ -91,7 +91,7 @@ In .NET, delegates are commonly used in event handlers, in defining asynchronous
...
 
-Generics were added to help programmers implement generic data structures. Before their arrival in order for a type such as the `List` type to be generic, it would have to work with elements that were of type `object`. This had various performance and semantic problems, along with possible subtle runtime errors. The most notorious of the latter is when a data structure contains, for instance, both integers and strings, and an `InvalidCastException` is thrown on working with the list's members.
+Generics were added to help programmers implement generic data structures. Before their arrival, in order for a type such as the `List` type to be generic, it would have to work with elements that were of type `object`. This had various performance and semantic problems, along with possible subtle runtime errors. The most notorious of the latter is when a data structure contains, for instance, both integers and strings, and an <xref:System:InvalidCastException> is thrown on working with the list's members.
```
gewarren:
```suggestion
Generics were added to help programmers implement generic data structures. Before their arrival, in order for a type such as the `List` type to be generic, it would have to work with elements that were of type `object`. This had various performance and semantic problems, along with possible subtle run-time errors. A common run-time error is when a data structure contains, for example, both integers and strings, and an <xref:System:InvalidCastException> is thrown while processing the list's members.
```

Youssef1313: @gewarren, Applied the suggestion. I feel like the "and" before "an InvalidCastException" is grammatically not correct and can be removed. May you confirm ? I'm not a native speaker so I might be mistaken.

gewarren (approved):
The "an" before InvalidCasException is correct. I'll merge now. Thanks for the fixes!

Youssef1313:
> 
> 
> The "an" before InvalidCasException is correct. I'll merge now. Thanks for the fixes!

I meant the "and".

> when a data structure contains, for example, both integers and strings, and an InvalidCastException is thrown while processing the list's members.

vs

> when a data structure contains, for example, both integers and strings, ~~and~~ an InvalidCastException is thrown while processing the list's members.

gewarren:
Ah, sorry. I think the "and" is required in this case because an InvalidCastException might not be thrown, depending on the processing. It changes the meaning to remove it.

PR #16271, opened 2019-12-15
Text/snippet mismatch

guardrex:
The quickest update is to make the text match the snippet ... only one file to change due to the snippet name (`SnippetWFDerived`) possibly requiring an update, too. If you want to go the other way and change both files, I understand ... "WithWind" is cool.

Cross-ref to that snippet :point_right: https://github.com/dotnet/samples/blob/master/snippets/core/system-text-json/csharp/WeatherForecast.cs#L82-L87

### DIFF
--- a/docs/standard/serialization/system-text-json-how-to.md
+++ b/docs/standard/serialization/system-text-json-how-to.md
@@ -455,7 +455,7 @@ To minimize escaping you can use <xref:System.Text.Encodings.Web.JavaScriptEncod
 
 ## Serialize properties of derived classes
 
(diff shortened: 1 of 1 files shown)


tdykstra (approved):
@guardrex Good catch, thanks for fixing!

guardrex:
Sure thing ... nice topic! Very helpful ...

I needed this to make a `System.Text.Json` version of a health checks JSON output delegate that was using `Newtonsoft.Json`. It went very smoothly ... and at this point in the year before finally getting some R&R 😵😩, I really *needed that.* :smile:🎁

Issue #16277, opened 2019-12-16
unmanaged type


paulomorgado:
&gt; In expression stackalloc T[E], T must be an unmanaged type and E must be an expression of type int.

There are no examples of unmanaged types and no mention on the docs of value types, which are the ones used in the examples.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ecde487e-96e5-85c2-ff87-227994f705ad
* Version Independent ID: f77588b3-9b31-f8ce-fad5-bf937337eb8c
* Content: [stackalloc operator - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/stackalloc)
* Content Source: [docs/csharp/language-reference/operators/stackalloc.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/operators/stackalloc.md)
* Product: **dotnet-csharp**
* Technology: **csharp-language-reference**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

pkulikov:
> There are no examples of unmanaged types...

Examples of (user-defined) unmanaged types are in the corresponding [reference article](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/unmanaged-types). I'm afraid that including them into the `stackalloc` article would make examples focused more on an unmanaged type, rather than on `stackalloc`. So, that would be the loss of focus. To me, for example, the main purpose of the `stackalloc` example is to provide the syntax how to write the `stackalloc` expression, doesn't really matter which type is allocated.

> ...no mention on the docs of value types, which are the ones used in the examples

Do you mean that every page that uses `int` (and other built-in numerics) in the examples should reference the corresponding reference articles?

Though, `int` in the text can be converted into the link.

paulomorgado:
In hindsight, I might have misunderstood what `unmanaged types` are.

PR #16278, opened 2019-12-16
stackalloc: added the link to int page

pkulikov:
I've also moved the paragraph to be the last one, as it's not so important (the compiler would provide that information in case of not using an unmanaged type) to be at the top.

Inspired by #16277


BillWagner (approved):
This LGTM @pkulikov 

I'll :shipit: now.

PR #16300, opened 2019-12-17
Update the link of "What's new in .NET Core 3.1"

imba-tjd:


mairaw (approved):
Good catch! Thanks @imba-tjd!

PR #16304, opened 2019-12-17
Add links to Core, standard, asp.net core

### LINKED ISSUE #2271: The .NET docs homepage became less useful
**Is your feature request related to a problem? Please describe.**

In the past days, the .NET docs homepage has been changed with a new design:

https://docs.microsoft.com/en-us/dotnet/

I used that page as a starting point for reaching products docs in a fast way. Now I find it much harder to use.

For example, often I want to open the first page of the ASP.NET Core docs, so that I can search articles through the sidebar.

In the previous website's version, it was easy to find the ASP.NET Core link in the middle of the screen:

immagine

With the current version, I have to scroll almost to the bottom of the page and then click this link:

immagine

That's not really intuitive.

Previously, there was also a link to the EF Core documentation. It's not there anymore unfortunately.

**Describe the solution you'd like**

The new page design might be good for new users, but it's not helpful for people that already know where they want to go.

There should be more quick links to specific products links. The "Programming languages" section is well designed, so it would be helpful to have a similar thing for products like ASP.NET Core and EF Core.

BillWagner:
Fixes MicrosoftDocs/Feedback#2271

We still want to optimize for the net-new developer, but we need to include the important links for our current .NET community.

### DIFF
--- a/docs/index.yml
+++ b/docs/index.yml
@@ -251,6 +251,15 @@ additionalContent:
       - title: ".NET for Apache Spark API reference"
         summary: API reference documentation for .NET for Apache Spark
         url: https://docs.microsoft.com/dotnet/api/?view=spark-dotnet
+      - title: ".NET Core"
+        summary: ".NET Core documentation"
+        url: core/index.md
+      - title: ".NET Standard"
+        summary: ".NET Standard documentation"
+        url: standard/index.md
+      - title: "ASP.NET Core"
+        summary: ASP.NET Core documentation"
+        url: /aspnet/core/
       # Card
       - title: "C# language reference"
         summary: C# language reference and specification
--- a/docs/index.yml
+++ b/docs/index.yml
@@ -73,6 +73,12 @@ conceptualContent:
         - url: standard/net-standard.md
           itemType: overview
           text: .NET Standard
+        - url: /aspnet/core/
+          itemType: overview
+          text: ASP.NET Core
+        - url: /ef/core/
+          itemType: overview
+          text: Entity Framework Core
         - url: standard/get-started.md
           itemType: get-started
           text: Get started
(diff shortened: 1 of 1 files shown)


docs/index.yml
```diff
@@ -251,6 +251,15 @@ additionalContent:
...
         summary: API reference documentation for .NET for Apache Spark
         url: https://docs.microsoft.com/dotnet/api/?view=spark-dotnet
+      - title: ".NET Core"
+        summary: ".NET Core documentation"
+        url: core/index.md
+      - title: ".NET Standard"
+        summary: ".NET Standard documentation"
+        url: standard/index.md
+      - title: "ASP.NET Core"
+        summary: ASP.NET Core documentation"
```
gewarren:
```suggestion
        summary: ASP.NET Core documentation
```

docs/index.yml
```diff
@@ -251,6 +251,15 @@ additionalContent:
       - title: ".NET for Apache Spark API reference"
         summary: API reference documentation for .NET for Apache Spark
         url: https://docs.microsoft.com/dotnet/api/?view=spark-dotnet
+      - title: ".NET Core"
```
gewarren: I don't think this is the right spot for these links, because it's under API reference. We already have links to .NET Core and .NET Standard in the first two cards. Maybe just move ASP.NET Core to the second card ".NET concepts"? Also, the feedback mentioned having a link to EF Core.

gewarren (changes requested):

docs/index.yml
BillWagner: Great suggestion!

BillWagner:
This is ready for another review @gewarren 

Great suggestions!

gewarren (approved):
LGTM! :shipit:

PR #16317, opened 2019-12-18
XML entities

TheConstructor:
It seems they were lost some day in the past.

## Summary

Describe your changes here.


Thraka (approved):
Thanks for catching this!

PR #16318, opened 2019-12-18
Incorrect `ref readonly` statement keyword order.

mayuki:
## Summary

Fix incorrect `ref readonly` statement keyword order.

Declaration for _Ref readonly local_ or _Ref return value_ must be `ref readonly`. The order of these keywords is fixed.
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/ref#ref-readonly-locals

### DIFF
--- a/docs/csharp/write-safe-efficient-code.md
+++ b/docs/csharp/write-safe-efficient-code.md
@@ -132,7 +132,7 @@ public struct Point3D
 }
 ```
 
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks for spotting this and making the fix @mayuki We appreciate it!

You should see the changes on the live site in the next day or two, on our regular publishing cycle. Thanks for making your first contribution to docs! 🎆

PR #16341, opened 2019-12-19
Escape the dollar sign

pkulikov:
[Formatting numeric results table](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/formatting-numeric-results-table)
The dollar sign (visible in the GitHub preview) disappears when published:
image

I escape it with `\\`, like the [Standard numeric format strings](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings) page does. However, that way the GitHub preview is not accurate anymore (for the both pages):
image


mairaw:
It looks fine on review:
image

It's probably because of the Latex extension that is custom to docs.

mairaw:
Build was stuck.

mairaw (approved):
Thanks @pkulikov!

PR #16344, opened 2019-12-19
Spark: Change Spark H2s

mamccrea:
## Summary

Updating .NET for Apache Spark H2s with correct branding.


BillWagner (approved):
Thanks @mamccrea 

I've approved this, and you can :shipit: when ready.

I also updated the code so the incorrect heading won't return next month.

PR #16364, opened 2019-12-20
Remove title suffix from markdown files

Youssef1313:
The title suffix already exists in the docfx settings.

### DIFF
--- a/docs/visual-basic/language-reference/operators/nameof.md
+++ b/docs/visual-basic/language-reference/operators/nameof.md
@@ -1,11 +1,11 @@
 ---
-title: "NameOf operator - Visual Basic"
+title: "NameOf operator"
 description: "Learn how to use the NameOf operator in Visual Basic"
 ms.date: 10/27/2019
 helpviewer_keywords:
   - "NameOf operator [Visual Basic]"
 ---
-# NameOf operator - Visual Basic
+# NameOf operator
 
 The `NameOf` operator obtains the name of a variable, type, or member as the string constant:
 
--- a/docs/visual-basic/language-reference/operators/nameof.md
+++ b/docs/visual-basic/language-reference/operators/nameof.md
@@ -5,7 +5,7 @@ ms.date: 10/27/2019
 helpviewer_keywords:
   - "NameOf operator [Visual Basic]"
 ---
-# NameOf operator
+# NameOf operator - Visual Basic
 
 The `NameOf` operator obtains the name of a variable, type, or member as the string constant:
 
--- a/docs/core/additional-tools/dotnet-svcutil.xmlserializer-guide.md
+++ b/docs/core/additional-tools/dotnet-svcutil.xmlserializer-guide.md
@@ -1,5 +1,5 @@
 ---
-title: Using dotnet-svcutil.xmlserializer on .NET Core
+title: Using dotnet-svcutil.xmlserializer
 description: Learn how you can use the `dotnet-svcutil.xmlserializer` NuGet package to pre-generate a serialization assembly for .NET Core projects.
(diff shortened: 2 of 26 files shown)


docs/visual-basic/language-reference/operators/nameof.md
```diff
@@ -1,11 +1,11 @@
 ---
-title: "NameOf operator - Visual Basic"
+title: "NameOf operator"
 description: "Learn how to use the NameOf operator in Visual Basic"
 ms.date: 10/27/2019
 helpviewer_keywords:
   - "NameOf operator [Visual Basic]"
 ---
-# NameOf operator - Visual Basic
+# NameOf operator
```
Youssef1313: @mairaw, I'm not sure if the suffix should be deleted from the h1 or only from the title. Can you confirm ?

Youssef1313: Also, before I go with further changes, check if these changes are correct. And if there is some cases where I **shouldn't** do them.

cartermp (changes requested):
In many cases the prefix or suffix maps directly to the name of the URL - for example, `F# types` is the title because the name of the url is `fsharp-types`, which is also how we typically refer to that concept. I think it's also good to still have things like `f# language reference` and the like.

docs/visual-basic/language-reference/operators/nameof.md
mairaw: You can keep them in H1. The suffix is only needed for SEO branding in the title metadata

mairaw:
@cartermp his changes are good. We've been adding brand suffixes to the titles via docfx.json so F# shows up repeated
image

mairaw (approved):
Thanks @Youssef1313. LGTM.

I'd open multiple PRs to not have too many areas involved in a single PR. I can try to run a query in our database to see if I can find other duplicate titles.

Youssef1313:
> I can try to run a query in our database to see if I can find other duplicate titles.

That would be easier than doing that manually.

mairaw:
For sure. I could probably paste the results in a spreadsheet here.

mairaw:
Here it is:

[titles.xlsx](https://github.com/dotnet/docs/files/3989976/titles.xlsx)

Youssef1313:
Most of the links I've looked at in the link doesn't have duplications in the title.
For example, `.NET Application Architecture Guidance | Microsoft Docs`

image

mairaw:
Yes, that is just the list of all titles. 

50 is not much but I think smaller PRs are quicker and easier to review. If @cartermp approves or make suggestions, then we can merge and start a new one quickly.

Youssef1313:
> 
> 
> Yes, that is just the list of all titles.
> 
> 50 is not much but I think smaller PRs are quicker and easier to review. If @cartermp approves or make suggestions, then we can merge and start a new one quickly.

Okay, I'll keep this one with the current changes and open a new one after it's reviewed and merged.

cartermp (approved):

Youssef1313:
@mairaw, For an article like this <https://docs.microsoft.com/en-us/dotnet/framework/wcf/accessing-services-using-a-wcf-client>. If I changed the title to `"Accessing Services Using a Client"` instead of `"Accessing Services Using a WCF Client"`, the title will be the same of another article <https://docs.microsoft.com/en-us/dotnet/framework/wcf/feature-details/accessing-services-using-a-client>. I'm not sure what to do in that case. Are the two articles kind of duplicate and should be combined into one article ? Or they're different things. I'm not familiar with WCF.

Youssef1313:
ping @mairaw 

+ There are other examples:

image

PR #16400, opened 2019-12-25
Update byrefs.md

ebresafegaga:
outref<'T> should be written to, not read from. In the Outref semantics section.


cartermp (approved):
Thanks!

PR #16418, opened 2019-12-27
Changed _integrationEventLogService to _catalogIntegrationEventService

napoleonjones:
Changed texted to reference _catalogIntegrationEventService

## Summary

Document referenced _integrationEventLogService which does not exist in the sample code.  Changed texted to reference _catalogIntegrationEventService


nishanil (approved):
LGTM. Good catch!

PR #16439, opened 2019-12-29
Typo "[Try…Catch…Finally]"→"[Try...Catch...Finally Statement]"

hyoshioka0128:
https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/operators/await-operator

## Summary

Describe your changes here.


docs/visual-basic/language-reference/operators/await-operator.md
```diff
@@ -52,7 +52,7 @@ Await AsyncMethodThatReturnsTask()
...
+An `Await` expression can occur only in the body of an immediately enclosing method or lambda expression that is marked by an `Async` modifier. The term *Await* serves as a keyword only in that context. Elsewhere, it is interpreted as an identifier. Within the async method or lambda expression, an `Await` expression cannot occur in a query expression, in the `catch` or `finally` block of a [Try…Catch…Finally Statement](../../../visual-basic/language-reference/statements/try-catch-finally-statement.md), in the loop control variable expression of a `For` or `For Each` loop, or in the body of a [SyncLock](../../../visual-basic/language-reference/statements/synclock-statement.md) statement.
```
Youssef1313:
```suggestion
An `Await` expression can occur only in the body of an immediately enclosing method or lambda expression that is marked by an `Async` modifier. The term *Await* serves as a keyword only in that context. Elsewhere, it is interpreted as an identifier. Within the `Async` method or lambda expression, an `Await` expression cannot occur in a query expression, in the `Catch` or `Finally` block of a [Try…Catch…Finally statement](../statements/try-catch-finally-statement.md), in the loop control variable expression of a `For` or `For Each` loop, or in the body of a [SyncLock](../statements/synclock-statement.md) statement.
```

@BillWagner, I think it was allowed to use await inside a catch block in some version of C#. So, I'm expecting the same to Visual Basic.

mairaw (approved):
The changes look good @hyoshioka0128. Thanks!

@Youssef1313 if you believe if this paragraph is not accurate, please open a new issue so we can follow up separately with @KathleenDollard.

PR #16452, opened 2019-12-29
Add error messages to C# compiler page titles

ChrisMaddock:
Contributes to #4698. This is currently done on top of #16449 - once that's merged, I'll rebase this one.

I used a regex to add the error messages to the title of each compiler page for SEO, as suggested in #4698. (It also standardises to a single new line after the title - to explain the other change in this PR!)

I can push the rest of the files next, but here's the first 50 - just to check the format is as we want. 🙂 

-------------

Here's what the regular expression looked like. I then also converted the titles to sentence-case.

```
(?s)---
title: \"(.*?) - "
(.*?)
---
# (.*?)$

(.*?)$
```

```
---
title: \"C# $1 - $4"
$2
---
# $3

$4
```


BillWagner:
I'll also wait on this pending answers on #16449

BillWagner:
I'm going to close this.

I'm afraid that recovering this would be more work than restarting the task, once we have redirections in place.

PR #16478, opened 2019-12-31
Replace Win32 tokens

nxtn:


Thraka (approved):
LGTM. Quick scan and it looks good.

PR #16547, opened 2020-01-06
Make sentence more accurate

Youssef1313:
It seems that:

> Changes made from instance A will be visible immediately to instances B and C if they access the field.

Implies that one can do something like the following:

```csharp
public class Program
{
  public static void Main()
  {
    var a = new Class1();
    a.MyStaticField = "Youssef";
  }
}

public class Class1
{
  public static string MyStaticField;
}
```

which causes a compile-time error.
So, I changed the sentence and stated that the static field can't be accessed by an instance name.


BillWagner (approved):
This LGTM. I'll :shipit:

PR #16548, opened 2020-01-06
Remove author from articles - core/diagnostics

Youssef1313:
The author and ms.author are already set for `"docs/core/diagnostics/**/**.md"` globally in docfx.json


mairaw (approved):
Thank you for helping me clean up those values @Youssef1313!

PR #16556, opened 2020-01-06
Add .NET Fx to .NET Core breaking changes doc to TOC

gewarren:
Follow up to #16331


gewarren:
I don't think I excluded it from the TOC on purpose originally, but LMK if you think this works Maira.

mairaw (approved):
Change LGTM. I also think that this specific article needs a note that it's WIP. I'm sure there's more than just those breaking changes between the two platforms.

PR #16576, opened 2020-01-07
fix casing of curl ssl option

scottaddie:
Use the same casing as displayed in the curl manpage.

image


mairaw (approved):
LGTM. Thanks @scottaddie.

PR #16599, opened 2020-01-09
Fix MD004

Youssef1313:


mamccrea (approved):

tdykstra (commented):
@Youssef1313 Thank you for this contribution, but the team discussed this rule and decided to disable it.

Youssef1313:
@tdykstra, Hmm, I see the occurrences aren't too large.
My **own** opinion is to disable MD rules when there are too much violations. Otherwise, I'd enable them with the exception of some rules that must be enabled. For instance, the html rule, because sometimes there is no choice other than using html tags.

tdykstra:
@Youssef1313 I can't speak for the other team members who joined in this decision, but speaking for myself, I see several issues with MD004:

* I question the intrinsic value of the rule. "Consistency for consistency's sake" is not necessarily a good thing.  For some issues, enforcing consistency is more like nitpicking than something that ensures more readable Markdown. It doesn't take a lot of mental energy to understand that either hyphens or asterisks at the beginning of lines can translate to bullets.
* The default rule doesn't give you consistency across a doc set, only within an article.  So you could still have variation even within one PR, which undercuts any possible benefits of this kind of consistency. Actually it makes things worse because you have to keep track of what a given article has when entering new text, in order not to run afoul of MD004.
* Different writers have different habits and preferences for asterisk vs. hyphen. There's no consensus about the benefits of one over the other being great enough to decree that one or the other must be used exclusively across the repo. Personally I prefer asterisks, and I'm not inclined to switch to hyphens; nor would I be inclined to force the hyphen folks to switch to asterisks.

Youssef1313:
Good points. Thanks for letting me know the reasoning behind disabling it.

PR #16604, opened 2020-01-09
Inform developers that DotNetCliToolReference is outdated

Tyrrrz:
## Summary

Added information that developers should use .NET Core local tools now instead of `DotNetCliToolReference`.


tdykstra (approved):
@Tyrrrz Thanks!

PR #16610, opened 2020-01-09
Unsupported APIs - binary serialization

gewarren:
- add note about binary serialization
- remove two explicit interface implementations for serialization

Related to #16608


mairaw (approved):
LGTM

Issue #13813, opened 2019-08-12
[desktop-guide wpf] Scrub+port the XAML-Services documentation


Thraka:
Related to #10649

XAML services doc set is used a lot throughout the existing documentation. As the existing documentation is ported to the Desktop Guide, the XAML Services docset continues to be referenced.

- The XAML Services probably doesn't need the term ".NET Framework" should be replaced with ".NET" to make it generic.

Steps:

- [x] Move articles from Framework to Desktop-WPF (#16193)
- [x] Scrub articles for .NET Framework to make generic (#16629)
- [x] Fix badly formatted content (indentation and heading spaces) (#17216)
- [x] ~Get articles edited~ Cancelled. Will do later in a different PR.

https://docs.microsoft.com/en-us/dotnet/framework/xaml-services/

PR #16629, opened 2020-01-10
Change references from .NET Framework to .NET (where appropriate)

Thraka:
## Summary

Changed a lot of references from **.NET Framework** to a generic **.NET** to make it apply to both framework and .NET Core. Further, this will make it apply to .NET 5 in the future.

Contributes to #13813


tdykstra (approved):
LGTM

docs/desktop-wpf/xaml-services/basic-reading-writing.md
```diff
@@ -8,7 +8,7 @@ ms.assetid: 6ac27fad-3687-4d7a-add1-3e90675fdfde
...
+<xref:System.Xaml.XamlServices> is a class provided by .NET that can be used to address XAML scenarios that don't require specific access to the XAML node stream or to XAML type system information obtained from those nodes. <xref:System.Xaml.XamlServices> API can be summarized as follows: `Load` or `Parse` to support a XAML load path, `Save` to support a XAML save path, and `Transform` to provide a technique that joins a load path and save path. `Transform` can be used to change from one XAML schema to another. This topic summarizes each of these API classifications and describes the differences between particular method overloads.
```
Youssef1313: Doesn't that imply that it's provided by all the .NET ecyosystem, which includes .NET Standard and Xamarin ?

The API docs <https://docs.microsoft.com/en-us/dotnet/api/system.xaml.xamlservices?view=netframework-4.8> state it's applied to .NET Core and .NET Framework, but doesn't mention .NET Standard or Xamarin

Thraka: It could be interpreted that way, but I don't think it needs to be. Saying something is from .NET doesn't necessarily mean it's supported by everything that implements .NET in some form. You'll still need to check any API for compatibility. .NET Standard is a specification set of APIs, by saying something is .NET compatible, you don't automatically imply it's supported by .NET Standard.

@mairaw What do you think?

PR #16648, opened 2020-01-13
Update Threading docs to de-emphasis Thread.Abort

### LINKED ISSUE #16616: Threading docs: de-emphasis Thread.Abort
The current documentation makes one think as if Thread.Abort is a preferred option to stop threads and mentions cooperative cancellation only as secondary choice. For example:

https://docs.microsoft.com/en-us/dotnet/standard/threading/using-threads-and-threading

>To terminate the execution of a thread, use the Thread.Abort method. [...] Beginning with the .NET Framework 4, you can use the System.Threading.CancellationToken to cancel a thread cooperatively.

https://docs.microsoft.com/en-us/dotnet/standard/threading/destroying-threads

>The Thread.Abort method is used to stop a managed thread permanently.

However, Thread.Abort is not supported on .NET Core, and in .NET Framework, even older versions, isn't really a primary way to go. This is **.NET guide**, so it must cover both .NET implementations. 

I think docs should be updated to put information like that:

- Cooperative cancellation as a primary scenario for .NET Core & newer .NET Framework: most people should use it 
- When cooperative cancellation is not possible (like running 3rd-party code): in .NET Framework, use Thread.Abort, in .NET Core, split that code into separate process.
- Older .NET Framework versions, where CancellationToken is not supported: manually craft cooperative cancellation with sync techniques, like volatile boolean field. (Is it needed? .NET Framework 3.5 is supported until 2028 so probably yes.)
---
#### Document Details

* ID: 0987cd07-0591-5ac9-536d-3b804f5332e2
* Version Independent ID: 208fb732-d9f4-8150-4675-befaa08a3a2c
* Content: [Using threads and threading](https://docs.microsoft.com/en-us/dotnet/standard/threading/using-threads-and-threading)
* Content Source: [docs/standard/threading/using-threads-and-threading.md](https://github.com/dotnet/docs/blob/master/docs/standard/threading/using-threads-and-threading.md)
* Product: **dotnet**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

MSDN-WhiteKnight:
## Summary

Mention cooperative cancellation model as a primary choice and reflect that Thread.Abort is not supported on .NET Core.

Fixes #16616

### DIFF
--- a/docs/standard/threading/destroying-threads.md
+++ b/docs/standard/threading/destroying-threads.md
@@ -11,8 +11,9 @@ helpviewer_keywords:
 ms.assetid: df54e648-c5d1-47c9-bd29-8e4438c1db6d
 ---
 # Destroying threads
-The <xref:System.Threading.Thread.Abort%2A?displayProperty=nameWithType> method is used to stop a managed thread permanently. When you call <xref:System.Threading.Thread.Abort%2A>, the common language runtime throws a <xref:System.Threading.ThreadAbortException> in the target thread, which the target thread can catch. For more information, see <xref:System.Threading.Thread.Abort%2A?displayProperty=nameWithType>.  
-  
+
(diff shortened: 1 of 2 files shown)


BillWagner (approved):
This is really well stated @MSDN-WhiteKnight 

I've reviewed everything, and I'll :shipit: now.

Thanks for helping us improve this area.

PR #16650, opened 2020-01-13
String interpolation: note boxing

### LINKED ISSUE #16618: String interpolation can cause boxing which can be avoided using ToString()
The following code causes boxing of the 'salary' value type. 

```
double salary = 1200;
string text = $"Salary {salary:C2}";
Console.WriteLine(text);
```

Boxing is evil because it is “slow”. It creates a reference type (the boxed value) which the garbage collector has to clean up (which is also “slow”).  You can easily do this by adding a ToString().
```
double salary = 1200;
string text = $"Salary {salary.ToString("C2")}";
Console.WriteLine(text);
```

I think this should be mentioned in the docs.

---
#### Document Details

* ID: d0665571-35b7-0841-5719-b69456b35cfb
* Version Independent ID: 0410420b-3b12-6d96-40cf-56f22ac19a1d
* Content: [$ - string interpolation - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated#feedback)
* Content Source: [docs/csharp/language-reference/tokens/interpolated.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/tokens/interpolated.md)
* Product: **dotnet-csharp**
* GitHub Login: @pkulikov
* Microsoft Alias: **wiwagn**

pkulikov:
Fixes #16618

### DIFF
--- a/docs/csharp/language-reference/tokens/interpolated.md
+++ b/docs/csharp/language-reference/tokens/interpolated.md
@@ -1,7 +1,7 @@
 ---
 title: "$ - string interpolation - C# reference"
 description: String interpolation provides a more readable and convenient syntax to format string output than traditional string composite formatting.
-ms.date: 09/02/2019
+ms.date: 01/13/2020
 f1_keywords: 
   - "$_CSharpKeyword"
   - "$"
@@ -85,6 +85,23 @@ If you are new to string interpolation, see the [String interpolation in C#](../
 If an interpolated string has the type `string`, it's typically transformed into a <xref:System.String.Format%2A?displayProperty=nameWithType> method call. The compiler may replace <xref:System.String.Format%2A?displayProperty=nameWithType> with <xref:System.String.Concat%2A?displayProperty=nameWithType> if the analyzed behavior would be equivalent to concatenation.
 
 If an interpolated string has the type <xref:System.IFormattable> or <xref:System.FormattableString>, the compiler generates a call to the <xref:System.Runtime.CompilerServices.FormattableStringFactory.Create%2A?displayProperty=nameWithType> method.
+
+> [!NOTE]
+> String interpolation can cause [boxing](../../programming-guide/types/boxing-and-unboxing.md), as it happens in the following example:
+>
+> ```csharp-interactive
+> double pi = Math.PI;
+> string line = $"The pi constant is {pi:F5}";
+> Console.WriteLine(line);  // output: The pi constant is 3.14159
+> ```
+>
+> If you need to avoid boxing, format an instance of a value type explicitly, as the following example shows:
+>
+> ```csharp-interactive
+> double pi = Math.PI;
+> string line = $"The pi constant is {pi.ToString("F5")}";
+> Console.WriteLine(line);  // output: The pi constant is 3.14159
+> ```
 
 ## C# language specification
 
--- a/docs/csharp/language-reference/tokens/interpolated.md
+++ b/docs/csharp/language-reference/tokens/interpolated.md
@@ -103,7 +103,7 @@ If an interpolated string has the type <xref:System.IFormattable> or <xref:Syste
 > Console.WriteLine(line);  // output: The pi constant is 3.14159
 > ```
 >
-> However, that also might result in heap allocations of the additional `string` instances. Benchmark your code to choose the implementation that fits your needs.
+> However, that also might result in heap allocations of the additional `string` instances. The performance difference between the two preceding examples depends on .NET implementation. Do benchmarking to choose the code that fits your needs.
 
 ## C# language specification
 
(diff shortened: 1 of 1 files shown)


docs/csharp/language-reference/tokens/interpolated.md
```diff
@@ -86,6 +86,23 @@ If an interpolated string has the type `string`, it's typically transformed into
...
+> [!NOTE]
+> String interpolation can cause [boxing](../../programming-guide/types/boxing-and-unboxing.md), as it happens in the following example:
+>
+> ```csharp-interactive
+> double pi = Math.PI;
+> string line = $"The pi constant is {pi:F5}";
+> Console.WriteLine(line);  // output: The pi constant is 3.14159
+> ```
+>
+> If you need to avoid boxing, format an instance of a value type explicitly, as the following example shows:
```
svick: Avoiding boxing is not a goal by itself, avoiding heap allocations is. And as I showed in https://github.com/dotnet/docs/issues/16618#issuecomment-573315069, using `ToString()` this way can result in *more* allocations. So I think I would add some hand-waving and say something like:

> If you need to avoid heap allocations caused by boxing, it can be advantageous to format an instance of a value type explicitly, as the following example shows:

Also, it could make sense to use a type where `ToString()` actually avoids allocations on a recent version of .Net Core. I believe any `enum` type would be suitable for this purpose:

* it's a value type,
* it doesn't implement `ISpanFormattable` (which is the interface that helps avoiding intermediate string allocations),
* it supports format strings.

pkulikov: @svick thanks; I've updated the note to provide more context

BillWagner:
I'd like to have @stephentoub weigh in on this PR. I like what you've written @pkulikov but I'm always nervous about low-level performance guidelines. The teams that work on these classes are always looking for opportunities to improve performance in these areas: even small improvements matter because of how often these constructs are used.

I'm glad that you've called out the need to benchmark these choices in your application before making a decision.

docs/csharp/language-reference/tokens/interpolated.md
```diff
@@ -86,6 +86,25 @@ If an interpolated string has the type `string`, it's typically transformed into
...
+> double pi = Math.PI;
+> string line = $"The pi constant is {pi:F5}";
+> Console.WriteLine(line);  // output: The pi constant is 3.14159
+> ```
+>
+> To avoid heap allocations caused by boxing, consider formatting an instance of a value type explicitly, as the following example shows:
+>
+> ```csharp-interactive
+> double pi = Math.PI;
+> string line = $"The pi constant is {pi.ToString("F5")}";
```
stephentoub: On .NET Core, this latter example is actually going to allocate more, not less, than the boxing version. String interpolation like this uses String.Format, and for primitive types like double, we can format them directly into the rented buffer we use as part of the String.Format.  So beyond the overall string that's being allocated, the first example allocates a boxed double, whereas the latter example allocates a string for a formatted double, and the latter is actually larger.  On 64-bit, the former allocates 104B and the latter allocates 120B.

If someone's interested in performance around interpolation, there are lots of subtleties to understand; I worry a note like this one will actually muddy the waters more than it helps.

My $.02.

pkulikov: @stephentoub indeed, @svick presented the performance difference in the [comment to the related issue](https://github.com/dotnet/docs/issues/16618#issuecomment-573315069) to show that that difference depends on .NET implementation.

> If someone's interested in performance around interpolation, there are lots of subtleties to understand; I worry a note like this one will actually muddy the waters more than it helps.

@stephentoub @BillWagner what about using that note only for its last sentence? That is, mention that both variants (boxing/no boxing) would allocate and the actual difference in performance depends on .NET implementation. Thus, do benchmarking if you're interested in performance.

If we leave the note in, I would like to make sure that it doesn't favor one of the ways.

Also, I've updated the text of the last paragraph to mention that actual difference in performance is dependent on .NET implementation.

pkulikov: Also, maybe it's good to show how difficult it is. Otherwise, the "no boxing always better" approach might be non-critically applied to the string interpolation usage.

stephentoub: My $.02:
.NET Core is the future, especially when someone is focused on this kind of perf difference.  I would suggest this PR just be closed. It's great to tell people to benchmark when they're looking at performance, but that's not specific to string interpolation.

pkulikov: @stephentoub thank you for clarifying comments
@BillWagner I'm fine with both closing (and closing related issue then) and merging this PR

BillWagner:
After the discussion, I am closing this PR.

I'll also close #16618, with further explanation as to why we're not updating the docs to address this concern.

PR #16665, opened 2020-01-14
Moved the default values table

pkulikov:
Moved the article from the *keywords* folder to the *builtin-types* folder where it looks to fit better. As for TOC, it's updated according to the trend of getting rid of the "Reference tables for types" node.

### DIFF
--- a/docs/csharp/language-reference/builtin-types/default-values.md
+++ b/docs/csharp/language-reference/builtin-types/default-values.md
@@ -1,25 +1,24 @@
 ---
-title: "Default values table - C# reference"
-description: Learn what are the default values of C# types.
+title: "Default values of C# types - C# reference"
 ms.date: 12/18/2019
 helpviewer_keywords: 
   - "default [C#]"
   - "parameterless constructor [C#]"
 ---
-# Default values table (C# reference)
+# Default values of C# types (C# reference)
 
 The following table shows the default values of C# types:
 
 |Type|Default value|
 |---------|------------------|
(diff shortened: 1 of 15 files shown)


docs/csharp/language-reference/builtin-types/default-values.md
```diff
@@ -1,25 +1,24 @@
 ---
-title: "Default values table - C# reference"
-description: Learn what are the default values of C# types.
```
Youssef1313: Why the description is deleted ?

pkulikov: It hasn't added anything to the title

BillWagner: I'd prefer adding back the description, and adding some context that helps someone who sees this page in search results. Something like: 

```yml
- description: Learn the default values of C# types such as bool, char, int, float, double and more.
```

BillWagner (changes requested):
I like this @pkulikov 

I have one suggestion to restore the `description` metadata, and provide information for those readers that see the description in search rersults.

Once that's updated, I'll :shipit:

docs/csharp/language-reference/builtin-types/default-values.md
pkulikov: @BillWagner thank you for the suggestion; I've updated the metadata

BillWagner (approved):

BillWagner:
Thanks @pkulikov 

PR #16677, opened 2020-01-15
Link to alternative polymorphic converter

tdykstra:


docs/standard/serialization/system-text-json-converters-how-to.md
```diff
@@ -276,6 +276,8 @@ The converter can deserialize JSON that was created by using the same converter
 ]
 ```
 
+The converter code in the preceding example reads and writes each property manually. An alternative is to call `Deserialize` or `Serialize` to do some of the work. For an example, see [this StackOverflow post](https://stackoverflow.com/a/59744873/12509023).
```
ahsonkhan: I don't know if we generally link to StackOverflow in docs. Otherwise, this is fine by me :)

ahsonkhan (approved):

PR #16703, opened 2020-01-15
Added Longjump interop limitations documentation.

ivdiazsa:
## Summary

As we investigated through .NET Runtime [Issue #1445](https://github.com/dotnet/runtime/issues/1445), we noticed there is currently no documentation describing the limitations of `longjmp` when dealing with interop exceptions on Unix. This PR adds this missing information.


docs/standard/exceptions/handling-com-interop-exceptions.md
```diff
@@ -22,6 +22,13 @@ Managed and unmanaged code can work together to handle exceptions. If a method t
   
  If a method fails in unmanaged code, an exception can be passed to a managed code segment. The topic [HRESULTS and Exceptions](../../../docs/framework/interop/how-to-map-hresults-and-exceptions.md) contains a table showing how HRESULTS map to runtime exception objects.
```
jkotas: This topic is about exception handling and COM. It is all Windows specific. This should be in the part of documentation that talks about platform-neutral behaviors. Maybe there should be a new top level file that talks about exception handling interop?

docs/standard/exceptions/handling-com-interop-exceptions.md
```diff
@@ -22,6 +22,13 @@ Managed and unmanaged code can work together to handle exceptions. If a method t
   
  If a method fails in unmanaged code, an exception can be passed to a managed code segment. The topic [HRESULTS and Exceptions](../../../docs/framework/interop/how-to-map-hresults-and-exceptions.md) contains a table showing how HRESULTS map to runtime exception objects.  
 
+## Longjump Behaviors
+Longjump interop is held to the same constraints as exception handling and is currently only supported on Windows.
```
jkotas: Nit: The documentation always talks about what the behavior is currently. I believe that we avoid saying "currently" in the documentation - it is implied.

docs/standard/exceptions/handling-com-interop-exceptions.md
```diff
@@ -22,6 +22,13 @@ Managed and unmanaged code can work together to handle exceptions. If a method t
...
+
+When unmanaged code moves on through a `longjmp` instruction, the runtime observes different behaviors between Windows and Unix operating systems. On Windows, there is support to unwind the stack properly when dealing with exceptions thrown by external components. On the other hand, Unix does not support exception interop as currently the Unix ABI has no definition for it. Therefore, exceptions thrown this way end up resulting in unpredictable behaviors and potential crashes.
```
jkotas: `longjmp` is not an instruction. It is a standard C library function.

docs/standard/exceptions/handling-com-interop-exceptions.md
jkotas: Now that I am thinking about it, we may declare setjmp/longjmp interop as unsupported everywhere, even on Windows. We do not seem to have any tests that it actually works, and there are likely number of cases where it does not work well even on Windows, given the long list of portability issues that setjmp/longjmp comes with.

docs/standard/exceptions/handling-com-interop-exceptions.md
ivdiazsa: I think we should create a new file in this case. I glanced over the doc files under _native-interop_ and all of them refer to COM interop and other things. On the other hand, the _exceptions_ directory has several how to's, but the only interop file is the one I modified here. Maybe we could name it something like `interop-exceptions-limitations` or `interop-exceptions-handling`?

jkotas: I like `interop-exceptions-handling`. Or make it just `interop-exceptions`

docs/standard/exceptions/handling-com-interop-exceptions.md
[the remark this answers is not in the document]
ivdiazsa: So, taking into account your thoughts from your last comment, we could draft the paragraphs like this:

- The first paragraph should only refer to specifying C++ exception interop is not officially supported due to the portability issues that arise since managed code can't know how the C++ exception mechanisms work under the covers.
- The second paragraph explains the `longjmp` case.

What do you think? How can we improve this?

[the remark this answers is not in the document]
jkotas: > C++ exception interop is not officially supported due to the portability issues

... on non-Windows platforms.

> officially supported

For the docs - omit "officially". If something is mentioned in the docs as supported, it implies that it is supported officially.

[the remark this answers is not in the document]
jkotas: Looks good otherwise.

You can also reference https://www.mono-project.com/docs/advanced/pinvoke/#runtime-exception-propagation (e.g. as a link at the end - check existing topics for how it is done) or borrow some of the ideas for how to explain this from it.

docs/standard/exceptions/handling-com-interop-exceptions.md
ivdiazsa: I just realized the current file is called `handling-com-interop-exceptions.md`. I think it might be a bit confusing to name the new one just `interop-exceptions`. Maybe something like `platform-neutral-interop-exceptions.md` would work?

ivdiazsa:
Rewriting of the doc in the new file is complete. I replaced "C++" with "unmanaged code" but I'm not sure whether it is 100% correct here.

docs/standard/exceptions/handling-com-interop-exceptions.md
jkotas: None of the platform-neutral docs under https://docs.microsoft.com/en-us/dotnet/standard/native-interop/ has platform-neutral prefix, so I do not think this one should  have it either.

docs/standard/exceptions/platform-neutral-interop-exceptions.md
```diff
@@ -0,0 +1,27 @@
...
+Longjump interop is held to the same constraints as exception handling and is only supported on Windows.
+
+When unmanaged code moves on through the `longjmp` C library function, the runtime observes different behaviors between Windows and Unix operating systems. On Windows, there is support to unwind the stack properly when dealing with exceptions thrown by external components. On the other hand, Unix does not support exception interop as currently the Unix ABI has no definition for it. Therefore, exceptions here end up resulting in unpredictable behaviors and potential crashes.
+
+For further information, take a look at `longjmp` [documentation](https://docs.microsoft.com/cpp/c-runtime-library/reference/longjmp).
+
+## See also
+
+- [Exceptions](index.md)
+- [Interop with Native Libraries](https://www.mono-project.com/docs/advanced/pinvoke)
```
jkotas:
```suggestion
- [Interop with Native Libraries](https://www.mono-project.com/docs/advanced/pinvoke/#runtime-exception-propagation)
```

docs/standard/exceptions/handling-com-interop-exceptions.md
ivdiazsa: Fair point. Will change it.

docs/standard/native-interop/interop-exceptions.md
```diff
@@ -0,0 +1,27 @@
+---
+title: "Platform Neutral Interop Exceptions"
```
jkotas:
```suggestion
title: "Exceptions interoperability"
```

jkotas: I believe "Interop Exceptions" would mean "exceptions that are meant to be used for interop". It is not what this doc is about.

docs/standard/native-interop/interop-exceptions.md
```diff
@@ -0,0 +1,27 @@
...
+ms.date: "01/16/2020"
+ms.technology: dotnet-standard
+helpviewer_keywords:
+  - "unmanaged code, exceptions"
+  - "exceptions, unmanaged code"
+  - "interop, exceptions"
+  - "exceptions, interop"
+  - "stack, jump"
+  - "jump, stack"
+description: Information the use of unmanaged code interop exceptions.
```
jkotas: The other files in this directory have either description that starts with "Learn how to ..." or do not have description at all. Match one of these? I think it would be fine to not have the description for this doc - the title describes it well enough.

docs/standard/native-interop/interop-exceptions.md
```diff
@@ -0,0 +1,27 @@
+---
+title: "Platform Neutral Interop Exceptions"
+ms.date: "01/16/2020"
+ms.technology: dotnet-standard
+helpviewer_keywords:
+  - "unmanaged code, exceptions"
+  - "exceptions, unmanaged code"
+  - "interop, exceptions"
+  - "exceptions, interop"
+  - "stack, jump"
```
jkotas: "stack jump" does not seem like a combination of keywords that somebody looking for this doc is going to use. Bing or Google does not show anything even remotely related when searching for "stack jump .NET".

docs/standard/native-interop/interop-exceptions.md
```diff
@@ -0,0 +1,27 @@
...
+  - "jump, stack"
+description: Information the use of unmanaged code interop exceptions.
+---
+# Working with Interop Exceptions in Unmanaged Code
+Unmanaged code exception interop is not supported due to the portability issues that arise, since managed code can't know how these exception mechanisms work under the covers in non-Windows platforms.
+
+## Longjump Behaviors
+Longjump interop is held to the same constraints as exception handling and is only supported on Windows.
```
jkotas: As I have mentioned in to other comment, I would document setjmp/longjump as unsupported on all platforms. I do not think we can declare as supported with confidence.

docs/standard/native-interop/interop-exceptions.md
ivdiazsa: I had misunderstood the concept this whole time. Thanks for the clarification!

ivdiazsa: Maybe we should also rename the file to match the title? To `exceptions-interoperability.md`?

jkotas: Looking at the files under https://github.com/dotnet/docs/tree/master/docs/standard/native-interop , there does not seem to be uniform file naming scheme.

I will leave it up to you to pick the filename. I do not have strong opinion.

docs/standard/native-interop/exceptions-interoperability.md
```diff
@@ -0,0 +1,24 @@
...
+title: "Exceptions Interoperability"
+ms.date: "01/16/2020"
+ms.technology: dotnet-standard
+helpviewer_keywords:
+  - "unmanaged code, exceptions"
+  - "exceptions, unmanaged code"
+  - "interop, exceptions"
+  - "exceptions, interop"
+---
+# Working with Interop Exceptions in Unmanaged Code
```
jkotas: We do support interop for exceptions on Windows. It is well-tested area and it is fully supported, e.g. as part of managed C++.

We do not support interop for exceptions on non-Windows.

We should declare setjmp/longjmp as unsupported everywhere. It is not tested well, and it is unlikely that it works 100% even though it happens to work for simple cases.

I am sorry about the confusion ... hope that this makes it clear.

ivdiazsa: Let me see if I got this correctly. In the first section, `Working with Interop Exceptions in Unmanaged Code`, I can add the Windows support at the beginning, making the line look something like _"Unmanaged code exception interop is not supported in non-Windows platforms, due to..."_.

Then, rephrase the second section to remove the section that states that on Windows there is support to unwind the stack and deal with external exceptions. Leave the Unix ABI not supporting section as is, as well as the unpredictable behaviors and potential crashes section.

Something like this?

jkotas: > Leave the Unix ABI not supporting section as is

Yes, but it should be moved into the "Exceptions chapter".

The section about setjmp/longjump should just say that it is not supported. It does need to explain in detail why it is not supported.

docs/standard/native-interop/exceptions-interoperability.md
```diff
@@ -0,0 +1,22 @@
...
+Unmanaged code exception interop is not supported in non-Windows platforms, due to the portability issues that arise, since managed code can't know how these exception mechanisms work under the covers. The Unix ABI has no definition for this and therefore, exceptions end up resulting in unpredictable behaviors and potential crashes.
```
jkotas: I would also explicitly mention: `Unmanaged code exception interop is supported on Windows platforms only.`

docs/standard/native-interop/exceptions-interoperability.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Setjmp/Longjmp Behaviors
+`Setjmp`/`Longjmp` interop is held to the same constraints as exception handling as it is not supported to skip over managed frames this way.
```
jkotas:
```suggestion
The interop with `setjmp`/`longjmp` C functions is not supported. It is not supported to use `longjmp` to skip over managed frames.
```

docs/standard/native-interop/exceptions-interoperability.md
ivdiazsa: Ok! Would this be added, or replacing `not supported in non-Windows platforms`? I ask because of the potential redundancy.

jkotas: It should be added.

jkotas: The doc should be explicit about both the behavior on Windows and the behavior on non-Windows.

docs/standard/native-interop/exceptions-interoperability.md
jkotas: Could you please apply this suggestion as well?

ivdiazsa: Sorry Jan, for some reason, it didn't show up in the email and I overlooked it when reviewing here. Let me do it.

jkotas:
Thanks! I believe that this says the correct thing now. I will let @tdykstra to hammer our the style.

tdykstra (changes requested):
Some suggestions for you to consider, plus this needs to be added to the table of contents, i.e. https://github.com/dotnet/docs/blob/master/docs/standard/toc.yml
in the native interop section add something like

```
  - name: Exceptions
    href: native-interop/exceptions-interoperability.md
```

tdykstra (approved):

PR #16710, opened 2020-01-16
Fix generic xrefs in visual basic async article

Youssef1313:
Contributes to #16709


mairaw (approved):
Good idea to use the VB syntax for those. Thanks @Youssef1313!

PR #16711, opened 2020-01-16
Fix code fencing

pkulikov:
- Code-fenced one occurrence of "object" where it refers to the type
- Removed code fencing where the "string" word refers to an instance


BillWagner (approved):

I'll :shipit: now.

PR #16733, opened 2020-01-17
remove extra slash in VS release notes URL

scottaddie:


mairaw (approved):
Thank you @scottaddie!

PR #16737, opened 2020-01-17
Add information on default interface members - added in C# 8.0

### LINKED ISSUE #12490: Language reference: Major updates to interface pages
Default interface members change several rules on what can be included in interfaces. It introduces new overload resolution scenarios. These need to be described on the appropriate language reference and programming guide pages.

### LINKED ISSUE #16061: Are static methods allowed in Interfaces nowadays ?
I've read in a post at stackflow at the following link:
https://stackoverflow.com/questions/259026/why-doesnt-c-sharp-allow-static-methods-to-implement-an-interface
and also at the following the link:
https://dotnetfiddle.net/Lrzy6y
It's written that in C# version 8.0, static methods are allowed in an interface. Please confirm, is it so? if so, then it needs to be mentioned in this article as well, otherwise, it could be misleading and needs to be updated of the current situation as well.

---
#### Document Details

* ID: d97db038-978f-fff3-4e68-110619dc86f2
* Version Independent ID: 3c933595-0ec1-2632-55c9-934af6e72f5d
* Content: [static modifier - C# Reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/static#feedback)
* Content Source: [docs/csharp/language-reference/keywords/static.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/keywords/static.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixes #12490 
Fixes #16061 

Interfaces can now contain implementations of some member types.

- Include information on default interface members and what new member types are allowed in interfaces.
- Rewrite the opening to match the new uses of interfaces.
- Refer to the correct section of the language standard.
- Add a reference to the C# 8.0 default interface members specification.

### DIFF
--- a/docs/csharp/language-reference/keywords/interface.md
+++ b/docs/csharp/language-reference/keywords/interface.md
@@ -34,7 +34,14 @@ These preceding member declarations typically do not contain a body. Beginning w
 - Member declarations using the explicit interface implementation syntax.
 - Explicit access modifiers (the default access is [`public`](access-modifiers.md)).
 
-Interfaces may not contain instance state. While static fields are now permitted, instance fields are not permitted in interfaces. [Instance auto-properties](../../programming-guide/classes-and-structs/auto-implemented-properties.md) are not supported in interfaces, as they would implicitly declare a hidden field.
+Interfaces may not contain instance state. While static fields are now permitted, instance fields are not permitted in interfaces. [Instance auto-properties](../../programming-guide/classes-and-structs/auto-implemented-properties.md) are not supported in interfaces, as they would implicitly declare a hidden field. This rule has a subtle effect on property declarations. In an interface declaration, the following does not declare an auto-implemented property as it does in a `class` or `struct`. Instead, it declares a property that doesn't have a default implementation, but must be implemented in any type that implements the interface:
+
+```csharp
+public interface INamed
+{
+  public string Name {get; set;}
+}
+```
 
 An interface can inherit from one or more base interfaces. When an interface [overrides a method](override.md) implemented in a base interface, it must use the explicit interface implementation syntax.
 
--- a/docs/csharp/language-reference/keywords/interface.md
+++ b/docs/csharp/language-reference/keywords/interface.md
@@ -29,7 +29,7 @@ These preceding member declarations typically do not contain a body. Beginning w
 - [Constants](const.md)
 - [Operators](../operators/operator-overloading.md)
 - [Static constructor](../../programming-guide/classes-and-structs/constructors.md#static-constructors).
-- [Nested types](../../programming-guide/classes-and-structs/nested-types)
+- [Nested types](../../programming-guide/classes-and-structs/nested-types.md)
 - [Static fields, methods, properties, indexers, and events](static.md)
 - Member declarations using the explicit interface implementation syntax.
 - Explicit access modifiers (the default access is [`public`](access-modifiers.md)).
--- a/docs/csharp/language-reference/keywords/interface.md
+++ b/docs/csharp/language-reference/keywords/interface.md
@@ -7,7 +7,7 @@ helpviewer_keywords:
   - "interface keyword [C#]"
 ms.assetid: 7da38e81-4f99-4bc5-b07d-c986b687eeba
 ---
-# interface (C# Reference)
+# :::no-loc text="interface"::: (C# Reference)
 
 An interface defines a contract. Any [`class`](class.md) or [`struct`](struct.md) that implements that contract must provide an implementation of the members defined in the interface. Beginning with C# 8.0, an interface may define a default implementation for members. It may also define [`static`](static.md) members in order to provide a single implementation for common functionality. In the following example, class `ImplementationClass` must implement a method named `SampleMethod` that has no parameters and returns `void`.
 
--- a/docs/csharp/language-reference/keywords/interface.md
+++ b/docs/csharp/language-reference/keywords/interface.md
@@ -1,6 +1,6 @@
 ---
 title: "interface - C# Reference"
-ms.date: 01/17/2019
+ms.date: 01/17/2020
 f1_keywords: 
   - "interface_CSharpKeyword"
 helpviewer_keywords: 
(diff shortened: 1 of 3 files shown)


docs/csharp/language-reference/keywords/interface.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: "interface - C# Reference"
-ms.date: 07/20/2015
+ms.date: 01/17/2019
```
Youssef1313:
```suggestion
ms.date: 01/17/2020
```

docs/csharp/language-reference/keywords/interface.md
```diff
@@ -9,31 +9,40 @@ ms.assetid: 7da38e81-4f99-4bc5-b07d-c986b687eeba
 ---
 # interface (C# Reference)
```
Youssef1313: Should "interface" be added to no-loc in the metadata?

pkulikov: It looks that H1 (or C# keywords) are treated in some special way, for example:
https://docs.microsoft.com/ru-ru/dotnet/csharp/language-reference/keywords/interface

Also note that the C# keywords are not translated in the TOC.
So, I guess some "magic" is applied to the C# keywords met in the titles (and, thus, in TOCs)

Youssef1313: Yes it seems that it doesn't get localized in the heading. But in the text under the H1 it gets localized, which I think not desirable.

pkulikov: If it's used as a keyword, it is usually code-fenced and not localized. However, when it's used as a term, it should be localized and not code-fenced.

pkulikov: So, maybe, in this particular case, the title should be localized as well.

BillWagner:
set the status back to work in progress. I think with a couple small edits, I can address three more issues related to language reference and DIM in this PR.

gewarren:
I'll remove my request until you're ready for review, if that's okay.

docs/csharp/language-reference/keywords/interface.md
BillWagner: Doh!

docs/csharp/language-reference/keywords/interface.md
BillWagner: This shouldn't be localized. I'll add it to the metadata.

BillWagner:
@gewarren I've finished the remaining updates. This should be ready.

docs/csharp/language-reference/keywords/static.md
```diff
@@ -10,48 +10,46 @@ ms.assetid: 5509e215-2183-4da3-bab4-6b7e607a4fdf
...
+Classes, interfaces, and static classes may have static constructors. Static constructors are called at some point between when the program starts and the class is instantiated.
 
 > [!NOTE]
 > The `static` keyword has more limited uses than in C++. To compare with the C++ keyword, see [Storage classes (C++)](/cpp/cpp/storage-classes-cpp#static).
 
-To demonstrate static members, consider a class that represents a company employee. Assume that the class contains a method to count employees and a field to store the number of employees. Both the method and the field do not belong to any instance employee. Instead they belong to the company class. Therefore, they should be declared as static members of the class.
+To demonstrate static members, consider a class that represents a company employee. Assume that the class contains a method to count employees and a field to store the number of employees. Both the method and the field don't belong to any instance employee. Instead they belong to the company class. They should be declared as `static` members of the class.
```
gewarren:
```suggestion
To demonstrate static members, consider a class that represents a company employee. Assume that the class contains a method to count employees and a field to store the number of employees. Both the method and the field don't belong to any one employee instance. Instead, they belong to the class of employees as a whole. They should be declared as `static` members of the class.
```

docs/csharp/language-reference/keywords/interface.md
```diff
@@ -1,39 +1,55 @@
...
+
+- [Constants](const.md)
+- [Operators](../operators/operator-overloading.md)
+- [Static constructor](../../programming-guide/classes-and-structs/constructors.md#static-constructors).
+- [Nested types](../../programming-guide/classes-and-structs/nested-types.md)
+- [Static fields, methods, properties, indexers, and events](static.md)
+- Member declarations using the explicit interface implementation syntax.
+- Explicit access modifiers (the default access is [`public`](access-modifiers.md)).
+
+Interfaces may not contain instance state. While static fields are now permitted, instance fields are not permitted in interfaces. [Instance auto-properties](../../programming-guide/classes-and-structs/auto-implemented-properties.md) are not supported in interfaces, as they would implicitly declare a hidden field. This rule has a subtle effect on property declarations. In an interface declaration, the following does not declare an auto-implemented property as it does in a `class` or `struct`. Instead, it declares a property that doesn't have a default implementation, but must be implemented in any type that implements the interface:
```
gewarren:
```suggestion
Interfaces may not contain instance state. While static fields are now permitted, instance fields are not permitted in interfaces. [Instance auto-properties](../../programming-guide/classes-and-structs/auto-implemented-properties.md) are not supported in interfaces, as they would implicitly declare a hidden field. This rule has a subtle effect on property declarations. In an interface declaration, the following code does not declare an auto-implemented property as it does in a `class` or `struct`. Instead, it declares a property that doesn't have a default implementation but must be implemented in any type that implements the interface:
```

gewarren (approved):
Looks good overall! Thanks for adding the clarification. I was confused :)

docs/csharp/language-reference/keywords/interface.md
```diff
@@ -1,39 +1,48 @@
...
 - [Methods](../../programming-guide/classes-and-structs/methods.md)
-
 - [Properties](../../programming-guide/classes-and-structs/using-properties.md)
-
 - [Indexers](../../programming-guide/indexers/using-indexers.md)
-
 - [Events](event.md)
 
-An interface can inherit from one or more base interfaces.
+These preceding member declarations typically do not contain a body. Beginning with C# 8.0, an interface may declare a body, called a *default implementation*. Members with bodies permit the interface to provide a "default" implementation for the method in classes and structs that do not provide an overriding implementation. In addition, beginning with C# 8.0, an interface may include:
```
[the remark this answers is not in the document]
BillWagner: I'd like to get your thoughts here.  Providing a body for events, indexers and properties is legal, but in practice it's not generally useful. Declaring an instance property almost always implies declaring an instance field, which isn't legal. 

I can mention it, but any example would be contrived.

docs/csharp/language-reference/keywords/static.md
```diff
@@ -10,48 +10,46 @@ ms.assetid: 5509e215-2183-4da3-bab4-6b7e607a4fdf
...
 
-To refer to the static member `x`, use the fully qualified name, `MyBaseC.MyStruct.x`, unless the member is accessible from the same scope:
+To refer to the `static` member `x`, use the fully qualified name, `MyBaseC.MyStruct.x`, unless the member is accessible from the same scope:
 
 ```csharp
 Console.WriteLine(MyBaseC.MyStruct.x);
 ```
 
-While an instance of a class contains a separate copy of all instance fields of the class, there is only one copy of each static field.
+While an instance of a class contains a separate copy of all instance fields of the class, there's only one copy of each static field.
```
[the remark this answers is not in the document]
BillWagner: My intent was that all instances would be code-fenced. I'll check again.

docs/csharp/language-reference/keywords/interface.md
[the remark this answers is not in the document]
gewarren: I think it's fine to leave it as-is then.

PR #16750, opened 2020-01-19
Code fence the null keyword

pkulikov:


BillWagner (approved):
Well, this is quick and easy to review @pkulikov 

I'll :shipit: now.

PR #16786, opened 2020-01-22
Added new inheritdoc xml documentation

mikadumont:
@kendrahavens @TerryGLee


docs/csharp/programming-guide/xmldoc/recommended-tags-for-documentation-comments.md
```diff
@@ -18,14 +18,14 @@ The C# compiler processes documentation comments in your code and formats them a
...
-|[\<c>](./code-inline.md)|[\<para>](./para.md)|[\<see>](./see.md)*|  
+|||||  
+|---|---|---|---|
+|[\<c>](./code-inline.md)|[\<para>](./para.md)|[\<see>](./see.md)*|[\<value>](./value.md)  
 |[\<code>](./code.md)|[\<param>](./param.md)*|[\<seealso>](./seealso.md)*|  
 |[\<example>](./example.md)|[\<paramref>](./paramref.md)|[\<summary>](./summary.md)|  
 |[\<exception>](./exception.md)*|[\<permission>](./permission.md)*|[\<typeparam>](./typeparam.md)*|  
 |[\<include>](./include.md)*|[\<remarks>](./remarks.md)|[\<typeparamref>](./typeparamref.md)|  
-|[\<list>](./list.md)|[\<returns>](./returns.md)|[\<value>](./value.md)|  
+|[\<list>](./list.md)|[\<inheritdoc>](./inheritdoc.md)|[\<returns>](./returns.md)|
```
Youssef1313: It was ordered alphabetically. Now It's not. The new inheritdoc should come just after the include one.

docs/csharp/programming-guide/xmldoc/inheritdoc.md
```diff
@@ -0,0 +1,32 @@
...
+## Syntax  
+  
+```xml  
+<inheritdoc/> 
+```  
+  
+## InheritDoc  
+Inherit XML comments from base classes, interfaces, and similar methods. This eliminates unwanted copying and pasting of duplicate XML comments and automatically keeps XML comments sychronized. 
+  
+## Remarks
```
Youssef1313:
```suggestion
## Remarks

```

docs/csharp/programming-guide/xmldoc/inheritdoc.md
```diff
@@ -0,0 +1,32 @@
...
+ms.assetid: 46d329b1-5b84-4537-9e17-73ca97313e4e
+---
+# \<inheritdoc> (C# Programming Guide)
+## Syntax  
+  
+```xml  
+<inheritdoc/> 
+```  
+  
+## InheritDoc
```
Youssef1313:
```suggestion
## InheritDoc

```

docs/csharp/programming-guide/xmldoc/inheritdoc.md
```diff
@@ -0,0 +1,32 @@
...
+ms.date: 01/21/2020
+f1_keywords: 
+  - "inheritdoc"
+  - "<inheritdoc>"
+helpviewer_keywords: 
+  - "<inheritdoc> C# XML tag"
+  - "inheritdoc C# XML tag"
+ms.assetid: 46d329b1-5b84-4537-9e17-73ca97313e4e
+---
+# \<inheritdoc> (C# Programming Guide)
```
Youssef1313:
```suggestion
# \<inheritdoc> (C# Programming Guide)

```

docs/csharp/programming-guide/xmldoc/inheritdoc.md
```diff
@@ -0,0 +1,32 @@
...
+<inheritdoc/> 
+```  
+  
+## InheritDoc  
+Inherit XML comments from base classes, interfaces, and similar methods. This eliminates unwanted copying and pasting of duplicate XML comments and automatically keeps XML comments sychronized. 
+  
+## Remarks  
+Add your XML comments in base classes and let InheritDoc copy the comments to implementing classes.
+
+Add your XML comments in interfaces and let InheritDoc copy the comments to implementing classes.
```
Youssef1313: Looks like a bit of unnecessary repetition here.

I would suggest:

"Add your XML comments in base classes or interfaces and let InheritDoc copy the comments to implementing classes."

docs/csharp/programming-guide/xmldoc/inheritdoc.md
```diff
@@ -0,0 +1,32 @@
...
+
+Add your XML comments to your synchronous methods and let InheritDoc copy the comments to your asynchronous versions of the same methods.
```
Youssef1313: As a reader, I'd ask what's the link between the synchronous method and its asynchronous version ?

Is that inferred by the method names ?
So, if I have a two methods called "Connect" and "ConnectAsync", the async one will take the comments from the synchronous one ? Or there is a special way to mark the "Connect" as the synchronous version of "ConnectAsync" ?
What if the methods are "MyConnect" and "ConnectAsync"? How would InheritDoc know that "ConnectAsync" is the async version of "MyConnect" ?

I think the documentation should somehow clarify those.

Youssef1313 (changes requested):

docs/csharp/programming-guide/xmldoc/inheritdoc.md
BillWagner: I think the intent here is for this scenario.  Consider this interface:

```csharp
public interface ISomeTaskThings
{
    Task<int> CountAsync();
}
```

The `inheritDoc` attribute would apply on this implementation:

```csharp
public class TaskThingsImpl : ISomeTaskThings
{
    public async Task<int> CountAsync()
    {
        // ...
        return result;
    }
}
```

The addition of the `async` modifier is allowed.

BillWagner (commented):
I'm excited to see this @mikadumont 

I'd like to see a bit more information, and possibly an example.  When I read through this, here are the questions it raised for me:

- I assume the `<inheritDoc>` is only used on the derived class. Is that correct?
- What about deep hierarchies? Suppose I have this hierarchy:
```csharp
public class B {}
public class D : B {}
public class S : D {}
```
If I add `inheritDoc` to `S`, would it pick up comments from `D`, `B`, or both? What if both `D` and `B` have comments on the same virtual / override method? Would the comments be merged?
- Are comments only inherited for virtual methods? What about methods overridden using the `new` modifier? What about constructors with matching signatures?
- Are comments added for explicit interface implementations?
- Would close generic types work as well? Consider this:
```csharp
public EmployeeDictionary : Dictionary<int, Employee>
{
   // ...
}
```
Would this type inherit docs from `Dictionary<TKey, TDictionary>`?

Again, overall, this is great. Once updated to anticipate these questions, it will be great.

docs/csharp/programming-guide/xmldoc/inheritdoc.md
mikadumont: More infor on inheritdoc is found here: https://www.inheritdoc.io/

mikadumont: @BillWagner I am not sure how to include an example or where the current examples live. I can provide a screenshot but couldnt figure out how to create a snippet.

mikadumont: @BillWagner to answer your questions below: 

- Deep Hierarchies: 
The <inheritdoc/> tag works for derived types but not deep hierarchies. There is another syntax for InheritDoc which I wasnt sure where to put <inheritdoc cref ="B"/>. This will work for deep hierarchies. 

image

- Method overrides and Virtual Methods: 
In general quick Info will now display XML comments for method overrides and virtual methods without InheritDoc so that doc should be updated but not sure exactly where to update that. 

image

image

- Explicit Interface Implementations:
This will work with the <inheritdoc cref=""/> syntax. 
image

- Close generic types:
Not sure about the close generic types. 

- The async example you provided below only works if you have the <inheritdoc cref=""/> tag:

image

BillWagner (approved):
This LGTM @mikadumont 

Once you merge dotnet/samples#1930, close and re-open PR to trigger a new build. That should make sure the snippets are included correctly (they look right to me). Then, you can :shipit:

PR #16798, opened 2020-01-23
char See also: added link to Rune API

pkulikov:
Inspired by #15845 (while there is no conceptual docs in place, at least the API page can be referenced)


BillWagner:
ping @tdykstra 

This looks fine to me. Would you prefer waiting until you publish the Rune articles?

tdykstra:
No reason to wait.  In fact it’s a good idea to go ahead with links like this in relevant APIs as they don’t take time to do and will remain relevant after the Rune documentation is published.

BillWagner (approved):
Thanks for adding this @pkulikov 

I'll :shipit: now.  Thanks @tdykstra

PR #16803, opened 2020-01-23
Fix typo.

TChatzigiannakis:
## Summary

Fixes typo that uses the word "binder" instead of "builder" (as the other examples do).


cartermp (approved):
Thanks!

PR #16820, opened 2020-01-24
Update XML documentation comments

Youssef1313:
When I tried running csc with -doc option, I got different outputs in some articles. This needs to be well-reviewed just in case the original output is correct and I'm going something wrong.

(Hide white-spaces for easier diff)

### DIFF
--- a/docs/csharp/programming-guide/xmldoc/summary.md
+++ b/docs/csharp/programming-guide/xmldoc/summary.md
@@ -1,92 +1,98 @@
 ---
-title: "<summary> - C# Programming Guide"
+title: "<summary> - C# programming guide"
 ms.date: 07/20/2015
-f1_keywords: 
+f1_keywords:
   - "<summary>"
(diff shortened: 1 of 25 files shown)


docs/csharp/programming-guide/xmldoc/summary.md
```diff
@@ -1,92 +1,98 @@
...
+The previous example produces the following XML file.
+
+```xml
+<?xml version="1.0"?>
+<doc>
+    <assembly>
+        <name>CRefTest</name>
+    </assembly>
+    <members>
+        <member name="T:A">
```
BillWagner: I'll merge this, because what you've done does fix the issue.

In the medium term, I'd rather see the sample updated to use more reasonable type and member names.

BillWagner (approved):
Thanks for making this updates @Youssef1313 

This does fix important issues.

I'll :shipit: now.

PR #16826, opened 2020-01-24
Add language identifiers and minor changes

Youssef1313:


tdykstra (changes requested):
Thanks for the corrections. Most of this LGTM, but I'm not inclined to go with "output" on inline code blocks since it's not a supported language identifier.

Youssef1313:
> 
> 
> Thanks for the corrections. Most of this LGTM, but I'm not inclined to go with "output" on inline code blocks since it's not a supported language identifier.

It's currently in use:

image

When an identifier isn't supported, the name will just display without syntax highlighting. I *think* this is okay.

docs/framework/wcf/samples/custom-service-host.md
```diff
@@ -139,35 +139,35 @@ public class SelfDescribingServiceHostFactory : ServiceHostFactory
...
-```  
+```xml
+<%@ServiceHost Service="Microsoft.ServiceModel.Samples.CalculatorService"
+               Factory="Microsoft.ServiceModel.Samples.SelfDescribingServiceHostFactory"
+               language=c# Debug="true" %>
+```
   
  Here we have added an additional `Factory` attribute to the `@ServiceHost` directive, and passed the CLR type name of our custom factory as the attribute’s value. When IIS or WAS receives a message for this service, the WCF hosting infrastructure first creates an instance of the ServiceHostFactory and then instantiate the service host itself by calling `ServiceHostFactory.CreateServiceHost()`.  
   
 ## Running the Sample
```
mairaw:
```suggestion
## Run the sample  
```

mairaw (approved):
Changes LGTM. We've adopted the output style in the past @tdykstra but happy to reconsider.

tdykstra:
>When an identifier isn't supported, the name will just display without syntax highlighting.

That's correct, and some writers have used "output" in some cases. But in other cases, "text" or "console" have been used instead and may be a better fit. In some cases it may be best to leave off the identifier. The team discussed this and didn't reach a consensus on a standard approach.  That being the case, I believe it would usually be best to leave whatever the writer used as-is except where there's an obvious error like a typo or a missing supported language identifier. In this PR the changes I'd prefer to see reverted are the ones in the System.Text.Json docs; "output" might work for console output but it doesn't seem to me ideal for labeling error messages returned by a library.

tdykstra (approved):
Looks good, thanks again for the corrections.

PR #16848, opened 2020-01-26
Fix Typo: Code --> Core

davza:
Article was referring to "ASP.NET Code web application project", updated to read "ASP.NET Core web application project"


mairaw (approved):
Thank you @davza for fixing that typo. Your contribution should be live within a day.

PR #16849, opened 2020-01-27
Fix zone pivot groups yaml mime header

jdanyow:
Workaround for a bug in the localization pipeline.
https://ceapex.visualstudio.com/Engineering/_workitems/edit/167502


Youssef1313:
I think the target branch should be **master** not **live**.

PR #16855, opened 2020-01-27
update zone pivot group yaml mime header

jdanyow:
workaround for a bug in the localization pipeline
https://ceapex.visualstudio.com/Engineering/_workitems/edit/167502

### DIFF
--- a/docs/zone-pivot-groups.yml
+++ b/docs/zone-pivot-groups.yml
@@ -1,5 +1,4 @@
-# YamlMime:ZonePivotGroups
-# GitHub repository = 'dotnet/docs'
+### YamlMime:ZonePivotGroups
 groups:
  - id: operating-systems-set-one
    title: Operating System


Thraka:
@jdanyow 

Did someone in the org approve this? We were guided to this specific format, including the `# GitHub repository = 'dotnet/docs'` line which you removed.

Additionally, why are we implementing workarounds instead of fixing the problem in the loc pipeline?

jdanyow:
@Thraka we hit this bug in the localization once before, back in December. When we discovered it I created [a work item for the localization team](https://ceapex.visualstudio.com/Engineering/_workitems/edit/156212). Until they've fixed the underlying issue, we need to normalize the yaml mime header in the content (this PR).

Who advised you to add `# GitHub repository = 'dotnet/docs'`? I do not know what that comment is for. It's not something we're using in Docs build or templating afaik.

Thraka:
@jdanyow Internal link: https://mseng.visualstudio.com/TechnicalContent/_wiki/wikis/Applied_AI_Docs_Wiki/11711/Zone-Pivots

This was (is still?) the guidance I used and it had that header in there.

bodhigautam:
@Thraka We have a plan to accommodate the # YamlMime syntax in the loc pipeline in future. 
Besides ### YamlMime is not a workaround, its a valid syntax being used across many repos for example [learn-pr](https://github.com/MicrosoftDocs/learn-pr/blob/master/learn-pr/zone-pivot-groups.yml#L1) , [azure-docs-pr](https://github.com/MicrosoftDocs/azure-docs-pr/blob/live/articles/mariadb/index.yml#L1) etc. 
Why we don't need repository name is because then it will be bound to a locale. For example here if we use # GitHub repository = 'dotnet/docs' then the zone pivots of 'dotnet/docs' will only be considered. On loc someone will have to manually add the locales like dotnet/docs.de-de , dotnet/docs.ja-jp which is not optimal for an automated loc pipeline. Hence its advised to not keep the repo names in the content files. cc @jdanyow

Thraka (approved):

PR #16856, opened 2020-01-27
Remove redundant path segments

Youssef1313:
This was done using search and replace.

Find: "../../../../docs/framework/unmanaged-api/debugging/"
Replace: Nothing
Directory: "\docs\framework\unmanaged-api\debugging"

The PR is bug, but I think the find and replace is straightforward.


mairaw (approved):
Thanks @Youssef1313. I agree that this change is pretty straightforward.

PR #16857, opened 2020-01-27
Add some info about error with deb packages

### LINKED ISSUE #16670: Add comment about feed instability due to deb/rpm package publish in "Troubleshoot the package manager"
We should mention that the feed becomes unstable when packages get published: what are the symptoms and how long are you expected to have to wait for it to resolve itself? (At which point you should file an issue on dotnet/core.)

Example on Ubuntu:

```
  File has unexpected size (72463 != 72485). Mirror sync in progress? [IP: 13.91.48.226 443]
```

I think this notice should be added to all Deb and RPM pages. I believe the delay is different between Deb and RPM.

Idea from @drk-mtr over at https://github.com/dotnet/core/issues/4116#issuecomment-574311268.

@leecow @vivmishra @dleeapho 

---
#### Document Details

* ID: 172f0b88-61f2-d631-478f-e88af9be1f7d
* Version Independent ID: c610685f-57a2-048d-8dfa-146e19943434
* Content: [Install .NET Core on Ubuntu 19.04 package manager - .NET Core](https://docs.microsoft.com/en-us/dotnet/core/install/linux-package-manager-ubuntu-1904)
* Content Source: [docs/core/install/linux-package-manager-ubuntu-1904.md](https://github.com/dotnet/docs/blob/master/docs/core/install/linux-package-manager-ubuntu-1904.md)
* Product: **dotnet-core**
* GitHub Login: @Thraka
* Microsoft Alias: **adegeo**

Thraka:
## Summary

- Added better troubleshooting headers.
- Added a note about the recent feed problems and possible reason.

@dagood I'm unsure if the **Unable to locate** error section should be ported to the debian articles. the steps don't exactly align to ubuntu (the user isn't told to download a `packages-microsoft-prod.deb` file).

Fixes #16670


docs/core/install/includes/package-manager-heading-hack-pkgname.md
```diff
@@ -26,6 +26,6 @@ The version of the SDK or runtime to install. This article will always give the
 - Install the ASP.NET Core 3.1 runtime: `aspnetcore-runtime-3.1`
 - Install the .NET Core 2.1 runtime: `dotnet-runtime-2.1`
 
-### Troubleshoot
+### Combination failed
```
dagood: "Combination" seems weird here to me, it makes me think of a combination of .NET Core, SDK, and ASP.NET Core packages not working, but it's just talking about one package name. This extends to `package combination` below.

Something like "Installing the custom package name fails" seems clearer to me.

docs/core/install/linux-package-manager-debian10.md
```diff
@@ -69,3 +69,11 @@ sudo apt-get install dotnet-runtime-3.1
...
 
 [!INCLUDE [package-manager-switcher](./includes/package-manager-heading-hack-pkgname.md)]
+
+## Troubleshoot the package manager
+
+This section provides information on common errors you may get while using the package manager to install .NET Core.
+
+### Failed to fetch
+
+While installing the .NET Core package, you may see an error similar to `Failed to fetch ... File has unexpected size ... Mirror sync in progress?`. In general, this error means that the package feed for .NET Core is being upgraded with newer versions and you should try again later. If you continually receive this error for some time, please file an issue at <https://github.com/dotnet/core/issues>.
```
dagood: Consider mentioning 30 minutes as an expected max "normal" temporary outage.

dagood:
The "Unable to locate" section is a workaround for a bug in `packages-microsoft-prod.deb`, so yeah, no need to port where it's not used.

dagood (changes requested):
This is needed for RPM package repos as well, 2 hour max expected outage there.

docs/core/install/linux-package-manager-debian9.md
```diff
@@ -69,3 +69,11 @@ sudo apt-get install dotnet-runtime-3.1
...
 
 [!INCLUDE [package-manager-switcher](./includes/package-manager-heading-hack-pkgname.md)]
+
+## Troubleshoot the package manager
+
+This section provides information on common errors you may get while using the package manager to install .NET Core.
+
+### Failed to fetch
+
+While installing the .NET Core package, you may see an error similar to `Failed to fetch ... File has unexpected size ... Mirror sync in progress?`. Generally speaking, this error means that the package feed for .NET Core is being upgraded with newer package versions, and that you should try again later. During an upgrade, the package feed should not be unavailable for more than 2 hours. If you continually receive this error for more than 2 hours, please file an issue at <https://github.com/dotnet/core/issues>.
```
dagood: 30 minutes for the Debian package repos (includes Debian and Ubuntu). The 2 hours is for RPM feeds like Fedora, openSUSE, SLES, and CentOS. (Not RHEL because those use Red Hat repos, not Microsoft ones.)

Thraka: Whoops. See, I don't fully understand this stuff yet. 👎  I'll fix that

docs/core/install/linux-package-manager-centos7.md
```diff
@@ -55,3 +55,11 @@ sudo yum install dotnet-runtime-3.1
...
 
 [!INCLUDE [package-manager-switcher](./includes/package-manager-heading-hack-pkgname.md)]
+
+## Troubleshoot the package manager
+
+This section provides information on common errors you may get while using the package manager to install .NET Core.
+
+### Failed to fetch
+
+While installing the .NET Core package, you may see an error similar to `Failed to fetch ... File has unexpected size ... Mirror sync in progress?`. Generally speaking, this error means that the package feed for .NET Core is being upgraded with newer package versions, and that you should try again later. During an upgrade, the package feed should not be unavailable for more than 2 hours. If you continually receive this error for more than 2 hours, please file an issue at <https://github.com/dotnet/core/issues>.
```
dagood: Hmm, maybe this should be in something we `INCLUDE`? I'm not familiar with the docs infra, just remembered it from the diff context. 🙂 This bit may change (without going away completely) as we figure out what we can do about this. (Also makes it easier to review.)

dagood: The error's actually different for RPM repository clients, sorry I missed adding this to the actual doc issue: https://github.com/dotnet/core/issues/4167.

I think this would work:

> similar to `Signature verification failed for file 'repomd.xml' from repository 'packages-microsoft-com-prod'`.

dagood (approved):
Thanks, one more tweak for the RPM repos, then LGTM.

docs/core/install/linux-package-manager-centos7.md
Thraka: Yes I think so. I started without the include because I thought it was only the ubuntu items 😁 I'll convert it.

Thraka:
@dagood OK I changed it up. I didn't mean to re-request since you're signed off.

dagood (approved):

docs/core/install/includes/package-manager-failed-to-fetch-deb.md
```diff
@@ -0,0 +1,2 @@
+
+While installing the .NET Core package, you may see an error similar to `Failed to fetch ... File has unexpected size ... Mirror sync in progress?`. Generally speaking, this error means that the package feed for .NET Core is being upgraded with newer package versions, and that you should try again later. During an upgrade, the package feed should not be unavailable for more than 30 minutes. If you continually receive this error for more than 30 minutes, please file an issue at <https://github.com/dotnet/core/issues>.
```
tdykstra: We usually avoid use of "please".

docs/core/install/includes/package-manager-failed-to-fetch-rpm.md
```diff
@@ -0,0 +1,2 @@
+
+While installing the .NET Core package, you may see an error similar to `signature verification failed for file 'repomd.xml' from repository 'packages-microsoft-com-prod'`. Generally speaking, this error means that the package feed for .NET Core is being upgraded with newer package versions, and that you should try again later. During an upgrade, the package feed should not be unavailable for more than 2 hours. If you continually receive this error for more than 2 hours, please file an issue at <https://github.com/dotnet/core/issues>.
```
tdykstra: ~~please~~

tdykstra (approved):
LGTM

docs/core/install/includes/package-manager-failed-to-fetch-deb.md
Thraka: @tdykstra  I know. Acro complains that it doesn't add anything. But in this case, I think it does. We're really requesting users to file an issue so that we can be alerted to the problem. Thoughts?

tdykstra: It's your call -- if you think it's justified in this instance I'm fine with it. I agree it would sound a bit abrupt here without it.

dagood: I agree it makes sense for this message. Style guide on [please](https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/p/please):

> Avoid please except in situations where the customer is asked to do something inconvenient or the application or site is to blame for the situation.
> 
> Example
> The network connection was lost. Please reenter your password.

The service is to blame, and it's inconvenient to have to file an issue to let us know.

PR #16865, opened 2020-01-28
Fix code snippets in console-webapiclient

alimon808:
- Remove semicolon from Repository class name property
- Remove return statement from ProcessRepository method
- Add repositories assignment prior to return statement in ProcessRepository method


BillWagner (approved):
Thanks for making these changes @alimon808 

I'll :shipit: now.

PR #16878, opened 2020-01-29
Remove info about Item property being required for collection seriali…

miloszkukla:
Tested on .Net Core 2.2 and serialization and deserialization of ICollection or IEnumerable implementing type works without indexer being defined.


tdykstra:
@StephenMolloy @HongGit Please review or recommend who should review.

StephenMolloy:
@mconnew ?

mconnew:
The code is really difficult to grok but I believe that on .NET Framework, when using the legacy code generation based serializer (which is always used when serializing using soap encoding on .NET Framework), it will use the Indexer property to assign values to the collection. I believe the existing documentation is correct.

tdykstra:
> I believe the existing documentation is correct.

Closing this PR

PR #16881, opened 2020-01-29
Add more info on working with categorical data

### LINKED ISSUE #13590: Need clarity on concept of key type, including info on using Hashing to create the key type
@nicolehaugen commented on [Fri Jul 26 2019](https://github.com/dotnet/machinelearning/issues/4041)

When learning about machine learning in general, I didn't see reference to a 'key type' outside of ML.NET - as a result, this concept was difficult for me to understand from the current description provided in the docs.

It would be better to provide an example that clarifies how and why a key type is needed.  The example that someone shared with me and that I found to be helpful, is this:
* When you use MapValueToKey, it "learns" the keys based on the training set.
* Using the U.S. States as an example, it "learns" that ND maps to 1, MN to 2, SD to 3, etc.
* And, when you see a new value in the test set that isn't in the training set, it will map it to an "unknown" key (e.g. 0)

Another way to make a KeyType is to use Hashing.  This should also be mentioned and explained in the docs alongside the description of KeyTypes.

[AB#1596630](https://dev.azure.com/mseng/677da0fb-b067-4f77-b89b-f32c12bb8617/_workitems/edit/1596630)

[AB#1668720](https://dev.azure.com/mseng/677da0fb-b067-4f77-b89b-f32c12bb8617/_workitems/edit/1668720)

natke:
Fixes #13590 

[AB#1668720](https://dev.azure.com/mseng/677da0fb-b067-4f77-b89b-f32c12bb8617/_workitems/edit/1668720)


docs/machine-learning/how-to-guides/prepare-data-ml-net.md
```diff
@@ -182,58 +182,38 @@ The result of binning creates bin bounds of `[0,200000,Infinity]`. Therefore the
 
 ## Work with categorical data
 
-Non-numeric categorical data needs to be converted to a number before being used to build a machine learning model.
+One of the most common types of data is categorical data. Categorical data has a finite number of categories. For example, the states of the USA, or a list of the types of animals found in a set of pictures. Whether the categorical data are features or labels, they must be mapped onto a numerical value in so that they can be used to generate a machine learning model. There are a number of ways of working with categorical data in ML.NET, depending on the problem you are solving.
```
luisquintanilla:
```suggestion
One of the most common types of data is categorical data. Categorical data has a finite number of categories. For example, the states of the USA, or a list of the types of animals found in a set of pictures. Whether the categorical data are features or labels, they must be mapped onto a numerical value so they can be used to generate a machine learning model. There are a number of ways of working with categorical data in ML.NET, depending on the problem you are solving.
```

docs/machine-learning/how-to-guides/prepare-data-ml-net.md
```diff
@@ -182,58 +182,38 @@ The result of binning creates bin bounds of `[0,200000,Infinity]`. Therefore the
...
-        VehicleType="Sedan"
-    },
-    new CarData
-    {
-        Color="Black",
-        VehicleType="SUV"
-    }
-};
-```
+In ML.NET, a key is an integer value representing a category. Key value mapping is most often used to map string labels into unique integer values for training, then back to their string values when the model is used to make a prediction.
```
luisquintanilla:
```suggestion
In ML.NET, a key is an integer value that represents a category. Key value mapping is most often used to map string labels into unique integer values for training, then back to their string values when the model is used to make a prediction.
```

docs/machine-learning/how-to-guides/prepare-data-ml-net.md
```diff
@@ -182,58 +182,38 @@ The result of binning creates bin bounds of `[0,200000,Infinity]`. Therefore the
...
-IDataView transformedData = categoricalTransformer.Transform(data);
-```
+One hot encoding takes a finite set of values and maps them onto integers whose binary representation has a single `1` value in unique positions in the string. One hot encoding can be the best choice if there is no implicit ordering of the categorical data. The following table shows an example with zip codes as raw values.
 
-The resulting transform converts the text value of `VehicleType` to a number. The entries in the `VehicleType` column become the following when the transform is applied:
+|Raw value|One hot encoded value|
+|---------|---------------------|
+|98052|00...01|
+|98100|00...10|
+|||
```
luisquintanilla:
```suggestion
```

docs/machine-learning/how-to-guides/prepare-data-ml-net.md
```diff
@@ -182,58 +182,38 @@ The result of binning creates bin bounds of `[0,200000,Infinity]`. Therefore the
...
-    1, // SUV
-    2, // Sedan
-    1 // SUV
-]
-```
+The transform to convert categorical data to one-hot encoded numbers is [`OneHotEncoding`](xref:Microsoft.ML.CategoricalCatalog.OneHotEncoding%2A).
+
+### Hashing
+
+Hashing is another way to convert categorical data to numbers. A hash function maps data of an arbitrary size (a string of text for example) onto a number with fixed range. Hashing can be a fast and space-efficient way of vectorizing features. One notable example of hashing in machine learning is email spam filtering where, instead of maintaining a dictionary of known words, every word in the email is hashed and added to a large feature vector. Using hashing in this way avoids the problem of malicious spam filtering circumvention by the use of words that are not in the dictionary.
```
luisquintanilla:
```suggestion
Hashing is another way to convert categorical data to numbers. A hash function maps data of an arbitrary size (a string of text for example) onto a number with a fixed range. Hashing can be a fast and space-efficient way of vectorizing features. One notable example of hashing in machine learning is email spam filtering where, instead of maintaining a dictionary of known words, every word in the email is hashed and added to a large feature vector. Using hashing in this way avoids the problem of malicious spam filtering circumvention by the use of words that are not in the dictionary.
```

docs/machine-learning/how-to-guides/prepare-data-ml-net.md
```diff
@@ -253,7 +233,7 @@ ReviewData[] reviews = new ReviewData[]
 };
 ```
 
-The minimum step to convert text to a numerical vector representation is to use the [`FeaturizeText`](xref:Microsoft.ML.TextCatalog.FeaturizeText*) method. By using the [`FeaturizeText`](xref:Microsoft.ML.TextCatalog.FeaturizeText*) transform, a series of transformations is applied to the input text column resulting in a numerical vector representing the lp-normalized word and character ngrams.
+ML.NET provides the [`FeaturizeText`](xref:Microsoft.ML.TextCatalog.FeaturizeText%2A) transform that takes a texts string and creates a set of features from the text, by applying a series of individual transforms.
```
luisquintanilla:
```suggestion
ML.NET provides the [`FeaturizeText`](xref:Microsoft.ML.TextCatalog.FeaturizeText%2A) transform that takes a text's string value and creates a set of features from the text, by applying a series of individual transforms.
```

docs/machine-learning/how-to-guides/prepare-data-ml-net.md
```diff
@@ -18,7 +18,7 @@ Data is often unclean and sparse. ML.NET machine learning algorithms expect inpu
 
 Sometimes, not all data in a dataset is relevant for analysis. An approach to remove irrelevant data is filtering. The [`DataOperationsCatalog`](xref:Microsoft.ML.DataOperationsCatalog) contains a set of filter operations that take in an [`IDataView`](xref:Microsoft.ML.IDataView) containing all of the data and return an [IDataView](xref:Microsoft.ML.IDataView) containing only the data points of interest. It's important to note that because filter operations are not an [`IEstimator`](xref:Microsoft.ML.IEstimator%601) or [`ITransformer`](xref:Microsoft.ML.ITransformer) like those in the [`TransformsCatalog`](xref:Microsoft.ML.TransformsCatalog), they cannot be included as part of an [`EstimatorChain`](xref:Microsoft.ML.Data.EstimatorChain%601) or [`TransformerChain`](xref:Microsoft.ML.Data.TransformerChain%601) data preparation pipeline.
 
-Using the following input data which is loaded into an [`IDataView`](xref:Microsoft.ML.IDataView):
+Using the following input data and load it into an [`IDataView`](xref:Microsoft.ML.IDataView):
```
luisquintanilla: I'd fix the tenses to match here. Using / load. We got feedback on this before though if we said "Use the following input data and load it into an IDataView" confused users because the code to load the data into an IDataView is not there. So it was worded as it currently is to make it sound like..."This is what the data looks like and we assume it's been loaded into an IDataView". The reason for not including the load code is we don't want to focus on loading code one way or another (file / enumerable). We want to focus more on the input data / transforms and resulting output.

luisquintanilla (commented):
Nicely done. Few comments and suggestions.

docs/machine-learning/how-to-guides/prepare-data-ml-net.md
natke: Ok sure that makes sense. Let me have another look at the wording

natke: I've made them consistent and explicitly instructed to load into a variable called `data`. Let me know what you think

luisquintanilla (approved):
Thanks for making the changes. Looks good to me. The only thing I'd change is the unresolved suggestion in the One Hot Encoding table. There seems to be an extra blank row. If the intent is to show that there are other values in between I'd replace the blank column values with "...". Otherwise, I'd remove the extra row.

docs/machine-learning/how-to-guides/prepare-data-ml-net.md
luisquintanilla: Looks good

natke:
Thanks! I missed that. Will update.

luisquintanilla:
:shipit:

PR #16970, opened 2020-02-04
Moved and revised the built-in types table

pkulikov:


docs/csharp/language-reference/builtin-types/built-in-types.md
```diff
@@ -0,0 +1,48 @@
...
+  - "built-in C# types"
+ms.assetid: 54f901f2-bf2f-472c-ae8d-73e8ecfc57fe
+---
+# Built-in types (C# reference)
+
+The following table lists the C# built-in [value](value-types.md) types:
+
+|C# type keyword|.NET type|
+|--------------|-------------------------|
+|[bool](bool.md)|<xref:System.Boolean?displayProperty=nameWithType>|
```
BillWagner: We should code-fence all the reserved words, I think:

```suggestion
|[`bool`](bool.md)|<xref:System.Boolean?displayProperty=nameWithType>|
```

docs/csharp/language-reference/builtin-types/built-in-types.md
```diff
@@ -0,0 +1,48 @@
...
+|[long](integral-numeric-types.md)|<xref:System.Int64?displayProperty=nameWithType>|
+|[ulong](integral-numeric-types.md)|<xref:System.UInt64?displayProperty=nameWithType>|
+|[short](integral-numeric-types.md)|<xref:System.Int16?displayProperty=nameWithType>|
+|[ushort](integral-numeric-types.md)|<xref:System.UInt16?displayProperty=nameWithType>|
+
+The following table lists the C# built-in [reference](../keywords/reference-types.md) types:
+
+|C# type keyword|.NET type|
+|--------------|-------------------------|
+|[object](reference-types.md#the-object-type)|<xref:System.Object?displayProperty=nameWithType>|
```
BillWagner: Same (adding comment because of two tables)

```suggestion
|[`object`](reference-types.md#the-object-type)|<xref:System.Object?displayProperty=nameWithType>|
```

BillWagner (commented):
This looks great @pkulikov 

I think that we should code fence the reserved words. What are your thoughts?

(commenting rather than requesting changes so this stays in my queue.)

pkulikov:
> I think that we should code fence the reserved words. What are your thoughts?

@BillWagner I agree. However, I'm afraid code-fencing is removed from the link text, as for now. Maybe, it'll change in future. So, I've pushed the update with code-fencing. How does it look now?

BillWagner (approved):
Thanks for making the update. You are correct, the site still loses the code formatting.

I'll :shipit: now, and see what we can do with the site engineering teams.

PR #16980, opened 2020-02-05
Rename nonbacktracking subexpression to atomic group

danmoseley:
We currently call the `(?>...)` construction a "Nonbacktracking (or "greedy") subexpression."

As pointed out by @stephentoub , it is a use of greedy that is not standard. The standard term seems to be atomic group, with some variation eg atomic subexpression, once-only subpattern (PHP) etc.

Changed to atomic group throughout with acknowledgement in one location of the alternative terms.

BTW it feels like subexpression and nonbacktracking ought to have hyphens, but I'm not changing that here.

@stephentoub


danmosemsft:
PHP - once only subpattern: https://www.php.net/manual/en/regexp.reference.onlyonce.php
Regexegg: atomic group https://www.rexegg.com/regex-disambiguation.html
Regularexpressions-info: atomic group https://www.regular-expressions.info/atomic.html
Boost : independent sub-expressions https://www.boost.org/doc/libs/1_44_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html
RE2: possessive match (but given its nature doesn't need them) https://github.com/google/re2/wiki/Syntax
Ruby: atomic group: https://ruby-doc.org/core-2.7.0/Regexp.html
PCRE: atomic group http://www.pcre.org/pcre2.txt

stephentoub:
cc: @cyrusn

CyrusNajmabadi:
thnks! i can make the roslyn change when this goes in.

docs/standard/base-types/backtracking-in-regular-expressions.md
```diff
@@ -112,8 +112,8 @@ Backtracking occurs when a regular expression pattern contains optional [quantif
  [!code-csharp[System.Text.RegularExpressions.Regex.ctor#1](../../../samples/snippets/csharp/VS_Snippets_CLR_System/system.text.regularexpressions.regex.ctor/cs/ctor1.cs#1)]
  [!code-vb[System.Text.RegularExpressions.Regex.ctor#1](../../../samples/snippets/visualbasic/VS_Snippets_CLR_System/system.text.regularexpressions.regex.ctor/vb/ctor1.vb#1)]  
 
-### Nonbacktracking Subexpression  
- The `(?>` *subexpression*`)` language element suppresses backtracking in a subexpression. It is useful for preventing the performance problems associated with failed matches.  
+### Atomic groups
+ The `(?>` *subexpression*`)` language element suppresses backtracking in a subexpression. This can be useful for preventing the performance problems associated with failed matches.
```
stephentoub: Not part of your change, but I think the text as it was written is misleading.  Saying "suppresses backtracking in a subexpression" makes it sound (at least to me) like there won't be any backtracking for any constructs in the subexpression... and that's not true.  Atomic groups prevent the subexpression from giving up anything that was matched to anything that comes after it, meaning it prevents backtracking _into_ the subexpression.  For example, if you have `(?>\w*\d*)1`, there may be backtracking as part of the `\d` failing to match, but once the `\w*\d*` has successfully matched and we exit the atomic group, if the `1` fails to match, we won't then re-enter the atomic group to backtrack into the `\d*` or the `\w*`.  In effect, the atomic group puts up a barrier around it that causes any attempts at backtracking to skip over it.
```suggestion
 The `(?>` *subexpression*`)` language element suppresses backtracking into a subexpression. This can be useful for preventing the performance problems associated with failed matches.  
```

docs/standard/base-types/best-practices.md
```diff
@@ -170,7 +170,7 @@ Often, applications pay a performance penalty for using backtracking despite the
...
 
-If you determine that backtracking is not necessary, you can disable it by using the `(?>subexpression)` language element. The following example parses an input string by using two regular expressions. The first, `\b\p{Lu}\w*\b`, relies on backtracking. The second, `\b\p{Lu}(?>\w*)\b`, disables backtracking. As the output from the example shows, they both produce the same result.
+If you determine that backtracking is not necessary, you can disable it by using the `(?>subexpression)` language element, known as an atomic group. The following example parses an input string by using two regular expressions. The first, `\b\p{Lu}\w*\b`, relies on backtracking. The second, `\b\p{Lu}(?>\w*)\b`, disables backtracking. As the output from the example shows, they both produce the same result.
```
stephentoub: This is fine and not worth changing.  But fwiw, in .NET 5 we will now automatically convert the former expression to latter.

docs/standard/base-types/details-of-regular-expression-behavior.md
```diff
@@ -100,7 +100,7 @@ The .NET Framework regular expression engine is a backtracking regular expressio
...
+- Atomic groups: `(?>`*subexpression*`)`. This feature allows the backtracking engine to guarantee that a subexpression matches only the first match found for that subexpression, as if the expression were running independent of its containing expression. If you do not use this construct, backtracking searches from the larger expression can change the behavior of a subexpression. For example, the regular expression `(a+)\w` matches one or more "a" characters, along with a word character that follows the sequence of "a" characters, and assigns the sequence of "a" characters to the first capturing group, However, if the final character of the input string is also an "a", it is matched by the `\w` language element and is not included in the captured group.
```
stephentoub:
```suggestion
- Atomic groups: `(?>`*subexpression*`)`. This feature allows the backtracking engine to guarantee that a subexpression matches only the first match found for that subexpression, as if the expression were running independent of its containing expression. If you do not use this construct, backtracking searches from the larger expression can change the behavior of a subexpression. For example, the regular expression `(a+)\w` matches one or more "a" characters, along with a word character that follows the sequence of "a" characters, and assigns the sequence of "a" characters to the first capturing group. However, if the final character of the input string is also an "a", it is matched by the `\w` language element and is not included in the captured group.
```

stephentoub (approved):

stephentoub:
Thanks for updating this, @danmosemsft.

docs/standard/base-types/best-practices.md
danmosemsft: I agree that it is not useful to attempt to describe in the documentation when our implementation might optimize their pattern anyway.

docs/standard/base-types/backtracking-in-regular-expressions.md
danmosemsft: changed to
```
### Atomic groups
 The `(?>` *subexpression*`)` language element suppresses backtracking into the subexpression. Once it has successfully matched, it will not give up any part of its match to subsequent backtracking. For example, in the pattern `(?>\w*\d*)1`, if the `1` cannot be matched, the `\d*` will not give up any of its match even if that means it would allow the `1` to successfully match. Atomic groups can help prevent the performance problems associated with failed matches.
```

danmosemsft:
Feedback addressed

danmosemsft:
@Thraka does this look good to merge?

Thraka (approved):
LGTM!

PR #16984, opened 2020-02-05
Update upgrade-to-nullable-references.md

SimpleSamples:
## Summary

It helps to be specific about where to put the element.


BillWagner (approved):
Good catch @SimpleSamples 

I'll :shipit: now.

PR #16999, opened 2020-02-06
Update using-directive.md

doterik:
In this type of examples where errors are generated, I think it would be better to clearly indicate that this is not an example you just should copy and use in your program.

I have seen many of these and if you just browse the page and see this snippet, with a handy copy button, you very well could think: "_Aha.. that is also a way to do it._"

I know that the text points out that this is wrong, but who reads manuals 😎

Often the code snippet includes comments of errors but not always, and as I said this is just one example.

And of course, this is only my personal point of view.

## Summary

Added a clarifying comment about the compiler error.

### DIFF
--- a/docs/csharp/language-reference/keywords/using-directive.md
+++ b/docs/csharp/language-reference/keywords/using-directive.md
@@ -62,7 +62,7 @@ Create a `using` alias directive to make it easier to qualify an identifier to a
 
 ```csharp
 using s = System.Text;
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks for making this change @doterik 

I like your justification. We've chosen to put those descriptions in text so they are localized. From this PR, I'm learning that we should do both, so the comment is in the code as well as in localized text.

PR #17001, opened 2020-02-06
Revert "fix casing of curl ssl option"

am11:
Reverts dotnet/docs#16576

`-sSL` option (with single hyphen) is the short form of `-s, --silent`, `-S, --show-error`, `-L, --location`, so cURL can follow redirects. Not to be confused with `--ssl`, which only tells the lib to "try SSL"; we do not need to specify it with HTTPS URLs.

Reference manual: <kbd>**[cURL(1)](https://linux.die.net/man/1/curl)**</kbd>


scottaddie (approved):

scottaddie:
Thank you for bringing this to our attention, @am11!

PR #17012, opened 2020-02-07
Update coding-conventions.md

WindOfMind:
## Summary

My proposal is to recommend to use an explicit type in foreach loops as, currently, it clashes with the rule: Do not use var when the type is not apparent from the right side of the assignment.
Using var in foreach loops makes it harder to read as often it's not clear what the type of an element is. Especially, when it concerns code reviews, for quick understanding and checking what is going on a loop it's utterly nice to know the type. Tools, such as VS Code, Microsoft VS, can help with that, but even in that case, you will have to spend more time using IntelliSense (move your cursor and so on), it's faster to read code without any popping up hints.

### DIFF
--- a/docs/csharp/programming-guide/inside-a-program/coding-conventions.md
+++ b/docs/csharp/programming-guide/inside-a-program/coding-conventions.md
@@ -88,15 +88,12 @@ ms.assetid: f4f60de9-d49b-4fb6-bab1-20e19ea24710
   
 - Avoid the use of `var` in place of [dynamic](../../language-reference/builtin-types/reference-types.md).  
   
-- Use implicit typing to determine the type of the loop variable in [for](../../language-reference/keywords/for.md) and [foreach](../../language-reference/keywords/foreach-in.md) loops.  
+- Use implicit typing to determine the type of the loop variable in [for](../../language-reference/keywords/for.md) loops.  
   
      The following example uses implicit typing in a `for` statement.  
   
      [!code-csharp[csProgGuideCodingConventions#7](../../../../samples/snippets/csharp/VS_Snippets_VBCSharp/csprogguidecodingconventions/cs/program.cs#7)]  
-  
-     The following example uses implicit typing in a `foreach` statement.  
-  
-     [!code-csharp[csProgGuideCodingConventions#12](../../../../samples/snippets/csharp/VS_Snippets_VBCSharp/csprogguidecodingconventions/cs/program.cs#12)]  
+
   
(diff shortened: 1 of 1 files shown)


BillWagner:
@WindOfMind I'm reviewing this, and I'm a bit confused. Your description says you'd want to use explicit typing in `for` and `foreach` loops. But, the text you've changed recommends using *implicit* typing, which would recommend `var`. Which did you mean?

WindOfMind:
@BillWagner Hi. I am sorry that my description is confusing. I suggested not to use `var` in `foreach` loops as typically it's unclear what type of an element is. For example, `foreach (var item in items)` looks more obscure rather than `foreach (Item item in items)` or `foreach (string item in items)`. Regarding `for` loops, usually, it's easy to quickly understand a type from an initial value, like `for (var 
 i = 0; i < 10; i++)`, so here, IMO, it make sense to leave a `var`.
Maybe it's even worth to point the next:
```
Do not use implicit typing to determine the type of the loop variable in foreach loops.
The following example uses implicit typing in a for statement.
...
```
If you agree, I can add it to my PR.

BillWagner:
@WindOfMind I think that change is fine, with one caveat:  when using explicit typing, you must be careful not to accidentally change the type. Good examples are changing an IQueryable to an IEnumerable, which changes the execution of a query.

WindOfMind:
@BillWagner, good point! I hope that in most cases code analyzers will help, at least R# highlights it by default.  Can we add it as a note? Like:
```
> [!NOTE]
> Be careful not to accidentally change a type of an element of the iterable collection. For example,  it is easy to switch from <xref:System.Linq.IQueryable?displayProperty=nameWithType> to <xref:System.Collections.Generic.IEnumerable?displayProperty=nameWithType> in `foreach` statement, which changes the execution of a query.
```
Feel free to correct the text of the note.

If you agree, let me know, I will add the new point and the note to my PR.

BillWagner:
I like the text above @WindOfMind 

Once you add that, I'll give it a final review and merge this.

WindOfMind:
@BillWagner I've updated my PR. Also, I've created a PR for updating the corresponding code snippet. You can find it here https://github.com/dotnet/samples/pull/1994.
Feel free to make any improvements in my PR.

BillWagner (approved):
Thanks @WindOfMind 

I'll merge this now.

WindOfMind:
@BillWagner You're welcome. It was a pleasure working with you.

PR #17033, opened 2020-02-10
Updated information about mixing integral and floating-point types

pkulikov:
- Be more elaborate about how integral and floating-point types can be mixed in an expression and how not.
- Remove the paragraph about the special values because those are mentioned earlier in the article.


BillWagner (approved):
This all LGTM @pkulikov 

I'll :shipit: now.

PR #17035, opened 2020-02-10
Clarify when the parallel loop is terminated by an exception

### LINKED ISSUE #17015: "Exceptions in Parallel Loops" is misleading
**Issue description**

[How to: Handle Exceptions in Parallel Loops](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/how-to-handle-exceptions-in-parallel-loops) has a misleading description of behavior when exceptions are thrown.

> The Parallel.For and Parallel.ForEach overloads ... resemble regular for and foreach loops ... an unhandled exception causes the **loop** to terminate immediately.

This is confusing. In this context, it seems like **loop** refers to the loop inside the Parallel.ForEach(...) command which is iterating through each object in 'source' and using it to execute 'body'. Following this assumption, the following code should throw an exception:

```
var bools = new List<bool> { true, false };

Parallel.ForEach(
	source: bools,
	body: (b) =>
	{
		while (true) { if (b) throw new Exception(); }
	});
```

However, it does not - it runs infinitely, even after the first thread throws an exception. This is misleading and should be corrected.

I'm assuming what the author meant was that an exception thrown during any executions of the 'body' action will act like a normal exception. While this is true inside the body of the action, it is not true for how it is handled outside of the action by the Parallel.ForEach(...) code.

[Edited by gewarren to add links in Document Details]

---
#### Document Details

⚠ Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.

* ID: b8ef7804-93ba-c398-19c4-1df6a3c5b168
* Version Independent ID: 0e699280-a6da-d56f-d713-bfd48d74bd28
* Content: [How to: Handle Exceptions in Parallel Loops](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/how-to-handle-exceptions-in-parallel-loops)
* Content Source: [docs/standard/parallel-programming/how-to-handle-exceptions-in-parallel-loops.md](https://github.com/dotnet/docs/blob/master/docs/standard/parallel-programming/how-to-handle-exceptions-in-parallel-loops.md)
* Product: **dotnet**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

pkulikov:
Fixes #17015

### DIFF
--- a/docs/standard/parallel-programming/how-to-handle-exceptions-in-parallel-loops.md
+++ b/docs/standard/parallel-programming/how-to-handle-exceptions-in-parallel-loops.md
@@ -10,7 +10,7 @@ helpviewer_keywords:
 ms.assetid: 512f0d5a-4636-4875-b766-88f20044f143
 ---
 # How to: Handle Exceptions in Parallel Loops
-The <xref:System.Threading.Tasks.Parallel.For%2A?displayProperty=nameWithType> and <xref:System.Threading.Tasks.Parallel.ForEach%2A?displayProperty=nameWithType> overloads do not have any special mechanism to handle exceptions that might be thrown. In this respect, they resemble regular `for` and `foreach` loops (`For` and `For Each` in Visual Basic); an unhandled exception causes the loop to terminate immediately.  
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks @pkulikov 

This is a nice, concise, clarifying change. 
I'll :shipit: now.

PR #17046, opened 2020-02-11
Moved and revised the void keyword article

pkulikov:
Supported by dotnet/samples#1991


BillWagner:
closing and opening for a fresh build.

docs/csharp/language-reference/builtin-types/void.md
```diff
@@ -0,0 +1,25 @@
...
+
+You use `void` as the return type of a [method](../../programming-guide/classes-and-structs/methods.md) (or a [local function](../../programming-guide/classes-and-structs/local-functions.md)) to specify that the method doesn't return a value.
+
+[!code-csharp[void method](~/samples/csharp/language-reference/builtin-types/VoidType.cs#VoidExample)]
+
+You can also use `void` as a referent type to declare a pointer to an unknown type. For more information, see [Pointer types](../../programming-guide/unsafe-code-pointers/pointer-types.md).
+
+You cannot use `void` as the type of a variable.
+
+The type keyword `void` is an alias for the .NET <xref:System.Void?displayProperty=nameWithType> type.
```
BillWagner: I wonder what this adds. Because it's rare that anyone would use `System.Void` (outside of reflection). It's true, but I don't know how helpful it is. I'm tempted to ask to delete this paragraph.

BillWagner (commented):
Everything looks great to me @pkulikov 

I had one comment for discussion. What are your thoughts?

docs/csharp/language-reference/builtin-types/void.md
pkulikov: @BillWagner I think at least the link to the .NET type should present on the page, but I agree that the paragraph might be too much. So, I've removed the paragraph and added the link to the **See also** section. What do you think?

BillWagner: That's perfect.

BillWagner (approved):
This is great @pkulikov I like the resolution on `System.Void`.

I'll :shipit: now

PR #17069, opened 2020-02-13
Update named fields pattern matching example

jdodds:
The definition of `Shape` earlier in the file uses `width`, not `height`, for defining `Rectangle`


cartermp (approved):
Thanks!

PR #17077, opened 2020-02-13
Replace "user" with "use".

RobJohnston:
## Summary

MARS can replace the **use** of server-side cursors, not the **user**.


mairaw (approved):
Thank you @RobJohnston for your contribution. It should go live within a day.

PR #17082, opened 2020-02-13
Add ASP.NET Core API ref to the hub page

scottaddie:
Add an ASP.NET Core API ref card to the hub page, per our discussion today.


mairaw (approved):
LGTM. Thanks @scottaddie!

Issue #4827, opened 2018-03-30
The % (Remainder) operator is named incorrectly in the docs


aaronfranke:
https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/operators/modulus-operator.md

https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/operators/index.md

This operator is named incorrectly in the documentation. It should be "Remainder" not "Modulus".

In C#, the functionality of the `%` operator is the Remainder function. The Remainder function returns a value that is the remainder after standard division. For example, `10 % 3` is `1`, and this is true for both Remainder and Modulus. 

The difference between these two is with negative numbers. 

* The standard integer division `-10 / 3` is `-3` so the Remainder can be found with `-10 - (-3)*3` which is `-1`. Remainder of `a` and `b` returns a value on the range `(-b, b)` (both exclusive).

* However, the Modulus function of `-10` and `3` would return `2`. Modulus satisfies a similar formula but with floored division. Floor division of `-10` and `3` is `-4` instead of `-3`, and `-10 - (-4)*3` is `2`. Modulus of `a` and `b` returns a value on the range `[0, b)` (0 inclusive, b exclusive).

* C# does not have operators for Modulus or floored division, only Remainder and standard division. The docs should reflect this, so all documentation for `%` should be renamed to Remainder.

Here's Eric Lippert's thoughts on the issue: https://blogs.msdn.microsoft.com/ericlippert/2011/12/05/whats-the-difference-remainder-vs-modulus/

It's easy to see why this confusion exists, as Remainder and Modulus are equivalent functions for positive numbers. But they are not the same operations. Describing `%` as Modulus leads to confusion: https://stackoverflow.com/questions/10065080/mod-explanation

pkulikov:
@aaronfranke you are right, thank you for reporing the issue.While the operator page itself doesn't mention *modulus*, the URL address might be misleading (I would rename the filename). The description in the operators table should be fixed. I am willing to update both pages to add clarifications.

svick:
I'm not sure the terminology is as clear as you imply:

* Eric Lippert is very careful in his article to use the term *canonical modulus*, which is not the same as just "modulus".
* The Wikipedia article about this issue, [*Modulo operation*](https://en.wikipedia.org/wiki/Modulo_operation) says "modulus" is a synonym of "modulo", which covers all the options. And it describes the two options as "remainder of truncated division" and "remainder of  Euclidean division".

[...]

PR #17102, opened 2020-02-15
Modulo operator -> remainder operator

pkulikov:


BillWagner (approved):
Thanks @pkulikov 

You made me check the conversation in #4827 for this. We did agree that we should use "remainder" in all docs.

PR #17105, opened 2020-02-15
Fix the indentation for click event setup code examples

jdodds:
Just a couple whitespace changes to make these examples valid 🙂


cartermp (approved):
Thanks

PR #16312, opened 2019-12-18
Add struct breaking change

gewarren:
Fixes #9148

Preview link: https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/2.0-2.1?branch=pr-en-us-16312

Issue #9148, opened 2018-11-20
Breaking changes in .NET Core 2.1 structs


martinsuchan:
There are breaking changes in some .NET Core 2.1 structs - these structs now have private fields and cannot be instantiated without `new`. There should be probabaly some official "breaking changes" section in this document.
https://github.com/dotnet/corefx/issues/6185

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 3bc74517-2839-4b16-8539-d31aba49bcfd
* Version Independent ID: 5511dd8b-d5d6-a3fb-172a-f35ff530cfc8
* Content: [What's new in .NET Core 2.1](https://docs.microsoft.com/en-us/dotnet/core/whats-new/dotnet-core-2-1)
* Content Source: [docs/core/whats-new/dotnet-core-2-1.md](https://github.com/dotnet/docs/blob/master/docs/core/whats-new/dotnet-core-2-1.md)
* Product: **dotnet-core**
* GitHub Login: @rpetrusha
* Microsoft Alias: **ronpet**

gewarren:
@martinsuchan Thanks, we will document this as a breaking change.

PR #17114, opened 2020-02-18
Remove the note about the struct breaking change from the conceptual docs

pkulikov:
The original note has sounded as if it's applicable to ANY structure type, which is not true, as far as I understand. This PR introduces two fixes:

- As the original issue #9148 states, the change affects only certain .NET structs. Not any struct type.
- If one doesn't use the `new` operator, one cannot initialize private fields after declaration of a struct instance. One can avoid using the `new` operator only if all instance fields are public. Because affected .NET structs have private fields added, there are only two ways to initialize them: the `new` operator or `default`. So, I've removed the third alternative.

However, now, after the fixes applied, the note is very specific and, I believe, about a rare use case. I don't think that note should be in [this article](https://docs.microsoft.com/en-us/dotnet/csharp/structs), which is the general overview of structs, at all. I believe that having it at the [breaking changes](https://docs.microsoft.com/en-us/dotnet/core/compatibility/2.0-2.1) part of the docs is enough. 

So, I suggest to remove this note at all.
@BillWagner @gewarren @jaredpar what do you think?

Related to #9148 and #16312

### DIFF
--- a/docs/csharp/structs.md
+++ b/docs/csharp/structs.md
@@ -41,9 +41,6 @@ Structs share most of the same syntax as classes, although structs are more limi
 
 - Unlike classes, structs can be instantiated without using a `new` operator.
 
-   > [!NOTE]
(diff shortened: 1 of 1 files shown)


BillWagner:
@pkulikov @jaredpar 

> I believe that having it at the breaking changes part of the docs is enough.

I agree with this statement. The breaking change is not in a typical use case. Does anyone have evidence that this not should be more prominent?

jaredpar:
This makes sense to me.

pkulikov:
@BillWagner thanks, I've removed the note then.

BillWagner (approved):
Thanks @pkulikov 

I'll :shipit: once the build finishes.

Issue #4561, opened 2018-03-06
Do not claim that structs are on the stack


svick:
[The *new Operator* article](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/new-operator) says:

> Value-type objects such as structs are created on the stack, while reference-type objects such as classes are created on the heap.

Depending on how you understand it, this claim could be considered true, if you consider that value types are *created* on the stack, but then can be copied to the heap.

But I think it's still too confusing and too close to the common misconception of "value types are stored on the stack, reference types on the heap" and that the wording should be changed.

(But I'm not sure what should it be changed to, which is why I'm creating an issue and not a PR.)

BillWagner:
I'd like to start a discussion on this issue.

First, this is much larger than a single topic. A search of the repo for the phrase ["on the stack"](https://github.com/dotnet/docs/search?utf8=%E2%9C%93&q=%22on+the+stack%22&type=) produces more than 100 occurrences. Some are correct, in that they refer to a Stack collection. Most refer to the common inaccurate phrase that value types are "on the stack".

@EricLippert wrote a few great blog posts explaining that the relevant fact for value types is they are **copied by value**, not that they are stored on the stack. [1](https://blogs.msdn.microsoft.com/ericlippert/2010/09/30/the-truth-about-value-types/) [2](https://blogs.msdn.microsoft.com/ericlippert/2009/04/27/the-stack-is-an-implementation-detail-part-one/)

The recent additions of `in`, `readonly struct` and `ref struct` declarations make the copy semantics much more relevant than the storage implementation. 

I propose that we remove any description of stack storage from all documentation on value types in the language guide(s) and the .NET guide, and replace that explanation with a correct discussion on the copy semantics.

dopare:
I would suggest to add clear note that value types are not restricted to stack storage instead of removing all descriptions about stack storage.

jskeet:
While I'm always nervous about disagreeing with Bill, I'd argue that `ref struct` makes the stack/heap distinction more relevant than ever before. The rules around what you can do with a `ref struct` are designed to avoid it escaping the stack. (The implementation details of captured variables etc then definitely leak into the language.)

I think it would make sense to start with the "copy by value" part, but then include material about heap vs stack in later material.

ericlippert:
This is indeed a tricky topic with over a decade of misinformation and confusing documentation muddying the waters.  So indeed, first, do no further harm!  

[...]

PR #17120, opened 2020-02-19
Moved and revised the struct article

pkulikov:
This PR consolidates the information about structs from the following articles:

- https://docs.microsoft.com/en-us/dotnet/csharp/structs
- https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/structs
- https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/using-structs

Not in scope of this PR: (1) redirection of the mentioned three articles (that is the next PR), (2) describing `ref` and `readonly` structs (also, it will be covered by dedicated PRs).

Supported by dotnet/samples#2008
Contributes to #10726


docs/csharp/language-reference/builtin-types/struct.md
```diff
@@ -0,0 +1,56 @@
...
+  - "structs [C#], struct keyword"
+ms.assetid: ff3dd9b7-dc93-4720-8855-ef5558f65c7c
+---
+# Structure types (C# reference)
+
+A *structure type* (or *struct type*) is a [value type](value-types.md) that encapsulates data and related functionality. As the following example shows, you use the `struct` keyword to define a structure type:
+
+[!code-csharp[struct example](~/samples/csharp/language-reference/builtin-types/StructType.cs#StructExample)]
+
+Because a structure type is a value type, a variable of a structure type contains an instance of the type. That is called value semantics. By default, variables are copied on assignment, passing an argument to a method, or returning a method result. In the case of a structure-type variable, an instance of the type is copied (for the examples, see [Value types](value-types.md)). In high-performance scenarios that involve large structure types, you can avoid value copying and pass around an alias to a value-type variable. For information about how to do that, see [Write safe and efficient C# code](../../write-safe-efficient-code.md).
```
pkulikov: @BillWagner what do you think of the "alias to a value-type variable" wording?

I've noticed that in the docs the preferred wording is "by reference". However I'm afraid that it might be confusing and misleading. How?

Imagine the following code:

```csharp
int b = 0;
Change(ref b);
Console.WriteLine(b);

void Change(ref int a)
{
    a = 4;
}
```

The above code would print `4`. When we say that we pass `int` by reference, we might imply that there is reference semantics here, that *if `int` were a reference type*, the above code would be equivalent to the following (you wouldn't need `ref`):

```csharp
int b = 0;
Change(b);
Console.WriteLine(b);

void Change(int a)
{
    a = 4;
}
```

And that's not true. Because what is passed is not the reference to a variable but an alias to a variable. That's why whatever is done with that alias is also reflected on the "original" variable. Personally, thinking in aliases instead of references helped me a lot, especially with ref locals and ref returns.

However, if you think that it's not a good idea to deviate from the current docs language, I'll use "by reference" wording in this article as well.

pkulikov: > When we say that we pass `int` by reference, we might imply that there is reference semantics here, that _if `int` were a reference type_...

This is how I'm demonstrating the confusion that the [note in the `ref` article](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/ref#passing-an-argument-by-reference) warns about :).

BillWagner: I looked at the discussion on #4561, and "by reference:" would be preferred.

In time, I'd like to follow more of the recommendations Eric posted in https://github.com/dotnet/docs/issues/4561#issuecomment-371319556. For now though, let's keep the current language.

BillWagner: I'd push the last two sentences lower in the article. The important design choice is `class` vs. `struct`, and you've discussed the main points her and in the next paragraph.

Later, once that decision has been made, the reader may be interested in performance optimizations.

docs/csharp/language-reference/builtin-types/struct.md
```diff
@@ -0,0 +1,56 @@
...
+
+Typically, you define a structure type when you need to represent a small data structure that has value semantics. For example, .NET uses structure types to represent a number (both [integer](integral-numeric-types.md) and [real](floating-point-numeric-types.md)), a [Boolean value](bool.md), a [Unicode character](char.md), a [time instance](xref:System.DateTime). If you need a type with reference semantics (that is, a variable of the type contains a reference to an instance of that type, not the instance itself), define a [class](../keywords/class.md).
```
BillWagner: I think we should emphasize the point in the first sentence more: Structs are when the behavior of the type is secondary to its data. (On the other hand, classes are primarily about behavior, with data that support the behavior.)

docs/csharp/language-reference/builtin-types/struct.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Limitations with the design of a structure type
```
BillWagner: (All these make more sense when you think about the struct as *data*, not *behavior*.)

docs/csharp/language-reference/builtin-types/struct.md
```diff
@@ -0,0 +1,56 @@
...
+
+You use the [`new`](../operators/new-operator.md) operator to instantiate a structure type. In the case of the [built-in value types](value-types.md#built-in-value-types), use the corresponding literals to specify a value of the type.
+
+You can also instantiate a structure type without the `new` operator. However, you must initialize all instance fields before the first use of the instance. The following example shows how to do that:
+
+[!code-csharp[without new](~/samples/csharp/language-reference/builtin-types/StructType.cs#WithoutNew)]
+
+If a structure type contains private fields, you must use the `new` operator to instantiate the type.
+
+## Conversions
```
BillWagner: Here, before conversions is where I'd recommend adding the (brief) discussion on reference semantics.

docs/csharp/language-reference/builtin-types/struct.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Instantiation of a structure type
+
+You use the [`new`](../operators/new-operator.md) operator to instantiate a structure type. In the case of the [built-in value types](value-types.md#built-in-value-types), use the corresponding literals to specify a value of the type.
+
+You can also instantiate a structure type without the `new` operator. However, you must initialize all instance fields before the first use of the instance. The following example shows how to do that:
+
+[!code-csharp[without new](~/samples/csharp/language-reference/builtin-types/StructType.cs#WithoutNew)]
+
+If a structure type contains private fields, you must use the `new` operator to instantiate the type.
```
BillWagner: I"m not sure what you're trying to explain here. I don't think its correct in that arrays of a struct type can be created (`new` isn't called on each element).

BillWagner (changes requested):
This is great @pkulikov 

I think it's a concise description of structs. I had a couple comments where I think the language could be more clear.

Let me know what you think.

docs/csharp/language-reference/builtin-types/struct.md
pkulikov: @BillWagner I'm confused; by the "current language" you mean

1. the current language of the article, that is, "pass around an alias to a value-type variable" OR
1. the current language of the docs, that is, I should rephrase the article to use "by reference"?

I guess, you mean variant 2. Is that correct?

BillWagner: Sorry about that. Number 2 is correct. Thanks for asking.

docs/csharp/language-reference/builtin-types/struct.md
pkulikov: @BillWagner do you mean reference semantics with structure types, as per [this comment](https://github.com/dotnet/docs/pull/17120#discussion_r382241564)?

docs/csharp/language-reference/builtin-types/struct.md
pkulikov: Indeed, it's not correct; thank you for the array example. I'll remove that sentence.

docs/csharp/language-reference/builtin-types/struct.md
```diff
@@ -0,0 +1,56 @@
...
+
+- You cannot declare a parameterless constructor. Every structure type already provides an implicit parameterless constructor that returns the [default value](default-values.md) of the type.
+
+- You cannot initialize an instance field at its declaration. However, you can initialize a [static](../keywords/static.md) or [const](../keywords/const.md) field at its declaration.
+
+- A structure type cannot inherit from other class or structure type and it cannot be the base of a class. However, a structure type can implement [interfaces](../keywords/interface.md).
+
+- You cannot declare a [finalizer](../../programming-guide/classes-and-structs/destructors.md) within a structure type.
+
+## Instantiation of a structure type
```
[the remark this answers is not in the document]
pkulikov: > ...using a combination of any accessible constructor and accessible property initializers

That reads to me as that I can call a constructor and then use properties to initialize remaining members. For example, for given struct:
```csharp
    public struct Coords
    {
        public double X { get; set; }
        public double Y;

        public Coords(double y) => Y = y;

        public override string ToString() => $"{X:F2} {Y:F2}";
    }
```

I would be able to call 
```csharp
Coords a = new Coords(y: 5);
a.X = 4;
```

However, the above struct declaration doesn't compile because `X` is not assigned in the constructor. So one must use only `new` with properly defined constructor that initializes all instance data members. 

Probably, I misunderstand you @BillWagner. What do you mean by "using a combination of any accessible constructor and accessible property initializers"?

docs/csharp/language-reference/builtin-types/struct.md
BillWagner: yes.

docs/csharp/language-reference/builtin-types/struct.md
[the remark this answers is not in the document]
BillWagner: Ahh. I was thinking something like this.

From the struct author perspective, all fields must be initialized in all declared constructors:

```csharp
public struct Coords
{
    public double X { get; set; }
    public double Y;

    private Coords(double x, double y) => (X, Y) = (x, y);
    public Coords(double y) : this(0.0, y) { }
}
```

And then from the perspective of callers, they can use constructor args or accessible field accessors to set values:

```csharp
Coords a = new Coords(y: 5) {X = 4 };
```

Looking at this conversation, it's probably wise to separate those two perspectives.

Great conversation and discussion. I am confident this will lead to a great final version. 👯‍♂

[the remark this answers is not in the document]
pkulikov: Great! Thank you for the clarification and the example!

pkulikov:
@BillWagner thanks a lot for your review; it helped me to improve the article a lot. Please review the new version (as I've updated almost every paragraph, I think it would be easier to read it again all through rather than checking commits).

docs/csharp/language-reference/builtin-types/struct.md
```diff
@@ -0,0 +1,65 @@
...
+
+In C#, you must initialize a declared variable before it can be used. Because a structure-type variable cannot be `null` (unless it's a variable of a [nullable value type](nullable-value-types.md)), you must instantiate an instance of the corresponding type. There are several ways to do that.
+
+Typically, you instantiate a structure type by calling an appropriate constructor with the [`new`](../operators/new-operator.md) operator. Every structure type has at least one constructor. That's an implicit parameterless constructor, which produces the [default value](default-values.md) of the type. You can also use the [default](../operators/default.md) operator or literal to produce the default value of a type.
```
pkulikov: As it's a general recommendation to define *immutable* value types, I haven't mentioned object initializers here. However, the linked article about the `new` operator discusses object initializers.

docs/csharp/language-reference/builtin-types/struct.md
```diff
@@ -0,0 +1,65 @@
...
+
+If all instance fields of a structure type are accessible, you can also instantiate it without the `new` operator. In that case you must initialize all instance fields before the first use of the instance. The following example shows how to do that:
+
+[!code-csharp[without new](~/samples/csharp/language-reference/builtin-types/StructType.cs#WithoutNew)]
+
+In the case of the [built-in value types](value-types.md#built-in-value-types), use the corresponding literals to specify a value of the type.
+
+## Passing structure-type variables by reference
```
pkulikov: I intentionally don't use the words "reference semantics" here to avoid confusion between passing variables by reference and reference types. I see #4561 warns about that several times. By the way, thanks for the link to that issue. I've tried to update the article inline with the guidance there.

pkulikov (commented):
A couple of clarifications of made choices.

docs/csharp/language-reference/builtin-types/struct.md
BillWagner: I really like the way you've worded this.

BillWagner (approved):
This looks great now @pkulikov 
I'll merge both this and the associated samples PR.

BillWagner:
closing and reopening for a fresh build after merging the samples PR.

docs/csharp/language-reference/builtin-types/struct.md
pkulikov: > Have you been reading the discussions about records and initializations for future versions?

I've noticed them and looked through but didn't read in details and thoughtfully yet.

PR #17121, opened 2020-02-19
Update F# type attribute example

jdodds:
Make the struct definition valid, and qualify the reference to `Sequential`


cartermp (approved):
Thanks!

PR #17169, opened 2020-02-21
Updating Health check documentation to use built in .NET core classes…

DonaldRWood:
… instead of 3rd party tool

## Summary

The document at 

https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/monitor-app-health

has samples that do not match the text.  They appear to be leftovers from the previous edition with .NET core 2.2 where a non-Microsoft library was needed.  In 3.1 the HealthCheck class can be implemented using only Microsoft classes.

I changed the 1st and 3rd code sample to work correctly with these new classes.

### DIFF
--- a/docs/architecture/microservices/implement-resilient-applications/monitor-app-health.md
+++ b/docs/architecture/microservices/implement-resilient-applications/monitor-app-health.md
@@ -34,10 +34,11 @@ public void ConfigureServices(IServiceCollection services)
     // Registers required services for health checks
     services.AddHealthChecks()
         // Add a health check for a SQL Server database
-        .AddSqlServer(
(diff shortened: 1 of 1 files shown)


nishanil:
Thanks @DonaldRWood for the PR. @mvelosop Can you verify this on eShop too?

mvelosop:
Hi @DonaldRWood,

Thanks for the PR 😊. 

The code changes are correct. so this can be merged.

@nishanil, I'm approving this one and creating another one to fix a couple more related issues in this article.

mvelosop (approved):
Thanks @DonaldRWood

PR #17184, opened 2020-02-22
Fix incorrect argument for global tool docs

JeremySkinner:
## Summary

This sentence is referring to installing a global tool in the default location, but incorrectly says you should use `--tool-path` instead of `--global`.


tdykstra (approved):
@JeremySkinner Thanks for fixing that!

PR #17188, opened 2020-02-24
Mention the OtherFlags property for fsc cli args

jdodds:
This is more or less the only way to set extra cli args for fsc for F# projects targeting .NET core, makes sense to mention it here.


cartermp (approved):
Thanks! This is a good change.

PR #17194, opened 2020-02-24
Fix tagname to design action

Thraka:
We forgot to change the title of the TOC entry after renaming the file.


BillWagner (approved):
This LGTM @Thraka 

You can :shipit: when ready.

PR #17195, opened 2020-02-24
fix comment in interactive code block

beeisabelm:
this small fix prevents the system error `CS1056: Unexpected character "\" at run time.`

## Summary

fix comment as forward slashes in code block so it does not error out in csharp interactive 


BillWagner (approved):
Thanks for spotting this and making the fix @beeisabelm 

I'll :shipit: now.

Issue #12736, opened 2019-06-08
Var pattern: Unfortunate code example makes the "var pattern" appear without purpose


elgonzo:
The example code given for the "_var pattern_":
````
object[] items = { new Book("The Tempest"), new Person("John") };
foreach (var item in items) {
    if (item is var obj)
        Console.WriteLine($"Type: {obj.GetType().Name}, Value: {obj}"); 
}
````
is just a needlessly complicated way of writing
````
object[] items = { new Book("The Tempest"), new Person("John") };
foreach (var item in items)
    Console.WriteLine($"Type: {item.GetType().Name}, Value: {item}"); 
````
thus inadvertently making it look like as if the "_var pattern_" has no useful function or purpose.

It would benefit the documentation if it would show an example code demonstrating the "_var pattern_" in a meaningful way. Perhaps something like the code snippet in this answer on StackOverflow: https://stackoverflow.com/a/50660530/2819245 (No, i am not nor otherwise related to the author of this answer.)

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 3126aaa9-8f5f-04f3-2d75-96477b6273cd
* Version Independent ID: 3f4b3eac-11eb-2dd6-4f27-b24537367288
* Content: [is - C# Reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/is#var)
* Content Source: [docs/csharp/language-reference/keywords/is.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/keywords/is.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

**Comment by BillWagner** [comment]
Thanks for writing this @elgonzo That's a good example to add.

Another use is to check for `null`.  Consider this slightly modified version of the original example:

```csharp
object[] items = { new Book("The Tempest"), null, new Person("John"), null };
foreach (var item in items) {
    if (item is var obj)
        Console.WriteLine($"Type: {obj.GetType().Name}, Value: {obj}"); 
}
```

The `is` check will be `false` for the `null` values, so it won't throw a `NullReferenceException`.

I've added the "up for grabs" label to this issue, and added it to our backlog for when we next update this article.

elgonzo:
[...]

PR #17197, opened 2020-02-24
New code example for the "var pattern" (issue #12736)

mssteele:
Addresses #12736 

Documentation changes relating to the new code example.

Also see my separate change to the code sample itself in the dotnet/samples repo.
PR: https://github.com/dotnet/samples/pull/2019

### DIFF
--- a/docs/csharp/language-reference/keywords/is.md
+++ b/docs/csharp/language-reference/keywords/is.md
@@ -31,7 +31,7 @@ When using the type pattern to perform pattern matching, `is` tests whether an e
    expr is type varname
 ```
 
-where *expr* is an expression that evaluates to an instance of some type, *type* is the name of the type to which the result of *expr* is to be converted, and *varname* is the object to which the result of *expr* is converted if the `is` test is `true`. 
+Where *expr* is an expression that evaluates to an instance of some type, *type* is the name of the type to which the result of *expr* is to be converted, and *varname* is the object to which the result of *expr* is converted if the `is` test is `true`. 
 
 The `is` expression is `true` if *expr* isn't `null`, and any of the following is true:
 
--- a/docs/csharp/language-reference/keywords/is.md
+++ b/docs/csharp/language-reference/keywords/is.md
@@ -101,15 +101,21 @@ The following example shows a comparison of `null` checks:
 
 ### var pattern
 
-The `var` pattern is a catch-all for any type or value. The value of *expr* is always assigned to a local variable the same type as the compile time type of *expr*. The result of the `is` expression is always `true`. Its syntax is:
+A pattern match with the `var` pattern always succeeds. Its syntax is:
 
 ```csharp
    expr is var varname
 ```
 
-The following example uses the var pattern to assign an expression to a variable named `obj`. It then displays the value and the type of `obj`.
+where the value of *expr* is always assigned to a local variable named *varname*. *varname* is a variable of the same type as the compile time type of *expr*. 
+
+If *expr* is `null`, the `is` expression is still true and assigns `null` to *varname*. 
+
+One reason to use the `var` pattern is to create a temporary variable within a Boolean expression. For example:
 
 [!code-csharp[is#8](../../../../samples/snippets/csharp/language-reference/keywords/is/is-var-pattern8.cs#8)]
+
+Here the temporary variable is used to store the result of an expensive operation. The variable can then be used multiple times.
 
 ## C# language specification
   
--- a/docs/csharp/language-reference/keywords/is.md
+++ b/docs/csharp/language-reference/keywords/is.md
@@ -109,7 +109,7 @@ A pattern match with the `var` pattern always succeeds. Its syntax is:
 
 where the value of *expr* is always assigned to a local variable named *varname*. *varname* is a variable of the same type as the compile-time type of *expr*. 
 
-If *expr* evaluates to `null`, the `is` expression produces `true` and assigns `null` to *varname*. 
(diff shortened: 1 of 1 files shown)


docs/csharp/language-reference/keywords/is.md
```diff
@@ -101,16 +101,22 @@ The following example shows a comparison of `null` checks:
...
 
-The `var` pattern is a catch-all for any type or value. The value of *expr* is always assigned to a local variable the same type as the compile time type of *expr*. The result of the `is` expression is always `true`. Its syntax is:
+A pattern match with the `var` pattern always succeeds. Its syntax is:
 
 ```csharp
    expr is var varname
 ```
 
-The following example uses the var pattern to assign an expression to a variable named `obj`. It then displays the value and the type of `obj`.
+where the value of *expr* is always assigned to a local variable named *varname*. *varname* is a variable of the same type as the compile time type of *expr*.
```
pkulikov:
```suggestion
where the value of *expr* is always assigned to a local variable named *varname*. *varname* is a variable of the same type as the compile-time type of *expr*. 
```

docs/csharp/language-reference/keywords/is.md
```diff
@@ -101,16 +101,22 @@ The following example shows a comparison of `null` checks:
...
+A pattern match with the `var` pattern always succeeds. Its syntax is:
 
 ```csharp
    expr is var varname
 ```
 
-The following example uses the var pattern to assign an expression to a variable named `obj`. It then displays the value and the type of `obj`.
+where the value of *expr* is always assigned to a local variable named *varname*. *varname* is a variable of the same type as the compile time type of *expr*. 
+
+If *expr* is `null`, the `is` expression is still true and assigns `null` to *varname*.
```
pkulikov:
```suggestion
If *expr* evaluates to `null`, the `is` expression produces `true` and assigns `null` to *varname*. 
```

docs/csharp/language-reference/keywords/is.md
```diff
@@ -101,16 +101,22 @@ The following example shows a comparison of `null` checks:
...
 ```csharp
    expr is var varname
 ```
 
-The following example uses the var pattern to assign an expression to a variable named `obj`. It then displays the value and the type of `obj`.
+where the value of *expr* is always assigned to a local variable named *varname*. *varname* is a variable of the same type as the compile time type of *expr*. 
+
+If *expr* is `null`, the `is` expression is still true and assigns `null` to *varname*. 
+
+One reason to use the `var` pattern is to create a temporary variable within a Boolean expression. For example:
```
pkulikov:
```suggestion
You can use the `var` pattern to create a temporary variable within a Boolean expression, as the following example shows:
```

docs/csharp/language-reference/keywords/is.md
```diff
@@ -101,16 +101,22 @@ The following example shows a comparison of `null` checks:
...
 
 [!code-csharp[is#8](../../../../samples/snippets/csharp/language-reference/keywords/is/is-var-pattern8.cs#8)]
 
+Here the temporary variable is used to store the result of an expensive operation. The variable can then be used multiple times.
```
pkulikov:
```suggestion
In the preceding example, the temporary variable is used to store the result of an expensive operation. The variable can then be used multiple times.
```

pkulikov (commented):
Rephrased to match the usual docs language.

docs/csharp/language-reference/keywords/is.md
```diff
@@ -101,16 +101,22 @@ The following example shows a comparison of `null` checks:
...
+A pattern match with the `var` pattern always succeeds. Its syntax is:
 
 ```csharp
    expr is var varname
 ```
 
-The following example uses the var pattern to assign an expression to a variable named `obj`. It then displays the value and the type of `obj`.
+where the value of *expr* is always assigned to a local variable named *varname*. *varname* is a variable of the same type as the compile-time type of *expr*. 
+
+If *expr* evaluates to `null`, the `is` expression produces `true` and assigns `null` to *varname*.
```
BillWagner: Is it worth adding a clarifying comment here?  In most situations `is` produces `false` for `null`:

```suggestion
If *expr* evaluates to `null`, the `is` expression produces `true` and assigns `null` to *varname*.  The var pattern is one of the few uses of `is` that produces `true` for a `null` value.
```

BillWagner (changes requested):
I really like this change @mssteele 

I had one suggestion, because we've seen confusion because `is` produces `false` in most cases when the argument is `null`. What do you think?

Marking as "request changes" so it stays in my queue while we discuss.

docs/csharp/language-reference/keywords/is.md
Thraka: I don't think it's particularly useful. Side question, are there other uses of var that does produce true for null?

BillWagner: @Thraka Nope, `thing is var newVariableName` always succeeds, and always returns true.

But, `thing is string message` returns `false` if `thing` is `null`.

mssteele:
Yes good suggestion @BillWagner 
I think it is worth pointing out. It is a bit unintuitive when you compare it to your example where
`thing is string message` returns `false` if `thing` is `null` 
(Sorry - I should have probably commented next to your suggestion but I accepted it so then couldn't comment there!)

BillWagner (approved):

BillWagner:
Thanks for making these additional changes @mssteele 

I've approved this, and I'll :shipit: now.

Thanks again!

Issue #17199, opened 2020-02-25
F# library and interactive reference readiness


cartermp:
The F# library and interactive reference is now merged into a migration branch: https://github.com/dotnet/docs/pull/17004

There are many things that need to be fixed up before it can be considered for inclusion into the full F# documentation set. Here are some of them:

- [x] All snippet references need to be specified correctly in documents
- [ ] File names should be adjusted to be more URL-friendly
- [ ] Titles are simpler/shorter
- [ ] Parameter section uses a table
- [ ] File metadata should be updated (simplified and dates updated)
- [ ] HTML-encoded text should be cleaned up
- [x] All redundant files (e.g., Abstract classes) are removed
- [ ] All outdated information (Platforms, Version Information) removed
- [ ] All unnecessary info (e.g., "Assembly: FSharp.Core (in FSharp.Core.dll)") removed
- [ ] F# interactive reference is "hooked up" to the main F# TOC
- [ ] F# library reference is "hooked up" to the main F# TOC
- [ ] All links to the old MSDN site should be updated, and all mentions of the reference being incomplete must be updated
- [ ] Make markdownlint happy

Separately, the following need to also be added, but need not block moving the doc set from the migration branch:

- [ ] Migrate issues from https://github.com/MicrosoftDocs/visualfsharpdocs/issues
- [ ] Complete any in progress work from https://github.com/MicrosoftDocs/visualfsharpdocs/issues
- [ ] Complete F# 4.1 updates
- [ ] Complete F# 4.5 updates
- [ ] Complete F# 4.6 updates
- [ ] Complete F# 4.7 updates

aggieben:
It looks like this is still a good ways out - is there something we could have in the interim with regard to the core library reference, which currently exists nowhere?

akeeton:
@aggieben Yeah, this is really rough.  I managed to follow your trail from https://github.com/dotnet/docs/issues/17557.  I just wanted to see the immutable Map docs :(

cartermp:
Unfortunately I don't know how long this will take. That said, I wasn't aware that the MSDN links were all dead. That shouldn't have ever happened, since internally we agreed to migrate but not immediately redirect until the migration was complete. I'll raise this over email.

voronoipotato:
Anyone looking for them in the interim can find them here courtesy of @baronfel . 

https://baronfel.github.io/fsharp-core-api-docs/

baronfel:
[...]

PR #17200, opened 2020-02-25
[WIP] Update array.append

cartermp:
This update will serve as an example of how to update the F# core library reference docs as per #17199


docs/fsharp/library-reference/core-library-reference/array.append.md
```diff
@@ -0,0 +1,50 @@
+---
```
pkulikov: There is a duplication in the file path: *library-reference*

PR #17212, opened 2020-02-25
Update F# library reference snippet links

cartermp:
As per https://github.com/dotnet/docs/issues/17199


cartermp:
The remaining snippet warnings are coming from snippets that don't yet exist. I have open PRs to add them.

PR #17213, opened 2020-02-25
12). Add topic files 23 to 26

jameshkramer:
@mairaw @BillWagner
This PR is for user story #1576037.

It adds four topic files, makes corresponding deletions from the C# and VB folders, and adds redirection entries.

cc: @tfosmark

Contributes to #4728


mairaw (approved):
Thanks @jameshkramer!

PR #17219, opened 2020-02-26
Fix code style

pkulikov:
Otherwise, the colons are hard to notice.


BillWagner (approved):

I'll :shipit: now.

PR #17227, opened 2020-02-26
Update globalization.md

yurikburko:
Fixed hexadecimal representation of a LATIN CAPITAL LETTER A WITH GRAVE character code.

It seems that there is a mistake in description of sample code.

Code units for LATIN CAPITAL LETTER A WITH GRAVE character should be _**U+0041 and U+0300**_ but not **_U+0021 and U+007E_** as written.


tdykstra (approved):
@yurikburko Thanks!

PR #17240, opened 2020-02-27
Mention the precedence of the `switch` expression

### LINKED ISSUE #14698: Precedence of C# 8.0 switch expression
As discussed on [StackOverflow](https://stackoverflow.com/q/58088073/8528014), currently the documentation doesn't mention the precedence of the new `switch` expressions (which is between the categories `multiplicative` and `range`).

---
#### Document Details

* ID: 7eeac581-6235-2192-d909-e7fe3d21197f
* Version Independent ID: 9d8e2a57-1c18-8cba-9bd4-eef546fe2dcb
* Content: [C# operators - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/#feedback)
* Content Source: [docs/csharp/language-reference/operators/index.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/operators/index.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

pkulikov:
Fixes #14698

Lower than range `..`, higher than multiplicative; based on [this Roslyn source](https://github.com/dotnet/roslyn/blob/aca76082e096859514fbdbcdbd8faab7739a2c89/src/Compilers/CSharp/Portable/Parser/LanguageParser.cs#L8700-L8718).

However, from the following usage:
```csharp
var xs = new[] { 0, 1, 2, 3 };
var ys = xs[0 switch { 0 => 0 }..2];  // compiles because first `switch` then `..` is fine
var zs = xs[0..2 switch { 2 => 2 }];  // doesn't compile because first `..` is evaluated
```

it looks that both `..` and `switch` have the same precedence level. Should we put them on the same line in the table?

/cc @gafter


BillWagner:
The [patterns spec](https://docs.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-8.0/patterns#switch-expression) , and the [range spec](https://docs.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-8.0/ranges#systemrange) add the new grammars for each in the same location without mentioning a precedence between them. I'll defer to @gafter on the above question. Reviewing the other changes now.

BillWagner:
The [patterns spec](https://docs.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-8.0/patterns#switch-expression) , and the [range spec](https://docs.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-8.0/ranges#systemrange) add the new grammars for each in the same location without mentioning a precedence between them. I'll defer to @gafter on the above question. Reviewing the other changes now.

BillWagner:
@pkulikov We may not be ready for this change.  See dotnet/roslyn#36514 for a discussion of current ambiguities and a proposed solution.

Your analysis of the current implementation is correct, but this PR may be documenting current implementation errors rather than the correct behavior.

I'll keep this open for now. There are internal discussion when the behavior may be corrected.

pkulikov:
@BillWagner, thank you for the link. However I might have missed some information: the linked issue discusses the `..` (range) operator and its precedence. That one is already in the docs and published. Or the internal discussion also includes the `switch` expression? If not, I suggest to merge this PR as it is now (with `switch` on a separate line) because its main message—the `switch` precedence is lower than unary and higher than multiplicative—is correct and not going to be changed, I guess? The changes to the `..` precedence should be then addressed in a separate PR.

BillWagner:
@pkulikov The bug is that range is bound more tightly than switch, and that should be corrected.  The plan for that change is being discussed.

I'd like to wait and get more details on that change and its schedule. (It shouldn't be too long).

pkulikov:
@BillWagner ah, I see now, thank you for the update

BillWagner (approved):
This looks good, and I'll :shipit: 

We'll revisit when spec updates in this area are completed.

PR #17251, opened 2020-02-27
Make var usage consistent with other docs

johndec:
Edited paragraph on use of var to correspond with Coding Guidelines.

## Summary

The coding guidelines suggest the use of `var` in certain circumstances -- e.g. when the type is obvious on the right side, in for loops, and other cases. However, this document says it should only be used when its use is strictly required. This is a contradiction, so I'm clarifying the use cases here.


BillWagner (approved):
Thanks @johndec 

PR #17252, opened 2020-02-27
Change proj name in tutorial to microsoft-botsay

tdykstra:
Per suggestion from @Thraka , prevents uploading to NuGet without requiring user to come up with a unique name.

[Internal review URL](https://review.docs.microsoft.com/en-us/dotnet/core/tools/global-tools-how-to-create?branch=pr-en-us-17252)


Thraka (approved):
Just curious, did you test that you can't actually upload it? Is the requirement that the nuget package name can't be `Microsoft.BotSay`, the period instead of a dash?

tdykstra:
Good point on dash vs. period.  I changed it to a period.

tdykstra:
And I verified that it can't be uploaded. 😄

PR #17263, opened 2020-02-28
Fix code formatting

stefan-malcek:
The line below ``if`` statement had no indent.

## Summary

Added spaces in the line below the ``if`` statement.


tdykstra (approved):
@stefan-malcek Thanks!

PR #17271, opened 2020-02-29
Fix terminated I/O thread "raised when" field

andresff:
Perhaps a copy/paste mistake

## Summary

A minimum mistake


BillWagner (approved):
Thanks for catching this and fixing it @andresff  We appreciate it. 

PR #17333, opened 2020-03-04
Update delegates-lambdas.md

b4ry:
## Summary

- added a bit of clarification to the set of delegate types (if they return a value or not),
- fixed a typo

I did not create an issue since it does not look like one - just more like clarification.

### DIFF
--- a/docs/standard/delegates-lambdas.md
+++ b/docs/standard/delegates-lambdas.md
@@ -40,9 +40,9 @@ public class Program
 
 In order to streamline the development process, .NET includes a set of delegate types that programmers can reuse and not have to create new types. These are `Func<>`, `Action<>` and `Predicate<>`, and they can be used in various places throughout the .NET APIs without the need to define new delegate types. Of course, there are some differences between the three as you will see in their signatures which mostly have to do with the way they were meant to be used:
 
-* `Action<>` is used when there is a need to perform an action using the arguments of the delegate.
-* `Func<>` is used usually when you have a transformation on hand, that is, you need to transform the arguments of the delegate into a different result. Projections are a prime example of this.
-* `Predicate<>` is used when you need to determine if the argument satisfies the condition of the delegate. It can also be written as a `Func<T, bool>`.
+* `Action<>` is used when there is a need to perform an action using the arguments of the delegate. The method it encapsulates does not return a value.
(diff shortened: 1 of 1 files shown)


docs/standard/delegates-lambdas.md
```diff
@@ -40,9 +40,9 @@ public class Program
 
 In order to streamline the development process, .NET includes a set of delegate types that programmers can reuse and not have to create new types. These are `Func<>`, `Action<>` and `Predicate<>`, and they can be used in various places throughout the .NET APIs without the need to define new delegate types. Of course, there are some differences between the three as you will see in their signatures which mostly have to do with the way they were meant to be used:
 
-* `Action<>` is used when there is a need to perform an action using the arguments of the delegate.
-* `Func<>` is used usually when you have a transformation on hand, that is, you need to transform the arguments of the delegate into a different result. Projections are a prime example of this.
-* `Predicate<>` is used when you need to determine if the argument satisfies the condition of the delegate. It can also be written as a `Func<T, bool>`.
+* `Action<>` is used when there is a need to perform an action using the arguments of the delegate. The method it encapsulates does not return a value.
```
gewarren: I think it's a good idea to add the return values to the definitions here. I'm not 100% sure that "encapsulates" is technically accurate, since the method is assigned to the delegate type. @billwagner Can you comment if "encapsulates" is correct here?

docs/standard/delegates-lambdas.md
```diff
@@ -40,9 +40,9 @@ public class Program
...
+* `Func<>` is used usually when you have a transformation on hand, that is, you need to transform the arguments of the delegate into a different result. Projections are a prime example of this. The method it encapsulates returns a specified value.
+* `Predicate<>` is used when you need to determine if the argument satisfies the condition of the delegate. It can also be written as a `Func<T, bool>` which means the method it encapsulates returns a bool value.
```
gewarren:
```suggestion
* `Predicate<>` is used when you need to determine if the argument satisfies the condition of the delegate. It can also be written as a `Func<T, bool>`, which means the method returns a boolean value.
```

docs/standard/delegates-lambdas.md
b4ry: @gewarren 
Yes, I agree that we need some more insight into that if it creates confusion. To be honest, I was using msdn docs referring to these particular delegates :).

gewarren (approved):
Will go ahead and merge this now. Thanks again!

PR #17334, opened 2020-03-04
Correct last example

mrlife:
The last example was incorrect. I updated with intention to achieve the same result of 4. This differentiates the last example from the preceding one.

## Summary

Describe your changes here.

### DIFF
--- a/samples/snippets/csharp/language-reference/builtin-types/EnumType.cs
+++ b/samples/snippets/csharp/language-reference/builtin-types/EnumType.cs
@@ -71,7 +71,7 @@ namespace builtin_types
                 var b = (Season)1;
                 Console.WriteLine(b);  // output: Summer
 
-                var c = (Season)4;
+                var c = (int)Season.Winter;
                 Console.WriteLine(c);  // output: 4
             }
         }


samples/snippets/csharp/language-reference/builtin-types/EnumType.cs
```diff
@@ -71,7 +71,7 @@ public static void Main()
                 var b = (Season)1;
                 Console.WriteLine(b);  // output: Summer
 
-                var c = (Season)4;
+                var c = (int)Season.Winter;
```
pkulikov: That would make `c` equal 3, not 4 as intended. Intention here is to show that one can convert *any* integer to the enum type.

mrlife: Sorry about that, meant to change result to 3.

pkulikov: Then it becomes the example of converting enum to int. That's already covered by the first example. I don't think any changes are required here.

mrlife:
@pkulikov 

> Intention here is to show that one can convert any integer to the enum type.

My mistake. I was not sure the purpose of the example, so I learned something.

Issue #17297, opened 2020-03-03
Move snippets into the docs 


BillWagner:
We've found that keeping snippet (code) projects in a different repository than the articles where they are includes introduces unnecessary friction in updating articles.

So, we plan to move snippets into the dotnet/docs and dotnet/dotnet-api-docs repositorites.

The dotnet/samples repository will be the home of larger scenario-based samples that may be referenced from tutorials or other long-form explorations. In addition, the samples in the dotnet/samples repository are likely candidates for the samples repository.

## Task plan

1. Make sure all included snippets are in the /snippets folder in the dotnet/samples repository.
1. Update all snippet references to the /samples/snippets folder, while referencing the snippets in the dotnet/samples repository.
1. Move all snippets into dotnet/docs/samples/snippets and dotnet/dotnet-api-docs/samples/snippets as one massive PR (each repo).
1. Remove the samples repo dependency, and fix build errors. 
1. Remove the /snippets content in the dotnet/samples repo.
1. Update the contributor guide with the snippet and sample guidelines below.
1. Remove unreferenced snippets in dotnet/dotnet-api-docs/snippets. (Leverage the clean-repo tool for this task)
1. Remove unreferenced snippets in dotnet/docs/snippets. (Leverage the clean-repo tool for this task)
1. We configure samples CI on the conceptual and API reference repositories.
1. Add dependabot to the docs and dotnet-api-docs repositories.
1. After that task is completed, any new significant edits should move snippets into the correct folder structure in the content repo (see below).
1. All samples remaining in the samples repository could be configured for the samples browser.

Note that we will be concurrently enabling .NET Interactive (formerly known as Try.NET) for samples. That work should be done in separate PRs from moving samples and snippets into a new repo or folder structure.

## Guidelines for snippet projects

- They are projects (not solutions), and the `.sln` file is not checked in. (There are exceptions for multi-project tutorials.)
- Language specific projects are in sibling folders using names following our inline code convention: `csharp`, `fsharp`, and `vb`. These folders are not needed in the language guide area, where only one language is assumed.
- When multiple versions of a project are needed to represent different steps in a tutorial, inline code can (maybe should) be used for interim steps. Only the final version is included using the include syntax.

[...]

PR #17337, opened 2020-03-04
C# reference: moved snippets for the revised articles

pkulikov:
Revised articles are those about operators and built-in types.
Related to #17297
@BillWagner @Thraka Is that the correct understanding of the new guidelines?


BillWagner (approved):
This LGTM @pkulikov 

Yes, this is one of the correct locations.  We're trying to be flexible enough for different reasonable locations. We have some areas where there is one project per folder, others where there is one project per article, and some, like XAML based projects, where there are multiple steps per article. One standard won't work well, but we want enough consistency to make it easy to maintain.

Thanks again. I'll :shipit: now.

PR #17338, opened 2020-03-04
Add custom metadata for ML.NET tooling articles.

natke:
Add metadata to distinguish between API and tooling docs for docs performance reporting


luisquintanilla (approved):
Looks good.

PR #17341, opened 2020-03-05
Apply dotnetcli dev lang to command block

scottaddie:
Use the `dotnetcli` dev lang so that we get syntax highlighting for the .NET Core CLI command.


BillWagner (approved):
This LGTM @scottaddie 

You can :shipit: when ready.

PR #17347, opened 2020-03-05
ref locals: mention ref assignment operator

pkulikov:


BillWagner (approved):
I like how you've highlighted this new feature inline rather than with the link at the bottom.

I'll :shipit: now.

PR #17349, opened 2020-03-05
TLS guidance - .NET Framework run on vs. targeting

karelz:
Clarify .NET Framework versions targeting vs. "run on"


davidsh (approved):
Overall LGTM. I like that we are focusing on targeting instead of what binaries it runs on.

It might be useful also if a sanity check was done by writing a small program, targetted to different .NET versions and printing out the default values of `System.Net.ServicePointManager.SecurityProtocol`

gewarren (approved):

karelz:
@davidsh I did the testing as part of this investigation:
* https://github.com/microsoft/dotnet/issues/1138#issuecomment-589779845
* https://github.com/microsoft/dotnet/issues/1138#issuecomment-589780844

PR #17368, opened 2020-03-06
Fix a few rules

BillWagner:
After adding the global rule, we noticed two issues.

First, it seems the ** syntax doesn't work as exptected. The docs say just ending with a trailing slash is better

Secondly, the newly imported snippets did not have any entries, so the entire team was tagged for any PR with samples changes.

This should fix those issues

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -19,157 +19,183 @@
 # This should make it easy to add new rules without breaking existing ones.
 
 ######## API landing page ###########
-/api/** @dotnet/docs
+/api/ @dotnet/docs
 
 ############ WHATS NEW ###############
-/docs/whats-new/** @billwagner
+/docs/whats-new/ @billwagner
 
 ############  GUIDES  ################
 # .NET Core
-/docs/core/** @dotnet/docs
+/docs/core/ @dotnet/docs
 
 # .NET Framework
-/docs/framework/** @dotnet/docs
+/docs/framework/ @dotnet/docs
 
 # .NET Standard
-/docs/standard/** @gewarren
+/docs/standard/ @gewarren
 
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -19,157 +19,183 @@
 # This should make it easy to add new rules without breaking existing ones.
 
 ######## API landing page ###########
-/api/** @dotnet/docs
+/api/ @dotnet/docs
```
Thraka: I think you still want ** on these items, otherwise it won't include child folders+items

.github/CODEOWNERS
```diff
@@ -19,157 +19,183 @@
...
+/docs/architecture/blazor-for-web-forms-developers/ @danroth27
 # gRPC:
-/docs/architecture/grpc-for-wcf-developers/** @shirhatti
+/docs/architecture/grpc-for-wcf-developers/ @shirhatti
+
+################# Samples folder ##############
+
+## These have been pulled directly from the samples repo:
+
+/samples/snippets/csharp/ @BillWagner
```
Thraka: again, you'll need ** on these to include child folders+items

.github/CODEOWNERS
```diff
@@ -19,157 +19,183 @@
...
+/api/ @dotnet/docs
 
 ############ WHATS NEW ###############
-/docs/whats-new/** @billwagner
+/docs/whats-new/ @billwagner
 
 ############  GUIDES  ################
 # .NET Core
-/docs/core/** @dotnet/docs
+/docs/core/ @dotnet/docs
```
Thraka: I think this is good. It just includes the root level items and assigns that to the team.

Thraka (approved):

PR #17380, opened 2020-03-08
Fixed the F# active patterns sample

Thecentury:
## Summary

Fixed F# active patterns sample — I guess, this is what was meant, otherwise the result of the first match is ignored.


cartermp (approved):
Thanks! Yes, this is an old sample that was definitely not doing things adequately.

PR #17389, opened 2020-03-09
Remove mairaw from global author/ms.author

gewarren:


docfx.json
```diff
@@ -176,7 +176,7 @@
       "author": {
         "docs/architecture/**/**.md": "nishanil",
         "docs/architecture/grpc-for-wcf-developers/**/**.md": "shirhatti",
-        "docs/core/**/**.md": "mairaw",
+        "docs/core/**/**.md": "dotnet/docs",
```
gewarren: Not sure if using a GH team will work here, but thought I'd try.

docfx.json
```diff
@@ -192,20 +192,20 @@
...
+        "docs/framework/get-started/**/**.md": "gewarren",
         "docs/framework/install/**/**.md": "Thraka",
         "docs/framework/migration-guide/**/**.md": "gewarren",
-        "docs/framework/misc/**/**.md": "mairaw",
-        "docs/framework/net-native/**/**.md": "mairaw",
+        "docs/framework/misc/**/**.md": "gewarren",
+        "docs/framework/net-native/**/**.md": "gewarren",
         "docs/framework/network-programming/**/**.md": "karelz",
-        "docs/framework/performance/**/**.md": "mairaw",
+        "docs/framework/performance/**/**.md": "billwagner",
```
gewarren: I used @BillWagner since he owns the threading docs which are also about perf in a way, but let me know if it should be someone else or just the dotnet/docs team.

docfx.json
```diff
@@ -176,7 +176,7 @@
       "author": {
         "docs/architecture/**/**.md": "nishanil",
         "docs/architecture/grpc-for-wcf-developers/**/**.md": "shirhatti",
-        "docs/core/**/**.md": "mairaw",
+        "docs/core/**/**.md": "tdykstra",
```
gewarren: Since we can't use the dotnet/docs GH team for author, I just rotated through each team member for /core, /framework, /standard/glob-loc, /standard/linq.

.github/CODEOWNERS
```diff
@@ -142,7 +142,7 @@
 # Exceptions
 /docs/standard/exceptions/ @tdykstra
 # GC
-/docs/standard/garbage-collection/ @BillWagner
+/docs/standard/garbage-collection/ @gewarren
```
BillWagner: cool

docfx.json
BillWagner: I just checked the [contributor guide](https://docsmetadatatool.azurewebsites.net/allowlists#ms.author). We can use our team alias, `dotnetcontent` here instead.  That would be better than rotating assignments. Thoughts?

BillWagner (approved):
This looks great @gewarren 

I checked the contributor guide. I suggested that we use our team alias instead of rotating assignments for ms.author. What are your thoughts here?

docfx.json
gewarren: Sure for ms.author, but even for author? It has to be a GitHub ID.

BillWagner: Ahh. According to the docfx folks, `@dotnet/docs` should work.

gewarren: I did try that on line 179 in [this commit](https://github.com/dotnet/docs/pull/17389/commits/26c282fd90e304154164d972556868bd0c24747b#diff-f14cdab6f1d762de4c71972b81dd008cR179), but the build had a warning that it wasn't a valid GitHub account.

BillWagner: *sigh* After I had the first response that GitHub teams were supported, one of the engineers wrote back to correct me that they aren't. Let's go with your original plan of cycling through each of us.

PR #17403, opened 2020-03-09
Size of the example struct is incorrect

ghost:
The example struct's size can be 8 bytes (32-bit), 16 bytes (64-bit with padding) or 12 bytes (64-bit with Pack = 1).

## Summary

The doc says that the example struct's size is 8 bytes. This is not always the case. I've tried to rephrase the sentence to avoid mentioning any concrete numbers (to be somewhat correct, at least 2 are needed, which in my opinion would bloat the text with extraneous information) while emphasizing the point being made (that regular arrays in structs are just references).


BillWagner (approved):
That is a great way to phrase the change @RtmK 

I'll :shipit: now. 

Thanks for helping us improve our docs!

PR #17405, opened 2020-03-09
configure dependabot

BillWagner:
Add a generated dependabot config file

Since we're moving the snippets here, we should keep the dependencies up to date.


tdykstra (approved):
LGTM

PR #17410, opened 2020-03-09
fix: MD009 in docs/csharp

nschonni:
## Summary

Enabled MD009 and ran Markdownlint with the `--fix` flag. Spit folders under docs/csharp into individual commits


BillWagner (approved):
Thanks @nschonni 

This was straightforward to review: turning on "hide whitespace" made it super easy. 👍

Issue #12512, opened 2019-05-24
Cannot implicitly convert type 'System.Data.Entity.Dbset to System.Data.Entity.Core.Objects.ObjectQuery'


Perustaja:
Running this code verbatim returns this error, everything goes smooth until

ObjectQuery&lt;Product&gt; products = dataEntities.Products;

returns this error, seems some others had the same issue. I have essentially the same code written out verbatim but with different names. 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 58339b81-2a36-2bb4-2fab-8d01596d5a70
* Version Independent ID: 75bc38d0-eba9-ea58-4cc5-ba38d90c613e
* Content: [Walkthrough: Display Data from a SQL Server Database in a DataGrid Control](https://docs.microsoft.com/en-us/dotnet/framework/wpf/controls/walkthrough-display-data-from-a-sql-server-database-in-a-datagrid-control#feedback)
* Content Source: [docs/framework/wpf/controls/walkthrough-display-data-from-a-sql-server-database-in-a-datagrid-control.md](https://github.com/dotnet/docs/blob/master/docs/framework/wpf/controls/walkthrough-display-data-from-a-sql-server-database-in-a-datagrid-control.md)
* Product: **dotnet-framework**
* Technology: **dotnet-wpf**
* GitHub Login: @dotnet-bot
* Microsoft Alias: **dotnetcontent**

PR #17421, opened 2020-03-10
Update MainWindow.xaml.cs

Da-vy:
## Summary

I actually found the solution here:
https://stackoverflow.com/a/12252423/12371154

Fix for this ticket: dotnet/docs#12512


BillWagner (approved):
Thanks @Da-vy 

This is great, and I'll :shipit: now.

PR #17422, opened 2020-03-10
Fix typo/syntax error in event tutorial

MatthewRiggott:
ValueChanged?.Invoke(this, e) should be onValueChanged?.Invoke(this, e)
## Summary

Fixing a typo as the given tutorial code does not compile.


Thraka (approved):
Thanks for spotting that!

PR #17454, opened 2020-03-12
Add Common Usage Patterns for Extension Methods and Update Guidance for Their Use

bsstahl:
## Summary

Add 3 Common Usage Patterns for extension methods and update the guidance on using Extension Methods to make it clear that, while not the top go-to option, using Extension Methods is a perfectly valid choice for many implementation scenarios.


docs/csharp/programming-guide/classes-and-structs/extension-methods.md
```diff
@@ -59,10 +60,38 @@ using System.Linq;
...
+public class DomainEntity
+{
+    public Int32 Id { get; set; }
+    public String FirstName { get; set; }
+    public String LastName { get; set; }
+}
+
+static class DomainEntityExtensions
+{
+    static String FullName(this DomainEntity value)
```
BillWagner: I have the same comment on all these samples. Our coding standard is to use the built in C# types rather than the library types:

```suggestion
    public int Id { get; set; }
    public string FirstName { get; set; }
    public string LastName { get; set; }
}

static class DomainEntityExtensions
{
    static string FullName(this DomainEntity value)
```

BillWagner (changes requested):
This is a great addition @bsstahl I like the way you've described the design options, and pointed to scenarios where extension methods are a great option.

I had a few small comments to address so that this code matches the coding style used throughout the C# samples. If you apply my suggestions, this is ready to merge.

docs/csharp/programming-guide/classes-and-structs/extension-methods.md
BillWagner: Thanks for making these updates @bsstahl

BillWagner (approved):

BillWagner:
After taking the suggestions, I've approved this PR, and I'll :shipit: now.

PR #17466, opened 2020-03-15
Fix inverted bool value for pragma case_sensitive_like

dionrhys:
## Summary

I think the instruction for enabling case-sensitive `LIKE` uses the wrong value in the current docs.

> You can easily make the LIKE operator case-sensitive by using the following pragma statement:

Current:

```sql
PRAGMA case_sensitive_like = 0
```

Should be:

```sql
PRAGMA case_sensitive_like = 1
```

https://sqlite.org/pragma.html#pragma_case_sensitive_like


bricelam (approved):

bricelam:
Oops. 😏 Off by one error?

dionrhys:
Likely particles from cosmic rays 😉 Cheers!

PR #17474, opened 2020-03-16
build: Trigger Markdownlint if CI changed

nschonni:
## Summary

Minor tweak to the build config, so that if the config YML, but no MD files are touched, the build is still tested


BillWagner (approved):
This is a good change @nschonni 
Thanks. I'll :shipit: now.

PR #17477, opened 2020-03-17
Update performing-arithmetic-operations.md

ry8806:
## Summary

Improved readability (removed uses of consecutive this...this)

Also (ln73) links to a method, but says "the xx class" - so corrected that link too (now goes to the TimeZoneInfo class)


Thraka (approved):
Thanks a bunch for fixing these issues!

PR #17482, opened 2020-03-17
Disables the C# addin for vscode when repo is opened

Thraka:
## Summary

Disables the C# (omnisharp) addin from processing the samples directory. If you load the repo root in VSCode, C# will start loading every csharp project in the repo. This config disables that.

You can still open individual project folders in vscode and the C# addin will load.


tdykstra (approved):
Should there be a newline at the end of the file?

PR #17484, opened 2020-03-17
Fix typo in code snippet

abonie:
`do` should be used here, because the if expression is missing an else branch - as explained in F# language reference where this code snippet is used: https://github.com/dotnet/docs/edit/master/docs/fsharp/language-reference/sequences.md


cartermp (approved):
Thanks!

PR #17487, opened 2020-03-17
clarify readonly members

### LINKED ISSUE #16510: C# 8.0 readonly members vs. "Never use mutable structs as in"
I think the documentation might contain conflicting or confusing statements about using mutable structs as in arguments.

The section [Declare readonly members when a struct can't be immutable
](https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#declare-readonly-members-when-a-struct-cant-be-immutable) informs that:
> In C# 8.0 and later, when a struct type is mutable, you should declare members that don't cause mutation to be readonly.
> ...
> Adding the readonly modifier to members that don't mutate state provides two related benefits. First, the compiler enforces your intent. That member can't mutate the struct's state, nor can it access a member that isn't also marked readonly. Second, **the compiler won't create defensive copies of in parameters when accessing a readonly member**. The compiler can make this optimization safely because it guarantees that the struct is not modified by a readonly member.

However, the section [Never use mutable structs as in in argument](https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#never-use-mutable-structs-as-in-in-argument) then says the following:
> The techniques described above explain how to avoid copies by returning references and passing values by reference. These techniques work best when the argument types are declared as readonly struct types. Otherwise, the compiler must create defensive copies in many situations to enforce the readonly-ness of any arguments.
> ...
> On first examination, you may have thought these accesses were safe. After all, a get accessor shouldn't modify the state of the object. **But there's no language rule that enforces that. It's only a common convention.** Any type could implement a get accessor that modified the internal state. **Without some language guarantee, the compiler must create a temporary copy of the argument before calling any member.**

From my point of view, the reader could be confused about whether or not the defensive copies are made, and what the best practice is. Moreover, the _Never use mutable structs as in in argument_ section references the `Point3D` definition, which has already been enhanced with `readonly` modifiers in the previous section.

Also, the latter section mentions that
> You shouldn't pass a nullable value type as an in argument. The Nullable\<T\> type isn't declared as a read-only struct.

I would like to ask if there have been any changes to this behavior with C# 8.0.

---
#### Document Details

* Content: [Write safe and efficient C# code](https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#declare-readonly-members-when-a-struct-cant-be-immutable)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixes #16510

Clarify the benefits of readonly members and readonly structs.

### DIFF
--- a/docs/csharp/write-safe-efficient-code.md
+++ b/docs/csharp/write-safe-efficient-code.md
@@ -67,7 +67,7 @@ Follow this recommendation whenever your design intent is to create an immutable
 
 ## Declare readonly members when a struct can't be immutable
 
-In C# 8.0 and later, when a struct type is mutable, you should declare members that don't cause mutation to be `readonly`. For example, the following is a mutable variation of the 3D point structure:
+In C# 8.0 and later, when a struct type is mutable, you should declare members that don't cause mutation to be `readonly`. Consider a different application that needs a 3D point structure, but must support mutability. The following version of the 3D point structure adds the `readonly` modifier only to those members that don't modify the structure. Follow this example when your design must support modifications to the struct by some members, but you still want the benefits of enforcing readonly on some members:
 
 ```csharp
 public struct Point3D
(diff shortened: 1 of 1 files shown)


tdykstra (approved):
LGTM, makes the point clearly.

PR #17508, opened 2020-03-19
Note that Rune is only in Core 3.x and 5

tdykstra:
Follow-up to #17448 
Also addresses some loc issues.

### DIFF
--- a/docs/standard/base-types/character-encoding-introduction.md
+++ b/docs/standard/base-types/character-encoding-introduction.md
@@ -2,7 +2,7 @@
 title: Introduction to character encoding in .NET
 description: Learn about character encoding and decoding in .NET.
 ms.date: 03/09/2020
-no-loc: [Rune, rune, runes, char, string]
+no-loc: [Rune, rune, runes, char, chars, string, strings]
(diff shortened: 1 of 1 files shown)


docs/standard/base-types/character-encoding-introduction.md
```diff
@@ -178,7 +179,7 @@ The following diagram illustrates the scalar value code points.
 
 ### The Rune type as a scalar value
 
-The `Rune` type, introduced in .NET Core 3.0, represents a Unicode scalar value.
+The <xref:System.Text.Rune?displayProperty=fullName> type represents a Unicode scalar value in .NET Core 3.x and .NET 5. **`Rune` is not available in .NET Core 2.x or .NET Framework 4.x.**
```
pkulikov: To make it future-proof?
```suggestion
Beginning with .NET Core 3.0, the <xref:System.Text.Rune?displayProperty=fullName> type represents a Unicode scalar value. **`Rune` is not available in .NET Core 2.x or .NET Framework 4.x.**
```

Thraka (approved):

PR #17510, opened 2020-03-19
Fix wrong note about default implementation

DAXaholic:
The text states that the default is `PowerStatus.ACPower`, however In the actual implementation the default is `PowerStatus.NoPower`.

https://github.com/dotnet/docs/blob/12d15a050e67a7fa66d743cb96c9ee331daf3c36/samples/snippets/csharp/tutorials/mixins-with-interfaces/ILight.cs#L24


BillWagner (approved):
Thanks for spotting this and submitting the fix @DAXaholic 

Thanks for contributing to docs! 🎆

PR #17511, opened 2020-03-19
Update private-protected.md

cloudn9:
## Summary

Deleted the "Compile with rows" info as suggested by @Youssef1313

Should Fix ##16748


BillWagner (approved):
Thanks for updating this @cloudn9 

I'll :shipit: now.

PR #17512, opened 2020-03-19
Add switch expression to language reference

### LINKED ISSUE #14889: Documentation on the switch expression is missing
This page is about the switch statement. Is it going to be expanded to include documentation about switch expression (https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#more-patterns-in-more-places) or is it going to be a separate page?

---
#### Document Details

* ID: 77d9cc47-5caf-184a-0ad6-cb968850be7a
* Version Independent ID: f058bf17-04e9-9e23-05c7-66c1864445ad
* Content: [C# switch statement](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/switch#feedback)
* Content Source: [docs/csharp/language-reference/keywords/switch.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/keywords/switch.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixes #14889 

Add a new article for the language reference on switch expressions.

Also, provide links between the articles on the switch statement and switch expression. I made that choice because the switch statement article was already rather long.

[Internal review link](https://review.docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression?branch=pr-en-us-17512)

### DIFF
--- a/docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
+++ b/docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
@@ -24,6 +24,29 @@ namespace operators
     {
         public static void Examples()
         {
+            InitialExample();
+
+            var collection = new int[]{1,2,3,4,5,6,7,8,9};
+            TypeExample(collection);
+            RecursiveExample(collection);
+            RecursiveExample(collection[0..0]);
+            RecursiveExample(collection[0..1]);
+            RecursiveExample(collection[0..2]);
+            CaseGuardExample(collection.AsEnumerable());
+            CaseGuardExample(collection[0..0].AsEnumerable());
+            CaseGuardExample(collection[0..1].AsEnumerable());
+            CaseGuardExample(collection[0..2].AsEnumerable());
+            try {
+                ExhaustiveExample(default(List<int>));
+            } catch (ArgumentNullException e)
+            {
+                Console.WriteLine($"Caught expected exception: {e.Message}");
+            }
+
+        }
+
+        private static void InitialExample()
+        {
             // <SnippetBasicStructure>
             var direction = Directions.Right;
             Console.WriteLine($"Map view direction is {direction}");
@@ -38,6 +61,76 @@ namespace operators
             Console.WriteLine($"Cardinal orientation is {orientation}");
             // </SnippetBasicStructure>
         }
+
+        private static void TypeExample<T>(IEnumerable<T> sequence)
+        {
+            // <SnippetTypePattern>
+            var third = sequence switch
+            {
+                System.Array array => array.GetValue(2),
+                IList<T> list => list[2],
+                IEnumerable<T> seq => seq.Skip(2).First(),
+            };
+            Console.WriteLine(third);
+            // </SnippetTypePattern>
+        }
+
+        private static void RecursiveExample<T>(IEnumerable<T> sequence)
+        {
+            // <SnippetRecursivePattern>
+            var third = sequence switch
+            {
+                System.Array { Length : 0}        => default,
+                System.Array { Length : 1} array => array.GetValue(0),
+                System.Array { Length : 2} array => array.GetValue(1),
+                System.Array array               => array.GetValue(2),
+                IList<T> list                    => list[2],
+                IEnumerable<T> seq               => seq.Skip(2).First(),
+            };
+            Console.WriteLine(third);
+            // </SnippetRecursivePattern>
+        }
+
+        private static void CaseGuardExample<T>(IEnumerable<T> sequence)
+        {
+            // <SnippetGuardCase>
+            var third = sequence switch
+            {
+                System.Array { Length : 0}                => default,
+                System.Array { Length : 1} array          => array.GetValue(0),
+                System.Array { Length : 2} array          => array.GetValue(1),
+                System.Array array                        => array.GetValue(2),
+                IEnumerable<T> list when !list.Any()      => default,
+                IEnumerable<T> list when list.Count() < 3 => list.Last(),
+                IList<T> list                             => list[2],
+                IEnumerable<T> seq                        => seq.Skip(2).First(),
+            };
+            Console.WriteLine(third);
+            // </SnippetGuardCase>
+        }
+
+        private static void ExhaustiveExample<T>(IEnumerable<T> sequence)
+        {
+            // <SnippetExhaustive>
+            var third = sequence switch
+            {
+                System.Array { Length : 0}       => default,
+                System.Array { Length : 1} array => array.GetValue(0),
+                System.Array { Length : 2} array => array.GetValue(1),
+                System.Array array               => array.GetValue(2),
+                IEnumerable<T> list 
+                    when !list.Any()             => default,
+                IEnumerable<T> list 
+                    when list.Count() < 3        => list.Last(),
+                IList<T> list                    => list[2],
+                null                             => throw new ArgumentNullException("null input"),
+                _                                => sequence.Skip(2).First(),
+            };
+            Console.WriteLine(third);
+            // </SnippetExhaustive>
+        }
+
+        
     }
 }
 
--- a/docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
+++ b/docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
@@ -70,68 +70,68 @@ namespace operators
         private static void TypeExample<T>(IEnumerable<T> sequence)
         {
             // <SnippetTypePattern>
-            var third = sequence switch
+            T thirdOrDefault = sequence switch
             {
                 System.Array array => array.GetValue(2),
-                IList<T> list => list[2],
+                IList<T> list      => list[2],
                 IEnumerable<T> seq => seq.Skip(2).First(),
             };
-            Console.WriteLine(third);
+            Console.WriteLine(thirdOrDefault);
             // </SnippetTypePattern>
         }
 
         private static void RecursiveExample<T>(IEnumerable<T> sequence)
         {
             // <SnippetRecursivePattern>
-            var third = sequence switch
+            T thirdOrDefault = sequence switch
             {
-                System.Array { Length : 0}        => default,
+                System.Array { Length : 0}       => default(T),
                 System.Array { Length : 1} array => array.GetValue(0),
                 System.Array { Length : 2} array => array.GetValue(1),
                 System.Array array               => array.GetValue(2),
                 IList<T> list                    => list[2],
                 IEnumerable<T> seq               => seq.Skip(2).First(),
             };
-            Console.WriteLine(third);
+            Console.WriteLine(thirdOrDefault);
             // </SnippetRecursivePattern>
         }
 
         private static void CaseGuardExample<T>(IEnumerable<T> sequence)
         {
             // <SnippetGuardCase>
-            var third = sequence switch
+            T thirdOrDefault = sequence switch
             {
-                System.Array { Length : 0}                => default,
+                System.Array { Length : 0}                => default(T),
                 System.Array { Length : 1} array          => array.GetValue(0),
                 System.Array { Length : 2} array          => array.GetValue(1),
                 System.Array array                        => array.GetValue(2),
-                IEnumerable<T> list when !list.Any()      => default,
+                IEnumerable<T> list when !list.Any()      => default(T),
                 IEnumerable<T> list when list.Count() < 3 => list.Last(),
                 IList<T> list                             => list[2],
                 IEnumerable<T> seq                        => seq.Skip(2).First(),
             };
-            Console.WriteLine(third);
+            Console.WriteLine(thirdOrDefault);
             // </SnippetGuardCase>
         }
 
         private static void ExhaustiveExample<T>(IEnumerable<T> sequence)
         {
             // <SnippetExhaustive>
-            var third = sequence switch
+            T thirdOrDefault = sequence switch
             {
-                System.Array { Length : 0}       => default,
+                System.Array { Length : 0}       => default(T),
                 System.Array { Length : 1} array => array.GetValue(0),
                 System.Array { Length : 2} array => array.GetValue(1),
                 System.Array array               => array.GetValue(2),
                 IEnumerable<T> list 
-                    when !list.Any()             => default,
+                    when !list.Any()             => default(T),
                 IEnumerable<T> list 
                     when list.Count() < 3        => list.Last(),
                 IList<T> list                    => list[2],
                 null                             => throw new ArgumentNullException(nameof(sequence)),
                 _                                => sequence.Skip(2).First(),
             };
-            Console.WriteLine(third);
+            Console.WriteLine(thirdOrDefault);
             // </SnippetExhaustive>
         }
 
--- a/docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
+++ b/docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
@@ -4,27 +4,48 @@ using System.Linq;
 
 namespace operators
 {
-    public enum Directions
-    {
-        Up,
-        Down,
-        Right,
-        Left
-    }
 
-    public enum Orientation
+    // <SnippetBasicStructure>
+    public static class SwitchExample
     {
-        North,
-        South,
-        East,
-        West
+        public enum Directions
+        {
+            Up,
+            Down,
+            Right,
+            Left
+        }
+
+        public enum Orientation
+        {
+            North,
+            South,
+            East,
+            West
+        }
+    
+        public static void Main()
+        {
+            var direction = Directions.Right;
+            Console.WriteLine($"Map view direction is {direction}");
+
+            var orientation = direction switch
+            {
+                Directions.Up    => Orientation.North,
+                Directions.Right => Orientation.East,
+                Directions.Down  => Orientation.South,
+                Directions.Left  => Orientation.West,
+            };
+            Console.WriteLine($"Cardinal orientation is {orientation}");
+        }
     }
+    // </SnippetBasicStructure>
 
     public static class SwitchExpressions
     {
         public static void Examples()
         {
-            InitialExample();
+            SwitchExample.Main();
 
             var collection = new int[]{1,2,3,4,5,6,7,8,9};
             TypeExample(collection);
@@ -45,22 +66,6 @@ namespace operators
 
         }
 
-        private static void InitialExample()
-        {
-            // <SnippetBasicStructure>
-            var direction = Directions.Right;
-            Console.WriteLine($"Map view direction is {direction}");
-
-            var orientation = direction switch
-            {
-                Directions.Up    => Orientation.North,
-                Directions.Right => Orientation.East,
-                Directions.Down  => Orientation.South,
-                Directions.Left  => Orientation.West,
-            };
-            Console.WriteLine($"Cardinal orientation is {orientation}");
-            // </SnippetBasicStructure>
-        }
 
         private static void TypeExample<T>(IEnumerable<T> sequence)
         {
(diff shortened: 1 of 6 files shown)


docs/csharp/language-reference/operators/switch.md
```diff
@@ -0,0 +1,49 @@
+---
+title: "switch expressions - C# reference"
```
pkulikov: All other articles are in single. "default value expressions" is an exception because there are two kinds of that.
```suggestion
title: "switch expression - C# reference"
```

docs/csharp/language-reference/operators/switch.md
```diff
@@ -0,0 +1,49 @@
...
+description: Learn how to use the C# switch expression for pattern matching and other data introspection
+ms.date: 03/19/2020
+---
+# switch expressions (C# reference)
```
pkulikov:
```suggestion
# switch expression (C# reference)
```

docs/csharp/language-reference/operators/switch.md
```diff
@@ -0,0 +1,49 @@
...
+Recursive patterns can examine properties of the range expression, but can't execute arbitrary code. You can use a *case guard*, specified in a `when` clause, to provide similar checks for other sequence types:
+
+:::code language="csharp" source="snippets/SwitchExpressions.cs" id="SnippetGuardCase":::
+
+Finally, you can add the `_` pattern and the `null` pattern to catch arguments that aren't processed by any other switch expression arm. That makes the switch expression *exhaustive*, meaning any possible value of the range expression is handled. The following example adds those expression arms:
+
+:::code language="csharp" source="snippets/SwitchExpressions.cs" id="SnippetExhaustive" interactive="csharp-interactive":::
+
+The preceding example adds a `null` pattern, and changes the `IEnumerable<T>` type pattern to a `_` pattern. The `null` pattern provides a null check as a switch expression arm. The expression for that arm throws a null argument exception. The `_` pattern matches all inputs that haven't been matched by a previous arm. It must come after the `null` check, or it would match `null` inputs.
+
```
pkulikov: Is there a language spec proposal available to link (instead of the C# language reference)?

docs/csharp/language-reference/operators/switch.md
```diff
@@ -0,0 +1,49 @@
...
+The preceding sample shows the basic elements of a switch expression:
+
+- The *range expression*: The preceding example uses the variable `direction` as the range expression.
+- The *switch expression arms*: Each switch expression arm contains a *pattern*, an optional *case guard*, the `=>` token, and an *expression*.
+
+The result of the *switch expression* is the value of the expression of the first *switch expression arm* whose *pattern* matches the *range expression* and whose *cause guard*, if present, evaluates to `true`. The *expression* on the right of the `=>` token can't be an expression statement.
+
+Each *switch expression arm* is evaluated in text order. The compiler issues an error when later switch expression arms can't be chosen because an earlier *switch expression arm* matches all its values.
+
+## Patterns and case guards
```
pkulikov: My following thoughts are definitely not something to act upon in this PR. I'm interested in how the pattern matching info is going to be consolidated in future.

In general, I observe some amount of duplication in the docs about patterns. Is that intentional? I wonder: are the patterns in this section unique to the `switch` expression? I guess one might use them with the `switch` statement or even with `is`. Ideally, it can be beneficial to have only one reference article that describes all available patterns (maybe in the order of introduction to the language). Then, the articles about `is`, `switch` (both statement and expression) would link that article.

docs/csharp/language-reference/operators/switch.md
```diff
@@ -0,0 +1,49 @@
...
+Finally, you can add the `_` pattern and the `null` pattern to catch arguments that aren't processed by any other switch expression arm. That makes the switch expression *exhaustive*, meaning any possible value of the range expression is handled. The following example adds those expression arms:
+
+:::code language="csharp" source="snippets/SwitchExpressions.cs" id="SnippetExhaustive" interactive="csharp-interactive":::
+
+The preceding example adds a `null` pattern, and changes the `IEnumerable<T>` type pattern to a `_` pattern. The `null` pattern provides a null check as a switch expression arm. The expression for that arm throws a null argument exception. The `_` pattern matches all inputs that haven't been matched by a previous arm. It must come after the `null` check, or it would match `null` inputs.
+
+## See also
+
+- [C# reference](../index.md)
+- [C# expressions and operators](index.md)
```
pkulikov: That article is still named "C# operators" and talks about operators only:
```suggestion
- [C# operators](index.md)
```
(I'm planning to move the material about expressions from the programming guide to that article; then, I would rename the article.)

docs/csharp/toc.yml
```diff
@@ -1611,6 +1611,9 @@
     - name: Member access operators and expressions
       href: language-reference/operators/member-access-operators.md
       displayName: "., [], ?., ?[], (), indexer, null-conditional, Elvis, invocation, ^, index from end, hat, .., range"
+    - name: switch expressions
```
pkulikov:
```suggestion
    - name: switch expression
```

pkulikov: Is it fine to move the TOC entry for the `switch` expression lower: just under `stackalloc`? The entries are sorted alphabetically, but only inside each of the two large groups: Overview articles and individual operator/expression articles are sorted separately.

docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
```diff
@@ -0,0 +1,137 @@
+using System;
```
Thraka: This file is supposed to be in a sub folder named after the article. (well all of these snippet files should be)

Thraka (approved):
Approved apart from the `snippets/article` folder naming.

pkulikov:
> Approved apart from the `snippets/article` folder naming.

@Thraka how would it help? I see only a large duplication between article name and the file name, because each folder would contain only one file. Is it necessary?

Also, according to #17297:

> For conceptual, we would put the project(s) in a snippets subfolder. The snippets subfolder should be in the same folder as the article by default. Where a snippets project contains code used in multiple articles, the snippets folder should be in the nearest parent folder of all relevant articles.

From that, I conclude that the current structure is fine.

docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
```diff
@@ -0,0 +1,137 @@
...
+            Console.WriteLine(third);
+            // </SnippetTypePattern>
+        }
+
+        private static void RecursiveExample<T>(IEnumerable<T> sequence)
+        {
+            // <SnippetRecursivePattern>
+            var third = sequence switch
+            {
+                System.Array { Length : 0}        => default,
```
pkulikov: From the above lines it's not easy (not at one glance) to figure out the type of `third`. I  suggest one of the following changes:
- `var` -> `T`
- `third` -> `thirdElement`
- `default` -> `default(T)`

docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
```diff
@@ -0,0 +1,137 @@
...
+            Console.WriteLine(third);
+            // </SnippetRecursivePattern>
+        }
+
+        private static void CaseGuardExample<T>(IEnumerable<T> sequence)
+        {
+            // <SnippetGuardCase>
+            var third = sequence switch
+            {
+                System.Array { Length : 0}                => default,
```
pkulikov: Same comment as to the above snippet: not easy to see the type of `third`.

docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
```diff
@@ -0,0 +1,137 @@
...
+                System.Array { Length : 0}       => default,
+                System.Array { Length : 1} array => array.GetValue(0),
+                System.Array { Length : 2} array => array.GetValue(1),
+                System.Array array               => array.GetValue(2),
+                IEnumerable<T> list 
+                    when !list.Any()             => default,
+                IEnumerable<T> list 
+                    when list.Count() < 3        => list.Last(),
+                IList<T> list                    => list[2],
+                null                             => throw new ArgumentNullException("null input"),
```
pkulikov: In the case of [one-parameter constructor](https://docs.microsoft.com/en-us/dotnet/api/system.argumentnullexception.-ctor?view=netframework-4.8#System_ArgumentNullException__ctor_System_String_), the argument should be the name of a parameter:
```suggestion
                null                             => throw new ArgumentNullException(nameof(sequence)),
```

docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        private static void TypeExample<T>(IEnumerable<T> sequence)
+        {
+            // <SnippetTypePattern>
+            var third = sequence switch
+            {
+                System.Array array => array.GetValue(2),
+                IList<T> list => list[2],
+                IEnumerable<T> seq => seq.Skip(2).First(),
```
pkulikov: Just a note: this is the only example where `=>` tokens are not aligned.

docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
```diff
@@ -0,0 +1,137 @@
...
+            };
+            Console.WriteLine(third);
```
pkulikov: I would move all the `Console.WriteLine` lines out of the snippets. Those snippets are not runnable in the published article, so those lines don't add much info. (If the method definitions are expression-bodied, those lines would be removed.)

docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
```diff
@@ -0,0 +1,137 @@
...
+                Directions.Left  => Orientation.West,
+            };
+            Console.WriteLine($"Cardinal orientation is {orientation}");
+            // </SnippetBasicStructure>
+        }
+
+        private static void TypeExample<T>(IEnumerable<T> sequence)
+        {
+            // <SnippetTypePattern>
+            var third = sequence switch
```
pkulikov: When published, readers wouldn't now the type of `sequence`. Is it important to know that `sequence` is `IEnumerable<T>`?

Alternative is to include the method signature into the snippet AND make a method definition **expression-bodied**.

docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
```diff
@@ -0,0 +1,137 @@
...
+                IEnumerable<T> seq => seq.Skip(2).First(),
+            };
+            Console.WriteLine(third);
+            // </SnippetTypePattern>
+        }
+
+        private static void RecursiveExample<T>(IEnumerable<T> sequence)
+        {
+            // <SnippetRecursivePattern>
+            var third = sequence switch
```
pkulikov: nit-pick: should be named differently because, in the case of short arrays, the result is not third. However, if expression-bodied, the variable name is not concern anymore.

docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
```diff
@@ -0,0 +1,137 @@
...
+                IEnumerable<T> seq               => seq.Skip(2).First(),
+            };
+            Console.WriteLine(third);
+            // </SnippetRecursivePattern>
+        }
+
+        private static void CaseGuardExample<T>(IEnumerable<T> sequence)
+        {
+            // <SnippetGuardCase>
+            var third = sequence switch
```
pkulikov: again nit-pick: it's not always third

docs/csharp/language-reference/operators/switch.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Basic example
+
+The `switch` expression provides for `switch`-like semantics in an expression context. It provides a concise syntax when the switch arms produce a value. The following example shows the structure of a switch expression. It translates values from an `enum` representing visual directions in an online map to the corresponding cardinal direction:
+
+:::code language="csharp" source="snippets/SwitchExpressions.cs" id="SnippetBasicStructure" interactive="csharp-interactive":::
+
+The preceding sample shows the basic elements of a switch expression:
+
+- The *range expression*: The preceding example uses the variable `direction` as the range expression.
```
pkulikov: Is it a widely-used name: *range expression*? It has several meanings: I would call `1..^1` also a range expression. If we consider `switch` as operator, then it's *argument expression*.

docs/csharp/language-reference/operators/switch.md
```diff
@@ -0,0 +1,49 @@
...
+- The *switch expression arms*: Each switch expression arm contains a *pattern*, an optional *case guard*, the `=>` token, and an *expression*.
+
+The result of the *switch expression* is the value of the expression of the first *switch expression arm* whose *pattern* matches the *range expression* and whose *cause guard*, if present, evaluates to `true`. The *expression* on the right of the `=>` token can't be an expression statement.
+
+Each *switch expression arm* is evaluated in text order. The compiler issues an error when later switch expression arms can't be chosen because an earlier *switch expression arm* matches all its values.
```
pkulikov: Now, it reads as an individual expression arm is evaluated from left to right, but I guess the intention is to say "from top to bottom":
```suggestion
The *switch expression arms* are evaluated in text order. The compiler issues an error when a lower *switch expression arm* can't be chosen because a higher *switch expression arm* matches all its values.
```

docs/csharp/language-reference/operators/switch.md
```diff
@@ -0,0 +1,49 @@
...
+
+:::code language="csharp" source="snippets/SwitchExpressions.cs" id="SnippetRecursivePattern":::
+
+Recursive patterns can examine properties of the range expression, but can't execute arbitrary code. You can use a *case guard*, specified in a `when` clause, to provide similar checks for other sequence types:
+
+:::code language="csharp" source="snippets/SwitchExpressions.cs" id="SnippetGuardCase":::
+
+Finally, you can add the `_` pattern and the `null` pattern to catch arguments that aren't processed by any other switch expression arm. That makes the switch expression *exhaustive*, meaning any possible value of the range expression is handled. The following example adds those expression arms:
+
+:::code language="csharp" source="snippets/SwitchExpressions.cs" id="SnippetExhaustive" interactive="csharp-interactive":::
```
pkulikov: The corresponding snippet doesn't look to be interactive.

docs/csharp/language-reference/operators/switch.md
```diff
@@ -0,0 +1,49 @@
...
+
+The preceding example adds a `null` pattern, and changes the `IEnumerable<T>` type pattern to a `_` pattern. The `null` pattern provides a null check as a switch expression arm. The expression for that arm throws a null argument exception. The `_` pattern matches all inputs that haven't been matched by a previous arm. It must come after the `null` check, or it would match `null` inputs.
```
pkulikov:
```suggestion
The preceding example adds a `null` pattern, and changes the `IEnumerable<T>` type pattern to a `_` pattern. The `null` pattern provides a null check as a switch expression arm. The expression for that arm throws an <xref:System.ArgumentNullException>. The `_` pattern matches all inputs that haven't been matched by previous arms. It must come after the `null` check, or it would match `null` inputs.
```

docs/csharp/language-reference/operators/switch.md
```diff
@@ -0,0 +1,49 @@
...
+---
+# switch expressions (C# reference)
+
+This article covers the `switch` expression, introduced in C# 8.0. For information on the `switch` statement, see the article on the [`switch` statement](../statements/switch.md) in the [statements](../statements/index.md) section.
+
+## Basic example
+
+The `switch` expression provides for `switch`-like semantics in an expression context. It provides a concise syntax when the switch arms produce a value. The following example shows the structure of a switch expression. It translates values from an `enum` representing visual directions in an online map to the corresponding cardinal direction:
+
+:::code language="csharp" source="snippets/SwitchExpressions.cs" id="SnippetBasicStructure" interactive="csharp-interactive":::
```
pkulikov: Is it the new interactive system in action? The snippet doesn't include `enum` definitions, but will be runnable anyway?

docs/csharp/language-reference/operators/snippets/SwitchExpressions.cs
BillWagner: @Thraka I like that this one sample program is used for all the articles in this folder. 

I'll switch the article name from *switch.md* to *switch-expression.md*.  

I will be updating the samples based on the other feedback this morning. (My goal is to make everything interactive). 

Once I do that, let's resolve this. I'd like to keep these snippets as one project for simplicity. Beyond that, I'm open to moving things around. Although, for scope, I might push for a 2nd PR to move samples.

Thraka:
> > Approved apart from the `snippets/article` folder naming.
> 
> @Thraka how would it help? I see only a large duplication between article name and the file name, because each folder would contain only one file. Is it necessary?
> 
> Also, according to #17297:
> 
> > For conceptual, we would put the project(s) in a snippets subfolder. The snippets subfolder should be in the same folder as the article by default. Where a snippets project contains code used in multiple articles, the snippets folder should be in the nearest parent folder of all relevant articles.
> 
> From that, I conclude that the current structure is fine.

It helps to avoid conflicts. For example, if C# 9 is out and there is something in the csproj you need to enable that you can't through code, you now have to create a new project file and manage file excludes in both projects and all sorts of other possible conflicts.

We also don't want to reuse samples across multiple articles, which having code snippets in the single root `snippets` folder would encourage. This is because you can inadvertently break the context or tone of one article while fixing or adjusting another.

@BillWagner Perhaps these could be moved into a single subfolder named `shared` with a csproj to indicate this content is spread among multiple files?

pkulikov:
@Thraka thanks for the clarification.

As I understand, each article folder is also supposed to contain its own *.csproj* file? In general, I agree with such approach, especially for the conceptual articles that are usually large and might require a lot of complex code. At the same time, the language reference articles sometimes are simpler and one shared project with one *.cs* file per article might be fine as well. That's why such format was chosen for the C# operators articles, I believe. I suggest to keep it as it is and move to the project-per-article approach only when necessary (to avoid conflicts if those occur in future).

> We also don't want to reuse the same snippets across multiple articles,

I also absolutely agree with that statement. I have been following that guideline all the time.

Thraka:
@pkulikov Yep. I think moving the content to a child `shared` folder helps keep the flexibility this section requires but future proofs it against having to have a second csproj for something else.

docs/csharp/language-reference/operators/switch.md
BillWagner: I added a link to the recursive patterns spec.

docs/csharp/language-reference/operators/switch.md
BillWagner: I agree with those thoughts. We're working on planning for major updates in time for .NET 5.0, and this is one of the topics being discussed. I'm hoping to have an issue and project for public discussion in the next few weeks.

docs/csharp/toc.yml
BillWagner: Good idea.

docs/csharp/language-reference/operators/switch.md
BillWagner: The new system is deployed. I'm making all the snippets in this article non-interactive because the new system uses scriptcs, which doesn't have full fidelity with C# 8.0 features.

docs/csharp/language-reference/operators/switch.md
BillWagner: That's a good point. Right now, that's the term used in the [recursive patterns spec](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/patterns)  for the grammar.

I think this may change. The spec currently uses the term "switch-expression" for the same concept in the `switch` statement.

For now though, this is consistent.

BillWagner:
>  I think moving the content to a child shared folder helps keep the flexibility this section requires but future proofs it against having to have a second csproj for something else.

I'd like to separate that into a standalone PR.

PR #17520, opened 2020-03-20
XML code blocks

cloudn9:
#9744
Should fix first 4 issues in the list of 267 entries

First 2 commits Verified, are not from #17520


Thraka (approved):
Thanks!!

PR #17523, opened 2020-03-20
Fix XML code blocks

cloudn9:
Should fix next issues from the list #9744


BillWagner (approved):
Thanks @cloudn9 

PR #17541, opened 2020-03-21
Adding another example from KB 301075

PiJoCoder:
The intention is to retire https://support.microsoft.com/en-us/help/301075/how-to-connect-to-a-database-and-run-a-command-by-using-ado-net-and-vi and transfer the code example to appropriate article.

## Summary

Describe your changes here.


BillWagner:
@stevestein Can you give this a review?

One concern I have is that the VB sample is not tagged correctly, so it appears when C# or C++ is selected as the language.

Another is that the sample was added without any accompanying text. The text to explain this sample should also be added.

BillWagner:
After internal discussion, I'm closing this PR.

While the information is useful, the additional sample doesn't belong in this article. Instead, a new article should replace the reference KB article, including both the text and the sample.

PiJoCoder:
If you guys don't see value in this sample, no worries. This KB article will be archived and no replacement needed for it. Thank you.

PR #17544, opened 2020-03-21
Remove exact line count

nxtn:
Continued from #16865

There are only two lines now.


BillWagner (approved):
This is a good resilient fix. Thanks @NextTurn 

I'll :shipit: now.

PR #17550, opened 2020-03-22
Remove outdated information from Regex docs

stephentoub:
Regex uses DynamicMethod, which can be collected.
cc: @jkotas


docs/standard/base-types/compilation-and-reuse-in-regular-expressions.md
```diff
@@ -19,11 +19,9 @@ You can optimize the performance of applications that make extensive use of regu
...
+ If a <xref:System.Text.RegularExpressions.Regex> object is constructed with the <xref:System.Text.RegularExpressions.RegexOptions.Compiled?displayProperty=nameWithType> option, it compiles the regular expression to explicit MSIL code instead of high-level regular expression internal instructions. This allows .NET's just-in-time (JIT) compiler to convert the expression to native machine code for higher performance.  The cost of constructing the <xref:System.Text.RegularExpressions.Regex> object may be higher, but the cost of performing matches with it is likely to be much smaller.
+
+ An alternative is to use precompiled regular expressions. You can compile all of your expressions into a reusable DLL by using the <xref:System.Text.RegularExpressions.Regex.CompileToAssembly%2A> method. This avoids the need to compile at runtime while still benefiting from the speed of compiled regular expressions.
```
jkotas: The next chapter is talking about precompiled regular expressions as well, but with a different meaning.

Change "Your application can take advantage of precompiled regular expressions in ..." to "Your application can reuse  regular expressions in ..."  ?

jkotas (approved):

docs/standard/base-types/compilation-and-reuse-in-regular-expressions.md
```diff
@@ -19,20 +19,18 @@ You can optimize the performance of applications that make extensive use of regu
...
   
- Your application can take advantage of precompiled regular expressions in one of the following two ways:  
+ Your application can reuse regular expressions in one of the following two ways:  
   
-- By using a static method of the <xref:System.Text.RegularExpressions.Regex> object to define the regular expression. If you are using a regular expression pattern that has already been defined in another static method call, the regular expression engine will retrieve it from the cache. If not, the engine will compile the regular expression and add it to the cache.  
+- By using a static method of the <xref:System.Text.RegularExpressions.Regex> object to define the regular expression. If you are using a regular expression pattern that has already been defined in another static method call, the regular expression engine will retrieve it from the cache, if it's still available. If not, or if it's not still available, the engine will compile the regular expression and add it to the cache.
```
Thraka: I think this makes it clearer.

```suggestion
- By using a static method of the <xref:System.Text.RegularExpressions.Regex> object to define the regular expression and add it to the cache. If you're using a regular expression pattern that has already been defined by another static method call, the regular expression engine will try to retrieve it from the cache. If it's not available in the cache the engine will compile the regular expression and add it to the cache.
```

Thraka (approved):
Let me know your thoughts on the suggestion.

docs/standard/base-types/compilation-and-reuse-in-regular-expressions.md
stephentoub: Thanks, but I personally think that makes it less clear.  The first sentence states that using the static method is going to add it to the cache, period. But it only adds it to the cache if it's not already there.

Thraka: OK. I updated the text to remove that intro clause. How is this?

stephentoub: Seems ok.

stephentoub:
Thanks

PR #17556, opened 2020-03-23
Fix incorrect filename for docker-compose override

MaartenGDev:
## Summary

There seems to be some inconsistency for the naming of the `docker-compose` override. This change makes both commands use the same filename. The filename corresponds with the following docs:

https://github.com/dotnet/docs/blob/master/docs/architecture/microservices/multi-container-microservice-net-applications/test-aspnet-core-services-web-apps.md#testing-in-eshoponcontainers


BillWagner (approved):
Thanks for spotting this and fixing it @MaartenGDev 

PR #17578, opened 2020-03-24
Clarify single and double quotation marks

jonzfisher:
## Summary
Clarify single and double quotation marks within verbatim string literal escape sequence `""`

The language `single quotation mark` suggests `""` will create a single quotation `'` when it will actually translate to one double quotation`"`.


BillWagner (approved):
I like the way you've addressed this @jonzfisher 

I'll :shipit: now. You should see the updates on our live site in the next day or two, on our regular publishing cycle.

Thanks again for contributing to docs! 🎆

jonzfisher:
Thanks for taking a look!

PR #17586, opened 2020-03-24
clarify nullable

### LINKED ISSUE #13421: C# Guide: update all occurrences of "Nullable Types", and "Nullable Value Types"
We often used "nullable type" where we meant "nullable value type" (because all reference types were nullable). These should be checked, and we must clarify if it applies to all nullables, or only nullable value types.

BillWagner:
Content that refers to "nullable" often refers to "nullable value types", introduced in .NET 2.0. Now that we also have nullable and non-nullable reference types, this needed to be clarified in many locations.

In addition, found a breadcrumb entry that was redirected.

Fixes #13421


docs/csharp/how-to/safely-cast-using-pattern-matching-is-and-as-operators.md
```diff
@@ -25,7 +25,7 @@ The preceding sample demonstrates other features of pattern matching to use with
 
 The preceding sample also shows how you use the type pattern in a `switch` statement where the variable may be one of many different types.
 
-If you want to test if a variable is a given type, but not assign it to a new variable, you can use the `is` and `as` operators for reference types and nullable types. The following code shows how to use the `is` and `as` statements that were part of the C# language before pattern matching was introduced to test if a variable is of a given type:
+If you want to test if a variable is a given type, but not assign it to a new variable, you can use the `is` and `as` operators for reference types, nullable reference types, and nullable value types. The following code shows how to use the `is` and `as` statements that were part of the C# language before pattern matching was introduced to test if a variable is of a given type:
```
tdykstra: Since "reference types" includes "nullable reference types," is it necessary to add the latter phrase?

docs/csharp/misc/cs0458.md
```diff
@@ -10,12 +10,12 @@ ms.assetid: 0986c620-b4bc-4e4b-976f-88359cfa3a45
 # Compiler Warning (level 2) CS0458
 The result of the expression is always 'null' of type 'type name'  
   
- This warning is caused by a `nullable` expression that always results in `null`.  
+ This warning is caused by a nullable value expression that always results in `null`.
```
tdykstra:
```suggestion
 This warning is caused by a nullable value type expression that always results in `null`.  
```

docs/csharp/nullable-attributes.md
```diff
@@ -24,7 +24,7 @@ Updating your library for nullable references requires more than sprinkling `?`
 
 This work takes time. Let's start with strategies to make your library or application nullable-aware, while balancing other requirements and deliverables. You'll see how to balance ongoing development enabling nullable reference types. You'll learn challenges for generic type definitions. You'll learn to apply attributes to describe pre- and post-conditions on individual APIs.
 
-## Choose a nullable strategy
+## Choose a nullable reference strategy
```
tdykstra:
```suggestion
## Choose a strategy for nullable reference types
```

docs/csharp/pattern-matching.md
```diff
@@ -198,7 +198,7 @@ Finally, you can add a `null` case to ensure the argument isn't `null`:
 
 The special behavior for the `null` pattern is interesting because the constant
 `null` in the pattern doesn't have a type but can be converted to any reference
-type or nullable type. Rather than convert a `null` to any type, the language
+type, nullable reference type, or nullable value type. Rather than convert a `null` to any type, the language
```
tdykstra: Same question wrt whether we need to add "nullable reference type" here.

tdykstra (approved):
Looks great, just a couple of minor suggestions.

docs/csharp/how-to/safely-cast-using-pattern-matching-is-and-as-operators.md
BillWagner: I wasn't sure. I think it's better to remove it.

PR #17594, opened 2020-03-25
Update CODEOWNERS Azure area

Thraka:
## Summary

Add @CamSoper as owner of the azure stuff.

Is this correct?


BillWagner (approved):
This matches my understanding. 

@CamSoper should weigh in as well.

PR #17606, opened 2020-03-26
Structure types: moved in the readonly struct section

pkulikov:
Contributes to #10726


BillWagner (approved):
This is really great, @pkulikov 

I'll :shipit: now.

PR #17613, opened 2020-03-26
Update environment title to be consistent with aka.ms link

natke:


luisquintanilla (approved):
Looks good. Not critical but a change for later. In the tool the step is Scenario not Scenarios. It'd be good to make that change to stay consistent.

PR #17634, opened 2020-03-27
Clean up /standard TOC

gewarren:
Clean up /standard TOC and other small touch-ups.

In preparation for the new landing page.

### DIFF
--- a/docs/standard/get-started.md
+++ b/docs/standard/get-started.md
@@ -12,21 +12,23 @@ There are a number of ways to get started with .NET. Because .NET is a massive p
 
 ## Get started using .NET languages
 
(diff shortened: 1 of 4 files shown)


tdykstra (approved):
LGTM so far as it goes -- this would be an opportunity to bring back into the .NET Standard TOC the sub-sections that were split into small disconnected TOC's (e.g., see Security, Serialization, etc.).

gewarren:
> LGTM so far as it goes -- this would be an opportunity to bring back into the .NET Standard TOC the sub-sections that were split into small disconnected TOC's (e.g., see Security, Serialization, etc.).

Yes, I'll add those in in a subsequent PR. Thanks for reviewing!

PR #17643, opened 2020-03-29
Update iterators.md

Vasilisdm:
Added an explanation about what is actually an iterator.

## Summary

Just added the definition of what the iterator actually is, I did since I searched it my self.

### DIFF
--- a/docs/csharp/iterators.md
+++ b/docs/csharp/iterators.md
@@ -12,8 +12,8 @@ Almost every program you write will have some need to iterate
 over a collection. You'll write code that examines every item in
 a collection.
 
-You'll also create iterator(which is actually an object, that enables a programmer to traverse a container, particularly lists) methods which are methods that produces an
-iterator for the elements of that class. These can be used for:
+You'll also create iterator methods which are methods that produces an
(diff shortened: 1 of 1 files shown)


docs/csharp/iterators.md
```diff
@@ -12,7 +12,7 @@ Almost every program you write will have some need to iterate
 over a collection. You'll write code that examines every item in
 a collection.
 
-You'll also create iterator methods which are methods that produces an
+You'll also create iterator(which is actually an object, that enables a programmer to traverse a container, particularly lists) methods which are methods that produces an
 iterator for the elements of that class. These can be used for:
```
BillWagner: I would change this just a bit:

```suggestion
You'll also create iterator methods which are methods that produces an
iterator (which is an object that traverses a container, particularly lists) for the elements of that class. These can be used for:
```

BillWagner (changes requested):
This is a great addition @Vasilisdm It's important that we define an iterator in the opening of this article.

I made a suggestion to move the text added, and a couple small edits for style. If you accept those edits, I'll merge this.

BillWagner (approved):

PR #17653, opened 2020-03-31
Updated documentation with the path where xsd can be found

furoTmark:

### DIFF
--- a/docs/standard/serialization/xml-schema-def-tool-gen.md
+++ b/docs/standard/serialization/xml-schema-def-tool-gen.md
@@ -9,11 +9,11 @@ helpviewer_keywords:
 ms.assetid: 51f0edc3-993d-4051-b7f2-77753694d3d1
 ---
 # How to: Use the XML Schema Definition Tool to Generate Classes and XML Schema Documents
-The XML Schema Definition tool (Xsd.exe) allows you to generate an XML schema that describes a class or to generate the class defined by an XML schema. The following procedures show how to perform these operations. 
+The XML Schema Definition tool (Xsd.exe) allows you to generate an XML schema that describes a class or to generate the class defined by an XML schema. The following procedures show how to perform these operations.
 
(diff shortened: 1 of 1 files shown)


docs/standard/serialization/xml-schema-def-tool-gen.md
```diff
@@ -9,8 +9,11 @@ helpviewer_keywords:
 ms.assetid: 51f0edc3-993d-4051-b7f2-77753694d3d1
 ---
 # How to: Use the XML Schema Definition Tool to Generate Classes and XML Schema Documents
-The XML Schema Definition tool (Xsd.exe) allows you to generate an XML schema that describes a class or to generate the class defined by an XML schema. The following procedures show how to perform these operations.  
-  
+The XML Schema Definition tool (Xsd.exe) allows you to generate an XML schema that describes a class or to generate the class defined by an XML schema. The following procedures show how to perform these operations.
+
+The XML Schema Definition tool (Xsd.exe) usually can be found in the following path:  
+C:\Program Files (x86)\Microsoft SDKs\Windows\ ***\<version\>*** \bin\NETFX ***\<version\>*** Tools
```
Thraka, line 15: We don't call out `<version>` syntax with styling separate from the rest of the path anymore. We assume most people understand placeholders. Paths/files as a whole should be styled with *italics*. We've started to use `{ }` syntax for placeholders as it removes confusion around XML.

Please accept this suggestion:

```suggestion
The XML Schema Definition tool (Xsd.exe) usually can be found in the following path:\
*C:\Program Files (x86)\Microsoft SDKs\Windows\{version}\bin\NETFX {version} Tools\*
```

Additionally, you can use the `\` syntax at the end of a line to enforce a line break instead of two spaces. Easier to spot 😄

Thraka (changes requested):
Thank you for updating the article and pointing this out!

Thraka (approved):

PR #17659, opened 2020-03-31
Clarified the type of a null-conditional expression for value-type expressions

### LINKED ISSUE #17618: Null conditional operator ?. needs explicit ValueType impact clarification
### Null conditional operator ?. needs explicit ValueType impact clarification

The current docs have an aside following the example in the **Null-conditional operators ?. and ?[]** section following the example:

> The preceding example also uses the null-coalescing operator ?? to specify an alternative expression to evaluate in case the result of a null-conditional operation is null.

I suggest more explicit guidance for ValueType T comparisons, lifted up into the main part of the explanation of the section:

> When a ValueType field or property (int, bool, struct, etc.) is preceded by the ?. operator, you should consider either:
> 
> - Using '??' to supply a default value T to compare with, if value-oriented comparison is intended.
> 
> - Comparing to null instead of some value T, if merely checking if the preceding operation chain is fully-defined (and value-oriented comparison does not also matter).

Examples:
```c#
int[ ] a = null;
// all of these are perfectly fine with the compiler
if (a?.Length == 0) return; // WRONG! looks right, but only works when a is defined, not when a is null
if (a?.Length == null) return; // works, but only if you care whether a is defined and not its value
if (a?.Length == null || a?.Length == 0) return; // works as both definition and value check, but worse than the original code 'a == null || a.length == 0' 
if ((a?.Length??0) == 0) return;  // works as value check and maybe an implicit definition check, hence this whole doc comment
```

I realize you can make the transitive argument that the '?.' doc leads to Nullable<T>, and Nullable<T> behavior is separately documented; but this is a definite case where no compiler warning will happen and it will merrily cause you a lot of grief if this gets by the eyeball test while coding and debugging (assuming your eyeballs even realize it is a problem). It also flies in the face of C#, in that usually the most elegant, simple representation does what you think it does. Here, it does not. Just look above at the example that doesn't work, which is the most likely result of a dev using '?.' instead of the old 'a == null || a.Length == 0' clause. Many devs would simply remove the 'a == null ||' and add the '?'.

Try this:
```c#
if (a == null || a.Length == 0) return;
```
Edit it the simplest way that looks like it should work:
```c#
if (a?.Length == 0) return;
```
A dev new to this feature will do this and it is incorrect.

This related issue [C# Language operators: ?. Null Conditional operator #1368](https://github.com/dotnet/docs/issues/1368) reinforces the need for more explicit guidance, just look at how this has to be worded in talking about it:

> The result is null if the LHS is null, right if LHS is not null.

pkulikov:
Fixes #17618

/cc: @redmondgiant

### DIFF
--- a/docs/csharp/language-reference/operators/member-access-operators.md
+++ b/docs/csharp/language-reference/operators/member-access-operators.md
@@ -215,4 +215,3 @@ For more information about indices and ranges, see the [feature proposal note](~
 - [C# operators](index.md)
 - [?? (null-coalescing operator)](null-coalescing-operator.md)
 - [:: operator](namespace-alias-qualifier.md)
- 
--- a/docs/csharp/language-reference/operators/member-access-operators.md
+++ b/docs/csharp/language-reference/operators/member-access-operators.md
@@ -124,7 +124,7 @@ The following example demonstrates the usage of the `?.` and `?[]` operators:
 
 The preceding example also uses the [null-coalescing operator `??`](null-coalescing-operator.md) to specify an alternative expression to evaluate in case the result of a null-conditional operation is `null`.
 
-If `a.x` or `a[x]` is of a non-nullable value type `T`, `a?.x` or `a?[x]` is of the corresponding nullable value type `T?`. If you need an expression of type `T`, apply the null-coalescing operator `??` to a null-conditional expression, as the following example shows:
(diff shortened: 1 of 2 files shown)


BillWagner:
This LGTM. I'll approve this and :shipit: later today.

@redmondgiant Let us know if you have any feedback.

redmondgiant:
Looks good thanks.

BillWagner (approved):
Thanks again @pkulikov 

This is great work. I'll :shipit: now.

PR #17661, opened 2020-04-01
Events have private invocation

breyed:


BillWagner (approved):
Thanks for adding these clarifying remarks @breyed 
I like the way you've described this.

I'll :shipit: now.

PR #17664, opened 2020-04-01
Profiling APIs are supported in production

noahfalk:
The .Net runtime team has supported using the ICorProfiler APIs in production environments for a long time (a decade or more?). They are tested with that goal in mind and issues reported by customers are investigated and fixed. The text here claimed otherwise so I want to remove that contradictory messaging.

@davmason 

cc @jkotas @GrabYourPitchforks @stephentoub


jkotas (approved):

davmason (approved):
Thanks!

BillWagner (approved):
I'll :shipit: now.

Thanks @noahfalk

PR #17669, opened 2020-04-01
Add note about runtime-dependent was known as framework dependent

Thraka:


BillWagner (approved):
This LGTM @Thraka 

You can :shipit: when ready.

PR #17674, opened 2020-04-01
Updated the example of the lock statement

### LINKED ISSUE #17642: Lock could be held for a shorter period of time
In general, it's a good idea to hold a lock for as short a time as possible, so that one thread doesn't have to wait for another thread to perform potentially slow operations that have nothing to do with the locked resource.

The console I/O in the example code, for instance, is an example of a potentially slow operation (much slower than simple arithmetic calculations).  There's no need for the console write calls that display the old and new balance to be performed with the lock held.  Instead, the two balance values could be copied to temporary local variables (while holding the lock) and then printed only after releasing the lock, something like this (for the debit case):

public decimal Debit(decimal amount)
{
    decimal return_value;

    // These must be declared outside the lock so that they can still be
    // accessed after the lock is released.
    decimal old_balance, new_balance;

    // Hold the lock for as short a period of time as possible, only long
    // enough to perform the arithmetic.
    lock (balanceLock)
    {
        old_balance = balance;

        if (old_balance &gt;= amount)
        {
            new_balance = old_balance - amount;
            balance = new_balance;
            return_value = amount;
        }
        else
        {
            new_balance = old_balance;
            // (no change to balance)
            return_value = 0;
        }
    }

    Console.WriteLine($"Balance before debit :{old_balance, 5}");
    Console.WriteLine($"Amount to remove     :{amount, 5}");
    Console.WriteLine($"Balance after debit  :{new_balance, 5}");

    return return_value;
}

---
#### Document Details

* ID: b544eeb3-082e-1c67-eb76-f5ce0acee45e
* Version Independent ID: d4765a2e-ac77-e947-165c-5a0f88f4e8e2
* Content: [lock statement - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/lock-statement#feedback)
* Content Source: [docs/csharp/language-reference/keywords/lock-statement.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/keywords/lock-statement.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

pkulikov:
Fixes #17642

### DIFF
--- a/docs/csharp/language-reference/keywords/lock-statement.md
+++ b/docs/csharp/language-reference/keywords/lock-statement.md
@@ -1,7 +1,7 @@
 ---
 title: "lock statement - C# reference"
 description: "Use the C# lock statement to synchronize thread access to a shared resource"
-ms.date: 10/01/2018
+ms.date: 04/02/2020
 f1_keywords: 
   - "lock_CSharpKeyword"
   - "lock"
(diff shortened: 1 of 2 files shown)


BillWagner (approved):

PR #17685, opened 2020-04-02
Fix typo

levi-rocha:
## Summary

I think this is what it meant to say considering the context. There's no reason it should specify a "Mac" rather than a machine because the environment mentioned is cross-platform.


BillWagner (approved):
This LGTM. I'll :shipit:

PR #17689, opened 2020-04-03
Add note about usesocketshttphandler

### LINKED ISSUE #16088: System.Net.Http.UseSocketsHttpHandler setting not available for .NET 5
Update https://docs.microsoft.com/en-us/dotnet/core/run-time-config/networking to note that the `System.Net.Http.UseSocketsHttpHandler` setting is removed for .NET 5.

gewarren:
Fixes #16088


jkotas (approved):
Thanks

PR #17697, opened 2020-04-03
Fix build warning

tdykstra:


gewarren (approved):
I think it's good to get rid of the xref here for now. That warning has become incredibly annoying.

PR #17717, opened 2020-04-06
Update portability-analyzer.md

shaggygi:
Fix wording.


docs/standard/analyzers/portability-analyzer.md
```diff
@@ -69,6 +69,6 @@ The **Details** section of the report lists the APIs missing from any of the sel
...
+You may find a Missing Assemblies section in your report. It tells you that this list of assemblies are referenced by your analyzed assemblies and were not analyzed. If it's an assembly that you own, include it in the Api portability analyzer run so that you can get API level detailed portability report for it. If it's a third party library, check if there is a newer version supporting your target platform. If so, consider moving to the newer version. Eventually, you would expect this list to include all the third party assemblies that your app depends on and confirmed that they have a version supporting your target platform.
```
gewarren, line 72:
```suggestion
You may find a Missing Assemblies section in your report. This section contains a list of assemblies that are referenced by your analyzed assemblies and were not analyzed. If it's an assembly that you own, include it in the API portability analyzer run so that you can get a detailed, API-level portability report for it. If it's a third-party library, check if there is a newer version that supports your target platform, and consider moving to the newer version. Eventually, the list should include all the third-party assemblies that your app depends on that have a version supporting your target platform.
```

I like your suggestions. I made a few more suggestions to clarify the intent.

BillWagner:
@gewarren I think this has been updated per your comments.

gewarren (approved):
Thanks for the ping Bill!

PR #17733, opened 2020-04-06
Clarify note about older HTTP stacks

gewarren:


docs/core/run-time-config/networking.md
```diff
@@ -19,14 +21,19 @@ ms.topic: reference
 
 ## UseSocketsHttpHandler
 
-- Configures whether high-level networking APIs, such as <xref:System.Net.Http.HttpClient>, use <xref:System.Net.Http.SocketsHttpHandler?displayProperty=nameWithType> or the implementation of <xref:System.Net.Http.HttpClientHandler?displayProperty=nameWithType> that's based on [libcurl](https://curl.haxx.se/libcurl/).
-- Default: Use <xref:System.Net.Http.SocketsHttpHandler?displayProperty=nameWithType> (`true`).
+- Configures whether <xref:System.Net.Http.HttpClientHandler?displayProperty=nameWithType> uses <xref:System.Net.Http.SocketsHttpHandler?displayProperty=nameWithType> or older HTTP protocol stacks (<xref:System.Net.Http.WinHttpHandler> on Windows and `CurlHandler`, an internal class implemented on top of [libcurl](https://curl.haxx.se/libcurl/), on Linux).
```
tdykstra, line 24: This note refers to CurlHandler (built on libcurl) but the sections below don't mention CurlHandler and only refer to libcurl directly.  Is that intentional?

```suggestion
- Configures whether <xref:System.Net.Http.HttpClientHandler?displayProperty=nameWithType> uses <xref:System.Net.Http.SocketsHttpHandler?displayProperty=nameWithType> or older HTTP protocol stacks. Examples of older HTTP protocol stacks, such as <xref:System.Net.Http.WinHttpHandler> on Windows and `CurlHandler`  on Linux. `CurlHandler` is an internal class implemented on top of [libcurl](https://curl.haxx.se/libcurl/).
```

tdykstra (approved):
LGTM

docs/core/run-time-config/networking.md
gewarren: Yes, that was intentional just to be less wordy in the table.

gewarren: Thank you for the suggestion, but I think it changes the intent a bit so I'm going to keep it as-is. They are two specific HTTP stacks, not just examples.

PR #17741, opened 2020-04-07
Fix typos

levi-rocha:
## Summary

Fixes a couple of typos in the document


Thraka (approved):
Thanks for spotting and fixing those!

PR #17456, opened 2020-03-13
C# operator reference: introduce expressions

pkulikov:
Per discussion in #14889
@BillWagner while working on it, I've not managed to convert `await` into the expression. Somehow, with the "operator" word the article still flows smoother.

The changes in the `stackalloc` article are done as a part of #17450

PR #17763, opened 2020-04-08
Renamed cast operator to cast expression

pkulikov:
An addition to #17456, also inline with how it's named in [language spec](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#cast-expressions).


BillWagner (approved):
This LGTM @pkulikov 

thanks again for all your contributions!

PR #17774, opened 2020-04-09
Updated attribute to correct name

philipwolfe:
I think the correct name for the attribute is *serializeAs* and not serializedAs.  It is correct in the example and I checked a couple of other examples on the web.

## Summary

Describe your changes here.


gewarren (approved):
@philipwolfe Yes, I agree with you. Thanks for fixing this!

PR #17778, opened 2020-04-10
Fix constructor link display text

gewarren:
?displayProperty=nameWithType displays as "Classname.Classname()". This does not seem desirable to me as you don't/can't call a constructor that way. Better to just display "Classname()".


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

gewarren:
Thanks for the review, Bill.

PR #17791, opened 2020-04-10
Fix constructor xref link display

gewarren:
More like in #17778.

Displays as **StringBuilder(Int32)** instead of **StringBuilder.StringBuilder(Int32)**, for example.


tdykstra (approved):
LGTM

scottaddie (approved):

PR #17800, opened 2020-04-11
Readonly struct members can access non-readonly members

pkulikov:
The following code compiles and runs:

```csharp
using System;

struct Point 
{
    public double x;
    public override string ToString()
    {
        x = 4;
        return x.ToString();
    }
}
struct Container
{
    readonly Point p;

    public readonly void M()
    {
        string s = p.ToString();
        Console.WriteLine(s);  //  output: 4
        string s1 = this.ToString();
        Console.WriteLine(s1);  // output: 0
    }
    public override string ToString() => p.x.ToString();
}

class Program
{
    static void Main()
    {
        var c = new Container();
        c.M();
    }
}
```

Both calls, `string s = p.ToString();` and `string s1 = this.ToString();` create defensive copies. [Remark: the `this.ToString()` results in the warning about an implicit copy of `this`, while `p.ToString()` — not, no warnings. As far as I understand, that is because the introduction of such a warning would be a breaking change.]


BillWagner (approved):
Good catch @pkulikov 
I had written that description poorly. Thanks for fixing it.

I'll :shipit: now.

PR #17805, opened 2020-04-12
Changed the meaning

Vasilisdm:
Made more clear that you are not able to pass an array as since it is a reference type and a value one.

## Summary

Describe your changes here.


BillWagner:
@Vasilisdm I'm going to close this PR, as the original text was more instructive: The point is that passing a reference type (a class) by value means the called code can't reassign the reference. However, it can modify the state of the object referred to.

Vasilisdm:
Let's agree to disagree then. I am just saying that it might be confusing for a newcomer, but its your call.

PR #17836, opened 2020-04-14
Align version in sample with narrative

sthussey:
The narrative is about building a Netcore 3.1 app, but the sample showed 2.2.

## Summary

Update the Getting Started documentation sample of the project file to show netcore 3.1 rather than 2.2


tdykstra (approved):
Thanks!

PR #17852, opened 2020-04-15
code improvement suggestions

sharpist:
In line 15 changed the Boolean condition.
Prime cause: this boolean expression have not targeted the case with a parameter equal to 1.


BillWagner (approved):
Thanks for spotting and making this change @sharpist 

I'll :shipit: now.

PR #3776, opened 2017-11-23
Add tuple deconstructing assignment documentation with code snippet

snhardin:
# Add tuple deconstructing assignment documentation with code snippet

## Summary

I have added more detail for deconstructing a tuple into variables that have already been declared.

Fixes #3690

## Details

So, up until #3690, I had no idea that you could deconstruct tuples into pre-declared variables. I agree that this feature should be mentioned in the "Deconstructing tuples and other types" page and, well, here's this PR.

The only page I wound up changing was _deconstruct.md_. I mentioned the possibility of deconstructing into declared variables, added a code example, and added a note about mixing declared variables with undeclared variables on the left side of a tuple deconstruction statement.

I did not modify _tuples.md_ because it actually already mentions this possibility (although only briefly) and includes sample code that is similar to that of @BillWagner's in #3690. **Advice for additional changes here is welcome.**

Advice is also welcome as to where it would be a good place to mention the swap one-liner that was also recommended in the open issue.

## Suggested Reviewers

@BillWagner because he commented on the original issue.

PR #17854, opened 2020-04-15
Fix ms.date

pkulikov:
Year 2016 doesn't look correct for the article about C# 7.0 feature (which was released in 2017). I've updated the date based on the last large update of the article: #3776


BillWagner (approved):
Thanks for spotting this and making the updates @pkulikov 

I'll :shipit: now.

PR #17856, opened 2020-04-15
Rename function parameter name that could be misleading

lilymaybenichou:
exeConfigName lead me to pass the config file path. Which did not produce expected results. 
Passing the executable file path worked as expected

## Summary

Describe your changes here.
exeConfigName parameter name lead me to pass the config file path. Which did not produce expected results. 
Passing the executable file path worked as expected


BillWagner (approved):
Thanks for spotting and making this change @lilymaybenichou I agree that the updated name is more clear.

I'll :shipit: now.  You should see the changes on our live site in the next day or two, on our regular publishing cycle.  

PR #17864, opened 2020-04-15
Change SSL to TLS per 1679050

TimShererWithAquent:
## Summary

This is [1679050](https://dev.azure.com/mseng/TechnicalContent/_workitems/edit/1679050) Content Update: replace SSL in docs

@msmbaldwin - Author is @HongGit 

Author - please review these as soon as you can. According to our SLA, review time is 5 business days. I'll follow up in email in a couple of days, in case you are buried under github at-mentions. 

Hi Hong - per our emails yesterday, I've done one directory as starter. We usually include fewer than 20 articles in a PR. The changes here are not major. Hopefully, this won't be as big a change as it seemed.


TimShererWithAquent:
As noted in [the work item](https://dev.azure.com/mseng/TechnicalContent/_workitems/edit/1679050), we're narrowing the scope to include only azure-docs-pr.

PR #17876, opened 2020-04-16
Update using-directive.md

Rick-Anderson:


BillWagner:
@Rick-Anderson Here, the original text is needed: It's meant to drive readers to a different article for a different use of `using`.

I'll close this.

Rick-Anderson:
I was thinking also is not exactly correct because one is a directive and the other a statement. But it's close enough. Probably too verbose to say 

The `using` keyword  is different than the `using` directive. The `using` keyword  used to create *using statements*, which help en...

That was my point.  aslo mean in addition, but it's not really additional, it's different.

PR #17884, opened 2020-04-16
"Explore C# 6": Fixed code on step 6

Domanser:
## Summary

Two code snippets have different logic:

1. This code doesn't depend on result of MoveNext() (true or false).
    bool? hasMore = s?.ToCharArray()?.GetEnumerator()?.MoveNext();
    Console.WriteLine(hasMore.HasValue);
2. And this code depends on result of MoveNext().
    bool hasMore = s?.ToCharArray()?.GetEnumerator()?.MoveNext() ?? false;
    Console.WriteLine(hasMore);

If string s is "" (empty), first code returns true, and second one returns false.
So, changes have been made in order to make those code snippets equivalent.


BillWagner (approved):
Thanks for making this change @Domanser It does improve the code.

I'll :shipit: now.

PR #17888, opened 2020-04-16
Clarify thread-safe event raising

### LINKED ISSUE #17867: Could the team add more information about the topic "thread-safe-delegate-invocation"
Could the team please add some more detailed explanation about the topic thread-safe-delegate-invocation
I mean thats a very interesting and essential topic imo. And there is only one line written:
"Use the ?. operator to check if a delegate is non-null and invoke it in a thread-safe way (for example, when you raise an event), as the following code shows"

Could the team explain why the ?. operator makes the invocation thread safe?
Sincerely and thanks in advance.

---
#### Document Details

* ID: 0a25abe4-2b4e-119e-85e9-af19ed06e8c0
* Version Independent ID: 682b8c86-e49a-1d6d-8562-c9f919082bdc
* Content: [Member access operators and expressions - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#thread-safe-delegate-invocation)
* Content Source: [docs/csharp/language-reference/operators/member-access-operators.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/operators/member-access-operators.md)
* Product: **dotnet-csharp**
* GitHub Login: @pkulikov
* Microsoft Alias: **wiwagn**

pkulikov:
Fixes #17867

### DIFF
--- a/docs/csharp/language-reference/operators/member-access-operators.md
+++ b/docs/csharp/language-reference/operators/member-access-operators.md
@@ -1,7 +1,7 @@
 ---
 title: "Member access operators and expressions - C# reference"
 description: "Learn about C# operators that you can use to access type members."
-ms.date: 03/31/2020
+ms.date: 04/17/2020
 author: pkulikov
 f1_keywords: 
   - "._CSharpKeyword"
@@ -149,6 +149,8 @@ if (handler != null)
     handler(…);
 }
 ```
+
+That is a thread-safe way to ensure that only a non-null `handler` is invoked. Because delegate instances are immutable, no thread can change the value referenced by the `handler` local variable. In particular, if the code executed by another thread unsubscribes from the `PropertyChanged` event and `PropertyChanged` becomes `null` before `handler` is invoked, the value referenced by `handler` remains unaffected.
 
 ## Invocation expression ()
 
--- a/docs/csharp/language-reference/operators/member-access-operators.md
+++ b/docs/csharp/language-reference/operators/member-access-operators.md
@@ -150,7 +150,7 @@ if (handler != null)
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/csharp/language-reference/operators/member-access-operators.md
```diff
@@ -150,6 +150,8 @@ if (handler != null)
 }
 ```
 
+That is a thread-safe way to ensure that only a non-null `handler` is invoked. Because delegate instances are immutable, no thread can change the value referenced by the `handler` local variable. In particular, if the code executed by another thread unsubscribes from the `PropertyChanged` event and `PropertyChanged` becomes `null` before `handler` is invoked, the value referenced by `handler` remains unaffected.
```
BillWagner, line 153: I'd add the following to re-introduce the `?.` operator:

```suggestion
That is a thread-safe way to ensure that only a non-null `handler` is invoked. Because delegate instances are immutable, no thread can change the value referenced by the `handler` local variable. In particular, if the code executed by another thread unsubscribes from the `PropertyChanged` event and `PropertyChanged` becomes `null` before `handler` is invoked, the value referenced by `handler` remains unaffected. The `?.` operator evaluates its left operand no more than once, guaranteeing that it cannot be changed to `null` after being verified as non-null.
```

BillWagner (changes requested):
Thanks @pkulikov 

I added one extra sentence to explain how `?.` ensures the correct behavior.

docs/csharp/language-reference/operators/member-access-operators.md
pkulikov: Thanks @BillWagner. I've applied the suggestion with the minor update ("left" -> "left-hand")

BillWagner (approved):

PR #17897, opened 2020-04-17
Replace redirected links

gewarren:


Thraka (approved):
~Only thing I think we have to do is restore those linux installer links :) everything else is optional.~

Oops.. I left this on the wrong PR 😁

Thraka (approved):
LGTM! Thank you 😄

PR #17898, opened 2020-04-17
Replace redirects

gewarren:


luisquintanilla (approved):
Looks good to me. Thanks for making these updates.

PR #17901, opened 2020-04-17
Fix signature of Async.StartChild

kentcb:
## Summary

Just fixing the signature of `Async.StartChild` to use `*` instead of an errant `-` in the tuple parameter.


cartermp (approved):
Thanks!

PR #17904, opened 2020-04-17
More simple fixes to tuple signatures

kentcb:
## Summary

Sorry, I already fixed one case in #17901, but then went on to find these other cases.


cartermp (approved):
Thanks, I think these got accidentally adjusted in a docs formatting pass converting `*` to `-` in lists. cc @gewarren

PR #17939, opened 2020-04-19
Update snippet1901.fs

kentcb:
## Summary

Make it clear that semicolons are optional for labels on separate lines in a record.


cartermp (approved):
Thanks, good find

PR #17944, opened 2020-04-19
Structure types: moved in the ref struct section

pkulikov:
Contributes to #10726 (@BillWagner could you please check the discussion in that issue)


docs/csharp/language-reference/builtin-types/snippets/StructType.cs
```diff
@@ -95,4 +95,22 @@ public int Counter
...
+        {
+            public Span<int> Inputs;
+            public Span<int> Outputs;
+        }
+        // </SnippetRefStruct>
+
+        // <SnippetReadonlyRef>
+        public readonly ref struct Example
+        {
+            // Type member definitions
```
BillWagner, line 112: Consider adding an integer, or a boolean here to show examples of those types in a ref struct.

docs/csharp/language-reference/builtin-types/struct.md
```diff
@@ -96,9 +96,32 @@ In the case of the [built-in value types](value-types.md#built-in-value-types),
...
+- A `ref` struct can't be the element type of an array.
+- A `ref` struct can't be a declared type of a field of a class or a non-`ref` struct.
+- A `ref` struct can't implement interfaces.
+- A `ref` struct can't be boxed to <xref:System.ValueType?displayProperty=nameWithType> or <xref:System.Object?displayProperty=nameWithType>.
+- A `ref` struct can't be a type argument.
+- A `ref` struct variable can't be captured by a [lambda expression](../../programming-guide/statements-expressions-operators/lambda-expressions.md) or a [local function](../../programming-guide/classes-and-structs/local-functions.md).
+- A `ref` struct variable can't be used in an [`async`](../keywords/async.md) method. However, you can use `ref` struct variables in synchronous methods, for example, in those that return <xref:System.Threading.Tasks.Task> or <xref:System.Threading.Tasks.Task%601>.
+- A `ref` struct variable can't be used in [iterators](../../iterators.md).
+
+In particular, you use a `ref` struct type if you want to encapsulate other `ref` struct types:
```
BillWagner, line 112: This is a bit misleading. It's true, but `ref struct` types can also contain unmanaged types (int, short, etc).

BillWagner (changes requested):
This is great work @pkulikov 

I had a couple comments, and then it's ready to :shipit:

docs/csharp/language-reference/builtin-types/snippets/StructType.cs
pkulikov: @BillWagner should I also add a span?

In general, is it worth to create custom structs **without** `ref` struct fields? I've thought that the main reason to define a custom `ref` struct is to encapsulate already existing ones.

docs/csharp/language-reference/builtin-types/struct.md
pkulikov: I'll swap the sentence. Its purpose is to introduce one possible use case of `ref` structs.

By the way, `ref` structs can also contain fields of reference types

docs/csharp/language-reference/builtin-types/snippets/StructType.cs
BillWagner: > In general, is it worth to create custom structs without ref struct fields? I've thought that the main reason to define a custom ref struct is to encapsulate already existing ones.

I think most scenarios outside of corefx will be combining existing `ref struct` types. That said, many `ref struct` and `readonly ref struct` types in the corefx library contain a mix of other `ref struct` types and primitive unmanaged types as members. It's worth mentioning that it is legal.

pkulikov: @BillWagner I've updated both sample examples to include some simple types. Do you think that would be enough?

docs/csharp/language-reference/builtin-types/struct.md
pkulikov: @BillWagner I've rephrased that sentence completely. Please check the latest commit.

BillWagner (approved):

BillWagner:
I like the way this turned out. It's clear, correct, and doesn't go so deep as to try and replicate what's in the spec for the feature. 

I'll :shipit: now.

PR #17977, opened 2020-04-21
Fix typos in "C# reserved attributes".

teo-tsirpanis:


BillWagner (approved):
Thanks for spotting this and fixing it @teo-tsirpanis 

PR #17984, opened 2020-04-21
Update character-encoding-introduction.md

ForNeVeR:
Fix unicode character name: `U+00E1` is `LATIN SMALL LETTER A WITH ACUTE`, not `LETTER E` :)


Thraka:
Indeed. Thanks for spotting and especially for fixing it!!

Thraka (approved):

Issue #17918, opened 2020-04-17
Clean up orphaned articles in repo


gewarren:
Related to #17649.

PR #17986, opened 2020-04-21
Orphans - C# and serialization

gewarren:
Contributes to #17918.


docs/csharp/programming-guide/classes-and-structs/local-functions.md
```diff
@@ -69,6 +69,59 @@ As we did with the method iterator, we can refactor the code from this example t
...
+declare a local function, the argument types and return type are part of the function declaration. Instead of being part of the body of the lambda expression, the argument types and return type are part of the lambda expression's variable type declaration. Those two differences may result in clearer code.
+
+Local functions have different rules for definite assignment than lambda expressions. A local function declaration can be referenced from any code location where it is in scope. A lambda expression must be assigned to a delegate variable before it can be accessed (or called through the delegate referencing the lambda expression.) Notice that the version using the lambda expression must declare and initialize the lambda expression, `nthFactorial` before defining it. Not doing so results in a compile time error for referencing `nthFactorial` before assigning it. These differences mean that recursive algorithms are easier to create using local functions. You can declare and define a local function that calls itself. Lambda expressions must be declared, and assigned a default value before they can be re-assigned to a body that references the same lambda expression.
```
tdykstra, line 87: moving period to after closing paren, removing comma
```suggestion
Local functions have different rules for definite assignment than lambda expressions. A local function declaration can be referenced from any code location where it is in scope. A lambda expression must be assigned to a delegate variable before it can be accessed (or called through the delegate referencing the lambda expression). Notice that the version using the lambda expression must declare and initialize the lambda expression `nthFactorial` before defining it. Not doing so results in a compile time error for referencing `nthFactorial` before assigning it. These differences mean that recursive algorithms are easier to create using local functions. You can declare and define a local function that calls itself. Lambda expressions must be declared, and assigned a default value before they can be re-assigned to a body that references the same lambda expression.
```

tdykstra (approved):
LGTM, one minor punctuation suggestion

Issue #17988, opened 2020-04-21
null-forgiving operator *can have* effect at run time


springy76:
Quote from the current doc:
> The null-forgiving operator has no effect at run time. It only affects the compiler's static flow analysis by changing the null state of the expression. At run time, expression x! evaluates to the result of the underlying expression x.

The document currently focuses primarily on the usage of suppressing warnings of code flow analysis.

But when combining the null-forgiving operator with a leading null-conditional member-access operator the compiler will indeed emit different code and this will have effect at run time -- which comes unexpected.

Example:
```C#
void Main()
{
	var names = T1(); // `names == null` now
	var names2 = T2(); // throws
}

IEnumerable<string> T1()
{
	FileInfo? fi = null;
	return fi?.Directory?.EnumerateFileSystemInfos().Select(fsi => fsi.Name);
}

IEnumerable<string> T2()
{
	FileInfo? fi = null;
	return fi?.Directory?.EnumerateFileSystemInfos()!.Select(fsi => fsi.Name);
}
```

This happens because in `T2()` the code actually compiles to `return (fi?.Directory?.EnumerateFileSystemInfos()).Select(fsi => fsi.Name);` (notice the additional parentheses).

I'm sure this is by design and is covered by the sentence "expression x! evaluates to the result of the underlying expression x". But since any article or documentation I have read about the null-forgiving operator so far, focused only about getting rid of warnings, this (change in) behavior was very unexpected, at least to me.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0471f98c-f5cd-00d2-a705-396682f73bcb
* Version Independent ID: a24b862d-c462-a040-8470-9771c74c4e6f
* Content: [! (null-forgiving) operator - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving)
* Content Source: [docs/csharp/language-reference/operators/null-forgiving.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/operators/null-forgiving.md)
* Product: **dotnet-csharp**
* Technology: **csharp-language-reference**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

pkulikov:
@springy76 thanks!

I've simplified your example a bit:

```csharp
using System;
using System.Collections.Generic;
using System.Linq;
#nullable enable

class Program
{
    private static void Main()
    {
        try
[...]

PR #18001, opened 2020-04-22
Correct operator precedence

pkulikov:
Based on discussions in #17988, I've made two changes in the precedence table:
- moved `x?.y` and `x?[y]` (null-conditionals) from the primary precedence row to the unary precedence row (putting it to the primary expressions group was not correct)
- added `x!` (null-forgiving) to the table to the primary precedence row

### DIFF
--- a/docs/csharp/language-reference/operators/index.md
+++ b/docs/csharp/language-reference/operators/index.md
@@ -1,6 +1,6 @@
 ---
 title: "C# operators - C# reference"
-ms.date: 08/20/2019
+ms.date: 04/22/2020
 f1_keywords: 
   - "cs.operators"
 helpviewer_keywords: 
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks for making this set of fixes @pkulikov 

I'll :shipit: now.

AntonLapounov:
@pkulikov @BillWagner I don't think that was a correct fix. Actually, `x?.y` cannot have a lower precedence than `x!`, because `x?.y!.z` is evaluated as `((x?.y)!).z` rather than `x?.((y!).z)`. Furthermore, `x?.y` has a _higher_ precedence than unary operators, since, for instance, `+"foo"?.Length` is evaluated as `+("foo"?.Length)` rather than `(+"foo")?.Length`. However, `x?.y` has a _lower_ precedence than `x.y`, because `x.y?.z.t` is evaluated as `(x.y)?.(z.t)` rather than `((x.y)?.z).t`.

If you really wanted to classify primary expressions by precedence, you would have to introduce more rows to the table: `x.y` > `x?.y` > `x!` > unary operators. In any case, it may be worth adding `x?.y` and `x?[y]` null-conditional operators to the list of right-associative operators: `x?.y?.z` is evaluated as `x?.(y?.z)` rather than `(x?.y)?.z`.

pkulikov:
@AntonLapounov @BillWagner 
it's hard to reason about the precedence of the null-conditional operators because those are evaluated in their own manner according to [the spec](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#null-conditional-operator)

`x?.y!.z` is evaluated as `((x?.y)!).z` because of the definitions and procedure in the linked spec. Furthermore, `x?.((y!).z)` doesn't look to be a correct expression. `y` is an identifier, so you cannot apply the `!` operator to it. Also, after a (null-conditional) member access `?.` one must use identifier as well, not an expression, so `x?.(...)` is not correct.

> `x?.y?.z` is evaluated as `x?.(y?.z)` rather than `(x?.y)?.z`

That cannot be as `x?.(...)` is not an expression. Also, the null-conditional operators are *unary* operators (they accept only one argument - a primary expression on the left). How to define associativness for unary operators?

Thanks for the remark about the other unary operators, Indeed, the null-conditional operators behave as if they have a higher precedence than unary `-`, `+`, `!` (bool negation)...

AntonLapounov:
@pkulikov The null-conditional operators are binary as they take `primary_expression` as their left side and `null_conditional_operations` as their right side. For instance, `x?.y.z+t` is interpreted as `(?(x, .y.z))+t`. Here the `?(x, .y.z)` operator either applies the second argument `.y.z` to `x` (if `x` is not `null`) or returns `null` of the appropriate type (if `x` is `null`).

> `x?.y!.z` is evaluated as `((x?.y)!).z` because of the definitions and procedure in the linked spec. 

Right. Those definitions and procedure _imply_ precedence of operators. For instance, `x+y*z` is parsed/interpreted as `x+(y*z)` rather than `(x+y)*z`, so we say that the `*` operator has a higher precedence than the `+` operator (in other words, `*` has 'stronger' binding). The same way the `?.` operator has a higher precedence than the `!` operator, because `?.` has stronger binding.

> That cannot be as `x?.(...)` is not an expression.

Sorry, probably I was not clear enough, because you are missing the point.  I used parentheses to show how expressions are parsed/interpreted according to the spec and never meant those to be valid C# expressions. Please re-read my comment keeping that in mind.

pkulikov:
@AntonLapounov if an operator is binary, both its operands are supposed to be valid expressions. `null_conditional_operations` is not a valid expression. To evaluate a binary operator, one should be able to get values of both its operands (if necessary). What is the value of `.y.z`?

If a null-conditional operator is unary or binary. Here is our disagreement. I think it's unary as it takes only one argument (a primary expression). At least, [spec](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#unary-operators) refers to null-conditional expression as unary expression. And after that, the spec says

> If the operand of a `unary_expression` has the compile-time type...

That implies a unary expression (and thus, a null-conditional expression) has one operand. It's according to the C# spec.

When we talk about `?.` and `!`, they both are unary postfix operators. That means that in expression `a?.b.c.d.e!.f` their precedence doesn't matter. Precedence decides to which operator a given operand binds. In the case of unary postfix operations, *every* operand has only one possibility to bind. To the operation on its right. So, evaluation is done simply from left to right. That's why in `a?.b.c.d.e!.f`, the `?.` is evaluated first. To evaluate it, the compiler should form a null-conditional expression that can contain only member and index access operations. `!` "breaks" that sequence. That's why `a?.b.c.d.e!.f` is interpreted as `(a?.b.c.d.e)!.f`

AntonLapounov:
Here is the first sentence of the spec you linked:
> An expression is a sequence of operators and operands.

Later the spec repeats that expressions are constructed from operands and operators.  Now let's consider the expression `x?.y` and identify that sequence of operators and operands:
* `x` is an operand;
* `?.`  is an operator;
* `y` is ???

I think one has to agree that `y` is also an operand; otherwise, we need to modify the definition of the expression to introduce more lexeme kinds.  And if `y` is an operand, what operator does it belong to?  Clearly, it belongs to `?.`, there are no other operators present.  The spec may mention the left argument of `?.` as the (main) _operand_ and the right argument as the _list of operations_ — I don't find that description wrong and it does not make the operator unary since it requires both the left and the right arguments.  Moreover, it follows from the spec that this operator is right-associative.

> What is the value of `.y.z`?

The second argument is the list of operations to apply to the first argument.  You might consider it to be an expression tree that takes `t` and returns `t.y.z`.

> That's why in `a?.b.c.d.e!.f`, the `?.` is evaluated first. To evaluate it, the compiler should form a null-conditional expression that can contain only member and index access operations. `!` "breaks" that sequence. That's why `a?.b.c.d.e!.f` is interpreted as `(a?.b.c.d.e)!.f`

Roughly speaking, the whole point of the operator precedence table is to describe in a systematic way what "breaks the sequence" when we parse arguments of operators.  For example, when we parse the second argument of `+` in `a + b * c.d / e > f`, how do we know to stop at `>`?  It is because the precedence of `>` is lower than the precedence of `+` (and the precedence of `*` and `/` is higher).  Not every grammar may be described this way.  Parsing rules may be complicated and context-dependent, in which case other means, like EBNF, are used to describe them.

pkulikov:
@AntonLapounov thanks for the discussion; I've learnt new things.

AntonLapounov:
@pkulikov Thank you, too! I have also learnt new things about C# grammar.

PR #18007, opened 2020-04-22
C# array guide initialisation length

TimMcGilly:
## Summary

Adds description of setting length in C# array initialisation on https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/arrays/. 

This makes it closer to the VB.Net documentation array example (https://docs.microsoft.com/en-us/dotnet/visual-basic/programming-guide/language-features/arrays/) and C# single array initialisation. (https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/arrays/single-dimensional-arrays)

This is important as the number passed to array initialisation in C# is the length whereas in VB.Net is the maximum index so length + 1.


BillWagner (approved):
Thanks for making this update @TimMcGilly We appreciate it.

PR #18020, opened 2020-04-22
Fix incorrect default value for GCHeapHardLimit

bdrans:
## Summary

GCHeapHardLimit default to the greater of 20MB or 75% of the memory limit. See https://devblogs.microsoft.com/dotnet/running-with-server-gc-in-a-small-container-scenario-part-1-hard-limit-for-the-gc-heap/

### DIFF
--- a/docs/core/run-time-config/garbage-collector.md
+++ b/docs/core/run-time-config/garbage-collector.md
@@ -234,7 +234,7 @@ Example:
 
 - Specifies the maximum commit size, in bytes, for the GC heap and GC bookkeeping.
 - This setting only applies to 64-bit computers.
-- The default value, which only applies in certain cases, is the lesser of 20 MB or 75% of the memory limit on the container. The default value applies if:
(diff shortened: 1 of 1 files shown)


gewarren:
I was wondering about this when I wrote it. To me it doesn't make sense that the limit is the greater, because what if 20 MB is more than 75% of the memory limit on the container - then GC could use all the memory on the container. @Maoni0 which is correct?

Maoni0:
what I said is in the blog is correct. this value is to enforce it so the minimum is not too small. 75% is just a number that we artificially chose when we are not provided with a value so it's not a contract but will be met unless the container's limit is too small in which case we may use up all the memory. for most cases we will be aiming at 75%. note that just because we specified 20mb doesn't mean GC will try to use 20mb. when the memory load is high in the container, GC will do the normal thing which is to be more aggressive at doing full compacting GCs (if GC deems it productive) to keep the heap size down.

gewarren (approved):
@bdrans Thank you!

gewarren:
Thanks for clarifying @Maoni0 .

PR #18021, opened 2020-04-22
Delete 3 orphaned ML files

gewarren:
Also alphabetized redirection file a bit.

Contributes to #17918.


tdykstra (approved):
LGTM, but I think redirect_document_id should be true only if we're changing the URL for the same article, or false when redirecting to a different article.

luisquintanilla (approved):
Looks good

PR #18037, opened 2020-04-23
Correct precedence of the null-conditional operators

pkulikov:
~~By [this comment](https://github.com/dotnet/docs/pull/18001#issuecomment-618339295) the null-conditional operators behave as if their precedence is higher than one of unary operators. I guess that's true because the null-conditional operators are post-fix, and post-fix operators precede pre-fix operators. Still, `x?.y` and `x?[y]` are not primary expressions. Thus, the separate row for them.~~

Update: partially reverts #18001 (the null-conditional operators changes)


docs/csharp/language-reference/operators/index.md
```diff
@@ -37,7 +37,8 @@ The following table lists the C# operators starting with the highest precedence
...
-| [+x](arithmetic-operators.md#unary-plus-and-minus-operators), [-x](arithmetic-operators.md#unary-plus-and-minus-operators), [\!x](boolean-logical-operators.md#logical-negation-operator-), [~x](bitwise-and-shift-operators.md#bitwise-complement-operator-), [++x](arithmetic-operators.md#increment-operator-), [--x](arithmetic-operators.md#decrement-operator---), [^x](member-access-operators.md#index-from-end-operator-), [x?.y](member-access-operators.md#null-conditional-operators--and-), [x?[y]](member-access-operators.md#null-conditional-operators--and-), [(T)x](type-testing-and-cast.md#cast-expression), [await](await.md), [&x](pointer-related-operators.md#address-of-operator-), [*x](pointer-related-operators.md#pointer-indirection-operator-), [true and false](true-false-operators.md) | Unary |
+| `x?.y`, `x?[y]` | [Null-conditional operators](member-access-operators.md#null-conditional-operators--and-) |
```
pkulikov, line 40: @BillWagner [the more I think](https://github.com/dotnet/docs/issues/17988#issuecomment-618819640), the more I see that the following three solutions might make sense for the precedence table:
- bring `?.` and `?[]` back to the first row (pro: it's the same precedence as they are unary post-fix; con: null-conditional operators are not primary (however, we can change the name of the first row from "primary" to "primary and null-conditional")
- place `?.` and `?[]` into the separate row as the current PR does
- remove `?.` and `?[]` from the precedence table at all (for example, the spec doesn't mention the precedence of `?.` and `?[]` explicitly)

What do you think?

AntonLapounov: Placing just `?.` and `?[]` in a separate row is not correct. Option 1 is clearly better as it does not introduce more confusion.

pkulikov: @BillWagner I've updated the table as [per comment](https://github.com/dotnet/docs/issues/17988#issuecomment-620310292)

BillWagner (approved):
Thanks for participating in all the conversations on this issue and PR @pkulikov 

I'll :shipit: now.

PR #18087, opened 2020-04-25
Add discussion of platform sensitive struct/union layout

AaronRobinsonMSFT:
## Summary

Add discussion for platform sensitive struct/union layout.

/cc @gewarren @danmosemsft @elinor-fung @jkoritzinsky

### DIFF
--- a/docs/framework/interop/marshaling-classes-structures-and-unions.md
+++ b/docs/framework/interop/marshaling-classes-structures-and-unions.md
@@ -280,7 +280,9 @@ public struct STRRET_64
 
 Failure to properly consider the native layout in an interop scenario can result in random crashes or worse, incorrect computations.
 
-.NET assemblies by default can run in both a 32-bit and 64-bit version of the .NET runtime so deciding which of the above definitions to use must be deferred until run time. An example of how to make this decision is below.
+By default, .NET assemblies can run in both a 32-bit and 64-bit version of the .NET runtime. The app must wait until run time to decide which of the previous definitions to use.
+
+The following code snippet shows an example of how to choose between the 32-bit and 64-bit definition at run time.
 
 ```CSharp
 if (IntPtr.Size == 8)
--- a/docs/framework/interop/marshaling-classes-structures-and-unions.md
+++ b/docs/framework/interop/marshaling-classes-structures-and-unions.md
@@ -36,7 +36,7 @@ The following table lists marshaling options for classes, structures, and unions
 |Array of structures with integers and strings by reference.|Passes an array of structures that contain integers and strings as an Out parameter. The called function allocates memory for the array.|[OutArrayOfStructs Sample](#outarrayofstructs-sample)|
 |Unions with value types.|Passes unions with value types (integer and double).|[Unions sample](#unions-sample)|
 |Unions with mixed types.|Passes unions with mixed types (integer and string).|[Unions sample](#unions-sample)|
-|Struct with platform specific layout.|Passes type with native packing definitions.|[Platform sample](#platform-sample)|
+|Struct with platform-specific layout.|Passes a type with native-packing definitions.|[Platform sample](#platform-sample)|
 |Null values in structure.|Passes a null reference (**Nothing** in Visual Basic) instead of a reference to a value type.|[HandleRef sample](https://docs.microsoft.com/previous-versions/dotnet/netframework-3.0/hc662t8k(v=vs.85))|
 
 ## Structures sample
--- a/docs/framework/interop/marshaling-classes-structures-and-unions.md
+++ b/docs/framework/interop/marshaling-classes-structures-and-unions.md
@@ -280,6 +280,20 @@ public struct STRRET_64
 
 Notice not only the size but also the field offsets are different. Failure to properly consider the native layout in an interop scenario can result in random crashes or worse, incorrect computation.
 
+.NET assemblies by default can run in both a 32-bit and 64-bit version of the .NET runtime so deciding which of the above definitions to use must be deferred until run time. An example of how to make this decision is below.
+
+```CSharp
+if (IntPtr.Size == 8)
+{
+    // Use the STRRET_64 definition
+}
+else
+{
+    Debug.Assert(IntPtr.Size == 4);
+    // Use the STRRET_32 definition
+}
+```
+
 ## SysTime sample
 
 This sample demonstrates how to pass a pointer to a class to an unmanaged function that expects a pointer to a structure.
--- a/docs/framework/interop/marshaling-classes-structures-and-unions.md
+++ b/docs/framework/interop/marshaling-classes-structures-and-unions.md
@@ -258,7 +258,7 @@ public struct STRRET_32
 }
 ```
 
-Where as on a 64-bit process, the correct layout would be:
+On a 64-bit process, the size *and* field offsets are different. The correct layout is:
 
 ``` CSharp
 [StructLayout(LayoutKind.Explicit, Size = 272)]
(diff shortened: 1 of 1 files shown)


AaronRobinsonMSFT:
This new content applies to both .NET Framework and .NET Core.

docs/framework/interop/marshaling-classes-structures-and-unions.md
```diff
@@ -216,6 +217,69 @@ The `NativeMethods` class contains the prototypes for the `TestUnion` and `TestU
...
+    [FieldOffset(8)]
+    public IntPtr pOleStr;
+
+    [FieldOffset(8)]
+    public uint uOffset;
+
+    [FieldOffset(8)]
+    public IntPtr cStr;
+}
+```
```
jkotas, line 279: It would be nice to also add snippet here that shows the use, e.g. `if (IntPtr.Size == 8) { ... } else { ... }`

jkotas (approved):

docs/framework/interop/marshaling-classes-structures-and-unions.md
```diff
@@ -36,6 +36,7 @@ The following table lists marshaling options for classes, structures, and unions
 |Array of structures with integers and strings by reference.|Passes an array of structures that contain integers and strings as an Out parameter. The called function allocates memory for the array.|[OutArrayOfStructs Sample](#outarrayofstructs-sample)|
 |Unions with value types.|Passes unions with value types (integer and double).|[Unions sample](#unions-sample)|
 |Unions with mixed types.|Passes unions with mixed types (integer and string).|[Unions sample](#unions-sample)|
+|Struct with platform specific layout.|Passes type with native packing definitions.|[Platform sample](#platform-sample)|
```
gewarren, line 39:
```suggestion
|Struct with platform-specific layout.|Passes a type with native-packing definitions.|[Platform sample](#platform-sample)|
```

docs/framework/interop/marshaling-classes-structures-and-unions.md
```diff
@@ -216,6 +217,83 @@ The `NativeMethods` class contains the prototypes for the `TestUnion` and `TestU
 [!code-csharp[Conceptual.Interop.Marshaling#29](~/samples/snippets/csharp/VS_Snippets_CLR/conceptual.interop.marshaling/cs/unions.cs#29)]
 [!code-vb[Conceptual.Interop.Marshaling#29](~/samples/snippets/visualbasic/VS_Snippets_CLR/conceptual.interop.marshaling/vb/unions.vb#29)]
 
+## Platform sample
+
+In some scenarios `struct` and `union` layouts can be different depending on the targeted platform. For example, consider the [`STRRET`](https://docs.microsoft.com/windows/win32/api/shtypes/ns-shtypes-strret) type when defined in a COM scenario:
```
gewarren, line 222:
```suggestion
In some scenarios, `struct` and `union` layouts can differ depending on the targeted platform. For example, consider the [`STRRET`](/windows/win32/api/shtypes/ns-shtypes-strret) type when defined in a COM scenario:
```

docs/framework/interop/marshaling-classes-structures-and-unions.md
```diff
@@ -216,6 +217,83 @@ The `NativeMethods` class contains the prototypes for the `TestUnion` and `TestU
...
+        {
+        /* [case()][string] */ LPWSTR pOleStr;
+        /* [case()] */ UINT uOffset;
+        /* [case()] */ char cStr[ 260 ];
+        }  DUMMYUNIONNAME;
+    }  STRRET;
+#include <poppack.h>
+```
+
+The above `struct` is being declared with Windows' headers that influence the memory layout of the type. When defined in a managed environment these layout details are needed to properly interoperate with native code.
```
gewarren, line 239:
```suggestion
The above `struct` is declared with Windows' headers that influence the memory layout of the type. When defined in a managed environment, these layout details are needed to properly interoperate with native code.
```

docs/framework/interop/marshaling-classes-structures-and-unions.md
```diff
@@ -216,6 +217,83 @@ The `NativeMethods` class contains the prototypes for the `TestUnion` and `TestU
...
+
+The correct managed definition of this type in a 32-bit process would be:
```
gewarren, line 241:
```suggestion
The correct managed definition of this type in a 32-bit process is:
```

docs/framework/interop/marshaling-classes-structures-and-unions.md
```diff
@@ -216,6 +217,83 @@ The `NativeMethods` class contains the prototypes for the `TestUnion` and `TestU
...
+
+    [FieldOffset(8)]
+    public uint uOffset;
+
+    [FieldOffset(8)]
+    public IntPtr cStr;
+}
+```
+
+Notice not only the size but also the field offsets are different. Failure to properly consider the native layout in an interop scenario can result in random crashes or worse, incorrect computation.
```
gewarren, line 281:
```suggestion
Failure to properly consider the native layout in an interop scenario can result in random crashes or worse, incorrect computations.
```

docs/framework/interop/marshaling-classes-structures-and-unions.md
```diff
@@ -216,6 +217,83 @@ The `NativeMethods` class contains the prototypes for the `TestUnion` and `TestU
...
+
+    [FieldOffset(4)]
+    public uint uOffset;
+
+    [FieldOffset(4)]
+    public IntPtr cStr;
+}
+```
+
+Where as on a 64-bit process, the correct layout would be:
```
gewarren, line 261:
```suggestion
On a 64-bit process, the size *and* field offsets are different. The correct layout is:
```

docs/framework/interop/marshaling-classes-structures-and-unions.md
```diff
@@ -216,6 +217,83 @@ The `NativeMethods` class contains the prototypes for the `TestUnion` and `TestU
...
+    public uint uOffset;
+
+    [FieldOffset(8)]
+    public IntPtr cStr;
+}
+```
+
+Notice not only the size but also the field offsets are different. Failure to properly consider the native layout in an interop scenario can result in random crashes or worse, incorrect computation.
+
+.NET assemblies by default can run in both a 32-bit and 64-bit version of the .NET runtime so deciding which of the above definitions to use must be deferred until run time. An example of how to make this decision is below.
```
gewarren, line 283:
```suggestion
By default, .NET assemblies can run in both a 32-bit and 64-bit version of the .NET runtime. The app must wait until run time to decide which of the previous definitions to use.

The following code snippet shows an example of how to choose between the 32-bit and 64-bit definition at run time.
```

gewarren (approved):
Thanks for adding this. I left some grammatical/flow suggestions.

jkoritzinsky:
Another thing that might be worth mentioning here is that if instead of creating one explicit struct, if you create a sequential structure for STREET with a nested explicit structure with everything at offset 0 that represents the implicit union field, then you don't need two definitions to account for bitness. Alignment considerations should automatically handle layout cross-bitness in that scenario.

AaronRobinsonMSFT:
> Another thing that might be worth mentioning here is that if instead of creating one explicit struct, if you create a sequential structure for STREET with a nested explicit structure with everything at offset 0 that represents the implicit union field, then you don't need two definitions to account for bitness. Alignment considerations should automatically handle layout cross-bitness in that scenario.

@jkoritzinsky That could be possible, but that may not be true for COM-RPC. How the structs are defined and that they can be different based on bitness is the intent of this sample. The above workaround could work in some scenarios but it could also fail in others. We should avoid those kind of tricks in official documentation.

gewarren:
Looks good. I'm not sure if you have write permissions, so I'm merging this now.

PR #18102, opened 2020-04-27
Clean up the .NET Core "Containerize an app with Docker" tutorial 

IEvangelist:
## Summary

- Took ownership of docker containers with .NET Core
- Adjust directory structure
- Use runtime, not aspnet base image
- Added tabs for Windows and Linux
- Use docker slugs instead of console (title cased as they are supported but not labeled on the platform)
- Fix mismatching image ids
- Provide a name to the container
- Made `<kbd>` callouts consistent
- Modernized C# bits
- Use active voice where possible
- Detailed the MCR, repository vs image vs registry
- Added example of passing a parameter to the container for counting app


docs/core/docker/build-container.md
```diff
@@ -22,58 +22,71 @@ You'll learn to:
...
 
-```console
-> dotnet run
+```dotnetcli
+dotnet run
 Hello World!
 ```
 
-The default template creates an app that prints to the terminal and then exits. For this tutorial, you'll use an app that loops indefinitely. Open the *Program.cs* file in a text editor. It should currently look like the following code:
+The default template creates an app that prints to the terminal and then exits immediately terminates. For this tutorial, you'll use an app that loops indefinitely. Open the *Program.cs* file in a text editor.
```
Thraka, line 73:
```suggestion
The default template creates an app that prints to the terminal and immediately terminates. For this tutorial, you'll use an app that loops indefinitely. Open the *Program.cs* file in a text editor.
```

docs/core/docker/build-container.md
```diff
@@ -22,58 +22,71 @@ You'll learn to:
...
+        ├──Program.cs
+        └──obj
+            ├──NetCore.Docker.csproj.nuget.dgspec.json
+            ├──NetCore.Docker.csproj.nuget.g.props
+            ├──NetCore.Docker.csproj.nuget.g.targets
+            ├──project.assets.json
+            └──project.nuget.cache
+```
+
+The `dotnet new` command creates a new folder named *App* and generates a "Hello World" console application. First, changes directories and navigate into the *App* folder.
```
Thraka, line 60:
```suggestion
The `dotnet new` command creates a new folder named *App* and generates a "Hello World" console application. First, change directories and navigate into the *App* folder.
```

docs/core/docker/build-container.md
```diff
@@ -22,58 +22,71 @@ You'll learn to:
...
+
+```dotnetcli
+cd App
 ```
 
-The `dotnet new` command creates a new folder named *app* and generates a "Hello World" app. Enter the *app* folder and run the command `dotnet run`. You'll see the following output:
+Execute the `dotnet run` command. The application will run, and print `Hello World!` below the command:
 
-```console
```
Thraka, line 66: This article may not be using the latest styles we've come up with, but currently we use `dotnetcli` only when the only commands run are a dotnet command. Otherwise, it's possible that coloring is wrong. You can kind of see that with the `dontnetcli` codeblock above, `cd` is colored.

So we've adopted using `dotnetcli` when the only command (without output) is the `dotnet` command. If it's a command + output, use `console` or separate them with something like:

> Execute the `dotnet run` command.
> -- Codeblock dotnetcli with dotnet run command
> The application prints "Hello World".
> -- Codeblock console with with hello world printed

docs/core/docker/build-container.md
```diff
@@ -122,205 +135,194 @@ Counter: 4
...
-> [!NOTE]
+> [!IMPORTANT]
 > Any parameters after `--` are not passed to the `dotnet run` command and instead are passed to your application.
 
 ## Publish .NET Core app
 
-Before you add your .NET Core app to the Docker image, publish it. You want to make sure that the container runs the published version of the app when it's started.
-
-From the working folder, enter the *app* folder with the example source code and run the following command:
+Before adding the .NET Core app to the Docker image, first it must be published. It is best to have the container run the published version of the app. To publish the app, execute the following command:
```
Thraka, line 143: We're not supposed to use the word "Execute" which, in translation, may not mean run, but mean kill 😄 There are a few places where you've used this word.

https://styleguides.azurewebsites.net/Styleguide/Read?id=2700&topicid=33512

docs/core/docker/build-container.md
```diff
@@ -122,205 +135,194 @@ Counter: 4
...
-  FROM mcr.microsoft.com/dotnet/core/aspnet:3.1
-  ```
-
-- .NET Core runtime
-
-  ```dockerfile
-  FROM mcr.microsoft.com/dotnet/core/runtime:3.1
-  ```
+```dockerfile
+FROM mcr.microsoft.com/dotnet/core/runtime:3.1
```
Thraka, line 186: The product group wants to push aspnet runtime as the runtime of choice as its the most compatible of the two (since it includes asp.net core stuff).

And, with the initial WARNING I added, and this section that points to ASP.NET (but leaves the option for non-ASP) the amount of complaints and bad feedback of "can't get my app to work" have pretty much disappeared. A lot of people are using ASP.NET Core and end up here...

docs/core/docker/build-container.md
```diff
@@ -122,205 +135,194 @@ Counter: 4
...
-If you're using Linux or macOS, use the `ls` command to get a directory listing and verify that the *myapp.dll* file was created.
+#### [Linux](#tab/linux)
+
+Use the `ls` command to get a directory listing and verify that the *NetCore.Docker.dll* file was created.
 
 ```bash
 me@DESKTOP:/docker-working/app$ ls bin/Release/netcoreapp3.1/publish
-myapp.deps.json  myapp.dll  myapp.pdb  myapp.runtimeconfig.json
+NetCore.Docker.deps.json  NetCore.Docker.dll  NetCore.Docker.pdb  NetCore.Docker.runtimeconfig.json
 ```
```
Thraka, line 178: You're missing the `---` end tabs region syntax. Also, the guide says that the tabs should be single `#` headers. https://review.docs.microsoft.com/en-us/help/contribute/validation-ref/tabbed-conceptual?branch=master

I don't know if the `#` header makes a difference, but in general we do that. If we use some automation to change them in the future, it would be good to have them consistent. 😄

Thraka (approved):
Looking great! Some things to consider before merging. A bug to fix related to tabs too.

docs/core/docker/build-container.md
IEvangelist: Great catch, I wish that this feature was automated and validated... Also, the `#` was giving me warnings in the markdown lint (as it treated those as H1's), I know that they work regardless of H1-6, as I used them in the Azure docs a lot.

docs/core/docker/build-container.md
IEvangelist: This one I'd like to push back on. We emphasize that if they're doing ASP.NET Core, there is a different article for that and that this article is explicitly using .NET Core. Without that, our docs would have zero examples of using the `runtime` image - if that is the case, why bother having it at all? I believe that this is the most appropriate place for this usage. Perhaps a compromise would be to include an alert letting them know about the aspnet image?

docs/core/docker/build-container.md
IEvangelist: Good catch

docs/core/docker/build-container.md
Thraka: OK keep it then :D I have disabled the warning in the linter then reenabled it later in the article, but that is more work than necessary. I like your approach better 😄

PR #18123, opened 2020-04-28
Corrected contributor guide link on hubs and other docs.

IEvangelist:
## Summary

Corrected contributor guide link on hubs and other docs. Noticed that the current link takes users to a MD file in GitHub, which simply points to another doc.


BillWagner (approved):
Thanks for spotting this and submitting the fix @IEvangelist 

You can :shipit: now.

PR #18124, opened 2020-04-28
Add note about interactions in operators

### LINKED ISSUE #17988: null-forgiving operator *can have* effect at run time

BillWagner:
Fixes #17988.  See https://github.com/dotnet/docs/issues/17988#issuecomment-620310292

The null conditional operators and null forgiving operators interact in interesting ways. Add notes on both pages to explain those interactions.

### DIFF
--- a/docs/csharp/language-reference/operators/member-access-operators.md
+++ b/docs/csharp/language-reference/operators/member-access-operators.md
@@ -132,6 +132,9 @@ In the preceding example, if you don't use the `??` operator, `numbers?.Length <
 
 The null-conditional member access operator `?.` is also known as the Elvis operator.
 
+> [!NOTE]
+> The null conditional operators can interact with the [null forgiving operator](null-forgiving.md) in unexpected ways. In C# 8 the expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation `z` is evaluated even if `x` is null, which may result in a <xref:System.NullReferenceException>.
+
 ### Thread-safe delegate invocation
 
 Use the `?.` operator to check if a delegate is non-null and invoke it in a thread-safe way (for example, when you [raise an event](../../../standard/events/how-to-raise-and-consume-events.md)), as the following code shows:
--- a/docs/csharp/language-reference/operators/member-access-operators.md
+++ b/docs/csharp/language-reference/operators/member-access-operators.md
@@ -133,7 +133,7 @@ In the preceding example, if you don't use the `??` operator, `numbers?.Length <
 The null-conditional member access operator `?.` is also known as the Elvis operator.
 
 > [!NOTE]
-> In C# 8, the null-conditional operators can interact with the [null-forgiving operator](null-forgiving.md) in unexpected ways. For example, the expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation, `z` is evaluated even if `x` is `null`, which may result in a <xref:System.NullReferenceException>.
+> In C# 8, the null-conditional operators interacts with the [null-forgiving operator](null-forgiving.md) in an unexpected way. For example, the expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation, `z` is evaluated even if `x` is `null`, which may result in a <xref:System.NullReferenceException>.
 
 ### Thread-safe delegate invocation
 
--- a/docs/csharp/language-reference/operators/member-access-operators.md
+++ b/docs/csharp/language-reference/operators/member-access-operators.md
@@ -133,7 +133,7 @@ In the preceding example, if you don't use the `??` operator, `numbers?.Length <
 The null-conditional member access operator `?.` is also known as the Elvis operator.
 
 > [!NOTE]
-> The null conditional operators can interact with the [null forgiving operator](null-forgiving.md) in unexpected ways. In C# 8 the expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation `z` is evaluated even if `x` is null, which may result in a <xref:System.NullReferenceException>.
+> The null-conditional operators can interact with the [null-forgiving-operator](null-forgiving.md) in unexpected ways. In C# 8 the expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation `z` is evaluated even if `x` is null, which may result in a <xref:System.NullReferenceException>.
 
 ### Thread-safe delegate invocation
 
--- a/docs/csharp/language-reference/operators/member-access-operators.md
+++ b/docs/csharp/language-reference/operators/member-access-operators.md
@@ -133,7 +133,7 @@ In the preceding example, if you don't use the `??` operator, `numbers?.Length <
 The null-conditional member access operator `?.` is also known as the Elvis operator.
 
 > [!NOTE]
-> The null-conditional operators can interact with the [null-forgiving operator](null-forgiving.md) in unexpected ways. In C# 8 the expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation `z` is evaluated even if `x` is null, which may result in a <xref:System.NullReferenceException>.
+> In C# 8, the null-conditional operators can interact with the [null-forgiving operator](null-forgiving.md) in unexpected ways. For example, the expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation, `z` is evaluated even if `x` is `null`, which may result in a <xref:System.NullReferenceException>.
 
 ### Thread-safe delegate invocation
 
--- a/docs/csharp/language-reference/operators/member-access-operators.md
+++ b/docs/csharp/language-reference/operators/member-access-operators.md
@@ -133,7 +133,7 @@ In the preceding example, if you don't use the `??` operator, `numbers?.Length <
 The null-conditional member access operator `?.` is also known as the Elvis operator.
 
 > [!NOTE]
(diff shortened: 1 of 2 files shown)


docs/csharp/language-reference/operators/member-access-operators.md
```diff
@@ -132,6 +132,9 @@ In the preceding example, if you don't use the `??` operator, `numbers?.Length <
 
 The null-conditional member access operator `?.` is also known as the Elvis operator.
 
+> [!NOTE]
+> The null conditional operators can interact with the [null forgiving operator](null-forgiving.md) in unexpected ways. In C# 8 the expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation `z` is evaluated even if `x` is null, which may result in a <xref:System.NullReferenceException>.
```
AntonLapounov, line 136: For the sake of consistency, could you place add a hyphen in 'null-conditional' and 'null-forgiving' on both pages?

AntonLapounov (approved):

docs/csharp/language-reference/operators/member-access-operators.md
```diff
@@ -133,7 +133,7 @@ In the preceding example, if you don't use the `??` operator, `numbers?.Length <
 The null-conditional member access operator `?.` is also known as the Elvis operator.
 
 > [!NOTE]
-> The null conditional operators can interact with the [null forgiving operator](null-forgiving.md) in unexpected ways. In C# 8 the expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation `z` is evaluated even if `x` is null, which may result in a <xref:System.NullReferenceException>.
+> The null-conditional operators can interact with the [null-forgiving-operator](null-forgiving.md) in unexpected ways. In C# 8 the expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation `z` is evaluated even if `x` is null, which may result in a <xref:System.NullReferenceException>.
```
AntonLapounov, line 136: Sorry for nitpicking, but I think the hyphen before 'operator' is not needed.

BillWagner: Thanks. Fixed it. It's that kind of day. 🤦

docs/csharp/language-reference/operators/member-access-operators.md
```diff
@@ -132,6 +132,9 @@ In the preceding example, if you don't use the `??` operator, `numbers?.Length <
 
 The null-conditional member access operator `?.` is also known as the Elvis operator.
 
+> [!NOTE]
+> The null-conditional operators can interact with the [null-forgiving operator](null-forgiving.md) in unexpected ways. In C# 8 the expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation `z` is evaluated even if `x` is null, which may result in a <xref:System.NullReferenceException>.
```
pkulikov, line 136:
```suggestion
> In C# 8.0, a null-conditional expression doesn't include the [null-forgiving operator](null-forgiving.md). That is, expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation, `z` is evaluated even if `x` is `null`, which may result in a <xref:System.NullReferenceException>.
```

docs/csharp/language-reference/operators/null-forgiving.md
```diff
@@ -13,6 +13,9 @@ Available in C# 8.0 and later, the unary postfix `!` operator is the null-forgiv
 
 The null-forgiving operator has no effect at run time. It only affects the compiler's static flow analysis by changing the null state of the expression. At run time, expression `x!` evaluates to the result of the underlying expression `x`.
 
+> [!NOTE]
+> The null-forgiving operator can interact with the [null-conditional operators](member-access-operators.md#null-conditional-operators--and-) in unexpected ways. In C# 8 the expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation `z` is evaluated even if `x` is null, which may result in a <xref:System.NullReferenceException>.
```
pkulikov, line 17:
```suggestion
> In C# 8.0, a [null-conditional](member-access-operators.md#null-conditional-operators--and-) expression doesn't include the null-forgiving operator. That is, expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation, `z` is evaluated even if `x` is `null`, which may result in a <xref:System.NullReferenceException>.
```

pkulikov (commented):
After all those discussions, I wouldn't call that behavior "unexpected". (That is, the docs shouldn't use subjective language.)

docs/csharp/language-reference/operators/null-forgiving.md
BillWagner: That's a good point @pkulikov I made some edits, but I did leave the word "unexpected" in. In this instance, it's appropriate.

pkulikov: @BillWagner thanks, then I have two nit-picks:
- Should "can interact" be "interacts"? As "can interact" suggests that it doesn't interact always like that
- Should "unexpected ways" be "unexpected way"?

docs/csharp/language-reference/operators/member-access-operators.md
```diff
@@ -132,6 +132,9 @@ In the preceding example, if you don't use the `??` operator, `numbers?.Length <
 
 The null-conditional member access operator `?.` is also known as the Elvis operator.
 
+> [!NOTE]
+> In C# 8, the null-conditional operators interacts with the [null-forgiving operator](null-forgiving.md) in an unexpected way. For example, the expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation, `z` is evaluated even if `x` is `null`, which may result in a <xref:System.NullReferenceException>.
```
AntonLapounov, line 136: @BillWagner Sorry, I did not look at the last commit yesterday and a grammar error crept in: "operators interacts".  Also I rather agree with @pkulikov's comment:
> After all those discussions, I wouldn't call that behavior "unexpected". (That is, the docs shouldn't use subjective language.)

I would prefer a neutral description like this:
> In C# 8, the [null-forgiving operator](null-forgiving.md) terminates the list of preceding null-conditional operations. For example, the expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation, `z` is evaluated even if `x` is `null`, which may result in a <xref:System.NullReferenceException>.

I changed only the first sentence to describe "unexpected interaction" in more precise words. If you consider that an improvement, please let me know if I should create a PR for that.

docs/csharp/language-reference/operators/null-forgiving.md
```diff
@@ -13,6 +13,9 @@ Available in C# 8.0 and later, the unary postfix `!` operator is the null-forgiv
 
 The null-forgiving operator has no effect at run time. It only affects the compiler's static flow analysis by changing the null state of the expression. At run time, expression `x!` evaluates to the result of the underlying expression `x`.
 
+> [!NOTE]
+> In C# 8 the null-forgiving operator interacts with the [null-conditional operators](member-access-operators.md#null-conditional-operators--and-) in an unexpected way. The expression `x?.y!.z` is parsed as `(x?.y)!.z`. Due to this interpretation `z` is evaluated even if `x` is `null`, which may result in a <xref:System.NullReferenceException>.
```
AntonLapounov, line 17: I know commas are frequently optional in English. Comparing to the previous note, this note misses commas after "In C# 8" and "Due to this interpretation". It also lost the "For example" part. Not sure if it was intentional.

docs/csharp/language-reference/operators/member-access-operators.md
BillWagner: Thanks @AntonLapounov I like that improved wording. I'll watch for a PR.

docs/csharp/language-reference/operators/null-forgiving.md
BillWagner: It was not intentional. If you want to fix this in the same PR (along with the updated wording), I'll merge it.

PR #18150, opened 2020-04-29
Clarified wording for dependency-version behavior.

samrueby:
## Summary

This documentation misleadingly said that NuGet **always** looks for the lowest dependency version. This is misleading because it is only the default behavior and is easily overridden. Changed _always_ to _default_ and provided a link to where the option can be overridden and that 'lowest' is the default.


JamesNK (approved):

gewarren:
Thank you @samrueby

PR #18158, opened 2020-04-29
Correct .NET Core SDK and CLI overview links

rainersigwald:
These had gotten swapped somehow.

cc @mairaw


gewarren (approved):
Good catch, thanks!

PR #18024, opened 2020-04-23
Update Hub + Landing pages to prep for L2 removal

wadepickett:
PR updates part 1 for #18011

[Internal Review - Hub](https://review.docs.microsoft.com/en-us/dotnet/?branch=pr-en-us-18024)
[Internal Review - Framework Landing Page](https://review.docs.microsoft.com/en-us/dotnet/framework/?branch=pr-en-us-18024)
[Internal Review - Standard Landing Page](https://review.docs.microsoft.com/en-us/dotnet/standard/?branch=pr-en-us-18024)
[Internal Review - What's New](https://review.docs.microsoft.com/en-us/dotnet/whats-new/?branch=pr-en-us-18024)
[Internal Review - Visual Basic Landing Page](https://review.docs.microsoft.com/en-us/dotnet/visual-basic/?branch=pr-en-us-18024)

Hub page:
- Add Download .NET highlight card to top left.
- Add a "Community" link plus a .NET Foundation link to the open source card

What's New page:
- Add a “Community” link to the “Get Involved” card.

.NET Framework landing page:
- Add link to .NET Download on "Install .NET Framework section"

.NET Standard landing Page:
- Add link to .NET Download as first link in Learn about .NET section

What's New page:
- Add Community and .NET Foundation links to "Get involved - contribute" section.

The L2 itself will be removed in a separate PR once these changes are merged.

PR #18159, opened 2020-04-29
Fix link to .NET Core CLI overview

gewarren:
Introduced by #18024 and #18158 by mistake.

### DIFF
--- a/docs/index.yml
+++ b/docs/index.yml
@@ -109,7 +109,7 @@ conceptualContent:
         - url: https://dotnet.microsoft.com/download
           itemType: download
           text: Download the .NET Core SDK
-        - url: core/sdk.md
+        - url: core/tools/index.md
           itemType: overview
           text: .NET Core CLI overview
         - url: core/porting/index.md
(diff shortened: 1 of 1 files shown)


docs/index.yml
```diff
@@ -109,7 +109,7 @@ conceptualContent:
         - url: https://dotnet.microsoft.com/download
```
tdykstra, line 109:
```suggestion
        - url: https://dotnet.microsoft.com/download/dotnet-core/3.1
```

tdykstra (approved):
LGTM, suggestion to make the download link more closely fit the link text.

wadepickett (approved):
Thanks for the fix.  This is correct.

docs/index.yml
gewarren: The problem with the suggestion is that it will need to be updated for each release. Since "Download .NET" is the first giant box on the page, how about we change this link back to https://docs.microsoft.com/en-us/dotnet/core/sdk? (But as `core/sdk.md`.)

gewarren: image

tdykstra: Good idea 👍

PR #18168, opened 2020-04-29
Add UriBuilder breaking change

### LINKED ISSUE #18161: UriBuilder.Query and UriBuilder.Fragment behavior changed in Core
## UriBuilder.Query and UriBuilder.Fragment behavior changed in Core

Moving from Framework to Core, UriBuilder behavior for Query and Fragment properties was changed. It now no longer prepends a leading ? for query and # for fragment when one is already present.

This results in a difference in behavior
```c#
new UriBuilder { Query = "?foo" }.Query

// ??foo on Framework
// ?foo on Core
```

### Version introduced

.NET Core 1.0

### Old behavior

Always prepend a leading ? or #.

### New behavior

Prepend a leading ? or # only if one isn't already present.

### Reason for change

Make the behavior more intuitive - you don't end up with double ?? or ## unless you specify them explicitly. See more discussion at dotnet/corefx#3647

Also fixes behavior when appending to the property
```c#
var builder = new UriBuilder();
builder.Query = "one=1";
builder.Query += "&two=2";
builder.Query += "&three=3";
builder.Query += "&four=4";
            
Console.WriteLine(builder.Query);

// ????one=1&two=2&three=3&four=4 on Framework
// ?one=1&two=2&three=3&four=4 on Core
```

### Recommended action

- Document the change

### Category

- Core .NET libraries

### Affected APIs

[UriBuilder.Query](https://docs.microsoft.com/en-us/dotnet/api/system.uribuilder.query?view=net-5.0)
[UriBuilder.Fragment](https://docs.microsoft.com/en-us/dotnet/api/system.uribuilder.fragment?view=net-5.0)

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #18161

### DIFF
--- a/docs/core/compatibility/corefx.md
+++ b/docs/core/compatibility/corefx.md
@@ -30,6 +30,7 @@ The following breaking changes are documented on this page:
 | [OpenSSL versions on macOS](#openssl-versions-on-macos) | 2.1 |
 | [UnauthorizedAccessException thrown by FileSystemInfo.Attributes](#unauthorizedaccessexception-thrown-by-filesysteminfoattributes) | 1.0 |
 | [Handling corrupted-process-state exceptions is not supported](#handling-corrupted-state-exceptions-is-not-supported) | 1.0 |
+| [UriBuilder properties no longer prepend leading characters](#uribuilder-properties-no-longer-prepend-leading-characters) | 1.0 |
 
 ## .NET Core 3.0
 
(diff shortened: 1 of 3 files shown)


gewarren:
We also need to update https://docs.microsoft.com/en-us/dotnet/api/system.uribuilder.fragment?view=netcore-3.1 and https://docs.microsoft.com/en-us/dotnet/api/system.uribuilder.query?view=netcore-3.1.

tdykstra (approved):
LGTM

PR #18200, opened 2020-05-01
Update dotnet-install-scripts

### LINKED ISSUE #14842: Install script is not creating registry values
The install script for Windows is not creating registry values like the Installer on dot.net does. This is a problem because I'm programmatically using these registry keys to determine if .NET Core is installed and use the script if not but since the script is not creating the registry keys its repeating itself.

Could it be the installing script just downloads the zipped binaries and copy them to a folder instead of downloading the installer and installs it? In this case the docs need to tell that and there should be an option to download the installers instead.

---
#### Document Details

* ID: a2e484a9-5e0c-eb3e-1413-7b507c4fa07d
* Version Independent ID: 5bc7caa5-f1c2-d7d5-fe02-80c65cca9672
* Content: [dotnet-install scripts - .NET Core CLI](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script#feedback)
* Content Source: [docs/core/tools/dotnet-install-script.md](https://github.com/dotnet/docs/blob/master/docs/core/tools/dotnet-install-script.md)
* Product: **dotnet-core**
* Technology: **dotnet-cli**
* GitHub Login: @mairaw
* Microsoft Alias: **mairaw**

tdykstra:
Fixes #14842

### DIFF
--- a/docs/core/tools/dotnet-install-script.md
+++ b/docs/core/tools/dotnet-install-script.md
@@ -1,7 +1,7 @@
 ---
 title: dotnet-install scripts
 description: Learn about the dotnet-install scripts to install the .NET Core SDK and the shared runtime.
-ms.date: 01/23/2020
+ms.date: 04/30/2020
 ---
 # dotnet-install scripts reference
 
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @tdykstra 

You can :shipit: when ready.

PR #18217, opened 2020-05-01
Add arm64 RIDS  to the portable RID list

KhArtNJava:
## Summary

Added new RIDs for ARM64 devices.


BillWagner:
Ping @Thraka for review. It looks like this is an area you understand better.

tdykstra:
Conflict with #18214, which was just merged.

Thraka (changes requested):
@KhArtNJava I've fixed the merge conflict. I wouldn't add only Ubuntu-arm64 to the new `Ubuntu` section. Can you please fill out some other common Ubuntu values? We do point to the `runtime.json` file as the definitive list, so we don't have to list everything, but having some other Ubuntu besides arm would be good.

Thraka:
@KhArtNJava I'm going to close this for now. Please feel free to reopen when you can find some time to work on this a bit. Thanks!

Issue #18207, opened 2020-05-01
"x op= y" is not equivalent to "x = x op y"


carsonRadtke:
Consider this example:

bool b = false;
String s = null;
b = b && s.Length &gt; 0; // no error because evaluation is lazy
b &= s.Length &gt; 0; // crashes (null reference exception) because evaluation is eager

Thus "x op= y" cannot always be replaced by "x = x op y"

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6d2d3240-ed65-94f2-335b-3ea483c51a5d
* Version Independent ID: 91eaa72e-c2fc-e0d3-ca11-994aadf02271
* Content: [Boolean logical operators - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators#feedback)
* Content Source: [docs/csharp/language-reference/operators/boolean-logical-operators.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/operators/boolean-logical-operators.md)
* Product: **dotnet-csharp**
* Technology: **csharp-language-reference**
* GitHub Login: @pkulikov
* Microsoft Alias: **wiwagn**

BillWagner:
Thanks for writing this @carsonRadtke 

The doc is correct, but misleading.

`&=` is the compound operator form of `&` not `&&`. The `&` operator does not short-circuit, so throwing an exception is correct, and equivalent:

```csharp
b = b && s.Length > 0; // no error because && short-circuits.
b = b & s.Length > 0; // NullReferenceException because & does not short-circuit.
b &= s.Length > 0; // NullReferenceException because &= evaluates both operands.
```csharp

This statement (from the article) should be emphasized more: 

> The conditional logical operators and don't support compound assignment. `&&` `||`.

This statement means there is no `&&=` or `||=` operators. 

Note the extra "and" should also be fixed.

carsonRadtke:
@BillWagner thank you for your response, I missed the "The conditional logical operators && and || don't support compound assignment."

PR #18229, opened 2020-05-02
Emphasize that there are no &&= and ||= operators

pkulikov:
As result of #18207


BillWagner (approved):
Thanks @pkulikov 
I think this added note will make this stand out more.

PR #18242, opened 2020-05-03
Removed an unused type in code listing.

ebresafegaga:
The BananaPeel class is unsued and I don't think it should be in the code listing.


cartermp (approved):
Thanks!

PR #18244, opened 2020-05-03
Update statically-resolved-type-parameters.md

ebresafegaga:
inline functions can be parameterized with a standard generic type parameter.

### DIFF
--- a/docs/fsharp/language-reference/generics/statically-resolved-type-parameters.md
+++ b/docs/fsharp/language-reference/generics/statically-resolved-type-parameters.md
@@ -28,7 +28,7 @@ The following table summarizes the similarities and differences between the two 
 |Member constraints|Cannot be used with member constraints.|Can be used with member constraints.|
 |Code generation|A type (or method) with standard generic type parameters results in the generation of a single generic type or method.|Multiple instantiations of types and methods are generated, one for each type that is needed.|
 |Use with types|Can be used on types.|Cannot be used on types.|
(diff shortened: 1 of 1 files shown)


cartermp (changes requested):
This isn't quite right. Adding the standard generic type parameters will no longer use static type parameters, forcing the compiler to specialize the function to a specific type. This should be added to the sentence.

ebresafegaga:
> This isn't quite right. Adding the standard generic type parameters will no longer use static type parameters, forcing the compiler to specialize the function to a specific type. This should be added to the sentence.

Yes, that's true.  But can an inline function have standard generic type parameters (CLR generics) without any static type parameters on it?  The docs says this is not possible at all.

cartermp:
It can, yes. But not in the context of static types, which is probably why this was explicitly called out as not possible here. I think it could get reworded, but it need to be explicit about the fact that this forces specialization with CLR generics and thus doesn't let you do things like generic arithmetic

adegeo:
@cartermp What should we do with this PR?

cartermp:
As it stands we'll need more clarification in this PR to take it further. Right now the document is correct, though could be improved. The current change would no longer make it correct.

IEvangelist:
Hi @ebresafegaga - just working backwards through open PRs. Are you going to make the adjustments that Phillip suggested, or continue the discussion?

adegeo:
Closing as abandoned.

PR #18303, opened 2020-05-07
Update to Console WebApiClient page

OwainWilliams:
## Summary

Removing the need to rename the namespace from ConsoleApp as the setup instructions names the namespace correctly.


BillWagner (approved):
Thanks for spotting this and fixing it @OwainWilliams We appreciate it.

PR #18326, opened 2020-05-08
Add read from stream example for Utf8JsonReader

### LINKED ISSUE #16941: Add Utf8JsonReader code example for partial reads
See https://github.com/dotnet/dotnet-api-docs/issues/3831 and 
https://github.com/dotnet/runtime/issues/2265#issuecomment-580594288

---
#### Document Details

* Product: **dotnet**
* GitHub Login: @tdykstra

tdykstra:
Fixes #16941

[Internal review URL](https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to?branch=pr-en-us-18326#use-utf8jsonreader)


samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,78 @@
...
+            }
+
+            Console.WriteLine($"Found property name: {reader.GetString()}");
+            Console.WriteLine($"String in buffer is: {System.Text.Encoding.UTF8.GetString(span)}");
+            if (!reader.Read())
+            {
+                // Not enough of the JSON is in the buffer to complete a read.
+                GetMoreBytesFromStream(stream, ref buffer, ref span, ref reader);
+                reader.Read();
+            }
```
Jozkee, line 50: In this particular scenario, calling `GetMoreBytesFromStream` once works, but in the case that your property value is too big, you may need to call this method more than once so I would suggest to change this to a while loop.
```suggestion
            while (!reader.Read())
            {
                // Not enough of the JSON is in the buffer to complete a read.
                GetMoreBytesFromStream(stream, ref buffer, ref span, ref reader);
            }
```

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,78 @@
...
+            System.Console.WriteLine($"Found property value: {reader.GetString()}");
```
Jozkee, line 51: I don't know if saying "Found" is correct here since you are not "searching" for the property value, it is right next to the property name.

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        private static void GetMoreBytesFromStream(MemoryStream stream, ref byte[] buffer, ref Span<byte> span, ref Utf8JsonReader reader)
+        {
+            if (reader.BytesConsumed < span.Length)
+            {
+                ReadOnlySpan<byte> leftover = span.Slice((int)reader.BytesConsumed);
+
+                if (leftover.Length >= span.Length)
```
Jozkee, line 60:
```suggestion
                if (leftover.Length == span.Length)
```

It is not possible that we have more leftover bytes than the buffer size.

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -775,6 +775,18 @@ Here's a JSON sample that the preceding code can read. The resulting summary mes
...
 
+### Read from a stream using Utf8JsonReader
+
+When using the `Utf8JsonReader` to read from a stream, the following rules apply:
+
+* The buffer must be at least as big as the largest token.
+* The buffer must be at least as big as the largest whitespace.
+* The reader doesn't keep track of the data it has read until it completely reads the next <xref:System.Text.Json.Utf8JsonReader.TokenType%2A> in the JSON payload. So when there are bytes left over in the buffer, you have to pass them to the reader again. You can use <xref:System.Text.Json.Utf8JsonReader.BytesConsumed%2A> to determine how many bytes are left over.
+
+The following code illustrates how to read from a stream. The code starts with a 10-byte buffer and doubles the buffer size each time it finds that the size is not big enough.
```
Jozkee, line 786:
```suggestion
The following code illustrates how to read from a stream. The code starts with a 10-byte buffer and doubles the buffer size each time it finds that the size is not big enough to allocate the largest token or whitespace.
```
Should we be more specific here?

Jozkee:
It looks great, just a few suggestions in the code and in the snippet description. 
cc @ahsonkhan @layomia in case they want to add more comments.

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
tdykstra:
```suggestion
            System.Console.WriteLine($"Got property value: {reader.GetString()}");
```

BillWagner (approved):
This LGTM as well @tdykstra 

You can :shipit: when ready.

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,77 @@
...
+                ""Date"": ""2019-08-01T00:00:00-07:00"",
+                ""Temperature"": 25,
+                ""TemperatureRanges"": {
+                    ""Cold"": { ""High"": 20, ""Low"": -10 },
+                    ""Hot"": { ""High"": 60, ""Low"": 20 }
+                },
+                ""Summary"": ""Hot"",
+            }";
+
+            var bytes = System.Text.Encoding.UTF8.GetBytes(jsonString.Trim());
```
ahsonkhan, line 21: nit: Don't use `var` where it isn't clear from the right hand side what the type should be (just by reading the code with no a priori knowledge). That is the coding style guideline we follow in the dotnet/runtime repo.

What that translates to is, unless you are using `new` on the R.H.S. or an explicit cast/generic type, don't use var.

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,77 @@
...
+            var stream = new MemoryStream(bytes);
+
+            var buffer = new byte[10];
+            var span = new Span<byte>(buffer);
+
+            // Fill the buffer
+            var read = stream.Read(span);
```
ahsonkhan, line 28: Same nit here - don't use `var`

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.IO;
+using System.Text.Json;
+
+namespace SystemTextJsonSamples
+{
+    public class Utf8ReaderPartialRead
+    {
+        public static void Run()
+        {
+            var jsonString = @"{
```
ahsonkhan, line 11: It might be too complicated (or not feasible), but maybe we can show a more realistic example where the user is trying to read/process a much larger JSON payload files (in the GBs), which makes it clear why partially reading from a stream is an important scenario. Most small payloads would just fit in memory and reading partial/streaming data for that isn't really useful.

Similarly, reading from a `FileStream` is more useful in motivating the use case, rather than `MemoryStream`.

Ignore the suggestion if not possible.

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,77 @@
...
+            }
+
+            Console.WriteLine($"Found property name: {reader.GetString()}");
+            Console.WriteLine($"String in buffer is: {System.Text.Encoding.UTF8.GetString(span)}");
+            while (!reader.Read())
+            {
+                // Not enough of the JSON is in the buffer to complete a read.
+                GetMoreBytesFromStream(stream, ref buffer, ref span, ref reader);
+            }
+            System.Console.WriteLine($"Got property value: {reader.GetString()}");
```
ahsonkhan, line 50: Remove `System.` from such calls. Makes the sample unnecessarily verbose and we already have `using System;` above.
```suggestion
            Console.WriteLine($"Got property value: {reader.GetString()}");
```

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                if (!reader.Read())
+                {
+                    // Not enough of the JSON is in the buffer to complete a read.
+                    GetMoreBytesFromStream(stream, ref buffer, ref span, ref reader);
+                }
+            }
+
+            Console.WriteLine($"Found property name: {reader.GetString()}");
+            Console.WriteLine($"String in buffer is: {System.Text.Encoding.UTF8.GetString(span)}");
```
ahsonkhan, line 44: Similarly here and elsewhere, add a `using System.Text.Encoding` directive at the top.
```suggestion
            Console.WriteLine($"String in buffer is: {UTF8.GetString(span)}");
```

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,77 @@
...
+
+                leftover.CopyTo(span);
+                stream.Read(span.Slice(leftover.Length));
+            }
+            else
+            {
+                stream.Read(span);
+            }
+            Console.WriteLine($"String in buffer is: {System.Text.Encoding.UTF8.GetString(span)}");
+            reader = new Utf8JsonReader(span, false, reader.CurrentState);
```
ahsonkhan, line 74: When stream is out of data, i.e. `stream.Read` returns 0, we need to pass `isFinalBlock: true` to the `Utf8JsonReader`, rather than false. Otherwise, the reader will keep expecting more data which will lead to incorrect validation of the JSON payload while parsing.

https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.read?view=netcore-3.1#System_IO_Stream_Read_System_Span_System_Byte__

Also, nit, use named parameters.
```suggestion
            reader = new Utf8JsonReader(span, isFinalBlock: false, reader.CurrentState);
```

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        private static void GetMoreBytesFromStream(MemoryStream stream, ref byte[] buffer, ref Span<byte> span, ref Utf8JsonReader reader)
+        {
+            if (reader.BytesConsumed < span.Length)
+            {
+                ReadOnlySpan<byte> leftover = span.Slice((int)reader.BytesConsumed);
+
+                if (leftover.Length == span.Length)
+                {
+                    Array.Resize(ref buffer, buffer.Length * 2);
```
ahsonkhan, line 61: This will try to grow indefinitely if the token size is too large and eventually throw OutOfMemoryException. This can happen if the JSON contains a token that is ~1 GB in size, because ~2*1GB becomes too large to fit into an int32 buffer.

That should almost never happen in practice (and the caller might actually be fine with the unbounded growth and OOM in that case anyway), but maybe add a comment calling that out?

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,77 @@
...
+            var bytes = System.Text.Encoding.UTF8.GetBytes(jsonString.Trim());
+            var stream = new MemoryStream(bytes);
+
+            var buffer = new byte[10];
+            var span = new Span<byte>(buffer);
+
+            // Fill the buffer
+            var read = stream.Read(span);
+
+            var reader = new Utf8JsonReader(span, false, state: default);
```
ahsonkhan, line 30: Same nit here, used named parameters.
```suggestion
            var reader = new Utf8JsonReader(span, isFinalBlock: false, state: default);
```

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
ahsonkhan, line 21: Trimming isn't necessary. Remove it.
```suggestion
            byte[] bytes = Encoding.UTF8.GetBytes(jsonString);
```

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,77 @@
...
+            Console.WriteLine($"String in buffer is: {System.Text.Encoding.UTF8.GetString(span)}");
+            while (!reader.Read())
+            {
+                // Not enough of the JSON is in the buffer to complete a read.
+                GetMoreBytesFromStream(stream, ref buffer, ref span, ref reader);
+            }
+            System.Console.WriteLine($"Got property value: {reader.GetString()}");
+        }
+
+        private static void GetMoreBytesFromStream(MemoryStream stream, ref byte[] buffer, ref Span<byte> span, ref Utf8JsonReader reader)
```
ahsonkhan, line 53: I don't think you need to keep passing the `span` parameter around. Using the `buffer` should be enough since span is just a view over the `byte[]` anyway and their size is identical.

You could pass in the `byte[] buffer` to the `Utf8JsonReader` directly since arrays and spans are implicitly convertible. @Jozkee - could you help Tom with this refactor and cleanup?

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,77 @@
...
+            while (reader.TokenType != JsonTokenType.PropertyName || reader.GetString() != "Summary")
+            {
+                if (!reader.Read())
+                {
+                    // Not enough of the JSON is in the buffer to complete a read.
+                    GetMoreBytesFromStream(stream, ref buffer, ref span, ref reader);
+                }
+            }
+
+            Console.WriteLine($"Found property name: {reader.GetString()}");
```
ahsonkhan, line 43: As the reader of the code, I can't tell what is being written to  console and why. Should we place what we expect the output to be in comment?

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,77 @@
...
+            var span = new Span<byte>(buffer);
+
+            // Fill the buffer
+            var read = stream.Read(span);
+
+            var reader = new Utf8JsonReader(span, false, state: default);
+            Console.WriteLine($"String in buffer is: {System.Text.Encoding.UTF8.GetString(span)}");
+
+            // Search for "Summary" property name
+            while (reader.TokenType != JsonTokenType.PropertyName || reader.GetString() != "Summary")
```
ahsonkhan, line 34: Use ValueTextEquals when comparing and searching for a property name. It is more efficient because you don't have to create/allocate a string for all the properties that you are not looking for along the way.
https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.valuetextequals?view=netcore-3.1#System_Text_Json_Utf8JsonReader_ValueTextEquals_System_String_

```suggestion
            while (reader.TokenType != JsonTokenType.PropertyName || !reader.ValueTextEquals("Summary"))
```

tdykstra:
@jozkee / @ahsonkhan please review code and text changes.

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -776,6 +776,22 @@ Here's a JSON sample that the preceding code can read. The resulting summary mes
...
+
+* The buffer containing the partial JSON payload must be at least as big as the largest JSON token within it so that the reader can make forward progress.
+* The buffer must be at least as big as the largest sequence of white space within the JSON.
+* The reader doesn't keep track of the data it has read until it completely reads the next <xref:System.Text.Json.Utf8JsonReader.TokenType%2A> in the JSON payload. So when there are bytes left over in the buffer, you have to pass them to the reader again. You can use <xref:System.Text.Json.Utf8JsonReader.BytesConsumed%2A> to determine how many bytes are left over.
+
+The following code illustrates how to read from a stream. The example shows a <xref:System.IO.MemoryStream>, but similar code will work with a <xref:System.IO.FileStream>. The code starts with a 4KB buffer and doubles the buffer size each time it finds that the size is not big enough to fit a complete JSON token, which is required for the reader to make forward progress on the JSON payload. The JSON sample provided in the snippet triggers a buffer size increase only if you set a very small initial buffer size, for example, 10 bytes.
```
ahsonkhan, line 789: The code currently doesn't start with a 4KB buffer, so currently this comment is a bit confusing. I can provide a re-wording suggestion later.

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,78 @@
...
+            }";
+
+            byte[] bytes = Encoding.UTF8.GetBytes(jsonString);
+            var stream = new MemoryStream(bytes);
+
+            var buffer = new byte[10];
+            var span = new Span<byte>(buffer);
+
+            // Fill the buffer
+            stream.Read(span);
```
ahsonkhan, line 29: If the stream is closed/empty at this point, `stream.Read` will return 0 and then the `Utf8JsonReader` will throw. We may want to check against that or at least add a comment that we are assuming the stream has enough JSON data to fill the 10-byte buffer.

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,78 @@
...
+                    ""Hot"": { ""High"": 60, ""Low"": 20 }
+                },
+                ""Summary"": ""Hot"",
+            }";
+
+            byte[] bytes = Encoding.UTF8.GetBytes(jsonString);
+            var stream = new MemoryStream(bytes);
+
+            var buffer = new byte[10];
+            var span = new Span<byte>(buffer);
```
ahsonkhan, line 26: You don't need this local. You can pass `buffer` directly to all the places that take `span`.

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,78 @@
...
+                ""Temperature"": 25,
+                ""TemperatureRanges"": {
+                    ""Cold"": { ""High"": 20, ""Low"": -10 },
+                    ""Hot"": { ""High"": 60, ""Low"": 20 }
+                },
+                ""Summary"": ""Hot"",
+            }";
+
+            byte[] bytes = Encoding.UTF8.GetBytes(jsonString);
+            var stream = new MemoryStream(bytes);
```
ahsonkhan, line 23: In the case of reading from a filestream that might contain a UTF-8 BOM at the start, the caller needs to strip those three bytes from the buffer before passing the bytes to the `Utf8JsonReader`. Otherwise the reader would throw (since the BOM is not considered a valid part of the JSON).

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,78 @@
...
+            }
+            Console.WriteLine($"Got property value: {reader.GetString()}");
+        }
+
+        private static void GetMoreBytesFromStream(MemoryStream stream, ref byte[] buffer, ref Utf8JsonReader reader)
+        {
+            int bytesRead;
+            if (reader.BytesConsumed < buffer.Length)
+            {
+                ReadOnlySpan<byte> leftover = buffer.AsSpan().Slice((int)reader.BytesConsumed);
```
ahsonkhan, line 59: Use the AsSpan() overloads that take ints rather than slicing after. Here/elsewhere:
https://docs.microsoft.com/en-us/dotnet/api/system.memoryextensions.asspan?view=netcore-3.1#System_MemoryExtensions_AsSpan__1___0___System_Int32_

```suggestion
                ReadOnlySpan<byte> leftover = buffer.AsSpan((int)reader.BytesConsumed);
```

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,78 @@
...
+
+                if (leftover.Length == buffer.Length)
+                {
+                    Array.Resize(ref buffer, buffer.Length * 2);
+                    Console.WriteLine($"Increased buffer size to {buffer.Length}");
+                }
+
+                leftover.CopyTo(buffer.AsSpan());
```
ahsonkhan, line 67: nit: The implicit cast should just work. You can remove `AsSpan()` here.
```suggestion
                leftover.CopyTo(buffer);
```

docs/standard/serialization/system-text-json-how-to.md
tdykstra: One of my code changes was to change the buffer definition to 4096, but I changed it back to 10 for testing and apparently neglected to restore 4096.

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
tdykstra: Where can I find an example of code that strips the BOM?

ahsonkhan: This isn't exactly applicable here, but it gives the general idea:
https://github.com/dotnet/runtime/blob/b9524694bc8f8b3676a6bcf25eb7586194693b38/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializer.Read.Stream.cs#L108-L137
https://github.com/dotnet/runtime/blob/b9524694bc8f8b3676a6bcf25eb7586194693b38/src/libraries/System.Text.Json/src/System/Text/Json/Document/JsonDocument.Parse.cs#L570-L607

Sorry, I am heads down on some other work, so I can't give a drag-n-drop answer right now. Maybe others on the team can chime in.

A comment might be sufficient for now, to avoid blocking on it.

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,79 @@
...
+            var stream = new MemoryStream(bytes);
+
+            var buffer = new byte[4096];
+
+            // Fill the buffer.
+            // For this snippet, we're assuming the stream is open and has data.
+            // If it might be closed or empty, check if the return value is 0.
+            stream.Read(buffer);
+
+            var reader = new Utf8JsonReader(buffer, isFinalBlock: false, state: default);
```
ahsonkhan, line 32: I am curious, does this sample successfully run when the buffer is larger than the payload, i.e. 4096. I would have thought that `isFinalBlock: false` when the entire JSON is provided in one chunk would have caused an exception.

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -776,6 +776,24 @@ Here's a JSON sample that the preceding code can read. The resulting summary mes
...
+
+When using the `Utf8JsonReader` to read from a stream, the following rules apply:
+
+* The buffer containing the partial JSON payload must be at least as big as the largest JSON token within it so that the reader can make forward progress.
+* The buffer must be at least as big as the largest sequence of white space within the JSON.
+* The reader doesn't keep track of the data it has read until it completely reads the next <xref:System.Text.Json.Utf8JsonReader.TokenType%2A> in the JSON payload. So when there are bytes left over in the buffer, you have to pass them to the reader again. You can use <xref:System.Text.Json.Utf8JsonReader.BytesConsumed%2A> to determine how many bytes are left over.
+
+The following code illustrates how to read from a stream. The example shows a <xref:System.IO.MemoryStream>. Similar code will work with a <xref:System.IO.FileStream>, except when the FileStream contains a UTF-8 BOM at the start. In that case, you need to strip those three bytes from the buffer before passing the remaining bytes to the `Utf8JsonReader`. Otherwise the reader would throw an exception, since the BOM is not considered a valid part of the JSON.
```
ahsonkhan, line 789: Nice!

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,79 @@
...
+            {
+                if (!reader.Read())
+                {
+                    // Not enough of the JSON is in the buffer to complete a read.
+                    GetMoreBytesFromStream(stream, ref buffer, ref reader);
+                }
+            }
+
+            // Found the "Summary" property name.
+            Console.WriteLine($"String in buffer is: {Encoding.UTF8.GetString(buffer)}");
```
ahsonkhan, line 46: What does  this output? Do we care and if so, should we show the reader in the comment as well of what this would output? The same applies to all other `Console.WriteLine` calls.

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
```diff
@@ -0,0 +1,79 @@
...
+            while (!reader.Read())
+            {
+                // Not enough of the JSON is in the buffer to complete a read.
+                GetMoreBytesFromStream(stream, ref buffer, ref reader);
+            }
+            Console.WriteLine($"Got property value: {reader.GetString()}");
```
ahsonkhan, line 52: Same here, what does the output look like? I am guessing it is `"Hot"`

ahsonkhan (approved):
Otherwise, looks good.

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
tdykstra: It doesn't cause an exception, it works fine at 4096.

docs/standard/serialization/system-text-json-how-to.md
tdykstra:
```suggestion
The following code illustrates how to read from a stream. The example shows a <xref:System.IO.MemoryStream>. Similar code will work with a <xref:System.IO.FileStream>, except when the `FileStream` contains a UTF-8 BOM at the start. In that case, you need to strip those three bytes from the buffer before passing the remaining bytes to the `Utf8JsonReader`. Otherwise the reader would throw an exception, since the BOM is not considered a valid part of the JSON.
```

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
tdykstra: What it shows depends on the buffer size -- at 4096, which the snippet has now, the entire sample JSON is displayed.  If you change it to 10, it shows 10 byte sections of the JSON at a time until the buffer is increased to 20 and then to 40, after which the sections shown are correspondingly bigger. It basically just gives a visual record of when and why the buffer is increased.

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
tdykstra: Yes that's what it shows.  I'll add a comment to that effect.

samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs
tdykstra: Thanks, I'll go with a comment for now and create an issue to refine the presentation later.

jubba2001:
The sample "samples/snippets/core/system-text-json/csharp/Utf8ReaderPartialRead.cs" does not work correctly.
The code will return an empty "Summary" value if its length is greater than the size of the buffer, for example:

           var jsonString = @"{
                ""Date"": ""2019-08-01T00:00:00-07:00"",
                ""Temperature"": 25,
                ""TemperatureRanges"": {
                    ""Cold"": { ""High"": 20, ""Low"": -10 },
                    ""Hot"": { ""High"": 60, ""Low"": 20 }
                },
                ""Summary"": ""A_1234567890_B_1234567890_C_1234567890_D_1234567890"",
            }";

In this case, the example returning a value should be modified to look like this: 
 
           while (!reader.Read())
            {
                // Not enough of the JSON is in the buffer to complete a read.
                GetMoreBytesFromStream(stream, ref buffer, ref span, ref reader);
                //reader.Read();
            }
            Console.WriteLine($"Found property value: {reader.GetString()}");

PR #18351, opened 2020-05-11
1657393: Add missing descriptions (1 of 2)

TimShererWithAquent:
## Summary

We are adding descriptions to improve SEO. Suggestions are welcome, but keep in mind our 100-160 char range. 

@IEvangelist - You're listed as author for these topics. Please review when you get a chance. Our SLA specifies 5 business days for review, but let me know if you need more time. (Also, let me know if you are not the right person to approve. We go by the github author tag.)

Fixes [1657393](https://dev.azure.com/mseng/TechnicalContent/_workitems/edit/1657393)


docs/standard/assembly/contents.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Assembly contents
+description: A static assembly can consist the assembly manifest containing assembly metadata, type metadata, code that implements types, and a set of resources.
```
IEvangelist, line 3: Everything else looks good, except for this one. The description doesn't make sense to me - could you please reword it?

IEvangelist (changes requested):

docs/standard/assembly/contents.md
TimShererWithAquent: @IEvangelist  - How about this version: 

This article describes the contents of a .NET assembly, which can include assembly metadata, type metadata, MSIL code, and resources.

I would have just said that the article describes the contents, but that was way under the 100 char min.

IEvangelist: Sounds good :shipit:

IEvangelist (approved):
Looks good to me :shipit:

Issue #15506, opened 2019-10-29
Review `dotnet` documentation


dsplaisted:
https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet

tdykstra:
@dsplaisted The CLI ref docs were recently reviewed and updated for 3.x -- does that address what you had in mind when created these "Review..." issues for the various dotnet command pages?  In other words, can we close these issues now or is further review needed?

wli3:
I reviewed everything SDK owns. Runtime folks need to review runtime arguments and environment variables

tdykstra:
>Runtime folks need to review runtime arguments and environment variables

@vitek-karas  You reviewed a PR for the `dotnet` command CLI ref doc relatively recently (https://github.com/dotnet/docs/pull/17183) -- do you want to take a look at the current state of the doc for this issue?

tdykstra:
@vitek-karas ping

vitek-karas:
I created #18368 which adds the missing options and makes other improvements in the area.

PR #18368, opened 2020-05-12
Add missing .NET Core runtime options and env. variables

vitek-karas:
Added command line options which were missing from the documentation. Also added several environment variables which were not documented.

Improved documentation for some of the options already listed.

Related to #15506


docs/core/tools/dotnet.md
```diff
@@ -300,10 +316,27 @@ dotnet myapp.dll
...
+  Specifies a directory to which a single-file application is extracted before it is executed.
+
+  For more information, see [Single-file executables](../whats-new/dotnet-core-3-0.md#single-file-executables).
+
 - `COREHOST_TRACE`, `COREHOST_TRACEFILE`, `COREHOST_TRACE_VERBOSITY`
 
   Controls diagnostics tracing from the hosting components, such as `dotnet.exe`, `hostfxr`, and `hostpolicy`.
 
+  * `COREHOST_TRACE=[0/1]` - default is `0` - tracing disabled. If set to `1` the diagnostics tracing is enabled.
+  * `COREHOST_TRACEFILE=<file path>` **Available starting with .NET Core 3.x.** - only has effect if tracing is enabled via `COREHOST_TRACE=1`. When set the tracing information is written to the specified file, otherwise the tracing information is written to `stderr`.
```
tdykstra, line 330:
```suggestion
  * `COREHOST_TRACEFILE=<file path>` - only has effect if tracing is enabled via `COREHOST_TRACE=1`. When set, the tracing information is written to the specified file, otherwise the tracing information is written to `stderr`. **Available starting with .NET Core 3.x.** 
```

docs/core/tools/dotnet.md
```diff
@@ -300,10 +316,27 @@ dotnet myapp.dll
...
+  * `COREHOST_TRACE_VERBOSITY=[1/2/3/4]` **Available starting with .NET Core 3.x.** - default is `4`. The setting is used only when tracing is enabled via `COREHOST_TRACE=1`.
```
tdykstra, line 331:
```suggestion
  * `COREHOST_TRACE_VERBOSITY=[1/2/3/4]` - default is `4`. The setting is used only when tracing is enabled via `COREHOST_TRACE=1`. **Available starting with .NET Core 3.x.**
```

docs/core/tools/dotnet.md
```diff
@@ -300,10 +316,27 @@ dotnet myapp.dll
...
+    * `4` - all tracing information is written
+    * `3` - only informational, warning and error messages are written
+    * `2` - only warning and error messages are written
+    * `1` - only error messages are written
+  `COREHOST_TRACEFILE` and `COREHOST_TRACE_VERBOSITY` is only supported starting with .NET Core 3.0.
```
tdykstra, line 336:
```suggestion

  `COREHOST_TRACEFILE` and `COREHOST_TRACE_VERBOSITY` are supported starting with .NET Core 3.0.
```

docs/core/tools/dotnet.md
```diff
@@ -300,10 +316,27 @@ dotnet myapp.dll
...
+
+  Typical way to get a detailed trace information about the application startup is to set `COREHOST_TRACE=1` and `COREHOST_TRACEFILE=host_trace.txt` and then run the application. A new file `host_trace.txt` will be created in the current directory with the detailed information.
```
tdykstra, line 338:
```suggestion
  The typical way to get detailed trace information about application startup is to set `COREHOST_TRACE=1` and `COREHOST_TRACEFILE=host_trace.txt` and then run the application. A new file `host_trace.txt` will be created in the current directory with the detailed information.
```

docs/core/whats-new/dotnet-core-3-0.md
```diff
@@ -229,6 +229,8 @@ Don't roll forward. Only bind to specified version. This policy isn't recommende
 
 Besides the **Disable** setting, all settings will use the highest available patch version.
 
+By default if the requested version (as specified in `.runtimeconfig.json` for the application) is a release version, only release versions are considered for roll forward. Any pre-release versions are ignored. Only if there are no matching release version, then also pre-release versions are taken into account. This behavior can be changed by setting `DOTNET_ROLL_FORWARD_TO_PRERELEASE=1` in which case all versions are always considered.
```
tdykstra, line 232:
```suggestion
By default, if the requested version (as specified in `.runtimeconfig.json` for the application) is a release version, only release versions are considered for roll forward. Any pre-release versions are ignored. If there is no matching release version, then pre-release versions are taken into account. This behavior can be changed by setting `DOTNET_ROLL_FORWARD_TO_PRERELEASE=1`, in which case all versions are always considered.
```

docs/core/tools/dotnet.md
```diff
@@ -114,24 +114,14 @@ The following options are available when `dotnet` runs an application. For examp
...
-
-  - `0` - Disable even minor version roll forward.
-  - `1` - Roll forward on minor version, but not on major version. This is the default behavior.
-  - `2` - Roll forward on minor and major versions.
-
-   For more information, see [Roll forward](../whats-new/dotnet-core-2-1.md#roll-forward).
-
 - **`--roll-forward <SETTING>`** **Available starting with .NET Core SDK 3.0.**
 
   Controls how roll forward is applied to the app. The `SETTING` can be one of the following values. If not specified, `Minor` is the default.
```
tdykstra, line 127:
```suggestion
  Controls how roll forward is applied to the app. The `SETTING` can be one of the following values. If not specified, `Minor` is the default.
  
```

tdykstra (approved):
Thank you very much for all these updates and corrections!  I made some minor stylistic suggestions.

gewarren (commented):
Should we have a "dotnet" page in the run-time config settings section of the TOC for these environment variables?

https://docs.microsoft.com/en-us/dotnet/core/run-time-config/

tdykstra:
@gewarren How about adding a section heading to https://docs.microsoft.com/en-us/dotnet/core/run-time-config/ and link to the env variables section of this doc?

vitek-karas:
I know this is confusing (maybe we need better terminology). Basically there are two sets of "properties" and "configuration variables".

The "host" configuration:
* Command line options as described in this doc
* Env. variables which start with the `DOTNET_` (meant to be used by developers, admins, ...) and `COREHOST_` (diagnostics only)
* Some portion of `.runtimeconfig.json` file
* Almost all of `.deps.json` file

The "runtime" configuration:
* Properties portion of the `.runtimeconfig.json`
* Env. variables which start with the `COMPLUS_` prefix

The doc I'm updating is talking solely about the "host" configuration options. The other doc you mention talks only about the "runtime" configuration options.

The problem is that for normal user there's no real difference between "host" and "runtime" - they both are some tech which runs the app. So we refer to both as "runtime" I guess.

tdykstra:
@vitek-karas Thanks for the runtime vs. host explanation!

gewarren:
Thanks for the explanation. I do think it will be confusing to users and we could help to clarify the difference and interlink between these pages. Also, there are some [profiling config](https://github.com/dotnet/aspnetcore/issues/12505) options documented in the run-time config section that start with `CORECLR_`. Are those host or runtime options?

vitek-karas:
Both of those would fall under the true "runtime" options in my explanation above. I think it's unfortunate we don't have a standardized naming scheme for the env. variables (unlike the `.runtimeconfig.json` properties where the naming is relatively consistent).

PR #18376, opened 2020-05-12
Remove single-project sln files

Youssef1313:
This was done automatically using the following code:

```csharp
void Main()
{
    foreach (string s in Directory.GetFiles(@"D:\dotnet\docs\docs", "*.sln", SearchOption.AllDirectories))
    {
        if (CountEndProject(File.ReadAllText(s)) == 1) File.Delete(s);
    }
}

int CountEndProject(string fileContent)
{
    return (fileContent.Length - fileContent.Replace("EndProject","").Length) / "EndProject".Length;
}
```


BillWagner (approved):
Thanks @Youssef1313 We appreciate it.

I'll :shipit: now.

PR #18382, opened 2020-05-12
1657393: Add missing descriptions (3 of 3)

TimShererWithAquent:
## Summary

Thanks for jumping in on the first two of these so fast!

We are adding descriptions to improve SEO, per [1657393](https://dev.azure.com/mseng/TechnicalContent/_workitems/edit/1657393). Suggestions are welcome, but keep in mind our 100-160 char range. 

@tdykstra - You're listed as author for these topics. Please review when you get a chance. Our SLA specifies 5 business days for review, but let me know if you need more time. (Also, let me know if you are not the right person to approve. We go by the github author tag.)

Fixes #1657393


tdykstra (approved):
LGTM

PR #18388, opened 2020-05-12
Delete unused powershell snippets

Youssef1313:
They're only used in dotnet-api-docs, and they exist there.


BillWagner (commented):
Thanks for spotting this @Youssef1313 

The build is clean, so I'll :shipit:

BillWagner (approved):

PR #18393, opened 2020-05-13
Correcting mislabeled slot numbers

noahfalk:
Previously some slot numbers were 0-based counts, some were 1-based counts, and some were 2-based counts.
Now all slot numbers are consistently 1-based counts.


hoyosjs (approved):
Thanks for fixing this

IEvangelist (approved):
Looks good to me :shipit:

PR #18441, opened 2020-05-14
Add breaking change for Process.StartInfo

### LINKED ISSUE #18310: Process.StartInfo changes
Reading the `StartInfo` property of a process that you did not start (within the current process, using `Process.Start`) will throw `InvalidOperationException`

### Version introduced

.NET Core v1.0

### Old behavior

In .NET Framework, it will return a dummy `ProcessStartInfo` object. This object will return default values from all its properties, except `EnvironmentVariables`, which will return the current process's environment variables.

However, it is [documented](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.startinfo?view=netcore-3.1#remarks) not to do this:  "You should only access the StartInfo property on a Process object returned by the Start method. For example, you should not access the StartInfo property on a Process object returned by GetProcesses. Otherwise, on .NET Core the StartInfo property will throw an InvalidOperationException and on .NET Framework it will return a dummy ProcessStartInfo object."

### New behavior

It throws `InvalidOperationException`

### Reason for change

The object returned was a dummy object, with no value; it was also documented that it should not be gotten.

### Recommended action

Stop calling this property getter.

### Category

CoreFX

### Affected APIs

[Process.Start](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.start?view=netcore-3.0)
[ProcessStartInfo](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo?view=netcore-3.0)

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #18310

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/corefx?branch=pr-en-us-18441#processstartinfo-throws-invalidoperationexception-for-processes-you-didnt-start)

### DIFF
--- a/docs/core/compatibility/corefx.md
+++ b/docs/core/compatibility/corefx.md
@@ -32,6 +32,7 @@ The following breaking changes are documented on this page:
 | [UnauthorizedAccessException thrown by FileSystemInfo.Attributes](#unauthorizedaccessexception-thrown-by-filesysteminfoattributes) | 1.0 |
 | [Handling corrupted-process-state exceptions is not supported](#handling-corrupted-state-exceptions-is-not-supported) | 1.0 |
 | [UriBuilder properties no longer prepend leading characters](#uribuilder-properties-no-longer-prepend-leading-characters) | 1.0 |
+| [Process.StartInfo throws InvalidOperationException for processes you didn't start](#processstartinfo-throws-invalidoperationexception-for-processes-you-didnt-start) | 1.0 |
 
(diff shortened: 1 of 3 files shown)


gewarren:
I'm using the terminology "read the property" instead of "access its getter". Is that okay?

danmosemsft:
Yes, any regularization of terms is good. thanks

danmoseley (approved):

PR #18454, opened 2020-05-15
Enclose languages with quotes in dotnet new

Youssef1313:
As noted in the article,

> Some shells interpret `#` as a special character. In those cases, enclose the language parameter value in quotes. For example, `dotnet new console -lang "F#"`.

Also, the syntax highlighting interpret the `#` as a comment (it appears in green color).


tdykstra (approved):

tdykstra:
Thanks!

PR #18485, opened 2020-05-18
Move system-text-json snippets

Youssef1313:
@Thraka, I couldn't follow the exact structure here, because the same files are being used in different articles. So, I kept the folder name as `system-text-json` and the files in that folder are referenced from different articles, all those articles are `system-text-json-<something>.md`.


tdykstra:
Thanks, but I just made a PR a while ago to take care of this.

Youssef1313:
@tdykstra Oh sorry, I didn't notice it 😄

PR #18490, opened 2020-05-18
Definition of examineNumber was never used

xerxesb:
## Summary

The second sample for Partial Active Patterns in F# defined a function `examineNumber`, but this function was never used in printing the output of squares and cubes. I believe it is leftover after an older example was changed, and should be removed as it's confusing for someone new to active patterns.


cartermp (approved):
Thanks for the catch!

PR #18492, opened 2020-05-18
Add globalization API breaking change

### LINKED ISSUE #18423: String comparison, sorting and globalization use ICU instead of NLS in Windows 19H1+
## Globalization support change on Windows 10 19H1+

We changed the underlying library that we use in order to get localization data, perform string/char sorting and comparisons. If an application is running on Windows 19H1 or later which ships with ICU native library, we will use ICU as the library to perform this operations instead of Windows NLS APIs. If we fail to load/find ICU on the OS we will fallback and use NLS.

### Version introduced

.NET 5.0 Preview4

### Old behavior

Globalization behavior was mainly derived from Windows NLS native APIs.

### New behavior

Globalization behavior will be mainly derived from ICU library shipped with Windows.

### Reason for change

1. Provide apps the same behavior in between different OSs (Linux, macOS, Windows).
2. Allow apps control the globalization behavior by supporting having the apps use custom-built ICU library.

### Recommended action

If customers want to fall back to use NLS, they can add a runtime switch in their runtimeconfig.template.json:
```js
{
    "configProperties": {
        "System.Globalization.UseNls": true
    }
}
```

They can also set an environment variable:
`DOTNET_SYSTEM_GLOBALIZATION_USENLS=true`

### Category

- Core
- Core .NET libraries
- Globalization

### Affected APIs

  - [System.String](https://docs.microsoft.com/en-us/dotnet/api/system.string?view=net-5.0)
  - [System.Span](https://docs.microsoft.com/en-us/dotnet/api/system.span-1?view=net-5.0)
  - [System.Globalization namespace classes](https://docs.microsoft.com/en-us/dotnet/api/system.globalization?view=net-5.0)

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #18423

[Breaking change preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/globalization?branch=pr-en-us-18492#globalization-apis-use-icu-libraries-on-windows).

[Run-time config preview link](https://review.docs.microsoft.com/en-us/dotnet/core/run-time-config/globalization?branch=pr-en-us-18492#nls).

### DIFF
--- a/docs/core/compatibility/globalization.md
+++ b/docs/core/compatibility/globalization.md
@@ -9,10 +9,15 @@ The following breaking changes are documented on this page:
 
 | Breaking change | Version introduced |
 | - | :-: |
+| [Globalization APIs use ICU libraries on Windows](#globalization-apis-use-icu-libraries-on-windows) | 5.0 |
 | [StringInfo and TextElementEnumerator are now UAX29-compliant](#stringinfo-and-textelementenumerator-are-now-uax29-compliant) | 5.0 |
 | ["C" locale maps to the invariant locale](#c-locale-maps-to-the-invariant-locale) | 3.0 |
 
 ## .NET 5.0
+
+[!INCLUDE [icu-globalization-api](../../../includes/core-changes/globalization/5.0/icu-globalization-api.md)]
+
+***
 
 [!INCLUDE [uax29-compliant-grapheme-enumeration](../../../includes/core-changes/globalization/5.0/uax29-compliant-grapheme-enumeration.md)]
 
--- a/docs/core/run-time-config/globalization.md
+++ b/docs/core/run-time-config/globalization.md
@@ -80,7 +80,7 @@ Project file:
 ## NLS
 
 - Determines whether .NET uses National Language Support (NLS) or International Components for Unicode (ICU) globalization APIs for Windows apps. .NET 5.0 and later versions use ICU globalization APIs by default on Windows 10 May 2019 Update and later versions.
-- Default: Use ICU globalization APIs (`false`).
+- If you omit this setting, .NET uses ICU globalization APIs by default. This is equivalent to setting the value to `false`.
 - For more information, see [Globalization APIs use ICU libraries on Windows](../compatibility/3.1-5.0.md#globalization-apis-use-icu-libraries-on-windows).
 
 | | Setting name | Values | Introduced |
--- a/docs/core/run-time-config/globalization.md
+++ b/docs/core/run-time-config/globalization.md
@@ -1,7 +1,7 @@
 ---
 title: Globalization config settings
 description: Learn about run-time settings that configure globalization aspects of a .NET Core app, for example, how it parses Japanese dates.
-ms.date: 11/27/2019
+ms.date: 05/18/2020
 ms.topic: reference
 ---
 # Run-time configuration options for globalization
@@ -76,3 +76,14 @@ Project file:
 | - | - | - |
 | **runtimeconfig.json** | `Switch.System.Globalization.FormatJapaneseFirstYearAsANumber` | `false` - format as "Gannen"<br/>`true` - format as number |
 | **Environment variable** | N/A | N/A |
+
+## NLS
+
+- Determines whether Windows uses National Language Support (NLS) or International Components for Unicode (ICU) globalization APIs. Starting in .NET 5.0, .NET libraries use ICU globalization APIs by default.
+- Default: Use ICU globalization APIs (`false`).
+- For more information, see [Globalization breaking changes](../compatibility/3.1-5.0.md#globalization).
+
+| | Setting name | Values | Introduced |
+| - | - | - |
+| **runtimeconfig.json** | `System.Globalization.UseNls` | `false` - Use ICU globalization APIs<br/>`true` - Use NLS globalization APIs | .NET 5.0 |
+| **Environment variable** | `DOTNET_SYSTEM_GLOBALIZATION_USENLS` | `false` - Use ICU globalization APIs<br/>`true` - Use NLS globalization APIs | .NET 5.0 |
--- a/docs/core/run-time-config/globalization.md
+++ b/docs/core/run-time-config/globalization.md
@@ -79,9 +79,9 @@ Project file:
 
 ## NLS
 
-- Determines whether Windows uses National Language Support (NLS) or International Components for Unicode (ICU) globalization APIs. Starting in .NET 5.0, .NET libraries use ICU globalization APIs by default.
+- Determines whether .NET uses National Language Support (NLS) or International Components for Unicode (ICU) globalization APIs for Windows apps. .NET 5.0 and later versions use ICU globalization APIs by default on Windows 10.
 - Default: Use ICU globalization APIs (`false`).
-- For more information, see [Globalization breaking changes](../compatibility/3.1-5.0.md#globalization).
+- For more information, see [Globalization APIs use ICU libraries on Windows](../compatibility/3.1-5.0.md#globalization-apis-use-icu-libraries-on-windows).
 
 | | Setting name | Values | Introduced |
 | - | - | - |
--- a/docs/core/run-time-config/globalization.md
+++ b/docs/core/run-time-config/globalization.md
@@ -79,11 +79,11 @@ Project file:
 
 ## NLS
 
-- Determines whether .NET uses National Language Support (NLS) or International Components for Unicode (ICU) globalization APIs for Windows apps. .NET 5.0 and later versions use ICU globalization APIs by default on Windows 10.
+- Determines whether .NET uses National Language Support (NLS) or International Components for Unicode (ICU) globalization APIs for Windows apps. .NET 5.0 and later versions use ICU globalization APIs by default on Windows 10 May 2019 Update and later versions.
 - Default: Use ICU globalization APIs (`false`).
 - For more information, see [Globalization APIs use ICU libraries on Windows](../compatibility/3.1-5.0.md#globalization-apis-use-icu-libraries-on-windows).
 
 | | Setting name | Values | Introduced |
-| - | - | - |
+| - | - | - | - |
 | **runtimeconfig.json** | `System.Globalization.UseNls` | `false` - Use ICU globalization APIs<br/>`true` - Use NLS globalization APIs | .NET 5.0 |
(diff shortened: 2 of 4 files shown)


docs/core/compatibility/globalization.md
```diff
@@ -9,11 +9,16 @@ The following breaking changes are documented on this page:
 
 | Breaking change | Version introduced |
 | - | :-: |
+| [Globalization APIs use ICU libraries on Windows](#globalization-apis-use-icu-libraries-on-windows) | 5.0 |
```
safern, line 12: Do we usually mention the `Preview` that it happened?

docs/core/run-time-config/globalization.md
```diff
@@ -76,3 +76,14 @@ Project file:
 | - | - | - |
 | **runtimeconfig.json** | `Switch.System.Globalization.FormatJapaneseFirstYearAsANumber` | `false` - format as "Gannen"<br/>`true` - format as number |
 | **Environment variable** | N/A | N/A |
+
+## NLS
+
+- Determines whether .NET uses National Language Support (NLS) or International Components for Unicode (ICU) globalization APIs for Windows apps. .NET 5.0 and later versions use ICU globalization APIs by default on Windows 10.
```
safern, line 82: Might be worth mentioning that this is on Windows 10 19H1 or later?

docs/core/run-time-config/globalization.md
```diff
@@ -76,3 +76,14 @@ Project file:
...
+- Default: Use ICU globalization APIs (`false`).
+- For more information, see [Globalization APIs use ICU libraries on Windows](../compatibility/3.1-5.0.md#globalization-apis-use-icu-libraries-on-windows).
+
+| | Setting name | Values | Introduced |
```
safern, line 86: It seems like this table has a syntax error:

image

docs/core/compatibility/globalization.md
gewarren: Not usually in this navigation table.

docs/core/run-time-config/globalization.md
gewarren: Oops, thanks.

docs/core/run-time-config/globalization.md
```diff
@@ -76,3 +76,14 @@ Project file:
 | - | - | - |
 | **runtimeconfig.json** | `Switch.System.Globalization.FormatJapaneseFirstYearAsANumber` | `false` - format as "Gannen"<br/>`true` - format as number |
 | **Environment variable** | N/A | N/A |
+
+## NLS
+
+- Determines whether .NET uses National Language Support (NLS) or International Components for Unicode (ICU) globalization APIs for Windows apps. .NET 5.0 and later versions use ICU globalization APIs by default on Windows 10 May 2019 Update and later versions.
+- Default: Use ICU globalization APIs (`false`).
```
safern, line 83: This should instead be: 
```suggestion
- Default: Use NLS globalization APIs (`false`).
```

or 

```suggestion
- Default: Use ICU globalization APIs (`true`).
```

gewarren: This is trying to show the default value if you don't explicitly set it, which is to use ICU, correct?

safern: Right, if the switch is not present, the default behavior is to use ICU.

includes/core-changes/globalization/5.0/icu-globalization-api.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Category
+
+Globalization
+
+### Affected APIs
+
+- <xref:System.Span%601?displayProperty=fullName>
+- <xref:System.String?displayProperty=fullName>
+- <xref:System.Globalization.CompareInfo?displayProperty=fullName>
```
tarekgh, line 32: we shouldn't list different classes inside the globalization namespace, instead, we should say the most of Globalization namespace types will get affected.

includes/core-changes/globalization/5.0/icu-globalization-api.md
```diff
@@ -0,0 +1,46 @@
+### Globalization APIs use ICU libraries on Windows
+
+.NET 5.0 and later versions use [International Components for Unicode (ICU)](http://site.icu-project.org/home) libraries for globalization functionality when running on Windows 10 May 2019 Update or later.
+
+#### Change description
+
+Previously, .NET libraries used [National Language Support (NLS)](/windows/win32/intl/national-language-support) APIs for globalization functionality. For example, NLS functions were used to set the user's locale, sort strings, and format dates and times in the appropriate culture.
```
tarekgh, line 7: > NLS functions were used to set the user's locale, sort strings, and format dates and times in the appropriate culture.

I am not sure if this is clear. it is better to rephrase it to something like:
`NLS functions were used to get culture data (e.g. date and time format patterns, currency symbols,...etc.), compare strings and do string casing in the appropriate culture`

tarekgh (approved):
added a couple of comments. LGTM otherwise. Thanks for getting this written.

safern (approved):
Thanks @gewarren

docs/core/run-time-config/globalization.md
gewarren: So should I leave the default out altogether? Your suggestion is not making sense to me.

safern: Now I understand what this is saying. However the (`false`) inside the parenthesis was confusing.

but this is saying that the default is to use ICU... maybe remove the `false` part?

gewarren: Thanks @safern. I agree that it might be confusing as written. Just to understand better, what did you think the `false` referred to? I'm trying to devise a better way to display defaults for this setting and all the others that currently list it in the same way.

includes/core-changes/globalization/5.0/icu-globalization-api.md
```diff
@@ -0,0 +1,40 @@
...
+
+#### Change description
+
+Previously, .NET libraries used [National Language Support (NLS)](/windows/win32/intl/national-language-support) APIs for globalization functionality. For example, NLS functions were used to get culture data, such as date and time format patterns, compare strings, and perform string casing in the appropriate culture.
+
+Starting in .NET 5.0, if an app is running on Windows 10 May 2019 Update or later, .NET libraries use [ICU](http://site.icu-project.org/home) globalization APIs. Windows 10 May 2019 Update and later versions ship with the ICU native library. If the .NET runtime can't load ICU, it uses NLS instead.
+
+This change was introduced for two reasons:
+
+- Apps have the same globalization behavior across platforms, including Linux, MacOS, and Windows.
```
scottaddie, line 13:
```suggestion
- Apps have the same globalization behavior across platforms, including Linux, macOS, and Windows.
```

includes/core-changes/globalization/5.0/icu-globalization-api.md
```diff
@@ -0,0 +1,40 @@
...
+- Apps can control globalization behavior by using custom ICU libraries.
+
+#### Version introduced
+
+.NET 5.0 Preview 4
+
+### Recommended action
+
+No action is required on the part of the developer. However, if you wish to continue using NLS globalization APIs, you can set a [run-time switch](../../../../docs/core/run-time-config/globalization.md#nls) to revert to that behavior.
```
scottaddie, line 22:
```suggestion
No action is required on the part of the developer. However, if you wish to continue using NLS globalization APIs, you can set a [runtime switch](../../../../docs/core/run-time-config/globalization.md#nls) to revert to that behavior.
```

includes/core-changes/globalization/5.0/icu-globalization-api.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Affected APIs
+
+- <xref:System.Span%601?displayProperty=fullName>
+- <xref:System.String?displayProperty=fullName>
+- Most types in the <xref:System.Globalization?displayProperty=fullName> namespace
+
+<!--
+
+- `T:System.Span%601`
```
scottaddie, line 36:
```suggestion
### Affected APIs

- `T:System.Span%601`
```

docs/core/run-time-config/globalization.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Globalization config settings
 description: Learn about run-time settings that configure globalization aspects of a .NET Core app, for example, how it parses Japanese dates.
-ms.date: 11/27/2019
+ms.date: 05/18/2020
 ms.topic: reference
 ---
 # Run-time configuration options for globalization
```
scottaddie, line 7:
```suggestion
# Runtime configuration options for globalization
```

docs/core/run-time-config/globalization.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Globalization config settings
 description: Learn about run-time settings that configure globalization aspects of a .NET Core app, for example, how it parses Japanese dates.
```
scottaddie, line 3:
```suggestion
description: Learn about runtime settings that configure globalization aspects of a .NET Core app, for example, how it parses Japanese dates.
```

includes/core-changes/globalization/5.0/icu-globalization-api.md
gewarren: I already have the heading on line 28.

includes/core-changes/globalization/5.0/icu-globalization-api.md
gewarren: We had a debate about this on the .NET docs team. Are the switches actually configuring the .NET runtime, or are they configuring behavior at run time. We decided they configure the .NET runtime, but only for your app, not for other apps that might be running (unless you are setting an environment variable). They also configure the run-time behavior of your app. So either way is correct. LMK if you disagree.

docs/core/run-time-config/globalization.md
safern: I understood `false` referred to `Use ICU globalization APIs` — but after your response I was then able to understand it referred to using NLS and that the default was referring to it using ICU. 

Does that make sense?

includes/core-changes/globalization/5.0/icu-globalization-api.md
gewarren: Oh I see, it should be in the commented section too. Thanks!

docs/core/run-time-config/globalization.md
gewarren: Yes, that makes sense, thanks!

PR #18499, opened 2020-05-19
Fix name of dotnet local tools .json file

akoeplinger:
It is called `dotnet-tools.json`, not `tool-manifest.json`


tdykstra (approved):
Thanks!

PR #18546, opened 2020-05-21
Correct that System.Byte has MaxValue field not property/method

Youssef1313:


IEvangelist (approved):
Looks good to me :shipit:

PR #18568, opened 2020-05-22
Make Console Messages Consistent With The Comments

0xNeshi:
Rewrite the child Task.Run Console messages to be consistent with other examples and with messages in the commented output example

## Summary

- Changed "Child starting." to "Attached child starting."
- Changed "Child completing." to "Attached child completing."
- Add a missing dot

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_Misc/tpl_childtasks/cs/child1a.cs
+++ b/samples/snippets/csharp/VS_Snippets_Misc/tpl_childtasks/cs/child1a.cs
@@ -10,9 +10,9 @@ public class Example
       var parent = Task.Run(() => {
             Console.WriteLine("Parent task executing.");
             var child = Task.Factory.StartNew(() => {
-                  Console.WriteLine("Child starting.");
+                  Console.WriteLine("Attached child starting.");
                   Thread.SpinWait(5000000);
-                  Console.WriteLine("Child completing.");
+                  Console.WriteLine("Attached child completing.");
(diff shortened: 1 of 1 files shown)


IEvangelist (approved):
Looks good to me :shipit:

samples/snippets/csharp/VS_Snippets_Misc/tpl_childtasks/cs/child1a.cs
```diff
@@ -10,17 +10,17 @@ public static void Main()
       var parent = Task.Run(() => {
             Console.WriteLine("Parent task executing.");
             var child = Task.Factory.StartNew(() => {
-                  Console.WriteLine("Child starting.");
+                  Console.WriteLine("Attached child starting.");
                   Thread.SpinWait(5000000);
-                  Console.WriteLine("Child completing.");
+                  Console.WriteLine("Attached child completing.");
```
Youssef1313, line 15: I think "completed" sounds more correct than "completing"

```suggestion
                  Console.WriteLine("Attached child completed.");
```

BillWagner: I would prefer leaving this unchanged. The `WriteLine` call is inside the child thread. For that reason, it hasn't completed when this is written. (If it were written in the parent thread, "completed" would be better.

BillWagner (approved):
I'll also approve this. I'll :shipit: now.

PR #18578, opened 2020-05-22
Change slash to dash

Youssef1313:
See #7732


BillWagner (approved):
Thanks again @Youssef1313 

I'll :shipit: now.

PR #18579, opened 2020-05-22
Add small tip about Main args (can't be null)

Youssef1313:
I think there is no cases where args could be null. So, I added a small tip for readers who may think that accessing `Length` property could throw a NullReferenceException.


BillWagner (approved):
I verified this with the C# standard. If `args` is specified to `Main`, the runtime will always pass an array. The array may be empty, but it is never `null`.

Thanks again @Youssef1313 I'll :shipit: now.

PR #18599, opened 2020-05-23
Fix snippet to match docs

Youssef1313:
This reverts one bad change in https://github.com/dotnet/samples/pull/1479


BillWagner (approved):
Thanks for fixing this @Youssef1313 

I'll :shipit: now.

PR #18600, opened 2020-05-23
Delete two unused files

Youssef1313:


BillWagner (approved):
The build completed without warnings, so I'll :shipit: 

Thanks @Youssef1313

Issue #18607, opened 2020-05-24
Change "ID" to "Id"


robloo:
Identifier should be abbreviated "Id" instead of "ID" following Microsoft's naming guidelines. 
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 3d85f4d3-fbb1-2aba-9acb-a1b5122bc484
* Version Independent ID: 19163540-1a7d-2524-370a-8b20d5735d90
* Content: [Auto-Implemented Properties - C# Programming Guide](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/auto-implemented-properties)
* Content Source: [docs/csharp/programming-guide/classes-and-structs/auto-implemented-properties.md](https://github.com/dotnet/docs/blob/master/docs/csharp/programming-guide/classes-and-structs/auto-implemented-properties.md)
* Product: **dotnet-csharp**
* Technology: **csharp-fundamentals**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
This was fixed in #18610.  

While I agree with the comment that "there may be other occurrences", a quick search didn't turn any up.

PR #18610, opened 2020-05-24
Match naming conventions (ID->Id)

Youssef1313:
Contributes to #18607

(Making it contributes instead of Fixes because there could be other snippets too with the same issue)

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_VBCSharp/csProgGuideLINQ/CS/csRef30LangFeatures_2.cs
+++ b/samples/snippets/csharp/VS_Snippets_VBCSharp/csProgGuideLINQ/CS/csRef30LangFeatures_2.cs
@@ -18,11 +18,11 @@ namespace AutoImplMutable
         public int CustomerId { get; set; }
 
         // Constructor
-        public Customer(double purchases, string name, int ID)
+        public Customer(double purchases, string name, int id)
         {
             TotalPurchases = purchases;
             Name = name;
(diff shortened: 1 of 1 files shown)


samples/snippets/csharp/VS_Snippets_VBCSharp/csProgGuideLINQ/CS/csRef30LangFeatures_2.cs
```diff
@@ -15,14 +15,14 @@ class Customer
         // Auto-implemented properties for trivial get and set
         public double TotalPurchases { get; set; }
         public string Name { get; set; }
-        public int CustomerID { get; set; }
+        public int CustomerId { get; set; }
 
         // Constructor
         public Customer(double purchases, string name, int ID)
```
NextTurn, line 21: parameter name

nxtn (changes requested):

IEvangelist (approved):
Looks good to me :shipit:

samples/snippets/csharp/VS_Snippets_VBCSharp/csProgGuideLINQ/CS/csRef30LangFeatures_2.cs
IEvangelist: I agree with @NextTurn - @Youssef1313 are you able to update the parameter name `ID` as well?

PR #18625, opened 2020-05-25
fix typo dotnet restore --use-lock-file (instead of --use-lockfile)

pirluq:
## Summary

There was a typo in the dotnet restore documentation
--use-lock-file (instead of --use-lockfile)


tdykstra (approved):
Thanks!

PR #18626, opened 2020-05-25
Using "effect" instead "affect"

AlexZazu:
I believe these two words have similar meaning, but I believe using effect in these two paragraphs would be more correct. Sorry if I made a mistake (I am Spanish speaker).

## Summary

Describe your changes here.


IEvangelist (approved):
Good catch, you're correct!

Looks good to me :shipit:

PR #18655, opened 2020-05-27
Remove moved snippets from CODEOWNERS

Youssef1313:
These directories doesn't exist now.


IEvangelist (approved):
Looks good to me :shipit:

PR #18678, opened 2020-05-28
fix: MD034/no-bare-urls

nschonni:
## Summary

Follow on PR from the last MD034 that included the auto clean-ups. This enables the rule, but does affect rendering on the markdown:
- Change from emphasis sections for commandline examples to code sections, and remove escaping
- Code fence OData examples with spaces
- Few quoted URLs that are flagged by v0.22, but not the latest v0.23.1, so bumped CI


IEvangelist (approved):

PR #18691, opened 2020-05-29
.NET Framework -> .NET

gewarren:
I only looked through the csharp/programming-guide directory as I was looking up XML doc comments.

Contributes to #18697.


BillWagner (approved):
This looks great @gewarren 

Its ready to :shipit:

Should we make a similar sweep through the other C# sections, and other .NET sections like the VB guide?

PR #18696, opened 2020-05-29
Implement the async dispose pattern

### LINKED ISSUE #18623: Guidance for implementing IAsyncDisposable
It would be great if this documentation mentioned implementing [`IAsyncDisposable`](https://docs.microsoft.com/en-us/dotnet/api/system.iasyncdisposable) for asynchronous disposal and provided some guidance for how to extend the disposal pattern to include asynchronous disposal.

Related:
* https://stackoverflow.com/q/58112716
* dotnet/roslyn-analyzers#3675

---
#### Document Details

* ID: 572ef6c6-5066-4a14-76a0-c3fb4b62e604
* Version Independent ID: 9c754da1-c234-fb12-580b-0fcf9ffe2c20
* Content: [Implement a Dispose method](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose)
* Content Source: [docs/standard/garbage-collection/implementing-dispose.md](https://github.com/dotnet/docs/blob/master/docs/standard/garbage-collection/implementing-dispose.md)
* Product: **dotnet**
* GitHub Login: @gewarren

IEvangelist:
## Summary

- Ran the collapse relative links in current working directory
- New article with guidance on implementing the async dispose pattern
- Remove self-referencing link in *implement-dispose.md* 🤦‍♂️ 
- Correct return type of `DisposeAsync` callout, and link to new article
- Use `Utf8JsonWriter` as example in cascading async dispose, and link to new migration guide
- Add TOC entry, and leverage `displayName` for searching "await using", etc.
- Provide clarity around stacked async usings in errant conditions

Fixes #18623

### Preview

✔️ [Implement a DisposeAsync method](https://review.docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync?branch=pr-en-us-18696)

### DIFF
--- a/docs/standard/garbage-collection/implementing-disposeasync.md
+++ b/docs/standard/garbage-collection/implementing-disposeasync.md
@@ -12,13 +12,13 @@ helpviewer_keywords:
 
 # Implement a DisposeAsync method
 
-The <xref:System.IAsyncDisposable> interface was introduced as part of C# 8.0. You implement the <xref:System.IAsyncDisposable.DisposeAsync?displayProperty=nameWithType> method when you need to perform resource cleanup, just as you would when [implementing a Dispose method](implementing-dispose.md). One of the key differences however, is that this implementation allows for asynchronous cleanup operations. The <xref:System.IAsyncDisposable.DisposeAsync> returns a <xref:System.Threading.Tasks.ValueTask> that represents the asynchronous dispose operation.
+The <xref:System.IAsyncDisposable?displayProperty=nameWithType> interface was introduced as part of C# 8.0. You implement the <xref:System.IAsyncDisposable.DisposeAsync?displayProperty=nameWithType> method when you need to perform resource cleanup, just as you would when [implementing a Dispose method](implementing-dispose.md). One of the key differences however, is that this implementation allows for asynchronous cleanup operations. The <xref:System.IAsyncDisposable.DisposeAsync> returns a <xref:System.Threading.Tasks.ValueTask> that represents the asynchronous dispose operation.
 
-It is typical that when implementing the <xref:System.IAsyncDisposable> interface, classes will also implement the <xref:System.IDisposable> interface. All of the guidance for implementing the dispose pattern is also relevant to the asynchronous implementation. This article assumes that you're already familiar with how to [implement a Dispose method](implementing-dispose.md).
+It is typical that when implementing the <xref:System.IAsyncDisposable> interface, classes will also implement the <xref:System.IDisposable> interface. All of the guidance for implementing the dispose pattern applies to the asynchronous implementation. This article assumes that you're already familiar with how to [implement a Dispose method](implementing-dispose.md).
 
 ## DisposeAsync() and DisposeAsyncCore()
 
-The <xref:System.IAsyncDisposable> interface requires the implementation of a single parameterless method, <xref:System.IAsyncDisposable.DisposeAsync>. Also, any non-sealed class should have an additional `DisposeAsyncCore()` method that also returns a <xref:System.Threading.Tasks.ValueTask>.
+The <xref:System.IAsyncDisposable> interface declares a single parameterless method, <xref:System.IAsyncDisposable.DisposeAsync>. Any non-sealed class should have an additional `DisposeAsyncCore()` method that also returns a <xref:System.Threading.Tasks.ValueTask>.
 
 - A `public` <xref:System.IAsyncDisposable.DisposeAsync?displayProperty=nameWithType> implementation that has no parameters.
 - A `protected virtual ValueTask DisposeAsyncCore()` method whose signature is:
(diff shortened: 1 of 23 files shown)


samples/snippets/csharp/VS_Snippets_CLR/conceptual.asyncdisposable/disposeasync.cs
```diff
@@ -0,0 +1,53 @@
...
+        Dispose(false);
+        GC.SuppressFinalize(this);
+    }
+
+    protected virtual async ValueTask DisposeAsyncCore()
+    {
+        // Cascade async dispose calls
+        if (_jsonWriter != null)
+        {
+            await _jsonWriter.DisposeAsync();
```
BillWagner, line 26: From my later comment, should this set `_jsonWriter` to null after disposing asynchronously?

```suggestion
            await _jsonWriter.DisposeAsync();
            _jsonWriter = null;
```

samples/snippets/csharp/VS_Snippets_CLR/conceptual.asyncdisposable/disposeasync.cs
```diff
@@ -0,0 +1,53 @@
...
+    protected virtual void Dispose(bool disposing)
+    {
+        if (_disposed)
+        {
+            return;
+        }
+
+        if (disposing)
+        {
+            // TODO: dispose managed state (managed objects).
```
BillWagner, line 45: I would think that `_jsonWriter` would need to be synchronously disposed of here, correct?

```suggestion
            _jsonWriter?.Dispose();
            // TODO: dispose managed state (managed objects).
```

docs/standard/garbage-collection/implementing-disposeasync.md
```diff
@@ -0,0 +1,89 @@
...
+    await DisposeAsyncCore();
+
+    // Dispose of managed resources.
+    Dispose(false);
+    // Suppress finalization.
+    GC.SuppressFinalize(this);
+}
+```
+
+> [!NOTE]
```
BillWagner, line 51: This paragraph was confusing to me. It took several reads to get to my current understanding (which may be wrong). Here's what I think you're trying to say. It matches my code comments below:

1. If a type implements `IAsyncDisposable`, it likely also implements `IDisposable`. 
1. `DisposeAsyncCore` disposes managed resources *asynchronously*, so you don't want to dispose of them *synchronously* as well. Therefore, call `Dispose(false)` instead of `Dispose(true)`.
1. The developers that use your type may dispose of it synchronously or asynchronously. Be prepared for either. (See my comments on the code)

I may be mis-understanding this, but that's what I got from this.

BillWagner (changes requested):
This is great @IEvangelist 

I had some comments, and I'd like to see how they are resolved to improve my understanding of this.

docs/standard/garbage-collection/implementing-disposeasync.md
IEvangelist: These are all correct statements. I'll try to rephrase this more appropriately.

BillWagner (approved):

BillWagner:
Those additional changes look great. You can :shipit: when ready.

sharwell:
@IEvangelist The name `DisposeAsyncCore()` conflicts with the async naming suffix. Since it appears to provide little to no value over making `DisposeAsync()` virtual, my recommendation is to remove it outright. An alternative would be renaming the method to `DisposeCoreAsync()`.

IEvangelist:
Hi @sharwell - That is a good catch, and I do agree with the naming, however; I'd like to ensure that @bartonjs is okay with this - as he's the one who originally proposed this. Jeremy, what do you think?

sharwell:
@bartonjs  Looking at https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#the-disposeasynccore-method, it seems even more clear that `DisposeAsyncCore()` isn't necessary. Everything the article claims `DisposeAsyncCore()` provides can be done with less code (and less complexity) by `DisposeAsync()` itself.

bartonjs:
Meta point: This is documentation supporting a published guideline, and it is correct as-is:

> DO implement `DisposeAsync` as only calling and awaiting `DisposeAsyncCore()`, calling `Dispose(false)`, and calling `GC.SuppressFinalize(this)`, in that order—except on sealed types.

(Framework Design Guidelines, 3rd edition, pg. 392)

> The name DisposeAsyncCore() conflicts with the async naming suffix.

No, it doesn't.  The Async guidance is to name a method with an Async suffix, and the Template Method Pattern guidance is to name a method with the Core suffix.  Since Template Method Pattern was applied later, "AsyncCore" is the correct combined suffix.

I'm pretty sure that there's an explanation like that somewhere in the book, but the best I'm finding is:

> DO use the suffix “Async” for methods returning IAsyncEnumerable<T>.
>
> Methods that return IAsyncEnumerable<T> follow the same naming
conventions as methods returning Tasks with the Task-Based Async
Pattern, including using the combined suffix “AsyncCore” when making
use of the Template Method Pattern (section 9.9).

> Since it appears to provide little to no value over making DisposeAsync() virtual, my recommendation is to remove it outright.

It does: It provides a place to call GC.SuppressFinalize(this), which is part of the official guidance.

PR #18724, opened 2020-06-01
Collapse relative links in visual-basic/

nxtn:
Contributes to #10653


IEvangelist (approved):
Looks good to me, we'll :shipit: - thank you 🙏.

The tooling is really making this a much quicker experience 🎉

NextTurn:
Seriously speaking, that tool runs really slowly and becomes slower as more editors are open (1 second per document when 1K editors are open). 🤣

IEvangelist:
Yeah, we’ll have to change that functionality – ideally it would work silently, using the native NodeJS file system APIs instead. Thanks for letting me know.

Issue #18611, opened 2020-05-24
Duplicate paragraph in Architecting-Cloud-Native-NET-Apps-Azure


amlaminaar:
There are repeated content at several place in this book e.g. at Page 37
"Containers are immutable. Once you define a container, you can recreate and run it exactly the same way. This immutability lends itself to component-based design. If some parts of an application evolve differently than others, why redeploy the entire app when you can just deploy the parts that change most frequently? Different features and cross-cutting concerns of an app can be broken up into separate units. Figure 3-2 shows how a monolithic app can take advantage of containers and microservices by delegating certain features or functionality. The remaining functionality in the app itself has also been containerized."

and Page 38 /39 
"Containerized services require automated management. It wouldn’t be feasible to manually administer a large set of independently deployed containers. For example, consider the following tasks:
• How will container instances be provisioned across a cluster of many machines?
• Once deployed, how will containers discover and communicate with each other?
• How can containers scale in or out on-demand?
• How do you monitor the health of each container?
• How do you protect a container against hardware and software failures?
• How do upgrade containers for a live application with zero downtime?
Container orchestrators address and automate these and other concerns."

and Page 46
You can also add Container Orchestration Support, also shown in Figure 3-8. By default, the orchestrator uses Kubernetes and Helm. Once you’ve chosen the orchestrator, a azds.yaml file is added to the project root and a charts folder is added containing the Helm charts used to configure and deploy the application to Kubernetes. Figure 3-9 shows the resulting files in a new project.

I would request team to ensure that there is no duplicate paragraph throughout ebook. 

(Edited by gewarren to add document details.)

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 89573850-fe00-e1a3-de51-57394b3ac131
* Version Independent ID: e5b1affd-a12a-bacb-f510-9abb242be819
* Content: [Architecting Cloud Native .NET Applications for Azure](https://docs.microsoft.com/en-us/dotnet/architecture/cloud-native/)
* Content Source: [docs/architecture/cloud-native/index.md](https://github.com/dotnet/docs/blob/master/docs/architecture/cloud-native/index.md)
* Product: **dotnet-architecture**
* Technology: **cloud-native**
* GitHub Login: @ardalis
* Microsoft Alias: **nanil**

IEvangelist:
[...]

PR #18725, opened 2020-06-01
Fixed duplicate section 18718

robvet:
Fixed duplicate section for container/orchestrator section. 

Part of #18611


IEvangelist (approved):

PR #18741, opened 2020-06-02
Fixed typo on app settings overview

alfredmyers:
Fixed typo on app settings overview


Thraka (approved):
Thanks for spotting and fixing this!

PR #18777, opened 2020-06-03
Fix enum value casing

MihaZupan:
Fixes the casings to what the validation scheme in VS is happy with.

Actual config parsing code is case-insensitive, this just gets rid of the warnings when users copy-paste the config.


karelz (approved):
Thanks @MihaZupan!

Thraka (approved):

PR #18800, opened 2020-06-04
Parsing datetime

### LINKED ISSUE #18399: Add how-to/conversions samples to sample browser
Currently, the how-to/conversions samples for C# and VB are zipped in the zip file is referenced in an article:

image

This isn't good because changes to made to the sample may be forgotten to be zipped again. That can cause the zip file to be out-of-sync with the actual sample.

It's better to add it the sample browser and delete the zip file.

IEvangelist:
## Summary

- To be clear, nothing has been added to the samples browser. It's not worth duplicating this code, as it is a small amount and then causes duplicated code - which makes the problem even larger. Instead, removed zip files and references to them.
- Code clean up
- Doc clean up
- Made all C# bits interactive
- Corrected commented output to match actual output

Fixes #18399


samples/snippets/visualbasic/how-to/conversions/Program.vb
```diff
@@ -14,7 +14,7 @@ Module Program
         Dim MyDateTime As DateTime = DateTime.Parse(MyString)
         Console.WriteLine(MyDateTime)
         ' Displays the following output on a system whose culture is en-US:
-        '       1/1/2009 12:00:00 AM
+        '       1/1/2009 00:00:00
         ' </Snippet1>
```
Youssef1313, line 18: I think (but haven't tested yet) that the output depends on the OS time settings. It can appear in different formats.
Maybe say "Displays output similar to the following:" ?

IEvangelist: I was basing this on the interactive experience. We can assume that developers will use the interactive C# in the browser... so I want that to at least be consistent.

Youssef1313: > We can assume that developers will use the interactive C# in the browser.

VB doesn't have interactive experience + I believe I met the "displays output similar to" pattern in many samples before.

Or maybe avoid all that by editing the sample code itself - if appropriate after reviewing the article - to have a consistent output?

Thraka (commented):
We shouldn't link to the snippets here in the articles as something people use for reference. Really that is what the samples and samples browser is for. However, with the fact that the article was already doing this, I think it's OK to keep doing it.

Because of this, the fact that you're not migrating to the samples browser, you shouldn't have "Fixes 12345" in the comment as it will close the issue. Use "Contributes to 12345" instead.

IEvangelist:
> We shouldn't link to the snippets here in the articles as something people use for reference. Really that is what the samples and samples browser is for. However, with the fact that the article was already doing this, I think it's OK to keep doing it.
> 
> Because of this, the fact that you're not migrating to the samples browser, you shouldn't have "Fixes 12345" in the comment as it will close the issue. Use "Contributes to 12345" instead.

I updated the description. My point is that I didn't think that these samples were worth having in the browser and that the doc itself was enough, as the code was such a trivial amount.

docs/csharp/how-to/concatenate-multiple-strings.md
```diff
@@ -49,7 +49,7 @@ combines an array of words by adding a space between each word in the array:
 
 [!code-csharp-interactive[string concatenation using LINQ expressions](../../../samples/snippets/csharp/how-to/strings/Concatenate.cs#6)]  
 
-You can try these samples by looking at the [sample code](https://github.com/dotnet/docs/tree/master/samples/snippets/csharp/how-to/strings). Or, you can download the samples [as a zip file](../../../samples/snippets/csharp/how-to/strings.zip).
+You can try these samples by looking at the [sample code](https://github.com/dotnet/docs/tree/master/samples/snippets/csharp/how-to/strings).
```
BillWagner, line 52: I recommend removing this paragraph from all these articles because we are not adding this project to the samples browser.

/cc @Thraka for his thoughts.

docs/standard/base-types/parsing-datetime.md
```diff
@@ -50,38 +50,38 @@ The following example illustrates the use of the <xref:System.DateTime.Parse%2A?
...
-  
-[!code-csharp[Parsing.DateAndTime#2](../../../samples/snippets/csharp/how-to/conversions/StringToDateTime.cs#2)]
+You can also explicitly define the culture whose formatting conventions are used when you parse a string. You specify one of the standard <xref:System.Globalization.DateTimeFormatInfo> objects returned by the <xref:System.Globalization.CultureInfo.DateTimeFormat%2A?displayProperty=nameWithType> property. The following example uses a format provider to parse a German string into a <xref:System.DateTime>. It creates a <xref:System.Globalization.CultureInfo> representing the `de-DE` culture. That `CultureInfo` object ensures successful parsing of this particular string. This precludes whatever setting is in the <xref:System.Threading.Thread.CurrentCulture> of the <xref:System.Threading.Thread.CurrentThread>.
+
+[!code-csharp-interactive[Parsing.DateAndTime#2](../../../samples/snippets/csharp/how-to/conversions/StringToDateTime.cs#2)]
```
BillWagner, line 60: consider using the new syntax, as long as you're making significant changes in these articles.

BillWagner (approved):
This LGTM @IEvangelist 

I had a couple thoughts, and then it's ready to :shipit:

docs/csharp/how-to/compare-strings.md
```diff
@@ -36,7 +36,7 @@ By default, the most common operations:
 
 perform a case-sensitive ordinal comparison and, if necessary, use the current culture. The following example demonstrates that:
 
-[!code-csharp-interactive[Comparing strings using an ordinal comparison](../../../samples/snippets/csharp/how-to/strings/CompareStrings.cs#1)]
+:::code language="csharp-interactive" source="../../../samples/snippets/csharp/how-to/strings/CompareStrings.cs" id="1":::
```
Youssef1313, line 39: I think the id should be `Snippet1` not just `1` ? The build hasn't finished yet, so I'm not sure.

I also think, based on [this](https://docs.microsoft.com/en-us/contribute/code-in-docs#code-snippets-included-by-reference), that the language should be csharp only and set the interactive attribute?

docs/csharp/how-to/concatenate-multiple-strings.md
```diff
@@ -14,46 +14,44 @@ ms.assetid: 8e16736f-4096-4f3f-be0f-9d4c3ff63520
...
-  
-[!code-csharp-interactive[string concatenation using string builder](../../../samples/snippets/csharp/how-to/strings/Concatenate.cs#4)]  
+In other cases, you may be combining strings in a loop where you don't know how many source strings you're combining, and the actual number of source strings may be large. The <xref:System.Text.StringBuilder> class was designed for these scenarios. The following code uses the <xref:System.Text.StringBuilder.Append%2A> method of the <xref:System.Text.StringBuilder> class to concatenate strings.
 
-You can read more about the [reasons to choose string concatenation or the `StringBuilder` class](xref:System.Text.StringBuilder#StringAndSB).
+:::code language="csharp-interactive" source="../../../samples/snippets/csharp/how-to/strings/Concatenate.cs" id="4":::
+
+You can read more about the [reasons to choose string concatenation or the `StringBuilder` class](xref:System.Text.StringBuilder%23StringAndSB).
```
Youssef1313, line 38:
```suggestion
You can read more about the [reasons to choose string concatenation or the `StringBuilder` class](xref:System.Text.StringBuilder#StringAndSB).
```

docs/csharp/how-to/compare-strings.md
Thraka: `Snippet` is actually optional when referring to a snippet. No matter what ID you create/use the word `Snippet` is always removed before it looks for a snippet.

Youssef1313: There was build errors caused by omitting the "Snippet" part (for the new syntax only) in previous PRs and was only solved by including the "Snippet" part. Maybe it was a bug that got fixed.

Thraka:
> > We shouldn't link to the snippets here in the articles as something people use for reference. Really that is what the samples and samples browser is for. However, with the fact that the article was already doing this, I think it's OK to keep doing it.
> > Because of this, the fact that you're not migrating to the samples browser, you shouldn't have "Fixes 12345" in the comment as it will close the issue. Use "Contributes to 12345" instead.
> 
> I updated the description. My point is that I didn't think that these samples were worth having in the browser and that the doc itself was enough, as the code was such a trivial amount.

@IEvangelist 

I didn't get to your response until now. Given that, I think you should mark this as fixes that issue and close it 😄 If we're deleting the zip because the article can cover it all, and we're not moving it to samples, the original issue is completed. Mark it as fixed by this issue so that there will be an appropriate link and then manually close the other issue.

Good job!

PR #18809, opened 2020-06-05
Fix anchors to headers in switch.md

WindOfMind:
## Summary

Currently, `Pattern matching with the switch statement` and `The case statement and the when clause` paragraphs are displayed in a browser (Chrome) as a link (when you put a cursor on it the browser highlights the whole text as a link). I propose to use valid anchors links to headers in md docs.


BillWagner (approved):
Thanks for submitting this fix @WindOfMind We appreciate it.

I've reviewed the changes, and I'll :shipit:

Thanks again.

PR #18829, opened 2020-06-06
Collapse relative links in framework/wcf/

nxtn:
Contributes to #10653


IEvangelist (approved):

Not sure if you saw this or have had a chance to use it yet, but I updated the Docs Authoring Pack to silently update files - when there is more than 20 files to change. https://github.com/microsoft/vscode-docs-authoring/pull/619

NextTurn:
I'd guess I won't have a chance to try it because we've finally finished #10653. ;)

IEvangelist:
> I'd guess I won't have a chance to try it because we've finally finished #10653. ;)

Just in time I suppose... perhaps, I'll run it at the top of the directory -- to see if I catch any others.

PR #18832, opened 2020-06-06
Fix wrongly indented list item

Youssef1313:
With the indentation, it shows as a nested item while it shouldn't


tdykstra (approved):
Thanks!

PR #18837, opened 2020-06-07
Format codes based on .editorconfig, Part 1

nxtn:


IEvangelist (approved):

PR #18838, opened 2020-06-07
Format codes based on .editorconfig, Part 2

nxtn:
Whitespace-only changes.


IEvangelist (approved):

PR #18868, opened 2020-06-09
Format codes based on .editorconfig, Part 5

nxtn:


BillWagner (approved):
Approving as all whitespace.

I'll :shipit: now.

Thanks again @NextTurn

PR #18869, opened 2020-06-09
Format codes based on .editorconfig, Part 6

nxtn:


samples/snippets/visualbasic/VS_Snippets_CLR/CustomAttributeData/VB/source.vb
```diff
@@ -76,11 +76,11 @@ End Class
...
 <Example(ExampleKind.SecondKind, _
-         New String() { "String array argument, line 1", _
+         New String() {"String array argument, line 1", _
                         "String array argument, line 2", _
-                        "String array argument, line 3" }, _
-         Note := "This is a note on the class.", _
-         Numbers := New Integer() { 53, 57, 59 })> _
+                        "String array argument, line 3"}, _
+         Note:="This is a note on the class.", _
+         Numbers:=New Integer() {53, 57, 59})> _
```
Thraka, line 83: This one seems like an odd rule. why remove spacing around the operator?

Thraka (approved):

samples/snippets/visualbasic/VS_Snippets_CLR/CustomAttributeData/VB/source.vb
NextTurn: This is indeed a bit odd, but it's VS' default rule as well.

BillWagner: @NextTurn Copying that .editorconfig to this repo would be a great change. (We should also copy it to the dotnet/dotnet-api-docs repository). 

I'll merge this now, and the addition of editorconfig can be another PR.

PR #18871, opened 2020-06-09
Remove linux from sdk/runtime pages

Thraka:
A future PR will redesign the install SDK/Runtime pages. For now, the bash automation section was leaking into the linux pivot. This has been removed. Now the linux pivot only points them to the new article.


IEvangelist (approved):

PR #18905, opened 2020-06-11
Format codes based on .editorconfig, Part 9

nxtn:
Whitespace-only changes.


BillWagner (approved):
Thanks again @NextTurn 

I'll :shipit: now.

PR #18927, opened 2020-06-12
Format codes based on .editorconfig, Part 10

nxtn:
Whitespace-only changes.


BillWagner (approved):
Shipping based on whitespace only diff.

PR #18972, opened 2020-06-16
Wording

bb-froggy:
The Pattern is implemented, but the linked page provides the overview.


IEvangelist (approved):

PR #18985, opened 2020-06-17
Format codes based on .editorconfig, Part 16

nxtn:


BillWagner (approved):
Thanks again @NextTurn 

I reviewed the changes as all whitespace, and I'll :shipit:

PR #18989, opened 2020-06-17
Corrections to the lists of Linux package dependencies

mthalman:
The current set of Linux package dependencies across the supported set of distros have some inaccuracies.  There are cases of packages that aren't listed, packages that shouldn't be listed, and updates to package versions that are needed.

cc @Thraka, @dagood, @MichaelSimons

### DIFF
--- a/docs/core/install/dependencies.md
+++ b/docs/core/install/dependencies.md
@@ -241,16 +241,17 @@ Based on your linux distribution, you may need to install additional dependencie
 
 Ubuntu distributions require the following libraries to be installed:
 
-- liblttng-ust0
-- libcurl3 (for 14.x and 16.x)
-- libcurl4 (for 18.x)
-- libssl1.0.0
-- libkrb5-3
-- zlib1g
+- libc6
+- libgcc1
+- libgssapi-krb5-2
 - libicu52 (for 14.x)
 - libicu55 (for 16.x)
-- libicu57 (for 17.x)
 - libicu60 (for 18.x)
+- libicu66 (for 20.x)
+- libssl1.0.0 (for 14.x, 16.x)
+- libssl1.1 (for 18.x, 20.x)
+- libstdc++6
+- zlib1g
 
 For .NET Core apps that use the *System.Drawing.Common* assembly, you also need the following dependency:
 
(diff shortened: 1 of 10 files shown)


dagood:
+cc @NikolaMilosavljevic

Thraka:
I was just having a conversation about this with @leecow 

I just need the lists corrected by someone with the knowledge and I can get it updated.

docs/core/install/dependencies.md
```diff
@@ -241,16 +241,17 @@ Based on your linux distribution, you may need to install additional dependencie
 
 Ubuntu distributions require the following libraries to be installed:
 
-- liblttng-ust0
-- libcurl3 (for 14.x and 16.x)
-- libcurl4 (for 18.x)
-- libssl1.0.0
-- libkrb5-3
-- zlib1g
+- libc6
```
leecow, line 244: Where did you get the requirements for libc6 and libgcc1? I have not seen a need to install these independently for .NET scenarios.

docs/core/install/linux-debian.md
```diff
@@ -122,7 +122,7 @@ sudo apt-get update; \
 
 ## Dependencies
 
-[!INCLUDE [linux-install-dependencies](includes/linux-install-dependencies.md)]
+[!INCLUDE [linux-rpm-install-dependencies](includes/linux-rpm-install-dependencies.md)]
```
leecow, line 125: The dependency list for Debian is not the same as what is listed in the rpm document. 

Debian 10: libicu63, libssl1.1
Debian 9: libicu57, libssl1.1

leecow (changes requested):
Left a few comments for changes and a question.

docs/core/install/dependencies.md
mthalman: From here:
* https://github.com/dotnet/runtime/blob/a755e742cad93fb173b47a95f68168c29ae95fb6/src/installer/pkg/packaging/deb/dotnet-runtime-deps-debian_config.json#L32
* https://github.com/dotnet/dotnet-docker/blob/75e441fdb26bdffa5f58437cd77048586016ec96/src/runtime-deps/3.1/bionic/amd64/Dockerfile#L8

This was before my time so I don't know the history.  Perhaps @dagood or @MichaelSimons has more context.

dagood: As far as Git can tell me, these originated in https://github.com/dotnet/core-setup/pull/3317.

docs/core/install/linux-debian.md
mthalman: Fixed

leecow (approved):
lgtm

PR #19003, opened 2020-06-18
Format codes based on .editorconfig, Part 18

nxtn:


BillWagner (approved):
Merging as whitespace only changes.

Thanks again @NextTurn

PR #19004, opened 2020-06-18
Format codes based on .editorconfig, Part 19

nxtn:


BillWagner (approved):
Also merging as whitespace only changes.

Thanks again @NextTurn

PR #19005, opened 2020-06-18
Format codes based on .editorconfig, Part 20

nxtn:


BillWagner (approved):
Thanks again. 

I'll :shipit: as whitespace only changes.

PR #19008, opened 2020-06-18
Use "Attribute" suffix in example defining a new attribute

kevinmkane:
## Summary

The example to create a custom attribute in C# creates an attribute called `Author` and defines it using a class called `Author`. Although this works, it causes warning CA1710, which encourages using preferred suffixes for types like this. This changes the name of the class to `AuthorAttribute` to follow this guidance.


BillWagner (approved):
Thanks for making this change @kevinmkane 

I appreciate the clear description in the PR, that made review easy. Thank you! After reviewing, I'll approve this and :shipit:

kevinmkane:
You're welcome, @BillWagner. Thanks for the fast review.

PR #19198, opened 2020-06-24
Clean error reporting for pr code checks

adegeo:
- Only report a single error line for an error instead of 3
- Checks for a `(line,col)` in the error message and reports that in a way GitHub can associate with the file.


BillWagner (approved):

PR #19201, opened 2020-06-24
Choosing between anonymous and tuple types

### LINKED ISSUE #1575: C# Conceptual: Tuples vs. ValueTuples vs. Anonymous Types vs. Struct vs Class
There are several different syntax elements developers can use to create types. An important question for developers is when to choose which syntax.

This can be especially difficult with syntax elements that create very similar types (like anonymous types and Value Tuples).

See the comment on [this topic](https://docs.microsoft.com/en-us/dotnet/articles/csharp/csharp-7#tuples) for a direct customer question.

IEvangelist:
## Summary

Added new article explaining some of the key differences between anonymous types, `System.Tuple`, and `System.ValueTuple` types.

### Preview

✔️ [Choosing between anonymous and tuple types](https://review.docs.microsoft.com/en-us/dotnet/standard/design-guidelines/choosing-between-anonymous-and-tuple?branch=pr-en-us-19201)

Fixes #1575

### DIFF
--- a/docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
+++ b/docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
@@ -7,7 +7,7 @@ ms.technology: dotnet-standard
 
 # Choosing between anonymous and tuple types
 
-As a developer, choosing the appropriate type involves considering a types usability, performance, and tradeoffs compared to other types. Anonymous types have been available since C# 3.0, while generic <xref:System.Tuple%602?displayProperty=nameWithType> types were introduced with .NET Framework 4.0. Since then new options have been introduced with language level support, such as <xref:System.ValueTuple%602?displayProperty=nameWithType> - which as the name implies, provide a value type with the flexibility of anonymous types. In this article, you'll learn when it's appropriate to choose one type over the other.
+Choosing the appropriate type involves considering its usability, performance, and tradeoffs compared to other types. Anonymous types have been available since C# 3.0, while generic <xref:System.Tuple%602?displayProperty=nameWithType> types were introduced with .NET Framework 4.0. Since then new options have been introduced with language level support, such as <xref:System.ValueTuple%602?displayProperty=nameWithType> - which as the name implies, provide a value type with the flexibility of anonymous types. In this article, you'll learn when it's appropriate to choose one type over the other.
 
 ## Usability and functionality
 
@@ -106,11 +106,11 @@ You might want to always use <xref:System.ValueTuple> over <xref:System.Tuple>, 
 
 ### Serialization
 
-One important consideration when choosing a type, is whether or not it will need to be serialized. Serialization is the process of converting the state of an object into a form that can be persisted or transported. For more information, see [serialization](../../csharp/programming-guide/concepts/serialization/index.md). With anonymous types, and tuple types it is not recommended to use them if you need to serialization. While it is possible to use them with serialization, it is not fully supported, and requires workarounds.
+One important consideration when choosing a type, is whether or not it will need to be serialized. Serialization is the process of converting the state of an object into a form that can be persisted or transported. For more information, see [serialization](../../csharp/programming-guide/concepts/serialization/index.md). When serialization is important, creating a `class` or `struct` is preferred over anonymous types or tuple types.
 
 ## Performance
 
-Performance is always a topic of discussion when referring to value types vs reference types, often leading to conversations of stack versus heap allocations. However, performance between these types is fairly similar with each other - in other words, performance differences are negligible. The micro-optimizations you might gain from choosing tuples over anonymous types is more often than not irrelevant.
+Performance between these types depends on the scenario. The major impact involves the tradeoff between allocations and copying. In most scenarios, the impact is small. When major impacts could arise, measurements should be taken to inform the decision.
 
 ## Conclusion
 
--- a/docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
+++ b/docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
@@ -1,13 +1,13 @@
 ---
-title: "Choosing between anonymous and tuple types"
+title: Choosing between anonymous and tuple types
 description: Learn how to decide whether to design a type as a class, or to design a type as a struct.
-ms.date: 06/24/2020
+ms.date: 06/26/2020
 ms.technology: dotnet-standard
 ---
 
 # Choosing between anonymous and tuple types
 
-As a developer, choosing the appropriate type can be scary task. Anonymous types have been available since C# 3.0, while generic <xref:System.Tuple%602?displayProperty=nameWithType> types were introduced with .NET Framework 4.0. There are more modern options that exist, such as <xref:System.ValueTuple%602?displayProperty=nameWithType> - which as the name implies, provide a value type with the flexibility of anonymous types. In this article, you'll learn when it's appropriate to choose one type over the other.
+As a developer, choosing the appropriate type involves considering a types usability, performance, and tradeoffs compared to other types. Anonymous types have been available since C# 3.0, while generic <xref:System.Tuple%602?displayProperty=nameWithType> types were introduced with .NET Framework 4.0. Since then new options have been introduced with language level support, such as <xref:System.ValueTuple%602?displayProperty=nameWithType> - which as the name implies, provide a value type with the flexibility of anonymous types. In this article, you'll learn when it's appropriate to choose one type over the other.
 
 ## Usability and functionality
 
@@ -65,7 +65,7 @@ foreach (var tuple in
 }
 ```
 
-With the <xref:System.Tuple%602?displayProperty=nameWithType>, the instance exposes numbered item properties, such as `Item1`, and `Item2`. These property names can make it more difficult to understand the intent of the property values, as the property name only provides the ordinal. Furthermore, the `System.Tuple` types are reference `class` types. The <xref:System.ValueTuple%602?displayProperty=nameWithType> however, is a value `struct` type.
+With the <xref:System.Tuple%602?displayProperty=nameWithType>, the instance exposes numbered item properties, such as `Item1`, and `Item2`. These property names can make it more difficult to understand the intent of the property values, as the property name only provides the ordinal. Furthermore, the `System.Tuple` types are reference `class` types. The <xref:System.ValueTuple%602?displayProperty=nameWithType> however, is a value `struct` type. The following C# snippet, uses `ValueTuple<string, long>` to project into. In doing so, it assigns using a literal syntax.
 
 ```csharp-interactive
 var dates = new[]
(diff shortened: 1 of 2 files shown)


docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
```diff
@@ -0,0 +1,114 @@
+---
+title: "Choosing between anonymous and tuple types"
+description: Learn how to decide whether to design a type as a class, or to design a type as a struct.
+ms.date: 06/24/2020
+ms.technology: dotnet-standard
+---
+
+# Choosing between anonymous and tuple types
+
+As a developer, choosing the appropriate type can be scary task. Anonymous types have been available since C# 3.0, while generic <xref:System.Tuple%602?displayProperty=nameWithType> types were introduced with .NET Framework 4.0. There are more modern options that exist, such as <xref:System.ValueTuple%602?displayProperty=nameWithType> - which as the name implies, provide a value type with the flexibility of anonymous types. In this article, you'll learn when it's appropriate to choose one type over the other.
```
BillWagner, line 10: I'd rework this paragraph.  I wouldn't use the word "scary", but it does involve tradeoffs.

Secondly, I wouldn't refer to Tuples as "more modern". They were added in a later release in C#, but the concepts have been around for longer.

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Usability and functionality
+
+With the advent of C# 3.0, when anonymous types were introduced so to were Language-Integrated Query (LINQ) expressions. With LINQ, developers often project results from queries into anonymous types that hold a few select properties from the objects their working with. Consider the following example, that instantiates an array of <xref:System.DateTime> objects, and iterates through them projecting into an anonymous type with two properties.
```
BillWagner, line 14: Some grammar suggestions:

```suggestion
Anonymous types were introduced in C# 3.0 with Language-Integrated Query (LINQ) expressions. With LINQ, developers often project results from queries into anonymous types that hold a few select properties from the objects they're working with. Consider the following example, that instantiates an array of <xref:System.DateTime> objects, and iterates through them projecting into an anonymous type with two properties.
```

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
```diff
@@ -0,0 +1,114 @@
...
+
+foreach (var anonymous in
+             dates.Select(
+                 date => new { Formatted = $"{date:MMM dd, yyyy hh:mm zzz}", date.Ticks }))
+{
+    Console.WriteLine($"Ticks: {anonymous.Ticks}, formatted: {anonymous.Formatted}");
+}
+```
+
+Anonymous types are instantiated by using the [`new`](../../csharp/language-reference/operators/new-operator.md) operator, and the property names and types are inferred from the declaration. If two or more anonymous object initializers in an assembly specify a sequence of properties that are in the same order and that have the same names and types, the compiler treats the objects as instances of the same type. They share the same compiler-generated type information.
```
BillWagner, line 32: I suggest slightly stronger wording:
```suggestion
Anonymous types are instantiated by using the [`new`](../../csharp/language-reference/operators/new-operator.md) operator, and the property names and types are inferred from the declaration. If two or more anonymous object initializers in the same assembly specify a sequence of properties that are in the same order and that have the same names and types, the compiler treats the objects as instances of the same type. They share the same compiler-generated type information.
```

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
```diff
@@ -0,0 +1,114 @@
...
+
+The previous C# snippet projects an anonymous type that would be defined with two properties, much like the following compiler-generated C# class:
```
BillWagner, line 34: change verb form:

```suggestion
The previous C# snippet projects an anonymous type with two properties, much like the following compiler-generated C# class:
```

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
```diff
@@ -0,0 +1,114 @@
...
+
+    public f__AnonymousType0(string formatted, long ticks)
+    {
+        Formatted = formatted;
+        Ticks = ticks;
+    }
+}
+```
+
+For more information, see [anonymous types](../../csharp/programming-guide/classes-and-structs/anonymous-types.md). The same functionality exists with tuples when projecting into LINQ queries, you can select properties and put them into tuples. These tuples flow through the query, just as anonymous types would. Now consider the following example using the `System.Tuple<string, long>`.
```
BillWagner, line 50: nit:

```suggestion
For more information, see [anonymous types](../../csharp/programming-guide/classes-and-structs/anonymous-types.md). The same functionality exists with tuples when projecting into LINQ queries, you can select properties into tuples. These tuples flow through the query, just as anonymous types would. Now consider the following example using the `System.Tuple<string, long>`.
```

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
```diff
@@ -0,0 +1,114 @@
...
+
+foreach (var (formatted, ticks) in
+            dates.Select(
+                date => (Formatted: $"{date:MMM dd, yyyy at hh:mm zzz}", date.Ticks)))
+{
+    Console.WriteLine($"Ticks: {ticks}, formatted: {formatted}");
+}
+```
+
+The previous examples are all functionally equivalent, however; there are slight differences in their usability and their underlying implementations. One nicety with <xref:System.ValueTuple> is the ability to deconstruct.
```
BillWagner, line 86: I'd add a clause here:

```suggestion
The previous examples are all functionally equivalent, however; there are slight differences in their usability and their underlying implementations. One nicety with <xref:System.ValueTuple> is the ability to deconstruct the tuple into its members.
```

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Limitations
+
+While it might seem as though you should always use <xref:System.ValueTuple> over <xref:System.Tuple>, and anonymous types - there are some limitations that you should consider. The <xref:System.ValueTuple> types are mutable, whereas <xref:System.Tuple> are read-only. Anonymous types can be used in expression trees, while tuples cannot.
```
BillWagner, line 90: I'd say "tradeoffs" instead of "limitations":

```suggestion
While it might seem as though you should always use <xref:System.ValueTuple> over <xref:System.Tuple>, and anonymous types - there are tradeoffs you should consider. The <xref:System.ValueTuple> types are mutable, whereas <xref:System.Tuple> are read-only. Anonymous types can be used in expression trees, while tuples cannot.
```

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Key differences
+
+| Name                     | Type     | Custom property name | Deconstruction support | Expression tree support |
```
BillWagner, line 94: Consider either adding a column, or note on the "Type" column that System.Tuple and System.ValueTuple are public, but anonymous types are internal to an assembly.

Another important tradeoff is that you can't use anonymous types as a parameter to method (other than as a generic type argument). That's because anonymous types don't have names.

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
```diff
@@ -0,0 +1,114 @@
...
+            dates.Select(
+                date => (Formatted: $"{date:MMM dd, yyyy at hh:mm zzz}", date.Ticks)))
+{
+    Console.WriteLine($"Ticks: {ticks}, formatted: {formatted}");
+}
+```
+
+The previous examples are all functionally equivalent, however; there are slight differences in their usability and their underlying implementations. One nicety with <xref:System.ValueTuple> is the ability to deconstruct.
+
+## Limitations
```
BillWagner, line 88: I'd change this header to something like "tradeoffs"

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
```diff
@@ -0,0 +1,114 @@
...
+var dates = new[]
+{
+    DateTime.UtcNow.AddHours(-1),
+    DateTime.UtcNow,
+    DateTime.UtcNow.AddHours(1),
+};
+
+foreach (var (formatted, ticks) in
+            dates.Select(
+                date => (Formatted: $"{date:MMM dd, yyyy at hh:mm zzz}", date.Ticks)))
```
BillWagner, line 80: This would be a really good place to point out the language support for tuples use `System.ValueTuple`. That's more than deconstruct, which you mention below. In addition, tuples support:

- tuple assignment
- tuple equality check.

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
BillWagner, line 94: Also, add a column to compare named members. Both anonymous types and tuples (using ValueTuple) support named members. System.Tuple has properties named `Item1`, `Item2` etc.

BillWagner (changes requested):
This is a great start @IEvangelist 

I have some suggestions for additional facets to compare. I'd like to take a second looks after you add those areas.

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
IEvangelist: The table already has that `Custom property name`, should I change the column name?

BillWagner: Anonymous types, System.Tuples and System.ValueTuples could be used with that function. However, anonymous types can't be arguments to non-generic methods.

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
```diff
@@ -1,13 +1,13 @@
...
 ---
 
 # Choosing between anonymous and tuple types
 
-As a developer, choosing the appropriate type can be scary task. Anonymous types have been available since C# 3.0, while generic <xref:System.Tuple%602?displayProperty=nameWithType> types were introduced with .NET Framework 4.0. There are more modern options that exist, such as <xref:System.ValueTuple%602?displayProperty=nameWithType> - which as the name implies, provide a value type with the flexibility of anonymous types. In this article, you'll learn when it's appropriate to choose one type over the other.
+As a developer, choosing the appropriate type involves considering a types usability, performance, and tradeoffs compared to other types. Anonymous types have been available since C# 3.0, while generic <xref:System.Tuple%602?displayProperty=nameWithType> types were introduced with .NET Framework 4.0. Since then new options have been introduced with language level support, such as <xref:System.ValueTuple%602?displayProperty=nameWithType> - which as the name implies, provide a value type with the flexibility of anonymous types. In this article, you'll learn when it's appropriate to choose one type over the other.
```
BillWagner, line 10: nits:
```suggestion
Choosing the appropriate type involves considering its usability, performance, and tradeoffs compared to other types. Anonymous types have been available since C# 3.0, while generic <xref:System.Tuple%602?displayProperty=nameWithType> types were introduced with .NET Framework 4.0. Since then new options have been introduced with language level support, such as <xref:System.ValueTuple%602?displayProperty=nameWithType> - which as the name implies, provide a value type with the flexibility of anonymous types. In this article, you'll learn when it's appropriate to choose one type over the other.
```

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
```diff
@@ -104,6 +104,10 @@ While it might seem as though you should always use <xref:System.ValueTuple> ove
 | <xref:System.Tuple>      | `public`        | `class`  | ❌                   | ❌                     | ✔️                     |
 | <xref:System.ValueTuple> | `public`        | `struct` | ✔️                   | ✔️                     | ❌                     |
 
+### Serialization
+
+One important consideration when choosing a type, is whether or not it will need to be serialized. Serialization is the process of converting the state of an object into a form that can be persisted or transported. For more information, see [serialization](../../csharp/programming-guide/concepts/serialization/index.md). With anonymous types, and tuple types it is not recommended to use them if you need to serialization. While it is possible to use them with serialization, it is not fully supported, and requires workarounds.
```
BillWagner, line 109: nits:

```suggestion
One important consideration when choosing a type, is whether or not it will need to be serialized. Serialization is the process of converting the state of an object into a form that can be persisted or transported. For more information, see [serialization](../../csharp/programming-guide/concepts/serialization/index.md). When serialization is important, creating a `class` or `struct` is preferred over anonymous types or tuple types.
```

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
```diff
@@ -83,23 +83,30 @@ foreach (var (formatted, ticks) in
...
+| <xref:System.Tuple>      | `public`        | `class`  | ❌                   | ❌                     | ✔️                     |
+| <xref:System.ValueTuple> | `public`        | `struct` | ✔️                   | ✔️                     | ❌                     |
 
 ## Performance
 
-Performance is always a topic of discussion when referring to value types vs reference types, often leading to conversations of stack versus heap allocations. However, performance between these types is fairly similar with each other - in other words, performance is irrelevant. The micro-optimizations you might gain from choosing tuples over anonymous types is more often than not negligible.
+Performance is always a topic of discussion when referring to value types vs reference types, often leading to conversations of stack versus heap allocations. However, performance between these types is fairly similar with each other - in other words, performance differences are negligible. The micro-optimizations you might gain from choosing tuples over anonymous types is more often than not irrelevant.
```
BillWagner, line 113: I'll suggest wording similar to the following. The performance differences *could* be major, but it depends on how either type is used:

```suggestion
Performance between these types depends on the scenario. The major impact involves the tradeoff between allocations and copying. In most scenarios, the impact is small. When major impacts could arise, measurements should be takent to inform the decision.
```

BillWagner (approved):
This looks great @IEvangelist 

I had a few small wording suggestions, and then :shipit:

docs/standard/design-guidelines/choosing-between-anonymous-and-tuple.md
```diff
@@ -0,0 +1,125 @@
+---
+title: Choosing between anonymous and tuple types
+description: Learn how to decide whether to design a type as a class, or to design a type as a struct.
```
pkulikov, line 3: @IEvangelist sorry for the late feedback, but this description doesn't match the article

IEvangelist: Ah, yes! Good catch... updating now.

PR #19239, opened 2020-06-27
Format codes based on .editorconfig, Part 22

nxtn:
As you can see in the "Revert due to compilation errors" commits, a number of projects can't be compiled. It's too early to enable the build checks.


NextTurn:
OK. The passing rate is 15%. :)

BillWagner (approved):
Thanks again @NextTurn 

We'll :shipit: now.

BillWagner:
@NextTurn 
> It's too early to enable the build checks.

We haven't set this check as a blocking check yet. We knew this would expose a number of issues. In most cases, PRs modify one or two samples, and the build errors provide information that help us improve.

In the work you're doing for formatting code, we can tell by inspection that the change won't introduce new errors. 

ping @dotnet/docs  Should we formally agree to merge these editorconfig PRs without mandating a build fix?

The other option is to combine source code configuration fixes with creating buildable projects that address both concerns.

tdykstra:
I vote for

>combine source code configuration fixes with creating buildable projects that address both concerns.

PR #19240, opened 2020-06-27
Format codes based on .editorconfig, Part 23

nxtn:
According to the pass rate calculated in #19239, this PR starts with 1K file updates.


NextTurn:
OK. 33 files survived.

BillWagner (approved):
This LGTM. I'll :shipit:

PR #19249, opened 2020-06-28
Improved explanation of nullable Boolean logical operators

### LINKED ISSUE #19237: Nullable boolean operator explanation
The section on nullable boolean operators is a bit verbose with a table defining the semantics, and I think it could be improved by replacing the table with a description of them that makes logical sense. I remember first looking at it thinking how daunting it would be to memorize that table. I think a good explanation would cover the following points:

1. They always evaluate both their operands.
1. They work as you'd expect for non-null values.
1. For `null` operands, `&` will return `false` if either operand is `false`, `|` will return `true` if either operand is `true`, and they return null in all other cases. This is like short-circuiting in that a definite answer can be determined even if you "don't know" whether the other operand is either true or false, but it still evaluates both expressions.

I actually ended up covering this somewhat in a [comment](https://stackoverflow.com/questions/4349016/and-operation-cannot-be-applied-between-nullable-bools/4349293#comment110612189_4349293) on Stack Overflow.

---
#### Document Details

* ID: 6d2d3240-ed65-94f2-335b-3ea483c51a5d
* Version Independent ID: 91eaa72e-c2fc-e0d3-ca11-994aadf02271
* Content: [Boolean logical operators - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators#nullable-boolean-logical-operators)
* Content Source: [docs/csharp/language-reference/operators/boolean-logical-operators.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/operators/boolean-logical-operators.md)
* Product: **dotnet-csharp**
* GitHub Login: @pkulikov
* Microsoft Alias: **wiwagn**

pkulikov:
Fixes #19237

### DIFF
--- a/docs/csharp/language-reference/operators/boolean-logical-operators.md
+++ b/docs/csharp/language-reference/operators/boolean-logical-operators.md
@@ -1,7 +1,7 @@
 ---
 title: "Boolean logical operators - C# reference"
 description: "Learn about C# operators that perform logical negation, conjunction (AND), and inclusive and exclusive disjunction (OR) operations with Boolean operands."
-ms.date: 09/27/2019
+ms.date: 06/29/2020
 author: pkulikov
 f1_keywords: 
   - "!_CSharpKeyword"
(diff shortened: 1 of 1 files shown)


adamyakes:
@pkulikov Do you think it's worth rementioning that they always evaluate both their operands?

pkulikov:
> @pkulikov Do you think it's worth rementioning that they always evaluate both their operands?

No; however, I've added the links to the corresponding sections that discuss that.

IEvangelist (approved):

PR #19252, opened 2020-06-29
Use 3.0.3 as "realistic" example

robertmuehsig:
## Summary

The article mentioned .NET Core 3.0.4 or 3.0.5 to demonstrate the general idea behind the version selection process. .NET 3.0 is out of support and the latest patch release is version 3.0.3. 
If I understand the versioning correctly it would be good to use 3.0.3 as a realistic example otherwise it might confuse people like me :)

### DIFF
--- a/docs/core/versions/selection.md
+++ b/docs/core/versions/selection.md
@@ -75,7 +75,7 @@ A given SDK supports a fixed set of frameworks, capped to the target framework o
 
 When you run an application from source with [`dotnet run`](../tools/dotnet-run.md), from a [**framework-dependent deployment**](../deploying/index.md#publish-runtime-dependent) with [`dotnet myapp.dll`](../tools/dotnet.md#description), or from a [**framework-dependent executable**](../deploying/index.md#publish-runtime-dependent) with `myapp.exe`, the `dotnet` executable is the **host** for the application.
 
-The host chooses the latest patch version installed on the machine. For example, if you specified `netcoreapp3.0` in your project file, and `3.0.3` is the latest .NET runtime installed, the `3.0.3` runtime is used.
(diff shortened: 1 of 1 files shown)


docs/core/versions/selection.md
```diff
@@ -75,13 +75,13 @@ A given SDK supports a fixed set of frameworks, capped to the target framework o
 
 When you run an application from source with [`dotnet run`](../tools/dotnet-run.md), from a [**framework-dependent deployment**](../deploying/index.md#publish-runtime-dependent) with [`dotnet myapp.dll`](../tools/dotnet.md#description), or from a [**framework-dependent executable**](../deploying/index.md#publish-runtime-dependent) with `myapp.exe`, the `dotnet` executable is the **host** for the application.
 
-The host chooses the latest patch version installed on the machine. For example, if you specified `netcoreapp3.0` in your project file, and `3.0.4` is the latest .NET runtime installed, the `3.0.4` runtime is used.
+The host chooses the latest patch version installed on the machine. For example, if you specified `netcoreapp3.0` in your project file, and `3.0.3` is the latest .NET runtime installed, the `3.0.3` runtime is used.
```
BillWagner, line 78: Let's change this from 3.0.3 to 3.0.2. I think it's important that the host will use the latest installed patch, even if a newer patch version is available. 
```suggestion
The host chooses the latest patch version installed on the machine. For example, if you specified `netcoreapp3.0` in your project file, and `3.0.2` is the latest .NET runtime installed, the `3.0.2` runtime is used.
```

BillWagner (changes requested):
Thanks for making these updates @robertmuehsig 

I agree that using version numbers that did ship will make a more compelling explanation.

I do have one suggestion. Once you accept that suggestion, I'll :shipit:

BillWagner (approved):

BillWagner:
Thanks again @robertmuehsig I'll :shipit: now.

PR #19277, opened 2020-07-01
Standardize sample projects, Part 4

nxtn:
Removes `<LangVersion>latest</LangVersion>`.

For `<LangVersion>` with specific version numbers, do we treat those as the minimum version requirements?


BillWagner (approved):
Thanks again @NextTurn 

BillWagner:
@NextTurn 

> For `<LangVersion>` with specific version numbers, do we treat those as the minimum version requirements?

In almost all cases, that assumption is correct. My recollection is that most of them will be setting a 7.x point release when "latest" meant "7.0", or "latest major". There may be a very small number of samples that are set to an older version where we wanted to show older syntax to contrast that with a newer syntax. I think those are mostly limited to demonstrating breaking changes. Examples would be where the scope of the `foreach` iteration variable changed, and where "better betterness" changes overload resolution.

PR #19287, opened 2020-07-02
Standardize sample projects, Part 6

nxtn:
Migrates...


IEvangelist (approved):

PR #19325, opened 2020-07-05
Remove redundant ToString calls

drewnoakes:
These calls are redundant. Removing them makes the samples simpler.

PR #19326, opened 2020-07-05
Remove redundant ToString calls

drewnoakes:
These calls are redundant. Removing them makes the samples simpler.

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR_System/system.DateTimeOffset.Conceptual/cs/Conceptual2.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR_System/system.DateTimeOffset.Conceptual/cs/Conceptual2.cs
@@ -16,14 +16,14 @@ public class DateManipulation
       DateTime utcTime = DateTime.UtcNow;
 
       Console.WriteLine("Difference between {0} and {1} time: {2}:{3} hours",
-                        localTime.Kind.ToString(),
-                        utcTime.Kind.ToString(),
+                        localTime.Kind,
+                        utcTime.Kind,
                         (localTime - utcTime).Hours,
                         (localTime - utcTime).Minutes);
       Console.WriteLine("The {0} time is {1} the {2} time.",
-                        localTime.Kind.ToString(),
+                        localTime.Kind,
                         Enum.GetName(typeof(TimeComparison), localTime.CompareTo(utcTime)),
-                        utcTime.Kind.ToString());
+                        utcTime.Kind);
    }
 }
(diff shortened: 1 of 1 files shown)


samples/snippets/csharp/VS_Snippets_CLR_System/system.DateTimeOffset.Conceptual/cs/Conceptual2.cs
```diff
@@ -16,14 +16,14 @@ public static void Main()
       DateTime utcTime = DateTime.UtcNow;
 
       Console.WriteLine("Difference between {0} and {1} time: {2}:{3} hours",
-                        localTime.Kind.ToString(),
-                        utcTime.Kind.ToString(),
+                        localTime.Kind,
```
IEvangelist, line 19: 😵 Project missing. A project (and optionally a solution file) must be in this directory or one of the parent directories to validate and build this code.

IEvangelist (changes requested):

samples/snippets/csharp/VS_Snippets_CLR_System/system.DateTimeOffset.Conceptual/cs/Conceptual2.cs
drewnoakes: That may be the case, but I don't see what that has to do with this suggested change. I don't have the bandwidth right now to work on docs infrastructure.

drewnoakes: FWIW this similar PR was merged despite the same problem: https://github.com/dotnet/docs/pull/19325/files

BillWagner: @IEvangelist I merged the similar PR based on the fact that they were straightforward changes, and didn't involve any updates to the logic or associated articles. 

While I'd also prefer having all samples under our CI system, these do improve quality. We're not worse off than before.

IEvangelist: Thank you @BillWagner and @drewnoakes...  I see that now. I was just basing this on what @adegeo was pushing for with the CI build system. While I agree that these changes do improve the quality, we should probably figure out a way to disable the build system somehow conditionally.

IEvangelist (approved):

PR #19342, opened 2020-07-07
Standardize sample projects, Part 9

nxtn:
`EnableDefaultCompileItems` and `EnableDefaultEmbeddedResourceItems` are still TODO.


BillWagner (approved):
Thanks @NextTurn 

I'll :shipit: now. I think that the CI build failure is because the PR list is picking up the deleted SLN file. I'll look at the logic for that.

PR #19355, opened 2020-07-07
Update about.md to include download button link for .NET Core

### LINKED ISSUE #18012: Follow up - Remove 2nd Layer - L2 - Navigation part 2
After the L2 navigation is removed, these are items to address that were not a blocker to removing the L2 but we should consider for follow up.

- Analyze  Get Started button click traffic to see if we can improve getting developers directly to where they really might want go. For example there are a lot of referrals to that link from the Framework install pages. Do they just want to start their downloads?  The hypothesis may require some testing over time.  
- Verify if there’s any relevant place to add the [about link](https://dotnet.microsoft.com/learn/dotnet/what-is-dotnet)
- Check where there’s high traffic to the download pages and move the link to an action button above the fold if too low. [Example of one above the fold](https://docs.microsoft.com/en-us/dotnet/framework/install/on-windows-10#net-framework-48)

wadepickett:
[Internal Review](https://review.docs.microsoft.com/en-us/dotnet/core/about?branch=pr-en-us-19355)

(See download button at top of topic)

Fixes #18012  (Partly)

- Added Download .NET core button above the fold.   I decided not to make an entirely new section just for that and seemed reasonable rather to put it at the end of the first section.  Developers need to be able to select Linux, macOS and Windows, so the link is to the top level download page rather than to a more direct download.

### DIFF
--- a/docs/core/about.md
+++ b/docs/core/about.md
@@ -5,6 +5,9 @@ ms.date: 03/26/2020
 ---
 # .NET Core overview
 
+> [!div class="button"]
+> [Download .NET Core](https://dotnet.microsoft.com/download)
+
 .NET Core has the following characteristics:
 
 - **Cross platform:** Runs on Windows, macOS, and Linux [operating systems](https://github.com/dotnet/core/blob/master/os-lifecycle-policy.md).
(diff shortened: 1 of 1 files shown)


wadepickett:
@tdykstra do you think the download button should be at the very top or is it OK at the bottom of the first section?  I didn't want to mess up SEO by making it the very first text item.

tdykstra (approved):
It looks strange to me after the intro section, I would put it directly under the H1.  I doubt it will substantially hurt SEO, but we could check in a couple of months and move it if necessary.

PR #19360, opened 2020-07-08
Standardize sample projects, Part 10

nxtn:


BillWagner (approved):
Thanks again @NextTurn 

I'll :shipit: now.

PR #19365, opened 2020-07-08
Enhancements to OOP article

Youssef1313:


docs/csharp/programming-guide/concepts/object-oriented-programming.md
```diff
@@ -5,8 +5,9 @@ ms.assetid: 89574786-65ef-4335-88bc-fbacd094f183
 ---
 # Object-Oriented programming (C#)
 
-C# provides full support for object-oriented programming including encapsulation, inheritance, and polymorphism.
+C# provides full support for object-oriented programming including abstraction, encapsulation, inheritance, and polymorphism.
 
+- *Abstraction* means hiding the unnecessary details from the class consumers.
```
Youssef1313, line 10: cc: @BillWagner @pkulikov, you may come up with a better phrasing here.

docs/csharp/programming-guide/concepts/object-oriented-programming.md
```diff
@@ -23,7 +24,7 @@ class SampleClass
 }
 ```
 
-C# also provides types called *structures* that are useful when you don't need support for inheritance or polymorphism.
+C# also provides types called *structures* that are useful when you don't need support for inheritance or polymorphism. For more information, see [Choosing between class and struct](../../../standard/design-guidelines/choosing-between-class-and-struct.md).
```
Youssef1313, line 27: Choosing between struct and classes is more deep than just whether the inheritance and polymorphism are needed or not. So I included the article that explains that in details.

docs/csharp/programming-guide/concepts/object-oriented-programming.md
pkulikov: @Youssef1313 what about

> *Abstraction* means hiding implementation details from type consumers

At least, "class" should be replaced with "type", as structure types also can abstract details away.

IEvangelist (approved):

PR #19380, opened 2020-07-09
Add missing whitespace

Hoffs:
## Summary

Fixes simple typo of missing whitespace between option and its value.

As per dotnet tool uninstall output it should be `--tool-path <PATH>`


IEvangelist (approved):

PR #19382, opened 2020-07-09
Added npm and docker support for dependabot

IEvangelist:
## Summary

Added npm and docker support for dependabot. https://dependabot.com/docs/config-file/#example-configyml-files

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -341,3 +341,13 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+- package_manager: "javascript"
+  directory: "/samples/snippets"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
+- package_manager: "docker"
+  directory: "/samples/snippets"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 10


gewarren (approved):

.github/dependabot.yml
```diff
@@ -341,3 +341,13 @@ updates:
...
+- package_manager: "javascript"
+  directory: "/samples/snippets"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
+- package_manager: "docker"
+  directory: "/samples/snippets"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 10
```
Youssef1313, line 353: Will dependabot look for project files in nested directories?
If so, this whole file can be shrinked.

adegeo:
Should the setting use a glob pattern that searches any `snippets` folder in the entire repo? As we move things away from the `samples/snippets` and into `docs/.../folder/snippets` paths, these won't currently be picked up.

Youssef1313:
@adegeo dependabot doesn't support wildcards AFAIK.

IEvangelist:
This PR is no longer needed....

PR #19385, opened 2020-07-09
Typo fix in creating-a-new-windows-form.md

petershh:
Removed a point after "Windows Forms Coordinates" link


adegeo (approved):
Thanks for spotting that discrepancy !

PR #19389, opened 2020-07-09
Fix "tar balls" to "tarballs"

JasonCard:
Currently, "tar balls" is being over-localized by Machine Translation in Italian.
The [source string](https://docs.microsoft.com/en-us/dotnet/core/introduction) is currently “Zips and tar balls”, but it seems like the computing term should be one word “tarballs”? 
I tried in Bing translator and changing the source to “Tarballs” should fix the Italian Machine Translation back to keeping it in English.

Fixes [https://github.com/dotnet/docs.it-it/pull/83](https://github.com/dotnet/docs.it-it/pull/83)


mairaw (approved):
Thanks @JasonCard!

PR #19390, opened 2020-07-09
Fix C# interactive experience

pkulikov:
Somehow, as it is, `min` and `max` local variables clash with the same-named variables in a local function. This is the error message, when running the snippet from the article in C# interactive (code runs without problems on desktop):

> (20,9): error CS0136: A local or parameter named 'min' cannot be declared in this scope because that name is used in an enclosing local scope to define a local or parameter

So, I've just renamed the variables and it worked in interactive.

Link to the article in question: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-tuples#use-cases-of-tuples


BillWagner (approved):
Thanks @pkulikov 

I'll :shipit: now.

BillWagner:
ping @LadyNaggaga @jonsequitur for awareness.

PR #18819, opened 2020-06-05
VS Mac .NET Core tutorials

tdykstra:
Fixes #18512 
Fixes #18513 
Fixes #18749 

In addition to creating the VS Mac series, this PR makes the 3 "swim lanes" as similar as possible, so some changes were made to the VS and VS Code series as well.

PR #19408, opened 2020-07-10
Delete unused snippets

Youssef1313:
"using-on-mac-vs-full-solution.md" was removed in #18819, so these samples became unused.

cc: @tdykstra


Youssef1313:
> ##[error]😵 Project missing. A project (and optionally a solution file) must be in this directory or one of the parent directories to validate and build this code.

@adegeo, Is that related to my changes?

adegeo:
We don't have logic yet to handle deleted files. So these errors can be ignored.

adegeo:
@tdykstra I'll let you review. I didn't see anything wrong with this but I believe you made changes in this area last.

tdykstra (approved):

PR #19410, opened 2020-07-10
Update outdated landing page

Youssef1313:
`vb-with-visual-studio.md` was deleted and now redirects to `with-visual-studio`, which is more than "Hello World"


adegeo (approved):
LGTM!

PR #19416, opened 2020-07-10
Vector.Lerp breaking change

### LINKED ISSUE #18937: Vector2/4 Lerp behavior changed for .NET 5
## Vector2/3 Lerp behavior changed for .NET 5

`Vector2.Lerp(value1, value2, amount)` and `Vector4.Lerp(value1, value2, amount)` were implemented as `value1 + (value2 - value1) * amount`. However, due to floating-point rounding error this will not always return `value2` when `amount == 1.0f`.

The implementation was updated to use the same algorithm as `Vector3.Lerp(value1, value2, amount)` which is `(value1 * (1.0f - amount)) + (value2 * amount)`. This algorithm correctly accounts for the rounding error and also allows the algorithm to be freely optimized using `FusedMultiplyAdd` when available.

### Version introduced

.NET 5

### Old behavior

For certain inputs, when `amount` was `1.0f` the result would not be precisely `value2`.

### New behavior

When `amount` is `1.0f`, the result will be precisely `value2`.

### Reason for change

The result would not be correctly interpolated as you could never get `value2` as a result for certain inputs.

### Recommended action

If maintaining the old behavior is desired, the user could implement their own `Lerp` function that uses the previous algorithm:  `value1 + (value2 - value1) * amount`

### Category

Core .NET libraries

### Affected APIs

[Vector2.Lerp(Vector2, Vector2, Single)](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.vector2.lerp?view=net-5.0)
[Vector4.Lerp(Vector4, Vector4, Single)](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.vector4.lerp?view=net-5.0)

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #18937

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/3.1-5.0?branch=pr-en-us-19416#behavior-change-for-vector2lerp-and-vector4lerp).

### DIFF
--- a/docs/core/compatibility/3.1-5.0.md
+++ b/docs/core/compatibility/3.1-5.0.md
@@ -92,9 +92,14 @@ If you're migrating from version 3.1 of .NET Core, ASP.NET Core, or EF Core to v
 
 ## Core .NET libraries
 
+- [Behavior change for Vector2.Lerp and Vector4.Lerp](#behavior-change-for-vector2lerp-and-vector4lerp)
 - [SSE and SSE2 CompareGreaterThan methods properly handle NaN inputs](#sse-and-sse2-comparegreaterthan-methods-properly-handle-nan-inputs)
 - [CounterSet.CreateCounterSetInstance now throws InvalidOperationException if instance already exist](#countersetcreatecountersetinstance-now-throws-invalidoperationexception-if-instance-already-exists)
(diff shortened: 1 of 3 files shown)


tdykstra (approved):
LGTM

gewarren:
Thanks Tom!

PR #19430, opened 2020-07-11
Re-generate dependabot.yml file

Youssef1313:
Generated by https://github.com/BillWagner/dependadotnet/tree/version-2-syntax

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,11 @@
 version: 2
 updates:
 - package-ecosystem: nuget
+  directory: "/samples/snippets/csharp/new-in-6" # new-in-6.csproj
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
+- package-ecosystem: nuget
   directory: "/samples/snippets/csharp/pipelines" # Pipes.csproj
   schedule:
     interval: daily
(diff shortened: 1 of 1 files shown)


Youssef1313 (commented):
The file became too long, I was expecting small changes due to samples that are moved from ~/samples to ~/docs.

@BillWagner, Is there some kind of filtering that you don't have in your fork?

---

**Edit:** The current large number is because some project files that was in the old format (didn't contain "PackageReference") was converted to the new sdk-style project which now have "PackageReference". Now, it may exceed dependabot limits.

BillWagner:
Waiting on this PR. We have a question to he dependabot team regarding configuration and navigating child folders. Pending that outcome, the configuration may get much simpler.

Youssef1313:
That would be really great.

Youssef1313:
@BillWagner, their response to @IEvangelist is that it's not supported currently.

I've a small workaround to bypass the limits. What if we put a sln file in the repository's root, and put **all** projects in it.
The complexity of keeping track of new/moved projects will be the same, but I think it will bypass the maximum number of entries as the config file will have only one sln file.

What do you think?

IEvangelist:
> @BillWagner, their response to @IEvangelist is that it's not supported currently.
> 
> I've a small workaround to bypass the limits. What if we put a sln file in the repository's root, and put **all** projects in it.
> The complexity of keeping track of new/moved projects will be the same, but I think it will bypass the maximum number of entries as the config file will have only one sln file.
> 
> What do you think?

I defer to @BillWagner, I'm not sure how I feel about that. It's not a bad idea.

BillWagner:
I like that idea as well. Ping @adegeo for thoughts as well.

adegeo:
That won't work. The CI system would find that root sln and always compile it no matter what code file we change anywhere in the repo. And, to me, it sounds like too much work to maintain.

Personally I don't like a global dependabot and I don't think we need it. 🗡️ As an alternative what if we just added dependabot config for specific projects where we know we want to have the latest package references.

BillWagner:
Does Dependabot support multiple config files in a repo? I didn't see that.

Lacking that, how do we want to proceed?

Youssef1313:
@BillWagner, I'm thinking of making a C# tool that directly updates the dependencies. However, I'm not familiar with making GitHub bots, so if I done that, the tool will be need to run manually and making a PR manually. What do you think?

adegeo:
Well I mean that if dependabot only runs on a project that is defined in its root repo config file (I think that's how it works??) then it should remain empty until we have a project we want to specifically declare in it as being monitored by the bot.

BillWagner:
@Youssef1313  I think we should close this.

Here's why:
- There's an issue for dependabot to support the folders with *sln* files (as opposed to each project). That helps.
- There's another issue for TFMs (no action yet).

What do you think?

Youssef1313:
I agree.

Issue #19440, opened 2020-07-11
Explain the usage of "await using" statement.


Youssef1313:
I'm not able to find any article that explains (or mentions) the usage of `await using`.

https://stackoverflow.com/questions/58791938/c-sharp-8-understanding-await-using-syntax

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 07164632-3cf8-864e-cd1e-485b73ed95a0
* Version Independent ID: 67bfb75a-8dd8-ecf3-ce15-2a531833e8b5
* Content: [await operator - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/await)
* Content Source: [docs/csharp/language-reference/operators/await.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/operators/await.md)
* Product: **dotnet-csharp**
* Technology: **csharp-language-reference**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

pkulikov:
@Youssef1313 thanks, I'll update the doc with the links to the following sections:
- https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#asynchronous-disposable
- https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#using-async-disposable

PR #19441, opened 2020-07-11
Document errors 8400, 8401, 8403, 8410, and 8411

Youssef1313:
Starting from [here](https://github.com/dotnet/roslyn/blob/bf301b0f12e9a84080f9d959fb66ae95380cb121/src/Compilers/CSharp/Portable/Errors/ErrorCode.cs#L1586).

### DIFF
--- a/docs/csharp/language-reference/compiler-messages/cs8400.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8400.md
@@ -11,7 +11,6 @@ author: Youssef1313
 
 Feature 'feature' is not available in C# 8.0. Please use language version \<version> or greater.
 
-
 ## To correct this error
 
 Upgrade your project to at least the version mentioned in the error.
--- a/docs/csharp/language-reference/compiler-messages/cs8400.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8400.md
@@ -1,0 +1,21 @@
+---
+title: Compiler Error CS8400
+ms.date: 07/11/2020
+f1_keywords:
+  - "CS8400"
+helpviewer_keywords:
+  - "CS8400"
+author: Youssef1313
+---
+# Compiler Error CS8400
+
+Feature 'feature' is not available in C# 8.0. Please use language version <version> or greater.
+
+
+## To correct this error
+
+Upgrade your project to at least the version mentioned in the error.
+
+## See also
+
+- [C# language versioning](../configure-language-version.md)
--- a/docs/csharp/language-reference/compiler-messages/cs8400.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8400.md
@@ -9,7 +9,7 @@ author: Youssef1313
 ---
 # Compiler Error CS8400
 
-Feature 'feature' is not available in C# 8.0. Please use language version <version> or greater.
+Feature 'feature' is not available in C# 8.0. Please use language version \<version> or greater.
 
 
 ## To correct this error
--- a/docs/csharp/language-reference/compiler-messages/cs8401.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8401.md
@@ -15,8 +15,10 @@ In older versions before C# 8.0, verbatim interpolated strings were only allowed
 
 ## To correct this error
 
-1. Replace `@$` with `$@`.
-2. Upgrade to C# 8.0 or later.
+Do one of the following options:
+
+- Replace `@$` with `$@`.
+- Upgrade to C# 8.0 or later.
 
 ## See also
 
--- a/docs/csharp/language-reference/compiler-messages/cs8401.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8401.md
@@ -15,10 +15,7 @@ In older versions before C# 8.0, verbatim interpolated strings were only allowed
 
 ## To correct this error
 
-Do one of the following options:
-
-- Replace `@$` with `$@`.
-- Upgrade to C# 8.0 or later.
+Either replace `@$` with `$@`, or upgrade to C# 8.0 or later.
 
 ## See also
 
--- a/docs/csharp/language-reference/compiler-messages/cs8401.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8401.md
@@ -1,0 +1,23 @@
+---
+title: Compiler Error CS8401
+ms.date: 07/11/2020
+f1_keywords:
+  - "CS8401"
+helpviewer_keywords:
+  - "CS8401"
+author: Youssef1313
+---
+# Compiler Error CS8401
+
+To use '@$' instead of '$@' for an interpolated verbatim string, please use language version '8.0' or greater.
+
+In older versions before C# 8.0, verbatim interpolated strings were only allowed using `$@`. Starting with C# 8.0, you can use `$@` or `@$`.
+
+## To correct this error
+
+1. Replace `@$` with `$@`.
+2. Upgrade to C# 8.0 or later.
+
+## See also
+
+- [String interpolation in C#](../../tutorials/string-interpolation.md)
--- a/docs/csharp/language-reference/compiler-messages/cs8403.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8403.md
@@ -9,7 +9,7 @@ author: Youssef1313
 ---
 # Compiler Error CS8403
 
-Method 'method' with an iterator block must be 'async' to return '{IAsyncEnumerable<T>|IAsyncEnumerator<T>}'
+Method 'method' with an iterator block must be 'async' to return '{IAsyncEnumerable\<T>|IAsyncEnumerator\<T>}'
 
 ## To correct this error
 
--- a/docs/csharp/language-reference/compiler-messages/cs8403.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8403.md
@@ -1,0 +1,20 @@
+---
+title: Compiler Error CS8403
+ms.date: 07/11/2020
+f1_keywords:
+  - "CS8403"
+helpviewer_keywords:
+  - "CS8403"
+author: Youssef1313
+---
+# Compiler Error CS8403
+
+Method 'method' with an iterator block must be 'async' to return '{IAsyncEnumerable<T>|IAsyncEnumerator<T>}'
+
+## To correct this error
+
+Mark your method with `async` modifier.
+
+## See also
+
+- [`async` (C# Reference)](../keywords/async.md)
--- a/docs/csharp/language-reference/compiler-messages/cs8403.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8403.md
@@ -13,7 +13,7 @@ Method 'method' with an iterator block must be 'async' to return '{IAsyncEnumera
 
 ## To correct this error
 
-Mark your method with `async` modifier.
+Mark your method with the `async` modifier.
 
 ## See also
 
--- a/docs/csharp/language-reference/compiler-messages/cs8410.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8410.md
@@ -37,3 +37,7 @@ class Example
 {
 }
 ```
+
+## See also
+
+- [Implement a DisposeAsync method](../../../standard/garbage-collection/implementing-disposeasync.md)
--- a/docs/csharp/language-reference/compiler-messages/cs8410.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8410.md
@@ -1,0 +1,34 @@
+---
+title: Compiler Error CS8410
+ms.date: 07/11/2020
+f1_keywords:
+  - "CS8410"
+helpviewer_keywords:
+  - "CS8410"
+author: Youssef1313
+---
+# Compiler Error CS8410
+
+'type': type used in an asynchronous using statement must be implicitly convertible to 'System.IAsyncDisposable' or implement a suitable 'DisposeAsync' method.
+
+The expression inside an `await using` statement must have a `DisposeAsync` method.
+
+## To correct this error
+
+Use `using` instead of `await using`.
+
+## Example
+
+<!-- TODO: The current example is a little bit poor. -->
+
+```csharp
+class C
+{
+    public static async System.Threading.Tasks.Task Main()
+    {
+        await using (Main)
+        {
+        }
+    }
+}
+```
--- a/docs/csharp/language-reference/compiler-messages/cs8410.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8410.md
@@ -22,9 +22,9 @@ Use `using` instead of `await using`.
 ```csharp
 using System.Threading.Tasks;
 
-static class Program 
+static class Program
 {
-    static async Task Main() 
+    static async Task Main()
     {
         // error CS8410: 'Example': type used in an asynchronous using statement
         // must be implicitly convertible to 'System.IAsyncDisposable' or implement
--- a/docs/csharp/language-reference/compiler-messages/cs8410.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8410.md
@@ -22,7 +22,7 @@ Remove the `await using` keywords, or implement a suitable `DisposeAsync` method
 ```csharp
 using System.Threading.Tasks;
 
-static class Program
+class Program
 {
     static async Task Main()
     {
--- a/docs/csharp/language-reference/compiler-messages/cs8410.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8410.md
@@ -19,8 +19,6 @@ Use `using` instead of `await using`.
 
 ## Example
 
-<!-- TODO: The current example is a little bit poor. -->
-
 ```csharp
 using System.Threading.Tasks;
 
@@ -28,7 +26,7 @@ static class Program
 {
     static async Task Main() 
     {
-        // error CS8410: 'ImplicitReader': type used in an asynchronous using statement
+        // error CS8410: 'Example': type used in an asynchronous using statement
         // must be implicitly convertible to 'System.IAsyncDisposable' or implement
         // a suitable 'DisposeAsync' method.
         await using var reader = new Example();
--- a/docs/csharp/language-reference/compiler-messages/cs8410.md
+++ b/docs/csharp/language-reference/compiler-messages/cs8410.md
@@ -22,13 +22,20 @@ Use `using` instead of `await using`.
 <!-- TODO: The current example is a little bit poor. -->
 
 ```csharp
-class C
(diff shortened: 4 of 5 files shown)


Youssef1313:
@BillWagner, Some of these newly-documented error messages doesn't have examples, and others have a little bit unrealistic examples (from roslyn's unit tests). Hope you can come up with better examples.

BillWagner:
Thanks @Youssef1313 

Let me think about the possible examples for the two you have noted with TODO comments. I'll respond with more details later.

docs/csharp/language-reference/compiler-messages/cs8410.md
```diff
@@ -0,0 +1,34 @@
...
+
+The expression inside an `await using` statement must have a `DisposeAsync` method.
+
+## To correct this error
+
+Use `using` instead of `await using`.
+
+## Example
+
+<!-- TODO: The current example is a little bit poor. -->
```
BillWagner, line 22: I think a really good example would be using `StreamReader`, because `StreamWriter` does have a `DisposeAsync` method, but `StreamReader` does not.

BillWagner (commented):
Chose "Comment" rather than "Request changes" so it stays in my queue.

I added a couple suggestions for updates @Youssef1313 What do you you think?

docs/csharp/language-reference/compiler-messages/cs8411.md
```diff
@@ -0,0 +1,34 @@
...
+
+`await foreach` statement operates only on types having a definition of `GetAsyncEnumerator`, such as `IAsyncEnumerable<T>`.
+
+## To correct this error
+
+<!-- TODO -->
+
+## Example
+
+<!-- TODO: The current example is a little bit poor. -->
```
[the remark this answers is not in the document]
Youssef1313: I tried that but unfortunately it doesn't generate cs8411.

image

The compile error in that case is weird. Is that a compiler bug or I'm missing something?

docs/csharp/language-reference/compiler-messages/cs8410.md
Youssef1313: I tried StreamReader but it generated CS8417 instead of CS8410 (Both errors are similar). Not sure when the compiler generates which one.

IEvangelist: Is this supposed to be an example of code that doesn't compile? If so, you'd show an example of something that is errant.

[Here is a SharpLab.io] (https://sharplab.io/#v2:EYLgtghgzgLgpgJwD4AEBMBGAsAKBQBgAIUMA6ASQHkBuXA4jAVlpzowDZi0GB2Q3AN65CIhpxQAOYpwCyEAJYA7ABQBKfjlGEhmraJQBOYkWUA3CAkII4EACaJCAXkKK4Ad0IBlGNYhgASjb2CMoARKGq6gD0UYQAIgD2cFCKAOQwhAAWEKZwhPIwwnoiMYbGhGYWVkEOzq4e3r5gAOoIBYhhEdGx2VDx8lAADglQcACCUACeigDGRcU6xSIAvvOEqzjLQA) showing that the `StreamWriter` can be used with `await using` but the `StreamReader` cannot.

Youssef1313: @IEvangelist Yes the examples are of code that doesn't compile, but it's supposed to generate a **specific** compile error. In this specific one, we need CS8410, but with StreamReader it generates CS8417. I'm not sure why the compiler team made these two different error code while they look just the same for me. It's unclear to me when the compiler would generate CS8410 and when it would generate CS8417.

IEvangelist: Let me see what I can dig up

IEvangelist: Hey @Youssef1313 - try this [SharpLab.io](https://sharplab.io/#v2:EYLgtghgzgLgpgJwD4AEAMACFBGArAbgFgAoEnANiwCYtsB2DEgbxIzdspQA4tKBZCAEsAdgAoAlI2LsMLaTPYoAnFkwA3CAgwI4EACaIMAXgzC4AdwwBJMAAcANoIDGgmACVdBhBKLy2AXxJA0mIUGhsHZ1cPfURmIKA===).

docs/csharp/language-reference/compiler-messages/cs8410.md
```diff
@@ -0,0 +1,34 @@
...
+---
+# Compiler Error CS8410
+
+'type': type used in an asynchronous using statement must be implicitly convertible to 'System.IAsyncDisposable' or implement a suitable 'DisposeAsync' method.
+
+The expression inside an `await using` statement must have a `DisposeAsync` method.
+
+## To correct this error
+
+Use `using` instead of `await using`.
```
IEvangelist, line 18: This is not the correct messaging for this error message. This actually corresponds to `ERR_NoConvToIAsyncDispWrongAsync = 8417`, see [resource](https://github.com/dotnet/roslyn/blob/dbdae0b756975b86f3411c3ec9e01df4afbe1fc8/src/Compilers/CSharp/Portable/CSharpResources.resx#L3004-L3009)

docs/csharp/language-reference/compiler-messages/cs8410.md
```diff
@@ -0,0 +1,34 @@
...
+```csharp
+class C
+{
+    public static async System.Threading.Tasks.Task Main()
+    {
+        await using (Main)
+        {
+        }
+    }
+}
```
IEvangelist, line 33:
```suggestion
using System.Threading.Tasks;

static class Program 
{
    static async Task Main() 
    {
        // error CS8410: 'ImplicitReader': type used in an asynchronous using statement
        // must be implicitly convertible to 'System.IAsyncDisposable' or implement
        // a suitable 'DisposeAsync' method.
        await using var reader = new ImplicitReader();
    }
}

class ImplicitReader
{
}
```

IEvangelist (changes requested):

docs/csharp/language-reference/compiler-messages/cs8400.md
```diff
@@ -0,0 +1,20 @@
...
+ms.date: 07/11/2020
+f1_keywords:
+  - "CS8400"
+helpviewer_keywords:
+  - "CS8400"
+author: Youssef1313
+---
+# Compiler Error CS8400
+
+Feature 'feature' is not available in C# 8.0. Please use language version \<version> or greater.
```
IEvangelist, line 12:
```suggestion
Feature is not available in C# 8.0. Please use language version \<version> or greater.
```

docs/csharp/language-reference/compiler-messages/cs8410.md
```diff
@@ -0,0 +1,34 @@
...
+```
```
IEvangelist, line 41: Consider adding a see also and linking to the Implement DisposeAsync article.

```markdown
## See also

- [Implement a DisposeAsync method](../../../standard/garbage-collection/implementing-disposeasync.md)
```

docs/csharp/language-reference/compiler-messages/cs8410.md
Youssef1313: @IEvangelist, this line isn't referred to as being the message of the error. It just suggests the common fix for this error.

docs/csharp/language-reference/compiler-messages/cs8410.md
Youssef1313: @IEvangelist, Your example shows the error message we want here. But any idea why StreamReader generates a different message?

docs/csharp/language-reference/compiler-messages/cs8400.md
Youssef1313: @IEvangelist I think other existing error messages keeps placeholders like this.

docs/csharp/language-reference/compiler-messages/cs8410.md
Youssef1313: Sure. I even opened https://github.com/dotnet/docs/issues/19440 specifically for this. But looks like I forgot 😄

docs/csharp/language-reference/compiler-messages/cs8411.md
```diff
@@ -0,0 +1,34 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+public class C
+{
+    async Task M(IAsyncEnumerator<int> enumerator)
+    {
+        await foreach (int i in enumerator) { }
+    }
+}
+```
```
IEvangelist, line 34:
```suggestion
```csharp
using System.Collections.Generic;
using System.Threading.Tasks;

class Program
{
    async Task Main(IAsyncEnumerator<int> enumerator)
    {
        // error CS8411: Asynchronous foreach statement cannot operate on variables 
        // of type 'IAsyncEnumerator<T>' because 'IAsyncEnumerator<T>' does not 
        // contain a suitable public instance definition for 'GetAsyncEnumerator'
        await foreach (int i in enumerator) 
        {
        }
    }
}
```
```

File: docs/csharp/language-reference/compiler-messages/cs8410.md (hunk 1 above)
Comment (IEvangelist) [reply docs/csharp/language-reference/compiler-messages/cs8410.md]: The reason is that the `StreamReader` implements `IDisposable`, this goes back to my earlier comment that removing the `await` is not valid for `8410` as a suggestion to fix it. That only applies to 8417.

[See this](https://sharplab.io/#v2:EYLgtghgzgLgpgJwD4AEBMBGAsAKBQBgAIUMBWAbl1xIDZi1iMB2Q3Ab10K8bpQA5idALIQAlgDsAFAEpWOboQ7yF3FAE5iRAG4QEhBHAgATRIQC8hcXADuhAEqGTCGZWUriGgoR16joqAAOAPZQEMAANnAOxqYWVrYAIv7BoRFRjogunO5yOZreuoQAHuaWNvYZztKueV4+hACepfGESYEhYZHRTlluhAC+uIM41AzdpmwDVHgMbSmd6TF6IIQAknMdaezZXCgAzMQALK3JIXAyilM4/UA=).

```csharp
using System;
using System.Threading.Tasks;

static class Program 
{
    static async Task Main() 
    {
        await using var reader = new Reader();
        await using var disposableReader = new DisposableReader();
        
        using var x = new Reader();
        using var y = new DisposableReader();
    }
}

class Reader { }

class DisposableReader : IDisposable
{
    public void Dispose() { }
}
```

docs/csharp/language-reference/compiler-messages/cs8410.md
IEvangelist: [See this please.](https://sharplab.io/#v2:EYLgtghgzgLgpgJwD4AEBMBGAsAKBQBgAIUMBWAbl1xIDZi1iMB2Q3Ab10K8bpQA5idALIQAlgDsAFAEpWOboQ7yF3FAE5iRAG4QEhBHAgATRIQC8hcXADuhAEqGTCGZWUriGgoR16joqAAOAPZQEMAANnAOxqYWVrYAIv7BoRFRjogunO5yOZreuoQAHuaWNvYZztKueV4+hACepfGESYEhYZHRTlluhAC+uIM41AzdpmwDVHgMbSmd6TF6IIQAknMdaezZXCgAzMQALK3JIXAyilM4/UA=)

Youssef1313: Oh got it. Then what you recommend to put under "To correct this error"?

IEvangelist: The only thing that makes sense is to remove both the `await` and the `using`. If the type is neither an implementation of `IDisposable` or `IAsyncDisposable`, I think that's the only logic suggestion that can be made. @BillWagner, do you agree?

docs/csharp/language-reference/compiler-messages/cs8403.md
```diff
@@ -0,0 +1,20 @@
...
+  - "CS8403"
+author: Youssef1313
+---
+# Compiler Error CS8403
+
+Method 'method' with an iterator block must be 'async' to return '{IAsyncEnumerable\<T>|IAsyncEnumerator\<T>}'
+
+## To correct this error
+
+Mark your method with `async` modifier.
```
IEvangelist, line 16:
```suggestion
Mark your method with the `async` modifier.
```

docs/csharp/language-reference/compiler-messages/cs8410.md
BillWagner: Yes, I agree. If a type is not disposable (synchronously or asynchronously), the only fix is to remove the `using`

docs/csharp/language-reference/compiler-messages/cs8410.md
```diff
@@ -0,0 +1,43 @@
...
+        // must be implicitly convertible to 'System.IAsyncDisposable' or implement
+        // a suitable 'DisposeAsync' method.
+        await using var reader = new Example();
+    }
+}
+
+class Example
+{
+}
+```
```
IEvangelist, line 39:
```suggestion
```csharp
using System.Threading.Tasks;

static class Program
{
    static async Task Main()
    {
        // error CS8410: 'Example': type used in an asynchronous using statement
        // must be implicitly convertible to 'System.IAsyncDisposable' or implement
        // a suitable 'DisposeAsync' method.
        await using var example = new Example();
    }
}

class Example
{
}
```
```

File: docs/csharp/language-reference/compiler-messages/cs8410.md (hunk 3 above)
Comment (Youssef1313) [reply docs/csharp/language-reference/compiler-messages/cs8410.md:18]: @BillWagner @IEvangelist

I think we may also add "implement a suitable `DisposeAsync` method" as another option. What do you think?

Comment (IEvangelist) [reply docs/csharp/language-reference/compiler-messages/cs8410.md:18]: Here is an updated version of the [SharpLab.io](https://sharplab.io/#v2:EYLgtghgzgLgpgJwD4AEBMBGAsAKBQBgAIUMBWAbl1xIDZi1iMB2Q3Ab10K8bpQA5idALIQAlgDsAFAEpWOboQ7yF3APSrCAYQDKfACwYiAWkIwAFqKiFLhAHIB5ACqEAZqIAeEYABs4hAO6i5qZmfghwYAD2AG4Q3oSRLoQABigAnMmcKlzpxESxCIThEAAmiIQAvITicP6EAEpwpYgylMrZWdnqdpGEiAiRCJ0qBUVNZQgMVTV1jc0IrcMKS2oaOvrMhCbmNjZunj5+gcHmYRExcQlJqRkrOWl5hKMllgAOkVBevnMTldW1hAAIm8Pl84D8WtI2tkuHdCN1bL1+oM4QQnhBCi8oO9PocIZM/jMgSDcd9xpDodwAL64Gk4agMfGKQh0hnE7GgvHkwogQgASWBHNJcHYwxQAGZiHp2Ti4DJmXSqUA===)

Comment (Youssef1313) [reply docs/csharp/language-reference/compiler-messages/cs8410.md:18]: These are very useful when documenting cs8417. We may even link the two pages as they're a bit related.

File: docs/csharp/language-reference/compiler-messages/cs8401.md [hunk 10]
```diff
@@ -0,0 +1,25 @@
...
+To use '@$' instead of '$@' for an interpolated verbatim string, please use language version '8.0' or greater.
+
+In older versions before C# 8.0, verbatim interpolated strings were only allowed using `$@`. Starting with C# 8.0, you can use `$@` or `@$`.
+
+## To correct this error
+
+Do one of the following options:
+
+- Replace `@$` with `$@`.
+- Upgrade to C# 8.0 or later.
```
IEvangelist, line 21:
```suggestion
Either replace `@$` with `$@`, or upgrade to C# 8.0 or later.
```

This formatting might have made sense if there were more than two options. Instead, make it a simple sentence.

docs/csharp/language-reference/compiler-messages/cs8410.md
```diff
@@ -0,0 +1,43 @@
...
+## To correct this error
+
+Remove the `await using` keywords, or implement a suitable `DisposeAsync` method.
+
+## Example
+
+```csharp
+using System.Threading.Tasks;
+
+static class Program
```
IEvangelist, line 25:
```suggestion
class Program
```

docs/csharp/language-reference/compiler-messages/cs8411.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Example
+
+```csharp
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+class Program
+{
+    async Task Main(IAsyncEnumerator<int> enumerator)
```
IEvangelist, line 28:
```suggestion
    async Task Example(IAsyncEnumerator<int> enumerator)
```

docs/csharp/language-reference/compiler-messages/cs8410.md
Youssef1313: I think it's okay to keep the static keyword? that class contains only a static main method.

docs/csharp/language-reference/compiler-messages/cs8400.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Compiler Error CS8400
+ms.date: 07/11/2020
+f1_keywords:
+  - "CS8400"
+helpviewer_keywords:
+  - "CS8400"
+author: Youssef1313
+---
```
Youssef1313, line 9: I see other compile-messages have ms.assetid in the metadata. Is it required? What is it for? How to get the correct value for it?

docs/csharp/language-reference/compiler-messages/cs8411.md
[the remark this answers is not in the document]
BillWagner: That makes sense, but it's weird. `MoveNext` returns `bool`, not `Task<bool>` (or similar).

Let's revert to the original example.

docs/csharp/language-reference/compiler-messages/cs8400.md
IEvangelist: I believe that the `ms.assetid` is a carry over from old documents that were ported from MSDN days. It is not needed, nor is it applicable for new docs.

docs/csharp/language-reference/compiler-messages/cs8410.md
IEvangelist: But it is not necessary. In fact to be consistent with the templates of a new C# console application, it should be omitted.

BillWagner (approved):
This LGTM now @Youssef1313 

I'll :shipit: now.

BillWagner:
@IEvangelist It looks like all your requested changes have been addressed. Do you want to take a look, and approve and merge if so?

IEvangelist (approved):

PR #19451, opened 2020-07-13
Update async.md

danespinosa:
## Summary

*  Replaced "take care" by tread carefully since in line 226 is the way the document tries to emphasize to be careful around LINQ expressions and I think it is a clearer way to warn the reader about these scenarios.

### DIFF
--- a/docs/csharp/async.md
+++ b/docs/csharp/async.md
@@ -201,7 +201,7 @@ public static async Task<User[]> GetUsersAsync(IEnumerable<int> userIds)
 }
 ```
 
-Although it's less code, take care when mixing LINQ with asynchronous code. Because LINQ uses deferred (lazy) execution, async calls won't happen immediately as they do in a `foreach` loop unless you force the generated sequence to iterate with a call to `.ToList()` or `.ToArray()`.
+Although it's less code, tread carefully when mixing LINQ with asynchronous code. Because LINQ uses deferred (lazy) execution, async calls won't happen immediately as they do in a `foreach` loop unless you force the generated sequence to iterate with a call to `.ToList()` or `.ToArray()`.
 
 ## Important info and advice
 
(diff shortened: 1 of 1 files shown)


docs/csharp/async.md
```diff
@@ -201,7 +201,7 @@ public static async Task<User[]> GetUsersAsync(IEnumerable<int> userIds)
 }
 ```
 
-Although it's less code, take care when mixing LINQ with asynchronous code. Because LINQ uses deferred (lazy) execution, async calls won't happen immediately as they do in a `foreach` loop unless you force the generated sequence to iterate with a call to `.ToList()` or `.ToArray()`.
+Although it's less code, tread carefully when mixing LINQ with asynchronous code. Because LINQ uses deferred (lazy) execution, async calls won't happen immediately as they do in a `foreach` loop unless you force the generated sequence to iterate with a call to `.ToList()` or `.ToArray()`.
```
BillWagner, line 204: I did a little checking, and "tread carefully" won't translate well into other languages. How about this change instead:

```suggestion
Although it's less code, use caution when mixing LINQ with asynchronous code. Because LINQ uses deferred (lazy) execution, async calls won't happen immediately as they do in a `foreach` loop unless you force the generated sequence to iterate with a call to `.ToList()` or `.ToArray()`.
```

BillWagner (changes requested):
Thanks for making this update @danespinosa We appreciate it. I'm suggesting an alternative wording due to likely issues translating the phrase.

If you accept my suggestion, I'll :shipit:

docs/csharp/async.md
danespinosa: Thanks @BillWagner I accepted the suggestions!

BillWagner (approved):

BillWagner:
Thanks @danespinosa I've now merged this.

PR #19514, opened 2020-07-15
Address concerns about compiler magic with indexers

### LINKED ISSUE #19510: Explain why we can't have an Item property when we declare an indexer property with its default name
Nowhere in the official docs does it state how exactly an indexer is born to fruition with the compiler.

Because that's not mentioned, one would greatly be confused when they attempt to create a property named "Item", when they have a named indexer property defined with the default index name (Item).

It would be wise to share with the user base that the way the compiler handles an indexer, is that it creates a property/methods for us, which can be used by languages not supporting indexers directly, that is:

If we have an object `SomeObject`, and it has an indexer with a single `int` argument, in C# if we would write the following: `SomeObject a = new SomeObject(); a[5];`, a language not supporting this would write something like this instead `SomeObject a = new SomeObject(); a.get_items(5);`

TL;DR: Any magic under the hood is worth pointing out.

---
#### Document Details

* ID: 5323b38b-8dc5-8f50-f5e7-b58ca0e88ad7
* Version Independent ID: 3d9a83cd-98e3-fc6e-727b-32c8e9228c69
* Content: [Using Indexers - C# Programming Guide](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/indexers/using-indexers)
* Content Source: [docs/csharp/programming-guide/indexers/using-indexers.md](https://github.com/dotnet/docs/blob/master/docs/csharp/programming-guide/indexers/using-indexers.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

IEvangelist:
## Summary

- Add alert about indexer compiler generated property
- Add details about `Item`
- General clean up / edit pass
- Fix formatting
- Modernize C# bits

Fixes #19510

### Preview

✔️ [Using indexers (C# Programming Guide)](https://review.docs.microsoft.com/en-us/dotnet/csharp/programming-guide/indexers/using-indexers?branch=pr-en-us-19514)


BillWagner (approved):
This LGTM @IEvangelist, pending how you want to address the snippet build errors.

You can :shipit: once that's resolved.

PR #19520, opened 2020-07-15
Delete orphaned images (ML)

gewarren:
Contributes to #19353.


Youssef1313 (changes requested):
I think images under the snippets/image-classification/csharp/assets/images should probably remain the same. They're used in the code.

gewarren:
> I think images under the snippets/image-classification/csharp/assets/images should probably remain the same. They're used in the code.

Good catch!

luisquintanilla (approved):
Thanks @gewarren and @Youssef1313. :shipit:

PR #19511, opened 2020-07-15
Update index.md

IEvangelist:
## Summary

Provide a link to a relevant tutorial that further details default interface methods. 

Fixes #19470

PR #19529, opened 2020-07-16
Fix build warnings

gewarren:
Introduced in #19511.


adegeo (approved):

IEvangelist:
I see what happened... the interfaces link actually worked, and I verified that in the build preview. But the other must not have, and I missed checking that one. Thank you @gewarren 🙏

PR #19584, opened 2020-07-19
Fix tiny typo in async.md

lv1il0s:
## Summary

Fixed tiny typo: 
should only to be used -> should only be used


BillWagner (approved):
Thanks for spotting this and fixing it @lv1il0s 

I'll :shipit: now.

PR #19592, opened 2020-07-20
Update cancellationex13.cs

harold-b:
The assignment to the field `internalToken` in the constructor appears to be superfluous.

## Summary

I removed an unneeded line in the constructor with the intention to improve clarity. The `internalToken` field is assigned again in `DoWork`, where it is actually referenced.


BillWagner (approved):
Thanks for making this change @harold-b 

harold-b:
Thank you, BIll

PR #19622, opened 2020-07-21
Fix typos in databricks commands

usmanmohammed:
## Summary

The destination path of the databricks copy command is invalid and this breaks the copy job.

Also think the release zip filename needs to be changed considering the previous steps used ```publish.zip``` as the filename.
Otherwise, there should be a step stating that the zip file should be renamed to ```mySparkApp.zip```


luisquintanilla (approved):
Thanks for your contribution @usmanmohammed . Looks good to me.

PR #19629, opened 2020-07-22
Revised foreach article

pkulikov:
Two major revisions:

- I've tried to simplify the explanations about the type of an iteration variable and moved it to the separate section
- Moved examples directly to the corresponding reference text


BillWagner (approved):
Thanks @pkulikov 

I really like these improvements. It integrates the new features into the existing behavior very well.

I'll :shipit: now.

PR #19671, opened 2020-07-24
Update dotnet build docs

richardwerkman:
## Summary

The -s flag is not supported for dotnet build as stated earlier in these docs. So it should be removed as an option.


Youssef1313 (changes requested):
Ops, for some reason I saw it as issue not a PR when I opened #19672 😄 

I'd suggest you make an additional change which I have made in #19672

richardwerkman:
@Youssef1313 I didn't see that case of the flag. Updated my PR!

tdykstra (approved):
Thanks @richardwerkman !

PR #19678, opened 2020-07-24
Upgrade Snippets.5000 workflow

adegeo:
- LocateProjects tool adds more support for scenarios regarding _snippets.5000.json_, _global.json_, and deleted files.
- Workflow updated to support changing/adding a _snippets.5000.json_ to trigger the workflow.


BillWagner (approved):
LGTM. Let's :shipit:

PR #19694, opened 2020-07-26
Operators overview: update link to switch expression reference

pkulikov:


BillWagner (approved):
Good catch. Thanks @pkulikov 

I'll :shipit: now.

PR #19695, opened 2020-07-26
Mention that switch expression (operator) is not overloadable

pkulikov:


BillWagner (approved):
Thanks @pkulikov 

I'll :shipit: now.

PR #19700, opened 2020-07-27
Add F1 keywords for combination keywords

davidwengier:
## Summary

In preparation of fixing https://github.com/dotnet/roslyn/issues/46142 we need to add keywords to the combination modifier pages.

Is there anything else I need to do for this to work, or will indexing pick up these automatically?

### DIFF
--- a/docs/csharp/language-reference/keywords/private-protected.md
+++ b/docs/csharp/language-reference/keywords/private-protected.md
@@ -1,6 +1,8 @@
 ---
 title: "private protected - C# Reference"
 ms.date: 11/15/2017
+f1_keywords:
+  - "privateprotected_CSharpKeyword"
 author: "sputier"
 ---
 # private protected (C# Reference)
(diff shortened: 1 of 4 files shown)


davidwengier:
Opened https://github.com/dotnet/roslyn/pull/46343 which will use these new keywords. I'll hold off merging that until this is merged, and published.

BillWagner (approved):
Thanks @davidwengier 

This is all that's needed.

I'll :shipit: now.

davidwengier:
Thanks @BillWagner. Is there any way to track when this goes live, or rather when the search index might be updated?

Otherwise I'll just keep hitting this URL until it works:
https://msdn.microsoft.com/query/dev16.query?appId=Dev16IDEF1&l=EN-US&k=k(privateprotected_CSharpKeyword);k(DevLang-csharp)&rd=true

davidwengier:
Looks like my link works! :)

BillWagner:
@davidwengier We publish once a day, so it should be live now.

PR #19713, opened 2020-07-27
removed reference to DocFX

smaillet-ms:
## Summary
DocFX support for this tags is incomplete and there is confusion on that since it is listed here, implying it is supported. See [DocFX issue ](https://github.com/dotnet/docfx/issues/2328#issuecomment-663845633) for more details.


BillWagner (approved):
Thanks for adding this clarification @smaillet-ms 

I'll :shipit: now

PR #19714, opened 2020-07-27
Snippets add NuGet ENV VAR to automatically trigger restore

adegeo:
@BillWagner
This should automatically trigger nuget restore on Visual Studio MSBUILD processes, at least according to the docs here: https://docs.microsoft.com/en-us/nuget/consume-packages/package-restore#restore-packages-manually-using-visual-studio

Same change that was completed in the samples repo.


BillWagner (approved):
LGTM. I'll :shipit:

PR #19721, opened 2020-07-27
Add Vector<T> breaking change

### LINKED ISSUE #19197: Vector<T> consistently throws NotSupportedException for .NET 5
## Vector<T> consistently throws NotSupportedException for .NET 5

`Vector<T>` is designed to work with only a limited subset of types. These types are currently `byte`, `sbyte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `float`, and `double`. 

When `T` is not one of these supported types, you are meant to get a `NotSupportedException`. However, due to some code paths that exist to better support hardware acceleration, this exception was not always thrown.

### Version introduced

.NET 5

### Old behavior

For certain functions, you would not get a `NotSupportedException` when using an unsupported `T`. For example, `Vector<bool> + Vector<bool>` would return `default` rather than throwing. However, this would only occur for platforms which had no hardware acceleration, such as ARM32, or if hardware acceleration was disabled for the current platform.

### New behavior

Unsupported `T` will now consistently throw `NotSupportException` on all hardware and configurations.

### Reason for change

Outside the explicitly supported types, `Vector<T>` does not behave as expected and the behavior was inconsistent across various platforms and hardware configurations.

### Recommended action

Do not use unsupported `T`. In .NET 5 and prior, the supported types are `byte`, `sbyte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `float`, and `double`.

In a future revision of .NET, this support could be expanded to additional types and the relevant docs should be checked to determine which types are valid for use.

### Category

- Core .NET libraries

### Affected APIs

This change impacts all members of [Vector<T>](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.vector-1?view=net-5.0)

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #19197.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/3.1-5.0?branch=pr-en-us-19721#vector-always-throws-notsupportedexception-for-unsupported-types).

### DIFF
--- a/docs/core/compatibility/3.1-5.0.md
+++ b/docs/core/compatibility/3.1-5.0.md
@@ -1,7 +1,7 @@
 ---
 title: Breaking changes, version 3.1 to 5.0
 description: Lists the breaking changes from version 3.1 to version 5.0 of .NET, ASP.NET Core, and EF Core.
-ms.date: 07/17/2020
+ms.date: 07/27/2020
 ---
 # Breaking changes for migration from version 3.1 to 5.0
 
--- a/docs/core/compatibility/3.1-5.0.md
+++ b/docs/core/compatibility/3.1-5.0.md
@@ -122,11 +122,16 @@ If you're migrating from version 3.1 of .NET Core, ASP.NET Core, or EF Core to v
 
 ## Core .NET libraries
 
+- [Vector<T> always throws NotSupportedException for unsupported types](#vectort-always-throws-notsupportedexception-for-unsupported-types)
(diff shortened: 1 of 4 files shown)


adegeo (approved):
LGTM, I would just double check that change to the icu-globalization-api file renders correctly.

gewarren:
Thanks for the quick review Andy! The ICU change isn't actually visible in the doc. I think those commented out sections are supposed to be used by a breaking change code analyzer.

PR #19728, opened 2020-07-28
DataTable.Columns.Add only works if you add all columns

DanielBailey1001001:
## Summary

The example suggests you can add a column to bypass the security, with no other work needed, but ReadXml then does not import any records because the schema is wrong. This note lets us know to add all columns if we wish this technique to work.

### DIFF
--- a/docs/framework/data/adonet/dataset-datatable-dataview/security-guidance.md
+++ b/docs/framework/data/adonet/dataset-datatable-dataview/security-guidance.md
@@ -33,6 +33,8 @@ If the incoming XML data contains an object whose type is not in this list:
 
 When loading XML into an existing `DataSet` or `DataTable` instance, the existing column definitions are also taken into account. If the table already contains a column definition of a custom type, that type is temporarily added to the allow list for the duration of the XML deserialization operation.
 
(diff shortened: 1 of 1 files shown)


docs/framework/data/adonet/dataset-datatable-dataview/security-guidance.md
```diff
@@ -33,6 +33,8 @@ If the incoming XML data contains an object whose type is not in this list:
 
 When loading XML into an existing `DataSet` or `DataTable` instance, the existing column definitions are also taken into account. If the table already contains a column definition of a custom type, that type is temporarily added to the allow list for the duration of the XML deserialization operation.
 
+Note: Once you add columns to a DataTable, ReadXml will not read the schema from the Xml, and if the schema does not match it will also not read in the records, so you will need to add all the columns yourself to use this method.
```
IEvangelist, line 36:
```suggestion
> [!NOTE]
> Once you add columns to a `DataTable`, `ReadXml` will not read the schema from the XML, and if the schema does not match it will also not read in the records, so you will need to add all the columns yourself to use this method.
```

IEvangelist (changes requested):
Hi 👋 @DanielBailey1001001
Thank you so much for posting this issue. We really appreciate it.

Update note to be an actual alert style note, enclose code fence methods, and capitalize XML.

IEvangelist (approved):

PR #19731, opened 2020-07-28
Add config for build system

BillWagner:
The roslyn samples must be built using Visual Studio.

This unblocks:

#19214
#19194
#19193
#19190
#19189
#19188
#19146

### DIFF
--- a/samples/snippets/csharp/roslyn-sdk/Tutorials/MakeConstTestProject/MakeConstTestProject/Program.cs
+++ b/samples/snippets/csharp/roslyn-sdk/Tutorials/MakeConstTestProject/MakeConstTestProject/Program.cs
@@ -10,7 +10,8 @@ namespace MakeConstTestProject
             int j = 2;
             int k = i + j;
 
-            int x = "abc";
+            // uncomment for analyzer test:
+            //int x = "abc";
 
             object s = "abc";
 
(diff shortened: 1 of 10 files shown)


samples/snippets/csharp/roslyn-sdk/Tutorials/MakeConstTestProject/MakeConstTestProject/Program.cs
```diff
@@ -10,7 +10,8 @@ static void Main(string[] args)
             int j = 2;
             int k = i + j;
 
-            int x = "abc";
+            // uncomment for analyzer test:
+            //int x = "abc";
```
Youssef1313, line 14: Isn't the intention of the analyzer is to use const when possible?
In that case, shouldn't this be `string x` instead of `int x`?

Unless I'm missing something, I think this line would generate an error instead of a warning from the analyzer.

BillWagner: The point of this line is that you must write analyzers so that they work correctly on code that doesn't compile cleanly. An analyzer that relies on the syntax model will be called when the code is syntactically valid, but has errors found during the semantic phase (as this error will be).

mamccrea (approved):

PR #19734, opened 2020-07-28
Mentioned one more use of the when keyword

pkulikov:


BillWagner (approved):
Thanks @pkulikov 

I'll :shipit: now.

Issue #19743, opened 2020-07-29
C# language reference consolidation: lambda expressions


pkulikov:
I suggest to move [Lambda expressions](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/lambda-expressions) article from the programming guide to the language reference section, under the "C# operators and expressions" TOC node: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/ (`=>` operator is already there)

Furthermore, [How to use lambda expressions in a query (C# Programming Guide)](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-use-lambda-expressions-in-a-query) may be merged into the main article about lambda expressions. Moreover, that article already contains the section about queries: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/lambda-expressions#lambdas-with-the-standard-query-operators

After that, a quite outdated [Anonymous functions](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/anonymous-functions) article can be removed/redirected to the lambda expressions article.

@BillWagner what do you think?

pkulikov:
@tdykstra, @BillWagner if this is a good suggestion to realize, please assign this issue to me.

BillWagner:
Thanks @pkulikov I do like this idea, and I've assigned it to you.

PR #19744, opened 2020-07-29
Moved and revised expressions reference

pkulikov:
This revision is the full rewrite, please review carefully.

The follow up of #17456
Possible continuation of this is #19743


docs/csharp/language-reference/operators/index.md
```diff
@@ -1,6 +1,7 @@
 ---
-title: "C# operators - C# reference"
-ms.date: 04/28/2020
+title: "C# operators and expressions - C# reference"
+description: "Learn about C# operators and expressions, operator precedence and operator associativity"
```
Youssef1313, line 3: nit

```suggestion
description: "Learn about C# operators and expressions, operator precedence, and operator associativity"
```

docs/csharp/language-reference/operators/snippets/Overview.cs
```diff
@@ -0,0 +1,33 @@
...
+            c = a >= 100 ? b : c / 10;
+            
+            string s;
+            char l;
+            s = "String literal";
+            a = s.Length;
+            l = s[3];
+            
+            List<int> numbers;
+            numbers = new List<int>(new[] { 1, 2, 3 });
```
Youssef1313, line 26: Unless there is a reason you have in mind for having them as two separate statements, I'd combine them.

```suggestion
            var numbers = new List<int>(new[] { 1, 2, 3 });
```

Youssef1313 (changes requested):

docs/csharp/language-reference/operators/index.md
```diff
@@ -10,11 +11,31 @@ helpviewer_keywords:
...
 
-In an [expression](../../programming-guide/statements-expressions-operators/expressions.md), operator precedence and associativity determine the order in which the operations are performed. You can use parentheses to change the order of evaluation imposed by operator precedence and associativity.
+The simplest C# expressions are literals (which may be [integer](../builtin-types/integral-numeric-types.md#integer-literals) or [real](../builtin-types/floating-point-numeric-types.md#real-literals) numbers, [characters](../builtin-types/char.md#literals) or [strings](../builtin-types/reference-types.md#the-string-types), [Boolean](../builtin-types/bool.md) `true` or `false`) and simple names (which usually denote variables). With use of operators, you get more complex expressions. Operator [precedence](#operator-precedence) and [associativity](#operator-associativity) determine the order in which the operations in an expression are performed. You can use parentheses to change the order of evaluation imposed by operator precedence and associativity.
```
Youssef1313, line 18:
```suggestion
The simplest C# expressions are literals (which may be [integer](../builtin-types/integral-numeric-types.md#integer-literals) or [real](../builtin-types/floating-point-numeric-types.md#real-literals) numbers, [characters](../builtin-types/char.md#literals) or [strings](../builtin-types/reference-types.md#the-string-type), [Boolean](../builtin-types/bool.md) `true` or `false`) and simple names (which usually denote variables). With use of operators, you get more complex expressions. Operator [precedence](#operator-precedence) and [associativity](#operator-associativity) determine the order in which the operations in an expression are performed. You can use parentheses to change the order of evaluation imposed by operator precedence and associativity.
```

docs/csharp/language-reference/operators/snippets/Overview.cs
pkulikov: I didn't want to mix declarations and assignments, but I think now that actually would be fine.

docs/csharp/language-reference/operators/index.md
```diff
@@ -10,11 +11,31 @@ helpviewer_keywords:
...
+- [Lambda expressions](../../programming-guide/statements-expressions-operators/lambda-expressions.md) that allow you to create anonymous functions
+- [Query expressions](../keywords/query-keywords.md) that allow you to use query capabilities directly in C#
+
+Typically, an expression produces a result. A [`void`](../builtin-types/void.md) method invocation is an example of an expression that doesn't produce a result. Any [method invocation](member-access-operators.md#invocation-expression-) is also an example of a *statement expression*, which is an expression that you can use as a statement, that is, in the following form:
+
+```csharp
+statement-expression;
+```
+
+Other statement expressions are [assignment](assignment-operator.md), [await](await.md), [new object instantiation](new-operator.md), [increment](arithmetic-operators.md#increment-operator-), and [decrement](arithmetic-operators.md#decrement-operator---) expressions. When a statement expression is used as a statement, its result, if any, is discarded.
```
tdykstra, line 36: The items identified as "statement expressions" link to articles that call them operators, which is likely to be confusing.  It might be worth making this list into a bulleted list with an expression example for each.

docs/csharp/language-reference/operators/index.md
```diff
@@ -10,11 +11,31 @@ helpviewer_keywords:
...
+C# provides a number of operators. Many of them are supported by the [built-in types](../builtin-types/built-in-types.md) and allow you to perform basic operations with values of those types. For example, [arithmetic operators](arithmetic-operators.md) perform arithmetic operations with numeric operands and [Boolean logical operators](boolean-logical-operators.md) perform logical operations with [`bool`](../builtin-types/bool.md) operands. Others can be used in a broader scope. For example, you can use [conditional operator `?:`](conditional-operator.md) or [`switch` expression](switch-expression.md) to select which operand expression to evaluate based on a Boolean condition or a matched pattern, respectively.
```
tdykstra, line 16: The last two sentences seem unnecessary and are hard to follow without examples.  I would consider omitting them.

On the other hand, I think the mention of operator overloading is important here and shouldn't be dropped.

docs/csharp/language-reference/operators/index.md
```diff
@@ -10,11 +11,31 @@ helpviewer_keywords:
...
 
-In an [expression](../../programming-guide/statements-expressions-operators/expressions.md), operator precedence and associativity determine the order in which the operations are performed. You can use parentheses to change the order of evaluation imposed by operator precedence and associativity.
+The simplest C# expressions are literals (which may be [integer](../builtin-types/integral-numeric-types.md#integer-literals) or [real](../builtin-types/floating-point-numeric-types.md#real-literals) numbers, [characters](../builtin-types/char.md#literals) or [strings](../builtin-types/reference-types.md#the-string-type), [Boolean](../builtin-types/bool.md) `true` or `false`) and simple names (which usually denote variables). With use of operators, you get more complex expressions. Operator [precedence](#operator-precedence) and [associativity](#operator-associativity) determine the order in which the operations in an expression are performed. You can use parentheses to change the order of evaluation imposed by operator precedence and associativity.
```
tdykstra, line 18: char instances aren't really "characters".
"simple names" also seems like inexact language and leaves the reader wondering what "simple names" are that don't denote variables.

It isn't necessary at this point to enumerate all of the kinds of literals or names -- perhaps simplify this intro text as something like "The simplest C# expressions are literals and variable names. These can be combined into complex expressions by using operators."

docs/csharp/language-reference/operators/index.md
```diff
@@ -10,11 +11,31 @@ helpviewer_keywords:
...
+
+[!code-csharp[expression examples](snippets/Overview.cs#Expressions)]
+
+Additionally, C# provides the following forms of expressions:
+
+- [Interpolated string expressions](../tokens/interpolated.md) that provide convenient syntax to create formatted strings
+- [Lambda expressions](../../programming-guide/statements-expressions-operators/lambda-expressions.md) that allow you to create anonymous functions
+- [Query expressions](../keywords/query-keywords.md) that allow you to use query capabilities directly in C#
+
+Typically, an expression produces a result. A [`void`](../builtin-types/void.md) method invocation is an example of an expression that doesn't produce a result. Any [method invocation](member-access-operators.md#invocation-expression-) is also an example of a *statement expression*, which is an expression that you can use as a statement, that is, in the following form:
```
tdykstra, line 30: "...in the following form:" -- after the preceding examples, the difference here seems to be the absence of an = sign, but "any method invocation" would also include those that have a return value.  Perhaps "is also an example of" should be "may be used as"?

docs/csharp/language-reference/operators/index.md
```diff
@@ -10,11 +11,31 @@ helpviewer_keywords:
...
+The simplest C# expressions are literals (which may be [integer](../builtin-types/integral-numeric-types.md#integer-literals) or [real](../builtin-types/floating-point-numeric-types.md#real-literals) numbers, [characters](../builtin-types/char.md#literals) or [strings](../builtin-types/reference-types.md#the-string-type), [Boolean](../builtin-types/bool.md) `true` or `false`) and simple names (which usually denote variables). With use of operators, you get more complex expressions. Operator [precedence](#operator-precedence) and [associativity](#operator-associativity) determine the order in which the operations in an expression are performed. You can use parentheses to change the order of evaluation imposed by operator precedence and associativity.
+
+In the following code, the right-hand operands of assignments are examples of expressions:
+
+[!code-csharp[expression examples](snippets/Overview.cs#Expressions)]
+
+Additionally, C# provides the following forms of expressions:
```
tdykstra, line 24: Or make it sound less like a comprehensive list -- "Here are some other kinds of expressions that C# provides:"

tdykstra (commented):
This looks good @pkulikov! I just had some suggestions for simplifying the language and adding examples.

docs/csharp/language-reference/operators/index.md
BillWagner, line 16: I agree with @tdykstra. This is a hard overview to get right. I think you're running into the same concerns we had trying to organize the operators into groups. How about this:

```suggestion
C# provides a number of operators. Many of them are supported by the [built-in types](../builtin-types/built-in-types.md) and allow you to perform basic operations with values of those types. For example, [arithmetic operators](arithmetic-operators.md) perform arithmetic operations with numeric operands and [Boolean logical operators](boolean-logical-operators.md) perform logical operations with [`bool`](../builtin-types/bool.md) operands. Other operators do not form obvious groups. For example, you can use [conditional operator `?:`](conditional-operator.md) or [`switch` expression](switch-expression.md) to select which operand expression to evaluate based on a Boolean condition or a matched pattern, respectively. Certain operators can be [overloaded](operator-overloading.md). With operator overloading, you can specify the operator behavior for the operands of a user-defined type. 
```

docs/csharp/language-reference/operators/index.md
BillWagner: I agree. Let's avoid making this a comprehensive list.

docs/csharp/language-reference/operators/index.md
BillWagner: This is one of the more confusing areas of the language spec (because `void` really isn't a type). Here's how I would rework this paragraph:

1. Typically, an expression produces a result.
1. A *statement expression* can be used as a statement, meaning the result (if any), is discarded.
1. Invoking a method whose return type is `void` must be used as a statement, because it does not produce a result.
1. The only expressions that can be used as statements are:
  - method invocation
  - `await`
  - object creation
  - assignment
  - pre and post increment and decrement

I think reworking these two paragraph along those lines will be more clear.

docs/csharp/language-reference/operators/index.md
BillWagner: I think reworking these two paragraph per my previous comment may make examples unnecessary. What do you think @tdykstra ?

docs/csharp/language-reference/operators/lambda-operator.md
```diff
@@ -38,7 +38,7 @@ An expression body definition has the following general syntax:
 member => expression;
 ```
 
-where `expression` is a valid expression. The return type of `expression` must be implicitly convertible to the member's return type. If the member's return type is `void` or if the member is a constructor, a finalizer, or a property `set` accessor, `expression` must be a [*statement expression*](~/_csharplang/spec/statements.md#expression-statements), which can be of any type.
+where `expression` is a valid expression. The return type of `expression` must be implicitly convertible to the member's return type. If the member's return type is `void` or if the member is a constructor, a finalizer, or a property or indexer `set` accessor, `expression` must be a [*statement expression*](~/_csharplang/spec/statements.md#expression-statements), which can be of any type.
```
BillWagner, line 41: I would change the last clause to make it clear what "which" refers to:

```suggestion
where `expression` is a valid expression. The return type of `expression` must be implicitly convertible to the member's return type. If the member's return type is `void` or if the member is a constructor, a finalizer, or a property or indexer `set` accessor, `expression` must be a [*statement expression*](~/_csharplang/spec/statements.md#expression-statements). The statement expression can return any type, because it will be discarded.
```

BillWagner (changes requested):
Thanks @pkulikov 

This is good work.  I had a few suggestions, mostly around the same areas @tdykstra mentioned. Once we resolve those, I think this is ready to merge.

docs/csharp/language-reference/operators/index.md
tdykstra: Some parts of your outline after "The only expressions that can be used as statements are:" don't seem obvious to me without examples to explain them:

  - method invocation: this seems obvious
  - `await`: I don't think of "await" as an expression but more like just a keyword you put in front of a method invocation.
  - object creation: I usually think of this as `var x = new Poco();` -- if you discard the result from `new Poco()` what have you accomplished.
  - assignment: same issue here -- assignment I think of as `a = b`, `b` being the expression and `a = b` being the statement; not clear why the later should be called an expression statement
  - pre and post increment and decrement - this one seems obvious

pkulikov:
@tdykstra and @BillWagner thank you for the feedback. I've made considerable reworking: made the text more focused and added examples. Please review again.

docs/csharp/language-reference/operators/index.md
pkulikov: @tdykstra @BillWagner thank you for your thoughts and suggestions. However, as this topic looks to be more confusing than I thought, I prefer to remove it at all instead of going into more explanations (which would make the article harder to follow). All in all, expression statements are discussed in the [Statements](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/statements) article. That appears to be a more proper place for them.

docs/csharp/language-reference/operators/index.md
```diff
@@ -10,11 +11,45 @@ helpviewer_keywords:
...
+
+The simplest C# expressions are literals (for example, [integer](../builtin-types/integral-numeric-types.md#integer-literals) and [real](../builtin-types/floating-point-numeric-types.md#real-literals) numbers) and names of variables. You can combine them into complex expressions by using operators. Operator [precedence](#operator-precedence) and [associativity](#operator-associativity) determine the order in which the operations in an expression are performed. You can use parentheses to change the order of evaluation imposed by operator precedence and associativity.
+
+In the following code, examples of expressions are at the right-hand side of assignments:
+
+[!code-csharp[expression examples](snippets/Overview.cs#Expressions)]
+
+Typically, an expression produces a result and can be included into another expression. A [`void`](../builtin-types/void.md) method call is an example of an expression that doesn't produce a result. It can be used only as a [statement](../../programming-guide/statements-expressions-operators/statements.md), as the following example shows:
```
tdykstra, line 32:
```suggestion
Typically, an expression produces a result and can be included in another expression. A [`void`](../builtin-types/void.md) method call is an example of an expression that doesn't produce a result. It can be used only as a [statement](../../programming-guide/statements-expressions-operators/statements.md), as the following example shows:
```

docs/csharp/language-reference/operators/index.md
```diff
@@ -10,11 +11,45 @@ helpviewer_keywords:
...
+
+```csharp
+Console.WriteLine("Hello, world!");
+```
+
+Here are some other kinds of expressions that C# provides:
+
+- [Interpolated string expressions](../tokens/interpolated.md) that provide convenient syntax to create formatted strings:
+
+  [!code-csharp-interactive[interpolated string](snippets/Overview.cs#InterpolatedString)]
```
tdykstra, line 42: If you used the new syntax here, I believe you could highlight the line(s) referred to in the text, i.e., in this case the `var message =` line.

tdykstra (approved):
Looks good!

docs/csharp/language-reference/operators/index.md
Youssef1313: If highlighting by line number will be added, an *access-by-line.txt* file will be needed.

pkulikov: @tdykstra thanks for a suggestion, but I'll leave it as it is now: the snippet is really small.

BillWagner (approved):

BillWagner:
This is great @pkulikov / @tdykstra 

I'll :shipit: now.

PR #18456, opened 2020-05-15
Unit testing order

IEvangelist:
## Summary

- Wrote new article "Order unit tests" for MSTest, xUnit, and NUnit
- Remove duplicate TOC entry, "Unit testing" and "Overview" were the same
- Add `xref` where possible in "Selective unit tests" doc
- Allow TOC to scope to test framework
- Add `dotnet/samples` repo to *.openpublishing.publish.config.json*
- Corrected redirect link, https://xunit.net/
- Add zone pivots def to *docs/zone-pivot-groups.yml*
- Use zone pivots for different test frameworks
  - Overview, Run select unit tests, and Order unit tests
- Break out long `dotnet test --filter` commands into code blocks for readability
- Normalize *docfx.json* file
- Fixed heading that use "ing"
- Add `dotnet *` links where missing
- Add `displayName` to help with filtering
- Rewrite VB article per #16129

Fixes #18432, fixes #16129, closes #8117 (as this was explored as part of this effort)

### Preview links

✔️ [Unit testing overview](https://review.docs.microsoft.com/en-us/dotnet/core/testing/?branch=pr-en-us-18456)
✔️ [Run selective unit tests](https://review.docs.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests?branch=pr-en-us-18456)
✔️ [Order unit tests](https://review.docs.microsoft.com/en-us/dotnet/core/testing/order-unit-tests?branch=pr-en-us-18456)

PR #19754, opened 2020-07-30
Remove samples repository reference

Youssef1313:
This setting used to exist before. I'm not 100% sure about what it does, but I guess it makes the live docs dependent on the live branch of the samples repo instead of master.

I'm also curious why docs is now dependent again on samples?

cc: @adegeo

### DIFF
--- a/.openpublishing.publish.config.json
+++ b/.openpublishing.publish.config.json
@@ -79,11 +79,6 @@
       "path_to_root": "iot-samples",
       "url": "https://github.com/MicrosoftDocs/dotnet-iot-assets",
       "branch": "master"
-    },
-    {
-      "path_to_root": "dotnet-samples",
-      "url": "https://github.com/dotnet/samples",
-      "branch": "master"
     }
   ],
   "docs_build_engine": {
--- a/docs/core/testing/order-unit-tests.md
+++ b/docs/core/testing/order-unit-tests.md
@@ -25,7 +25,7 @@ With MSTest, tests are automatically ordered by their test name.
 > [!NOTE]
 > A test named `Test14` will run before `Test2` even though the number  `2` is less than `14`. This is because, test name ordering uses the text name of the test.
 
-:::code language="csharp" source="~/dotnet-samples/csharp/unit-testing/MSTest.Project/ByAlphabeticalOrder.cs":::
+:::code language="csharp" source="snippets/order-unit-tests/csharp/MSTest.Project/ByAlphabeticalOrder.cs":::
 
 :::zone-end
 :::zone pivot="xunit"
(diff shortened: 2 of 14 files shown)


BillWagner:
@Youssef1313 

> I guess it makes the live docs dependent on the live branch of the samples repo instead of master.

Yes, that's exactly what it does.

I agree that we *should* be able to remove the dependency on the `dotnet/samples` repo at this time. Do you want to update this PR to try that instead? (If the build is clean, we should be ready to merge it).

Youssef1313:
The build won't be clean. There is a sample from samples repo that is referenced here in docs.
Should I move the sample?
cc: @Evangelist

IEvangelist:
> The build won't be clean. There is a sample from samples repo that is referenced here in docs.
> Should I move the sample?
> cc: @IEvangelist

Yeah, we added that a while back. The only concern that I have with removing it, and copying code back over is the concern of having code that is copied in two places. Ideally we should avoid that.

Youssef1313:
I totally agree that code shouldn't be in 2 places. That's harder to maintain. But on the other hand, there was a work to avoid dependency on the other repo.
What about adding the sample here and deleting it from the other repo? Is there any concerns about that?

adegeo:
@IEvangelist Duplication is fine when it comes to full samples. Full samples should be clean and perfect code that a user can download and use. Snippets shouldn't be. This is just a problem we have to live with if we want samples for the samples browser. 😦 

@Youssef1313 Can you elaborate on which sample dependency is specifically being talked about?

Youssef1313:
@adegeo https://github.com/dotnet/samples/tree/master/csharp/unit-testing

It's used in https://github.com/dotnet/docs/blob/9511adb1d654883c948b39363174f774ece7d444/docs/core/testing/order-unit-tests.md

adegeo:
Ahh yes. We need to clean that up. @IEvangelist 😄

Youssef1313:
@adegeo What's the correct action to do here?

IEvangelist:
> @adegeo What's the correct action to do here?

Hi @Youssef1313,

According to @adegeo - the correct action is to remove the mapping altogether, and then fix references to it. The fixes would be to create a local _snippets_ directory next to the articles, and copy the projects, and source code local to the docs repo. Does that make sense?

IEvangelist:
> Ahh yes. We need to clean that up. @IEvangelist 😄

We talked about this before, when I initially brought it up to the team and got permission to add it this way. I'm fine with it's removal too.

Youssef1313:
> 
> 
> > @adegeo What's the correct action to do here?
> 
> Hi @Youssef1313,
> 
> According to @adegeo - the correct action is to remove the mapping altogether, and then fix references to it. The fixes would be to create a local _snippets_ directory next to the articles, and copy the projects, and source code local to the docs repo. Does that make sense?

So you're okay with the duplication?
I'll finish it soon once I get home.

BillWagner:
"ok with the duplication" is the best way to phrase it. It's the least painful of all the practical solutions.

Ideally, we'd like cross-repo build enhancements that would enable a single code source for larger samples and included in article snippets. We've asked for that feature, and it's under investigation.

BillWagner:
@Youssef1313 I'm going over older PRs, and I'm wondering if this one should be closed. Is it still relevant?  

/cc @adegeo

adegeo:
@BillWagner OK I looked this over. It looks like we had removed reference to the samples repo in my migration pr but with #18456 addressing unit test articles, it was added it back in.

 Considering we don't want samples and docs mixed, we should copy the unit test snippets into the local snippets folder and reference them there and remove the link to the sample repo. The file using these samples is `core\testing\order-unit-tests.md`

Perhaps we should scrap this PR and open an issue to discuss/track?

cc @IEvangelist

IEvangelist:
I'm still conflicted on this, but I'm happy to support anyone's strong opinions. :)

tdykstra:
I like the suggestion from @adegeo 

>copy the unit test snippets into the local snippets folder and reference them there and remove the link to the sample repo.

BillWagner:
@Youssef1313 

This is still in draft mode. Do you want to complete it, or do you want to close it?

Youssef1313:
Sorry for the delay. I copied the samples from dotnet/samples and removed the dependency on dotnet/samples repository.

BillWagner (approved):
Thanks @Youssef1313 

I'll :shipit: now.

PR #19757, opened 2020-07-30
Update to Schannel TLS registry entry

felgar73:
Changed "Enable" dword value option to "1" instead of "0xFFFFFFFF" as this is not correct.

## Summary

Describe your changes here.


IEvangelist:
Hi @felgar73 - thank you for this PR.

Hey @karelz - wanted to verify with you that this is accurate, if so I'll merge this PR. Thanks

wfurt (approved):
LGTM. 
I think any non-zero value may work but I did not test it.
This looks more consistent and does work for sure. 

Since we are touching this, can you add TLS 1.3 as well. 
SSL 2.0 will generally not work. For rest, perhaps we should add note that this is subject to OS support e.g. adding the entry may or may not yield the given protocol.

BillWagner (approved):
This LGTM. I'll :shipit: now.

PR #19789, opened 2020-08-02
Delete unused images

Youssef1313:
@gewarren, FYI. (Related to #19777)


Youssef1313:
Does the build detect if this breaks something?

gewarren:
> Does the build detect if this breaks something?

Yes, build would flag a broken image link. How did you find these images?

Youssef1313:
> How did you find these images?

@gewarren I wrote a small tool that gets the images from every subfolder, and searches if the image is referenced in any markdown or not.

nishanil (approved):
Looks good to me!

PR #19813, opened 2020-08-04
Update security-guidance.md

hihayak:


stevestein (approved):
Per email with Hiroyuki

PR #19825, opened 2020-08-04
Normalize source files and fix encoding

sharwell:
* Normalize text files in the repository

        git add --all --renormalize

* Fix encoding for **\*.vb** and **\*.cs** (should be `utf-8-bom`)


Youssef1313:
Can this PR be also extended to make sure cs and vb files ends with an empty line (i.e. follows the `insert_final_newline = true` editorconfig option). (If the docs team accepts that)

cc: @tdykstra

Youssef1313:
Can this PR be also extended to make sure cs and vb files ends with an empty line (i.e. follows the `insert_final_newline = true` editorconfig option). (If the docs team accepts that)

cc: @tdykstra

tdykstra (approved):
Thanks @sharwell, and @Youssef1313 for initiating the discussion. There are CI warnings for missing proj files but we can let those go for this PR.

tdykstra:
>Can this PR be also extended to make sure cs and vb files ends with an empty line

I don't think the benefit of that is worth doing a bulk change.

tdykstra:
@sharwell Your PRs in the samples repo warn against squash-merging -- does that apply here as well?

sharwell:
@tdykstra it's not as big an issue here because it's only a few dozen files (there it was over 20,000). If a merge commit is allowed, it would still be my preference.

PR #19826, opened 2020-08-04
Fixed typo on sessionCookieLifetime

martyncoup:
Existing variable was assigning to configuration, not Configuration.

## Summary

The existing variable used lowercase for configuration instead of starting with an uppercase 'C' for the sessionCookieLifetime value in the sample.


BillWagner (approved):
Thanks for spotting this and making the fix @martyncoup 

I'll :shipit: now.

PR #19830, opened 2020-08-04
F1 keywords

gewarren:
- Add missing F1 keywords according to https://ceapex.visualstudio.com/Engineering/_workitems/edit/228223.
- Remove MSDN note from F# docs.


docs/csharp/language-reference/builtin-types/bool.md
```diff
@@ -2,8 +2,12 @@
...
 ms.date: 11/26/2019
 f1_keywords: 
-  - "bool"
-  - "bool_CSharpKeyword"
+  - bool
+  - bool_CSharpKeyword
+  - "true"
+  - "false"
+  - true_CSharpKeyword
+  - false_CSharpKeyword
```
adegeo, line 10: Just curious, is there some sort of format or syntax to these keywords? Why is true_CSharpKeyword special? I see all sorts of other ones too.

adegeo (approved):

docs/csharp/language-reference/builtin-types/bool.md
gewarren: I'm not sure why it's special, I'm basically just following a spreadsheet that said which keywords were missing, I think based on keywords that existed in the old MSDN topic and keywords that users pressed F1 on. But I'm not 100% sure.

Youssef1313: They're defined in Roslyn repository AFAIK.

Youssef1313: https://github.com/dotnet/roslyn/blob/master/src/VisualStudio/CSharp/Impl/LanguageService/CSharpHelpContextService.cs

docs/csharp/language-reference/operators/arithmetic-operators.md
```diff
@@ -11,6 +11,9 @@ f1_keywords:
   - "%_CSharpKeyword"
   - "+_CSharpKeyword"
   - "-_CSharpKeyword"
+  - "%=_CSharpKeyword"
+  - "*=_CSharpKeyword"
+  - "_CSharpKeyword"
```
Youssef1313, line 16: This one looks strange to me. Based on my understanding, [this line](https://github.com/dotnet/roslyn/blob/master/src/VisualStudio/CSharp/Impl/LanguageService/CSharpHelpContextService.cs#L50) in Roslyn shouldn't have `text` being empty. There should be something before the underscore I think.

docs/fsharp/language-reference/arrays.md
```diff
@@ -5,9 +5,6 @@ ms.date: 05/16/2016
 ---
 # Arrays
 
-> [!NOTE]
-> The API reference link will take you to MSDN.  The docs.microsoft.com API reference is not complete.
```
Youssef1313, line 9: The links doesn't even take to MSDN, they take to docs.microsoft.com with 404.

I think the note may be updated instead of completely deleted if there is a known issue with F# API reference?

gewarren: @cartermp It seems like you're aware of the broken links. Since the links no longer go to MSDN (they redirect to docs, but 404), I removed the notes. Should I temporarily add a reference to https://fsharp.github.io/fsharp-core-docs/?

docs/csharp/language-reference/operators/arithmetic-operators.md
gewarren: In this case, I think the language service is not picking up anything where the user's cursor is. I think it's good to have a fallback page so they at least land in the vicinity of where they might find help. But it might be better to add this keyword here instead: https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/operators/index.md. Thoughts @BillWagner ?

docs/fsharp/language-reference/arrays.md
cartermp: Yeah, I think that's a good alternative for now.

docs/csharp/language-reference/operators/arithmetic-operators.md
Youssef1313: If Roslyn generates `_CSharpKeyword` without anything before the underscore when the cursor isn't on a keyword or something like that, I think it should be fixed in Roslyn not here. Nothing before the underscore is meaningless, and Roslyn shouldn't generate it like that.

@gewarren Do you know who from Roslyn team we can ask?

gewarren: @Youssef1313 I don't know who to contact in the Roslyn team. You could file an issue there. Either way, I don't think you can guarantee that Roslyn will always capture text, and it's easy and doesn't hurt to add this here.

Youssef1313: Then probably the c# landing page would be a better place?

gewarren: I put the keyword at the root of the language reference docs, so /language-reference/index.md.

Issue #19836, opened 2020-08-05
Rendering issue in Application compatibility in the .NET Framework


hecongy:
**Issue description**

There should be .NET framework version lists both in [Runtime changes](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/application-compatibility#runtime-changes) section and [Retargeting changes](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/application-compatibility#retargeting-changes) section  in the doc [Application compatibility in the .NET Framework](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/application-compatibility). But there are rendering issues.

**Target framework**

Check the .NET target framework(s) being used, and include the version number(s).

- [ ] .NET Core
- [x] .NET Framework
- [ ] .NET Standard

If using the .NET Core SDK, include `dotnet --info` output. If using .NET Framework without the .NET Core SDK, include info from Visual Studio's **Help** > **About Microsoft Visual Studio** dialog.

<details>
<summary><strong>dotnet --info output</strong> or <strong>About VS info</strong></summary>

```console
<replace>
```
</details>

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 4ee6a407-16fe-6537-e587-99908c9a8248
* Version Independent ID: 920fcce6-df42-ff50-dae2-cd7110a8a2f3
* Content: [Runtime and retargeting changes - .NET Framework](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/application-compatibility)
* Content Source: [docs/framework/migration-guide/application-compatibility.md](https://github.com/dotnet/docs/blob/master/docs/framework/migration-guide/application-compatibility.md)
* Product: **dotnet-framework**
* Technology: **dotnet-appcompat**
* GitHub Login: @gewarren
* Microsoft Alias: **gewarren**

adegeo:
@gewarren The version selector include isn't working.

gewarren:
Shoot, it's broken here too: https://docs.microsoft.com/en-us/dotnet/core/compatibility/breaking-changes#by-version. @hecongy Thanks for letting us know!

PR #19844, opened 2020-08-05
Try to fix multi-selectors

gewarren:
Related to #19836. Replaced include files in /core/ folder (since they're only used in one place). Doesn't fix the problem, but probably still a good idea to remove the includes.


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

gewarren:
Thanks Bill!

PR #19845, opened 2020-08-05
C# reference TOC: moved the Types node one level up

pkulikov:
- The Types TOC node in the C# reference seems to be worth moving out of the Keywords node as it's more than about just keywords (there are articles covering nullable value types, unmanaged types, nullable reference types)

- Also, removed "C#" from titles as those are redundant in the C# guide TOC


BillWagner (approved):
I like this change @pkulikov 

I'll :shipit: now.

PR #19846, opened 2020-08-05
switch expression: fixed the name of a pattern

pkulikov:
Throughout the docs, this is called "constant pattern":
- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/switch#constant-pattern
- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/is#constant-pattern

As well as in the language spec proposal: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-7.0/pattern-matching#constant-pattern


BillWagner (approved):
Thanks @pkulikov 

I'll :shipit: now.

PR #19874, opened 2020-08-06
Use when filter rather than if else throw.

ILMTitan:
## Summary

Change the documentation to use a when filter.
The when filter has existed since C# 6 and is the more correct way to handle this situation.


IEvangelist (approved):

PR #19879, opened 2020-08-06
Cryptography.Oid property setters

### LINKED ISSUE #19680: Breaking Change Doc: System.Security.Cryptography.Oid is now functionally init-only
## System.Security.Cryptography.Oid is now functionally init-only

The `System.Security.Cryptography.Oid` class, which is used to represent ASN.1 Object Identifier values and their "friendly" names, was previously fully mutable -- which was often a surprise or simply overlooked.  Now the property setters will throw `PlatformNotSupportedException` when a call attempts to change the value after it has already been assigned.

### Version introduced

.NET 5 (Preview 8)

### Old behavior

The property setters on `System.Security.Cryptography.Oid` can always be used to change the `Value` or `FriendlyName` properties.

### New behavior

The property setters can be called to initialize the property value.  Once the property has a value, either from a constructor parameter or a previous call to the property setter, the property setter will throw `PlatformNotSupportedException`.

### Reason for change

Allow the reuse of `Oid` objects as part of return values in public API to reduce object allocation profiles and avoid the need for creating temporary "defensive" copies when `Oid` values are used as inputs.

### Recommended action

Avoid using the `Oid` property setters other than for object initialization.  Instead of changing the value on an existing object, use a new instance to represent a new value.

### Category

Cryptography

### Affected APIs

* [Oid.FriendlyName](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.oid.friendlyname)
* [Oid.Value](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.oid.value)

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #19680

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/3.1-5.0?branch=pr-en-us-19879#systemsecuritycryptographyoid-is-functionally-init-only).

### DIFF
--- a/docs/core/compatibility/3.1-5.0.md
+++ b/docs/core/compatibility/3.1-5.0.md
@@ -168,6 +168,14 @@ If you're migrating from version 3.1 of .NET Core, ASP.NET Core, or EF Core to v
 
 ***
 
+## Cryptography
+
+- [System.Security.Cryptography.Oid is functionally init-only](#systemsecuritycryptographyoid-is-functionally-init-only)
+
+[!INCLUDE [cryptography-oid-init-only](../../../includes/core-changes/cryptography/5.0/cryptography-oid-init-only.md)]
+
+***
+
 ## Globalization
 
 - [StringInfo and TextElementEnumerator are now UAX29-compliant](#stringinfo-and-textelementenumerator-are-now-uax29-compliant)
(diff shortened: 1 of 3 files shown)


tdykstra (approved):
LGTM

gewarren:
Thanks Tom!

PR #19904, opened 2020-08-07
Unique assembly per identity per context

lbargaoanu:
I think the idea is to load a unique assembly per identity per context, so it's only relevant if the same version is already loaded in the load-from context, not the load context.


adegeo:
@GrabYourPitchforks and @steveharter Can you please comment on this suggested docs change? Thanks!

adegeo:
Sent email to @steveharter

steveharter (approved):
The change looks correct.

It is possible to have both the assembly already loaded from the default load context as well as a different assembly (in a different path) loaded via LoadFrom -- they will be different Assembly instances even though they may have the same identity.

However, when using LoadFrom more than once with the same identity, the first instance is returned even though they have different paths.

One nit not covered here is that calling LoadFrom using the path from an already loaded assembly from the default load context returns the same Assembly instance (the same physical file is not loaded again). Probably not worth calling out.

adegeo:
Thanks for the contribution @lbargaoanu

PR #19912, opened 2020-08-08
A/W suffix for P/Invokes

### LINKED ISSUE #19681: Remove A/W probing for P/Invokes on non-Windows platforms
## Remove A/W probing for P/Invokes on non-Windows platforms

The Windows [function convention](https://docs.microsoft.com/windows/win32/intl/conventions-for-function-prototypes) of having a A/W suffix is now only respected on the Windows platform. Both the CoreCLR and Mono runtimes have continued this probing on non-Windows platform where the A/W suffix is not a convention. With this change the automatic addition of a A/W suffix is no longer performed on non-Windows platforms. The semantics of both runtimes on the Windows platform remain unchanged.

See https://github.com/dotnet/runtime/issues/33246 for additional details.

### Version introduced

.NET 5.0

### Old behavior

During export discovery for P/Invokes an A/W prefix was added to the export name when probing. This was done on all platforms.

### New behavior

During export discovery for P/Invokes an A/W prefix was added to the export name when probing only on the Windows platform.

### Reason for change

This change was made to simplify the cross-platform probing. It is overhead that should not be incurred by the developer given non-Windows platforms don't contain this semantic.

### Recommended action

In order to mitigate this change, the user should manually add the desired prefix on non-Windows platforms.

In the example below, if the export on non-Windows was `FooW` the following declaration would previously suffice:

```CSharp
[DllImport(...)]
extern static void Foo();
```

It is now recommended the developer update the declaration with the desired prefix:

```CSharp
[DllImport(...)]
extern static void FooW();
```

### Category

- Interop

### Affected APIs

[DllImportAttribute](https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.dllimportattribute).

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #19681.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/3.1-5.0?branch=pr-en-us-19912#no-aw-suffix-probing-on-non-windows-platforms).

### DIFF
--- a/docs/core/compatibility/3.1-5.0.md
+++ b/docs/core/compatibility/3.1-5.0.md
@@ -188,7 +188,12 @@ If you're migrating from version 3.1 of .NET Core, ASP.NET Core, or EF Core to v
 
 ## Interop
 
+- [No A/W suffix probing on non-Windows platforms](#no-aw-suffix-probing-on-non-windows-platforms)
 - [Built-in support for WinRT is removed from .NET](#built-in-support-for-winrt-is-removed-from-net)
+
+[!INCLUDE [function-suffix-pinvoke](../../../includes/core-changes/interop/5.0/function-suffix-pinvoke.md)]
+
+***
 
 [!INCLUDE [built-in-support-for-winrt-removed](~/includes/core-changes/interop/5.0/built-in-support-for-winrt-removed.md)]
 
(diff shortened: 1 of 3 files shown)


AaronRobinsonMSFT (approved):

BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

PR #19917, opened 2020-08-08
F# printf "%%" prints a percent sign

Tarmil:
## Summary

Document that F# printf "%%" prints a percent sign.


cartermp (approved):
Thanks!

PR #19921, opened 2020-08-09
Fix dotnet-sdk-2.1 version recommendation: SDK

dagood:
I think this is wrong and the SDK package should use `<sdk version>`:

> - `dotnet-sdk-[major].[minor]` - Installs the latest sdk for specific runtime
>   - **Version:** \<runtime version>

It looks like this started with a mistake in my suggestion at https://github.com/dotnet/docs/pull/13196#discussion_r301634357, so I doubt there was any reason for it to be `<runtime version>`.

/cc @tmds @omajid


omajid (approved):
Thanks for catching and fixing this!

tdykstra (approved):
Thanks!

PR #19925, opened 2020-08-09
Moved the article about lambda expressions to the language reference

pkulikov:
Contributes to #19743
No revisions, only links/titles/TOC updates.


IEvangelist (approved):
Looks good to me, we'll :shipit: - thank you 🙏. I'll let @BillWagner merge it, so that he can at least have eyes on this.

PR #19937, opened 2020-08-10
Updated DefaultOperator.cs

### LINKED ISSUE #19919: Code snippet throws ArgumentNullException for a positive scenario
The code snippet for **Default Literal** section uses a **Display** method to show the output of the **InitializeArray** method. This works fine for all the examples mentioned in the code snippet however, the examples fail to test the path when length parameter for InitializeArray is negative. As per the code in InitializeArray, the method should return a **default value** when length is negative. However, since the default value for an **array is null**, the method returns a null value which then causes the Display method to throw a **System.ArgumentNullException**.

The case I used was `Display(InitializeArray<int>(-1));`.

I propose the following change to Display method to handle this scenario:
`static void Display<T>(T[] values) =>Console.WriteLine(values == null ? "null" : $"[ {string.Join(", ", values)} ]");`
The InitializeArray method which handles a negative length parameter by returning a default value. 

---
#### Document Details

* ID: 08344c62-6194-1057-e409-35d65fcab2cd
* Version Independent ID: 53f08a66-f8b6-7953-d260-e52694934613
* Content: [default value expressions - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/default)
* Content Source: [docs/csharp/language-reference/operators/default.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/operators/default.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

akashmanoj213:
## Summary

Updated InitializeArray() method to throw ArgumentOutOfRangeException when length parameter is negative.

Fixes #19919

### DIFF
--- a/docs/csharp/language-reference/operators/snippets/shared/DefaultOperator.cs
+++ b/docs/csharp/language-reference/operators/snippets/shared/DefaultOperator.cs
@@ -39,7 +39,7 @@ namespace operators
             {
                 if (length < 0)
                 {
-                    return default;
+                    throw new ArgumentOutOfRangeException(nameof(length), "Array length must be nonnegative.");
                 }
 
                 var array = new T[length];


BillWagner (approved):
This LGTM @akashmanoj213 

I'll :shipit: now.

Thanks for helping us improve and making your first contribution to docs! 🎆

PR #19940, opened 2020-08-10
Set ms.topic

tdykstra:
Set ms.topic for how-to docs in /standard and /framework

### DIFF
--- a/docfx.json
+++ b/docfx.json
@@ -186,6 +186,7 @@
         "docs/azure/**/**.md": "conceptual",
         "docs/core/tutorials/**.md": "tutorial",
         "docs/csharp/language-reference/**/**.md": "language-reference",
+        "docs/framework/**/*how-to*.md": "how-to",
         "docs/framework/additional-apis/**/**.md": "managed-reference",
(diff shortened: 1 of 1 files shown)


docfx.json
```diff
@@ -186,9 +186,11 @@
         "docs/azure/**/**.md": "conceptual",
         "docs/core/tutorials/**.md": "tutorial",
         "docs/csharp/language-reference/**/**.md": "language-reference",
+        "docs/framework/**/*how-to*.md": "how-to",
```
Youssef1313, line 189: It's a really edge case, but what if "how" part of the file name was actually the end of another word. Wouldn't it be safer to remove the asterisk before "how"?

tdykstra: >Is there a way to define an optional dash before "how"?

If you find such a way I'm happy to add it, but we don't need 100% accuracy, if there are exceptions they can be overridden in the doc itself.  This just establishes a default setting, which without it is "conceptual" for everything.

BillWagner (approved):
This LGTM @tdykstra 

I'll :shipit:

Issue #13682, opened 2019-08-01
[Code Gallery] Migrate Async Sample: Accessing the Web Walkthrough


rpetrusha:
## [Code Gallery] Migrate Async Sample: Accessing the Web Walkthrough

**Sample URL**

<https://code.msdn.microsoft.com/Async-Sample-Accessing-the-9c10497f>

**Sample information**

14,179 downloads, a 5.0 rating (with 3 reviewers)

**Usage**

- docs\csharp\language-reference\keywords\await.md 
- docs\csharp\programming-guide\concepts\async\how-to-extend-the-async-walkthrough-by-using-task-whenall.md
- docs\csharp\programming-guide\concepts\async\how-to-extend-the-async-walkthrough-by-using-task-whenall.md  
- docs\csharp\programming-guide\concepts\async\task-asynchronous-programming-model.ms
- docs\csharp\programming-guide\concepts\async\walkthrough-accessing-the-web-by-using-async-and-await.md
- docs\visual-basic\language-reference\modifiers\async.md  
- docs\visual-basic\language-reference\operators\await-operator.md  
- docs\visual-basic\programming-guide\concepts\async\how-to-extend-the-async-walkthrough-by-using-task-whenall.md 
- docs\visual-basic\programming-guide\concepts\async\walkthrough-accessing-the-web-by-using-async-and-await.md
- docs\visual-basic\programming-guide\concepts\async\index.md

Issue #13680, opened 2019-08-01
[Code Gallery] Migrate Async Sample: Asynchronous Programming with Async and Await


rpetrusha:
## [Code Gallery] Migrate Async Sample: Asynchronous Programming with Async and Await

**Sample URL**

<https://code.msdn.microsoft.com/Async-Sample-Example-from-9b9f505c>

**Sample information**

The sample has 42,474 downloads and a 4-star rating (10 reviewers).

**Usage**

- docs\csharp\programming-guide\concepts\async\task-asynchronyous-programming-model.md
- docs\csharp\programming-guide\concepts\async\task-asynchronous-programming-model.md
- docs\visual-basic\programming-guide\concepts\async\index.md
- docs\visual-basic\programming-guide\concepts\async\index.md

**Migration Information**
- [x] **C# sample**: to be migrated to *samples/async-and-await/cs*  
- [x] **Visual Basic sample**: migrated to *samples/async-and-await/vb* by dotnet/samples#1380.

rskar-git:
@rpetrusha

I have done a migration of this for visual basic - project is named AsyncFirstExampleVB.

It is placed under samples\concepts\asynchronous-programming\vb

Pull request: https://github.com/dotnet/samples/pull/1380

rskar-git:
@rpetrusha

I have made the requested changes. I am new to github, and hope you've been aware by it - please let me know if there's something else I need to do to in github after making changes. Thx!

PR #19970, opened 2020-08-11
Update tutorial content, modernize WPF app and C#.

### LINKED ISSUE #19474: Race Condition
I believe this code has a race condition because resultsTextBox.Text is set by multiple threads in the DisplayResult method. Please either correct, let me know why I'm wrong, or at least notate in the article that there is a race condition so it won't encourage bad practices and leave people wondering. 

Thank you!
---
#### Document Details

* ID: 702427b0-0bcc-1152-75a5-c909717723c0
* Version Independent ID: 0dbb92ca-4166-ceaf-1083-1280703a9c15
* Content: [How to extend the async walkthrough by using Task.WhenAll (C#)](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/how-to-extend-the-async-walkthrough-by-using-task-whenall)
* Content Source: [docs/csharp/programming-guide/concepts/async/how-to-extend-the-async-walkthrough-by-using-task-whenall.md](https://github.com/dotnet/docs/blob/master/docs/csharp/programming-guide/concepts/async/how-to-extend-the-async-walkthrough-by-using-task-whenall.md)
* Product: **dotnet-csharp**
* Technology: **csharp-async**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

IEvangelist:
## Summary

Addresses the concerns of a race condition, as well as potential app crash.

- Updated corresponding samples, relies on #dotnet/samples#3819
- Remove MSDN links, replace with sample browser
- Heavy edits to tutorial flow and content
- Simplify tutorials
- Modernize C#

Fixes #19474

## Preview links

✔️ [Tutorial: Access the web with async and await (C#)](https://review.docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await?branch=pr-en-us-19970)
✔️ [Extend the async tutorial using Task.WhenAll (C#)](https://review.docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/how-to-extend-the-async-walkthrough-by-using-task-whenall?branch=pr-en-us-19970)

### DIFF
--- a/docs/csharp/programming-guide/concepts/async/how-to-extend-the-async-walkthrough-by-using-task-whenall.md
+++ b/docs/csharp/programming-guide/concepts/async/how-to-extend-the-async-walkthrough-by-using-task-whenall.md
@@ -1,7 +1,7 @@
 ---
 title: "How to extend the async walkthrough by using Task.WhenAll (C#)"
 description: Learn how to improve performance of the async solution in C# by using Task.WhenAll. This method asynchronously awaits multiple asynchronous operations.
-ms.date: 08/11/2020
+ms.date: 08/12/2020
 ms.assetid: f6927ef2-dc6c-43f8-bc82-bbeac42de423
 ---
 
@@ -161,33 +161,7 @@ Navigate to the `SumPageSizesAsync` method.
 
 1. Finally, wrap the updates to the results textbox and the call to <xref:System.Diagnostics.Stopwatch.Stop?displayProperty=nameWithType> in the <xref:System.Windows.Threading.Dispatcher.BeginInvoke%2A?displayProperty=nameWithType>. Be sure to `await` the call.
 
-```csharp
-await Dispatcher.BeginInvoke(() =>
-{
-    stopwatch.Stop();
-
-    _resultsTextBox.Text += $"\nTotal bytes returned:  {total:#,#}";
-    _resultsTextBox.Text += $"\nElapsed time:          {stopwatch.Elapsed}\n";
-});
-```
-
-The updated version of the `SumPageSizesAsync` method should be as follows:
-
-```csharp
-async Task SumPageSizesAsync()
-{
-    var stopwatch = new Stopwatch();
-    stopwatch.Start();
-
-    IEnumerable<Task<int>> downloadTasksQuery =
-        from url in _urlList
-        select ProcessURLAsync(url, _client);
-
-    Task<int>[] downloadTasks = downloadTasksQuery.ToArray();
-
-    int[] lengths = await Task.WhenAll(downloadTasks);
-    int total = lengths.Sum();
-
+    ```csharp
     await Dispatcher.BeginInvoke(() =>
     {
         stopwatch.Stop();
@@ -195,8 +169,34 @@ async Task SumPageSizesAsync()
         _resultsTextBox.Text += $"\nTotal bytes returned:  {total:#,#}";
         _resultsTextBox.Text += $"\nElapsed time:          {stopwatch.Elapsed}\n";
     });
-}
-```
+    ```
+
+    The updated version of the `SumPageSizesAsync` method should be as follows:
+
+    ```csharp
+    async Task SumPageSizesAsync()
+    {
+        var stopwatch = new Stopwatch();
+        stopwatch.Start();
+
+        IEnumerable<Task<int>> downloadTasksQuery =
+            from url in _urlList
+            select ProcessURLAsync(url, _client);
+
+        Task<int>[] downloadTasks = downloadTasksQuery.ToArray();
+
+        int[] lengths = await Task.WhenAll(downloadTasks);
+        int total = lengths.Sum();
+
+        await Dispatcher.BeginInvoke(() =>
+        {
+            stopwatch.Stop();
+
+            _resultsTextBox.Text += $"\nTotal bytes returned:  {total:#,#}";
+            _resultsTextBox.Text += $"\nElapsed time:          {stopwatch.Elapsed}\n";
+        });
+    }
+    ```
 
 ## Test the Task.WhenAll solutions
 
--- a/docs/csharp/programming-guide/concepts/async/how-to-extend-the-async-walkthrough-by-using-task-whenall.md
+++ b/docs/csharp/programming-guide/concepts/async/how-to-extend-the-async-walkthrough-by-using-task-whenall.md
@@ -1,433 +1,332 @@
 ---
 title: "How to extend the async walkthrough by using Task.WhenAll (C#)"
 description: Learn how to improve performance of the async solution in C# by using Task.WhenAll. This method asynchronously awaits multiple asynchronous operations.
-ms.date: 07/20/2015
+ms.date: 08/11/2020
 ms.assetid: f6927ef2-dc6c-43f8-bc82-bbeac42de423
 ---
-# How to extend the async walkthrough by using Task.WhenAll (C#)
 
-You can improve the performance of the async solution in [Walkthrough: Accessing the Web by Using async and await (C#)](./walkthrough-accessing-the-web-by-using-async-and-await.md) by using the <xref:System.Threading.Tasks.Task.WhenAll%2A?displayProperty=nameWithType> method. This method asynchronously awaits multiple asynchronous operations, which are represented as a collection of tasks.
+# Extend the async tutorial using Task.WhenAll (C#)
 
-You might have noticed in the walkthrough that the websites download at different rates. Sometimes one of the websites is very slow, which delays all the remaining downloads. When you run the asynchronous solutions that you build in the walkthrough, you can end the program easily if you don't want to wait, but a better option would be to start all the downloads at the same time and let faster downloads continue without waiting for the one that’s delayed.
+You can improve the performance of the async solution in [Tutorial: Access the web with async and await (C#)](walkthrough-accessing-the-web-by-using-async-and-await.md) by using the <xref:System.Threading.Tasks.Task.WhenAll%2A?displayProperty=nameWithType> method. This method asynchronously awaits multiple asynchronous operations, which are represented as a collection of tasks.
 
-You apply the `Task.WhenAll` method to a collection of tasks. The application of `WhenAll` returns a single task that isn’t complete until every task in the collection is completed. The tasks appear to run in parallel, but no additional threads are created. The tasks can complete in any order.
+You might have noticed in the tutorial that the websites download at different rates. Sometimes one of the websites is slow, which delays all the remaining downloads. When you run the asynchronous solutions that you build in the walkthrough, you can end the program easily if you don't want to wait, but a better option would be to start all the downloads at the same time and let faster downloads continue without waiting for the one that's delayed.
+
+You apply the `Task.WhenAll` method to a collection of tasks. The application of `WhenAll` returns a single task that isn't complete until every task in the collection is completed. The tasks appear to run in parallel, but no additional threads are created. The tasks can complete in any order.
 
 > [!IMPORTANT]
-> The following procedures describe extensions to the async applications that are developed in [Walkthrough: Accessing the Web by Using async and await (C#)](./walkthrough-accessing-the-web-by-using-async-and-await.md). You can develop the applications by either completing the walkthrough or downloading the code from [Developer Code Samples](https://code.msdn.microsoft.com/Async-Sample-Accessing-the-9c10497f).
+> The following procedures describe extensions to the async applications that are developed in [Tutorial: Access the web with async and await (C#)](walkthrough-accessing-the-web-by-using-async-and-await.md). You can develop the applications by either completing the tutorial or downloading the code from [Code samples browser](https://docs.microsoft.com/samples/dotnet/samples/async-and-await-cs).
 >
-> To run the example, you must have Visual Studio 2012 or later installed on your computer.
+> To run the example, you must have Visual Studio 2019 or later installed on your computer.
 
-### To add Task.WhenAll to your GetURLContentsAsync solution
+## Update the DisplayResults method
 
-1. Add the `ProcessURLAsync` method to the first application that's developed in [Walkthrough: Accessing the Web by Using async and await (C#)](./walkthrough-accessing-the-web-by-using-async-and-await.md).
+Open the *MainWindow.xaml.cs* file, and navigate to the `DisplayResults` method. If you downloaded the code from [Code samples browser](https://docs.microsoft.com/samples/dotnet/samples/async-and-await-cs), open the *SerialAsyncExample* project.
 
-    - If you downloaded the code from  [Developer Code Samples](https://code.msdn.microsoft.com/Async-Sample-Accessing-the-9c10497f), open the AsyncWalkthrough project, and then add `ProcessURLAsync` to the MainWindow.xaml.cs file.
+The `DisplayResults` method used string concatenation to append the `url` and the corresponding download size as an interpolated string to the `_resultsTextBox.Text`. Apply the following updates.
 
-    - If you developed the code by completing the walkthrough, add `ProcessURLAsync` to the application that includes the `GetURLContentsAsync` method. The MainWindow.xaml.cs file for this application is the first example in the "Complete Code Examples from the Walkthrough" section.
-
-    The `ProcessURLAsync` method consolidates the actions in the body of the `foreach` loop in `SumPageSizesAsync` in the original walkthrough. The method asynchronously downloads the contents of a specified website as a byte array, and then displays and returns the length of the byte array.
-
-    ```csharp
-    private async Task<int> ProcessURLAsync(string url)
-    {
-        var byteArray = await GetURLContentsAsync(url);
-        DisplayResults(url, byteArray);
-        return byteArray.Length;
-    }
-    ```
-
-2. Comment out or delete the `foreach` loop in `SumPageSizesAsync`, as the following code shows.
-
-    ```csharp
-    //var total = 0;
-    //foreach (var url in urlList)
-    //{
-    //    byte[] urlContents = await GetURLContentsAsync(url);
-
-    //    // The previous line abbreviates the following two assignment statements.
-    //    // GetURLContentsAsync returns a Task<T>. At completion, the task
-    //    // produces a byte array.
-    //    //Task<byte[]> getContentsTask = GetURLContentsAsync(url);
-    //    //byte[] urlContents = await getContentsTask;
-
-    //    DisplayResults(url, urlContents);
-
-    //    // Update the total.
-    //    total += urlContents.Length;
-    //}
-    ```
-
-3. Create a collection of tasks. The following code defines a [query](../linq/index.md) that, when executed by the <xref:System.Linq.Enumerable.ToArray%2A> method, creates a collection of tasks that download the contents of each website. The tasks are started when the query is evaluated.
-
-    Add the following code to method `SumPageSizesAsync` after the declaration of `urlList`.
-
-    ```csharp
-    // Create a query.
-    IEnumerable<Task<int>> downloadTasksQuery =
-        from url in urlList select ProcessURLAsync(url);
-
-    // Use ToArray to execute the query and start the download tasks.
-    Task<int>[] downloadTasks = downloadTasksQuery.ToArray();
-    ```
-
-4. Apply `Task.WhenAll` to the collection of tasks, `downloadTasks`. `Task.WhenAll` returns a single task that finishes when all the tasks in the collection of tasks have completed.
-
-    In the following example, the `await` expression awaits the completion of the single task that `WhenAll` returns. The expression evaluates to an array of integers, where each integer is the length of a downloaded website. Add the following code to `SumPageSizesAsync`, just after the code that you added in the previous step.
-
-    ```csharp
-    // Await the completion of all the running tasks.
-    int[] lengths = await Task.WhenAll(downloadTasks);
-
-    //// The previous line is equivalent to the following two statements.
-    //Task<int[]> whenAllTask = Task.WhenAll(downloadTasks);
-    //int[] lengths = await whenAllTask;
-    ```
-
-5. Finally, use the <xref:System.Linq.Enumerable.Sum%2A> method to calculate the sum of the lengths of all the websites. Add the following line to `SumPageSizesAsync`.
-
-    ```csharp
-    int total = lengths.Sum();
-    ```
-
-### To add Task.WhenAll to the HttpClient.GetByteArrayAsync solution
-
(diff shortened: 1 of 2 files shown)


docs/csharp/programming-guide/concepts/async/how-to-extend-the-async-walkthrough-by-using-task-whenall.md
```diff
@@ -1,438 +1,338 @@
...
+
+You apply the `Task.WhenAll` method to a collection of tasks. The application of `WhenAll` returns a single task that isn't complete until every task in the collection is completed. The tasks appear to run in parallel, but no additional threads are created. The tasks can complete in any order.
 
 > [!IMPORTANT]
-> The following procedures describe extensions to the async applications that are developed in [Walkthrough: Accessing the Web by Using async and await (C#)](./walkthrough-accessing-the-web-by-using-async-and-await.md). You can develop the applications by either completing the walkthrough or downloading the code from [Developer Code Samples](https://code.msdn.microsoft.com/Async-Sample-Accessing-the-9c10497f).
+> The following procedures describe extensions to the async applications that are developed in [Tutorial: Access the web with async and await (C#)](walkthrough-accessing-the-web-by-using-async-and-await.md). You can develop the applications by either completing the tutorial or downloading the code from [Code samples browser](https://docs.microsoft.com/samples/dotnet/samples/async-and-await-cs).
```
Youssef1313, line 17: I think the old link refers to https://github.com/dotnet/docs/issues/13682, while the new link refers to https://github.com/dotnet/docs/issues/13680, which are different code samples.

docs/csharp/programming-guide/concepts/async/how-to-extend-the-async-walkthrough-by-using-task-whenall.md
```diff
@@ -1,438 +1,338 @@
...
-    //    //byte[] urlContents = await getContentsTask;
+```csharp
+Task DisplayResultsAsync(string url, byte[] content) =>
+    Dispatcher.BeginInvoke(() =>
+        _resultsTextBox.Text += $"{url,-60} {content.Length,10:#,#}\n")
+                .Task;
+```
 
-    //    DisplayResults(url, urlContents);
+The use of `Dispatcher` is required to ensure that all UI updates are correctly dispatched to the UI thread. Later, you use  that queues work on the thread pool.
```
Youssef1313, line 39: Wow, never heard of the Dispatcher before. That's really a great improvement.

docs/csharp/programming-guide/concepts/async/how-to-extend-the-async-walkthrough-by-using-task-whenall.md
```diff
@@ -1,438 +1,338 @@
...
-### To test the Task.WhenAll solutions
-
-- For either solution, choose the F5 key to run the program, and then choose the **Start** button. The output should resemble the output from the async solutions in [Walkthrough: Accessing the Web by Using async and await (C#)](./walkthrough-accessing-the-web-by-using-async-and-await.md). However, notice that the websites appear in a different order each time.
-
-## Example
-
-The following code shows the extensions to the project that uses the `GetURLContentsAsync` method to download content from the web.
+1. Finally, wrap the updates to the results textbox and the call to <xref:System.Diagnostics.Stopwatch.Stop?displayProperty=nameWithType> in the <xref:System.Windows.Threading.Dispatcher.BeginInvoke%2A?displayProperty=nameWithType>. Be sure to `await` the call.
 
 ```csharp
```
Youssef1313, line 164: The content under this list item needs to be indented.

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
```diff
@@ -1,379 +1,371 @@
...
 
-For more information about the Async feature, see [Asynchronous Programming with async and await (C#)](./index.md).
+For more information about async feature, see [Asynchronous programming with async and await (C#)](index.md).
 
-This walkthrough starts with a synchronous Windows Presentation Foundation (WPF) application that sums the number of bytes in a list of websites. The walkthrough then converts the application to an asynchronous solution by using the new features.
+This tutorial starts with a synchronous Windows Presentation Foundation (WPF) application that sums the number of bytes in a list of websites. The walkthrough then converts the application to an asynchronous solution by using the async features.
 
-If you don't want to build the applications yourself, you can download [Async Sample: Accessing the Web Walkthrough (C# and Visual Basic)](https://code.msdn.microsoft.com/Async-Sample-Accessing-the-9c10497f).
+If you don't want to build the applications yourself, you can download [Async Sample: Asynchronous programming with async and await in C#](https://docs.microsoft.com/samples/dotnet/samples/async-and-await-cs).
```
Youssef1313, line 17: Same comment about the different samples applies here.

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
```diff
@@ -1,379 +1,371 @@
...
 
 > [!NOTE]
-> To run the examples, you must have Visual Studio 2012 or newer and the .NET Framework 4.5 or newer installed on your computer.
+> To run the examples, you must have Visual Studio 2019 (v16.7.0) or newer and the .NET 5.0 or newer installed on your computer.
```
Youssef1313, line 20:
```suggestion
> To run the examples, you must have Visual Studio 2019 (v16.7.0) or newer and .NET 5.0 or newer installed on your computer.
```

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
```diff
@@ -1,379 +1,371 @@
...
+    using System.Windows;
     ```
 
 ## Create a synchronous app
 
-1. In the design window, MainWindow.xaml, double-click the **Start** button to create the `startButton_Click` event handler in MainWindow.xaml.cs.
-
-2. In MainWindow.xaml.cs, copy the following code into the body of `startButton_Click`:
+1. In the design window, *MainWindow.xaml*, double-click the **Start** button to create the click event handler in *MainWindow.xaml.cs*.
+1. Select the <kbd>F2</kbd> key to rename the handler to `OnStartButtonClick`.
```
Youssef1313, line 98: Based on [this Visual Studio doc](https://docs.microsoft.com/en-us/visualstudio/ide/reference/rename?view=vs-2019), the rename shortcut is CTRL+R, CTRL+R

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
```diff
@@ -1,379 +1,371 @@
...
+    - `SumPageSizes`, which gets a list of webpage URLs from the `_urlList` field and then calls `ProcessUrl` and `DisplayResults` to process each URL.
+    - `ProcessUrl`, which downloads the contents of each URL and returns the contents as a byte array.
+    - `DisplayResults`, which displays the number of bytes in the byte array for each URL.
 
-    Copy the following four methods, and then paste them under the `startButton_Click` event handler in MainWindow.xaml.cs:
+    Copy the following code, and paste it above the `OnStartButtonClick` event handler in *MainWindow.xaml.cs*:
 
     ```csharp
-    private void SumPageSizes()
+    readonly IEnumerable<string> _urlList = new string[]
```
Youssef1313, line 121:
```suggestion
    private readonly IEnumerable<string> _urlList = new string[]
```

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
```diff
@@ -1,379 +1,371 @@
...
+    Copy the following methods, and paste them below the `OnStartButtonClick` event handler in *MainWindow.xaml.cs*:
 
-            // Update the total.
-            total += urlContents.Length;
-        }
+    ```csharp
+    void SumPageSizes()
+    {
+        var stopwatch = new Stopwatch();
+        stopwatch.Start();
```
Youssef1313, line 138: nit: these can be combined into one line using the Stopwatch.StartNew static method.

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
```diff
@@ -1,379 +1,371 @@
...
-            "https://msdn.microsoft.com/library/ff730837.aspx"
-        };
-        return urls;
+        stopwatch.Stop();
+        _resultsTextBox.Text += $"\nTotal bytes returned:  {total:#,#}";
+        _resultsTextBox.Text += $"\nElapsed time:          {stopwatch.Elapsed}\n";
     }
 
-    private byte[] GetURLContents(string url)
+    int ProcessUrl(string url)
```
Youssef1313, line 147: Access modifier?

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
```diff
@@ -1,379 +1,371 @@
...
     > [!NOTE]
-    > As you follow the steps in this walkthrough, several compiler errors appear. You can ignore them and continue with the walkthrough.
+    > As you follow the steps in this tutorial, several compiler errors appear. You can ignore them and continue with the tutorial.
 
-     Change the method that's called in the third line of `GetURLContents` from `GetResponse` to the asynchronous, task-based <xref:System.Net.WebRequest.GetResponseAsync%2A> method.
+     Change the method that's called in `ProcessUrl` from `GetResponse` to the asynchronous, task-based <xref:System.Net.WebRequest.GetResponseAsync%2A> method.
 
     ```csharp
-    using (WebResponse response = webReq.GetResponseAsync())
+    using WebResponse response = webReq.GetResponseAsync();
```
Youssef1313, line 200: Very nice to see the new C# features are used more and more in docs. I've seen developers that write new C# code that looks 10-year-old 😄.

Youssef1313 (commented):
Are there parallel VB docs for these?

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
IEvangelist: Have you tried <kbd>F2</kbd>? That works in both Visual Studio and VS Code, not sure about <kbd>Ctrl+R</kbd>?

IEvangelist:
> Are there parallel VB docs for these?

Indeed there are, I started updating them and the samples - but it was taking a long time. https://docs.microsoft.com/dotnet/visual-basic/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await. I might circle back around.

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
IEvangelist: I prefer to omit the default access modifiers.

docs/csharp/programming-guide/concepts/async/how-to-extend-the-async-walkthrough-by-using-task-whenall.md
IEvangelist: The old link was referring to a single method in an old sample source code project. I'm updating this to based on extended a migrated tutorial, and extending that specific tutorial.

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
Youssef1313: The dotnet/runtime [coding style document](https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/coding-style.md) mentions:

> We always specify the visibility, even if it's the default (e.g. private string _foo not string _foo). Visibility should be the first modifier (e.g. public abstract not abstract public).

There is also a roslyn analyzer (IDE0040) for that.

If the analyzer is enabled by default, it may be better to include the modifier. If it's not enabled by default, I'll be okay with either adding it or omitting it.

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
Youssef1313: I'm not sure, but I think <kbd>F2</kbd> is a ReSharper thing. Is there a way to open Visual Studio with the default settings but without also losing them later?

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
```diff
@@ -1,379 +1,369 @@
...
-    }
-
-    private byte[] GetURLContents(string url)
-    {
-        // The downloaded resource ends up in the variable named content.
-        var content = new MemoryStream();
-
-        // Initialize an HttpWebRequest for the current URL.
+        using var memoryStream = new MemoryStream();
         var webReq = (HttpWebRequest)WebRequest.Create(url);
```
Youssef1313, line 149: The API reference for HttpWebRequest says:

> We don't recommend that you use HttpWebRequest for new development. Instead, use the System.Net.Http.HttpClient class.

Do you want to me to open an issue to address this separately in another PR or address it in this PR (and the dotnet/samples PR too)?

BillWagner: Since we're making this many updates, I'd add it to this PR

/cc @IEvangelist

BillWagner (commented):
I'd like @adegeo to look here as well.

I'm concerned with the amount of code copied into the text, which defeats the purpose of the build system, and could cause the sample code and the text to drift apart over time.

I don't have a good solution. Hoping Andy does.

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
IEvangelist: The tutorial explicitly covers updating from `HttpWebRequest` to `HttpClient`. I think we should leave it, thoughts @BillWagner?

Youssef1313: Oh, My bad. I see that the final code uses HttpClient. It should be okay. But probably the other sample that's explained to be extended here may need to get updated to HttpClient? In that case, it may be another task.

Youssef1313:
> I'm concerned with the amount of code copied into the text, which defeats the purpose of the build system, and could cause the sample code and the text to drift apart over time.

Why not ask the engineering team to allow adding samples to the samples browser from this repository, then move the sample here in delete it from dotnet/samples? and at the same time, it will be maintained in the article and exists in the samples browser without duplication in two repos.

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
IEvangelist: What other examples are you referring to? The effort would more than likely include articles that refer to them.

adegeo (changes requested):
A few bugs and suggestions. I stopped reviewing around the "Convert" headers. I wanted to convey my concerns before I continued.

I think this article is a bit misguided. There are really two concepts:

1. How do I upgrade my existing app that does synchronous web access to asynchronous web access?
2. How do I do asynchronous web access?

The title of the article really lends itself to `#2`, but the content of the tutorial really lends itself to `#1`.

**First**, the audience getting here is probably not really concerned with creating an app from scratch and copy\pasting a bunch of content and possibly missing a step and not being able to compile. Really, it's just extra overhead for them. To simplify the tutorial and get people up-and-running faster would be to have the synchronous version of the app for download. The start of the tutorial tells them to download, extract, and edit the app. Now the focus of the tutorial would be converting synchronous GUI to asynchronous GUI as related to accessing the web.

**Secondly**, and probably more important, by using a WPF app we limit the audience who can actually use this tutorial to Windows-only users. This tutorial is in the C# programming guide and we should really focus on all OS platforms here unless the title specifically calls out "WPF" or "WinForms". However, with this being a basic concept, **asynch web access**, we shouldn't use a GUI.

**Third**, it seems there are multiple articles here all related to this app, yet the titles and instructions don't really create a cohesive "walkthrough" experience. If the goal of these articles is to create, convert, and demonstrate the different asynchronous techniques, these articles don't have the necessary header/footer scaffolding to truly guide a user.

While updating and modernizing these is important, I think dumping the WPF/WinForms model and using pure console code lends itself to a wider audience. This whole area is really out-of-date and needs to be modernized. This PR could be the start of that.

Thoughts?

IEvangelist:
> A few bugs and suggestions. I stopped reviewing around the "Convert" headers. I wanted to convey my concerns before I continued.
> 
> I think this article is a bit misguided. There are really two concepts:
> 
> 1. How do I upgrade my existing app that does synchronous web access to asynchronous web access?
> 2. How do I do asynchronous web access?
> 
> The title of the article really lends itself to `#2`, but the content of the tutorial really lends itself to `#1`.
> 
> **First**, the audience getting here is probably not really concerned with creating an app from scratch and copy\pasting a bunch of content and possibly missing a step and not being able to compile. Really, it's just extra overhead for them. To simplify the tutorial and get people up-and-running faster would be to have the synchronous version of the app for download. The start of the tutorial tells them to download, extract, and edit the app. Now the focus of the tutorial would be converting synchronous GUI to asynchronous GUI as related to accessing the web.
> 
> **Secondly**, and probably more important, by using a WPF app we limit the audience who can actually use this tutorial to Windows-only users. This tutorial is in the C# programming guide and we should really focus on all OS platforms here unless the title specifically calls out "WPF" or "WinForms". However, with this being a basic concept, **asynch web access**, we shouldn't use a GUI.
> 
> **Third**, it seems there are multiple articles here all related to this app, yet the titles and instructions don't really create a cohesive "walkthrough" experience. If the goal of these articles is to create, convert, and demonstrate the different asynchronous techniques, these articles don't have the necessary header/footer scaffolding to truly guide a user.
> 
> While updating and modernizing these is important, I think dumping the WPF/WinForms model and using pure console code lends itself to a wider audience. This whole area is really out-of-date and needs to be modernized. This PR could be the start of that.
> 
> Thoughts?

This is where I struggle a bit too. Since this content is so dated, I questioned whether or not it should simply be removed - and new content introduced? That would actually be a lot less time consuming. The heavy edits to existing, dated content seems to take a REALLY long time to do well... But I still believe there is value here. Not everything needs to be general. It's okay to have things that are specific to Windows, as there are still users who are writing WPF apps for Windows that would benefit from these and keeping this content updated is impactful.

I agree with the suggestions about simplifying the tutorial by prompting a download as a starting point. The way it is written today, it started from several potential places, either a sample, or from coding it up from scratch. It's a mess, and confusing, and very dated. But the reason I'm in here is that customers were relying on it, and reporting concerns - so what's the correct approach? Is it better to make tiny edits to address their concern, or do sweeping efforts to dramatically modernize and improve the content? I always assume the latter.

What do you propose I do?

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
```diff
@@ -1,379 +1,369 @@
...
+    using System.Windows;
     ```
 
 ## Create a synchronous app
 
-1. In the design window, MainWindow.xaml, double-click the **Start** button to create the `startButton_Click` event handler in MainWindow.xaml.cs.
-
-2. In MainWindow.xaml.cs, copy the following code into the body of `startButton_Click`:
+1. In the design window, *MainWindow.xaml*, double-click the **Start** button to create the click event handler in *MainWindow.xaml.cs*.
+1. Select the <kbd>F2</kbd> key to rename the handler to `OnStartButtonClick`.
```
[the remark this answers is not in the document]
IEvangelist: Interesting, I think I could have done better by referring to the start button click event handler as just that - rather than the function name. It lends itself to more of the conceptual aspect of documenting the source code, opposed to directly referencing a member. It also future-proofs the doc, so that the code itself can evolve.

But I am against the default naming that the tooling generates. Just because the tooling names something, doesn't mean we have to document it that way. I prefer following best practices in the naming conventions of modern C#. Event handlers should be named `OnObjectEventName`. Method names shouldn't include `_` and should use Pascal casing.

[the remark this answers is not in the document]
adegeo: Well the point I was bringing out is that our events pattern article for C# is wrong when it says to use `OnObjectEventName`, perhaps @BillWagner can comment on this. [WPF](https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/UIElement.cs#L3564), Winforms, and the guidance from the Framework 2.0 design book (which we still have published on the website) says that `OnObjectEventName` is used to raise the events and is implemented in class defining the event as a protected method. It's not supposed to be the event handler name.

adegeo:
The CSharp programming guide isn't really my domain, so I wouldn't have final say, I'm just voicing what I think makes sense :D I agree that we don't have to just get rid of WPF/WinForms all together, but when describing key concepts, we should keep it as generic as possible. We need those concepts demonstrated in a way that every user can benefit from them and run them. If there is a specific scenario, such as **converting synch GUI event handling (button clicks etc) to async**, that is completely appropriate.

I think we should rip out all 10 or so of these articles, figure out what concepts they are describing, boil that down to the required articles (the two or so conceptual articles already here are comprehensive and may really already cover this) and then decide what user-problem scenarios we want to cover. Do we need a multi-part walkthrough to describe these? Do we need tutorials to describe these? Are how-to articles fine enough? Those questions can only really be answered once we understand the scope of this section.

docs/csharp/programming-guide/concepts/async/walkthrough-accessing-the-web-by-using-async-and-await.md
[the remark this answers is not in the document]
adegeo: OK. I was able to look at the new Framework Design Guidelines 3rd Edition book. It does confirm that `OnEventName` should be used for raising events and allowing an inherited class to override that method. The examples it used for adding a handler to event however didn't use `On` or `_` in them. They simply were just named `[EventName]Handler` or `[ObjectNameEventName]`.

PR #19978, opened 2020-08-12
Restore vs.dotnet in ms.custom

tdykstra:
contperfq1 should have been added instead of replacing what was already there in ms.custom

Related to #20021


BillWagner (approved):
This LGTM @tdykstra 

:shipit: when ready.

PR #20022, opened 2020-08-13
Environment.OSVersion breaking change

### LINKED ISSUE #19684: Environment.OSVersion will now return the correct value for the OS on Windows.
## Environment.OSVersion will now return the correct value for the OS

Previously `Environment.OSVersion` would return a version that may be incorrect on Windows if an application war running under compatibility mode, as mentioned [here](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversionexa#remarks).

Now `Environment.OSVersion` will return the correct value.

### Version introduced

.NET 5.0

### Old behavior

`Environment.OSVersion` would return a version that may be incorrect when an application was running under Windows compatibility mode.

### New behavior

`Environment.OSVersion` returns the actual version of the operating system.

### Reason for change

Most users of this API expect it to return the actual version of the operating system.  Most .NET applications do not specify their supported version in their application manifest, and thus get the default from the dotnet host.  As a result the compatibility shim rarely is meaningful for the application that is running and they will typically get the actual version, except in cases where windows releases a new version and an older dotnet host is still in use.  We felt that returning the actual version was more inline with the expectation of the users of this API.

### Recommended action

Review and test any code that you have that uses Environment.OSVersion to ensure it behaves as desired.

### Category

- Core .NET libraries

### Affected APIs

  [Environment.OSVersion](https://docs.microsoft.com/en-us/dotnet/api/system.environment.osversion)

https://github.com/dotnet/runtime/issues/33643
https://github.com/dotnet/runtime/pull/33651

gewarren:
Fixes #19684.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/3.1-5.0?branch=pr-en-us-20022#environmentosversion-returns-the-correct-operating-system-version).


includes/core-changes/corefx/5.0/environment-osversion-returns-correct-version.md
```diff
@@ -0,0 +1,37 @@
+### Environment.OSVersion returns the correct operating system version
+
+<xref:System.Environment.OSVersion?displayProperty=nameWithType> returns the actual version of the operating system (OS) instead of, for example, the OS that's selected for application compatibility.
+
+#### Change description
+
+In previous .NET version, <xref:System.Environment.OSVersion?displayProperty=nameWithType> returns an OS version that may be incorrect when an application runs under Windows compatibility mode. For more information, see [GetVersionExA function remarks](/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversionexa#remarks).
```
Youssef1313, line 7: The remarks section of the API reference says:

> The Environment.OSVersion property does not provide a reliable way to identify the exact operating system and its version. Therefore, we do not recommend that you use this method. Instead:
>
> - To identify the operating system platform, use the RuntimeInformation.IsOSPlatform method.
> - Avoid writing code that depends on a reported operating system version. Instead, check for the availability of the features that your application needs.

Should this be updated to clarify that it's no longer valid for .NET 5.0, and probably give a link to this breaking change doc?

gewarren: Thanks, I updated it here: https://github.com/dotnet/dotnet-api-docs/pull/4674. In general, I always update the API ref for breaking changes.

gewarren, line 7:
```suggestion
In previous .NET versions, <xref:System.Environment.OSVersion?displayProperty=nameWithType> returns an OS version that may be incorrect when an application runs under Windows compatibility mode. For more information, see [GetVersionExA function remarks](/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversionexa#remarks).
```

BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

gewarren:
Thanks Bill!

PR #20043, opened 2020-08-13
Improve perf on VB landing page

gewarren:
- For content perf improvement KR
- Move language reference sublist up so it's visible "above the fold" (lots of clicks)
- Remove links that weren't clicked very often to declutter the page
- Add more links to popular VB pages (e.g. what's new, data types, string functions)

image


tdykstra (approved):
LGTM

PR #20072, opened 2020-08-14
MulticastOption.Group doesn't accept null

### LINKED ISSUE #19723: MulticastOption.Group no longer accepts null
## MulticastOption.Group no longer accepts null

MulticastOption.Group is not supposed to accept null. If someone sets it to null, we will NRE inside of the Sockets implementation.

### Version introduced

5.0

### Old behavior

`null` could be set on `MulticastOption.Group` property, and later if the `MulticastOption` was passed to `Socket.SetSocketOption`, the runtime would throw a `System.NullReferenceException`.

### New behavior

Setting `MulticastOption.Group` property to `null` results in an `ArgumentNullException` during the set.

### Reason for change

The NullReferenceException was discovered when annotating the Sockets library with nullable annotations. To be consistent with the Framework Design Guidelines, the property was updated to throw the correct exception.

### Recommended action

Ensure you don't pass `null` into `MulticastOption.Group` setter.

### Category

- Networking

### Affected APIs

`[MulticastOption.Group Property](https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.multicastoption.group)

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #19723.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/3.1-5.0?branch=pr-en-us-20072#multicastoptiongroup-doesnt-accept-a-null-value).

### DIFF
--- a/docs/core/compatibility/3.1-5.0.md
+++ b/docs/core/compatibility/3.1-5.0.md
@@ -204,6 +204,14 @@ If you're migrating from version 3.1 of .NET Core, ASP.NET Core, or EF Core to v
 
 ***
 
+## Networking
+
+- [MulticastOption.Group doesn't accept a null value](#multicastoptiongroup-doesnt-accept-a-null-value)
+
+[!INCLUDE [multicastoption-group-doesnt-accept-null](../../../includes/core-changes/networking/5.0/multicastoption-group-doesnt-accept-null.md)]
+
+***
+
 ## Serialization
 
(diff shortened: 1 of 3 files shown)


tdykstra (approved):
LGTM

eerhardt (approved):
Looks good. Thanks for documenting this.

PR #20076, opened 2020-08-15
Standardize sample projects, Part 18

nxtn:


samples/snippets/csharp/VS_Snippets_CLR/conceptual.resources.locating/cs/assemblyinfo.cs
```diff
@@ -1,40 +0,0 @@
-﻿using System.Reflection;
```
Youssef1313, line 1: As *assemblyinfo.cs* files are deleted, should we explicitly set [GenerateAssemblyInfo](https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#generateassemblyinfo) to False in csproj?

NextTurn: > The documentation says the default value is True, so I'd expect them to be re-generated, is that correct?

Right. These deleted *AssemblyInfo.cs* files only contain auto-generated values.

BillWagner (approved):
Thanks again @NextTurn 

I'll :shipit: now.

gewarren:
There's a build warning introduced here. Can someone fix it?

Invalid code snippet link: '~/samples/snippets/csharp/VS_Snippets_CLR/conceptual.resources.locating/cs/assemblyinfo.cs'.

Youssef1313:
@gewarren I'll open a PR.

PR #20082, opened 2020-08-15
Enable markdownlint for samples directory

Youssef1313:
I'm not sure why it was ignored in the first place. But the build seem to pass.

cc: @adegeo


BillWagner (approved):
Thanks @Youssef1313 

We'd excluded it for speed. But, there are readme files so this is a good addition.

PR #20086, opened 2020-08-15
Remove incorrect section about span and slicing for F#

cartermp:
This was not a correct paragraph/section, but the example itself is still valuable to have.


BillWagner (approved):
LGTM. :shipit:

PR #20095, opened 2020-08-16
Update F# nullable operators to point to fsharp.core reference

cartermp:
No issue yet, but it'll happen one way or another


BillWagner (approved):
LGTM. :shipit:

PR #20099, opened 2020-08-16
Update fsharp.core links in F# open declarations doc

cartermp:
Also clean up the namespaces stuff since `Microsoft` is implied

### DIFF
--- a/docs/fsharp/language-reference/import-declarations-the-open-keyword.md
+++ b/docs/fsharp/language-reference/import-declarations-the-open-keyword.md
@@ -1,7 +1,7 @@
 ---
 title: "Import Declarations: The open Keyword"
 description: Learn about F# import declarations and how they specify a module or namespace whose elements you can reference without using a fully qualified name.
-ms.date: 04/04/2019
+ms.date: 08/15/2020
 ---
 # Import declarations: The `open` keyword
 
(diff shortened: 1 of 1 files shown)


docs/fsharp/language-reference/import-declarations-the-open-keyword.md
```diff
@@ -46,19 +43,19 @@ Some namespaces are so frequently used in F# code that they are opened implicitl
 
 |Namespace|Description|
 |---------|-----------|
-|`Microsoft.FSharp.Core`|Contains basic F# type definitions for built-in types such as `int` and `float`.|
-|`Microsoft.FSharp.Core.Operators`|Contains basic arithmetic operations such as `+` and `*`.|
-|`Microsoft.FSharp.Collections`|Contains immutable collection classes such as `List` and `Array`.|
-|`Microsoft.FSharp.Control`|Contains types for control constructs such as lazy evaluation and asynchronous workflows.|
-|`Microsoft.FSharp.Text`|Contains functions for formatted IO, such as the `printf` function.|
+|`FSharp.Core`|Contains basic F# type definitions for built-in types such as `int` and `float`.|
```
BillWagner, line 46: holy breaking change, Batman!.  :bat: really?

BillWagner (approved):
This LGTM. You can :shipit:

Should there be an additional breaking change note on the namespace changes? Or does the F# compiler just pick that up?

docs/fsharp/language-reference/import-declarations-the-open-keyword.md
cartermp: Oh, this one will tickle you. These namespaces are actually all auto-opened anyways, so the vast majority of code will never notice anything.

But it gets better. The F# compiler actually treats these namespaces as equivalent even if they're explicitly opened. It's kind of gross, but ... it is what it is, and does let us drop the prefix.

PR #20107, opened 2020-08-16
Remove useless link in F# Types article

cartermp:
This link went nowhere and just pointed at the root of docs.microsoft.com


IEvangelist (approved):

PR #20119, opened 2020-08-16
Update list module function link in F# first class functions doc

cartermp:


IEvangelist (approved):

PR #20120, opened 2020-08-16
Remove useless msdn link in F# pattern matching doc

cartermp:
This link lead just to the root of the docs.ms site. Better to just not have it,.


IEvangelist (approved):

PR #20121, opened 2020-08-16
Cache markdownlint-cli npm installation in GH action

Youssef1313:
This is taken from [GitHub docs](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#using-the-cache-action).

@adegeo, Do you know how to confirm that this actually works?

I wouldn't expect a huge time difference, the only npm dependency is markdownlint, which has small size. So, feel free to close this if you think it doesn't worth caching the dependency.

### DIFF
--- a/.github/workflows/markdownlint.yml
+++ b/.github/workflows/markdownlint.yml
@@ -23,7 +23,15 @@ jobs:
     - uses: actions/checkout@v2
     - name: Use Node.js
       uses: actions/setup-node@v1
+      env:
+        cache_name: cache-node-modules
       with:
+        path: ~/.npm
+        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-build-${{ env.cache-name }}-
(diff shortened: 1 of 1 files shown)


.github/workflows/markdownlint.yml
```diff
@@ -23,7 +23,15 @@ jobs:
     - uses: actions/checkout@v2
     - name: Use Node.js
       uses: actions/setup-node@v1
+      env:
+        cache_name: cache-node-modules
       with:
+        path: ~/.npm
```
Youssef1313, line 29: This gave a warning 😄 

```
##[warning]Unexpected input(s) 'path', 'key', 'restore-keys', valid inputs are ['always-auth', 'node-version', 'registry-url', 'scope', 'version']
```

Is GitHub docs outdated or I'm missing something?

adegeo:
Sorry for the delay @Youssef1313 

Nah it doesn't seem worth it. I can see if the workflow restored 10 different node dependencies, but we only have one node app here.

Is there a specific problem you ran into and that is why you're suggesting this? I'll close this but we can reopen it if needed.

PR #20128, opened 2020-08-17
Corrected grammar in section "Nullable types"

### LINKED ISSUE #20125: Is the "to" here in "Types and members", Tour of C# possibly unnecessary?
In "Nullable types", there is a sentence:

> The compiler also provides warnings when non-nullable references are assigned to a value that may be null.

I'm trying to translate it into Chinese and I found it hard to understand when there is a "to". How can I assign a reference "to" a value? Should it be, "when non-nullable references are assigned a value", that is, a value is assigned to the references?

---
#### 文档详细信息

⚠ *请勿编辑此部分。 docs.microsoft.com ➟ GitHub 问题链接必须具有此部分。*

* ID: 0b0e44dd-e0f2-9ace-006b-479a92fbf424
* Version Independent ID: 7811190e-6019-83d8-d504-c0b985368055
* Content: [Define types and their members - A tour of C#](https://docs.microsoft.com/zh-cn/dotnet/csharp/tour-of-csharp/types)
* Content Source: [docs/csharp/tour-of-csharp/types.md](https://github.com/dotnet/docs/blob/master/docs/csharp/tour-of-csharp/types.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

Xin-Shiyu:
## Summary

The sentence, "The compiler also provides warnings when non-nullable references are assigned __to__ a value that may be null" does not sound correct, as you cannot assign a reference "to" a value. Instead, a value being assigned to a reference should be the normal case. Removing the "to" here makes better sense.

Fixes #20125


BillWagner (approved):
Thanks for making this fix @Xin-Shiyu 

PR #20136, opened 2020-08-17
IDentifier -> Identifier

Youssef1313:


tdykstra (approved):
Thanks!  The capitalization of D was probably done deliberately to match up with the RID acronym, but it looks odd.

PR #20139, opened 2020-08-17
Update isnot-operator.md

Youssef1313:
See https://github.com/dotnet/roslyn/blob/bf26b0607760eabf50ee20e60513b7492decea96/docs/wiki/New-Language-Features-in-VB-14.md#typeof-expression-isnot-type-operator

### DIFF
--- a/docs/visual-basic/language-reference/operators/isnot-operator.md
+++ b/docs/visual-basic/language-reference/operators/isnot-operator.md
@@ -49,7 +49,7 @@ The following code example uses both the `Is` operator and the `IsNot` operator 
 
 ## Use TypeOf operator with IsNot operator
 
-Starting with Visual Basic 14, you can use the `TypeOf` operator with the `IsNot` operator. For example, instead of writing:
+Starting with Visual Basic 14, you can use the `TypeOf` operator with the `IsNot` operator to test whether an object is *not* compatible with a data type. For example:
 
 ```vb
 If TypeOf sender IsNot Button Then
--- a/docs/visual-basic/language-reference/operators/isnot-operator.md
+++ b/docs/visual-basic/language-reference/operators/isnot-operator.md
@@ -35,7 +35,7 @@ result = object1 IsNot object2
 
 ## Remarks
 
-The `IsNot` operator determines if two object references refer to different objects. However, it does not perform value comparisons. If `object1` and `object2` both refer to the exact same object instance, `result` is `False`; if they do not, `result` is `True`.
+The `IsNot` operator determines if two object references refer to different objects. However, it doesn't perform value comparisons. If `object1` and `object2` both refer to the exact same object instance, `result` is `False`; if they don't, `result` is `True`.
 
 `IsNot` is the opposite of the `Is` operator. The advantage of `IsNot` is that you can avoid awkward syntax with `Not` and `Is`, which can be difficult to read.
 
--- a/docs/visual-basic/language-reference/operators/isnot-operator.md
+++ b/docs/visual-basic/language-reference/operators/isnot-operator.md
@@ -50,12 +50,6 @@ The following code example uses both the `Is` operator and the `IsNot` operator 
 Starting with Visual Basic 14, you can use the `TypeOf` operator with the `IsNot` operator. For example, instead of writing:
 
 ```vb
-If Not TypeOf sender Is Button Then
-```
-
-You can use:
-
-```vb
 If TypeOf sender IsNot Button Then
 ```
 
--- a/docs/visual-basic/language-reference/operators/isnot-operator.md
+++ b/docs/visual-basic/language-reference/operators/isnot-operator.md
@@ -1,5 +1,5 @@
 ---
-title: "IsNot Operator"
+title: IsNot operator
 ms.date: 07/20/2015
 f1_keywords: 
   - "vb.isnot"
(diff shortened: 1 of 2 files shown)


docs/visual-basic/language-reference/operators/isnot-operator.md
```diff
@@ -18,26 +18,46 @@ result = object1 IsNot object2
...
- The following code example uses both the `Is` operator and the `IsNot` operator to accomplish the same comparison.
 
- [!code-vb[VbVbalrOperators#29](~/samples/snippets/visualbasic/VS_Snippets_VBCSharp/VbVbalrOperators/VB/Class1.vb#29)]
+The following code example uses both the `Is` operator and the `IsNot` operator to accomplish the same comparison.
+
+[!code-vb[VbVbalrOperators#29](~/samples/snippets/visualbasic/VS_Snippets_VBCSharp/VbVbalrOperators/VB/Class1.vb#29)]
+
+## Use TypeOf operator with IsNot operator
+
+Starting with Visual Basic 14, you can use TypeOf operator with IsNot operator. Instead of writing:
```
adegeo, line 50:
```suggestion
Starting with Visual Basic 14, you can use the `TypeOf` operator with the `IsNot` operator. For example, instead of writing:
```

docs/visual-basic/language-reference/operators/isnot-operator.md
```diff
@@ -18,26 +18,46 @@ result = object1 IsNot object2
...
+
+- `object2`
+
+  Required. Any `Object` variable or expression.
 
 ## Remarks
- The `IsNot` operator determines if two object references refer to different objects. However, it does not perform value comparisons. If `object1` and `object2` both refer to the exact same object instance, `result` is `False`; if they do not, `result` is `True`.
 
- `IsNot` is the opposite of the `Is` operator. The advantage of `IsNot` is that you can avoid awkward syntax with `Not` and `Is`, which can be difficult to read.
+The `IsNot` operator determines if two object references refer to different objects. However, it does not perform value comparisons. If `object1` and `object2` both refer to the exact same object instance, `result` is `False`; if they do not, `result` is `True`.
```
adegeo, line 36:
```suggestion
The `IsNot` operator determines if two object references refer to different objects. However, it doesn't perform value comparisons. If `object1` and `object2` both refer to the exact same object instance, `result` is `False`; if they don't, `result` is `True`.
```

docs/visual-basic/language-reference/operators/isnot-operator.md
```diff
@@ -18,26 +18,46 @@ result = object1 IsNot object2
...
+
+```vb
+If Not TypeOf sender Is Button Then
+```
+
+You can use:
+
+```vb
+If TypeOf sender IsNot Button Then
+```
```
adegeo, line 60: I think we can remove the "old way" and just simply say you can use it with the type of and show it. The previous section brings out that IsNot is better than Is and Not.

adegeo: Also, I think we should make a similar update to the `Is operator` article. That article does bring out that it can be used with `TypeOf`, but it's a note. I think it's nice to have it as a section, like you did here.

docs/visual-basic/language-reference/operators/isnot-operator.md
```diff
@@ -18,26 +18,46 @@ result = object1 IsNot object2
 ```
 
 ## Parts
- `result`
- Required. A `Boolean` value.
 
- `object1`
- Required. Any `Object` variable or expression.
+- `result`
```
adegeo, line 22: These `-` don't follow the convention of the other articles, so we should remove them.

adegeo (changes requested):
Thanks for the improvements! Minor changes requested.

docs/visual-basic/language-reference/operators/isnot-operator.md
Youssef1313: @adegeo I changed some articles to use list item a long time ago :cry:. So there will still be some articles that don't follow the convention of others.

docs/visual-basic/language-reference/operators/isnot-operator.md
```diff
@@ -18,26 +20,40 @@ result = object1 IsNot object2
...
- The following code example uses both the `Is` operator and the `IsNot` operator to accomplish the same comparison.
 
- [!code-vb[VbVbalrOperators#29](~/samples/snippets/visualbasic/VS_Snippets_VBCSharp/VbVbalrOperators/VB/Class1.vb#29)]
+The following code example uses both the `Is` operator and the `IsNot` operator to accomplish the same comparison.
+
+[!code-vb[VbVbalrOperators#29](~/samples/snippets/visualbasic/VS_Snippets_VBCSharp/VbVbalrOperators/VB/Class1.vb#29)]
+
+## Use TypeOf operator with IsNot operator
+
+Starting with Visual Basic 14, you can use the `TypeOf` operator with the `IsNot` operator. For example, instead of writing:
```
Youssef1313, line 52:
```suggestion
Starting with Visual Basic 14, you can use the `TypeOf` operator with the `IsNot` operator to test whether an object is *not* compatible with a data type. For example:
```

adegeo (approved):

PR #20145, opened 2020-08-18
Standardize sample projects, Part 20

nxtn:
Follow-up to #20076 /cc @gewarren 

See dotnet/sdk#12951 for why it's necessary to build the .vbproj with Visual Studio.


gewarren (approved):
@NextTurn Thank you!

PR #20137, opened 2020-08-17
Clean up redirected links

gewarren:
Contributes to #19777.

PR #20149, opened 2020-08-18
Create markdown-link-checker.yml

Youssef1313:
Trying to find a good working workflow that detects problem such as the ones being fixed in https://github.com/dotnet/docs/pull/20137.

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -9,4 +9,4 @@ jobs:
       - uses: actions/checkout@v2
       - name: Link Checker
         id: lc
-        uses: peter-evans/link-checker@v1
+        uses: peter-evans/link-checker@v1.2.2
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -1,2 +1,10 @@
-- name: Link Checker
-  uses: peter-evans/link-checker@v1
+name: Markdown link checker
+
+on: push
+
+jobs:
+  markdown-link-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - uses: peter-evans/link-checker@v1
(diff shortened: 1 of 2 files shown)


Youssef1313:
I can't get either of the two workflows I found to work.

@adegeo Any ideas?

adegeo:
Hrm, not sure? But I dont think this would work anyhow. Does the tool check xref links? Those can be put inside of markdown links: `[some link text](xref:api-path)`

The build system at least tells us of any non HTTP link AFAIK.

Youssef1313:
I think xrefs shouldn't be detected as relative link. These actions should only work in relative links.

adegeo:
The way I got the initial snippets testing working to work was doing it in like your fork first. So I think the workflows only trigger when you're working on the branch they are in. So create branch1 with the workflow configured and ready, then do things in branch2 which has a PR into branch1 where the workflow exists. And I think your trigger needs to be `on: pull_request` not `on: push` as you want to test the links of the files when a bad PR comes in, not after it's merged.

Youssef1313:
@adegeo The work flows indeed was triggered in my fork (they doesn't appear in PR but I can see in my fork). However, they doesn't seem to be doing any good.

One of the workflows (https://github.com/Youssef1313/docs/actions?query=workflow%3A%22Markdown+link+checker%22) passes  in 5 minutes, although they should report broken links.

The second one is actually very funny (https://github.com/Youssef1313/docs/actions?query=workflow%3A%22Check+Markdown+links%22), it fails after 6 hours with a timeout 😄

adegeo:
No idea really.. does the workflow work with PRs checking just the files in it? I would look into how to configure that script. A lot of scripts allow you set them up through properties or env variables.

Youssef1313:
@adegeo I'll try to play with it in my fork and see if I could get something to work. In the meanwhile, I'll close the PR and re-open if I could get to something working.

PR #20146, opened 2020-08-18
Merge master into live

ghost:
Please don't squash-merge this PR.

PR #20151, opened 2020-08-18
Fix build warning in Arrays.md (F#)

Youssef1313:
After merging this, it needs to be added to https://github.com/dotnet/docs/pull/20146. (Edit: That's now merged already. This fix will go in the next publish)


cartermp (approved):
Thanks for the catch!

PR #20154, opened 2020-08-18
Fix title suffix

Youssef1313:
A title suffix is applied globally and affects this file. So, the current title is "What's new for Visual Basic - Visual Basic".

After this fix, it will be "What's new - Visual Basic".

If you would like it to be "What's new for Visual Basic", let me know to override the titleSuffix.


BillWagner (approved):
This LGTM @Youssef1313 

I'll :shipit: now.

PR #20178, opened 2020-08-19
Fix typo

JasonCard:
## Summary

Fixing what appears to be typo where period is within the bold font. This was pointed out by translator who was not sure if the period was meant to be part of the identifier "Azure".


IEvangelist (approved):

PR #20188, opened 2020-08-19
Add `ms.custom: "updateeachrelease"` to keywords index

Youssef1313:
Usually, new C# versions come with new keywords.


BillWagner (approved):
Thanks @Youssef1313 

I'll merge this, although it's likely not necessary. We use the "updateeachrelease" tag to help us find articles that need updating where we are likely to miss them. We're unlikely to miss adding a new keyword here as we do the other work to document what it does.

PR #20208, opened 2020-08-20
Workaround localization issue

### LINKED ISSUE #3042: Wrong return value on tutorial page
Go to https://docs.microsoft.com/it-it/dotnet/csharp/tutorials/intro-to-csharp/list-collection?tutorial-step=3
The description of  the return value of the function IndexOf, in case the element is not founded, is indicated as 1 instead of -1 (at least in the italian version of the site)

Youssef1313:
Fixes MicrosoftDocs/feedback#3042

This might not be the ideal fix, but just workaround the issue.

cc: @JasonCard

### DIFF
--- a/docs/csharp/tutorials/intro-to-csharp/list-collection.yml
+++ b/docs/csharp/tutorials/intro-to-csharp/list-collection.yml
@@ -85,7 +85,7 @@ items:
       Console.WriteLine($"When an item is not found, IndexOf returns {notFound}");
     ```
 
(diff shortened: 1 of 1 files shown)


JasonCard:
I see `-1` is already used in the first paragraph, so this would make the content more consistent as well.

BillWagner (approved):
This LGTM @Youssef1313 

I'll :shipit: now.

PR #20212, opened 2020-08-20
Property name mistake

AbregoDev:
In the *Binding and BindingExpression* there's a mistake in the property name of the binding source.

## Summary

Describe your changes here.


adegeo (approved):
Thank you for catching this!

PR #20236, opened 2020-08-21
Enhance description

johypark:
## Summary

Enhance description on Enhance the app

PR #20240, opened 2020-08-21
Prevent possible over-localization

Youssef1313:
Addresses https://github.com/dotnet/docs/pull/20236#discussion_r474826733


tdykstra (commented):
If we add no loc for the name prompt we should do it for the press any key prompt as well.

Also the same text is in VS Code and VS Mac versions of this tutorial.

An alternative that might work better is to avoid repeating what's in the code -- e.g., "The code displays a prompt in the console window ..." and "Finally, it displays a prompt in the console window ..."

johypark (changes requested):

tdykstra (approved):
Thanks!

PR #20249, opened 2020-08-22
Fix breaking change category (Core libraries -> Code analysis)

Youssef1313:


Youssef1313 (commented):
@gewarren, the new folder in *core-changes* is called *codeanalysis*. Was that naming intentional without `-` (i.e. *code-analysis*)?

gewarren (approved):
Thanks for catching that. The folder name was intentional to match others like visualbasic and aspnetcore. Also, since this is an include file folder, the folder name isn't used in the published URL.

PR #20261, opened 2020-08-24
Use dotnetcli and update example to 3.1

Youssef1313:


IEvangelist (approved):

PR #20281, opened 2020-08-26
Change wording for referring to Core/5+

tdykstra:

### DIFF
--- a/docs/standard/glossary.md
+++ b/docs/standard/glossary.md
@@ -42,13 +42,13 @@ Base Class Library. Also known as *framework libraries*.
 
 A set of libraries that comprise the System.\* (and to a limited extent Microsoft.\*) namespaces. The BCL is a general purpose, lower-level framework that higher-level application frameworks, such as ASP.NET Core, build on.
 
-The source code of the BCL for [.NET 5 and later versions (including .NET Core 2.1-3.1)](#net-5-and-later-versions) is contained in the [.NET runtime repository](https://github.com/dotnet/runtime). The majority of the BCL APIs for this newer implementation of .NET are also available in .NET Framework, so you can think of this source code as a fork of the .NET Framework BCL source code.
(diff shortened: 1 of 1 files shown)


docs/standard/glossary.md
```diff
@@ -42,13 +42,13 @@ Base Class Library. Also known as *framework libraries*.
 
 A set of libraries that comprise the System.\* (and to a limited extent Microsoft.\*) namespaces. The BCL is a general purpose, lower-level framework that higher-level application frameworks, such as ASP.NET Core, build on.
 
-The source code of the BCL for [.NET 5 and later versions (including .NET Core 2.1-3.1)](#net-5-and-later-versions) is contained in the [.NET runtime repository](https://github.com/dotnet/runtime). The majority of the BCL APIs for this newer implementation of .NET are also available in .NET Framework, so you can think of this source code as a fork of the .NET Framework BCL source code.
+The source code of the BCL for [.NET 5 (and .NET Core) and later versions](#net-5-and-later-versions) is contained in the [.NET runtime repository](https://github.com/dotnet/runtime). The majority of the BCL APIs for this newer implementation of .NET are also available in .NET Framework, so you can think of this source code as a fork of the .NET Framework BCL source code.
```
Youssef1313, line 45: This might not be very accurate.
dotnet/runtime contains the source for .NET 5 and later. However, the source code for .NET Core is still in dotnet/corefx branches (e.g. https://github.com/dotnet/corefx/tree/release/3.1).

The source code for .NET Framework isn't in any of these repositories. It's in https://referencesource.microsoft.com/ (it also exists somewhere in GitHub, but I can't remember it).

---

> The majority of the BCL APIs for this newer implementation of .NET are also available in .NET Framework, so you can think of this source code as a fork of the .NET Framework BCL source code.

This is true, but the implementation has improved for many things.

For example, Regex existed since .NET Framework 1.1. However, the implementation should have changed a lot.

Youssef1313 (commented):
I also think the article can be moved to the new "fundamentals" section. It's not `.NET standard`-specific

docs/standard/glossary.md
BillWagner: I like Tom's version. It is accurate, but not too much information for a glossary. I don't think it's helpful to readers to explain the history of .NET in the glossary.

BillWagner (approved):
This LGTM @tdykstra 

You can :shipit: when ready.

PR #20287, opened 2020-08-26
Various clarifications in the F# style guidelines

dsyme:
## Summary

Various clarifications in the F# style guidelines 

Basically the guidance "prefer structs" is too strong and should be qualified by "high allocation"

cc @cartermp  @luisquintanilla 

[Preview URL](https://review.docs.microsoft.com/dotnet/fsharp/style-guide/conventions?branch=pr-en-us-20287)

### DIFF
--- a/docs/fsharp/style-guide/conventions.md
+++ b/docs/fsharp/style-guide/conventions.md
@@ -102,34 +102,19 @@ open System.IO
 open System.Reflection
 open System.Text
 
-open Microsoft.FSharp.Compiler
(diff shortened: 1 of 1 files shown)


luisquintanilla (approved):
:shipit:

cartermp:
The reason why this was worded the way it was is because of measuring the impact in real-world scenarios and benchmarks. In the case of single-case DUs, struct is always better. Same with small records and tuples. Note that this is specifically for the performance section and not generally speaking, so perhaps that should additionally be qualified.

PR #20316, opened 2020-08-27
Update `var` with info when used with NRT

Youssef1313:


docs/csharp/language-reference/keywords/var.md
```diff
@@ -19,6 +19,8 @@ int i = 10; // Explicitly typed.
 
 For more information, see [Implicitly Typed Local Variables](../../programming-guide/classes-and-structs/implicitly-typed-local-variables.md) and [Type Relationships in LINQ Query Operations](../../programming-guide/concepts/linq/type-relationships-in-linq-query-operations.md).
 
+When `var` is used with nullable reference types enabled, it always implies a nullable reference type even if the expression type isn't nullable.
```
IEvangelist, line 22:
```suggestion
> [!IMPORTANT]
> When `var` is used with nullable reference types enabled, it always implies a nullable reference type even if the expression type isn't nullable.
```

IEvangelist (commented):
Minor nit, wrap as Important.

IEvangelist (approved):

PR #20341, opened 2020-08-28
Fix null forgiving F1 keyword

### LINKED ISSUE #46988: Issue with `!_CSharpKeyword`
Both the not operator and null-forgiving operator have the same F1 help keyword. That causes the dotnet/docs repository to have the exact same f1 keyword in two articles. And unfortunately, pressing F1 in any `!` always redirects to the null-forgiving article. I think Roslyn should provide different f1 keywords so dotnet/docs can be updated to have different f1 keywords too.

cc: @BillWagner @gewarren

Youssef1313:
Fixes https://github.com/dotnet/roslyn/issues/46988

### DIFF
--- a/docs/csharp/language-reference/operators/null-forgiving.md
+++ b/docs/csharp/language-reference/operators/null-forgiving.md
@@ -3,7 +3,7 @@ description: "! (null-forgiving) operator - C# reference"
 title: "! (null-forgiving) operator - C# reference"
 ms.date: 10/11/2019
 f1_keywords: 
-  - "!_CSharpKeyword"
+  - "nullForgiving_CSharpKeyword"
 helpviewer_keywords: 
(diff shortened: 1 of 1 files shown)


Youssef1313:
@BillWagner Once https://github.com/dotnet/roslyn/pull/47516 is reviewed for correctness, this should be ready to merge, then Roslyn's PR.

davidwengier (approved):
I don't think there is any reason to wait for the Roslyn PR to be approved, having the same `f1_keyword` in two pages is a problem regardless of whether any code currently points to it.

davidwengier (approved):
I don't think there is any reason to wait for the Roslyn PR to be approved, having the same `f1_keyword` in two pages is a problem regardless of whether any code currently points to it.

PR #20368, opened 2020-08-31
Fix incorrect/oudated documentation for supported platforms

Youssef1313:
This is based on what I see in:

https://github.com/dotnet/runtime/blob/77b752b7cec435943668df438ab9cb33d5ece253/src/libraries/System.Security.Cryptography.Csp/ref/System.Security.Cryptography.Csp.cs#L47-L48

However, this PR still needs a product team review.

I also think this file should be somehow auto generated from runtime source code based on this new attribute.


Youssef1313:
Oops, that's a completely different API. Sorry, closing 😄

PR #20357, opened 2020-08-30
Improve minor syntax

johypark:
## Summary

Improve syntax, Fix IDE0049

I edited these things after reading library-with-visual-studio.

Please merge if you want, thanks.

PR #20397, opened 2020-09-01
Improve minor syntax

johypark:
## Summary

After #20357, reading testing-library-with-visual-studio, and some C# docs,
I realized that I made a mistake in my previous pull request.
Now I know that most people and also docs prefer starting with a lower case for variables and with an upper case for classes.

So I fixed it again including UnitTest1.cs.
Sorry to confuse.

### DIFF
--- a/docs/core/tutorials/snippets/library-with-visual-studio/csharp/ShowCase/Program.cs
+++ b/docs/core/tutorials/snippets/library-with-visual-studio/csharp/ShowCase/Program.cs
@@ -13,7 +13,7 @@ class Program
                 ResetConsole();
 
             string input = Console.ReadLine();
-            if (string.IsNullOrEmpty(input)) break;
+            if (String.IsNullOrEmpty(input)) break;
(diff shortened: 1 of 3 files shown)


tdykstra:
A common preference is to use language keywords when they are available, as your previous PR did.  I think we're fine with the previous changes.

JohyPark:
I have read these documents: 
[Types (C# Programming Guide)](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/types/)
[Strings (C# Programming Guide)](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/strings/)
[Why is “string” considered a simplified version of “String”?](https://stackoverflow.com/questions/32187486/why-is-string-considered-a-simplified-version-of-string)

tdykstra:
There's no right or wrong answer here, but my preference is to keep the language keywords.

JohyPark:
Thanks for the advice.

PR #20405, opened 2020-09-02
Fix grammatical error

xTyler73:
"(the method in this case.) This behavior"
should be
"(the method in this case). This behavior"
The period is placed outside of the bracket consistently throughout the rest of the documentation.
An example of a page that uses ")." 8 times: https://docs.microsoft.com/en-us/dotnet/csharp/tour-of-csharp/program-building-blocks


IEvangelist (approved):

PR #20409, opened 2020-09-02
Fix typo to be consistent with example

bnewcomb:
The reference to the expression was inconsistent with the corresponding example above it.  The member referenced in the interpolation expression should be Date, not Data.

## Summary

I modified the text to the correct vowel 'e' instead of the erroneous vowel 'a', making the text of the expression consistent with the example code and improving the value of the encapsulating text by decreasing the probability of confusion or higher orders of cognitive processing that have the potential to diminish the learning experience via the phenomenology of frustration that could negate the benefit of the intended clarification.


IEvangelist (approved):

PR #20411, opened 2020-09-02
using code sample fix

mohsin-mehmood:
Moved the return `return skippedLines;` line of code inside using code block

## Summary

Describe your changes here.


IEvangelist (approved):

PR #20413, opened 2020-09-02
Fix labeling Roslyn issues

Youssef1313:
Discovered in #20403 

The value for technology is set as "csharp-roslyn" not "csharp-csharp-roslyn":

https://github.com/dotnet/docs/blob/a0980c8fea1fe7ba30538d63e8fa5486af168a95/docfx.json#L438


IEvangelist (approved):

PR #20425, opened 2020-09-02
Add breaking change for CA2013

### LINKED ISSUE #19806: Breaking change: Roslyn analyzer CA2013 - Do not use ReferenceEquals with value types
## Breaking change: Roslyn analyzer CA2013 - Do not use ReferenceEquals with value types

This new Roslyn analyzer will emit a warning if it finds instances where the `System.Object.ReferenceEquals` method is used to test one or more value types for equality.

Depending on the project's settings, the warning could show up as a build break.

### Version introduced

This analyzer is included in .NET 5 Preview 8 SDK and turned on by default.
It can be turn them off by setting the MSBuild flag `EnableNETAnalyzers` to `false`.

### Old behavior

If the user compared two value types using `System.Object.ReferenceEquals`, they are boxed before they get passed to the method, which causes the method to return `false` even if they both represent the same instance of a value type.

### New behavior

The user will now see the CA2013 warning suggesting to instead use an appropriate equality operator, such as `==`.

### Reason for change

Using `System.Object.ReferenceEquals` to compare value types can lead to wrong results.

### Recommended action

Change the code to use an appropriate equality operator, such as `==`. This warning is not safe to suppress.

Refer to the documentation for rule CA2013 for a detailed explanation on when to expect this rule violation to be diagnosed.

https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2013

### Category

- Core .NET libraries

### Affected APIs

[System.Object.ReferenceEquals](https://docs.microsoft.com/en-us/dotnet/api/system.object.referenceequals) when used to compare two value types.

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #19806.
Also changes heading of CA1831 breaking change.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/3.1-5.0?branch=pr-en-us-20425#ca2013-do-not-use-referenceequals-with-value-types).


carlossanlop:
I don't think the heading for CA1831 should be changed. It should match the original title provided in the Roslyn Analyzer: https://github.com/dotnet/roslyn-analyzers/search?q=ca1831

carlossanlop:
@buyaa-n can you please review?

CamSoper (approved):

PR #20431, opened 2020-09-03
clarify array text and integrate tuple value type

bnewcomb:
Made text consistent with other parts of itself by updating lists to include the tuple value type.
Clarified text surrounding single-dimensional arrays and the description of a jagged array.


IEvangelist (approved):

PR #20416, opened 2020-09-02
Document System.Net.Cookie path related breaking changes

CarnaViire:
Described changes introduced by PR dotnet/runtime#39250

Fixes #20363

PR #20435, opened 2020-09-03
Fix cookie breaking change docs

CarnaViire:
Fixed @gewarren comments from PR #20416 (that was merged before I could react)


IEvangelist (approved):

PR #20440, opened 2020-09-03
Replaces "indexer" word with the "accessor"

mialkin:
`set` is an accessor, not an indexer


BillWagner (approved):
Thanks for spotting this and fixing it @mialkin 

PR #20446, opened 2020-09-03
Whats new in C# 9

### LINKED ISSUE #19969: What's new in C# 9.0
We want readers to find new content for C# 9 as soon as we update different areas in the C# Guide for C# 9.0. 

The new article created for this issue should link to each update as more C# 9 content publishes.

Part of #19716

BillWagner:
Add the What's new in C# 9 article.

Fixes #19969


docs/csharp/whats-new/snippets/whats-new-csharp9/PatternUtilities.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public static class PatternUtilities
+    {
+        // <IsLetterPattern>
+        public static bool IsLetter(this char c) =>
+            c is >= 'a' and <= 'z' or >= 'A' and <= 'Z';
+        // </IsLetterPattern>
+
+        // <IsLetterOrSeparatorPattern>
+        public static bool IsLetterIsSeparator(this char c) =>
```
Youssef1313, line 17: Nit: "IsLetterOrSeparator".

BillWagner:
The CI build is failing because `PrintMembers` didn't make the preview 8 build. It is in RC1, when this build error will go away.

gewarren:
> The CI build is failing because `PrintMembers` didn't make the preview 8 build. It is in RC1, when this build error will go away.

I don't think we should knowingly introduce a build warning. It means everyone will have to check the build report for every PR.

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+- Methods for value-based equality comparisons
+- Override for GetHashCode
+- Copy and Clone members
+- PrintMembers and ToString
+- Deconstruct method
+
+Records support inheritance. You can declare a new record derived from `Person` as follows:
+
+:::code language="csharp" source="snippets/whats-new-csharp9/RecordsExample.cs" ID="InheritedRecord":::
```
IEvangelist:
```suggestion
:::code language="csharp" source="snippets/whats-new-csharp9/RecordsExamples.cs" ID="InheritedRecord":::
```

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+You can also seal records to prevent further derivation:
+
+:::code language="csharp" source="snippets/whats-new-csharp9/RecordsExample.cs" ID="SealedRecord":::
```
IEvangelist:
```suggestion
:::code language="csharp" source="snippets/whats-new-csharp9/RecordsExamples.cs" ID="SealedRecord":::
```

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+:::code language="csharp" source="snippets/whats-new-csharp9/RecordsExample.cs" ID="RecordsEquality":::
```
IEvangelist:
```suggestion
:::code language="csharp" source="snippets/whats-new-csharp9/RecordsExamples.cs" ID="RecordsEquality":::
```

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+Record types make it easy to create immutable reference types in .NET. Historically, .NET types are largely classified as classes or anonymous types, which are reference types and structs or tuples, which are value types. While creating immutable value types is recommended, mutable value types don’t often introduce errors. That’s because value types are passed by value to methods so any changes are made to a copy of the original data.
+
+There are many advantages to immutable reference types as well. These advantages are more pronounced in concurrent programs with shared data. Unfortunately, C# forced you to write quite a bit of extra code to create immutable reference types. Records provide a type declaration for an immutable reference type that uses value semantics for equality. The synthesized methods for equality and hash codes consider two records equal if their properties are all equal. Consider this definition:
+
+:::code language="csharp" source="snippets/whats-new-csharp9/RecordsExample.cs" ID="RecordDefinition":::
```
IEvangelist:
```suggestion
:::code language="csharp" source="snippets/whats-new-csharp9/RecordsExamples.cs" ID="RecordDefinition":::
```

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+The compiler produces a `Deconstruct` method for positional records. The `Deconstruct` method has parameters that match the names of all public properties in the record type. The `Deconstruct` method can be used to deconstruct the record into its component properties:
+
+:::code language="csharp" source="snippets/whats-new-csharp9/PositionalRecords.cs" ID="DeconstructRecord":::
+
+Finally, records support ***with-expressions***. A ***with-expression*** instructs the compiler to create a something like copy of a record, but with specified properties modified:
+
+:::code language="csharp" source="snippets/whats-new-csharp9/PositionalRecords.cs" ID="Wither":::
+
+The above line creates a new `Person` record where the `LastName` property is a copy of `person`, and the `FirstName` is "Paul". You can set any number of properties in a with-expression.  Any of the synthesized members except the "clone" method may be written by you. If a record type has a method that matches the signature of any synthesized method, the compiler does not synthesize that method. The earlier `Dog` record example contains a hand coded <xref:System.ToString> method as an example.
```
IEvangelist:
```suggestion
The above line creates a new `Person` record where the `LastName` property is a copy of `person`, and the `FirstName` is "Paul". You can set any number of properties in a with-expression.  Any of the synthesized members except the "clone" method may be written by you. If a record type has a method that matches the signature of any synthesized method, the compiler does not synthesize that method. The earlier `Dog` record example contains a hand coded <xref:System.Object.ToString> method as an example.
```

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+:::code language="csharp" source="snippets/whats-new-csharp9/RecordsExample.cs" ID="RecordsEquality":::
+
+The compiler synthesizes two methods that support printed output: a <xref:System.ToString> override, and `PrintMembers`. The `PrintMembers` method returns a comma-separated list of property names and values. The <xref:System.ToString> override returns the string produced by `PrintMembers`, surrounded by `{` and `}`. For example, the <xref:System.ToString> method for `Student` returns a `string` like the following code:
```
IEvangelist, line 68:
```suggestion
The compiler synthesizes two methods that support printed output: a <xref:System.Object.ToString> override, and `PrintMembers`. The `PrintMembers` method returns a comma-separated list of property names and values. The <xref:System.Object.ToString> override returns the string produced by `PrintMembers`, surrounded by `{` and `}`. For example, the <xref:System.Object.ToString> method for `Student` returns a `string` like the following code:
```

docs/csharp/whats-new/snippets/whats-new-csharp9/PositionalRecords.cs
```diff
@@ -0,0 +1,60 @@
...
+                base.PrintMembers(s);
+                return $"{s.ToString()} is a dog";
+            }
+        }
+        // </RecordsWithMethods>
+
+        public static void PositionalExamples()
+        {
+            // <DeconstructRecord>
+            Person person = new Person("Bill", "Wagner");
```
IEvangelist, line 46:
```suggestion
            var person = new Person("Bill", "Wagner");
```

docs/csharp/whats-new/snippets/whats-new-csharp9/PositionalRecords.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            var (first, last) = person;
+            // Deconstruct is in the order declared in the record. (string first, string last) = p1;
```
IEvangelist, line 49:
```suggestion
```

Comments are not localized, and I believe that this code is self-documenting. The variable names imply their intent.

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+These declarations create the same functionality as the earlier version (with a couple extra features covered in the following section). These declarations end with a semicolon instead of brackets because these records do not add additional methods. You can add a body, and include any additional methods as well:
+
+:::code language="csharp" source="snippets/whats-new-csharp9/PositionalRecords.cs" ID="RecordsWithMethods":::
+
+The compiler produces a `Deconstruct` method for positional records. The `Deconstruct` method has parameters that match the names of all public properties in the record type. The `Deconstruct` method can be used to deconstruct the record into its component properties:
+
+:::code language="csharp" source="snippets/whats-new-csharp9/PositionalRecords.cs" ID="DeconstructRecord":::
+
+Finally, records support ***with-expressions***. A ***with-expression*** instructs the compiler to create a something like copy of a record, but with specified properties modified:
```
IEvangelist, line 86:
```suggestion
Finally, records support ***with-expressions***. A ***with-expression*** instructs the compiler to create a copy of a record, but *with* specified properties modified:
```

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Init only setters
+
+***Init only setters*** provide consistent syntax to initialize members of an object. Property initializers provide make it clear which value is setting which property. The downside is that those properties must be settable. Starting with C# 9.0, you can create `init` accessors, in addition of `set` accessors for properties and indexers. The main benefit is that callers can use property initializer syntax to set these values in creation expressions, but those properties are readonly once construction has completed. Init only setters provide a window to change state. That window closes when the construction phase ends. The construction phase effectively ends after all initialization, including property initializers and with-expressions have completed.
```
IEvangelist:
```suggestion
***Init only setters*** provide consistent syntax to initialize members of an object. Property initializers provide make it clear which value is setting which property. The downside is that those properties must be settable. Starting with C# 9.0, you can create `init` accessors, in addition of `get` accessors for properties and indexers. The main benefit is that callers can use property initializer syntax to set these values in creation expressions, but those properties are readonly once construction has completed. Init only setters provide a window to change state. That window closes when the construction phase ends. The construction phase effectively ends after all initialization, including property initializers and with-expressions have completed.
```

I think you meant to say "in addition to `get` accessors", right?  not, "in addition to `set` accessors". The `init` is only ever used in conjunction with `get`.

docs/csharp/whats-new/snippets/whats-new-csharp9/RecordsExamples.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+        // </SealedRecord>
+
+        public static void ExampleCode()
+        {
+            // <RecordsEquality>
+            Person p1 = new Person("Bill", "Wagner");
+            Student s1 = new Student("Bill", "Wagner", 11);
+
+            Console.WriteLine(s1 == p1); // false
```
IEvangelist, line 42:
```suggestion
            var person = new Person("Bill", "Wagner");
            var student = new Student("Bill", "Wagner", 11);

            Console.WriteLine(student == person); // false
```

Use `var` and improve variable naming

docs/csharp/whats-new/snippets/whats-new-csharp9/WeatherObservation.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NewCsharp9
+{
+    // <DeclareWeatherObservation>
+    public struct WeatherObservation
```
IEvangelist, line 6: Curious, could this be a `readonly struct` since it uses `init`?

docs/csharp/whats-new/snippets/whats-new-csharp9/WeatherObservation.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        public DateTime RecordedAt { get; init; }
+        public decimal TemperatureInCelsius { get; init; }
+        public decimal PressureInMillibars { get; init; }
+
+        public override string ToString() =>
+            $"At {RecordedAt.ToShortTimeString()} on {RecordedAt.ToShortDateString()}: Temp = {TemperatureInCelsius}, with {PressureInMillibars} pressure";
```
IEvangelist, line 13:
```suggestion
        public override string ToString() =>
            $"At {RecordedAt:h:mm tt} on {RecordedAt:M/d/yyyy}: " +
            $"Temp = {TemperatureInCelsius}, with {PressureInMillibars} pressure";
```

Concat string on new line for readability... trying to avoid horizontal scrolling, also used string interpolation format specifiers for date formatting to shorten it.

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+The preceding example for positional records demonstrates using an init-only setter to set a property using a with expression. You can declare init only setters in any type you write. For example, the following struct defines a weather observation structure:
+
+:::code language="csharp" source="snippets/whats-new-csharp9/WeatherObservation.cs" ID="DeclareWeatherObservation":::
+
+Callers can use property initializer syntax to set the values, while still preserving the immutability:
+
+:::code language="csharp" source="snippets/whats-new-csharp9/WeatherObservation.cs" ID="UseWeatherObservation":::
+
+But, changing an observation after initialization is an error:
```
IEvangelist, line 104: What error is caused?

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+```csharp
+// Error!
```
IEvangelist, line 107: Maybe comment with the `CS` error code here

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+        {
+            Console.WriteLine("Hello World!");
+        }
+    }
+}
+```
+
+There’s only one line of code that does anything. With top-level statements, you can replace all that boilerplate with the `using` statement and the single line that does the work:
+
+:::code language="csharp" source="snippets/whats-new-csharp9/Program.cs" ID="TopLevelStatements":::
```
IEvangelist, line 134: I'm curious if we should truly express this as a single line, in other words drop the `using System;` and fully qualify `Console`?

```csharp
System.Console.WriteLine("Hello World!");
```

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+Only one file in your application may use top-level statements. If the compiler finds top-level statements in multiple source files, it’s an error. It’s also an error if you combine top-level statements with a declared program entry point method, typically a `Main` method. In a sense, you can think that one file contains the statements that would normally be in the `Main` method of a `Program` class.  
+
+One of the most common uses for this feature is creating teaching materials. Beginner C# developers can write the canonical “Hello World!” in two lines of code. None of the extra ceremony is needed. However, seasoned developers will find many uses for this feature as well. Top-level statements enable a script-like experience for experimentation similar to what Jupyter notebooks provide. Top-level statements are great for small console programs and utilities. In addition, Azure functions are an ideal use case for top-level statements.
```
IEvangelist:
```suggestion
One of the most common uses for this feature is creating teaching materials. Beginner C# developers can write the canonical "Hello World!" in one (or two) lines of code. None of the extra ceremony is needed. However, seasoned developers will find many uses for this feature as well. Top-level statements enable a script-like experience for experimentation similar to what Jupyter notebooks provide. Top-level statements are great for small console programs and utilities. In addition, [Azure functions](/azure/azure-functions) are an ideal use case for top-level statements.
```

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+:::code language="csharp" source="snippets/whats-new-csharp9/PatternUtilities.cs" ID="IsLetterPattern":::
+
+Alternatively, with optional parentheses to make it clear that and has higher precedence than or:
+
+:::code language="csharp" source="snippets/whats-new-csharp9/PatternUtilities.cs" ID="IsLetterOrSeparatorPattern":::
+
+One of the most common uses is a new syntax for a null check:
+
+```csharp
+if (e is not null)
```
IEvangelist, line 164: Should you include previous ways that this was checked?

```csharp
if (!(e is null))
```

or 

```csharp
if (e != null)
```

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+Native sized integers, `nint` and `nuint`, are integer types. They are expressed by the underlying types <xref:System.IntPtr?displayProperty=nameWithType> and <xref:System.UIntPtr?displayProperty=nameWithType>. The compiler surfaces additional conversions and operations for these types as native ints. Native sized ints do not have constants for `MaxValue` or `MinValue`, except for `nuint.MinValue`, which has a `MinValue` of `0`. Other values cannot be expressed as constants because it would depend on the native size of an integer on the target machine. You can use constant values for `nint` in the range [`int.MinValue` .. `int.MaxValue`]. You can use constant values for `nuint` in the range [u`int.MinValue` .. `uint.MaxValue`]. The compiler performs constant folding for all unary and binary operators using the <xref:System.Int32?displayProperty=nameWithType> and <xref:System.UInt32?displayProperty=nameWithType> types. If the result does not fit in 32 bits, the operation is executed at runtime and is not considered a constant. Native sized integers can increase performance in scenarios where integer math is used extensively and needs to have the faster performance possible.
```
IEvangelist:
```suggestion
Native sized integers, `nint` and `nuint`, are integer types. They are expressed by the underlying types <xref:System.IntPtr?displayProperty=nameWithType> and <xref:System.UIntPtr?displayProperty=nameWithType>. The compiler surfaces additional conversions and operations for these types as native ints. Native sized ints do not have constants for `MaxValue` or `MinValue`, except for `nuint.MinValue`, which has a `MinValue` of `0`. Other values cannot be expressed as constants because it would depend on the native size of an integer on the target machine. You can use constant values for `nint` in the range [`int.MinValue` .. `int.MaxValue`]. You can use constant values for `nuint` in the range [`uint.MinValue` .. `uint.MaxValue`]. The compiler performs constant folding for all unary and binary operators using the <xref:System.Int32?displayProperty=nameWithType> and <xref:System.UInt32?displayProperty=nameWithType> types. If the result does not fit in 32 bits, the operation is executed at runtime and is not considered a constant. Native sized integers can increase performance in scenarios where integer math is used extensively and needs to have the fastest performance possible.
```

One of the `unit.MinValue` was incorrectly typed as "u`int.MinValue`". Also, fixed "faster", corrected as "fastest".

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,231 @@
...
+- Function pointers
+- Suppress emitting localsinit flag
+- Target-typed new expressions
+- static anonymous functions
+- Target-typed conditional expressions
+- Covariant return types
+- Lambda discard parameters
+- Attributes on local functions
+- Module initializers
+- New features for partial methods
```
Youssef1313, line 24: source generators is probably missing?

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+These features can improve performance in some scenarios. They should be used only after careful benchmarking both before and after adoption. Code involving native sized integers must be tested on multiple target platforms with different integer sizes. The other features require unsafe code.
+
+## Fit and finish features
+
+Many of the other features help you write code more efficiently. In C# 9.0, you can omit the type in a new expression when the created object’s type is already known. The most common use is in field declarations:
```
IEvangelist, line 186:
```suggestion
Many of the other features help you write code more efficiently. In C# 9.0, you can omit the type in a new expression when the created object's type is already known. The most common use is in field declarations:
```

docs/csharp/whats-new/snippets/whats-new-csharp9/FitAndFinish.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+
+    }
+
+    public class WeatherStation
+    {
+        public string Location { get; init; }
+
+        // <WeatherStationField>
+        private List<WeatherObservation> observations = new();
```
IEvangelist, line 21:
```suggestion
        private List<WeatherObservation> _observations = new();
```

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,231 @@
...
+- Module initializers
+- New features for partial methods
+
+C# 9.0 is supported on **.NET 5**. For more information, see [C# language versioning](../language-reference/configure-language-version.md).
+
+## Record types
+
+C# 9.0 introduces ***record types***, which are a reference type that provides synthesized methods to provide value semantics for equality. Records are immutable by default.
+
+Record types make it easy to create immutable reference types in .NET. Historically, .NET types are largely classified as reference types (including classes and anonymous types) and value types (including structs and tuples). While immutable value types are recommended, mutable value types don’t often introduce errors. Value types are passed by value to methods so changes are made to a copy of the original data.
```
Youssef1313, line 32: nit:

```suggestion
Record types make it easy to create immutable reference types in .NET. Historically, .NET types are largely classified as reference types (including classes and anonymous types) and value types (including structs and tuples). While immutable value types are recommended, mutable value types don't often introduce errors. Value types are passed by value to methods so changes are made to a copy of the original data.
```

gewarren:
> > The CI build is failing because `PrintMembers` didn't make the preview 8 build. It is in RC1, when this build error will go away.
> 
> I don't think we should knowingly introduce a build warning. It means everyone will have to check the build report for every PR.

Oh sorry, I misunderstood. If the error is only in the CI build, it may be okay. I thought it was an xref link.

docs/csharp/whats-new/snippets/whats-new-csharp9/FitAndFinish.cs
```diff
@@ -0,0 +1,50 @@
...
+            return new();
+        }
+    }
+
+    public class FitAndFinish
+    {
+        public static void Examples()
+        {
+            // <InitWeatherStation>
+            WeatherStation station = new() { Location = "Seattle, WA" };
```
IEvangelist, line 38: I had an ah-ha moment when looking at this, and I am VERY curious - this looks a lot like object initializers syntax, but omitting the type. With that said, in object initializers you could omit the `()` on the object, for example:

Code like this:

```csharp
WeatherStation station = new WeatherStation() { Location = "Seattle, WA" };
```

Could omit the `()`, like this:

```csharp
WeatherStation station = new WeatherStation { Location = "Seattle, WA" };
```

I'm curious, can we do the same thing now with the `new` syntax?

```csharp
WeatherStation station = new { Location = "Seattle, WA" };
```

I'm wondering because this looks like an anonymous object, instead of the `WeatherStation`... but wow, that would be really cool!

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+A similar feature improves the target type resolution of conditional expressions. With this change, the two expressions need not have an implicit conversion from one to the other, but may both have implicit conversions to a common type. You likely won’t notice this change. What you will notice is that some conditional expressions that previously required casts or wouldn’t compile now just work.
+
+Starting in C# 9.0, you can add the `static` modifier to lambda expressions or anonymous methods. That has the same effect as the `static` modifier on local functions: a static lambda or anonymous function cannot capture local variables or instance state. This prevents accidentally capturing other variables.
+
+Covariant return types enables a function that overrides a base class virtual function to change the return type to a type derived from the type declared in the base class. This can be useful for Records, and for other types that support virtual clone or factory methods.
+
+Next, you can use discards as parameters to lambda expressions. This convenience enables you to avoid naming the argument, and the compiler may be able to avoid using it. You use the `_` for any argument.
```
IEvangelist: Might be worth mentioning that it was possible to use `_` before for arguments, but you could only use one as it would simple be the name of the argument instance rather than a discard. With this additional functionality you can use more than one `_` and they're true discards, rather than a single variable named `_`.

docs/csharp/whats-new/csharp-9.md
Youssef1313: I think that *technically*, both reference types and value types are passed by value. However, the *value* of a passed reference type is a pointer in the heap.

The last sentence probably implies that reference types are passed by reference by default (which seem incorrect to me).

Certainly, the sentence doesn't say that reference types are passed by reference. So the sentence is completely correct. However, I think readers may interpret it incorrectly.

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,231 @@
...
+
+There are many advantages to immutable reference types as well. These advantages are more pronounced in concurrent programs with shared data. Unfortunately, C# forced you to write quite a bit of extra code to create immutable reference types. Records provide a type declaration for an immutable reference type that uses value semantics for equality. The synthesized methods for equality and hash codes consider two records equal if their properties are all equal. Consider this definition:
+
+:::code language="csharp" source="snippets/whats-new-csharp9/RecordsExamples.cs" ID="RecordDefinition":::
+
+That record definition creates a `Person` type that contains two readonly properties: `FirstName` and `LastName`. The `Person` type is a reference type. If you looked at the IL, it’s a class. It’s immutable in that none of the properties can be modified once it's been created. When you define a record type, the compiler synthesizes several other methods for you:
```
Youssef1313, line 38:
```suggestion
That record definition creates a `Person` type that contains two readonly properties: `FirstName` and `LastName`. The `Person` type is a reference type. If you looked at the IL, it's a class. It's immutable in that none of the properties can be modified once it's been created. When you define a record type, the compiler synthesizes several other methods for you:
```

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,233 @@
...
+
+The two features added for code generators are extensions to ***partial method syntax***, and ***module initializers***. First, the changes to partial methods. Before C# 9.0, partial methods are `private` but cannot specify an access modifier, have a `void` return, and cannot have `out` parameters. These restrictions meant that if no method implementation is provided, the compiler removes all calls to the partial method. C# 9.0 removes these restrictions, but requires that partial method declarations have an implementation. Code generators can provide that implementation. To avoid introducing a breaking change, the compiler considers any partial method without an access modifier to follow the old rules. If the partial method includes the `private` access modifier, the new rules govern that partial method.
+
+The second new feature for code generators is ***module initializers***. These are methods that have the <xref:System.Runtime.CompilerServices.ModuleInitializer> attribute attached to them. These methods will be called by the runtime when the assembly loads. A module initializer method:
```
IEvangelist: The `<xref:System.Runtime.CompilerServices.ModuleInitializer>` is not yet available, we may want to remove the xref for now?

IEvangelist (approved):
This is awesome, love the way that this came together and ❤️ all the C# 9 features. I had a few suggestions and questions, but otherwise - :shipit:

docs/csharp/whats-new/csharp-9.md
IEvangelist: Source Generators aren't a C# language feature. They are a new capability.

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,231 @@
...
+- Records have a consistent string representation generated for you.
+- Records support copy construction. Correct copy construction must include inheritance hierarchies, and properties added by developers.
+- Records can be copied with modification. These copy and modify operations supports non-destructive mutation.
+- All records support deconstruction.
+
+In addition to the familiar `Equals` overloads, `operator ==` and `operator !=`, the compiler synthesizes a new `EqualityContract` property. The property returns a `Type` object that matches the type of the record. If the base type is `object`, the property is `virtual`. If the base type is another record type, the property is an `override`. If the record type is `sealed`, the property is `sealed`. The synthesized `GetHashCode` uses the `GetHashCode` from all the public properties declared in the base type and the record type. These synthesized methods enforce value-based equality throughout an inheritance hierarchy. That means a `Student` will never be considered equal to a `Person` with the same name. The types of the two records must match as well as all properties shared among the record types being equal.
```
Youssef1313, line 62: nit:

```suggestion
In addition to the familiar `Equals` overloads, `operator ==`, and `operator !=`, the compiler synthesizes a new `EqualityContract` property. The property returns a `Type` object that matches the type of the record. If the base type is `object`, the property is `virtual`. If the base type is another record type, the property is an `override`. If the record type is `sealed`, the property is `sealed`. The synthesized `GetHashCode` uses the `GetHashCode` from all the public properties declared in the base type and the record type. These synthesized methods enforce value-based equality throughout an inheritance hierarchy. That means a `Student` will never be considered equal to a `Person` with the same name. The types of the two records must match as well as all properties shared among the record types being equal.
```

Youssef1313, line 62: I believe GetHashCode accounts for the synthesized EqualityContract property as well, not only the properties declared by the developer.

Youssef1313: Note: the text says GetHashCode uses the public properties.
But I used a private property in SharpLab and it seems GetHashCode uses this private property. Will that change in the official release? If not, the text should be updated.

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,231 @@
...
+
+Records also have a synthesized constructor and a "clone" method for creating copies. The synthesized constructor has one argument of the record type. It produces a new record with the same values for all properties of the record. This constructor is private if the record is sealed, otherwise it's protected. The synthesized "clone" method supports copy construction for record hierarchies. The term "clone" is in quotes because the actual name is compiler generated. You can't create a method named `Clone` in a record type. The synthesized "clone" method returns the type of record being copied using virtual dispatch. If a record type is `abstract`, the clone method is also `abstract`. If a record type is `sealed`, the clone method is `sealed`. If the base type of the record is `object`, the clone method is `virtual`. Otherwise, it's `override`. The result of all these rules is the equality is implemented consistently across any hierarchy of record types. Two records are equal to each other if their properties are equal and their types are the same, as shown in the following example:
```
Youssef1313, line 64: > If a record type is `sealed`, the clone method is `sealed`.

How a method can be sealed if it's non-`overrides`?

Youssef1313, line 64: > If the base type of the record is `object`, the clone method is `virtual`.

```
   [System.Runtime.CompilerServices.NullableContext(1)]
    public Person <Clone>$()
    {
        return new Person(this);
    }
```

I'm not seeing virtual in SharpLab.
The record I declared doesn't inherit any other time. Hence, the base type is object.

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,231 @@
...
+
+:::code language="csharp" source="snippets/whats-new-csharp9/RecordsExamples.cs" ID="RecordsEquality":::
+
+The compiler synthesizes two methods that support printed output: a <xref:System.ToString> override, and `PrintMembers`. The `PrintMembers` method returns a comma-separated list of property names and values. The <xref:System.ToString> override returns the string produced by `PrintMembers`, surrounded by `{` and `}`. For example, the <xref:System.ToString> method for `Student` returns a `string` like the following code:
```
Youssef1313, line 68: This is the PrintMembers method I'm seeing on SharpLab, it's private, returns true always, and it receives a StringBuilder to fill, this is confusing a little bit:


```
    [System.Runtime.CompilerServices.NullableContext(1)]
    private bool PrintMembers(StringBuilder builder)
    {
        builder.Append("FirstName");
        builder.Append(" = ");
        builder.Append((object)FirstName);
        return true;
    }
```

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,231 @@
...
+
+```text
+Student { LastName = Wagner, FirstName = Bill, Level = 11 }
+```
+
+The examples shown so far use traditional syntax to declare properties. There’s a more concise form called ***positional records***.  Here are the three record types defined earlier as positional records:
```
Youssef1313, line 74: nit:

```suggestion
The examples shown so far use traditional syntax to declare properties. There's a more concise form called ***positional records***.  Here are the three record types defined earlier as positional records:
```

docs/csharp/whats-new/snippets/whats-new-csharp9/WeatherObservation.cs
BillWagner: no.

docs/csharp/whats-new/snippets/whats-new-csharp9/FitAndFinish.cs
Youssef1313:
```csharp
using System;

Console.WriteLine();

Student x = new { FirstName = "Youssef" };

public class Student
{
}
```

>    error CS0029: Cannot implicitly convert type '<anonymous type: string FirstName>' to 'Student'

Looks like the parentheses cannot be omitted.

docs/csharp/whats-new/csharp-9.md
BillWagner: Which branch are you using?  It's virtual [here](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQBMQGoA+ABAzAAgTgGMB7BFfABUSlIDsBYAKAG8X9P89uBGABnwAxAJYJYAOQhg4+VgHM4MANz4Avhy49sA/ABloMKTLn5FK9Zs48a4hgAodggGZjYSPoIA2hgJT4AXgA+fCsuTntRcSNpOA8DSVj/APx7V2iPH1hfZRY1IA===)

Youssef1313: @BillWagner, I was using master. Unfortunately, I cannot repro.
I don't remember the exact code I wrote ;-(

Youssef1313: @BillWagner Alright, I found it. The virtual isn't applied if the record is sealed even if the base type is object.

I was playing around in SharpLab while reading the article, so I had the record sealed from the previous sentence :smile:

docs/csharp/whats-new/csharp-9.md
BillWagner: I thought about both of those.  There were so many ways to do null checks, I didn't want to list them. Consider these alternatives:
```csharp
// 1. 
_ = e ?? throw null;
// 2. 
if (e is object)
```

As for the operator overloading, I'll add that to my list for updating `null` check guidance for C# 9. (I'll be creating issues later today.)

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,237 @@
...
+
+Records also have a synthesized constructor and a "clone" method for creating copies. The synthesized constructor has one argument of the record type. It produces a new record with the same values for all properties of the record. This constructor is private if the record is sealed, otherwise it's protected. The synthesized "clone" method supports copy construction for record hierarchies. The term "clone" is in quotes because the actual name is compiler generated. You can't create a method named `Clone` in a record type. The synthesized "clone" method returns the type of record being copied using virtual dispatch. If a record type is `abstract`, the clone method is also `abstract`. If a record type is `sealed` and its base type isn't `object`, the clone method is `sealed`. If the base type of the record is `object`, the clone method is `virtual` if the record type isn't `sealed`. Otherwise, it's `override`. The result of all these rules is the equality is implemented consistently across any hierarchy of record types. Two records are equal to each other if their properties are equal and their types are the same, as shown in the following example:
```
Youssef1313, line 64: > If the base type of the record is `object`, the clone method is `virtual` if the record type isn't `sealed`. Otherwise, it's `override`.

Does "otherwise" refer to the first "if" (base type is object), or the second if "record type isn't sealed".

For a sealed record whose base type is object, I'm getting Clone as only public.
The sentence in all cases seems to contradict with this, I think?

https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQBMQGoA+ABAzAAijggBs4V8E4BjAewQoAVEpaA7AWACgBvbgXyA

Adding a repro this time :smile:

BillWagner: Good catch. Let me re-write this set of sentences. (I think it is just too close to the spec language, and isn't strict enough)

docs/csharp/whats-new/csharp-9.md
```diff
@@ -61,7 +61,17 @@ The compiler synthesizes different versions of the methods above. The method sig
...
+
+- If the record type is `abstract`, the "clone" method is also `abstract`.
+- For record types that aren't `abstract` when the base type is `object`:
+  - If the record is `sealed`, no additional modifiers are added to the "clone" method (meaning it is not `virtual`).
+  - If the record isn't `sealed`, the "clone" method is `virtual`.
+- For record types that aren't `abstract` when the base type is not `object`:
+  - If the record is `sealed`, the "clone" method is also `sealed`.
+  - If the record isn't `sealed`, the "clone" method is `override`.
```
Youssef1313, line 72: A final bit, if it's abstract and the base type is not object, clone method is both "abstract" and "override"

https://sharplab.io/#v2:EYLgtghgzgLgpgJwD4AEDMACFAmDBhAWACgBvYgX2OPQwmFgQgGMYME4mB7BAEwwEFiZIpSLVMdBs1bsuvDACEMIAUIpA===

```
    [System.Runtime.CompilerServices.NullableContext(1)]
    public abstract override A <Clone>$();

    protected B([System.Runtime.CompilerServices.Nullable(1)] B original)
        : base(original)
    {
    }
```

docs/csharp/whats-new/csharp-9.md
BillWagner: I'm going to leave that out of the what's new. It will be covered in the reference area.

docs/csharp/whats-new/csharp-9.md
BillWagner: My bad. It should be `ModuleInitializerAttribute` (for the class name)

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,247 @@
...
+- Function pointers
+- Suppress emitting localsinit flag
+- Target-typed new expressions
+- static anonymous functions
+- Target-typed conditional expressions
+- Covariant return types
+- Lambda discard parameters
+- Attributes on local functions
+- Module initializers
+- New features for partial methods
```
jcouv, line 24: The "Extension GetEnumerator" feature is missing

docs/csharp/whats-new/csharp-9.md
```diff
@@ -0,0 +1,247 @@
...
+
+C# 9.0 is supported on **.NET 5**. For more information, see [C# language versioning](../language-reference/configure-language-version.md).
+
+## Record types
+
+C# 9.0 introduces ***record types***, which are a reference type that provides synthesized methods to provide value semantics for equality. Records are immutable by default.
```
HaloFour, line 30: I know that this has already been merged but I have concerns about the heavy emphasis on the immutability of "records" in C# when there really is little in the language that enforces or even encourages that they can't be mutated.  I feel that this emphasis will lead to a very false and confusing expectation.

BillWagner: @HaloFour This is being discussed in #20601. Reading your comment, you may have better ideas for updating the description. I'll be watching that issue discussion before I make an update to this article.

PR #20447, opened 2020-09-04
Fix out of date explanation of named arguments

xTyler73:
the old sentence: "However, a positional argument cannot follow a named argument" is wrong as of C# 7.2

https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7-2#non-trailing-named-arguments
states: "Method calls may now use named arguments that precede positional arguments when those named arguments are in the correct positions."


BillWagner (approved):
Thanks for spotting this and fixing it @xTyler73 I'd missed this sentence when I updated docs for the 7.2 changes.

Thanks again for helping us improve our docs!

PR #20451, opened 2020-09-04
Initial commit

robvet:
## Summary

Pushing directory structure and inital documents as a template for new book.

Including TOC, Index, and the MDs for the first two chapters from Cloud Native.


nishanil:
Any reason why you are adding markdowns from cloud-native book?

robvet:
> Any reason why you are adding markdowns from cloud-native book?

Only as templates from which to start writing.  I can blank them out if you'd like.

nishanil:
Yes, please add pages/content only when required. Just TOC and index will be good enough to start with. You can update the TOC as you add pages and chapters.
Please check the lint errors.

nishanil:
Should it be "DAPR" or "Dapr"?

nishanil:
@robvet Can you fix warnings please?

PR #20467, opened 2020-09-04
Minor enhancements to the Microsoft.Data.Sqlite docs

bricelam:
Changes:

- Reccommend using a shared cache to mitigate the lack of async
- List the new SQLitePCLRaw.bundle_sqlite3 package

/cc @ajcvickers


IEvangelist (approved):

PR #20481, opened 2020-09-06
Add missing using

perlun:
The example was incomplete; `System.Threading.Thread` was being referenced without the namespace being properly imported.

### DIFF
--- a/samples/snippets/standard/assembly/unloading/complex_assemblyloadcontext.cs
+++ b/samples/snippets/standard/assembly/unloading/complex_assemblyloadcontext.cs
@@ -1,20 +1,26 @@
-﻿class TestAssemblyLoadContext : AssemblyLoadContext
+﻿using System.Reflection;
+using System.Runtime.Loader;
+
+namespace complex
 {
-    private AssemblyDependencyResolver _resolver;
-
-    public TestAssemblyLoadContext(string mainAssemblyToLoadPath) : base(isCollectible: true)
+    class TestAssemblyLoadContext : AssemblyLoadContext
     {
-        _resolver = new AssemblyDependencyResolver(mainAssemblyToLoadPath);
-    }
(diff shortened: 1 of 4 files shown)


IEvangelist:
Hi 👋 @perlun
Thank you so much for posting this issue. We really appreciate it. Could you please add a C# project to the *unloading* directory:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <StartupObject>test.Program</StartupObject>

</PropertyGroup>
```

We have a build validation that requires a project to compile from.

perlun:
@IEvangelist No probs, sounds good so we can avoid compilation issues in these examples going forward. I added a `.csproj` file, let's hope it passes CI runs this time.

perlun:
This was bit of a can of worms actually, but I think I have most problems resolved now. After adding the `.csproj` file, some other compilation errors were discovered. Opening the project in JetBrains Rider didn't work so well, since I didn't have the .NET 5 preview installed. After installing it, things worked much smoother and I could fix most of these, _but_...

...the `TestAssemblyLoadContext` name is being overloaded in two of these examples, which aint going to work. One of them would need to be renamed. Which one should I rename? Are there any references to the class names there (in the actual docs) that has to be fixed as well?

(Btw, https://dotnetcli.blob.core.windows.net is really slow where I am (400 KiB/s on a 100 Mbit connection), would be good to put up this behind a proper CDN like I presume the official downloads are. :sunglasses:)

IEvangelist:
Good to know, I'd be good with you adding a `namespace` for each of these.

perlun:
@IEvangelist 

> Good to know, I'd be good with you adding a `namespace` for each of these.

You've got it, compiles fine locally now. :+1:

IEvangelist (approved):

PR #20486, opened 2020-09-07
Delete outdated information

Youssef1313:


docs/csharp/language-reference/compiler-messages/cs0106.md
```diff
@@ -14,8 +14,6 @@ The modifier 'modifier' is not valid for this item
 
  A class or interface member was marked with an invalid access modifier. The following examples describe some of these invalid modifiers:
 
-- The [static](../keywords/static.md) modifier is not permitted on a [local function](../../programming-guide/classes-and-structs/local-functions.md).
```
Youssef1313, line 17: Note: I didn't update this with the version that allowed the usage of static function because the compiler shouldn't generate CS0106, but another error saying that the feature isn't allowed in version X but allowed in version Y and later.

svick:
What is the policy on documenting the behavior of older compilers? This page isn't versioned, so if I'm using an older version of the compiler (probably because I'm using an older version of VS), this is still the documentation for me.

It would be especially odd if this page kept the note about not allowing `static class` before VS 2005, but not mention at all what VS 2017 does. (I haven't actually tested VS 2017, but I base that on compiling with the corresponding version of the Roslyn NuGet package.)

Youssef1313:
@svick Hmm, that sounds very reasonable.
Probably change the wording to something like: "If you got this error with static local functions, make sure you use LangVersion 8 or higher"

The diagnostic area in the documentation needs a lot a lot of work ;-(

docs/csharp/language-reference/compiler-messages/cs0106.md
IEvangelist: Perhaps we should call out the versions of C# that it was not permitted in? I think this is a bit misleading as `static` was only recently added to local functions. We should be clear to state that here.

IEvangelist (changes requested):
Instead of removing this, version it. Call out which versions and maybe even a link to the conceptual docs.

docs/csharp/language-reference/compiler-messages/cs0106.md
Youssef1313: @IEvangelist The diagnostic for using static local function depends on the compiler version.
If the compiler version supports C# 8.0, but the used version is lower, the diagnostic message will **not** be CS0106, but a diagnostic that suggests upgrading to C# 8.0.

However, if the compiler version doesn't at all support C# 8.0, the diagnostic message will be CS0106.

I updated the sentence to describe that, but I'm not sure if you'll like a long sentence to achieve this level of accuracy, or be less accurate with a smaller sentence. I'm okay either way.

IEvangelist (approved):

PR #20536, opened 2020-09-09
Enhance docs\core\tutorials\index.md

johypark:
## Summary
Add links for macOS.

### DIFF
--- a/docs/core/tutorials/index.md
+++ b/docs/core/tutorials/index.md
@@ -41,7 +41,6 @@ Choose these tutorials if you want to use Visual Studio Code or some other code 
 - [Create a class library](library-with-visual-studio-mac.md)
 - [Unit test a class library](testing-library-with-visual-studio-mac.md)
 - [Install and use a package](/nuget/quickstart/install-and-use-a-package-in-visual-studio-mac)
-- [Create and publish a package](/nuget/quickstart/create-and-publish-a-package-using-the-dotnet-cli)
 - [Create an F# console app](../../fsharp/get-started/get-started-with-visual-studio-for-mac.md)
 
 ## Advanced topics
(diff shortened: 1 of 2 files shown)


docs/core/tutorials/index.md
```diff
@@ -36,8 +36,13 @@ Choose these tutorials if you want to use Visual Studio Code or some other code
 ## Use Visual Studio for Mac
 
 - [Create a console app](with-visual-studio-mac.md)
-- [Create a library and unit test it](library-with-visual-studio-mac.md)
+- [Debug an app](debugging-with-visual-studio-mac.md)
+- [Publish an app](publishing-with-visual-studio-mac.md)
+- [Create a class library](library-with-visual-studio-mac.md)
+- [Unit test a class library](testing-library-with-visual-studio-mac.md)
 - [Install and use a package](/nuget/quickstart/install-and-use-a-package-in-visual-studio-mac)
+- [Create and publish a package](/nuget/quickstart/create-and-publish-a-package-using-the-dotnet-cli)
```
tdykstra, line 44:
```suggestion
```

tdykstra (approved):
Thanks! One link I think we should omit because it's not actually about VS Mac.

docs/core/tutorials/index.md
JohyPark: @tdykstra Thank you for reviewing!

image
There's an issue that redirecting _create-and-publish-a-package-using-the-dotnet-cli_ If click that.

tdykstra: Good point. I put it there in the TOC because there isn't a corresponding VS Mac "create and publish" tutorial, so this is the best match we have to offer to make the TOC nodes for VS Mac correspond to the other tools' topics.  On second thought now that you mention it, maybe it would be better to omit the CLI link here too.

JohyPark: @tdykstra Thank you for your answer!

JohyPark: @tdykstra I updated TOC either, please check it before merging.
Thank you.

tdykstra (approved):
Looks good, thanks!

PR #20539, opened 2020-09-09
Fix build warnings due to en-us locale

Youssef1313:
The bot needs to be updated to avoid this. It also should probably be saying "GitHub" instead of "github".


CamSoper:
@Youssef1313 They told me they fixed the bot. I'll follow up.

CamSoper (approved):

PR #20600, opened 2020-09-11
Update example to explain CS0119

### LINKED ISSUE #20558: Interesting case to explain for CS0119
I believe the example stated in https://github.com/dotnet/roslyn/issues/47589 should be added to the doc page. The doc page will need to be updated again if https://github.com/dotnet/roslyn/issues/47589 is fixed to indicate which version has this fix.

---
#### Document Details

* ID: eec72ac2-345b-8257-9765-426726d3ac40
* Version Independent ID: 8f8bdfec-3503-2b36-8d67-aa4dbf253c3c
* Content: [Compiler Error CS0119](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0119)
* Content Source: [docs/csharp/misc/cs0119.md](https://github.com/dotnet/docs/blob/master/docs/csharp/misc/cs0119.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

Marusyk:
Update example to explain CS0119

Fixes #20558

### DIFF
--- a/docs/csharp/misc/cs0119.md
+++ b/docs/csharp/misc/cs0119.md
@@ -19,8 +19,7 @@ ms.assetid: 048924f1-378f-4021-bd20-299d3218f810
 - A method identifier is used as if it were a struct or class  
   
 ## Example  
- The following sample generates CS0119: 'C.B()' is a method, which is not valid in the given context 
-
+ The following sample generates CS0119: 'C.B()' is a method, which is not valid in the given context.
 ```csharp   
 public static class B
 {
--- a/docs/csharp/misc/cs0119.md
+++ b/docs/csharp/misc/cs0119.md
@@ -19,20 +19,23 @@ ms.assetid: 048924f1-378f-4021-bd20-299d3218f810
 - A method identifier is used as if it were a struct or class  
   
 ## Example  
- The following sample generates CS0119: 'C.B()' is a method, which is not valid in the given context. You can fix this error by changing the name of the method `C.B`, or using the fully qualified name for the class `B`.
+ The following sample generates CS0119: 'C.B()' is a method, which is not valid in the given context. You can fix this error by changing the name of the method `C.B`, or using the fully qualified name for the class `B` like `N2.B`.
 
 ```csharp
-public static class B
+namespace N2
 {
-    public static void X() {}
+    public static class B
+    {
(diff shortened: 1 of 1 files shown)


docs/csharp/misc/cs0119.md
```diff
@@ -19,18 +19,20 @@ ms.assetid: 048924f1-378f-4021-bd20-299d3218f810
 - A method identifier is used as if it were a struct or class  
   
 ## Example  
- The following sample generates CS0119.  
+ The following sample generates CS0119: 'C.B()' is a method, which is not valid in the given context
```
Youssef1313, line 22: I think this needs more context/explanation rather than giving just the code.

@BillWagner Is this case stated somewhere in the language spec. That would be a good reference here.

BillWagner: This error is related to the rules on binding, and overload resolution. Those rules get complicated very quickly, so I wouldn't put a reference here. See me other review comment.

docs/csharp/misc/cs0119.md
```diff
@@ -19,18 +19,20 @@ ms.assetid: 048924f1-378f-4021-bd20-299d3218f810
...
-using System;  
-public class MyClass
-{  
-   public static void Test() {}  
-  
-   public static void Main()  
-   {  
-      Console.WriteLine(Test.x);   // CS0119  
-   }  
+ The following sample generates CS0119: 'C.B()' is a method, which is not valid in the given context.
```
BillWagner, line 22: The important addition is how a developer can fix this error. I'd add the following:

```suggestion
 The following sample generates CS0119: 'C.B()' is a method, which is not valid in the given context. You can fix this error by changing the name of the method `C.B`, or using the fully qualified name for the class `B`.
```

To make this explanation more clear, it would be helpful to put `B` (the class) in a namespace like `N2`. Then, you can say "...for the class `N2.B`."

BillWagner (changes requested):
Thanks for adding this @Marusyk I had one suggestion, and then this would be ready to merge.

BillWagner (approved):

BillWagner:
Thanks again @Marusyk 
I'll :shipit: now

PR #20607, opened 2020-09-12
Add new troubleshoot note

### LINKED ISSUE #19564: Include x64-only warning in the troubleshooting section, not only at the top of the doc
It's common to skip through install pages IMO: e.g. if I'm installing on Ubuntu, I'm just looking for a command to set up the repo, then another command to install the package from the repo. Anything else can be skipped and read later if I need to, but I just want to get the install done. However, if I hit a problem because I'm running on ARM64, I go to the troubleshooting section, not re-read the doc from the top. So I think this info should go there where it'll have a better chance of being seen.

---
#### Document Details

* ID: 80707602-f283-e796-d3b8-3287fbc011dc
* Version Independent ID: 1d732b15-9e64-633a-000c-ad96d9d3818a
* Content: [Install .NET Core on Ubuntu - .NET Core](https://docs.microsoft.com/en-us/dotnet/core/install/linux-ubuntu)
* Content Source: [docs/core/install/linux-ubuntu.md](https://github.com/dotnet/docs/blob/master/docs/core/install/linux-ubuntu.md)
* Product: **dotnet-core**
* GitHub Login: @adegeo
* Microsoft Alias: **adegeo**

adegeo:
Fixes #19564

Add to troubleshooting section to add info about  x64


BillWagner (approved):
This LGTM @adegeo 

You can :shipit: when ready.

PR #20649, opened 2020-09-15
Snippets config: Remove CPP triggers; Update to .NET 5 RC 1

adegeo:

### DIFF
--- a/.github/workflows/build-validation.yml
+++ b/.github/workflows/build-validation.yml
@@ -9,19 +9,16 @@ on:
       - "**.cs"
       - "**.vb"
       - "**.fs"
-      - "**.cpp"
-      - "**.h"
       - "**.xaml"
       - "**.razor"
       - "**.cshtml"
       - "**.vbhtml"
-      - "**.*proj"
       - "**global.json"
       - "**snippets.5000.json"
     branches: [ master ]
 
 env:
-  DOTNET_INSTALLER_CHANNEL: 'release/5.0.1xx-preview8'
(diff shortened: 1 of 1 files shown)


.github/workflows/build-validation.yml
```diff
@@ -9,19 +9,16 @@ on:
       - "**.cs"
       - "**.vb"
       - "**.fs"
-      - "**.cpp"
-      - "**.h"
       - "**.xaml"
       - "**.razor"
       - "**.cshtml"
       - "**.vbhtml"
-      - "**.*proj"
```
BillWagner, line 18: We should keep the `csproj`, `vsproj`, and `vbproj` so that we build projects when dependencies change.

adegeo: DUH on me.

.github/workflows/build-validation.yml
```diff
@@ -9,19 +9,19 @@ on:
...
+      - "**.csproj"
+      - "**.vbproj"
+      - "**.fsproj"
       - "**global.json"
       - "**snippets.5000.json"
     branches: [ master ]
 
 env:
-  DOTNET_INSTALLER_CHANNEL: 'release/5.0.1xx-preview8'
+  DOTNET_INSTALLER_CHANNEL: 'release/5.0.1xx-preview7'
```
pkulikov, line 24: RC1 is released; would roll forward work?

BillWagner (approved):
This LGTM @adegeo 

I wonder about @pkulikov 's comment as well. Is it right just to move to RC 1?

adegeo:
I believe this will work. I'll test here and then do the edits in the other repos

PR #20652, opened 2020-09-15
fix: Docs Authoring Markdownlint rules

nschonni:
## Summary

I'm trying to split out the extra rules in https://github.com/microsoft/vscode-docs-authoring/pull/747
So I was running the rules from the commandline and wanted to pass along the ones I didn't think were false positives


adegeo:
I don't understand those rules considering the xref links work unescaped. 
🤷‍♂️

adegeo (approved):

nschonni:
Yeah, the `*` and backtick work in the URL bar, but I think the hash sign doesn't make sense escaped because it's used for headings. I left those ones out

PR #20736, opened 2020-09-22
Install-Scripts - Removing out of support `coherent` version option

bekir-ozturk:
Related issue https://github.com/dotnet/install-scripts/issues/64

Install scripts no longer allow downloading coherent versions and we have decided to remove that feature.
This PR updates the documentation to remove the section related to `coherent` version option.

### DIFF
--- a/docs/core/tools/dotnet-install-script.md
+++ b/docs/core/tools/dotnet-install-script.md
@@ -1,7 +1,7 @@
 ---
 title: dotnet-install scripts
 description: Learn about the dotnet-install scripts to install the .NET Core SDK and the shared runtime.
-ms.date: 04/30/2020
+ms.date: 09/22/2020
 ---
 # dotnet-install scripts reference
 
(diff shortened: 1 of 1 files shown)


docs/core/tools/dotnet-install-script.md
```diff
@@ -179,7 +179,6 @@ The install scripts do not update the registry on Windows. They just download th
   Represents a specific build version. The possible values are:
 
   - `latest` - Latest build on the channel (used with the `-Channel` option).
-  - `coherent` - Latest coherent build on the channel; uses the latest stable package combination (used with Branch name `-Channel` options).
```
KathleenDollard, line 182: We are dropping the coherent feature. We do not think it works correctly and it has not been a reasonable value for some time. 

Docs folks: Is it OK for us to just drop it, or do we need a note explaining the date on which we dropped it?

tdykstra: OK to just drop.

tdykstra (approved):

PR #20761, opened 2020-09-23
Single file documentation update

LakshanF:
## Summary

Additional API usage information for single-file and the non-bundling of of pdb files. Also added comments on the inability now with rare use of pinvoking into coreclr

https://github.com/dotnet/runtime/issues/38405


docs/core/deploying/single-file.md
```diff
@@ -11,9 +11,26 @@ Bundling all application-dependent files into a single binary provides an applic
...
+There are some caveats that you need to be aware for single-file use, first of which is the use of path information to locate a file relative to the location of your application. The <xref:System.Reflection.Assembly.Location?displayProperty=nameWithType> API will return an empty string, which is the default behavior for assemblies loaded from memory. The compiler will give a warning for this API during build time to alert the developer to the specific behavior. If the path to the application directory is needed, the <xref:System.AppContext.BaseDirectory?displayProperty=nameWithType> API will return the directory where the AppHost (the single-file bundle itself) resides. The table at the bottom of this section has runtime library API behavior in single-file use.
```
jkotas, line 16: > The compiler will give a warning for this API during build time to alert the developer to the specific behavior

Is it the case in .NET 5?

docs/core/deploying/single-file.md
```diff
@@ -11,9 +11,26 @@ Bundling all application-dependent files into a single binary provides an applic
...
+
+Managed C++ applications aren't well suited for single-file deployment and we recommend that you write applications in C# or another non-managed C++ language to be single-file compatible. There are some rare cases where a managed library takes a dependency into the runtime library, coreclr, via `pinvoke`. This will not be possible in single-file case since the runtime library is bundled with the application.
```
jkotas, line 20: Managed C++ application -> managed C++ components

jkotas, line 20: >  There are some rare cases where a managed library takes a dependency into the runtime library, coreclr, via `pinvoke`. This will not be possible in single-file case since the runtime library is bundled with the application.

I do not think we need to talk about this. Applications taking dependencies on coreclr internals is unsupported in general.

docs/core/deploying/single-file.md
```diff
@@ -11,9 +11,26 @@ Bundling all application-dependent files into a single binary provides an applic
...
+
+| API                                                                                     | Notes |
+| ---------------------------------------------------------------------------------------  | ------- |
+| Assembly.Location | Returns an empty string, use AppContext.BaseDirectory |
+| Module.FullyQualifiedName  | Returns a `<Unknown>` string ((just like it does in no-single-file for memory-loaded assemblies) |
+| Assembly.GetFile  | Will throw `IOException` (just like it does in no-single-file for memory-loaded assemblies) |
+| Assembly.GetFiles  | Will throw `IOException` (just like it does in no-single-file for memory-loaded assemblies) |
+| Assembly.CodeBase  | Will throw `PlatformNotSupportedException` (this API is obsolete in .NET Core 5.0) |
+| Assembly.EscapedCodeBase  | Will throw `PlatformNotSupportedException` (this API is obsolete in .NET Core 5.0) |
+| AssemblyName.CodeBase  | Will throw `PlatformNotSupportedException` |
```
jkotas, line 32: I do not think that's correct. I believe AssemblyName returns empty string for assemblies loaded from memory.

docs/core/deploying/single-file.md
```diff
@@ -11,9 +11,26 @@ Bundling all application-dependent files into a single binary provides an applic
...
+
+Managed C++ applications aren't well suited for single-file deployment and we recommend that you write applications in C# or another non-managed C++ language to be single-file compatible. There are some rare cases where a managed library takes a dependency into the runtime library, coreclr, via `pinvoke`. This will not be possible in single-file case since the runtime library is bundled with the application.
+
+The following table outlines the API behavior in a single-file application:
+
+| API                                                                                     | Notes |
+| ---------------------------------------------------------------------------------------  | ------- |
+| Assembly.Location | Returns an empty string, use AppContext.BaseDirectory |
+| Module.FullyQualifiedName  | Returns a `<Unknown>` string ((just like it does in no-single-file for memory-loaded assemblies) |
```
jkotas, line 27: Also mention `Module.Name` ?

docs/core/deploying/single-file.md
```diff
@@ -11,9 +11,26 @@ Bundling all application-dependent files into a single binary provides an applic
...
+
+Single-file doesn't bundle native libraries by default. On Linux, we prelink the runtime into the bundle and only application native libraries are deployed to the same directory as the single-file app. On Windows, we prelink only the hosting code and both the runtime and application native libraries are deployed to the same directory as the single-file app. This is to ensure a good debugging experience, which requires native files to be excluded from the single file. There is an option to set a flag, `IncludeNativeLibrariesForSelfExtract`, to include native libraries in the single file bundle, but these files will be extracted to a temporary directory in the client machine when the single file application is run. Similarly, single-file doesn't bundle debug files (PDB files) for a better debugging experience.
```
jkotas, line 18: > Similarly, single-file doesn't bundle debug files (PDB files) for a better debugging experience.

This makes it sound like that not bunding PDB files improves debugging experience that's not correct.

Do we want to mentioned embedded PDBs?

docs/core/deploying/single-file.md
LakshanF: yes. Andy added a source analyzer, https://github.com/dotnet/roslyn-analyzers/pull/3921

docs/core/deploying/single-file.md
LakshanF: You are right!

docs/core/deploying/single-file.md
jkotas: Add a link to the analyzer rule from here? https://docs.microsoft.com/en-us/visualstudio/code-quality/il3000

docs/core/deploying/single-file.md
```diff
@@ -11,9 +11,26 @@ Bundling all application-dependent files into a single binary provides an applic
...
 
-There are some caveats that you need to be aware for single-file use, first of which is the use of path information to locate a file relative to the location of your application. The <xref:System.Reflection.Assembly.Location?displayProperty=nameWithType> API will return an empty string, which is the default behavior for assemblies loaded from memory. The compiler will give a warning for this API during build time to alert the developer to the specific behavior. If the path to the application directory is needed, the <xref:System.AppContext.BaseDirectory?displayProperty=nameWithType> API will return the directory where the AppHost (the single-file bundle itself) resides. Managed C++ applications aren't well suited for single-file deployment and we recommend that you write applications in C# to be single-file compatible.
+## Impact of using SingleFile feature
```
agocke, line 14: Let's just call this out directly: `API Incompatibility`

docs/core/deploying/single-file.md
```diff
@@ -11,9 +11,26 @@ Bundling all application-dependent files into a single binary provides an applic
...
+
+Managed C++ applications aren't well suited for single-file deployment and we recommend that you write applications in C# or another non-managed C++ language to be single-file compatible. There are some rare cases where a managed library takes a dependency into the runtime library, coreclr, via `pinvoke`. This will not be possible in single-file case since the runtime library is bundled with the application.
+
+The following table outlines the API behavior in a single-file application:
+
+| API                                                                                     | Notes |
```
agocke, line 24: Let's move this up and remove the paragraph explanation for `Assembly.Location`. This info should probably be at the top since everyone will have to review their app for these calls.

docs/core/deploying/single-file.md
```diff
@@ -11,9 +11,26 @@ Bundling all application-dependent files into a single binary provides an applic
...
+| ---------------------------------------------------------------------------------------  | ------- |
+| Assembly.Location | Returns an empty string, use AppContext.BaseDirectory |
+| Module.FullyQualifiedName  | Returns a `<Unknown>` string ((just like it does in no-single-file for memory-loaded assemblies) |
+| Assembly.GetFile  | Will throw `IOException` (just like it does in no-single-file for memory-loaded assemblies) |
+| Assembly.GetFiles  | Will throw `IOException` (just like it does in no-single-file for memory-loaded assemblies) |
+| Assembly.CodeBase  | Will throw `PlatformNotSupportedException` (this API is obsolete in .NET Core 5.0) |
+| Assembly.EscapedCodeBase  | Will throw `PlatformNotSupportedException` (this API is obsolete in .NET Core 5.0) |
+| AssemblyName.CodeBase  | Will throw `PlatformNotSupportedException` |
+| AssemblyName.EscapedCodeBase  | Will throw `PlatformNotSupportedException` |
```
agocke, line 33: This is also wrong, although I thought these returned `null`, not empty string.

docs/core/deploying/single-file.md
agocke, line 27: This may also throw. For each of these I would consider not specifying the exception type, since I'm not sure we want to be too explicit on the contract.

docs/core/deploying/single-file.md
```diff
@@ -11,9 +11,26 @@ Bundling all application-dependent files into a single binary provides an applic
...
+Managed C++ applications aren't well suited for single-file deployment and we recommend that you write applications in C# or another non-managed C++ language to be single-file compatible. There are some rare cases where a managed library takes a dependency into the runtime library, coreclr, via `pinvoke`. This will not be possible in single-file case since the runtime library is bundled with the application.
+
+The following table outlines the API behavior in a single-file application:
+
+| API                                                                                     | Notes |
+| ---------------------------------------------------------------------------------------  | ------- |
+| Assembly.Location | Returns an empty string, use AppContext.BaseDirectory |
+| Module.FullyQualifiedName  | Returns a `<Unknown>` string ((just like it does in no-single-file for memory-loaded assemblies) |
+| Assembly.GetFile  | Will throw `IOException` (just like it does in no-single-file for memory-loaded assemblies) |
+| Assembly.GetFiles  | Will throw `IOException` (just like it does in no-single-file for memory-loaded assemblies) |
```
agocke, line 29:
```suggestion
| Module.FullyQualifiedName  | Returns a `<Unknown>` string |
| Assembly.GetFile  | Will throw `IOException` |
| Assembly.GetFiles  | Will throw `IOException` |
```

docs/core/deploying/single-file.md
```diff
@@ -11,9 +11,26 @@ Bundling all application-dependent files into a single binary provides an applic
...
+
+Managed C++ applications aren't well suited for single-file deployment and we recommend that you write applications in C# or another non-managed C++ language to be single-file compatible. There are some rare cases where a managed library takes a dependency into the runtime library, coreclr, via `pinvoke`. This will not be possible in single-file case since the runtime library is bundled with the application.
+
+The following table outlines the API behavior in a single-file application:
+
+| API                                                                                     | Notes |
+| ---------------------------------------------------------------------------------------  | ------- |
+| Assembly.Location | Returns an empty string, use AppContext.BaseDirectory |
```
agocke, line 26: Let's separate known behavior suggestions from a separate checklist of "recommended alternatives," including `AppContext.BaseDirectory` to access the exe parent directory, `Environment.GetCommandLineArgs()[0]` for the exe file name, etc.

docs/core/deploying/single-file.md
agocke: Yes, I would instead say something like, "to include PDBs inside the application, use the \<DebugType>embedded</> flag" +link

docs/core/deploying/single-file.md
LakshanF: I was considering this but was not sure if we have alternative recommendations for all the APIs and if not, what we should say.

docs/core/deploying/single-file.md
LakshanF: These 2 seem to be returning null

docs/core/deploying/single-file.md
LakshanF: Need to think more on this since it seems better for some of these APIs not to give prescriptive guidance but rather encourage to take a step back and re-think the app design for single-file use.

docs/core/deploying/single-file.md
```diff
@@ -11,10 +11,32 @@ Bundling all application-dependent files into a single binary provides an applic
...
+## API Incompatibility
+
+There are some caveats that you need to be aware for single-file use, first of which is the use of path information to locate a file relative to the location of your application. The <xref:System.Reflection.Assembly.Location?displayProperty=nameWithType> API will return an empty string, which is the default behavior for assemblies loaded from memory. The compiler will give a [warning](https://docs.microsoft.com/en-us/visualstudio/code-quality/il3000?view=vs-2019) for this API during build time to alert the developer to the specific behavior. If the path to the application directory is needed, the <xref:System.AppContext.BaseDirectory?displayProperty=nameWithType> API will return the directory where the AppHost (the single-file bundle itself) resides. There is also an option, `ExcludeFromSingleFile`, to exclude files from being bundled that is described [below](#exclude-files-from-being-embedded) in detail.
```
jkotas, line 16: Omit `en-us` and `?view=vs-2019` in the URL.

https://docs.microsoft.com/visualstudio/code-quality/il3000

The `en-us` is important to omit to make the links work well in localized version of the documentation

docs/core/deploying/single-file.md
LakshanF: We had a publish property for this in pre 5.0, IncludeSymbolsInSingleFile, but stopped supporting it in 5.0 with [this](https://github.com/dotnet/sdk/pull/12724/files#diff-7dc833d02f198e5a9f0ab88e229db18aR97) checkin.

@sbomer might have more context on the reason

docs/core/deploying/single-file.md
```diff
@@ -11,10 +11,32 @@ Bundling all application-dependent files into a single binary provides an applic
...
+
+The table below has the relevant runtime library API details for single-file use.
+
+| API                                                                                     | Note |
+| ---------------------------------------------------------------------------------------  | ------- |
+| Assembly.Location | Returns an empty string, use AppContext.BaseDirectory |
+| Module.FullyQualifiedName  | Returns a `<Unknown>` string |
+| `Module.Name`  | Returns a `<Unknown>` string |
```
vitek-karas, line 24: We should make all the other API names a "code" format as well

docs/core/deploying/single-file.md
```diff
@@ -11,10 +11,32 @@ Bundling all application-dependent files into a single binary provides an applic
...
+| Assembly.EscapedCodeBase  | Will throw `PlatformNotSupportedException` |
+| AssemblyName.CodeBase  | Returns null |
+| AssemblyName.EscapedCodeBase  | Returns null |
+
+## Other considerations
 
 Single-file doesn't bundle native libraries by default. On Linux, we prelink the runtime into the bundle and only application native libraries are deployed to the same directory as the single-file app. On Windows, we prelink only the hosting code and both the runtime and application native libraries are deployed to the same directory as the single-file app. This is to ensure a good debugging experience, which requires native files to be excluded from the single file. There is an option to set a flag, `IncludeNativeLibrariesForSelfExtract`, to include native libraries in the single file bundle, but these files will be extracted to a temporary directory in the client machine when the single file application is run.
 
+Single-file application will have all related PDB files alongside it and will not be bundled by default. If you want to include PDBs inside the application, set the `DebugType` to `embedded` as described [below](#include-pdb-files-inside-the-bundle) in detail.
```
vitek-karas, line 36: This is kind of "lying" - SDK will put all PDBs there - even those from potential NuGets and such - the embedded solution will only work for the assemblies which are built from source in the project...

docs/core/deploying/single-file.md
```diff
@@ -34,6 +56,22 @@ For example, to place some files in the publish directory but not bundle them in
 </ItemGroup>
 ```
 
+## Include PDB files inside the bundle
+
+The PDB file for an assembly can be bundled with single-file by setting the following property:
```
vitek-karas, line 61:
```suggestion
The PDB file for an assembly can be embedded into the assembly itself (the `.dll`) using the setting below. Since the symbols are part of the assembly, they will be part of the single-file application as well.
```

Probably needs better wording - but basically: embedding symbols does not "bundle them with single-file" - or not directly. I think in the context of this document it would be rather confusing.

Also - do we have some docs on PDB embedding - I would assume so. Maybe link to it.

docs/core/deploying/single-file.md
```diff
@@ -11,10 +11,32 @@ Bundling all application-dependent files into a single binary provides an applic
...
 
-There are some caveats that you need to be aware for single-file use, first of which is the use of path information to locate a file relative to the location of your application. The <xref:System.Reflection.Assembly.Location?displayProperty=nameWithType> API will return an empty string, which is the default behavior for assemblies loaded from memory. The compiler will give a warning for this API during build time to alert the developer to the specific behavior. If the path to the application directory is needed, the <xref:System.AppContext.BaseDirectory?displayProperty=nameWithType> API will return the directory where the AppHost (the single-file bundle itself) resides. Managed C++ applications aren't well suited for single-file deployment and we recommend that you write applications in C# to be single-file compatible.
+## API Incompatibility
```
adegeo, line 14:
```suggestion
## API incompatibility
```

docs/core/deploying/single-file.md
adegeo: @LakshanF These names should be inline code. Some are, some aren't. Otherwise, they'll be translated into other languages.

docs/core/deploying/single-file.md
LakshanF: Looks like the DebugType route was the direction we wanted to go based on the github discussion on this

docs/core/deploying/single-file.md
LakshanF: changed the wording to reflect this

docs/core/deploying/single-file.md
LakshanF: Couldn't quite find a relevant doc related to this but changed wording to reflect the feedback

adegeo:
@LakshanF your email used in your git is not the same as the one in your github profile, so it's not linking the commits to your account. Not a big deal I don't think, but you won't see your face on the "contributors" to the article 😄

LakshanF:
Hi Andy,
I believe this email is registered with my work profile and I have the necessary repo permissions, at least to the runtime. I can also see my ‘face’ on the existing doc that I’m trying to update currently 😊
Thanks
Lakshan

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

docs/core/deploying/single-file.md
```diff
@@ -11,10 +11,32 @@ Bundling all application-dependent files into a single binary provides an applic
...
+## API incompatibility
+
+There are some caveats that you need to be aware for single-file use, first of which is the use of path information to locate a file relative to the location of your application. The <xref:System.Reflection.Assembly.Location?displayProperty=nameWithType> API will return an empty string, which is the default behavior for assemblies loaded from memory. The compiler will give a [warning](https://docs.microsoft.com/visualstudio/code-quality/il3000) for this API during build time to alert the developer to the specific behavior. If the path to the application directory is needed, the <xref:System.AppContext.BaseDirectory?displayProperty=nameWithType> API will return the directory where the application executable resides. There is also an option, `ExcludeFromSingleFile`, to exclude files from being bundled that is described [below](#exclude-files-from-being-embedded) in detail.
```
agocke, line 16:
```suggestion
Some APIs are not compatible with single-file deployment and applications may require modification if they use these APIs. If you use a third-party framework or package, it's possible that they may also use one of these APIs and need modification. The most common cause of problems is dependence on file paths for files or DLLs shipped with the application.
```

docs/core/deploying/single-file.md
```diff
@@ -11,10 +11,32 @@ Bundling all application-dependent files into a single binary provides an applic
...
+| `Assembly.Location` | Returns an empty string, use `AppContext.BaseDirectory` |
+| `Module.FullyQualifiedName`  | Returns a `<Unknown>` string |
+| `Module.Name`  | Returns a `<Unknown>` string |
+| `Assembly.GetFile`  | Will throw `IOException` |
+| `Assembly.GetFiles`  | Will throw `IOException` |
+| `Assembly.CodeBase`  | Will throw `PlatformNotSupportedException` |
+| `Assembly.EscapedCodeBase`  | Will throw `PlatformNotSupportedException` |
+| `AssemblyName.CodeBase`  | Returns null |
+| `AssemblyName.EscapedCodeBase`  | Returns null |
+
```
agocke, line 31:
```suggestion

We have some recommendations for fixing common scenarios:

* To access files next to the executable, use <xref:System.AppContext.BaseDirectory?displayProperty=nameWithType>

* To find the file name of the executable, use the first element of <xref:System.Environment.GetCommandLineArgs()?displayProperty=nameWithType>

* To avoid shipping loose files entirely, consider using [embedded resources](https://docs.microsoft.com/en-us/dotnet/framework/resources/creating-resource-files-for-desktop-apps)
```

docs/core/deploying/single-file.md
```diff
@@ -11,10 +11,32 @@ Bundling all application-dependent files into a single binary provides an applic
...
+The table below has the relevant runtime library API details for single-file use.
+
+| API                                                                                     | Note |
+| ---------------------------------------------------------------------------------------  | ------- |
+| `Assembly.Location` | Returns an empty string, use `AppContext.BaseDirectory` |
+| `Module.FullyQualifiedName`  | Returns a `<Unknown>` string |
```
agocke, line 23:
```suggestion
| `Module.FullyQualifiedName`  | Returns "<Unknown>" or throws an exception |
```

docs/core/deploying/single-file.md
```diff
@@ -11,10 +11,32 @@ Bundling all application-dependent files into a single binary provides an applic
...
+
+The table below has the relevant runtime library API details for single-file use.
+
+| API                                                                                     | Note |
+| ---------------------------------------------------------------------------------------  | ------- |
+| `Assembly.Location` | Returns an empty string, use `AppContext.BaseDirectory` |
```
agocke, line 22:
```suggestion
| `Assembly.Location` | Returns an empty string |
```

docs/core/deploying/single-file.md
```diff
@@ -11,10 +11,40 @@ Bundling all application-dependent files into a single binary provides an applic
...
+## API incompatibility
+
+Some APIs are not compatible with single-file deployment and applications may require modification if they use these APIs. If you use a third-party framework or package, it's possible that they may also use one of these APIs and need modification. The most common cause of problems is dependence on file paths for files or DLLs shipped with the application.
+
+The table below has the relevant runtime library API details for single-file use.
+
+| API                            | Note                                                                   |
+|--------------------------------|------------------------------------------------------------------------|
+| `Assembly.Location`            | Returns an empty string.                                               |
+| `Module.FullyQualifiedName`    | Returns a string with the value of `<Unknown>` or throws an exception. |
```
adegeo, line 23: I did a bit of cleanup in this table. Do you have an idea as to which exception is thrown here? Saying "throws an exception" is pretty generic. 😄

agocke: Yeah, I prefer we don't specify an exception type here so people can't take a dependency on it. The exact behavior of this API may change through runtime refactorings or modifications.

adegeo (approved):

PR #20769, opened 2020-09-24
Deconstruct is for positional records only

BillWagner:
The `Deconstruct` method is only synthesized for positional records.

### DIFF
--- a/docs/csharp/whats-new/csharp-9.md
+++ b/docs/csharp/whats-new/csharp-9.md
@@ -42,7 +42,6 @@ The record definition creates a `Person` type that contains two readonly propert
 - Override for <xref:System.Object.GetHashCode>
 - Copy and Clone members
 - `PrintMembers` and <xref:System.Object.ToString>
-- `Deconstruct` method
 
 Records support inheritance. You can declare a new record derived from `Person` as follows:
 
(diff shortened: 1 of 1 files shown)


docs/csharp/whats-new/csharp-9.md
```diff
@@ -42,7 +42,6 @@ The record definition creates a `Person` type that contains two readonly propert
 - Override for <xref:System.Object.GetHashCode>
 - Copy and Clone members
 - `PrintMembers` and <xref:System.Object.ToString>
-- `Deconstruct` method
```
IEvangelist, line 45: We should probably add somewhere that an auto-generated `Deconstruct` method exists when declaring a `record` type with the following syntax:

```csharp
public record Person(string FirstName, string LastName);
```

But not when using the verbose syntax:

```csharp
public record Person
{
    ...
}
```

IEvangelist (approved):
I'm good with this, just think it might be worth calling out when you do get auto-deconstruct

docs/csharp/whats-new/csharp-9.md
BillWagner: I do call that out a couple paragraphs down, where I introduce positional records.

PR #20780, opened 2020-09-25
Renamed rules folder to quality-rules

gewarren:
We need to rename the folder because we'll soon be adding style rules as well, and they should go in a separate folder.

Since this has only been live for a few hours, I don't think any redirects are necessary.


adegeo (approved):
Perhaps we should publish right away?

mavasani (approved):

PR #20797, opened 2020-09-27
Fix syntactic error in recursion article

piotr-lewandowski:
The example `fib` function in recursive-functions-the-rec-keyword.md, is not valid F# code as it lacks the `rec` keyword.

## Summary
The first fibonacci example does not compile
``` fsharp
let fib n =
    match n with
    | 0 | 1 -> 1
    | n -> fib (n-1) + fib (n-2)
```
I tested it on my machine and a simple addition of the `rec` keyword is enough to make the example compile.


cartermp (approved):
Thanks!

PR #20835, opened 2020-09-29
Standardize sample projects, Part 53

nxtn:


BillWagner (approved):
Thanks again. This is also great.

I'll :shipit:

PR #20836, opened 2020-09-29
Use proper names

Youssef1313:
Replaces #19756


.markdownlint.json
```diff
@@ -39,5 +39,12 @@
     "MD041": false,
     "MD046": {
         "style": "fenced"
+    },
+    "proper-names": {
+        "code_blocks": false,
+        "names": [
+            "ASP.NET",
+            "PowerShell"
```
Youssef1313, line 47: I didn't add the other names as they have false positives.
If you want to completely remove this from markdownlint and take the fixes only, let me know.

BillWagner (approved):
Thanks @Youssef1313 
This is a big improvement. I'll :shipit: now.

PR #20818, opened 2020-09-28
Fix CA1835 code fix screenshot

carlossanlop:
Fixes https://github.com/dotnet/docs/issues/20820
When I created the screenshot, I didn't notice I was focusing the "expression value is never used" suggested change, instead of the actual CA1835 rule.
I made sure not to select `0` for the `offset` value because we are going to add an improvement to the rule to handle a special case: if `offset` is `0` and `count` is `buffer.Length`, then the suggestion will be to simply pass `buffer` (no `AsMemory` invocation). See https://github.com/dotnet/roslyn-analyzers/issues/4235
I also made sure to generate a screenshot in blue theme, because the original screenshot was in black theme and the rules say we should use the blue theme.

PR #20838, opened 2020-09-29
Fix CA1835 code fix screenshot (second version)

carlossanlop:
The original submission https://github.com/dotnet/docs/pull/20818 had a bug where offset+count would be larger than buffer.Length.
cc @stephentoub @gewarren


stephentoub:
Would you mind changing the `new CancellationToken()` to instead be `CancellationToken.None` or `default(CancellationToken)`?  It's not a big deal, though.

stephentoub (approved):

carlossanlop:
> _Would you mind changing the `new CancellationToken()` to instead be `CancellationToken.None` or `default(CancellationToken)`? It's not a big deal, though._

GitHub does not seem to be updating the comments when I leave the page open, so I saw your comment after I merged. I can definitely do that, no problem, @stephentoub.

PR #20848, opened 2020-09-30
Add breaking change for preprocessor symbols

### LINKED ISSUE #20692: Don't define NETCOREAPP3_1 when targeting .NET 5
## Don't define NETCOREAPP3_1 when targeting .NET 5

.NET 5 will now have the same behavior for defining preprocessor symbols as previous releases of .NET Core, that is, we only define symbols for what the project is actually targeting and no longer include defines for earlier versions.

### Version introduced

.NET 5 RC2

### Old behavior

When targeting `net5.0` we defined both `NETCOREAPP3_1` and `NET5_0` preprocessor symbols. The intent was that starting with .NET 5, conditional compilation symbols [would be accumulative](https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md#preprocessor-symbols).

### New behavior

We'll revert this change which will make the behavior consistent with how PR preprocessor symbols were defined in previous releases of .NET Core (1.0 - 3.1).

### Reason for change

Customer feedback. The new behavior was surprising and caused confusion on how to accommodate it. People assumed this was a bug in the C# compiler.

* [Issue](https://github.com/dotnet/sdk/issues/13377)
* [PR](https://github.com/dotnet/sdk/pull/13615)

### Recommended action

Ensure your `#if` logic doesn't assume that `NETCOREAPP3_1` is defined when the project is targeting `net5.0` or higher. Instead, assume that `NETCOREAPP3_1` is only defined when the project targets `netcoreapp3.1` exactly.

For example, if your project multitargets for .NET Core 2.1 and .NET Core 3.1 and you need to call APIs that were introduced in .NET Core 3.1, your `#if` logic should look as follows:

```C#
#if NETCOREAPP2_1 || NETCOREAPP3_0
    // Fallback behavior for old versions
#elif NETCOREAPP
    // Behavior for .NET Core 3.1 or later
#endif
```

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #20692.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/msbuild?branch=pr-en-us-20848#netcoreapp3_1-preprocessor-symbol-is-not-defined-when-targeting-net-5).

### DIFF
--- a/docs/core/compatibility/msbuild.md
+++ b/docs/core/compatibility/msbuild.md
@@ -9,7 +9,7 @@ The following breaking changes are documented on this page:
 
 | Breaking change | Version introduced |
 | - | - |
-| [NETCOREAPP3_1 preprocessor symbol is not defined when targeting .NET 5](#netcoreapp31-preprocessor-symbol-is-not-defined-when-targeting-net-5) | 5.0 |
+| [NETCOREAPP3_1 preprocessor symbol is not defined when targeting .NET 5](#netcoreapp3_1-preprocessor-symbol-is-not-defined-when-targeting-net-5) | 5.0 |
 | [PublishDepsFilePath behavior change](#publishdepsfilepath-behavior-change) | 5.0 |
 | [Directory.Packages.props files is imported by default](#directorypackagesprops-files-is-imported-by-default) | 5.0 |
 | [Resource manifest file name change](#resource-manifest-file-name-change) | 3.0 |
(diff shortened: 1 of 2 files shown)


BillWagner (approved):
This LGTM @gewarren 

:shipit:

gewarren:
Thanks Bill!

PR #20870, opened 2020-10-01
Clarify distinction between "assembly reference" and "version referenced by an assembly"

sambauwens:
## Summary

Reading the phrase "A strong-named assembly reference must exactly match the version referenced by an assembly", I am confused as to what "version referenced by an assembly" refers to, it seems to refer to the same thing as "assembly reference" in the first part of the phrase. So I was trying to find the expression that would clarify what is meant by the second part, which I think refer to the "the candidate assembly to be bound", which depends on the context (nuget, GAC...). I chose loaded assembly here, but I'm not sure it's accurate.


IEvangelist (approved):

Issue #11957, opened 2019-04-19
Replace "default constructor" with "parameterless constructor"


rpetrusha:
## Replace "default constructor" with "parameterless constructor"

For background, see #10517 and #11945 

The term "default constructor" is confusing because:
- It is similar to "implicit default constructor".
- It can be confused with a constructor that has default arguments.

It should be replaced with "parameterless constructor". This change should be made in:

- [ ] The Visual Basic documentation.
- [ ] The Managed Reference documentation (in dotnet/dotnet-api-docs).
- [ ] The .NET Guide.

keysangyonthan:
> ## Replace "default constructor" with "parameterless constructor"
> For background, see #10517 and #11945
> 
> The term "default constructor" is confusing because:
> 
> * It is similar to "implicit default constructor".
> * It can be confused with a constructor that has default arguments.
> 
> It should be replaced with "parameterless constructor". This change should be made in:
> 
> * [ ]  The Visual Basic documentation.
> * [ ]  The Managed Reference documentation (in dotnet/dotnet-api-docs).
> * [ ]  The .NET Guide.

Where are the documents, I can't find "default constructor" anywhere in the docs I have seen, can you please provide more information and directory. Thanks

alx168:
I'll do it.

PR #20873, opened 2020-10-01
Replace default constructor with parameterless

QwerMike:
## Summary

Updated the article "How to initialize objects by using an object
initializer". Replaced "default" constructor with "parameterless"
in the article itself and corresponding code snippets file.

Contributes to the issue #11957


BillWagner (approved):
Thanks for spotting this and fixing it @QwerMike 

PR #20888, opened 2020-10-02
Fix PerformanceCounters element

gewarren:
Also remove unnecessary text from TOC titles.


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

PR #20899, opened 2020-10-04
Change https to http for jsnlog.com, https does not resolve to the correct website.

Swimburger:
Unfortunately, the https version of jsnlog.com doesn't resolve to the correct website while the http version does.
If it is not allowed to have http-links (non-https-links) on docs, we should probably refer directly to the GitHub repository.

HTTPS:
image

HTTP: 
image


BillWagner (approved):
Thanks for making this change @Swimburger 

We do prefer https links when they are available, but we accept http links when necessary.

PR #20900, opened 2020-10-04
URL for React project has changed

Swimburger:
URL for React project has changed to https://reactjs.org/ from https://facebook.github.io/react/.
For now it redirects fine, so feel free to reject change. Though I want to change it in case the redirect is gone in the future.


adegeo (approved):
Thanks!

PR #20905, opened 2020-10-04
Fix link to a section

artemious7:
## Summary

Fixed a link which was displaying as plain text


BillWagner (approved):
Thanks for spotting this and fixing it @artemious7 

We appreciate it.

PR #20918, opened 2020-10-05
Update interpolated-strings.md

ericmutta:
Remove reference to C# conditional expression `(age == 1 ? "" : "s"` which can't be used in VB as written (the VB version has no colon so the problem of needing to escape it via parentheses goes away).


BillWagner (approved):
Thanks for making this update @ericmutta 

Thanks again.

ericmutta:
Many thanks for following up @BillWagner, much appreciated :+1:

PR #20919, opened 2020-10-05
Update interpolated-strings4.vb

ericmutta:
Remove the extra brackets around the `If()` expression since they were assumed to escape a colon character that doesn't exist in VB (but does in C#'s conditional operator). This code change is related to the textual changes made in PR #20918.

I have also removed a stray `</Snippet1>` tag that doesn't appear to have a matching opening tag anywhere.

### DIFF
--- a/samples/snippets/visualbasic/programming-guide/language-features/strings/interpolated-strings4.vb
+++ b/samples/snippets/visualbasic/programming-guide/language-features/strings/interpolated-strings4.vb
@@ -5,11 +5,10 @@
       Dim s1 = $"He asked, ""Is your name {name}?"", but didn't wait for a reply."
       Console.WriteLine(s1)
       
-      Dim s2 = $"{name} is {age:D3} year{(If(age = 1, "", "s"))} old."
+      Dim s2 = $"{name} is {age:D3} year{If(age = 1, "", "s")} old."
       Console.WriteLine(s2) 
    End Sub
 End Module
 ' The example displays the following output:
 '       He asked, "Is your name Horace?", but didn't wait for a reply.
 '       Horace is 034 years old.
-' </Snippet1>


ericmutta:
Hmmm...looks like the build failed (something about a missing project/solution file). Did removing the `</Snippet1>` tag somehow cause this?

BillWagner:
@ericmutta We've added a new sample CI system since these files were last updated. These samples don't include a `vbproj` file to build the samples.  

For these older samples, we've had to add a project file, and make sure that project file specifies which `Main` to use as the entry point. Often, we've added a new module with a `Main` method that calls all the existing methods in turn to run all the snippets in a folder.

Thanks again for all your help.

ericmutta:
Hi @BillWagner many thanks for the clarification! Is there anything I can do from my side to get the samples to build or should I leave that to you? (I imagine the procedure may require more privileges than I have as an external contributor).

BillWagner:
@ericmutta You should be able to add a project file to this PR, and that should be all that's needed to get the CI build to pass. As I mentioned in the previous comment, because the folder has multiple `Main` methods, the project file needs to specify one as the entry point.

BillWagner:
@ericmutta Will you add the project file to resolve the build issues? Or would you rather we did that?

ericmutta:
Hi @BillWagner, many thanks for following up and apologies I got carried away with work stuff! Please go ahead and add the project file on my behalf so I can observe how its done and use that knowledge for future PRs :+1:

BillWagner (approved):
I'll merge this despite the CI build issue.  Because Visual Basic doesn't have a way to specify multiple `Main` methods, this would require breaking up these snippets into different folders, with different projects. That's more work than justified for this change.

ericmutta:
Thanks for following up @BillWagner, much appreciated! :+1:

PR #20922, opened 2020-10-05
Remove statement about TLS 1.3 for .NET Core 3.0

kapsiR:
I think, TLS 1.3 won't be supported by .NET Core 3.0 since it's out of support and the implementation is coming with .NET 5, as stated in [Transport Layer Security (TLS) best practices with the .NET Framework](https://github.com/dotnet/docs/issues/4675#issuecomment-678421120).

This info should be removed, so we have no misleading information out there... 😉


adegeo (approved):
Thanks for finding that info. I think this is safe to remove. Cheers!

PR #20930, opened 2020-10-06
Update delimited regex

willwhitwellbailey:
The description given for the breakout of the regex was missing the pipe character.


willwhitwellbailey:
Turns out the expression is right. Documentation is not. Updating that instead.

PR #20931, opened 2020-10-06
Removing redundant pipe character

willwhitwellbailey:
## Summary

Pipe character is not used in code example nor in Pattern breakout following.


adegeo (approved):
Good catch. Thanks!

PR #20956, opened 2020-10-06
Breaking changes for JsonSerializer.Deserialize

### LINKED ISSUE #20815: Breaking change: deserializing JSON string into a char requires single-character string
## Deserializing JSON string into a char is now more strict to match user expectations

`JsonSerializer.Deserialize<char>(payload)` the `payload` must contain a single-character string for deserialization to succeed.
Related to https://github.com/dotnet/runtime/pull/528#issuecomment-589298591
### Version introduced

5.0

### Old behavior
When deserializing a char with JsonSerializer if payload consists multicharacter string first character was deserialized.

```csharp
public static void TestDeserializeToChar()
{
    // Before (3.1): Returns the first character - 'a'
    JsonSerializer.Deserialize<char>("\"abc\"");
}
```
### New behavior
When deserializing a char with JsonSerializer payload must contain a single-character string for deserialization to succeed, throws JsonException otherwise.

```csharp
public static void TestDeserializeToChar()
{
    // After (5.0): throw JsonException because payload has more than one character
    JsonSerializer.Deserialize<char>("\"abc\"");
}
```
### Reason for change

`JsonSerializer.Deserialize<TValue>("payload")`  parses the text representing a single JSON value into an instance of the type specified by a generic type parameter. Parsing should only succeed when the provided `payload` is valid for the specified generic type parameter. For char value type a valid payload is expected to be a single character string.

### Recommended action

When parsing a string to a char type using `JsonSerializer.Deserialize<char>(payload)`  make sure the `payload` is non-null, non-empty, and contain a single-character string for a successful deserialization.

### Category

- [ ] ASP.NET Core
- [ ] C#
- [ ] Code analysis
- [x] Core .NET libraries
- [ ] Cryptography
- [ ] Data
- [ ] Debugger
- [ ] Deployment
- [ ] Globalization
- [ ] Interop
- [ ] JIT
- [ ] LINQ
- [ ] MSBuild
- [ ] Networking
- [ ] Printing
- [ ] Security
- [x] Serialization
- [ ] Visual Basic
- [ ] Windows Forms
- [ ] XML, XSLT

### Affected APIs

[System.Text.Json.JsonSerializer.Deserialize(String, JsonSerializerOptions)](https://docs.microsoft.com/dotnet/api/system.text.json.jsonserializer.deserialize#System_Text_Json_JsonSerializer_Deserialize__1_System_String_System_Text_Json_JsonSerializerOptions_)

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #20815.

[Preview link - 5.0](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/serialization?branch=pr-en-us-20956#jsonserializerdeserialize-requires-single-character-string).


includes/core-changes/serialization/3.1/jsonserializer-deserialize-throws-jsonexception.md
```diff
@@ -0,0 +1,35 @@
+### JsonSerializer.Deserialize throws JsonException for null or empty strings
+
+When the type parameter is <xref:System.Char>, <xref:System.Text.Json.JsonSerializer.Deserialize%60%601(System.String,System.Text.Json.JsonSerializerOptions)?displayProperty=nameWithType> now throws a <xref:System.Text.Json.JsonException> instead of a <xref:System.NullReferenceException> or <xref:System.IndexOutOfRangeException>.
+
+#### Change description
+
+In previous .NET versions, when the type parameter is <xref:System.Char>, passing a null reference for the <xref:System.String> parameter of <xref:System.Text.Json.JsonSerializer.Deserialize%60%601(System.String,System.Text.Json.JsonSerializerOptions)?displayProperty=nameWithType> causes a <xref:System.NullReferenceException> to be thrown. Passing an empty string causes an <xref:System.IndexOutOfRangeException> to be thrown.
+
+In .NET Core 3.1 and later, when the type parameter is <xref:System.Char>, passing a null reference or an empty string causes a <xref:System.Text.Json.JsonException> to be thrown.
```
buyaa-n, line 9: NIT: I think adding the example code might make it more clear, up to you
```csharp
    // Before (3.0): throws NullReferenceException
    // After (3.1): throw JsonException
    JsonSerializer.Deserialize<char>("null");

    // Before (3.0): throws IndexOutOfRangeException
    // After (3.1): throw JsonException
    JsonSerializer.Deserialize<char>("");

File: includes/core-changes/serialization/5.0/deserializing-json-into-char-requires-single-character.md [hunk 2]
```diff
@@ -0,0 +1,37 @@
+### JsonSerializer.Deserialize requires single-character string
+
+When the type parameter is <xref:System.Char>, the string argument to <xref:System.Text.Json.JsonSerializer.Deserialize%60%601(System.String,System.Text.Json.JsonSerializerOptions)?displayProperty=nameWithType> must contain a single character for deserialization to succeed.
+
+#### Change description
+
+In previous .NET versions, if you pass a multi-character string to <xref:System.Text.Json.JsonSerializer.Deserialize%60%601(System.String,System.Text.Json.JsonSerializerOptions)?displayProperty=nameWithType> and the type parameter is <xref:System.Char>, the deserialization succeeds and only the first character is deserialized.
+
+In .NET 5.0 and later, when the type parameter is <xref:System.Char>, passing anything other than a single-character string causes a <xref:System.Text.Json.JsonException> to be thrown.
```
buyaa-n, line 9: NIT: same
```csharp
    // Before (5.0): Returns the first character - 'a'
    // After (5.0): throw JsonException because payload has more than one character
    JsonSerializer.Deserialize<char>("abc");

    // Correct usage
    JsonSerializer.Deserialize<char>("a");
```

buyaa-n (approved):
Left a NIT comment, overall looks good, thanks!

includes/core-changes/serialization/3.1/jsonserializer-deserialize-throws-jsonexception.md
gewarren: Okay, I will add it. But isn't `"null"` just a string that contains the characters n, u, l, and l?

buyaa-n: Good question, it might supposed to be `JsonSerializer.Deserialize<char>(null);`, but also "null" string might be serialized into `null` value instead of "null" string cc @layomia @ahsonkhan

includes/core-changes/serialization/5.0/deserializing-json-into-char-requires-single-character.md
```diff
@@ -0,0 +1,46 @@
...
+
+```csharp
+// .NET Core 3.0 and 3.1: Returns the first character 'a'.
+// .NET 5.0 and later: Throws JsonException because payload has more than one character.
+JsonSerializer.Deserialize<char>("abc");
+
+// Correct usage.
+JsonSerializer.Deserialize<char>("a");
```
ahsonkhan, line 17: Both these are invalid JSON because they are missing the string within embedded quotes, These should be:
```suggestion
JsonSerializer.Deserialize<char>("\"abc\"");

// Correct usage.
JsonSerializer.Deserialize<char>("\"a\"");
```

includes/core-changes/serialization/5.0/deserializing-json-into-char-requires-single-character.md
ahsonkhan: You are missing quotes in the input. Just the string `a` without quotes surrounding it, is invalid JSON.

includes/core-changes/serialization/5.0/deserializing-json-into-char-requires-single-character.md
gewarren: But then the first character is always ", so how do you pass a single-character string?

ahsonkhan: That **is** how you pass a single character string.

Here's an example valid, single-character JSON string:
`"a"`

In JSON, a string token is one that is surrounded with quotes.

Another valid JSON token, is number, so for example:
`1`

The way you would write that in C# is as follows (note, there are no extra quotes):
```C#
JsonSerializer.Deserialize<int>("1");
```

Maybe if you think of it as a character array rather than strings, that might clear up the need for the quotes:
```C#
byte[] arr = {(byte)'"', (byte)'a', (byte)'"'};
Console.WriteLine(JsonSerializer.Deserialize<char>(arr.AsSpan())); // valid, character 'a'

byte[] another = {(byte)'1', (byte)'2', (byte)'3'};
Console.WriteLine(JsonSerializer.Deserialize<int>(another.AsSpan())); // valid, number 123
```

ahsonkhan: You can explore various values in jsonlint to see:
https://jsonlint.com/

image

includes/core-changes/serialization/3.1/jsonserializer-deserialize-throws-jsonexception.md
```diff
@@ -0,0 +1,45 @@
...
+
+```csharp
+// .NET Core 3.0: Throws NullReferenceException.
+// .NET Core 3.1 and later: Throws JsonException.
+JsonSerializer.Deserialize<char>(null);
+
+// .NET Core 3.0: Throws IndexOutOfRangeException.
+// .NET Core 3.1 and later: Throws JsonException.
+JsonSerializer.Deserialize<char>("");
```
ahsonkhan, line 18:
```suggestion
JsonSerializer.Deserialize<char>("\"\"");
```

includes/core-changes/serialization/3.1/jsonserializer-deserialize-throws-jsonexception.md
```diff
@@ -0,0 +1,45 @@
...
+#### Change description
+
+In previous .NET versions, when the type parameter is <xref:System.Char>, passing a null reference for the <xref:System.String> parameter of <xref:System.Text.Json.JsonSerializer.Deserialize%60%601(System.String,System.Text.Json.JsonSerializerOptions)?displayProperty=nameWithType> causes a <xref:System.NullReferenceException> to be thrown. Passing an empty string causes an <xref:System.IndexOutOfRangeException> to be thrown.
+
+In .NET Core 3.1 and later, when the type parameter is <xref:System.Char>, passing a null reference or an empty string causes a <xref:System.Text.Json.JsonException> to be thrown.
+
+```csharp
+// .NET Core 3.0: Throws NullReferenceException.
+// .NET Core 3.1 and later: Throws JsonException.
+JsonSerializer.Deserialize<char>(null);
```
ahsonkhan, line 14: 1) This code won't compile without a cast:
> Compilation error (line 10, col 3): The call is ambiguous between the following methods or properties: 'System.Text.Json.JsonSerializer.Deserialize<char>(string, System.Text.Json.JsonSerializerOptions)' and 'System.Text.Json.JsonSerializer.Deserialize<char>(System.ReadOnlySpan<byte>, System.Text.Json.JsonSerializerOptions)'

2) It should be the following:
```suggestion
JsonSerializer.Deserialize<char>("\"null\"");
```

Processing **that** particular JSON string is what was causing a NullReferenceException on 3.0 and is now correctly throwing a JsonException with this message:
> Unhandled exception. System.Text.Json.JsonException: The JSON value could not be converted to System.Char. Path: $ | LineNumber: 0 | BytePositionInLine: 2.

includes/core-changes/serialization/3.1/jsonserializer-deserialize-throws-jsonexception.md
ahsonkhan: I responded to this question below:
https://github.com/dotnet/docs/pull/20956#discussion_r500707091

This is invalid:
```C#
JsonSerializer.Deserialize<char>(null);
```

The example needs to be a null string in quotes, i.e. the six characters ", n, u, l, l, ".

ahsonkhan:
@buyaa-n - can you please also update the examples in the breaking change issue https://github.com/dotnet/docs/issues/20815#issue-710570036

And please also include the `JsonSerializer.Serialize` change mentioned in https://github.com/dotnet/runtime/pull/528#issuecomment-589298591 (in addition to the **De**serialize changes you already covered)

ahsonkhan:
Also, to clarify, there is only a single truly breaking change that needs to be documented:

```C#
// This use to work "incorrectly" before and now it throws an exception
JsonSerializer.Deserialize<char>("\"abc\"");
```

The other examples are changing one exception type to another and those are not considered breaking. In fact, going from NullRefException to a more user-friendly exception is encouraged.

```C#
using System;
using System.Text.Json;

namespace TestNull
{
    public class Program
    {
        public class MyPoco
        {
            public char Character { get; set; }
        }

        {
            // === Changing exception types like this is NOT considered a breaking change doc. === 
            // Before (3.0): NullReferenceException
            // After (3.1): JsonException
            // After (5.0): JsonException
            JsonSerializer.Deserialize<MyPoco>("{\"Character\": null}");

            // === Changing exception types like this is NOT considered a breaking change doc. === 
            // Before (3.0): NullReferenceException
            // After (3.1): JsonException
            // After (5.0): JsonException
            JsonSerializer.Deserialize<MyPoco>("{\"Character\": \"null\"}");

            // === Changing exception types like this is NOT considered a breaking change doc. === 
            // Before (3.0): IndexOutOfRangeException
            // After (3.1): JsonException
            // After (5.0): JsonException
            JsonSerializer.Deserialize<MyPoco>("{\"Character\": \"\"}");

            // === This is the ONLY change that needs to be documented as it is the only one that is breaking === 
            // Before (3.0): Set Character to the first character - 'a'
            // Before (3.1): Set Character to the first character - 'a'
            // After (5.0): JsonException
            var val = JsonSerializer.Deserialize<MyPoco>("{\"Character\": \"abc\"}");
            Console.WriteLine(val.Character);
        }
    }
}
```

includes/core-changes/serialization/5.0/deserializing-json-into-char-requires-single-character.md
buyaa-n: > Both these are invalid JSON because they are missing the string within embedded quotes, These should be:

Ah, forgot about that

includes/core-changes/serialization/3.1/jsonserializer-deserialize-throws-jsonexception.md
buyaa-n: Now i confused, isn't `JsonSerializer.Deserialize<char>("\"null\"");` would be be parsed to "null" string instead of `null` value?

I think the original example was the correct one i. e. `JsonSerializer.Deserialize<char>("null");`

buyaa-n:
> The other examples are changing one exception type to another and those are not considered breaking. In fact, going from NullRefException to a more user-friendly exception is encouraged.

They were within the examples of your breaking change comment so thought all of them breaking 😛, but you are right , I will update the issue accordingly, @gewarren let's remove the NRE, IOOR exceptions section from the doc, sorry for the confusion 🙏

buyaa-n:
@gewarren the issues now update as needed

includes/core-changes/serialization/3.1/jsonserializer-deserialize-throws-jsonexception.md
ahsonkhan: > I think the original example was the correct one

In this particular case, both are correct and behave the same way (with quotes or without), so you are right, if we want to show an example, either works.

Both were failing with null reference exception in 3.0 and both throw json exception now in 3.1/5.0 (but this change isn't considered breaking). See the example code snippet, below:
https://github.com/dotnet/docs/pull/20956#issuecomment-704669841

buyaa-n (approved):
Thank you @gewarren looks great

PR #20957, opened 2020-10-06
Fix link to design document

terrajobst:
We shouldn't point to PRs but to the merged document /cc @glennc


glennc (approved):

gewarren (approved):
LGTM

adegeo:
@terrajobst I think we should just remove this link completely. What goal is a user going to accomplish by clicking on that link? Are they going to read that entire design spec? It seems a bit out of place.

PR #20965, opened 2020-10-07
Fix language identifiers for CAxxxx rules

Youssef1313:
Currently, both languages are displayed regardless of the chosen one in the drop down at the top of the page.
This should fix that. But needs to be confirmed in staging first.

This is a "Search & Replace" PR. No manual changes are made. (Replaces `language="vbnet"` with `language="vb"`


tdykstra (approved):
Confirmed in staging, thanks @Youssef1313

gewarren:
Interesting, these snippets were inserted using the Docs Authoring Pack. Thanks @Youssef1313 for noticing and fixing!

PR #20979, opened 2020-10-08
SslStream/NegotiateStream breaking change

### LINKED ISSUE #20865: NegotiateStream and SslStream will no longer fail on multiple Begin operations without End
### NegotiateStream and SslStream will no longer fail on multiple *Begin operations without *End
This is change in how we handle error cases on security streams.

### Version introduced
5.0 RC1

### Old behavior
In the past sequence of 
- BeginXx
- BeginXx

would throw NotSupportedException exception.
 
### New behavior
Since the API calls are now backed by Task based implementation, this is no longer the case. 

### Reason for change
Switching internal implementation from APM to Tasks improves performance and decreases code complexity. 

### Recommended action

none. It was always invalid to call multiple Begin* operations without End*.

### Category

- [ ] ASP.NET Core
- [ ] C#
- [ ] Code analysis
- [ ] Core .NET libraries
- [ ] Cryptography
- [ ] Data
- [ ] Debugger
- [ ] Deployment
- [ ] Globalization
- [ ] Interop
- [ ] JIT
- [ ] LINQ
- [ ] MSBuild
- [x] Networking
- [ ] Printing
- [x] Security
- [ ] Serialization
- [ ] Visual Basic
- [ ] Windows Forms
- [ ] XML, XSLT

### Affected APIs
[SslStream.BeginAuthenticateAsServer](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslstream.beginauthenticateasserver?view=net-5.0)
[SslStream.BeginAuthenticateAsClient](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslstream.beginauthenticateasclient?view=net-5.0)
[NegotiateStream.BeginAuthenticateAsServer](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.negotiatestream.beginauthenticateasserver?view=net-5.0
)
[NegotiateStream.BeginAuthenticateAsClient](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.negotiatestream.beginauthenticateasclient?view=net-5.0
)

---
#### Issue metadata

* Issue type: breaking-change

cc: @stephentoub

gewarren:
Fixes #20865.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/networking?branch=pr-en-us-20979#negotiatestream-and-sslstream-allow-successive-begin-operations).


gewarren:
It seems like the Remarks sections for each of these overloads may need to be updated. Can you comment on that @wfurt? E.g. for [SslStream.BeginAuthenticateAsServer](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslstream.beginauthenticateasserver?view=net-5.0#System_Net_Security_SslStream_BeginAuthenticateAsServer_System_Security_Cryptography_X509Certificates_X509Certificate_System_AsyncCallback_System_Object_).

wfurt:
Is there specific part of the remarks @gewarren? They generally seem good to me. The Begin* operation does not block and End* must be called to complete the operation. 

Calling another Begin before calling End was never allowed and that logic was enforced in the Begin/End functions. With the 5.0 change the Begin/End is implemented on top of underlying Task. So it is possible to call Begin even before End is finished as ling as the underlying Task is done. I think there is marginal window of opportunity when something would fail in the past but may work in 5.0

includes/core-changes/networking/5.0/negotiatestream-sslstream-dont-fail-on-successive-begin-calls.md
```diff
@@ -0,0 +1,41 @@
+### NegotiateStream and SslStream allow successive Begin operations
+
+Error cases on security streams are handled differently and successive calls to `BeginAuthenticateAsServer` or `BeginAuthenticateAsClient` no longer fail.
```
wfurt, line 3: per the discussion in the original notes, this should be something like "may no longer fail" The exclusiveness is now enforced by the underlying task instead of the Begin/End pair so  there is some reface condition. It is still generally not OK to call Begin multiple times without calling End.

gewarren:
> Is there specific part of the remarks @gewarren? They generally seem good to me. The Begin* operation does not block and End* must be called to complete the operation.
> 
> Calling another Begin before calling End was never allowed and that logic was enforced in the Begin/End functions. With the 5.0 change the Begin/End is implemented on top of underlying Task. So it is possible to call Begin even before End is finished as ling as the underlying Task is done. I think there is marginal window of opportunity when something would fail in the past but may work in 5.0

I just thought since the implementation was no longer using APM that the wording may need to be changed, but it sounds like the answer is "no". Thanks for checking!

wfurt (approved):
LGTM. thanks.

PR #20985, opened 2020-10-08
Add POST example

Doerge:
## Summary

An example of sending JSON data would be very helpful in this example.

### DIFF
--- a/docs/csharp/tutorials/console-webapiclient.md
+++ b/docs/csharp/tutorials/console-webapiclient.md
@@ -377,6 +377,13 @@ Console.WriteLine(repo.LastPush);
 
 Your version should now match the [finished sample](https://github.com/dotnet/samples/tree/master/csharp/getting-started/console-webapiclient).
 
+## Sending data
+To send data to a REST service, we need to serialize the data to Json and wrap it in `StringContent` when sending:
+```
+string json = JsonSerializer.Serialize(new { data = new { foo = 42 } });
+Task<HttpResponseMessage> task = client.PostAsync($"{BaseUrl}/items/{item.Id}/subitems", new StringContent(json));
+```
+
 ## Conclusion
 
 This tutorial showed you how to make web requests, parse the result, and display properties of
--- a/docs/csharp/tutorials/console-webapiclient.md
+++ b/docs/csharp/tutorials/console-webapiclient.md
@@ -382,9 +382,11 @@ Your version should now match the [finished sample](https://github.com/dotnet/sa
 To send data to a REST service, we need to serialize the data to Json and wrap it in `StringContent` when sending:
 
 ```csharp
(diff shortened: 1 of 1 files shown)


docs/csharp/tutorials/console-webapiclient.md
```diff
@@ -377,6 +377,13 @@ Console.WriteLine(repo.LastPush);
 
 Your version should now match the [finished sample](https://github.com/dotnet/samples/tree/master/csharp/getting-started/console-webapiclient).
 
+## Sending data
+To send data to a REST service, we need to serialize the data to Json and wrap it in `StringContent` when sending:
+```
```
adegeo, line 382: Let's fix up this markdown a bit
```suggestion
## Sending data

To send data to a REST service, we need to serialize the data to Json and wrap it in `StringContent` when sending:

```csharp
```

adegeo (changes requested):
Thank you for this change. However, there is one major problem that would need to be addressed. In addition to my inline suggestion, this should be changed further. This is a tutorial article and therefore has a lot of progressive code that a reader would write and run. The example you posted doesn't flow into that narrative and even uses variables never declared (`BaseUrl`). If you were to add an example of a POST operation, you would want to do something with the GitHub service demonstrated with this tutorial and not a generic example such as the one you added.

This article also references a public sample: <https://github.com/dotnet/samples/tree/master/csharp/getting-started/console-webapiclient> and it should also be updated with a POST example.

Doerge:
Hi @adegeo ,

I completely agree, and it also crossed my mind when I typed this up, but since POST is creating a resource, the user needs to get add credentials and then this will snowball into a pretty big PR with OTP password on top. In addition running a POST request will create something on their github repo's/profiles. I don't think that would be ideal, although a gist would be ok I guess.

I couldn't find any public APIs that uses POST. Any ideas? Otherwise I'm also OK with just closing this PR again. It really was just meant as a footnote for the next person.

Best,

Doerge:
That's sort of my best effort on this. I understand if it's not up to standard, in which case you can just close this PR again.

Thanks for your hard work on the docs! It really helps people like me who are new to the libs.

BillWagner:
@Doerge @adegeo 

I like the idea of extending this to cover REST clients in more detail. That would include POST, PUT, and DELETE.

However, I think that would make this article much longer, and would include many more concepts. I'll propose that this should be a separate tutorial that extends the existing tutorial. That would focus on REST. It could assume this tutorial as a pre-requisite so that it wouldn't need to explain the HTTPClient class, and async web requests.

Thoughts?

Doerge:
I (mis)used the tutorial more as a reference to figure out how to correctly use the async methods. It would have been fine for me with just one example of a POST/PUT appended to the current tutorial, but for someone new to REST there might be a different need.

BillWagner:
closing as stale. We like the concept, but this has grown out of date.

PR #20924, opened 2020-10-05
Update pattern matching tutorial for C# 9.0

BillWagner:
Fixes #20594

Add one example for or patterns and relational patterns. Also, show a pre-pattern algorithm for comparison.

Tagging @adegeo for review since he recently updated this as well.

PR #20987, opened 2020-10-08
Update version (C# 8 -> C# 9)

Youssef1313:
The article was updated for C# 9 in #20924, but looks like this was forgotten to be updated.


BillWagner (approved):
Thanks @Youssef1313 

I'll :shipit: now.

Issue #20799, opened 2020-09-28
Update F1 keywords for tokens with multiple uses


BillWagner:
A number of C# tokens have different semantics depending on where they appear in the syntax tree. Historically, these tokens have gone to a page for the token, and the reader needs to select a link to the page with the correct content. We can do better. Here are the tokens that have multiple uses, the current page, and the list of uses for that token:

- `class`: goes to class declaration page.
  - declare a class
  - specify the class constraint on a generic type parameter.
- `default`: goes to a landing page to select the use.
  - default label of a `switch` statement
  - default operator or default literal.
- `enum`: destination describes declaring an enum.
  - declare an `enum` type
  - generic constraint where the base type must be an `enum` (derived from `System.Enum`)
- `in`: destination is a landing page for user to decide.
  - specify contravariance in a generic type parameter
  - parameter or argument modifier, meaning pass by read-only reference
  - `in` as part of a `foreach` statement.
  - part of a `from` clause in a LINQ query expression
  - part of a `join` clause in a LINQ query expression
- `internal`: destination is the page for `internal` access.
  - `internal` access
  - part of `protected internal` access
- `new`: goes to the `new` expression page, with a link to the method modifier.
  - new expressions
  - new method declaration modifier
  - `new` constraint on generic type parameter
- `out`: landing page for the user to specify
  - specify covariance on a generic type parameter
  - parameter or argument modifier meaning pass by reference, set by called method
- `private`: goes to `private` access modifier
  - `private` access
  - part of `private protected` access modifier
- `protected`: goes to the protected access modifier 
  - `protected` access
  - part of `protected internal` access
  - part of `private protected` access
- `ref`: goes to the page for `ref` parameter / argument modifier, `ref` return and `ref` local. Page has link to `ref struct` usage.
  - `ref` parameter / argument modifier
  - `ref` return value
  - `ref` local variable.
  - `ref` struct declaration
- `static` page goes to static modifier on class, field, method, local function, or lambda declaration
  - `static` modifier.
  - part of the `using static` directive.
- `struct`: page goes to `struct` declarations (which includes `ref struct` and `readonly struct`.
  - declare a struct
  - specify the `struct` constraint on a generic type parameter.
- `this` page is for `this` referring to the current instance of a type
  - refer to the current instance of a type
  - parameter modifier for first argument of an extension method
- `using` goes to a landing page for different uses.
  - `using` directive
  - `using static` directive 
  - `using` statement
- `partial`: already has separate pages for type and method
[...]

PR #21034, opened 2020-10-11
Add F1 keywords for default token semantics

TheSench:
## Summary

This PR adds two separate f1_keywords for different semantics of the `default` keyword.  This allows the F1 help to route to the appropriate page based on the context, rather than having to route to a generic `default` page.

Fixes part of #20799, Roslyn changes under dotnet/roslyn#48392.

### DIFF
--- a/docs/csharp/language-reference/keywords/switch.md
+++ b/docs/csharp/language-reference/keywords/switch.md
@@ -8,7 +8,6 @@ f1_keywords:
   - "case"
   - "case_CSharpKeyword"
   - "defaultcase_CSharpKeyword"
-  - "defaultcase"
 helpviewer_keywords:
   - "switch statement [C#]"
   - "switch keyword [C#]"
--- a/docs/csharp/language-reference/keywords/switch.md
+++ b/docs/csharp/language-reference/keywords/switch.md
@@ -7,6 +7,8 @@ f1_keywords:
   - "switch"
   - "case"
   - "case_CSharpKeyword"
+  - "defaultcase_CSharpKeyword"
+  - "defaultcase"
 helpviewer_keywords:
   - "switch statement [C#]"
   - "switch keyword [C#]"
--- a/docs/csharp/language-reference/operators/default.md
+++ b/docs/csharp/language-reference/operators/default.md
@@ -6,7 +6,6 @@ f1_keywords:
   - "default_CSharpKeyword"
   - "default"
   - "defaultvalue_CSharpKeyword"
-  - "defaultvalue"
 helpviewer_keywords: 
   - "default keyword [C#]"
 ---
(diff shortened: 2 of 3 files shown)


docs/csharp/language-reference/keywords/switch.md
```diff
@@ -7,6 +7,8 @@ f1_keywords:
   - "switch"
   - "case"
   - "case_CSharpKeyword"
+  - "defaultcase_CSharpKeyword"
+  - "defaultcase"
```
Youssef1313, line 11: This looks redundant. Only `defaultcase_CSharpKeyword` is needed based on your Roslyn PR.

docs/csharp/language-reference/operators/default.md
```diff
@@ -5,6 +5,8 @@ ms.date: 03/13/2020
 f1_keywords:
   - "default_CSharpKeyword"
   - "default"
+  - "defaultvalue_CSharpKeyword"
+  - "defaultvalue"
```
Youssef1313, line 9: Same comment applies here.

docs/csharp/language-reference/operators/default.md
```diff
@@ -5,6 +5,8 @@ ms.date: 03/13/2020
 f1_keywords:
   - "default_CSharpKeyword"
```
Youssef1313, line 6: We can delete this to avoid duplicate f1_keywords.
But **don't** delete it from `keywords/default.md` for backcompat with older Roslyn versions.

docs/csharp/language-reference/operators/default.md
```diff
@@ -5,6 +5,8 @@ ms.date: 03/13/2020
...
   - "default"
```
Youssef1313, line 7: We can delete this to avoid duplicate f1_keywords.
But **don't** delete it from `keywords/default.md` for backcompat with older Roslyn versions.

```suggestion
```

docs/csharp/language-reference/keywords/switch.md
TheSench: Add, I wasn't aware, I kept seeing `foo` and `foo_CSharpKeyword`, so I thought both were necessary.  Thanks for letting me know.

docs/csharp/language-reference/operators/default.md
davidwengier: I think this is a good change.

davidwengier (approved):

docs/csharp/language-reference/operators/default.md
TheSench: As mentioned in the Roslyn PR, it also means one less added keyword, since we can now just use the existing `default_CSharpKeyword` one that's already on the operator page.

TheSench:
@davidwengier based on the discussion in dotnet/roslyn#48500, I opted to reuse the existing `default_CSharpKeyword` and move it from the generic keyword page to the operator page.

Let me know if I should undo this change and I'll just rebase those commits out so that the overall history is less polluted.

BillWagner (approved):
This LGTM @TheSench 

I'll approve this, and :shipit: now.

You should see the changes on the live site tomorrow, during our regular publishing cycle.  After that, the roslyn PR can be tested against the live site.

PR #21037, opened 2020-10-12
Add f1-keyword for `class` used as generic type constraint

TheSench:
## Summary

This PR adds an f1_keyword for the generic where clause context of the `class` keyword.  This allows the F1 help to route to this usage of the keyword, rather than always routing to the class declaration page.

Fixes part of #20799, Roslyn changes under dotnet/roslyn#48506.

## Alternative Solution

This PR is unnecessary if the existing `whereconstraint_CSharpKeyword` keyword can be used by Roslyn for this context.

### DIFF
--- a/docs/csharp/language-reference/keywords/where-generic-type-constraint.md
+++ b/docs/csharp/language-reference/keywords/where-generic-type-constraint.md
@@ -7,6 +7,7 @@ f1_keywords:
   - "whereconstraint"
   - "whereconstraint_CSharpKeyword"
   - "classconstraint_CSharpKeyword"
+  - "structconstraint_CSharpKeyword"
 helpviewer_keywords:
   - "where (generic type constraint) [C#]"
 ---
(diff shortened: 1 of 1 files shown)


BillWagner:
@TheSench This change looks good to me.

is there are related roslyn PR to reference? /cc @davidwengier

TheSench:
> @TheSench This change looks good to me.
> 
> is there are related roslyn PR to reference? /cc @davidwengier

Yes, it should be linked in the initial comment, but here it is for reference:
dotnet/roslyn#48506

TheSench:
It looks like I missed some styling issues. I'll get those cleaned up later today.

davidwengier (approved):

TheSench:
I've started discussion on the necessity of this change versus using the existing `whereconstraint_CSharpKeyword`.  I _think_ this change may still be needed (and if so, an additional one for `struct` must be added).

TheSench:
> I've started discussion on the necessity of this change versus using the existing `whereconstraint_CSharpKeyword`. I _think_ this change may still be needed (and if so, an additional one for `struct` must be added).

Since this PR is either completely unnecessary, or new keywords are needed for _both_ `class` and `struct`, I've gone ahead and added `structcontraint` to this PR.

PR #21038, opened 2020-10-12
Added missing `protected` keyword

simplysiby:
Added missing `protected` keyword to the code description

## Summary

The 'protected' keyword was missing from the code description. This was added back to avoid confusing the reader.


BillWagner (approved):
Thanks for spotting this and fixing it @simplysiby 

I've reviewed the change and I'll :shipit: now.

PR #21043, opened 2020-10-12
Random IV in encrypt data example

duracellko:
Already since .NET 2 and MSDN lot of examples about symmetric cryptography treat IV the same way as encryption key. Unfortunately this anti-pattern is quite often used in production code too. Main purpose of IV is to add some randomness to encrypted data. So that encrypting the same text gives different output each time. This way an attacker cannot tell if the encrypted message/password is the same or not.

Therefore I updated Encrypting data and Decrypting data using AES pages. There are following changes:
1. The updated example stores IV together with encrypted data and it is not part of source code as secret. 
2. Decrypting example is organized the same way as encrypting example. In "snippets" folder.
3. Streams are disposed by `using` statement and not `Close` method.

Open question: I used new `using` form in C# 8 without nested code block. Is this ok, or should the example be compatible with previous versions of C#?


docs/standard/security/decrypting-data.md
```diff
@@ -22,7 +22,7 @@ Decryption is the reverse operation of encryption. For secret-key encryption, yo
...
+The following example illustrates how to create a new instance of the default implementation class for the <xref:System.Security.Cryptography.Aes> algorithm. The instance is used to perform decryption on a <xref:System.Security.Cryptography.CryptoStream> object. This example first creates a new instance of the **Aes** implementation class. Next it reads IV value from a managed stream called `myStream`. Next it creates a **CryptoStream** object and initializes it to the value of the managed stream called `myStream`. Next, the **CreateDecryptor** method from the **Aes** class is passed the read IV and the same key that was used for encryption and are then passed to the **CryptoStream** constructor.
```
IEvangelist, line 25:
```suggestion
The following example illustrates how to create a new instance of the default implementation class for the <xref:System.Security.Cryptography.Aes> algorithm. The instance is used to perform decryption on a <xref:System.Security.Cryptography.CryptoStream> object. This example first creates a new instance of the <xref:System.Security.Cryptography.Aes> implementation class. It reads the initialization vector (IV) value from a managed stream variable, `myStream`. Next it instantiates a <xref:System.Security.Cryptography.CryptoStream> object and initializes it to the value of the `myStream` instance. The <xref:System.Security.Cryptography.SymmetricAlgorithm.CreateDecryptor%2A?displayProperty=nameWithType> method from the <xref:System.Security.Cryptography.Aes> instance is passed the IV value and the same key that was used for encryption.
```

docs/standard/security/decrypting-data.md
```diff
@@ -22,7 +22,7 @@ Decryption is the reverse operation of encryption. For secret-key encryption, yo
...
 
 ```vb
```
IEvangelist, line 27: If you're adding VB and C# code, all code examples must have functional equivalents. For example, this VB should have a C# equivalent - otherwise you end up with a mix of the two languages and the language dropdown doesn't work.

IEvangelist (changes requested):
Hi @duracellko

This looks good, thank you. We'll :shipit: when you make the suggested changes. Also, it is fine to use the C# 8 bits here - not a problem.

docs/standard/security/decrypting-data.md
duracellko: Am I missing something? The C# code is just below the VB code. Or should C# code be always first?
I didn't change this part and in currently published version both languages work.

duracellko:
I have a question. Would this repository join Hacktoberfest by adding tag? https://hacktoberfest.digitalocean.com/hacktoberfest-update

IEvangelist:
> I have a question. Would this repository join Hacktoberfest by adding tag? https://hacktoberfest.digitalocean.com/hacktoberfest-update

I'm actually not sure, let's ask @BillWagner about that.

docs/standard/security/decrypting-data.md
duracellko: But this VB snippet is followed by C#.

```vb
Dim aes As Aes = Aes.Create()
Dim cryptStream As New CryptoStream(myStream, aes.CreateDecryptor(key, iv), CryptoStreamMode.Read)
```

```csharp
Aes aes = Aes.Create();
CryptoStream cryptStream = new CryptoStream(myStream, aes.CreateDecryptor(key, iv), CryptoStreamMode.Read);
```

Unless we are talking about different snippet.

docs/standard/security/decrypting-data.md
```diff
@@ -34,98 +34,12 @@ Aes aes = Aes.Create();
...
-        //Catch any exceptions.
-        catch
-        {
-            Console.WriteLine("The decryption failed.");
-            throw;
-        }
-    }
-}
-```
+:::code language="csharp" source="snippets/decrypting-data/csharp/aes-edecrypt.cs":::
```
IEvangelist, line 39:
```suggestion
:::code language="csharp" source="snippets/decrypting-data/csharp/aes-decrypt.cs":::
```

IEvangelist:
Closed and reopened to trigger build validation... odd happens 😵

IEvangelist (approved):

PR #21045, opened 2020-10-12
Move ms extensions to top level of toc

### LINKED ISSUE #21023: Microsoft Extensions hides content
For me it's difficult to find config, DI, logging, etc because they are hidden under Microsoft Extensions 
I think Microsoft Extensions  is meaningless to new folks. All the other top level TOC are descriptive.

image

@tdykstra  is the TOC expert.

For example, filter by title on logging is even obscured:
image

I think @tdykstra  knows how to fix this. The 3 extension articles are advanced and not what beginners are looking for. It hides https://docs.microsoft.com/en-us/dotnet/core/extensions/logging

The extensions/logging articles might be better found under https://docs.microsoft.com/en-us/dotnet/core/diagnostics/logging-tracing

So far, I can't find a google search a customer is likely to use that finds the extensions/logging docs.

tdykstra:
Fixes #21023

### DIFF
--- a/docs/fundamentals/toc.yml
+++ b/docs/fundamentals/toc.yml
@@ -1503,44 +1503,42 @@ items:
       href: ../standard/attributes/retrieving-information-stored-in-attributes.md
   - name: Choose between anonymous and tuple types
     href: ../standard/base-types/choosing-between-anonymous-and-tuple.md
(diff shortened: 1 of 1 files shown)


IEvangelist (approved):

PR #21065, opened 2020-10-13
Added diagnostics overview to TOC

### LINKED ISSUE #20943: Overview doc for Fundamentals --> Tools and diagnostics TOC section

IEvangelist:
## Summary

Added diagnostics overview to TOC

Fixes #20943

### DIFF
--- a/docs/fundamentals/toc.yml
+++ b/docs/fundamentals/toc.yml
@@ -149,6 +149,8 @@ items:
     href: ../standard/whats-new/whats-new-in-dotnet-standard.md
 - name: Tools and diagnostics
   items:
+  - name: Overview
+    href: index.md
   - name: .NET Core SDK
     items:
     - name: Overview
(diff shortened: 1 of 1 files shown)


tdykstra (changes requested):
The tools and diagnostics TOC section includes SDK, CLI, global tools, etc., not just diagnostics.  This article only covers diagnostics and already is in the Overview node for that section, so it doesn't fit in this higher-level TOC node.

IEvangelist:
> The tools and diagnostics TOC section includes SDK, CLI, global tools, etc., not just diagnostics. This article only covers diagnostics and already is in the Overview node for that section, so it doesn't fit in this higher-level TOC node.

You are correct. I see that now. Do you have the details for the TOC changes that are being proposed, some of these new issues around it are vague and it is difficult to collaborate on?

tdykstra:
When you take up one of the overview-creation issues, ping me offline and we can go over specifically what content is needed.

PR #21104, opened 2020-10-14
Add info about object detection & GPU training

briacht:
## Summary

Have a few tooling updates, including Object detection in model builder and local GPU training for image classification


luisquintanilla (changes requested):
:shipit:

luisquintanilla (approved):

PR #21105, opened 2020-10-14
.NET Framework -> .NET

gewarren:
Updates a handful of instances of .NET Framework that should just be ".NET" in the /standard/ folder. There are still lots more to update though.

_(Hide whitespace changes in diff.)_

Related to #18691. Contributes to #21191.


docs/standard/attributes/applying-attributes.md
```diff
@@ -2,48 +2,51 @@
...
- [!code-vb[Conceptual.Attributes.Usage#3](../../../samples/snippets/visualbasic/VS_Snippets_CLR/conceptual.attributes.usage/vb/source1.vb#3)]  
-  
-## Applying Attributes at the Assembly Level  
- If you want to apply an attribute at the assembly level, use the **assembly** (`Assembly` in Visual Basic) keyword. The following code shows the **AssemblyTitleAttribute** applied at the assembly level.  
-  
+ [!code-vb[Conceptual.Attributes.Usage#3](../../../samples/snippets/visualbasic/VS_Snippets_CLR/conceptual.attributes.usage/vb/source1.vb#3)]
+
+## Apply attributes at the assembly level
+
+ If you want to apply an attribute at the assembly level, use the **assembly** (`Assembly` in Visual Basic) keyword. The following code shows the **AssemblyTitleAttribute** applied at the assembly level.
```
tdykstra, line 42: The C# keyword should be code format like the VB one, otherwise it could get translated. I would code-format the attribute names as well, but I'm not sure it's worth the effort to retrofit the whole doc that way.

```suggestion
 If you want to apply an attribute at the assembly level, use the `assembly` (`Assembly` in Visual Basic) keyword. The following code shows the **AssemblyTitleAttribute** applied at the assembly level.
```

tdykstra (approved):
Thanks for doing this, good timing as the new .NET intro doc links to many of these.  One nit suggestion.

Issue #15807, opened 2019-11-11
Add supported types article to System.Text.Json documentation


tdykstra:
Create supported types doc similar to what we have [here](https://docs.microsoft.com/en-us/dotnet/standard/serialization/binary-serialization#serializable-types) per [suggestion from @ahsonkhan](https://github.com/dotnet/docs/pull/14208#discussion_r324409240):

>I saw there was a somewhat exhaustive list for the other serializer (https://docs.microsoft.com/en-us/dotnet/standard/serialization/binary-serialization#serializable-types), but that isn't necessary here. At least not atm. Maybe in the future, we do this in a separate, linked page when the types supported grows.

Related to #12207 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6b74d77c-33e1-d44c-18c6-bd0b612f74de
* Version Independent ID: 9c6a7135-eb75-21a9-1d27-0880459523ca
* Content: [Migrate from Newtonsoft.Json to System.Text.Json - .NET](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to)
* Content Source: [docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md](https://github.com/dotnet/docs/blob/master/docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md)
* Product: **dotnet**
* GitHub Login: @tdykstra
* Microsoft Alias: **tdykstra**

tdykstra:
# [`JsonSerializer`](https://docs.microsoft.com/dotnet/api/system.text.json.jsonserializer?view=netcore-3.1) collection support in 3.1.x

All collections (any type that derives from [`IEnumerable`](https://docs.microsoft.com/dotnet/api/system.collections.ienumerable?view=netcore-3.1))
are supported for serialization with [`JsonSerializer`](https://docs.microsoft.com/dotnet/api/system.text.json.jsonserializer?view=netcore-3.1);
we simply call the [`.GetEnumerator()`](https://docs.microsoft.com/dotnet/api/system.collections.ienumerable.getenumerator?view=netcore-3.1)
method of the collection, and write the elements if they are supported. Deserialization is more complicated and is not supported for all collections.

This document gives a quick overview of which collections are supported.

## Built-in types

[...]

PR #21108, opened 2020-10-14
5.0 updates to System.Text.Json docs

### LINKED ISSUE #21313: Document new features in 5.0 for System.Text.Json

tdykstra:
Fixes #21313 

Document new features in 5.0 for System.Text.Json.

There are other outstanding issues for the System.Text.Json docs, this one addresses only the new features for 5.0.

[Internal review URL - how-to doc](https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to?branch=pr-en-us-21108&pivots=dotnet-5-0)
[Internal review URL - migration doc](https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to?branch=pr-en-us-21108&pivots=dotnet-5-0)
[Internal review URL - custom converters doc](https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?branch=pr-en-us-21108&pivots=dotnet-5-0)


adegeo:
Close/reopen to see if snippets system can download the latest .net rc or not...

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -698,6 +709,177 @@ With this option, the `Summary` property of the `WeatherForecastWithDefault` obj
...
+}
+Date: 10/20/2020 6:38:33 PM
+TemperatureC: 40
+Summary: Hot
+```
+
+::: zone-end
+
+::: zone pivot="dotnet-core-3-1"
+A constructors that specifies web defaults is not available in `System.Text.Json` 3.1.
```
gewarren, line 854:
```suggestion
A constructor that specifies web defaults is not available in `System.Text.Json` 3.1.
```

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -698,6 +709,177 @@ With this option, the `Summary` property of the `WeatherForecastWithDefault` obj
...
+::: zone-end
+
+::: zone pivot="dotnet-core-3-1"
+A constructor that takes an existing instance is not available in `System.Text.Json` 3.1.
+::: zone-end
+
+## Web defaults for JsonSerializerOptions
+
+::: zone pivot="dotnet-5-0"
+There is a <xref:System.Text.Json.JsonSerializerOptions> constructor that creates a new instance with [the default options that are used for web apps](xref:System.Text.Json.JsonSerializerDefaults.Web), as shown in the following example:
```
gewarren, line 830: Technically the constructor accepts a JsonSerializerDefaults object. It could be Web or something else, depending on what the user specifies.

gewarren (commented):
I know it's a WIP, but was reading through anyway.

docs/standard/serialization/system-text-json-how-to.md
tdykstra: Good point, I changed the wording so it doesn't imply this is the only possible usage of this constructor, though as a practical matter web defaults is the only scenario it's useful for at present.

docs/standard/serialization/system-text-json-converters-how-to.md
```diff
@@ -301,6 +326,24 @@ The following code registers the converter:
 
 [!code-csharp[](snippets/system-text-json-how-to/csharp/RoundtripStackOfT.cs?name=SnippetRegister)]
 
+## Handle null values
+
+By default, the serializer does not pass `null` to custom converters for reference types and primitive value types on serialization and deserialization:
```
steveharter, line 331: The serializer does pass `null` to value type converters, which normally would throw a JsonException. However, for some converters, like the built-in converter for the value type `JsonElement`, it writes a `null` value if the internal value is `JsonTokenType.Null`.

Also in 5.0 there is now a `HandleNull` public virtual that says whether or not to handle null.

At least this was the behavior in 3.1. cc @layomia ?

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -171,6 +220,150 @@ public class ExampleClass
...
 ```
 
+## Scenarios using JsonSerializer that are supported starting in 5.0
+
+### Quoted numbers
+
+::: zone pivot="dotnet-5-0"
+`Newtonsoft.Json` can serialize or deserialize numbers represented by JSON strings (surrounded by quotes). For example, it can accept: `{"DegreesCelsius":"23"}` instead of `{"DegreesCelsius":23}`. To enable that behavior in <xref:System.Text.Json>, set <xref:System.Text.Json.JsonSerializerOptions.NumberHandling%2A?displayProperty=nameWithType> to <xref:System.Text.Json.Serialization.JsonNumberHandling.WriteAsString> or <xref:System.Text.Json.Serialization.JsonNumberHandling.AllowReadingFromString>.
+
+If you're using `System.Text.Json` indirectly by using ASP.NET Core, you don't need to do anything to get behavior like `Newtonsoft.Json`. ASP.NET Core specifies [web defaults](system-text-json-how-to.md#web-defaults-for-jsonserializeroptions) when it uses `System.Text.Json`, which allows quoted numbers.
```
gewarren, line 230: Should it be "which allow" singular at the end?

gewarren (approved):
This looks like a lot of work!

The "how to serialize/deserialize" and "how to migrate" articles are both gigantic. It seems like this might be a good time to split them up. Maybe have a header node in the TOC with subarticles for each.

tdykstra:
>The "how to serialize/deserialize" and "how to migrate" articles are both gigantic. It seems like this might be a good time to split them up. Maybe have a header node in the TOC with subarticles for each.

Yes they are large. We discussed this issue when we did the doc planning, and the consensus of the product team (and what I've heard from other sources confirms it) is that people prefer the large articles and use find-on-page to go to what they want. For now I'm keeping to that plan, but I'll reconsider if I get feedback advocating break-up of the pages.

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
tdykstra: Could go either way -- "**specifies** web defaults, which allows..." or "specifies **web defaults**, which allow..."

docs/standard/serialization/system-text-json-converters-how-to.md
tdykstra: My summary of default behavior without the HandleNull set to true comes from https://github.com/dotnet/runtime/issues/34439

layomia: @steveharter is correct and dotnet/runtime#34439 should be updated.

```suggestion
By default, the serializer does not pass `null` to custom converters for reference types on serialization and deserialization. `JsonTokenType.Null` is passed to custom converters for value types on deserialization.
```

docs/standard/serialization/system-text-json-converters-how-to.md
```diff
@@ -301,6 +326,24 @@ The following code registers the converter:
...
+
+* For reference types, it returns a `null` instance on deserialization.
+* For primitive value types, it throws a `JsonException` or returns a default value, depending on if there's an internal converter.
```
layomia, line 334:
```suggestion
* For primitive value types, `JsonTokenType.Null` is passed to the converter on deserialization, which either throws a `JsonException` or returns a `default` value, depending on if there's an internal converter for the type, or given the implementation of a user-provided converter.
```

docs/standard/serialization/system-text-json-converters-how-to.md
```diff
@@ -301,6 +326,24 @@ The following code registers the converter:
...
+* It writes `null` directly with the writer on serialization.
+
+This behavior is primarily to optimize performance by skipping an extra call to the converter. In addition, it avoids forcing converters to check for `null` at the start of every `Read` and `Write` method override.
+
+For `Nullable<T>` value types, the serializer does pass `null` to converters because it isn't a valid CLR value for value types. This way, the converter can determine what to do with this "invalid" token.
```
layomia, line 339:
```suggestion
For  value types, the serializer passes `JsonTokenType.Null` values to converters on deserialization because it isn't a valid CLR value for value types. This way, the converter can determine what to do with this "invalid" token.
```

docs/standard/serialization/snippets/system-text-json-how-to-5-0/csharp/CustomConverterHandleNull.cs
```diff
@@ -0,0 +1,98 @@
...
+{
+    public class Point
+    {
+        public int X { get; }
+        public int Y { get; }
+
+        public Dictionary<int, int> AdditionalValues;
+
+        [JsonConverter(typeof(DescriptionConverter))]
+        public string Description { get; set; }
```
layomia, line 16: This is the only property needed for this example. Other properties can be removed to avoid distracting from the main scenario, particularly `AdditionalValues`.

docs/standard/serialization/snippets/system-text-json-how-to-5-0/csharp/CustomConverterHandleNull.cs
```diff
@@ -0,0 +1,98 @@
...
+            writer.WriteStringValue(value);
+        }
+    }
+
+    public class Program
+    {
+        private static JsonSerializerOptions options = 
+            new JsonSerializerOptions(JsonSerializerDefaults.Web)
+        {
+            IncludeFields = true,
```
layomia, line 48: This can be removed if the `AdditionalValues` prop is removed.

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -355,6 +434,19 @@ Here's an example type to serialize and JSON output:
 }
 ```
 
+::: zone pivot="dotnet-5-0"
+You can specify conditional exclusion by setting the [[JsonIgnore]](xref:System.Text.Json.Serialization.JsonIgnoreAttribute) attribute's `Condition` property. The <xref:System.Text.Json.Serialization.JsonIgnoreCondition> enum provides the following options:
+
+* `Always` - The property will always be ignored. If no `Condition` is specified, this option is assumed.
+* `Never` - The property will always be serialized and deserialized, regardless of the `IgnoreNullValues` global setting.
```
layomia, line 441:
```suggestion
* `Never` - The property will always be serialized and deserialized, regardless of the `DefaultIgnoreCondition` global setting.
```

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -108,15 +109,42 @@ A <xref:System.Text.Json.JsonSerializer.Serialize%2A> overload that takes a <xre
...
+::: zone pivot="dotnet-5-0"
+
+* .NET primitives that map to JavaScript primitives, such as numeric types, strings, and Boolean.
+* User-defined [plain old CLR objects (POCOs)](https://en.wikipedia.org/wiki/Plain_old_CLR_object).
+* One-dimensional and jagged arrays (`ArrayName[][]`).
+* `Dictionary<TKey,TValue>` where `TValue` is `object`, `JsonElement`, or a POCO.
+* Collections from the following namespaces.
+  * <xref:System.Collections>
+  * <xref:System.Collections.Generic>
+  * <xref:System.Collections.Immutable>
```
layomia, line 144: In both 5.0 and 3.1 `S.C.Concurrent`, `.Specialized`, and `.ObjectModel` collections are also supported - https://github.com/dotnet/docs/issues/15807#issuecomment-576941811.

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -108,15 +109,42 @@ A <xref:System.Text.Json.JsonSerializer.Serialize%2A> overload that takes a <xre
...
+::: zone-end
 
 Supported types include:
+::: zone pivot="dotnet-5-0"
+
+* .NET primitives that map to JavaScript primitives, such as numeric types, strings, and Boolean.
+* User-defined [plain old CLR objects (POCOs)](https://en.wikipedia.org/wiki/Plain_old_CLR_object).
+* One-dimensional and jagged arrays (`ArrayName[][]`).
+* `Dictionary<TKey,TValue>` where `TValue` is `object`, `JsonElement`, or a POCO.
+* Collections from the following namespaces.
```
layomia, line 141:
```suggestion
* Collections from the following namespaces, including dictionaries.
```

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -108,15 +109,42 @@ A <xref:System.Text.Json.JsonSerializer.Serialize%2A> overload that takes a <xre
...
+* [Fields](../../csharp/programming-guide/classes-and-structs/fields.md) are excluded.
+::: zone-end
 
 Supported types include:
+::: zone pivot="dotnet-5-0"
+
+* .NET primitives that map to JavaScript primitives, such as numeric types, strings, and Boolean.
+* User-defined [plain old CLR objects (POCOs)](https://en.wikipedia.org/wiki/Plain_old_CLR_object).
+* One-dimensional and jagged arrays (`ArrayName[][]`).
+* `Dictionary<TKey,TValue>` where `TValue` is `object`, `JsonElement`, or a POCO.
```
layomia, line 140: We support more dictionary types than this.

```suggestion
```

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -157,16 +186,31 @@ To deserialize from UTF-8, call a <xref:System.Text.Json.JsonSerializer.Deserial
...
 The following behaviors apply when deserializing JSON:
 
+::: zone pivot="dotnet-5-0"
+
 * By default, property name matching is case-sensitive. You can [specify case-insensitivity](#case-insensitive-property-matching).
 * If the JSON contains a value for a read-only property, the value is ignored and no exception is thrown.
-* Constructors for deserialization:
-  - On .NET Core 3.0 and 3.1, a parameterless constructor, which can be public, internal, or private, is used for deserialization.
-  - In .NET 5.0 and later, non-public constructors are ignored by the serializer. However, parameterized constructors can be used if a parameterless constructor isn't available.
+* Non-public constructors are ignored by the serializer. However, parameterized constructors can be used if a parameterless constructor isn't available.
```
layomia, line 193: The point about parameterized ctors doesn't necessarily follow.

```suggestion
* Non-public constructors are ignored by the serializer. 
* A parameterized constructor can be used if a parameterless constructor isn't available and there is only one public parameterized constructor on the type. If `[JsonConstructorAttribute]` is used, the specified constructor will be used for deserialization regardless of the presence of other constructors.
```

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -157,16 +186,31 @@ To deserialize from UTF-8, call a <xref:System.Text.Json.JsonSerializer.Deserial
...
+* Deserialization to immutable objects or read-only properties is supported. See [Immutable types and Records](#immutable-types-and-records).
+* By default, enums are supported as numbers. You can [serialize enum names as strings](#enums-as-strings).
+* By default, fields are excluded. You can [include fields](#include-fields).
+* By default, comments or trailing commas in the JSON throw exceptions. You can [allow comments and trailing commas](#allow-comments-and-trailing-commas).
+* The [default maximum depth](xref:System.Text.Json.JsonReaderOptions.MaxDepth) is 64.
+
+    When you use System.Text.Json indirectly in an ASP.NET Core app, some default behaviors are different. For more information, see [Web defaults for JsonSerializerOptions](#web-defaults-for-jsonserializeroptions).
```
layomia, line 200:
```suggestion
When you use System.Text.Json indirectly in an ASP.NET Core app, some default behaviors are different. For more information, see [Web defaults for JsonSerializerOptions](#web-defaults-for-jsonserializeroptions).
```

layomia, line 200: This statement is also true for 3.1. Maybe add a corresponding note and inline details about case sensitivity and property naming policy?

docs/standard/serialization/snippets/system-text-json-how-to-5-0/csharp/QuotedNumbers.cs
```diff
@@ -0,0 +1,59 @@
...
+            };
+
+            string forecastJson = JsonSerializer.Serialize<Forecast>
+                (forecast, options);
+            Console.WriteLine($"Output JSON:\n{forecastJson}");
+
+            options = new JsonSerializerOptions()
+            {
+                NumberHandling = JsonNumberHandling.AllowReadingFromString
+            };
```
layomia, line 38: Can we use the same `options` instance, but set `NumberHandling` to `JsonNumberHandling.AllowReadingFromString | JsonNumberHandling.WriteAsString`?

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -698,6 +802,118 @@ With this option, the `Summary` property of the `WeatherForecastWithDefault` obj
...
+::: zone-end
+
+::: zone pivot="dotnet-core-3-1"
+`JsonConstructorAttribute` and C# 9 Record support aren't available in `System.Text.Json` 3.1.
+::: zone-end
+
+## Copy JsonSerializerOptions
+
+::: zone pivot="dotnet-5-0"
+There is a <xref:System.Text.Json.JsonSerializerOptions> constructor that lets you create a new instance with the same options as an existing instance, as shown in the following example:
```
layomia, line 875: Can we link to the docs for the ctor?

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -698,6 +802,118 @@ With this option, the `Summary` property of the `WeatherForecastWithDefault` obj
...
+## Web defaults for JsonSerializerOptions
+
+::: zone pivot="dotnet-5-0"
+Here are the options that have different defaults for web apps:
+
+* <xref:System.Text.Json.JsonSerializerOptions.PropertyNameCaseInsensitive%2A> = `true`
+* <xref:System.Text.Json.JsonNamingPolicy> = <xref:System.Text.Json.JsonNamingPolicy.CamelCase>
+* <xref:System.Text.Json.JsonSerializerOptions.NumberHandling%2A> = <xref:System.Text.Json.Serialization.JsonNumberHandling.AllowReadingFromString>
+
+There's a <xref:System.Text.Json.JsonSerializerOptions> constructor that lets you create a new instance with [the default options that ASP.NET Core uses for web apps](xref:System.Text.Json.JsonSerializerDefaults.Web), as shown in the following example:
```
layomia, line 893: Perhaps we should generalize this section to be about the constructor, and the mention web defaults as one possilbe param. This should scale well if we have more defaults later on.

docs/standard/serialization/snippets/system-text-json-how-to-5-0/csharp/HttpClientExtensionMethods.cs
```diff
@@ -0,0 +1,44 @@
...
+        public string Name { get; set; }
+        public string Username { get; set; }
+        public string Email { get; set; }
+    }
+    public class Program
+    {
+        public static async Task Main()
+        {
+            var client = new HttpClient();
+            client.BaseAddress = new Uri("https://jsonplaceholder.typicode.com");
```
layomia, line 20: Do we have any Microsoft/dotnet provided dummy JSON producing endpoints? There was feedback previously about pointing to an external URL. FWIW I think this url is safe and we can keep using it if there's no other option.

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -35,6 +37,48 @@ The following table lists `Newtonsoft.Json` features and `System.Text.Json` equi
...
+| Minimal character escaping                            | ✔️ [Strict character escaping, configurable](#minimal-character-escaping) |
+| `NullValueHandling.Ignore` global setting             | ✔️ [IgnoreNullValues global option](system-text-json-how-to.md#exclude-all-null-value-properties) |
+| Allow comments                                        | ✔️ [ReadCommentHandling global setting](#comments) |
+| Allow trailing commas                                 | ✔️ [AllowTrailingCommas global setting](#trailing-commas) |
+| Custom converter registration                         | ✔️ [Order of precedence differs](#converter-registration-precedence) |
+| No maximum depth by default                           | ✔️ [Default maximum depth 64, configurable](#maximum-depth) |
+| `PreserveReferencesHandling` global setting           | ✔️ [ReferenceHandling global setting](#preserve-object-references-and-handle-loops) |
+| `ReferenceLoopHandling` global setting                | ✔️ [ReferenceHandling global setting](#preserve-object-references-and-handle-loops) |
+| Serialize or deserialize numbers in quotes            | ✔️ [NumberHandling global setting](#quoted-numbers) |
```
layomia, line 53: There are also per-property/field and per-type settings.

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -35,6 +37,48 @@ The following table lists `Newtonsoft.Json` features and `System.Text.Json` equi
...
 * Not supported, workaround is not practical or possible. If you rely on these `Newtonsoft.Json` features, migration will not be possible without significant changes.
 
+::: zone pivot="dotnet-5-0"
+| Newtonsoft.Json feature                               | System.Text.Json equivalent |
+|-------------------------------------------------------|-----------------------------|
+| Case-insensitive deserialization by default           | ✔️ [PropertyNameCaseInsensitive global setting](#case-insensitive-deserialization) |
+| Camel-case property names                             | ✔️ [PropertyNamingPolicy global setting](system-text-json-how-to.md#use-camel-case-for-all-json-property-names) |
+| Minimal character escaping                            | ✔️ [Strict character escaping, configurable](#minimal-character-escaping) |
+| `NullValueHandling.Ignore` global setting             | ✔️ [IgnoreNullValues global option](system-text-json-how-to.md#exclude-all-null-value-properties) |
```
layomia, line 46: For 5.0 we should use `DefaultIgnoreCondition`, i.e. `JsonIgnoreCondition.WhenWritingNull` instead.

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -35,6 +37,48 @@ The following table lists `Newtonsoft.Json` features and `System.Text.Json` equi
...
+| Allow comments                                        | ✔️ [ReadCommentHandling global setting](#comments) |
+| Allow trailing commas                                 | ✔️ [AllowTrailingCommas global setting](#trailing-commas) |
+| Custom converter registration                         | ✔️ [Order of precedence differs](#converter-registration-precedence) |
+| No maximum depth by default                           | ✔️ [Default maximum depth 64, configurable](#maximum-depth) |
+| `PreserveReferencesHandling` global setting           | ✔️ [ReferenceHandling global setting](#preserve-object-references-and-handle-loops) |
+| `ReferenceLoopHandling` global setting                | ✔️ [ReferenceHandling global setting](#preserve-object-references-and-handle-loops) |
+| Serialize or deserialize numbers in quotes            | ✔️ [NumberHandling global setting](#quoted-numbers) |
+| Deserialize to immutable classes and structs          | ✔️ [JsonConstructor, C# 9 Records](#deserialize-to-immutable-classes-and-structs) |
+| Support for fields                                    | ✔️ [IncludeFields global setting](#public-and-non-public-fields) |
```
layomia, line 55: There's also a per-field inclusion setting.

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -35,6 +37,48 @@ The following table lists `Newtonsoft.Json` features and `System.Text.Json` equi
...
+| `DefaultValueHandling` setting on `[JsonProperty]`    | ✔️ [JsonIgnore attribute](#conditionally-ignore-a-property)  |
+| Deserialize `Dictionary` with non-string key          | ✔️ [Supported](#dictionary-with-non-string-key) |
+| Support for non-public property setters and getters   | ✔️ [JsonInclude attribute](#internal-and-private-property-setters-and-getters) |
+| Support for a broad range of types                    | ⚠️ [Some types require custom converters](#types-without-built-in-support) |
+| Polymorphic serialization                             | ⚠️ [Not supported, workaround, sample](#polymorphic-serialization) |
+| Polymorphic deserialization                           | ⚠️ [Not supported, workaround, sample](#polymorphic-deserialization) |
+| Deserialize inferred type to `object` properties      | ⚠️ [Not supported, workaround, sample](#deserialization-of-object-properties) |
+| Deserialize JSON `null` literal to non-nullable value types | ⚠️ [Not supported, workaround, sample](#deserialize-null-to-non-nullable-type) |
+| `[JsonConstructor]` attribute                         | ⚠️ [Not supported, workaround, sample](#specify-constructor-to-use) |
```
layomia, line 66: This is supported now.

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -84,7 +129,11 @@ This is not an exhaustive list of `Newtonsoft.Json` features. The list includes
...
 
-If you're using `System.Text.Json` indirectly by using ASP.NET Core, you don't need to do anything to get behavior like `Newtonsoft.Json`. ASP.NET Core specifies the settings for [camel-casing property names](system-text-json-how-to.md#use-camel-case-for-all-json-property-names) and case-insensitive matching when it uses `System.Text.Json`. The default values are set in the [JsonOptions class](https://github.com/dotnet/aspnetcore/blob/1f56888ea03f6a113587a6c4ac4d8b2ded326ffa/src/Mvc/Mvc.Core/src/JsonOptions.cs#L22-L28).
+If you're using `System.Text.Json` indirectly by using ASP.NET Core, you don't need to do anything to get behavior like `Newtonsoft.Json`. ASP.NET Core specifies the settings for [camel-casing property names](system-text-json-how-to.md#use-camel-case-for-all-json-property-names) and case-insensitive matching when it uses `System.Text.Json`.
+
+::: zone pivot="dotnet-5-0"
+ASP.NET Core also enables [quoted numbers](#quoted-numbers) by default.
```
layomia, line 135:
```suggestion
ASP.NET Core also enables deserializing [quoted numbers](#quoted-numbers) by default.
```

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -171,6 +220,150 @@ public class ExampleClass
...
+* `DefaultContractResolver` lets you select properties to include or exclude, based on arbitrary criteria.
+* The `NullValueHandling` and `DefaultValueHandling` settings on `JsonSerializerSettings` let you specify that all null-value or default-value properties should be ignored.
+* The `NullValueHandling` and `DefaultValueHandling` settings on the `[JsonProperty]` attribute let you specify individual properties that should be ignored when set to null or the default value.
+
+<xref:System.Text.Json> provides the following ways to omit properties while serializing:
+
+::: zone pivot="dotnet-5-0"
+
+* The [[JsonIgnore]](system-text-json-how-to.md#exclude-individual-properties) attribute on a property causes the property to be omitted from the JSON during serialization.
+* The [IgnoreNullValues](system-text-json-how-to.md#exclude-all-null-value-properties) global option lets you exclude all null-value properties.
```
layomia, line 271:
```suggestion
* The [DefaultIgnoreCondition](system-text-json-how-to.md#exclude-all-null-value-properties) global option lets you exclude all null-value properties.
```

Maybe we should merge this with the info on L273.

FYI we are marking `IgnoreNullValues` as obsolete in .NET 6.0.

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -171,6 +220,150 @@ public class ExampleClass
...
+* The JSON that is created for the second `Person` object contains a reference to that identifier instead of property values.
+
+`Newtonsoft.Json` also has a `ReferenceLoopHandling` setting that lets you ignore circular references rather than throw an exception.
+
+::: zone pivot="dotnet-5-0"
+To preserve references and handle circular references in <xref:System.Text.Json>, set <xref:System.Text.Json.JsonSerializerOptions.ReferenceHandler%2A?displayProperty=nameWithType> to <xref:System.Text.Json.Serialization.ReferenceHandler.Preserve%2A>. The `ReferenceHandler.Preserve` setting is equivalent to `ReferenceLoopHandling` = `true` and `PreserveReferencesHandling` = `true` in `Newtonsoft.Json`.
+
+Some related `Newtonsoft.Json` features are not supported:
+
+* [ReferenceResolver](https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonSerializer_ReferenceResolver.htm)
```
layomia, line 344: We expose our own `ReferenceResolver` type which allows folks to implement custom preserve handling logic.

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -406,13 +550,17 @@ If you use a custom converter that follows the preceding sample:
 
 For more information about custom converters that recursively call `Serialize` or `Deserialize`, see the [Required properties](#required-properties) section earlier in this article.
 
-### Public and non-public fields
+### Internal and private property setters and getters
```
layomia, line 553:
```suggestion
### Non-`public` property setters and getters
```

here and elsewhere

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -698,6 +802,118 @@ With this option, the `Summary` property of the `WeatherForecastWithDefault` obj
...
+
+:::code language="csharp" source="snippets/system-text-json-how-to-5-0/csharp/PreserveReferences.cs" highlight="34":::
+
+Here are some other notes about preserving references:
+
+* `MaxDepth` validation is not affected.
+* The following types can be serialized as references but can't be deserialized as references:
+  * Immutable types
+  * Arrays
+  * Value types, such as structs
```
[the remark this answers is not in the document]
Jozkee: > We should also mention that we throw when reference metadata is found when deserializing these types.

We throw if an Immutable type or Array contains $`ref` or `$id`.
POCO value types can ignore `$id` but will throw if they contain `$ref`.

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -698,6 +802,118 @@ With this option, the `Summary` property of the `WeatherForecastWithDefault` obj
...
+* To determine if objects are equal, `System.Text.Json` uses the default equality of `Hashset` or `Dictionary`, which calls <xref:System.Object.GetHashCode%2A?displayProperty=nameWithType> or <xref:System.Object.Equals%2A?displayProperty=nameWithType>.
```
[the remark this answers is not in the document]
Jozkee: We don't use the same equality comparer as the Dictionary or Hashset, we use [`ReferenceEqualityComparer.Instance`](https://docs.microsoft.com/dotnet/api/system.collections.generic.referenceequalitycomparer.instance?view=net-5.0)

From docs:
> An IEqualityComparer<T> that uses reference equality (ReferenceEquals(Object, Object)) instead of value equality (Equals(Object)) when comparing two object instances.

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -698,6 +802,118 @@ With this option, the `Summary` property of the `WeatherForecastWithDefault` obj
...
+* An exception is thrown if there are properties other than `$ref` in a JSON reference object.
```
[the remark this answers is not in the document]
Jozkee: > What does "JSON reference object" mean?

We should probably say "JSON object that contains a $ref metadata" since that's what we use on the [`ReferenceHandler.Preserve`](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.serialization.referencehandler.preserve?view=net-5.0) notes.

Those other three bullets that you mentioned are already specified in said doc.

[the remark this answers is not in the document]
layomia: Nice, we should link to this doc then: https://docs.microsoft.com/en-us/dotnet/api/system.text.json.serialization.referencehandler.preserve?view=net-5.0.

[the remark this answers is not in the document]
Jozkee: We should at least link to [ReferenceHandler.Preserve](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.serialization.referencehandler.preserve?view=net-5.0) in this doc.

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -171,6 +220,150 @@ public class ExampleClass
...
+
+`Newtonsoft.Json` has a `PreserveReferencesHandling` setting on `JsonSerializerSettings` that lets you serialize by reference:
+
+* An identifier metadata is added to the JSON created for the first `Person` object.
+* The JSON that is created for the second `Person` object contains a reference to that identifier instead of property values.
+
+`Newtonsoft.Json` also has a `ReferenceLoopHandling` setting that lets you ignore circular references rather than throw an exception.
+
+::: zone pivot="dotnet-5-0"
+To preserve references and handle circular references in <xref:System.Text.Json>, set <xref:System.Text.Json.JsonSerializerOptions.ReferenceHandler%2A?displayProperty=nameWithType> to <xref:System.Text.Json.Serialization.ReferenceHandler.Preserve%2A>. The `ReferenceHandler.Preserve` setting is equivalent to `ReferenceLoopHandling` = `true` and `PreserveReferencesHandling` = `true` in `Newtonsoft.Json`.
```
Jozkee, line 340:
```suggestion
To preserve references and handle circular references in <xref:System.Text.Json>, set <xref:System.Text.Json.JsonSerializerOptions.ReferenceHandler%2A?displayProperty=nameWithType> to <xref:System.Text.Json.Serialization.ReferenceHandler.Preserve%2A>. The `ReferenceHandler.Preserve` setting is equivalent to `PreserveReferencesHandling` = `PreserveReferencesHandling.All` in `Newtonsoft.Json`.
```

I wouldn't say that `ReferenceHandler.Preserve` is equivalent to Newtonsoft's `ReferenceLoopHandling.Ignore`, we also have plans to add Ignore as a new feature for 6.0.

Also, [`PreserveReferencesHandling`](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_PreserveReferencesHandling.htm) is an enum, not a bool.

docs/standard/serialization/snippets/system-text-json-how-to-5-0/csharp/QuotedNumbers.cs
tdykstra: That results in `System.InvalidOperationException: Serializer options cannot be changed once serialization or deserialization has occurred.`

tdykstra: Thanks!

docs/standard/serialization/system-text-json-how-to.md
tdykstra: > generalize this section to be about the constructor, and then mention web defaults as one possible param. This should scale well if we have more defaults later on.

The emphasis in the conceptual docs is covering scenarios rather than APIs -- for constructor documentation we have the API ref, and for the scenario of needing to apply web defaults we have this section of a conceptual doc, which explains what the defaults are and how to apply them. An alternative is  to just note what the defaults are (since that doesn't show up in the API ref) and link to the constructor API ref, and omit the code example.

docs/standard/serialization/system-text-json-how-to.md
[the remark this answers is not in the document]
Jozkee: @tdykstra I think it would say: "this feature cannot be used to preserve value types and/or immutable types"
Reasons: 
* Value types: we don't have a reference per se.
* Immutable types: they are created until the payload is fully read, which makes it impossible to preserve a reference of itself that could be within the payload e.g: 
```cs
var options = new JsonSerializerOptions { ReferenceHandler = ReferenceHandler.Preserve };
var obj = JsonSerializer.Deserialize<ImmutableDictionary<string, object>>(@"{""$id"": ""1"", ""property1"": {""$ref"": ""1""} }", options);
```

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
tdykstra: How do you allow quoted numbers on a per-property or per-type basis?

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -246,7 +421,7 @@ To implement type inference for `object` properties, create a converter like the
 * `NullValueHandling` is set to `Ignore`, and
 * During deserialization, the JSON contains a null value for a non-nullable value type.
 
-In the same scenario, <xref:System.Text.Json> does throw an exception. (The corresponding null handling setting is <xref:System.Text.Json.JsonSerializerOptions.IgnoreNullValues?displayProperty=nameWithType>.)
+In the same scenario, <xref:System.Text.Json> does throw an exception. (For the corresponding null-handling setting in `System.Text.Json`, see [Conditionally ignore a property](#conditionally-ignore-a-property).)
```
layomia, line 424: Should this link to notes about per-property ignoring or the global setting i.e. `DefaultIgnoreCondition.WhenWritingDefault`?

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -274,7 +449,14 @@ After deserialization, the `Date` property has 1/1/0001 (`default(DateTimeOffset
 
 ### Deserialize to immutable classes and structs
 
-`Newtonsoft.Json` can deserialize to immutable classes and structs because it can use constructors that have parameters. <xref:System.Text.Json> supports only public parameterless constructors. As a workaround, you can call a constructor with parameters in a custom converter.
+`Newtonsoft.Json` can deserialize to immutable classes and structs because it can use constructors that have parameters.
+
+::: zone pivot="dotnet-5-0"
+In <xref:System.Text.Json>, use the x to specify use of a parameterized constructor. Records in C# 9 are also immutable and are supported as deserialization targets. For more information, see [Immutable types and Records](system-text-json-how-to.md#immutable-types-and-records).
```
layomia, line 455: > use the x

Hmm what did you mean to write here 😄

layomia (commented):
Looks great. LGTM after these comments are addressed.

docs/standard/serialization/snippets/system-text-json-how-to-5-0/csharp/HttpClientExtensionMethods.cs
tdykstra: I looked for and could not find a Microsoft/dotnet provided dummy JSON producing endpoint.

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -698,6 +821,122 @@ With this option, the `Summary` property of the `WeatherForecastWithDefault` obj
...
+
+  Metadata is expected (although not mandatory), and the deserializer tries to understand it.
+
+The following code illustrates use of the `Preserve` setting.
+
+:::code language="csharp" source="snippets/system-text-json-how-to-5-0/csharp/PreserveReferences.cs" highlight="34":::
+
+This feature can't be used to preserve value types or immutable types. On deserialization, the instance of an immutable type is created after the entire payload is read. So it would be impossible to deserialize the same instance if a reference to it appears within the JSON payload.
+
+For value types, immutable types, and arrays, no reference metadata is serialized. On deserialization, an exception is thrown if `$ref` or `$id` is found. However, POCO value types ignore `$id` to make it possible to deserialize payloads that were serialized by using Newtonsoft.Json.  Newtonsoft.Json does serialize metadata for such types.
```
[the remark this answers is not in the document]
Jozkee: We have other validations that interfere with the one described in the paragraph.

* For arrays: we validate first that `$id` is read before `$values`.
* For objects: we throw if we read `$values`.

This one is fine as it is, also IIRC value type collections with $id and $values can be deserialized correctly by ignoring the preserve semantics.

Jozkee, line 844:
```suggestion
For value types, immutable types, and arrays, no reference metadata is serialized. On deserialization, an exception is thrown if `$ref` or `$id` is found. However, value types ignore `$id` (and `$values` in the case of collections) to make it possible to deserialize payloads that were serialized by using Newtonsoft.Json.  Newtonsoft.Json does serialize metadata for such types.
```

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -336,9 +411,20 @@ Enum string names can be deserialized as well, as shown in the following example
 
 By default, all public properties are serialized. If you don't want some of them to appear in the JSON output, you have several options. This section explains how to exclude:
 
+::: zone pivot="dotnet-5-0"
+
+* [Individual properties](#exclude-individual-properties)
+* [All read-only properties](#exclude-all-read-only-properties)
```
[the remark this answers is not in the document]
tdykstra: The text that introduces the list specifies serialization, and the other links in the list also apply when serializing. Is there some reason to call it out specifically for this item in the list?

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -698,6 +821,122 @@ With this option, the `Summary` property of the `WeatherForecastWithDefault` obj
...
+
+The following code illustrates use of the `Preserve` setting.
+
+:::code language="csharp" source="snippets/system-text-json-how-to-5-0/csharp/PreserveReferences.cs" highlight="34":::
+
+This feature can't be used to preserve value types or immutable types. On deserialization, the instance of an immutable type is created after the entire payload is read. So it would be impossible to deserialize the same instance if a reference to it appears within the JSON payload.
+
+For value types, immutable types, and arrays, no reference metadata is serialized. On deserialization, an exception is thrown if `$ref` or `$id` is found. However, POCO value types ignore `$id` to make it possible to deserialize payloads that were serialized by using Newtonsoft.Json.  Newtonsoft.Json does serialize metadata for such types.
+
+To determine if objects are equal, `System.Text.Json` uses <xref:System.Collections.Generic.ReferenceEqualityComparer.Instance%2A>, which uses reference equality (<xref:System.Object.ReferenceEquals(System.Object,System.Object)>) instead of value equality (<xref:System.Object.Equals(System.Object)?displayProperty=nameWithType> when comparing two object instances.
```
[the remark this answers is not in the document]
Jozkee: @layomia this should render to what you are asking here:
`<xref:System.Collections.Generic.ReferenceEqualityComparer.Instance?displayProperty=nameWithType>`

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -375,8 +474,17 @@ Here's an example type to serialize and JSON output:
...
 
-### Exclude all null value properties
+### Exclude all null-value properties
+
+::: zone pivot="dotnet-5-0"
+To exclude all null-value properties, set the <xref:System.Text.Json.JsonSerializerOptions.DefaultIgnoreCondition> property to <xref:System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull>, as shown in the following example:
 
+:::code language="csharp" source="snippets/system-text-json-how-to-5-0/csharp/IgnoreNullOnSerialize.cs" highlight="28":::
+
+This setting applies to serialization. For information about deserialization, see [Ignore null when deserializing](#ignore-null-when-deserializing).
```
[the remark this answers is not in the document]
tdykstra: I'll remove the section.

docs/standard/serialization/system-text-json-how-to.md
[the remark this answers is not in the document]
tdykstra: >This one is fine as it is

So the recommendation is to not mention $values here, or add remarks per the two bullet points about arrays and objects?

docs/standard/serialization/system-text-json-how-to.md
```diff
@@ -698,6 +821,122 @@ With this option, the `Summary` property of the `WeatherForecastWithDefault` obj
...
+
+To determine if objects are equal, `System.Text.Json` uses <xref:System.Collections.Generic.ReferenceEqualityComparer.Instance%2A>, which uses reference equality (<xref:System.Object.ReferenceEquals(System.Object,System.Object)>) instead of value equality (<xref:System.Object.Equals(System.Object)?displayProperty=nameWithType> when comparing two object instances.
+
+For more information about how references are serialized and deserialized, see <xref:System.Text.Json.Serialization.ReferenceHandler.Preserve%2A?displayProperty=nameWithType>.
+
+The <xref:System.Text.Json.Serialization.ReferenceResolver> class defines the behavior of preserving references on serialization and deserialization. Create a derived class to specify custom behavior. For an example, see [GuidReferenceResolver](https://github.com/dotnet/dotnet-docs/blob/master/docs/standard/serialization/snippets/system-text-json-how-to-5-0/csharp/GuidReferenceResolver.cs).
```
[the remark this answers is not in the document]
tdykstra: Yes, this will be in the dotnet/docs repo after the PR is merged and I will go back and update this link.

docs/standard/serialization/system-text-json-how-to.md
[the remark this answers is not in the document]
Jozkee: I left a suggestion but wasn't posted as part of this thread:
https://github.com/dotnet/docs/pull/21108#discussion_r517637520

docs/standard/serialization/system-text-json-how-to.md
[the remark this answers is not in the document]
layomia: Per your comment and after re-reading the section, I withdraw the suggestion.

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -15,7 +17,7 @@ helpviewer_keywords:
 
 This article shows how to migrate from [Newtonsoft.Json](https://www.newtonsoft.com/json) to <xref:System.Text.Json>.
 
-The `System.Text.Json` namespace provides functionality for serializing to and deserializing from JavaScript Object Notation (JSON). The `System.Text.Json` library is included in the [.NET Core 3.0](https://aka.ms/netcore3download) shared framework. For other target frameworks, install the [System.Text.Json](https://www.nuget.org/packages/System.Text.Json) NuGet package. The package supports:
+The `System.Text.Json` namespace provides functionality for serializing to and deserializing from JavaScript Object Notation (JSON). The `System.Text.Json` library is included in the runtime for [.NET Core 3.0](https://aka.ms/netcore3download) and later versions. For other target frameworks, install the [System.Text.Json](https://www.nuget.org/packages/System.Text.Json) NuGet package. The package supports:
```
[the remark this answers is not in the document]
tdykstra: In the discussions about what to call runtime-dependent vs. framework-dependent deployments, the term "shared framework" was called out specifically as a term we should avoid. 

cc @terrajobst

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -171,6 +220,149 @@ public class ExampleClass
 The JSON value could not be converted to System.String.
 ```
 
+## Scenarios using JsonSerializer that are supported starting in 5.0
```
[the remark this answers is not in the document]
tdykstra: Good point, I reorganized to put all JsonSerializer sections under one H2.

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
tdykstra: I believe the equivalent here is JsonSerializerOptions.IgnoreNullValues, which I removed from the main how-to doc per your comment:

>I'm not sure we should advertise this. IgnoreNullValues is obsolete starting in .NET 6.0, and we couldn't think of when it makes sense to ignore a null value received when deserializing.

PR #21110, opened 2020-10-15
TargetFramework breaking change

### LINKED ISSUE #21024: TargetFramework change to net5.0 may break customer logic relying on the TargetFramework value
## TargetFramework changed from netcoreapp3.1 to net5.0
Any MSBuild logic a customer has that tries to parse or use the TargetFramework value may fail.

### Version introduced

.NET 5.0

### Old behavior
```xml
<TargetFramework>netcoreapp3.1</TargetFramework>
```

### New behavior
```xml
```

### Reason for change
With the move to .net 5 both to simplify the TargetFramework value to the simple alias and allow some projects to include a TargetPlatform in the TargetFramework property, we've changed the formatting.

### Recommended action
**Existing project file**
```xml
<PropertyGroup Condition="$(TargetFramework.StartsWith('netcoreapp'))">
```
**Suggested change**
```xml
<PropertyGroup Condition="'$([MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)'))' == '.NETCoreApp'">
```

### Category

- [ ] ASP.NET Core
- [ ] C#
- [ ] Code analysis
- [ ] Core .NET libraries
- [ ] Cryptography
- [ ] Data
- [ ] Debugger
- [ ] Deployment
- [ ] Globalization
- [ ] Interop
- [ ] JIT
- [ ] LINQ
- [x] MSBuild
- [ ] Networking
- [ ] Printing
- [ ] Security
- [ ] Serialization
- [ ] Visual Basic
- [ ] Windows Forms
- [ ] XML, XSLT

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #21024.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/msbuild?branch=pr-en-us-21110#targetframework-change-from-netcoreapp-to-net).

### DIFF
--- a/includes/core-changes/msbuild/5.0/targetframework-name-change.md
+++ b/includes/core-changes/msbuild/5.0/targetframework-name-change.md
@@ -1,0 +1,48 @@
+### TargetFramework change from netcoreapp to net
+
+The value for the MSBuild `TargetFramework` property changed from `netcoreapp3.1` to `net5.0`. This can break code that relies on parsing the value of `TargetFramework`.
+
+#### Version introduced
+
+5.0
+
+#### Change description
+
+In .NET Core 1.0 - 3.1, the value for the MSBuild `TargetFramework` property starts with `netcoreapp`, for example, `netcoreapp3.1` for apps that target .NET Core 3.1. Starting in .NET 5.0, this value is simplified to just start with `net`, for example, `net5.0` for .NET 5.0.
+
+For more information, see [The future of .NET Standard](https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/) and [Target framework names in .NET 5](https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md).
+
+#### Reason for change
+
+- Simplifies the `TargetFramework` value.
+- Enables projects to include a `TargetPlatform` in the `TargetFramework` property.
+
+#### Recommended action
+
+If you have logic that parses the value of `TargetFramework`, you'll need to update it. For example, the following MSBuild condition relies on the value of `TargetFramework`.
+
+```xml
+<PropertyGroup Condition="$(TargetFramework.StartsWith('netcoreapp'))">
+```
+
+For this requirement, you can update the code to compare the target framework identifier instead.
+
+```xml
+<PropertyGroup Condition="'$([MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)'))' == '.NETCoreApp'">
+```
+
+#### Category
+
+MSBuild
+
+#### Affected APIs
+
+N/A
+
+<!--
+
+#### Affected APIs
+
+Not detectable via API analysis.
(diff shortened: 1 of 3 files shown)


includes/core-changes/msbuild/5.0/targetframework-name-change.md
```diff
@@ -0,0 +1,48 @@
...
+If you have logic that parses the value of `TargetFramework`, you'll need to update it. For example, the following MSBuild condition relies on the value of `TargetFramework`.
+
+```xml
+<PropertyGroup Condition="$(TargetFramework.StartsWith('netcoreapp'))">
+```
+
+For this requirement, you can update the code to compare the target framework identifier instead.
+
+```xml
+<PropertyGroup Condition="'$([MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)'))' == '.NETCoreApp'">
```
Youssef1313, line 31: Is the condition more complication than necessary?

From https://github.com/dotnet/roslyn/pull/46306/files, It looks like `Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'"` should work, while much shorter.

BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

includes/core-changes/msbuild/5.0/targetframework-name-change.md
gewarren: @marcpopMSFT Can you comment on Youssef's suggestion?

marcpopMSFT: I think we want to recommend customers use the method rather than the property as it ensures it goes through the NuGet parsing code to get the live value from the TargetFramework.  @dsplaisted may have a different opinion and Youssef is right that most of the time, these should be equivalent.  The method is slightly safer though more complex.

Youssef1313: @marcpopMSFT Can you clarify when they can have different results?

dsplaisted: It depends on where the condition is.  In the body of the project file, the `TargetFramework` has not yet been parsed into the `TargetFrameworkIdentifier`, `TargetFrameworkVersion`, etc. properties.  So you have to use the parsing functions.  If you are putting the conditions in a .targets file that is evaluated after Microsoft.NET.TargetFrameworkInference.targets, then you can use the `TargetFrameworkIdentifier` etc properties.

To make it more confusing, it's also possible to use the component properties on ItemGroups and items, because MSBuild does evaluation in multiple passes and the items are evaluated after all of the properties.

Youssef1313: @dsplaisted Thanks for the clarification!

PR #21125, opened 2020-10-16
Add f1-keyword for `using static` combination

TheSench:
## Summary

This PR adds an f1_keyword for the `using static` directive, allowing the `using` and `static` keywords to route here when they are found together as part of that directive.

Fixes part of #20799, Roslyn changes under dotnet/roslyn#48660.


BillWagner (approved):
Thanks for submitting this @TheSench 

Thanks again.

PR #21152, opened 2020-10-19
Update set-up-windows-containers-with-powershell.md

stackoverjoe:
## Summary

The example Dockerfile did not work. Updated to get tag ltsc2019 which will be installed with powershell 7, so "RUN powershell.." switched to "RUN pwsh..".

### DIFF
--- a/docs/architecture/containerized-lifecycle/design-develop-containerized-apps/set-up-windows-containers-with-powershell.md
+++ b/docs/architecture/containerized-lifecycle/design-develop-containerized-apps/set-up-windows-containers-with-powershell.md
@@ -10,9 +10,9 @@ With [Windows Containers](/virtualization/windowscontainers/about/index), you ca
 To use Windows Containers, you just need to write Windows PowerShell commands in the DockerFile, as demonstrated in the following example:
 
 ```dockerfile
-FROM microsoft/windowsservercore
+FROM mcr.microsoft.com/windows/servercore:ltsc2019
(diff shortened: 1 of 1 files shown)


sughosneo:
Hi @stackoverjoe, thank you for submitting this PR. 

If you use `FROM mcr.microsoft.com/windows/servercore:ltsc2019` then by default it uses PowerShell version 5. 

image

And that's the reason while running the step `RUN pwsh -Command Add-WindowsFeature Web-Server` fails with following error. 

image

So it's better to keep the `RUN powershell ` step as is.

```Dockerfile
FROM mcr.microsoft.com/windows/servercore:ltsc2019
LABEL Description="IIS" Vendor="Microsoft" Version="10"
RUN powershell -Command Add-WindowsFeature Web-Server
CMD [ "ping", "localhost", "-t" ]
```

And that can be run successfully.

image

Reference 

https://devblogs.microsoft.com/dotnet/we-made-windows-server-core-container-images-40-smaller/

stackoverjoe:
Hi @sughosneo thank you for the reply.

I was mistaken by thinking that mcr.microsoft.com/powershell installed the same image on windows! Thank you for the correction!

sughosneo:
> Hi @sughosneo thank you for the reply.
> 
> I was mistaken by thinking that mcr.microsoft.com/powershell installed the same image on windows! Thank you for the correction!

Looks good ! Thank you

nishanil (approved):

PR #21172, opened 2020-10-21
C# 9 reference update: covariant return types

### LINKED ISSUE #20577: Language reference: Miscellaneous
There are other small changes needed for the language reference:

- `override` should include the new rules for covariant return types.
- `foreach` `in` should add that `GetEnumerator` could be an extension method.
- The article on discards should mention that lambda expressions can use discards as arguments.
- The article on local function definitions should add that attribute can be applied to local functions. (Consider nullable attributes as an example).
- The conditional expression `?:` page should be reviewed to ensure it is accurate with the new target type conditional expressions.

pkulikov:
Fixes #20577 

Related, typo fixes in the proposal note: dotnet/csharplang#4045 (@BillWagner could you please review and merge that as well)

### DIFF
--- a/docs/csharp/language-reference/keywords/override.md
+++ b/docs/csharp/language-reference/keywords/override.md
@@ -1,7 +1,7 @@
 ---
 description: "override modifier - C# Reference"
 title: "override modifier - C# Reference"
-ms.date: 07/20/2015
+ms.date: 10/22/2020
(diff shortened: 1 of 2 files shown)


BillWagner (approved):
This LGTM @pkulikov 

I'll :shipit: now.

PR #19843, opened 2020-08-05
Moved C# operator snippets into the shared subfolder

pkulikov:
Per https://github.com/dotnet/docs/pull/17512#issuecomment-601874268

PR #21179, opened 2020-10-21
Moved C# built-in types snippets into the shared subfolder

pkulikov:
Similar to #19843
Later, record snippets can go to its own subfolder, if necessary


BillWagner (approved):
Thanks again @pkulikov 

The build is clean, and the snippets are all included. I'll :shipit: now.

PR #21195, opened 2020-10-23
C# 9 reference update: init and readonly structs/members

pkulikov:
A recent change as described by dotnet/csharplang#3884


BillWagner (approved):
Thanks @pkulikov 

I'm glad you caught this change and updated these.

I'll :shipit: now.

PR #21206, opened 2020-10-23
Add tutorial on collecting traces with PerfCollect

sywhang:
## Summary

This adds a tutorial on collecting traces with perfcollect. 

cc @tommcdon

### DIFF
--- a/docs/core/diagnostics/trace-perfcollect-lttng.md
+++ b/docs/core/diagnostics/trace-perfcollect-lttng.md
@@ -1,0 +1,247 @@
+---
+title: Tracing .NET applications with LTTng.
+description: A tutorial that walks you through collecting a trace with LTTng and PerfCollect in .NET.
+ms.topic: tutorial
+ms.date: 10/23/2020
+---
+
+# Tracing .NET applications with LTTng and PerfCollect
+
+**This article applies to: ✔️** .NET Core 2.1 SDK and later versions
+
+When a performance problem is encountered on Linux, collecting a trace with PerfCollect can be used to gather detailed information about what was happening on the machine at the time of the performance problem.
+
+PerfCollect is a bash script that leverages [Linux Tracing Tookit-Next Generation (LTTng)](https://lttng.org) to collect events written from the runtime or any [EventSource](xref:System.Diagnostics.Tracing.EventListener), as well as [perf](https://perf.wiki.kernel.org/) to collect CPU samples of the target process.
+
+
+
+## Preparing Your Machine
+Follow these steps to prepare your machine to collect a performance trace with PerfCollect.
+
+1. Download Perfcollect.
+
+	> ```bash
+	> curl -OL http://aka.ms/perfcollect
+	> ```
+
+2. Make the script executable.
+
+	> ```bash
+	> chmod +x perfcollect
+	> ```
+
+3. Install tracing prerequisites - these are the actual tracing libraries.
+
+	> ```bash
+	> sudo ./perfcollect install
+	> ```
+
+    This will install the following prerequisites on your machine:
+    
+    1. `perf`: the Linux Performance Events sub-system and companion user-mode collection/viewer application. `perf` is part of the Linux kernel source, but is not usually installed by default.
+
+    2. `LTTng`: Used to capture event data emitted at runtime by CoreCLR. This data is then used to analyze the behavior of various runtime components such as the GC, JIT and thread pool.
+
+## Collecting a Trace ##
+1. Have two shells available - one for controlling tracing, referred to as **[Trace]**, and one for running the application, referred to as **[App]**.
+
+2. **[App]** Setup the application shell with the following environment variables - this enables tracing configuration of CoreCLR.
+
+	> ```bash
+	> export COMPlus_PerfMapEnabled=1
+	> export COMPlus_EnableEventLog=1
+	> ```
+
+3. **[Trace]** Start collection.
+
+	> ```bash
+	> sudo ./perfcollect collect sampleTrace
+	> ```
+
+	Expected Output:
+
+	> ```bash
+	> Collection started.  Press CTRL+C to stop.
+	> ```
+
+4. **[App]** Run the app - let it run as long as you need to in order to capture the performance problem.  Generally, you don't need very long.  As an example, for a CPU investigation, 5-10 seconds of the high CPU situation is usually enough.
+
+	> ```bash
+	> dotnet run
+	> ```
+
+5. **[Trace]** Stop collection - hit CTRL+C.
+
+	> ```bash
+	> ^C
+	> ...STOPPED.
+	>
+	> Starting post-processing. This may take some time.
+	>
+	> Generating native image symbol files
+	> ...SKIPPED
+	> Saving native symbols
+	> ...FINISHED
+	> Exporting perf.data file
+	> ...FINISHED
+	> Compressing trace files
+	> ...FINISHED
+	> Cleaning up artifacts
+	> ...FINISHED
+	>
+	> Trace saved to sampleTrace.trace.zip
+	> ```
+
+	The compressed trace file is now stored in the current working directory.
+
+## Resolving Framework Symbols ##
+Framework symbols need to be manually generated at the time the trace is collected.  They are different than app-level symbols because the framework is pre-compiled while apps are just-in-time-compiled.  For code like the framework that was precompiled to native
+code, you need a special tool called crossgen that knows how to generate the mapping from the native code to the name of the
+methods.
+
+PerfCollect can handle most of the details for you, but it needs to have the crossgen tool and by default this is NOT part of
+the standard .NET distribution. If it is not there it warns you and refers you to these instructions. To fix things you
+need to fetch EXACTLY the right version of crossgen for the runtime you happen to be
+using.  If you place the crossgen tool in the same directory as the .NET Runtime DLLs (e.g. libcoreclr.so), then PerfCollect can
+find it and add the framework symbols to the trace file for you.
+
+Normally when you create a .NET application, it just generates the DLL for the code you wrote, using a shared copy of the runtime
+for the rest.   However you can also generate what is called a 'self-contained' version of an application and this contains all
+runtime DLLs.  It turns out that the crossgen tool is part of the Nuget package that is used to create these self-contained apps, so
+one way of getting the right crossgen tool is to create a self contained package of any application.
+
+So you could do the following
+   >```bash
+   > mkdir helloWorld
+   > cd helloWorld
+   > dotnet new console
+   > dotnet publish --self-contained -r linux-x64
+   >```
+Which creates a new helloWorld application and builds it as a self-contained app.    The only subtlety here is that if you have
+multiple versions of the .NET Runtime installed, the instructions above will use the latest.  As long as your app also uses
+the latest (likely) then these instructions will work without modification.
+
+As a side effect of creating the self-contained application the dotnet tool will download a nuget package
+called runtime.linux-x64.microsoft.netcore.app and place it in
+the directory ~/.nuget/packages/runtime.linux-x64.microsoft.netcore.app/VERSION, where VERSION is the version number of
+your .NET Core runtime (e.g. 2.1.0).   Under that is a tools directory and inside there is the crossgen tool you need.
+Starting with .NET Core 3.0, the package location is ~/.nuget/packages/microsoft.netcore.app.runtime.linux-x64/VERSION.
+
+The crossgen tool needs to be put next to the runtime that is actually used by your application.   Typically your app uses the shared
+version of .NET Core that is installed at /usr/share/dotnet/shared/Microsoft.NETCore.App/VERSION where VERSION is the
+version number of the .NET Runtime.   This is a shared location, so you need to be super-user to modify it.   If the
+VERSION is 2.1.0 the commands to update crossgen would be
+   >```bash
+   > sudo bash
+   > cp ~/.nuget/packages/runtime.linux-x64.microsoft.netcore.app/2.1.0/tools/crossgen /usr/share/dotnet/shared/Microsoft.NETCore.App/2.1.0
+   >```
+
+Once you have done this, perfcollect will use crossgen to include framework symbols.  The warning that perfcollect used to
+issue should go away.   This only has to be one once per machine (until you update your runtime).
+
+### Alternative: Turn off use of precompiled code ###
+
+If you don't have the abiltiy to update the .NET Runtime (to add crossgen), or if the above procedure did not work
+for some reason, there is another approach to getting framework symbols.   You can tell the runtime to simply
+not use the precompiled framework code.   The code will be Just in time compiled and the special crossgen tool
+is not needed.   This works, but will increase startup time for your code by something like a second or two.  If you
+can tolerate that (you probably can), then this is an alternative.   You were already setting environment variables
+in order to get symbols, you simply need to add one more.
+	> ```bash
+	> export COMPlus_ZapDisable=1
+	> ```
+With this change you should get the symbols for all .NET code.
+
+## Getting Symbols For the Native Runtime ##
+
+Most of the time you are interested in your own code, which perfcollect resolves by default.   Sometimes it is very
+useful to see what is going on inside the .NET Framework DLLs (which is what the last section was about), but sometimes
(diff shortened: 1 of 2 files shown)


docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,201 @@
...
+description: A tutorial that walks you through collecting a trace with LTTng and perfcollect in .NET.
+ms.topic: tutorial
+ms.date: 10/23/2020
+---
+
+# Tracing .NET applications with LTTng and PerfCollect
+
+**This article applies to: ✔️** .NET Core 2.1 SDK and later versions
+
+When a performance problem is encountered on Linux, collecting a trace with `perfcollect` can be used to gather detailed information about what was happening on the machine at the time of the performance problem.
```
davmason, line 12: Super Nit: I think "When performance problems are encountered" sounds better.

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,201 @@
...
+
+    1. `perf`: the Linux Performance Events sub-system and companion user-mode collection/viewer application. `perf` is part of the Linux kernel source, but is not usually installed by default.
+
+    2. `LTTng`: Used to capture event data emitted at runtime by CoreCLR. This data is then used to analyze the behavior of various runtime components such as the GC, JIT and thread pool.
+
+## Collecting a Trace
+
+1. Have two shells available - one for controlling tracing, referred to as **[Trace]**, and one for running the application, referred to as **[App]**.
+
+2. **[App]** Setup the application shell with the following environment variables - this enables tracing configuration of CoreCLR.
```
davmason, line 48: Set up should be two words

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,201 @@
...
+    > sudo ./perfcollect collect sampleTrace
+    > ```
+
+    Expected Output:
+
+    > ```bash
+    > Collection started.  Press CTRL+C to stop.
+    > ```
+
+4. **[App]** Run the app - let it run as long as you need to in order to capture the performance problem.
```
davmason, line 67: Should we suggest a good rule of thumb value, like 30 seconds?

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,201 @@
...
+    > ...FINISHED
+    >
+    > Trace saved to sampleTrace.trace.zip
+    > ```
+
+    The compressed trace file is now stored in the current working directory.
+
+## Resolving Framework Symbols
+
+Framework symbols need to be manually generated at the time the trace is collected. They are different than app-level symbols because the framework is pre-compiled while apps are just-in-time-compiled. For code like the framework that was precompiled to native code, you need a special tool called crossgen that knows how to generate the mapping from the native code to the name of the methods.
```
davmason, line 99:
```suggestion
Framework symbols need to be manually generated at the time the trace is collected. They are different than app-level symbols because the framework is pre-compiled while app code is just-in-time-compiled. For framework code that was precompiled to native code, you need to call crossgen to generate the mapping from the native code to the name of the methods.
```

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,201 @@
...
+
+`perfcollect` can handle most of the details for you, but it needs to have the crossgen tool and by default this is not part of the standard .NET distribution. If it is not there it warns you and refers you to these instructions. To fix things you need to fetch exactly the right version of crossgen for the runtime you happen to be using. If you place the crossgen tool in the same directory as the .NET Runtime DLLs (e.g. libcoreclr.so), then `perfcollect` can find it and add the framework symbols to the trace file for you.
```
davmason, line 101:
```suggestion
`perfcollect` can handle most of the details for you, but it needs to have crossgen available. By default crossgen is not installed with .NET distribution. If crossgen is not there `perfcollect` warns you and refers you to these instructions. To fix things you need to fetch the right version of crossgen for the runtime you are using. If you place the crossgen tool in the same directory as the .NET Runtime DLLs (e.g. libcoreclr.so), then `perfcollect` can find it and add the framework symbols to the trace file for you.
```

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,201 @@
...
+
+Normally when you create a .NET application, it just generates the DLL for the code you wrote, using a shared copy of the runtime for the rest.   However you can also generate what is called a 'self-contained' version of an application and this contains all runtime DLLs. The crossgen tool is part of the Nuget package that is used to create self-contained apps, so one way of getting the right crossgen tool is to create a self-contained package of your application.
```
davmason, line 103: Do we refer to crossgen as "the crossgen tool" elsewhere? It sounds awkward to me.

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,201 @@
...
+For example:
+
+   >```bash
+   > mkdir helloWorld
+   > cd helloWorld
+   > dotnet new console
+   > dotnet publish --self-contained -r linux-x64
+   >```
+
+This creates a new Hello World application and builds it as a self-contained app. Note that if you have multiple versions of the .NET Runtime installed, the instructions above will use the latest. As long as your app also uses the latest (likely) then these instructions will work without modification.
```
davmason, line 114: What if they are not using the latest version?

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,201 @@
...
+
+After this, you should get symbolic names for the native dlls when you run `perfcollect`.
+
+## Collecting in a Docker Container
+
+For more information on how to use `perfcollect` in container environments, refer to [Collect diagnostics in containers](./diagnostics-in-containers.md) documentation.
+
+## Viewing a Trace
+
+Traces are best viewed using PerfView on Windows.  Note that we're currently looking into porting the analysis pieces of PerfView to Linux so that the entire investigation can occur on Linux.
```
davmason, line 175: Link to perfview the first time it's mentioned?

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,201 @@
...
+
+3. Run PerfView.exe
+
+    > ```cmd
+    > PerfView.exe <path to trace.zip file>
+    > ```
+
+### Select a View
+
+PerfView will display the list of views that are supported based on the data contained in the trace file.
```
davmason, line 191: Don't we have documentation for perfview we could link here?

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,201 @@
...
+    > ...FINISHED
+    > Cleaning up artifacts
+    > ...FINISHED
+    >
+    > Trace saved to sampleTrace.trace.zip
+    > ```
+
+    The compressed trace file is now stored in the current working directory.
+
+## Resolving Framework Symbols
```
noahfalk, line 97: @sdmaclea @brianrob - Am I remembering correctly that this step is no longer required as long as the developer is tracing .Net Core 3.0+ (which has Steve's perf jitdump changes) and they have a sufficiently recent version of perf? Do we have a good description of what environments will have the updated version of perf or is this something that the perfcollect could warn about during the install step?

What I am hoping we can do is insert a tiny piece of text in the workflow above that says something like:
````
Recent versions of .NET Core and the Linux perf tool support automatic resolution of method names for 
framework code. If you are working with .NET Core version 2.1 or perf version <= XXX an extra step is
 necessary. See the [instructions below](some_anchor_link) for details.
````

File: docs/core/diagnostics/trace-perfcollect-lttng.md [hunk 11]
```diff
@@ -0,0 +1,201 @@
...
+
+**This article applies to: ✔️** .NET Core 2.1 SDK and later versions
+
+When a performance problem is encountered on Linux, collecting a trace with `perfcollect` can be used to gather detailed information about what was happening on the machine at the time of the performance problem.
+
+`perfcollect` is a bash script that leverages [Linux Tracing Tookit-Next Generation (LTTng)](https://lttng.org) to collect events written from the runtime or any [EventSource](xref:System.Diagnostics.Tracing.EventListener), as well as [perf](https://perf.wiki.kernel.org/) to collect CPU samples of the target process.
+
+## Preparing Your Machine
+
+Follow these steps to prepare your machine to collect a performance trace with `perfcollect`.
```
noahfalk, line 18: I would add a warning here that if you are using a container, it will need to have the SYS_ADMIN capability. There is a more a generic callout about containers far below, but for anyone following the steps in order this is a requirement they need to act on now.

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,201 @@
...
+
+```bash
+export COMPlus_ZapDisable=1
+```
+
+With this change you should get the symbols for all .NET code.
+
+## Getting Symbols For the Native Runtime
+
+Most of the time you are interested in your own code, which `perfcollect` resolves by default. Sometimes it is very useful to see what is going on inside the .NET DLLs (which is what the last section was about), but sometimes what is going on in the native runtime dlls (typically libcoreclr.so), is interesting.  `perfcollect` will resolve the symbols for these when it converts its data, but only if the symbols for these native DLLs are present (and are beside the library they are for).
```
noahfalk, line 144: Similar to the framework symbols, I would put a line of text+link in chronological order within the workflow that tells users they need to address this concern before they take the trace.

@brianrob - would it be hard to get perfcollect to do this automatically in the future?

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,201 @@
...
+    sudo cp mySymbols/* /usr/share/dotnet/shared/Microsoft.NETCore.App/2.1.0
+    ```
+
+After this, you should get symbolic names for the native dlls when you run `perfcollect`.
+
+## Collecting in a Docker Container
+
+For more information on how to use `perfcollect` in container environments, refer to [Collect diagnostics in containers](./diagnostics-in-containers.md) documentation.
+
+## Viewing a Trace
```
noahfalk, line 173: I would move this trace viewing information above the section about resolving framework symbols. Assuming the trace was collected succesfully this is the logical next step in the workflow.

docs/core/diagnostics/trace-perfcollect-lttng.md
noahfalk, line 175: For a Linux developer this could be a fairly disheartening message and PerfView on Linux hasn't been a priority for us so far. Ideally in addition to PerfView we could give a little more love to the Linux fans with some minimal information about how to get started viewing the data with perf, perfcollect's own view command and/or TraceCompass.

https://codeblog.dotsandbrackets.com/profiling-net-core-app-linux/
https://tooslowexception.com/analyzing-runtime-coreclr-events-from-linux-trace-compass/

noahfalk (commented):
Great to see this : ) I think it needs a little more work to streamline the flow and flesh out the Linux story but its going in a good direction.

docs/core/diagnostics/trace-perfcollect-lttng.md
brianrob: Yup, but it's .NET 5.  So starting with .NET 5, you no longer need to do anything in this section to resolve framework symbols.

docs/core/diagnostics/trace-perfcollect-lttng.md
brianrob: This would not be hard to do.  In most cases, we get this for free.  The specific cases where this doesn't work is when the binary gets stripped - I believe this only happens for distro-specific packages (e.g. deb, rpm)  It would not be hard to detect whether or not we should download the symbols at collection time.

docs/core/diagnostics/trace-perfcollect-lttng.md
brianrob: Agreed, folks tend to not like this very much as an answer, and I think that's fair.  Calling out some of these community driven experiences is good.

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,241 @@
...
+    > ```bash
+    > sudo ./perfcollect install
+    > ```
+
+    This will install the following prerequisites on your machine:
+
+    1. `perf`: the Linux Performance Events sub-system and companion user-mode collection/viewer application. `perf` is part of the Linux kernel source, but is not usually installed by default.
+
+    2. `LTTng`: Used to capture event data emitted at runtime by CoreCLR. This data is then used to analyze the behavior of various runtime components such as the GC, JIT and thread pool.
+
```
brianrob, line 46: There is a PR currently in review that would allow users to choose to use dotnet-trace to capture runtime events instead of LTTng.  You may want to keep an eye on this to update the documentation as well.

sywhang: Yup - I am aware of the PR from Christophe. Thanks for the reminder :)

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,241 @@
...
+2. Download the symbols. If your installed version of the .NET Core runtime is 2.1.0 the command to do this is
+
+    ```bash
+    mkdir mySymbols
+    dotnet symbol --symbols --output mySymbols  /usr/share/dotnet/shared/Microsoft.NETCore.App/2.1.0/lib*.so
+    ```
+
+3. Copy the symbols to the correct place
+
```
brianrob, line 232: This may or may not be something that can be done, since users may not be able to write to this location.  Another option is to use `perf buildid-cache` to add the symbol files to the buildid-cache.  This essentially tells perf about them.

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,241 @@
...
+    1. `perf`: the Linux Performance Events sub-system and companion user-mode collection/viewer application. `perf` is part of the Linux kernel source, but is not usually installed by default.
+
+    2. `LTTng`: Used to capture event data emitted at runtime by CoreCLR. This data is then used to analyze the behavior of various runtime components such as the GC, JIT and thread pool.
+
+## Collecting a Trace
+
+1. Have two shells available - one for controlling tracing, referred to as **[Trace]**, and one for running the application, referred to as **[App]**.
+
+2. **[Trace]** Start collection.
+
```
brianrob, line 52: Not sure if you want to include it here, but there is also some support for threadtime traces as well.  They really only view well on PerfView, and I consider them experimental, just because I haven't fully proved them out.  They are collectible by specifying `-threadtime`.

sywhang: Thanks Brian. In that case, I think I'll leave out the `-threadtime` flag until we are more confident about using this. I will be updating the docs frequently in the future so maybe this is something we can add in the next iteration for this doc along with what Christophe is adding via his PR.

brianrob: Sounds good.

docs/core/diagnostics/trace-perfcollect-lttng.md
```diff
@@ -0,0 +1,247 @@
...
+    > Trace saved to sampleTrace.trace.zip
+    > ```
+
+    The compressed trace file is now stored in the current working directory.
+
+## Viewing a Trace
+
+There are number of options for viewing the trace that was collected. Traces are best viewed using [PerfView](http://aka.ms/perfview>) on Windows, but they can be viewed directly on Linux using `PerfCollect` itself or `TraceCompass`.
+
+Recent versions of .NET Core and the Linux perf tool support automatic resolution of method names for framework code. If you are working with .NET Core version 3.1 or less, an extra step is necessary. See [Resolving Framework Symbols](#resolving-framework-symbols) for details.
```
noahfalk, line 104: I'd suggest moving these callouts about resolving framework and runtime symbols upwards to the 'preparing your machine' section. If the user reads this in order then they have already collected the trace by this point. It is too late for them to act on these new instructions without starting over.

noahfalk (approved):
Looks great! Thanks Sung : ) One comment inline

PR #21207, opened 2020-10-24
Add enumconstraint_CSharpKeyword

TheSench:
## Summary

This PR adds an f1_keyword for the generic where clause context of the `enum` keyword.  This allows the F1 help to route to this usage of the keyword, rather than always routing to the class declaration page.

Fixes part of #20799, Roslyn changes under dotnet/roslyn#TBD.


TheSench:
This is stemming from this part of #20799:

> `enum`: destination describes declaring an enum.
> * declare an `enum` type
> * generic constraint where the base type must be an `enum` (derived from `System.Enum`)

Is `where T : enum` now valid, or does it have to be `where T : System.Enum`?  If the new syntax is valid, then this PR requires more work to actually update the document, versus just adding an F1 keyword.

BillWagner (approved):
Thanks @TheSench 

PR #21209, opened 2020-10-24
Add using-statement F1 keyword

TheSench:
## Summary

This PR adds an f1_keyword for the using-statement context of the `using` keyword and moves the default `using` keyword to the `using-directive` page.  This allows the F1 help to route this keyword to pages more specific to the context in which it is being used.

Fixes part of #20799, Roslyn changes under dotnet/roslyn#48898.

### DIFF
--- a/docs/csharp/language-reference/keywords/using-directive.md
+++ b/docs/csharp/language-reference/keywords/using-directive.md
@@ -2,6 +2,8 @@
 description: "using directive - C# Reference"
 title: "using directive - C# Reference"
 ms.date: 07/20/2015
+f1_keywords: 
+  - "using_CSharpKeyword"
 helpviewer_keywords:
   - "using directive [C#]"
 ms.assetid: b42b8e61-5e7e-439c-bb71-370094b44ae8
--- a/docs/csharp/language-reference/keywords/using.md
+++ b/docs/csharp/language-reference/keywords/using.md
@@ -3,7 +3,6 @@ description: "using keyword - C# Reference"
 title: "using keyword - C# Reference"
 ms.date: 04/05/2019
 f1_keywords: 
-  - "using_CSharpKeyword"
   - "using"
 helpviewer_keywords: 
   - "using keyword [C#]"
--- a/docs/csharp/language-reference/keywords/using-statement.md
+++ b/docs/csharp/language-reference/keywords/using-statement.md
@@ -2,6 +2,8 @@
 description: "using statement - C# Reference"
 title: "using statement - C# Reference"
 ms.date: 05/29/2020
+f1_keywords:
+  - "using-statement_CSharpKeyword"
 helpviewer_keywords:
   - "using statement [C#]"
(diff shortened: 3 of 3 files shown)


docs/csharp/language-reference/keywords/using-directive.md
```diff
@@ -2,6 +2,8 @@
 description: "using directive - C# Reference"
 title: "using directive - C# Reference"
 ms.date: 07/20/2015
+f1_keywords: 
+  - "using_CSharpKeyword"
```
davidwengier, line 6: This is also on referenced in `keywords/using.md`, which would need to be changed as part of this.

TheSench: Ah, shoot, I was supposed to have already removed that.  Thanks for pointing it out and reminding me.

docs/csharp/language-reference/keywords/using.md
```diff
@@ -3,7 +3,6 @@ description: "using keyword - C# Reference"
 title: "using keyword - C# Reference"
 ms.date: 04/05/2019
 f1_keywords: 
-  - "using_CSharpKeyword"
   - "using"
```
BillWagner, line 6: If this keyword won't be generated by Visual Studio F1, I'd like to remove this article. It has no content, and it's only purpose is to provide a single click to send readers where they want to get to.

We'd also need to check for links and remove links to this page.

@davidwengier do you know if "using" is generated from Visual Studio?

BillWagner (commented):
Thanks again @TheSench 

I had one question that we should resolve before merging this.

Once that's figured out, we can complete the changes and merge this.

docs/csharp/language-reference/keywords/using.md
davidwengier: Right now, Roslyn appends `_CSharpKeyword` for any language construct it is trying to access help for, so unless there is an old version of Visual Studio that might be triggering this (that we care about) so as far as the `f1_keywords` goes I don't expect this to be ever used.

TheSench: What about users browsing the documentation outside of F1-help?  If they were to go to the [keywords](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/) page, what should they see?  I think leaving in the `using` page for that purpose makes sense.  There's no "using-statement" _keyword_, they would expect a "using" keyword page.

davidwengier: Ping @BillWagner in case you didn't see this comment.

I'm happy from the Roslyn side of this, and the PR on that side is ready to go when this is in, but this question is for the docs point of view.

davidwengier (approved):

docs/csharp/language-reference/keywords/using.md
BillWagner: Sorry, missed this comment.  Thanks for the ping.

This is good to go, and I'll :shipit: now.

PR #21236, opened 2020-10-26
Create Runtime libraries toc section

tdykstra:
Contributes to https://github.com/dotnet/docs/projects/104

[Internal review URL](https://review.docs.microsoft.com/en-us/dotnet/fundamentals/?branch=pr-en-us-21236)


docs/fundamentals/toc.yml
```diff
@@ -1385,6 +1387,19 @@ items:
...
+    - name: Overview
+      href: ../standard/attributes/index.md
+      displayName: attributes
+    - name: Apply attributes
+      href: ../standard/attributes/applying-attributes.md
+    - name: Write custom attributes
+      href: ../standard/attributes/writing-custom-attributes.md
+    - name: Retrieve information stored in attributes
+      href: ../standard/attributes/retrieving-information-stored-in-attributes.md
+- name: Fundamental runtime components
```
IEvangelist, line 1401: Even though we talked about this earlier, in looking at it again - while I think it is appropriate to move all these things as you have done...I'm getting hung up on the "Fundamental runtime components" top level TOC node. Configuration is not a runtime component, it happens before runtime - does that make sense?

What about **Fundamentals** or **Fundamental components**?

tdykstra: Configuration is a runtime component, not a run-time component, it's an extension to the runtime libraries.  We could change the name to Framework libraries or Runtime libraries if that would be more readily understandable.

IEvangelist (approved):

PR #21262, opened 2020-10-28
Notes for including SSL dependencies

plooploops:
## Summary

These are notes to prevent trimming SSL dependencies with .netcore 3.1 / .net 5.0 apps.

### DIFF
--- a/docs/core/deploying/trim-self-contained.md
+++ b/docs/core/deploying/trim-self-contained.md
@@ -32,6 +32,39 @@ When the code is indirectly referencing an assembly through reflection, you can 
 </ItemGroup>
 ```
 
+### Support for SSL Certificates
+
+Suppose one of the scenarios is to load SSL certificates within an ASP.Net Core app.  We want to ensure that when trimming we also prevent trimming assemblies that will help with loading SSL certificates.
+
+We can update our project file to include the following for ASP.Net Core 3.1:
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk.Web">
+  <PropertyGroup>...</PropertyGroup>
+  <!--Include the following for .aspnetcore 3.1-->
+  <ItemGroup>
+    <TrimmerRootAssembly Include="System.Net" />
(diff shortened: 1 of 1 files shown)


docs/core/deploying/trim-self-contained.md
```diff
@@ -32,6 +32,39 @@ When the code is indirectly referencing an assembly through reflection, you can
 </ItemGroup>
 ```
 
+### Support for SSL Certificates
+
+Suppose one of the scenarios is to load SSL certificates within an ASP.NET Core app.  We want to ensure that when trimming we also prevent trimming assemblies that will help with loading SSL certificates.
```
adegeo, line 37: - Remove double space
- I would be more aggressive with the wording. I feel like "Suppose" is too related to suggesting they may want to do something when the it's more like problem a exists here is a solution.

```suggestion
If your app loads SSL certificates, such as in an ASP.NET Core app, you'll want to ensure that when trimming you prevent trimming assemblies that will help with loading SSL certificates.
```

docs/core/deploying/trim-self-contained.md
```diff
@@ -32,6 +32,39 @@ When the code is indirectly referencing an assembly through reflection, you can
 </ItemGroup>
 ```
 
+### Support for SSL Certificates
```
adegeo, line 35: - Normal casing for titles.

```suggestion
### Support for SSL certificates
```

adegeo (approved):
I left some comments for you to consider.

docs/core/deploying/trim-self-contained.md
plooploops: Adding in the suggestion, thank you!

docs/core/deploying/trim-self-contained.md
plooploops: Adding in the suggestion, thank you!

plooploops:
@adegeo thank you for the approval.  Let me know if there's other checks we need to do before merging.

PR #21263, opened 2020-10-28
Call out code style enforcement as experimental

jmarolf:
## Summary

Calls out that enforcing code style on build is experimental


jmarolf:
CC: @mavasani @mikadumont

mavasani (approved):

mikadumont (commented):
I believe the proper doc syntax for notes is the following:

[!NOTE] This feature is currently experimental and may change between the .NET 5 and .NET 6 releases.

mavasani:
@jmarolf Do you plan to do a follow-up PR for mentioning the new property page for [enforcing code style on build](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#enforcecodestyleinbuild) (or will that need to go into VS docs repo)? Eventually, once we release `Microsoft.CodeAnalysis.NetAnalyzers` 5.0.0 NuGet package post .NET 5.0 GA, we should also mention the NuGet package here as an alternative.

jmarolf:
@mavasani good idea, I'll submit a PR to the VS docs repo after this

PR #21264, opened 2020-10-29
Mention support for RFC 3339

layomia:

### DIFF
--- a/docs/standard/datetime/system-text-json-support.md
+++ b/docs/standard/datetime/system-text-json-support.md
@@ -185,7 +185,7 @@ The following levels of granularity are defined for parsing:
     2. "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'FFFFFFFZ"
     3. "yyyy'-'MM'-'dd'T'HH':'mm':'ss('+'/'-')HH':'mm"
     4. "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'FFFFFFF('+'/'-')HH':'mm"
-    
+
     This level of granularity is compliant with [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6), a widely adopted profile of ISO 8601 used for interchanging date and time information.
 
 If there are decimal fractions for seconds, there must be at least one digit; `2019-07-26T00:00:00.` is not allowed.
@@ -210,7 +210,6 @@ The following levels of granularity are defined for formatting:
         Used to format a <xref:System.DateTime> with fractional seconds but without offset information.
 
 2. 'Date time'
-
     1. "yyyy'-'MM'-'dd'T'HH':'mm':'ssZ"
 
         Used to format a <xref:System.DateTime> without fractional seconds but with a UTC offset.
@@ -226,7 +225,7 @@ The following levels of granularity are defined for formatting:
     4. "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'FFFFFFF('+'/'-')HH':'mm"
 
         Used to format a <xref:System.DateTime> or <xref:System.DateTimeOffset> with fractional seconds and with a local offset.
-    
+
     This level of granularity is compliant with [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6).
 
 If the [round-trip format](../base-types/standard-date-and-time-format-strings.md#the-round-trip-o-o-format-specifier) representation of a
--- a/docs/standard/datetime/system-text-json-support.md
+++ b/docs/standard/datetime/system-text-json-support.md
@@ -185,6 +185,8 @@ The following levels of granularity are defined for parsing:
     2. "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'FFFFFFFZ"
     3. "yyyy'-'MM'-'dd'T'HH':'mm':'ss('+'/'-')HH':'mm"
     4. "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'FFFFFFF('+'/'-')HH':'mm"
+    
+    This level of granularity is compliant with [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6), a widely adopted profile of ISO 8601 used for interchanging date and time information.
 
 If there are decimal fractions for seconds, there must be at least one digit; `2019-07-26T00:00:00.` is not allowed.
 While up to 16 fractional digits are allowed, only the first seven are parsed. Anything beyond that is considered a zero.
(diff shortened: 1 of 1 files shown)


docs/standard/datetime/system-text-json-support.md
```diff
@@ -186,6 +186,8 @@ The following levels of granularity are defined for parsing:
     3. "yyyy'-'MM'-'dd'T'HH':'mm':'ss('+'/'-')HH':'mm"
     4. "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'FFFFFFF('+'/'-')HH':'mm"
 
+    This level of granularity is compliant with [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6), a widely adopted profile of ISO 8601 used for interchanging date and time information.
```
ahsonkhan, line 189: How will this appear in the final doc? Don't have it indented.

```suggestion
This level of granularity is compliant with [Internet Date/Time Format from RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6), a widely adopted profile of ISO 8601 used for interchanging date and time information.
```

image

docs/standard/datetime/system-text-json-support.md
```diff
@@ -224,6 +226,8 @@ The following levels of granularity are defined for formatting:
 
         Used to format a <xref:System.DateTime> or <xref:System.DateTimeOffset> with fractional seconds and with a local offset.
 
+    This level of granularity is compliant with [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6).
```
ahsonkhan, line 229:
```suggestion
This level of granularity is compliant with [Internet Date/Time Format from RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6).
```

ahsonkhan (approved):

ahsonkhan:
Add a similar note/link to RFC 3339 here:
https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to#specify-date-format

And here:
https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to

docs/standard/datetime/system-text-json-support.md
layomia: The note only applies to granularity no `5.` Per the examples section of RFC 3339 (https://tools.ietf.org/html/rfc3339#section-5.8), it is the full date/time format that is compliant, not the other levels of granularity. I think the current indentation makes sense.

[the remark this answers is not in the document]
layomia: > For parsing, RFC 3339 seems to allow arbitrary many fractional seconds (at least 1).

Our parser supports up to 16, and only the first 7 are meaningful. Does it make sense to allow any arbitrary number of digits and not throw for those?

We allow up to 16 digits to be present to better interop with other languages. Java for instance supports up to 9 fractional digits in their DateTime formatting implementation: https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html (search for "Fraction: Outputs the nano-of-second" on the page). I think our current behavior here makes sense. Any change here would be breaking.

> maybe it is worth highlighting that a bit more (since that seems to be the only difference). It's mentioned below, but adding a sentence here to connect it could be useful to some.

Thanks. I'll add a note about how there are some restrictions in the System.Text.Json implementation.

> how can I write a converter to allow that?

You can write a custom converter by copying the internal parsing logic in the reader and modifying the section that processes the fractional digits.

docs/standard/datetime/system-text-json-support.md
layomia: Similar to https://github.com/dotnet/docs/pull/21264/files#r514584458, this note only applies to the `'Date time'` format, so I think the indentation is correct here.

docs/standard/datetime/system-text-json-support.md
ahsonkhan: I see. Isn't 4 also compliant?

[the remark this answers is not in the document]
ahsonkhan: Gotcha. And what about:

> will any other input string that satisfies RFC 3339 result in JsonSerializer.Deserialize throwing? Space delimiter, capital/lower case Z/T, etc.

docs/standard/datetime/system-text-json-support.md
```diff
@@ -186,6 +186,12 @@ The following levels of granularity are defined for parsing:
...
+    - RFC 3339 allows the "T" and "Z" characters to be "t" or "z" respectively, but allows applications to limit support to just the upper-case variants. The implementation in System.Text.Json requires them to be "T" and "Z". A <xref:System.Text.Json.JsonException> will be thrown if input payloads contain "t" or "z" when reading `DateTime` and `DateTimeOffset` instances.
+    - RFC 3339 specifies that the date and time sections are separated by "T", but allows applications to separate them by a space (" ") instead. System.Text.Json requires date and time sections to be separated with "T". A <xref:System.Text.Json.JsonException> will be thrown if input payloads contain a space (" ") when reading `DateTime` and `DateTimeOffset` instances.
```
ahsonkhan, line 193: Awesome!

Does RFC 3339 **allow** space to be disallowed, just like lower case t/z?

docs/standard/datetime/system-text-json-support.md
layomia: No, 4  is not compliant because it does not contain a `time-second` ("ss") section.

docs/standard/datetime/system-text-json-support.md
layomia: Hmm

```
ISO 8601 defines date and time separated by "T".
Applications using this syntax may choose, for the sake of
readability, to specify a full-date and full-time separated by
(say) a space character.
```

I interpret this to mean that applications can emit it, but it doesn't explicitly say that applications have to be able to parse it.

tdykstra (approved):
LGTM

docs/standard/datetime/system-text-json-support.md
ahsonkhan: Makes sense. Thanks.

> We can always add support for this if it becomes a problem.

That is a good stance to have for the time being, and we are also clearly documenting it now if someone needs to know.

ahsonkhan:
> Add a similar small/non-obtrusive note/link to RFC 3339 here:
> https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to#specify-date-format
> 
> And here:
> https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to

Can we add the links to these other docs too? Or maybe, cross-link to our DateTime concept doc to increase its visibility/discoverability. We want folks to be able to understand if and how well our DateTime serialization support fits with ISO 8601 **and** RFC 3339.

PR #21285, opened 2020-10-30
Update fundamentals.md

svemuri-dev:
## Summary

There is an issue with continuation of sentence . 


Maoni0:
the original text is more accurate but needs some work to make it even more accurate -
> An application's roots include static fields, local variables and parameters on a thread's stack, and CPU registers. Each root either refers to an object on the managed heap or is set to null. The garbage collector has access to the list of active roots that the just-in-time (JIT) compiler and the runtime maintain. 

should be

> An application's roots include static fields, local variables on a thread's stack, CPU registers, GC handles and finalize queue. Each root either refers to an object on the managed heap or is set to null. The garbage collector can ask the rest of the runtime for these roots. 

it's true that the kind of roots listed previous were maintained by the JIT but there are additional kinds of roots that are maintained by the other components in the runtime.

docs/standard/garbage-collection/fundamentals.md
```diff
@@ -64,7 +64,7 @@ Allocating memory from the managed heap is faster than unmanaged memory allocati
...
+The garbage collector's optimizing engine determines the best time to perform a collection based on the allocations being made. When the garbage collector performs a collection, it releases the memory for objects that are no longer being used by the application. It determines which objects are no longer being used by examining the application's *roots*. An application's roots include static fields, local variables and parameters on a thread's stack, and CPU registers. Each root either refers to an object on the managed heap or is set to null. The garbage collector has access to the list of active roots that the just-in-time (JIT) compiler maintains. Using this list, the garbage collector creates a graph that contains all the objects that are reachable from the roots.
```
gewarren, line 67:
```suggestion
The garbage collector's optimizing engine determines the best time to perform a collection based on the allocations being made. When the garbage collector performs a collection, it releases the memory for objects that are no longer being used by the application. It determines which objects are no longer being used by examining the application's *roots*. An application's roots include static fields, local variables on a thread's stack, CPU registers, GC handles, and the finalize queue. Each root either refers to an object on the managed heap or is set to null. The garbage collector can ask the rest of the runtime for these roots. Using this list, the garbage collector creates a graph that contains all the objects that are reachable from the roots.
```

gewarren (approved):

PR #21307, opened 2020-11-02
Changing from properties to fields

AYS17:
Those are referred to as instance fields in the "Program building blocks" in the "Fields" section.

## Summary

Describe your changes here.


BillWagner:
Thanks for submitting this @amadosoft 

The fix is correct, but to promote good C# practices, I'd prefer changing the text to say "instance properties".

What do you think?

amadosoft:
@BillWagner  The problem is in the context of the example code.
https://docs.microsoft.com/en-us/dotnet/csharp/tour-of-csharp/program-building-blocks#fields

I think that would require either removing the entire paragraph about "Fields" or just altering the example code. But you are the expert here.

BillWagner:
@amadosoft Good point.  Thanks for correcting me.

BillWagner (approved):
Thanks for making this change, and helping me through the change.

I'll approve this, and :shipit: now.

PR #21323, opened 2020-11-03
$rootcert vs $rootCert; Is this intentional, or a mistake?

whereistimbo:
I was trying to follow the instruction to generate Root CA and Intermediate CA and I noticed the variable name was different ($rootcert and $rootCert). Is it supposed to be different?

## Summary

Describe your changes here.


adegeo:
Sorry for the delay in reviewing this. I don't think it makes a difference for PowerShell, but it's not bad to have them match. I also spotted one more spot where the change can be made and went ahead and did it for you.

Thanks for the contribution!

adegeo (approved):

PR #21329, opened 2020-11-03
Don't overload the term "parse a string"

gewarren:
No need for a redirection on the renamed file since it hasn't gone live yet.

In .NET, to "parse" a string seems to mean to convert it to a base type. Not to extract its parts into substrings.

Although Merriam-Webster says parse means to divide a sentence into its grammatical parts. So .NET is using parse in a strange way here.


adegeo:
I think this can be a good change, but it should be used with caution. String parsing is a common concept that people will search for. Also, there is a bit of a conflict with the `Substring` method. If someone is searching on split but sees substring in the results they may skip it. The title shouldn't have both method names in it:  "Extract substrings using String.Split (C# Guide)"

One case where I don't think it would be good to change is with the renamed `extract-substrings.md`. The description is `Learn about different techniques to extract parts of a string, including String.Split, regular expressions, and String.Substring.` The common thing to do with this is extract those parts to parse, look at the characters, do something with it. So I think the word "parse" is better than "substring" (also for the reason I said above). Thinking of it like a language "parser" https://github.com/eliben/pycparser for example.

gewarren:
> One case where I don't think it would be good to change is with the renamed `extract-substrings.md`. The description is `Learn about different techniques to extract parts of a string, including String.Split, regular expressions, and String.Substring.` The common thing to do with this is extract those parts to parse, look at the characters, do something with it. So I think the word "parse" is better than "substring" (also for the reason I said above). Thinking of it like a language "parser" https://github.com/eliben/pycparser for example.

There is already a file called parsing-strings.md (about converting strings to base types) in the same folder, so I thinks it's a bad idea to leave it named as parse-strings.md when it means something completely different. I would say that a language parser is using the term "parse" as M-W defines it, which is to break it up into parts of grammar, not splitting it up based on delimiters or a pattern.

BillWagner (approved):
This LGTM @gewarren 

I have a vague recollection of a conversation with Ron that we chose "parse" because of SEO research.  Can we run a time-based experiment to see if that changes the visibility of these articles?

gewarren:
> This LGTM @gewarren
> 
> I have a vague recollection of a conversation with Ron that we chose "parse" because of SEO research. Can we run a time-based experiment to see if that changes the visibility of these articles?

Sure, I opened #21339 to track that.

PR #21344, opened 2020-11-04
Publish updated nullable reference type spec

BillWagner:
See dotnet/csharplang#4100

Move the included specification for nullable reference types to the C# 9 folder.

Update the appropriate title metadata.

Because nullable reference types was a C# 8.0 feature, I am publishing this speclet in both C# 8.0 and C# 9.0.

### DIFF
--- a/docs/csharp/language-reference/proposals/toc.yml
+++ b/docs/csharp/language-reference/proposals/toc.yml
@@ -63,7 +63,7 @@
     - name: Nullable reference types - proposal
       href: ../../../../_csharplang/proposals/csharp-8.0/nullable-reference-types.md
     - name: Nullable reference types - specification
-      href: ../../../../_csharplang/proposals/csharp-8.0/nullable-reference-types-specification.md
+      href: ../../../../_csharplang/proposals/csharp-9.0/nullable-reference-types-specification.md
     - name: Recursive pattern matching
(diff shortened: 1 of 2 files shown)


docs/csharp/language-reference/proposals/toc.yml
```diff
@@ -88,6 +88,8 @@
       href: ../../../../_csharplang/proposals/csharp-9.0/records.md
     - name: Top-level statements
       href: ../../../../_csharplang/proposals/csharp-9.0/top-level-statements.md
+    - name: Nullable reference types - specification
```
gewarren, line 91: Why does this appear twice?

gewarren (approved):

docs/csharp/language-reference/proposals/toc.yml
BillWagner: Because the feature was added in C# 8.0, and is now being updated for C# 9.0. I think it will help findability. On the other hand, do you think it will be confusing to appear twice?

gewarren: I think it's probably fine. I did notice in the preview that "proposal" should be plural in the top level node for 8 and 9.

gewarren: @BillWagner Not sure if you saw the previous comment.

PR #21348, opened 2020-11-05
Array initializer can be used in more places

AYS17:
Array initializer can be used to initialize "auto properties" as well. Specifying that it can be used with "Local variable and field declarations" might not be accurate. Also future versions of C# might add more possibilities(perhaps for method parameter array default value). So I think that saying "Array initialization" covers all current possible use cases and is also future-proof.

## Summary

Describe your changes here.


BillWagner (approved):
Thanks for making these changes @amadosoft 

I've reviewed them, and I'll :shipit: now.

PR #21363, opened 2020-11-06
Fix build warning for proposals/csharp-7.0/task-types.md

Youssef1313:
This PR fixes a build warning for not having a title for this file.

The other fix to the warning is to exclude this file in docfx.json.

I'm not sure which fix is the correct/preferred one.

The file was recently added in https://github.com/dotnet/csharplang/pull/4107.


BillWagner (approved):
Thanks @Youssef1313 

PR #21364, opened 2020-11-06
C# keywords index: fixed link to `unmanaged` reference

pkulikov:


BillWagner (approved):
Thanks for making this fix @pkulikov That is a better destination for the link.  I'll :shipit: now.

PR #21375, opened 2020-11-07
LastIndexOf breaking change

### LINKED ISSUE #21372: Breaking change: Improved String.LastIndexOf handling of empty values
## Breaking change: Improved String.LastIndexOf handling of empty values

`string.LastIndexOf` and related APIs now return correct values when searching for a zero-length (or zero-length equivalent) substring within a larger string.

### Version introduced

.NET 5.0

### Old behavior

In .NET Framework and .NET Core, `string.LastIndexOf` and related APIs could return an incorrect value when the caller searches for a zero-length substring.

```cs
Console.WriteLine("Hello".LastIndexOf("")); // prints '4' (incorrect)

ReadOnlySpan<char> span = "Hello";
Console.WriteLine(span.LastIndexOf("")); // prints '0' (incorrect)
```

### New behavior

Beginning with .NET 5.0, these APIs return the correct value for _LastIndexOf_.

```cs
Console.WriteLine("Hello".LastIndexOf("")); // prints '5' (correct)

ReadOnlySpan<char> span = "Hello";
Console.WriteLine(span.LastIndexOf("")); // prints '5' (correct)
```

In these examples, __5__ is the correct answer because `"Hello".Substring(5)` and `"Hello".AsSpan().Slice(5)` both produce an empty string, which is trivially equal to the empty substring being sought.

### Reason for change

This change was part of an overall bug fixing effort around string handling for .NET 5. This also helps unify our behavior between Windows and non-Windows platforms. See https://github.com/dotnet/runtime/issues/13383 and https://github.com/dotnet/runtime/issues/13382 for more information.

### Recommended action

Developers needn't take any action. The .NET 5.0 runtime provides the new behaviors automatically.

There is no compatibility switch to restore the old behavior.

### Category

- Core .NET libraries

### Affected APIs

- [`string.LastIndexOf`](https://docs.microsoft.com/dotnet/api/system.string.lastindexof)
- [`CompareInfo.LastIndexOf`](https://docs.microsoft.com/dotnet/api/system.globalization.compareinfo.lastindexof)
- [`MemoryExtensions.LastIndexOf`](https://docs.microsoft.com/dotnet/api/system.memoryextensions.lastindexof)

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #21372.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/corefx?branch=pr-en-us-21375#lastindexof-has-improved-handling-of-empty-search-strings).


includes/core-changes/corefx/5.0/lastindexof-improved-handing-of-empty-values.md
```diff
@@ -0,0 +1,59 @@
+### LastIndexOf has improved handling of empty search strings
```
Youssef1313, line 1: nit: File name has a typo (handing vs handling). It's not very important to fix as it doesn't appear in the URL.

includes/core-changes/corefx/5.0/lastindexof-improved-handing-of-empty-values.md
```diff
@@ -0,0 +1,59 @@
...
+
+There is no compatibility switch to restore the old behavior.
+
+#### Category
+
+Core .NET libraries
+
+#### Affected APIs
+
+- <xref:System.String.LastIndexOf%2A?displayProperty=fullName>
```
Youssef1313, line 47: This means all overloads? but I don't think it's correct. For example, the overload that takes a `char` isn't affected?

includes/core-changes/corefx/5.0/lastindexof-improved-handing-of-empty-values.md
```diff
@@ -0,0 +1,59 @@
...
+Starting with .NET 5.0, these APIs return the correct value for `LastIndexOf`.
+
+```csharp
+Console.WriteLine("Hello".LastIndexOf("")); // prints '5' (correct)
+
+ReadOnlySpan<char> span = "Hello";
+Console.WriteLine(span.LastIndexOf("")); // prints '5' (correct)
+```
+
+In these examples, `5` is the correct answer because `"Hello".Substring(5)` and `"Hello".AsSpan().Slice(5)` both produce an empty string, which is trivially equal to the empty substring that is sought.
```
Youssef1313, line 25: The API reference needs to reflect this.

includes/core-changes/corefx/5.0/lastindexof-improved-handing-of-empty-values.md
```diff
@@ -0,0 +1,59 @@
...
+
+This change was part of an overall bug fixing effort around string handling for .NET 5. It also helps unify our behavior between Windows and non-Windows platforms. For more information, see [dotnet/runtime#13383](https://github.com/dotnet/runtime/issues/13383) and [dotnet/runtime##13382](https://github.com/dotnet/runtime/issues/13382).
+
+#### Version introduced
+
+5.0
+
+#### Recommended action
+
+You don't need to take any action. The .NET 5.0 runtime provides the new behaviors automatically.
```
Youssef1313, line 37: I don't think this is correct.

---

```csharp
private static void M(string s, string s2)
{
    var index = s.LastIndexOf(s2);
    if (index > -1) Console.WriteLine(s[index]);
}
```

The above code was valid before this breaking change.

But with this breaking change, the developer should check the against s.Length or check that s is not empty before trying to access the index.

See #44376

GrabYourPitchforks: This code was never valid. See my response at https://github.com/dotnet/runtime/issues/44376.

includes/core-changes/corefx/5.0/lastindexof-improved-handing-of-empty-values.md
GrabYourPitchforks: What do you mean by this comment? True, the sentence I included in the breaking change notification is a _drastic_ simplification of how `IndexOf` and `LastIndexOf` work, and there are some counterexamples. But as a high-level overview it's a useful way to think about the API.

See https://github.com/dotnet/runtime/blob/a057f27e860c1a59aa12e7dd9be028fba82b037f/src/libraries/System.Private.CoreLib/src/System/String.Searching.cs#L223 if you're curious as to the real logic that underlies these APIs.

GrabYourPitchforks (approved):

includes/core-changes/corefx/5.0/lastindexof-improved-handing-of-empty-values.md
GrabYourPitchforks: Depends on the overload. I don't think it's worth discussing this kind of nuance in the doc, though.

includes/core-changes/corefx/5.0/lastindexof-improved-handing-of-empty-values.md
Youssef1313: @GrabYourPitchforks I meant [this](https://docs.microsoft.com/en-us/dotnet/api/system.string.lastindexof?view=net-5.0#System_String_LastIndexOf_System_String_System_Int32_System_Int32_), and [this](https://docs.microsoft.com/en-us/dotnet/api/system.string.lastindexof?view=net-5.0#System_String_LastIndexOf_System_String_), etc..

Currently it says that the method returns the following:

> The zero-based starting index position of value if that string is found, or -1 if it is not. **If value is Empty, the return value is the last index position in this instance.**

Note the bolded part isn't correct (similar wording is there for different overloads).

includes/core-changes/corefx/5.0/lastindexof-improved-handing-of-empty-values.md
Youssef1313: @GrabYourPitchforks I still think this may need developer action in some cases, see a possible example from dotnet/aspnetcore in https://github.com/dotnet/runtime/issues/44376#issuecomment-723549583.

includes/core-changes/corefx/5.0/lastindexof-improved-handing-of-empty-values.md
Youssef1313: @GrabYourPitchforks I think at least the commented section needs to mention the specific overloads, as it's used by API compat tool.

includes/core-changes/corefx/5.0/lastindexof-improved-handing-of-empty-values.md
GrabYourPitchforks: Ah, you're referring to the remarks at https://docs.microsoft.com/dotnet/api/system.string.lastindexof. I don't think those files are in this repo, so they'd not be included in this PR.

Youssef1313: Yep, they belong to dotnet-api-docs repo.

BillWagner (approved):
This LGTM @gewarren 

You can decide if any changes are warranted from the conversations.

includes/core-changes/corefx/5.0/lastindexof-improved-handing-of-empty-values.md
gewarren: I'm going to leave this as-is. The API compat tool will at least not miss any occurrences by specifying all overloads. If you know exactly which overloads are affected and are up to it, feel free to create a new PR.

includes/core-changes/corefx/5.0/lastindexof-improved-handing-of-empty-values.md
gewarren: I'll update the API docs now.

PR #21380, opened 2020-11-07
Add interfaces to the list

AYS17:
## Summary

Describe your changes here.


BillWagner (approved):
Thanks for adding this @amadosoft 

This is a correct addition now that interfaces can have default implementations. I'll :shipit: now.

PR #21420, opened 2020-11-09
Add suppress warning section to syslib pages

gewarren:


GrabYourPitchforks:
Should this instead be an include file? Might get a bit wordy to include this snippet as more files are added. Would we need to update the pages which don't include workarounds so that they have some linkable text?

GrabYourPitchforks (approved):

GrabYourPitchforks:
Keeping the copy+paste text for articles without a workaround section seems fine. If we update the articles to include workaround we can remove the custom text and instead use the shared include.

PR #21457, opened 2020-11-10
Upgrade all "Microsoft Extensions" bits to .NET 5

IEvangelist:
## Summary

Upgrade all "Microsoft Extensions" bits to .NET 5. Contributes to #21458


BillWagner (approved):
Woo hoo!  Let's :shipit:!

🎆

PR #21490, opened 2020-11-11
Add links to one-file binaries for diagnostics CLI tools

chrisnas:
## Summary

The direct download links are available in each tool documentation page

Fixes https://github.com/dotnet/diagnostics/issues/344

### DIFF
--- a/docs/core/diagnostics/dotnet-counters.md
+++ b/docs/core/diagnostics/dotnet-counters.md
@@ -18,7 +18,7 @@ dotnet tool install --global dotnet-counters
 Starting from .NET 5, it is now possible to directly download a one-file executable for your operating system:
 
 - Win ([x86](https://aka.ms/dotnet-counters/win-x86) | [x64](https://aka.ms/dotnet-counters/win-x64) | [arm](https://aka.ms/dotnet-counters/win-arm) | [arm-x64](https://aka.ms/dotnet-counters/win-arm64))
-- OSX ([x64](https://aka.ms/dotnet-counters/osx-x64))
+- macOS ([x64](https://aka.ms/dotnet-counters/osx-x64))
 - Linux ([x64](https://aka.ms/dotnet-counters/linux-x64) | [arm](https://aka.ms/dotnet-counters/linux-arm) | [arm64](https://aka.ms/dotnet-counters/linux-arm64) | [musl-x64](https://aka.ms/dotnet-counters/linux-musl-x64) | [musl-arm64](https://aka.ms/dotnet-counters/linux-musl-arm64))
 
 ## Synopsis
--- a/docs/core/diagnostics/dotnet-counters.md
+++ b/docs/core/diagnostics/dotnet-counters.md
@@ -9,13 +9,15 @@ ms.date: 02/26/2020
 
 ## Install dotnet-counters
 
+### Option 1 - Install as dotnet global tool:
+
 To install the latest release version of the `dotnet-counters` [NuGet package](https://www.nuget.org/packages/dotnet-counters), use the [dotnet tool install](../tools/dotnet-tool-install.md) command:
 
 ```dotnetcli
 dotnet tool install --global dotnet-counters
 ```
 
-Starting from .NET 5, it is now possible to directly download a one-file executable for your operating system:
+### Option 2 - Direct download:
 
 - Win ([x86](https://aka.ms/dotnet-counters/win-x86) | [x64](https://aka.ms/dotnet-counters/win-x64) | [arm](https://aka.ms/dotnet-counters/win-arm) | [arm-x64](https://aka.ms/dotnet-counters/win-arm64))
 - macOS ([x64](https://aka.ms/dotnet-counters/osx-x64))
--- a/docs/core/diagnostics/dotnet-counters.md
+++ b/docs/core/diagnostics/dotnet-counters.md
@@ -17,9 +17,9 @@ dotnet tool install --global dotnet-counters
 
 Starting from .NET 5, it is now possible to directly download a one-file executable for your operating system:
 
- - Win ([x86](https://aka.ms/dotnet-counters/win-x86) | [x64](https://aka.ms/dotnet-counters/win-x64) | [arm](https://aka.ms/dotnet-counters/win-arm) | [arm-x64](https://aka.ms/dotnet-counters/win-arm64))
- - OSX ([x64](https://aka.ms/dotnet-counters/osx-x64))
- - Linux ([x64](https://aka.ms/dotnet-counters/linux-x64) | [arm](https://aka.ms/dotnet-counters/linux-arm) | [arm64](https://aka.ms/dotnet-counters/linux-arm64) | [musl-x64](https://aka.ms/dotnet-counters/linux-musl-x64) | [musl-arm64](https://aka.ms/dotnet-counters/linux-musl-arm64))
+- Win ([x86](https://aka.ms/dotnet-counters/win-x86) | [x64](https://aka.ms/dotnet-counters/win-x64) | [arm](https://aka.ms/dotnet-counters/win-arm) | [arm-x64](https://aka.ms/dotnet-counters/win-arm64))
+- OSX ([x64](https://aka.ms/dotnet-counters/osx-x64))
+- Linux ([x64](https://aka.ms/dotnet-counters/linux-x64) | [arm](https://aka.ms/dotnet-counters/linux-arm) | [arm64](https://aka.ms/dotnet-counters/linux-arm64) | [musl-x64](https://aka.ms/dotnet-counters/linux-musl-x64) | [musl-arm64](https://aka.ms/dotnet-counters/linux-musl-arm64))
 
 ## Synopsis
 
--- a/docs/core/diagnostics/dotnet-counters.md
+++ b/docs/core/diagnostics/dotnet-counters.md
@@ -16,10 +16,10 @@ dotnet tool install --global dotnet-counters
 ```
 
 Starting from .NET 5, it is now possible to directly download a one-file executable for your operating system:
+
  - Win ([x86](https://aka.ms/dotnet-counters/win-x86) | [x64](https://aka.ms/dotnet-counters/win-x64) | [arm](https://aka.ms/dotnet-counters/win-arm) | [arm-x64](https://aka.ms/dotnet-counters/win-arm64))
  - OSX ([x64](https://aka.ms/dotnet-counters/osx-x64))
  - Linux ([x64](https://aka.ms/dotnet-counters/linux-x64) | [arm](https://aka.ms/dotnet-counters/linux-arm) | [arm64](https://aka.ms/dotnet-counters/linux-arm64) | [musl-x64](https://aka.ms/dotnet-counters/linux-musl-x64) | [musl-arm64](https://aka.ms/dotnet-counters/linux-musl-arm64))
-
 
 ## Synopsis
 
--- a/docs/core/diagnostics/dotnet-counters.md
+++ b/docs/core/diagnostics/dotnet-counters.md
@@ -15,6 +15,12 @@ To install the latest release version of the `dotnet-counters` [NuGet package](h
 dotnet tool install --global dotnet-counters
 ```
 
+Starting from .NET 5, it is now possible to directly download a one-file executable for your operating system:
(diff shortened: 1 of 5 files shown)


docs/core/diagnostics/dotnet-counters.md
```diff
@@ -15,6 +15,12 @@ To install the latest release version of the `dotnet-counters` [NuGet package](h
 dotnet tool install --global dotnet-counters
 ```
 
+Starting from .NET 5, it is now possible to directly download a one-file executable for your operating system:
```
sdmaclea, line 18: I don't think we need to mention `Starting from .NET 5`.  Here or in each document.

```suggestion
It is also possible to directly download a single-file executable:
```

I am not sure of the placement here.  It is possible the links could be below. It is also possible we might want them all on a single reference page.

We will need to update for Apple Silicon later this year too.

docs/core/diagnostics/dotnet-counters.md
```diff
@@ -15,6 +15,12 @@ To install the latest release version of the `dotnet-counters` [NuGet package](h
...
+
+- Win ([x86](https://aka.ms/dotnet-counters/win-x86) | [x64](https://aka.ms/dotnet-counters/win-x64) | [arm](https://aka.ms/dotnet-counters/win-arm) | [arm-x64](https://aka.ms/dotnet-counters/win-arm64))
+- OSX ([x64](https://aka.ms/dotnet-counters/osx-x64))
```
sdmaclea, line 21: `OSX` is an obsolete name.  Apple calls this `Mac OS`  we should do the same.

adegeo: macOS

docs/core/diagnostics/dotnet-counters.md
hoyosjs: I hadn't added the documentation because adding it per tool has two issues:
- It doesn't call out that you need a 3.1 runtime or any other caveats like access to TEMP in the environment (I already hit this with a few customers). 
- It is not a preferred mechanism. There's no update, cleanup, or scope story around this. 

I was thinking of just adding a note to each tool saying: `In case you need to use this tool in an environment where the SDK is not available, please refer to ...`. I was talking about this with @sywhang, but he caveated the one thing I don't like of my approach: now users need to check two pages. Any feedback @dotnet/dotnet-diag @chrisnas?

sywhang: Thought about this for a while last night, and what I think we can do is to update each tool's page to include a link to the main reference page that has the caveats mentioned. Since single-file tools are not the preferred mechanisms for the reasons Juan mentioned above, it's probably better to just mention that it is possible on each tools' page, and user can go through one more step to download this less-preferred mechanism if they need it.

chrisnas: Are you sure that .NET 5+ versions of the tools will work with .NET Core 3 apps?

docs/core/diagnostics/dotnet-counters.md
chrisnas: Good point: I'm updating the text to macOS (this "maxOS" syntax is the official Apple syntax)

docs/core/diagnostics/dotnet-counters.md
sywhang: Yes. Tools are written in a way that will fall back if it tries to use a feature that is not compatible, but for all features that are compatible with 3.1, the tools will/should work.

chrisnas: > I don't think we need to mention `Starting from .NET 5`. Here or in each document.
> 
Are you sure that .NET 5+ versions of the tools will work with .NET Core 3/3.1 apps?

sywhang: > Are you sure that .NET 5+ versions of the tools will work with .NET Core 3/3.1 apps?

Yes. Tools are written in a way that will fall back if it tries to use a feature that is not compatible, but for all features that are compatible with 3.1, the tools will/should work.

chrisnas: > * It doesn't call out that you need a 3.1 runtime or any other caveats like access to TEMP in the environment (I already hit this with a few customers).

That's why I added the ".NET 5 bla bla" at the beginning. But a better wording could be found such as "Starting with .NET Core 3.1 <or any relevant version>, bla bla"

> * It is not a preferred mechanism. There's no update, cleanup, or scope story around this.
> I was thinking of just adding a note to each tool saying: `In case you need to use this tool in an environment where the SDK is not available, please refer to ...`. I was talking about this with @sywhang, but he caveated the one thing I don't like of my approach: now users need to check two pages. Any feedback @dotnet/dotnet-diag @chrisnas?

This will definitively be the ONLY version used in production: nobody will install the SDK in production JUST to download a tool. And I'm not talking about containers that must be as small as possible. 
You are right about the known limitations: they must be stated clearly.
I was thinking about adding a "Download" section instead of adding the text under the "Install" section
The download link should be in the tool page, not in another side page.
If you look at what is done for SysInternals, each tools has its download link and there is another page for the Suite where all tools could be downloaded at once. That could be a solution to keep the links for each tool + a common page with the limitations + all the links

docs/core/diagnostics/dotnet-counters.md
chrisnas: BTW, I've seen that "dotnet-dump` isn't supported on macOS." in the tool page but the OSX binary is available: is there any problem here?

sywhang: outdated docs :) cc @mikem8361

docs/core/diagnostics/dotnet-counters.md
chrisnas: > > Are you sure that .NET 5+ versions of the tools will work with .NET Core 3/3.1 apps?
> 
> Yes. Tools are written in a way that will fall back if it tries to use a feature that is not compatible, but for all features that are compatible with 3.1, the tools will/should work.

The text mentions .NET SDK 3.0; nothing about 3.1 runtime: this is why I added the mention to .NET 5

docs/core/diagnostics/dotnet-counters.md
chrisnas: > outdated docs :) cc @mikem8361

Do you want me to remove that line in the same PR?

docs/core/diagnostics/dotnet-counters.md
hoyosjs: The regular tools are `netcoreapp2.1` with a roll-forward policy. The self-contained apps are 3.1 - as the first version of single file was added there. The tools are agnostic of the runtime the target app is running, as long as there's a proper pipe it will work (so the target app needs 3.1+, and potentially 5.0 for some features like events at startup and features that rely on reverse communication mechanisms)

docs/core/diagnostics/dotnet-counters.md
sywhang: Sure -- one thing to mention would be that on macOS, dotnet-dump only works for target apps running .NET 5.0+.

Thanks @chrisnas!

mikem8361: dotnet-dump is supported on MacOS for 5.0 runtimes. ELF core dumps are generated by `collect` and read by `analyze`.  MachO dumps are not supported.

docs/core/diagnostics/dotnet-counters.md
noahfalk: I don't think we should frame the download options as being more or less 'preferred'. We want customers to use the best option for their situation. For many of them, the best option is the single-file download.



In terms of the presentation here, how does this sound?
 
## Install dotnet-counters

**Option 1** - Install as a [dotnet global tool](../tools/dotnet-tool-install.md):
```dotnetcli
dotnet tool install --global dotnet-counters
```

**Option 2** - Direct download:
- Win ([x86](https://aka.ms/dotnet-counters/win-x86) | [x64](https://aka.ms/dotnet-counters/win-x64) | [arm](https://aka.ms/dotnet-counters/win-arm) | [arm-x64](https://aka.ms/dotnet-counters/win-arm64))
- macOS ([x64](https://aka.ms/dotnet-counters/osx-x64))
- Linux ([x64](https://aka.ms/dotnet-counters/linux-x64) | [arm](https://aka.ms/dotnet-counters/linux-arm) | [arm64](https://aka.ms/dotnet-counters/linux-arm64) | [musl-x64](https://aka.ms/dotnet-counters/linux-musl-x64) | [musl-arm64](https://aka.ms/dotnet-counters/linux-musl-arm64))

For guidance on installation options and system requirements see \<some_link_to_a_page_with_details>

chrisnas:
I just updated the text to follow @noahfalk idea + adding a comment about dotnet-dump on macOS

sywhang (approved):
LGTM - Thanks @chrisnas!

docs/core/diagnostics/dotnet-counters.md
```diff
@@ -9,12 +9,20 @@ ms.date: 02/26/2020
...
 
 ```dotnetcli
 dotnet tool install --global dotnet-counters
 ```
 
+### Option 2 - Direct download:
+
+- Win ([x86](https://aka.ms/dotnet-counters/win-x86) | [x64](https://aka.ms/dotnet-counters/win-x64) | [arm](https://aka.ms/dotnet-counters/win-arm) | [arm-x64](https://aka.ms/dotnet-counters/win-arm64))
+- macOS ([x64](https://aka.ms/dotnet-counters/osx-x64))
+- Linux ([x64](https://aka.ms/dotnet-counters/linux-x64) | [arm](https://aka.ms/dotnet-counters/linux-arm) | [arm64](https://aka.ms/dotnet-counters/linux-arm64) | [musl-x64](https://aka.ms/dotnet-counters/linux-musl-x64) | [musl-arm64](https://aka.ms/dotnet-counters/linux-musl-arm64))
```
adegeo, line 24: There are a few things I think we should address in the design of these pages.

01. The **Description** section's content should be moved to the intro of the article after the **APPLIES TO** info, and then the **Description** header removed.
01. I think the way the install area is designed, it doesn't look good on the [published site](https://review.docs.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-sos?branch=pr-en-us-21490). I think we should remove the extra headers, they draw attention away from the main section **Install**.

    Summary of changes:
    - Change header from `Install dotnet-counters` to simply `Install`.
    - Add intro text to the `Install` header to help SEO.
    - Use bullet points instead of H3 for the two types of install.
    - Add some introduction to the binary download.
    - Change download from bullets to table.

    ````suggestion
    ## Install

    There are two ways to download and install `dotnet-counters`:

    - **dotnet global tool:**

      To install the latest release version of the `dotnet-counters` [NuGet package](https://www.nuget.org/packages/dotnet-counters), use the [dotnet tool install](../tools/dotnet-tool-install.md) command:

      ```dotnetcli
      dotnet tool install --global dotnet-counters
      ```

    - **Direct download:**

      Download the tool executable that matches your platform:

      | OS  | Platform |
      | --- | -------- |
      | Windows | [x86](https://aka.ms/dotnet-counters/win-x86) \| [x64](https://aka.ms/dotnet-counters/win-x64) \| [arm](https://aka.ms/dotnet-counters/win-arm) \| [arm-x64](https://aka.ms/dotnet-counters/win-arm64) |
      | macOS   | [x64](https://aka.ms/dotnet-counters/osx-x64) |
      | Linux   | [x64](https://aka.ms/dotnet-counters/linux-x64) \| [arm](https://aka.ms/dotnet-counters/linux-arm) \| [arm64](https://aka.ms/dotnet-counters/linux-arm64) \| [musl-x64](https://aka.ms/dotnet-counters/linux-musl-x64) \| [musl-arm64](https://aka.ms/dotnet-counters/linux-musl-arm64) |
    ````

    ##### Preview:

    image

chrisnas:
@adege: Feel free to polish the layout of the page after this PR is accepted. You have much better ways to ensure consistency with the MS Docs pages and tooling to preview the changes from .md files  :^)

adegeo:
@chrisnas I can go in and edit your sources in your fork for this PR if that's OK with you. 😄

adegeo (approved):
I'll follow up with another PR that does the fixes we discussed.

Thanks for your contribution!

chrisnas:
Sorry about the delay but I was talking at a conference today...
Thanks for doing the proposed formatting/updates into another PR!

adege:
*(WRONG RECIPIENT WARNING)*

Hi,

I'd like to kindly inform you guys that I'm receiving these messages, which
are obviously supposed to go to someone else called *@adegeo*
It looks like there's a typo in *@adegeo*'s email settings, which I believe
is the reason why I'm getting her/his messages.

Just FYI, cheers!

-adege

PR #21497, opened 2020-11-11
Fix Dockerfile, COPY command

facundo91:
## Summary

COPY should recieve two parameters. https://docs.docker.com/engine/reference/builder/#copy
Without it, Docker throws a build error saying: 'Dockerfile parse error line 14: COPY requires at least two arguments, but only one was provided. Destination could not be determined.'


sughosneo:
Hi @facundo91, Thank you for submitting this PR. 
Good catch 👍  Yes, destination (current directory context) is missing in line:14. Changes look good.

nishanil (approved):

PR #21523, opened 2020-11-12
Emphasize the zero-copy nature of System.Memory primitives

SingleAccretion:
## Summary

After reading the documentation on `Span`-related types, it may not be apparent to the reader what their purpose is and how do they actually function. There is a very real-world example: we had a conversion on Discord today with an experienced developer who had read the "Memory<T> and Span<T> usage guidelines" doc and came away from it with an impression that using `Span` involves copying the memory to the stack - clearly not the idea we want to give people.

The PR also clarified that `Memory<T>` can be used to wrap arbitrary memory, not just managed.


docs/standard/memory-and-spans/index.md
```diff
@@ -16,7 +16,7 @@ Starting with .NET Core 2.1, .NET includes a number of interrelated types that r
 
 - <xref:System.ReadOnlySpan%601?displayProperty=nameWithType>, an immutable version of the <xref:System.Span%601> structure.
 
-- <xref:System.Memory%601?displayProperty=nameWithType>, a contiguous region of memory that is allocated on the managed heap rather than the stack. A <xref:System.Memory%601> instance can be backed by an array of type `T` or a <xref:System.String>. Because it can be stored on the managed heap, <xref:System.Memory%601> has none of the limitations of <xref:System.Span%601>.
+- <xref:System.Memory%601?displayProperty=nameWithType>, a wrapper over a contiguous region of memory. <xref:System.Memory%601?displayProperty=nameWithType> can also be stored on the managed heap rather than only on the stack. A <xref:System.Memory%601> instance can be backed by an array of type `T` or a <xref:System.String>. Because it can be stored on the managed heap, <xref:System.Memory%601> has none of the limitations of <xref:System.Span%601>.
```
GrabYourPitchforks, line 19: I don't think the inserted sentence adds clarity. The sentence at the end of this paragraph already says that `Memory<T>` instances can be stored on the heap.

Maybe the inserted sentence needs to be reworded? What point needs clarified?

docs/standard/memory-and-spans/index.md
```diff
@@ -30,6 +30,8 @@ Starting with .NET Core 2.1, .NET includes a number of interrelated types that r
 
 - <xref:System.MemoryExtensions?displayProperty=nameWithType>, a collection of extension methods for converting strings, arrays, and array segments to <xref:System.Memory%601> blocks.
 
+<xref:System.Span%601?displayProperty=nameWithType>, <xref:System.Memory%601?displayProperty=nameWithType>, and their readonly counterparts are designed to allow the creation of algorithms that avoid copying memory or allocating on the managed heap more than necessary. Creating them (either via `Slice` or their constructors) does not involve manipulating the underlying buffers: only the relevant references and offsets, which represent the "view" of the wrapped memory, are updated.
```
GrabYourPitchforks, line 33: Perhaps "manipulating" should be changed to "duplicating"? If there's confusion about how `Span<T>` and `Memory<T>` work, I don't think the confusion stems from consumers believing that the contents of the buffers will be altered. The confusion probably stems from when copies are made vs. when copies aren't made.

docs/standard/memory-and-spans/memory-t-usage-guidelines.md
```diff
@@ -8,9 +8,9 @@ helpviewer_keywords:
...
 
-Both <xref:System.Span%601> and <xref:System.Memory%601> are buffers of structured data that can be used in pipelines. That is, they are designed so that some or all of the data can be efficiently passed to components in the pipeline, which can process them and optionally modify the buffer. Because <xref:System.Memory%601> and its related types can be accessed by multiple components or by multiple threads, it's important that developers follow some standard usage guidelines to produce robust code.
+Both <xref:System.Span%601> and <xref:System.Memory%601> are wrappers over buffers of structured data that can be used in pipelines. That is, they are designed so that some or all of the data can be efficiently (as none of it is copied when new instances of <xref:System.Span%601> or <xref:System.Memory%601> are created) passed to components in the pipeline, which can process them and optionally modify the buffer. Because <xref:System.Memory%601> and its related types can be accessed by multiple components or by multiple threads, it's important that developers follow some standard usage guidelines to produce robust code.
```
GrabYourPitchforks, line 13: The point about no-copy is important, but I don't think it belongs as an interjection here. Suggest removing the interjection and instead inserting a new sentence after "... optionally modify the buffer." That new sentence can describe that the original buffer reference is passed and that no copy of the buffer is made.

docs/standard/memory-and-spans/index.md
SingleAccretion: The initial point was to add something after the first sentence "a wrapper over a contiguous region of memory", as it replaced the confusing (In my opinion) "a contiguous region of memory that is allocated on the managed heap rather than the stack". After rereading it, I can see it is actually redundant. I will remove it.

docs/standard/memory-and-spans/index.md
SingleAccretion: Indeed, that's the point I hope to get across. I will replace it as per the suggestion.

docs/standard/memory-and-spans/memory-t-usage-guidelines.md
SingleAccretion: After a bit of thinking, I decided to cut this part of the change as it would be effectively duplicating the new paragraph being added to the "Overview" section.

IEvangelist (approved):

PR #21531, opened 2020-11-12
Null-forgiving and null-conditional interaction is fixed

pkulikov:
This is the continuation of #17988

That "unexpected behavior" was fixed in dotnet/roslyn#47712. As far as I understand from my tests, that compiler change also has an effect for the **C# 8.0** code (at least, the code in https://github.com/dotnet/docs/issues/17988#issuecomment-617348340 doesn't throw anymore for C# 8.0). So, the current note text is not correct anymore.

This PR suggests the most radical fix that is according to the latest compiler. There are some alternatives, though, if we want to keep documenting previous compiler versions.

From the referenced PR, I assume that that compiler change has been shipped with the compiler in Visual Studio 16.8. That means that alternative wording of the note would be:

> In Visual Studio 16.7 and earlier, the null-forgiving operator...

I don't like it, as Visual Studio is not the only available editor. Another option is to mention .NET Core SDK version.
@BillWagner what do you think?


BillWagner (approved):
Thanks @pkulikov 

I apologize for missing this earlier. It looks great, and I'll :shipit: now.

Issue #21272, opened 2020-10-29
F# 5 docs update


cartermp:
This tracks work to update the docs site with F# 5 content:

- [ ] "What's new in F# 5" article, with index update
- [ ] Update most uses of `sprintf` to use interpolated strings, particularly in "What is F#?" and the F# Tour
- [ ] Ensure coherency in the installation of "get started" pages
- [ ] Language reference: interpolated strings
- [ ] Language reference: nameof
- [ ] Language reference: open type declarations
- [ ] Language reference: applicative computation expressions
- [ ] Language reference: quotations retain constraints
- [ ] Language reference: slices behavior for negative/nonexistent ones, fixed-index 3d and 4d slices
- [ ] Language reference: default interface member consumption
- [ ] Language reference: nullable value types (new article + account for interop)
- [ ] F# Interactive: emphasize package references

dburriss:
@cartermp I would be interested in helping out. Are there guides on recommended structuring of docs or something like that?

cartermp:
@dburriss thanks for offering! No real recommended structure, but for things like language reference articles I will typically copy-paste another similar article and then tweak it from there. Usually the right structure falls out of that.

dburriss:
Sorry @cartermp I had some other writing due Monday and looks like mostly done now.

luisquintanilla:
@cartermp I can take this one

How to make this article more beginner-friendly? https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/slices

AngelMunoz:
After reviewing the FSI document in the session call I took the following notes

- Move fsharp-interactive into a Tools section for the F# docs
    - Requires updating [toc.yml](https://github.com/dotnet/docs/blob/master/docs/fsharp/toc.yml) and [index.yml](https://github.com/dotnet/docs/blob/master/docs/fsharp/index.yml)
    - Add (if possible, not required) any kind of tutorial or walkthrough like in the tutorials  section that uses the FSI

cc @luisquintanilla @dsyme @cartermp 
I got my fork running I'll try to get a PR in the following days, Luis let me know how should we coordinate this in case you have feedback on the original article

lqdev:
@AngelMunoz you can open a PR against this repo and I can take a look.

BillWagner:
closing this as completed.

PR #21542, opened 2020-11-13
Add nameof and interpolated strings articles for F#

cartermp:
https://github.com/dotnet/docs/issues/21272


docs/fsharp/language-reference/interpolated-strings.md
```diff
@@ -0,0 +1,99 @@
...
+
+printfn $"Name: %s{name}, Age: %d{age}"
+
+// Error: type mismatch
+printfn $"Name: %s{age}, Age: %d{name}"
+```
+
+In the previous example, the code mistakenly passes the `age` value where `name` should be, and vice/versa. Because the interpolated strings use format specifiers, this is a compile error instead of a subtle runtime bug.
+
+All format specifiers covered un [plaintext formatting](plaintext-formatting.md) are valid inside of an interpolated string.
```
Youssef1313, line 56:
```suggestion
All format specifiers covered in [plaintext formatting](plaintext-formatting.md) are valid inside of an interpolated string.
```

IEvangelist (approved):

PR #21555, opened 2020-11-13
Update how-to-use-spinlock-for-low-level-synchronization.md

maa:
## Summary

typo fix


BillWagner (approved):
Thanks for spotting this and fixing it @maa

PR #21571, opened 2020-11-15
Update dependency-injection.md

oneolddev:
I had to read the original text twice. It referred to a _MessageService_ class and the example above it did not include this class.

I believe it would be clearer just to refer to the _Worker_ class.

## Summary

Describe your changes here.


IEvangelist (approved):

PR #21592, opened 2020-11-16
What's new in .NET 5: rephrase entry about records

pkulikov:
- Records make it easy to create immutable types, but they are not forced to be immutable. Users still can define public fields and mutable properties or methods that mutate the record's state
- "...behave like value types" - this is too broad. What is the behaviour of value types exactly? One may say that "a variable of a value type contains the instance of the type, not a reference to it", for example.


IEvangelist (approved):

PR #21595, opened 2020-11-16
update to new msprod and mstechnology values.

BillWagner:
We updated our allow list for ms.prod and ms.technology.

This PR updates the values used in the file metadata section of the docfx.json file.


BillWagner:
This fixes almost all the warnings introduced by the new metadata value edits.

The four that remain are for values that were on the list to be requested but weren't added. (Vague language because don't know if I edited the spreadsheet wrong, or if the values were missed by mistake). Either way, the fix for these is the same: I'll file a new request for those values.

tdykstra (approved):
LGTM

PR #21615, opened 2020-11-17
Update CA5379 doc

### LINKED ISSUE #4110: False-positive? CA5379: Do not use weak key derivation function algorithm
### Analyzer

**Diagnostic ID**: [CA5379](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca5379): `Do not use weak key derivation function algorithm`

### Analyzer source

**NuGet Package**: [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers)

**Version**: v3.3.0 (Latest)

### Describe the bug

CA5379 is triggered even though neither SHA1 nor MD5 is used anywhere. The algorithm is passed in via a property, like so:

```c#
internal class Pbkdf2PasswordHashAlgorithm : PasswordHashAlgorithm
{
    public HashAlgorithmName AlgorithmName { get; }

    public int HashSize { get; }

    public Pbkdf2PasswordHashAlgorithm(string algorithmId, HashAlgorithmName algorithmName, int hashSize) : base(algorithmId)
    {
        AlgorithmName = algorithmName;
        HashSize = hashSize;
    }

    public override byte[] CreateHash(byte[] password, byte[] salt, int iterations)
    {
        // CA5379 on this line:
        using var rfc2898 = new Rfc2898DeriveBytes(password, salt, iterations, AlgorithmName);
        return rfc2898.GetBytes(HashSize);
    }
```

### Expected behavior

Do not trigger a warning on the marked line.

### Actual behavior

Triggers the specified warning.

## Additional context

If this is by design (which I would find odd) then the rule documentation should at least be updated to specify that if the algorithm is not known at compile-time then the rule is also triggered. That said, I don't think this should trigger the rule...rather it should be triggered anywhere `HashAlgorithmName.SHA1` or `HashAlgorithmName.MD5` is referenced, or if a different constructor for `Rfc2898DeriveBytes` is used where the algorithm is implicitly selected to be `SHA1`, which is not the case here.

dotpaul:
## Summary

Rename rule title, note that the rule fires when it can't determine the algorithm, note .NET version may need to be upgraded

Fixes https://github.com/dotnet/roslyn-analyzers/issues/4110


docs/fundamentals/code-analysis/quality-rules/ca5379.md
```diff
@@ -19,7 +19,11 @@ f1_keywords:
 
 ## Cause
 
-Using <xref:System.Security.Cryptography.MD5?displayProperty=nameWithType> or <xref:System.Security.Cryptography.SHA1?displayProperty=nameWithType> when instantiating <xref:System.Security.Cryptography.Rfc2898DeriveBytes?displayProperty=nameWithType>.
+Using one of the following algorithms when instantiating <xref:System.Security.Cryptography.Rfc2898DeriveBytes?displayProperty=nameWithType>:
```
gewarren, line 22:
```suggestion
Use of one of the following algorithms when instantiating <xref:System.Security.Cryptography.Rfc2898DeriveBytes?displayProperty=nameWithType>:
```

docs/fundamentals/code-analysis/quality-rules/ca5379.md
```diff
@@ -19,7 +19,11 @@ f1_keywords:
...
+
+- <xref:System.Security.Cryptography.MD5?displayProperty=nameWithType>
+- <xref:System.Security.Cryptography.SHA1?displayProperty=nameWithType>
+- Or the rule can't determine the algorithm at compile time
```
gewarren, line 26:
```suggestion
- An algorithm that the rule can't determine at compile time
```

gewarren (approved):

dotpaul:
Thank you @gewarren !

PR #21624, opened 2020-11-18
Natural Language syntax Error

wuyuansushen:
made availably rather than made available


BillWagner (approved):
Thanks for spotting this and making the correction @wuyuansushen We appreciate it.

PR #21628, opened 2020-11-18
Clarify how `x is null` is evaluated

### LINKED ISSUE #21107: Clarify "x is null" behavior – is it like operator==, Equals(), ReferenceEquals() ?
I'm shamelessly copying the content of the closed PR here: #6855

> Please document the lovely new "x is null" syntax that was added to C# 7. I'm talking about this: Open VS 2017 then click menubar -> Tools -> Options -> Text Editor -> C# -> Code Style -> General -> "Prefer 'is null' over 'object.ReferenceEquals(..., null)'".
>
> C# now allows us to write:
if (x is null) {...}
instead of:
if (object.ReferenceEquals(x, null)) {...}
>
> Please document the difference between "x is null" versus "x == null", meaning the difference when the class contains an operator-overload for the "==" operator. Thanks!

I found that PR because I wanted to know exactly that: what does `x is null` do? I don't feel that question has been answered at all.
Does it call an overloaded `==` operator ?
Does it call `object.Equals()` ?
Does it call `object.ReferenceEquals()` ?

I'm guessing `ReferenceEquals` due to how the original PR is phrased, but the docs do not make that clear at all.
There is a sample that uses various ways to check for null and demonstrate that `object o = null` is indeed `null`, but don't clarify which of the 3 versions above is used to determine that.

BillWagner:
Fixes #21107

### DIFF
--- a/docs/csharp/language-reference/keywords/is.md
+++ b/docs/csharp/language-reference/keywords/is.md
@@ -99,6 +99,8 @@ The following example shows a comparison of `null` checks:
 
 [!code-csharp[is#11](../../../../samples/snippets/csharp/language-reference/keywords/is/is-const-pattern11.cs#11)]
 
+The expression `x is null` is computed differently for reference types and nullable value types. For nullable value types, it uses <xref:System.Nullable%601.HasValue?displayProperty=nameWithType>. For reference types, it uses `x == null`.
+
 ### var pattern
 
 A pattern match with the `var` pattern always succeeds. Its syntax is:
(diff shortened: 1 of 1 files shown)


gewarren (approved):
LGTM.

docs/csharp/language-reference/keywords/is.md
```diff
@@ -99,6 +94,8 @@ The following example shows a comparison of `null` checks:
 
 [!code-csharp[is#11](../../../../samples/snippets/csharp/language-reference/keywords/is/is-const-pattern11.cs#11)]
 
+The expression `x is null` is computed differently for reference types and nullable value types. For nullable value types, it uses <xref:System.Nullable%601.HasValue?displayProperty=nameWithType>. For reference types, it uses `x == null`.
```
pkulikov, line 97: @BillWagner is that true? What if `==` is overloaded for a reference type?

pkulikov: @BillWagner please check #21630

PR #21640, opened 2020-11-18
Update blittable concept for clarity with UnmanagedCallersOnlyAttribute.

jkoritzinsky:
## Summary

Clarify the definition of a "blittable" type and split the "contains blittable contents" concept from the "blittable type" concept to improve clarity around the requirements of argument and return types for UnmanagedCallersOnlyAttribute-marked methods.

### DIFF
--- a/docs/standard/native-interop/best-practices.md
+++ b/docs/standard/native-interop/best-practices.md
@@ -82,15 +82,21 @@ GUIDs are usable directly in signatures. Many Windows APIs take `GUID&` type ali
 
 ## Blittable types
 
-Blittable types are types that have the same bit-level representation in managed and native code. As such they do not need to be converted to another format to be marshaled to and from native code, and as this improves performance they should be preferred.
+Blittable types are types that have the same bit-level representation in managed and native code. As such they do not need to be converted to another format to be marshaled to and from native code, and as this improves performance they should be preferred. Some types are not blittable but are known to contain blittable contents. These types have similar optimizations as blittable types when they are not contained in another type, but are not considered blittable when in fields of structs or for the purposes of [`UnmanagedCallersOnlyAttribute`](xref:System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute).
 
 **Blittable types:**
 
 - `byte`, `sbyte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `single`, `double`
-- non-nested one-dimensional arrays of blittable types (for example, `int[]`)
-- structs and classes with fixed layout that only have blittable value types for instance fields
+- structs with fixed layout that only have blittable value types for instance fields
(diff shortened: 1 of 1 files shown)


docs/standard/native-interop/best-practices.md
```diff
@@ -82,25 +82,35 @@ GUIDs are usable directly in signatures. Many Windows APIs take `GUID&` type ali
...
 
-- `char`, `string`
+- `char`
+
+**Types with SOMETIMES blittable contents:**
+
+- `string`
 
-When blittable types are passed by reference, they're simply pinned by the marshaller instead of being copied to an intermediate buffer. (Classes are inherently passed by reference, structs are passed by reference when used with `ref` or `out`.)
+When blittable types are passed by reference with `in`, `ref`, or `out`, or when types with blittable contents are passed by value, they're simply pinned by the marshaller instead of being copied to an intermediate buffer.
```
AaronRobinsonMSFT, line 113: This is only true when the desired `CharSet` matches the runtime encoding (i.e. `CharSet.Unicode`). Unless I am missing what is being stated here.

AaronRobinsonMSFT: I see you mentioned this below. Perhaps clarify here as well.

docs/standard/native-interop/best-practices.md
```diff
@@ -112,9 +122,9 @@ public struct UnicodeCharStruct
 }
 ```
 
-`string` is blittable if it isn't contained in another type and it's being passed as an argument that is marked with `[MarshalAs(UnmanagedType.LPWStr)]` or the `[DllImport]` has `CharSet = CharSet.Unicode` set.
+`string` contains blittable contents if it isn't contained in another type and it's being passed as an argument that is marked with `[MarshalAs(UnmanagedType.LPWStr)]` or the `[DllImport]` has `CharSet = CharSet.Unicode` set.
 
-You can see if a type is blittable by attempting to create a pinned `GCHandle`. If the type isn't a string or considered blittable, `GCHandle.Alloc` will throw an `ArgumentException`.
+You can see if a type is blittable or contains blittable contents by attempting to create a pinned `GCHandle`. If the type isn't a string or considered blittable, `GCHandle.Alloc` will throw an `ArgumentException`.
```
AaronRobinsonMSFT, line 127: I personally prefer the `Marshal.SizeOf()` method as it has less side effects. Unless that doesn't handle all the known cases.

jkoritzinsky: Marshal.SizeOf works for non-blittable value types as well so we can’t use it.

gewarren:
@jkoritzinsky Is this ready to go?

jkoritzinsky:
Yes it's ready to go.

gewarren (approved):

PR #21644, opened 2020-11-19
Convert ASP.NET Core 5.0 breaking changes

gewarren:
Convert ASP.NET Core 5.0 breaking changes to new format.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/serialization/5.0/jsonserializer-allows-reading-numbers-as-strings?branch=pr-en-us-21644).


docs/core/compatibility/aspnet-core/5.0/authentication-aad-packages-obsolete.md
```diff
@@ -1,36 +1,39 @@
...
 
 The APIs weren't marked as obsolete.
 
-#### New behavior
+## New behavior
 
 The APIs are marked as obsolete.
 
-#### Reason for change
+## Reason for change
```
Youssef1313, line 28: I already pointed to that in the previous PR, but just as a reminder, this breaks the apiport tool.

That will need to be fixed before merging `release-breaking-changes` into master.

gewarren: @Youssef1313 Thanks, I thought you were just saying that "Category" and "Affected APIs" were required to be H3s, but now I see it's all the headings. But in that case, I don't see how the tool is working even with the existing format where we use H4 headings for these headings. The code appears to be expecting H3s.

```csharp
if (currentLine.StartsWith("###", StringComparison.Ordinal))
{
    switch (currentLine.Substring("###".Length).Trim().ToLowerInvariant())
    {
        case "scope":
            state = ParseState.Scope;
            break;
        case "version introduced":
        case "version broken":
            state = ParseState.VersionBroken;
            break;
        ...
```

@Lxiamail Does the ApiPort tool even work with the existing format? Who should I talk to about updating the tool to work with this new format where each breaking change is in a separate file, and the headings to parse are H2s?

Youssef1313: Yeah I didn't notice it was expecting H3 but the actual was H4. So it looks like it's already broken for a while? not sure.

gewarren:
This looks good to me, so I'm going to go ahead and merge it so I can work on the last set of conversions.

PR #21645, opened 2020-11-19
Simplify sentence

wuyuansushen:
Simplify the description.

### DIFF
--- a/docs/standard/generics/index.md
+++ b/docs/standard/generics/index.md
@@ -29,7 +29,7 @@ ms.assetid: 2994d786-c5c7-4666-ab23-4c83129fe39c
 ---
 # Generics in .NET
 
-Generics let you tailor a method, class, structure, or interface to the precise data type it acts upon. For example, instead of using the <xref:System.Collections.Hashtable> class, which allows keys and values to be of any type, you can use the <xref:System.Collections.Generic.Dictionary%602> generic class and specify the type allowed for the key and the type allowed for the value. Among the benefits of generics are increased code reusability and type safety.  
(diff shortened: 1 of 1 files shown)


docs/standard/generics/index.md
```diff
@@ -29,7 +29,7 @@ ms.assetid: 2994d786-c5c7-4666-ab23-4c83129fe39c
 ---
 # Generics in .NET
 
-Generics let you tailor a method, class, structure, or interface to the precise data type it acts upon. For example, instead of using the <xref:System.Collections.Hashtable> class, which allows keys and values to be of any type, you can use the <xref:System.Collections.Generic.Dictionary%602> generic class and specify the type allowed for the key and the type allowed for the value. Among the benefits of generics are increased code reusability and type safety.  
+Generics let you tailor a method, class, structure, or interface to the precise data type it acts upon. For example, instead of using the <xref:System.Collections.Hashtable> class, which allows keys and values to be of any type, you can use the <xref:System.Collections.Generic.Dictionary%602> generic class and specify the type allowed for the key and the value. Among the benefits of generics are increased code reusability and type safety.
```
adegeo, line 32: Perhaps pluralizing `type` will help indicate there are two types specified?

```suggestion
Generics let you tailor a method, class, structure, or interface to the precise data type it acts upon. For example, instead of using the <xref:System.Collections.Hashtable> class, which allows keys and values to be of any type, you can use the <xref:System.Collections.Generic.Dictionary%602> generic class and specify the types allowed for the key and the value. Among the benefits of generics are increased code reusability and type safety.  
```

adegeo (commented):
Hello, thank you for submitting this update. I can see that the original author probably wanted to indicate that there are two types, one for the key and one for the value. I'm afraid this change could be interpreted as a single type for both key and value. What about my suggestion to pluralize `type`?

adegeo (approved):
Thanks a bunch for helping out!

PR #21676, opened 2020-11-20
Fixed a bug where the demo source code for PLINQ

### LINKED ISSUE #21602: Code example for ordering with AsSequential() has no effect?
The following code in the docs

```
static void SequentialDemo()
        {
            var orders = GetOrders();

            // call to WithExecutionMode added by me
            var query = (from ord in orders.AsParallel().WithExecutionMode(ParallelExecutionMode.ForceParallelism)
                         orderby ord.CustomerID
                         select new
                         {
                             O = ord.CustomerID,
                             Details = ord.OrderID,
                             Date = ord.OrderDate,
                             Shipped = ord.ShippedDate
                         }).AsUnordered().AsSequential().Take(5);
            
           // Added by me to print the elements
            foreach (var item in query)
            {
                Console.WriteLine(item);
            }
        }
```
Has no difference when the call to `AsSequential` is omitted. Why is that the case? Furthermore, each time we run this, it the order is always different, never the same. I don't think this example is at all a good one. If we order by a second colum, say ord.OrderId, we can see that the same order is produced regardless if we have `AsSequential` in there or not.

---
#### Document Details

* ID: 6de472a3-3017-e6e2-3b27-a62f8bd18959
* Version Independent ID: c7ddd569-d9b5-a59f-86a7-29ef8675540d
* Content: [How to: Combine Parallel and Sequential LINQ Queries](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/how-to-combine-parallel-and-sequential-linq-queries?redirectedfrom=MSDN)
* Content Source: [docs/standard/parallel-programming/how-to-combine-parallel-and-sequential-linq-queries.md](https://github.com/dotnet/docs/blob/master/docs/standard/parallel-programming/how-to-combine-parallel-and-sequential-linq-queries.md)
* Product: **dotnet**
* GitHub Login: @IEvangelist
* Microsoft Alias: **dapine**

IEvangelist:
## Summary

Fixed a bug where the demo source code was incorrectly ordering by CustomerID.

Describe your changes here.

Contributes to #21602

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_Misc/plinq/cs/create1.cs
+++ b/samples/snippets/csharp/VS_Snippets_Misc/plinq/cs/create1.cs
@@ -2,37 +2,43 @@
 using System;
 using System.Linq;
 
-public class Example
+class ExampleForAll
 {
-   public static void Main()
-   {
-      var source = Enumerable.Range(100, 20000);
+    public static void Main()
+    {
+        var source = Enumerable.Range(100, 20000);
 
-      // Result sequence might be out of order.
-      var parallelQuery = from num in source.AsParallel()
-                          where num % 10 == 0
-                          select num;
+        // Result sequence might be out of order.
+        var parallelQuery =
+            from num in source.AsParallel()
+            where num % 10 == 0
+            select num;
 
-      // Process result sequence in parallel
-      parallelQuery.ForAll((e) => DoSomething(e));
+        // Process result sequence in parallel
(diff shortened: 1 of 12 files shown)


BillWagner:
@IEvangelist It looks like adding a project means we have to do some work on the samples due to multiple `Main` methods and multiple `Example` modules.

BillWagner (approved):
This is a great update @IEvangelist 

You can :shipit: when ready.

PR #21687, opened 2020-11-21
fix: whitelisting -> allowlisting

nschonni:
Flagged with a few other instances in the build. The others look related to command-line flags and can't be updated, but this one should be OK


sughosneo:
Hi @nschonni , thank you for submitting this PR.  
`IP whitelisting` is the security functionality for any API gateway to control the access for a trusted client. So if we change the functionality name wouldn't that mean different ? Thoughts ?

nishanil (approved):

PR #21688, opened 2020-11-21
fix small typo in computation-expressions.md

hede:
## Summary

Removed two extraneous "is"s from the explanations of `return` and `return!`


cartermp (approved):
Thanks!

PR #21690, opened 2020-11-21
fsharp: avoid vanity-alignment

### LINKED ISSUE #21487: F# style guidelines: Recommend non-vanity indentation
The following is a request to change the F# style guidelines.

In [section Formatting white space, subsection Place parameters on a new line for long definitions](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#place-parameters-on-a-new-line-for-long-definitions), the style guide says:

> If you have a very long function definition, place the parameters on new lines and indent them to match the indentation level of the subsequent parameter.
> 
> ```f#
> module M =
>     let LongFunctionWithLotsOfParameters(aVeryLongParam: AVeryLongTypeThatYouNeedToUse)
>                                         (aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse)
>                                         (aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse)
>                                         =
>         // ... the body of the method follows
> ```

As mentioned in https://github.com/fsprojects/fantomas/issues/657, this kind of indentation (AFAIK called "vanity indentation") has several drawbacks:

* Important code moved far to the right
* There is little space left for the actual code
* Refactoring names breaks the alignment (even if using tooling like Fantomas to auto-format code, the tooling may need to be explicitly invoked on all changed files, and AFAIK renames that extend names may cause compiler warnings)
* There is an unseemly gap on the left-hand side (personal opinion)

I would format the above example like this:

```f#
module M =
    let LongFunctionWithLotsOfParameters
        (aVeryLongParam: AVeryLongTypeThatYouNeedToUse)
        (aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse)
        (aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse)
        =
        // ... the body of the method follows
```

This has none of the mentioned drawbacks.

As seen in the linked section, this also goes for the recommendations for members and constructors.

Another section which has the same problem is [Formatting copy-and-update record expressions](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-copy-and-update-record-expressions), which has this example:

> ```f#
> let newState =
>     {
>         state with
>             F = Some {
>                     F1 = 0
>                     F2 = ""
>                 }
>     }
> ```

Notice how the indentation of the `F` value is sensitive to the name of `F`:

```f#
let newState =
    {
        state with
            AGorillaHoldingTheBananaAndTheEntireJungle = Some {
                                                             F1 = 0
                                                             F2 = ""
                                                         }
    }
```

A better approach would be:

```f#
let newState =
    {
        state with
            AGorillaHoldingTheBananaAndTheEntireJungle =
                Some {
                    F1 = 0
                    F2 = ""                    
                }
    }
```

---

Conversely, in the section [Formatting discriminated unions](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-discriminated-unions) (unrelated to explicit whitespace recommendations), my preferred approach is demonstrated:

> ```f#
> let tree1 =
>     BinaryNode
>         (BinaryNode(BinaryValue 1, BinaryValue 2),
>          BinaryNode(BinaryValue 3, BinaryValue 4))
> ```

Furthermore, in the section [Formatting pipeline operators](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-pipeline-operators), the non-vanity indentation kind is also shown (though vanity indentation would have `|>` aligned three more spaces to the right):

> ```f#
> // Preferred approach
> let methods2 =
>     System.AppDomain.CurrentDomain.GetAssemblies()
>     |> List.ofArray
>     |> List.map (fun assm -> assm.GetTypes())
>     |> Array.concat
>     |> List.ofArray
>     |> List.map (fun t -> t.GetMethods())
>     |> Array.concat
> 
> // Not OK
> let methods2 = System.AppDomain.CurrentDomain.GetAssemblies()
>             |> List.ofArray
>             |> List.map (fun assm -> assm.GetTypes())
>             |> Array.concat
>             |> List.ofArray
>             |> List.map (fun t -> t.GetMethods())
>             |> Array.concat
> ```

In light of these counter-examples, one could argue that changing recommending non-vanity indentation would make the style guide more consistent.

---

To fix this issue, I suggest the following:

* In the examples above, use the non-vanity kind of indentation
* Add a separate sub-section under "Formatting white space" explicitly stating that the non-vanity kind of indentation is preferred over vanity indentation for the aforementioned reasons. This section could, among other examples, show mutable setters (https://github.com/fsprojects/fantomas/issues/659) and the reverse pipeline operator (#21459).

If this is acceptable, I can help create a PR if desired, but I have no personal need to put my name on this change and really just want it changed regardless of who does the work.

/cc @nojaf @knocte @lydell @pragmatrix @cartermp

knocte:
Fixes https://github.com/dotnet/docs/issues/21487

### DIFF
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -94,6 +94,80 @@ let myFun (a: decimal) b c = a + b + c
 let myFunBad (a:decimal)(b)c = a + b + c
 ```
 
+### Avoid name-sensitive alignments
+
+In general, seek to avoid indentation and alignment that is sensitive to naming:
+
+```fsharp
+// OK
+let myLongValueName =
+    someExpression
+    |> anotherExpression
+
+
+// Bad
+let myLongValueName = someExpression
+                      |> anotherExpression
+```
+
+This is sometimes called “vanity alignment” or “vanity indentation”. The primary reasons for avoiding this are:
+
+* Important code is moved far to the right
+* There is less width left for the actual code
+* Renaming can break the alignment
+
+Do the same for `do`/`do!` in order to keep the indentation consistent with `let`/`let!`. Here is an example using `do` in a class:
+
+```fsharp
+// OK
+type Foo () =
+    let foo =
+        fooBarBaz
+        |> loremIpsumDolorSitAmet
+        |> theQuickBrownFoxJumpedOverTheLazyDog
+    do
+        fooBarBaz
+        |> loremIpsumDolorSitAmet
+        |> theQuickBrownFoxJumpedOverTheLazyDog
+
+// Bad - notice the "do" expression is indented one space less than the `let` expression
+type Foo () =
+    let foo =
+        fooBarBaz
+        |> loremIpsumDolorSitAmet
+        |> theQuickBrownFoxJumpedOverTheLazyDog
+    do fooBarBaz
+       |> loremIpsumDolorSitAmet
+       |> theQuickBrownFoxJumpedOverTheLazyDog
+```
+
+Here is an example with `do!` using 2 spaces of indentation (because with `do!` there is coincidentally no difference between the approaches when using 4 spaces of indentation):
+
+```fsharp
+// OK
+async {
+  let! foo =
+    fooBarBaz
+    |> loremIpsumDolorSitAmet
+    |> theQuickBrownFoxJumpedOverTheLazyDog
+  do!
+    fooBarBaz
+    |> loremIpsumDolorSitAmet
+    |> theQuickBrownFoxJumpedOverTheLazyDog
+}
+
+// Bad - notice the "do!" expression is indented two spaces more than the `let!` expression
+async {
+  let! foo =
+    fooBarBaz
+    |> loremIpsumDolorSitAmet
+    |> theQuickBrownFoxJumpedOverTheLazyDog
+  do! fooBarBaz
+      |> loremIpsumDolorSitAmet
+      |> theQuickBrownFoxJumpedOverTheLazyDog
+}
+```
+
 ### Place parameters on a new line for long definitions
 
 If you have a long function definition, place the parameters on new lines and indent them to match the indentation level of the subsequent parameter.
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -852,10 +852,10 @@ This also applies to mutable setters:
 
 ```fsharp
 // Preferred approach
-ctx.Response.Headers.[HeaderNames.ContentType]
-    <- Constants.jsonApiMediaType |> StringValues
-ctx.Response.Headers.[HeaderNames.ContentLength]
-    <- bytes.Length |> string |> StringValues
+ctx.Response.Headers.[HeaderNames.ContentType] <-
+    Constants.jsonApiMediaType |> StringValues
+ctx.Response.Headers.[HeaderNames.ContentLength] <-
+    bytes.Length |> string |> StringValues
 
 // Not OK
 ctx.Response.Headers.[HeaderNames.ContentType] <- Constants.jsonApiMediaType
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -179,7 +179,7 @@ module M =
         (aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse)
         (aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse)
         =
-            // ... the body of the method follows
+        // ... the body of the method follows
 ```
 
 This also applies to members, constructors, and parameters using tuples:
@@ -192,7 +192,7 @@ type TM() =
             aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
             aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse
         ) =
-            // ... the body of the method
+        // ... the body of the method
 
 type TC
     (
@@ -200,7 +200,7 @@ type TC
         aSecondVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse,
         aThirdVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse
     ) =
-        // ... the body of the class follows
+    // ... the body of the class follows
 ```
 
 If the parameters are currified or there's an explicit return type annotation, it is preferable to place the `=` character on a new line:
@@ -213,7 +213,7 @@ type C() =
             aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
             aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse
         ) : AReturnType =
-            // ... the body of the method
+        // ... the body of the method
     member _.LongMethodWithLotsOfCurrifiedParams
         (aVeryLongParam: AVeryLongTypeThatYouNeedToUse)
         (aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse)
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -814,7 +814,7 @@ let function1 arg1 arg2 arg3 arg4 =
     arg3 + arg4
 ```
 
-### Formatting pipeline operators or mutable setters
+### Formatting pipeline operators or mutable assignments
 
 Pipeline `|>` operators should go underneath the expressions they operate on.
 
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -100,42 +100,52 @@ If you have a long function definition, place the parameters on new lines and in
 
 ```fsharp
 module M =
-    let LongFunctionWithLotsOfParameters(aVeryLongParam: AVeryLongTypeThatYouNeedToUse)
-                                        (aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse)
-                                        (aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse)
-                                        =
-        // ... the body of the method follows
+    let LongFunctionWithLotsOfParameters
+        (aVeryLongParam: AVeryLongTypeThatYouNeedToUse)
+        (aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse)
+        (aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse)
+        =
+            // ... the body of the method follows
 ```
 
 This also applies to members, constructors, and parameters using tuples:
 
 ```fsharp
 type TM() =
-    member _.LongMethodWithLotsOfParameters(aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
-                                            aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
-                                            aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse) =
-        // ... the body of the method
+    member _.LongMethodWithLotsOfParameters
+        (
+            aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
+            aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
+            aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse
+        ) =
+            // ... the body of the method
 
-type TC(aVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse,
+type TC
+    (
+        aVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse,
         aSecondVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse,
-        aThirdVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse) =
-    // ... the body of the class follows
+        aThirdVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse
+    ) =
+        // ... the body of the class follows
 ```
 
 If the parameters are currified or there's an explicit return type annotation, it is preferable to place the `=` character on a new line:
 
 ```fsharp
 type C() =
-    member _.LongMethodWithLotsOfParameters(aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
-                                            aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
-                                            aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse)
-                                            : AReturnType =
-        // ... the body of the method
-    member _.LongMethodWithLotsOfCurrifiedParams(aVeryLongParam: AVeryLongTypeThatYouNeedToUse)
-                                                (aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse)
-                                                (aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse)
-                                                =
-        // ... the body of the method
+    member _.LongMethodWithLotsOfParameters
+        (
+            aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
+            aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
+            aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse
+        ) : AReturnType =
+            // ... the body of the method
+    member _.LongMethodWithLotsOfCurrifiedParams
+        (aVeryLongParam: AVeryLongTypeThatYouNeedToUse)
+        (aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse)
+        (aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse)
+        =
+            // ... the body of the method
 ```
 
 This is a way to avoid too long lines (in case return type might have long name) and have less line-damage when adding parameters.
@@ -529,13 +539,14 @@ And as with the record guidance, you may want to dedicate separate lines for the
 
 ```fsharp
 type S = { F1: int; F2: string }
-type State = { F:  S option }
+type State = { Foo: S option }
 
-let state = { F = Some { F1 = 1; F2 = "Hello" } }
+let state = { Foo = Some { F1 = 1; F2 = "Hello" } }
 let newState =
     {
         state with
-            F = Some {
+            Foo =
+                Some {
                     F1 = 0
                     F2 = ""
                 }
@@ -803,7 +814,7 @@ let function1 arg1 arg2 arg3 arg4 =
     arg3 + arg4
 ```
 
-### Formatting pipeline operators
+### Formatting pipeline operators or mutable setters
 
 Pipeline `|>` operators should go underneath the expressions they operate on.
 
@@ -826,6 +837,32 @@ let methods2 = System.AppDomain.CurrentDomain.GetAssemblies()
             |> List.ofArray
             |> List.map (fun t -> t.GetMethods())
             |> Array.concat
+
+// Not OK either
+let methods2 = System.AppDomain.CurrentDomain.GetAssemblies()
+               |> List.ofArray
+               |> List.map (fun assm -> assm.GetTypes())
+               |> Array.concat
+               |> List.ofArray
+               |> List.map (fun t -> t.GetMethods())
+               |> Array.concat
+```
+
+This also applies to mutable setters:
+
+```fsharp
+// Preferred approach
+ctx.Response.Headers.[HeaderNames.ContentType]
+    <- Constants.jsonApiMediaType |> StringValues
+ctx.Response.Headers.[HeaderNames.ContentLength]
+    <- bytes.Length |> string |> StringValues
+
+// Not OK
+ctx.Response.Headers.[HeaderNames.ContentType] <- Constants.jsonApiMediaType
+                                                  |> StringValues
+ctx.Response.Headers.[HeaderNames.ContentLength] <- bytes.Length
+                                                    |> string
+                                                    |> StringValues
 ```
 
 ### Formatting modules
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -219,7 +219,7 @@ type C() =
         (aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse)
         (aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse)
         =
-            // ... the body of the method
+        // ... the body of the method
 ```
 
 This is a way to avoid too long lines (in case return type might have long name) and have less line-damage when adding parameters.


docs/fsharp/style-guide/formatting.md
```diff
@@ -100,42 +100,52 @@ If you have a long function definition, place the parameters on new lines and in
...
-        // ... the body of the method
-
-type TC(aVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse,
+    member _.LongMethodWithLotsOfParameters
+        (
+            aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
+            aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
+            aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse
+        ) =
+            // ... the body of the method
```
cmeeren, line 121: IMHO this looks less noisy, and seems to be more consistent with the formatting of curried parameters as well as with how Fantomas currently formats similar stuff:

```suggestion
    member _.LongMethodWithLotsOfParameters
        (aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
         aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
         aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse) =
            // ... the body of the method
```

Also note the placement of `=` on the last parameter line, in line with the following part of the docs:

> If the parameters are currified or there's an explicit return type annotation, it is preferable to place the `=` character on a new line

docs/fsharp/style-guide/formatting.md
```diff
@@ -100,42 +100,52 @@ If you have a long function definition, place the parameters on new lines and in
...
+
+type TC
+    (
+        aVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse,
         aSecondVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse,
-        aThirdVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse) =
-    // ... the body of the class follows
+        aThirdVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse
+    ) =
+        // ... the body of the class follows
```
cmeeren, line 129: Same reasoning as previous comment (can't add a GH suggestion due to the deleted line, is that deletion intentional?)

```f#
type TC =
    (aVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse,
     aSecondVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse,
     aThirdVeryLongCtorParam: AVeryLongTypeThatYouNeedToUse) =
```

docs/fsharp/style-guide/formatting.md
```diff
@@ -100,42 +100,52 @@ If you have a long function definition, place the parameters on new lines and in
...
-                                                (aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse)
-                                                =
-        // ... the body of the method
+    member _.LongMethodWithLotsOfParameters
+        (
+            aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
+            aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
+            aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse
+        ) : AReturnType =
+            // ... the body of the method
```
cmeeren, line 142: Same reasoning as previous comment

```suggestion
    member _.LongMethodWithLotsOfParameters
        (aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
         aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
         aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse)
        : AReturnType =
            // ... the body of the method
```

docs/fsharp/style-guide/formatting.md
```diff
@@ -826,6 +837,32 @@ let methods2 = System.AppDomain.CurrentDomain.GetAssemblies()
...
+```
+
+This also applies to mutable setters:
+
+```fsharp
+// Preferred approach
+ctx.Response.Headers.[HeaderNames.ContentType]
+    <- Constants.jsonApiMediaType |> StringValues
+ctx.Response.Headers.[HeaderNames.ContentLength]
+    <- bytes.Length |> string |> StringValues
```
cmeeren, line 858: I would much rather keep this consistent with e.g. `->` in pattern matching and keep the `<-` on the original line. I think it looks cleaner, and if the "body" spans multiple lines, you don't have to indent them an extra 3 spaces. Here's what I suggest:

```suggestion
ctx.Response.Headers.[HeaderNames.ContentType] <-
    Constants.jsonApiMediaType |> StringValues
ctx.Response.Headers.[HeaderNames.ContentLength] <-
    bytes.Length |> string |> StringValues
```

Here's what I meant by my last remark, which I don't want, and is a quasi-vanity-alignment (not sensitive to names, but still needs extra indentation due to the assignment operator, which is against the spirit of this PR):

```f#
ctx.Response.Headers.[HeaderNames.ContentType]
    <- some stuff
       |> that spans
       |> several lines
       |> and must be indented
ctx.Response.Headers.[HeaderNames.ContentLength]
    <- let x = 2
       x |> something |> somethingElse
```

docs/fsharp/style-guide/formatting.md
```diff
@@ -803,7 +814,7 @@ let function1 arg1 arg2 arg3 arg4 =
     arg3 + arg4
 ```
 
-### Formatting pipeline operators
+### Formatting pipeline operators or mutable setters
```
cmeeren, line 817:
```suggestion
### Formatting pipeline operators or assignments
```

cmeeren (changes requested):
Apart from my specific comments below, I am also missing an explicit mention somewhere that vanity alignment/indentation should not be used.

docs/fsharp/style-guide/formatting.md
cartermp: The downside to the approach you mention @cmeeren is that it doesn't distinguish the body of the member from the parameter list very well.

cmeeren: I agree, and personally I always intent the parameter list one level extra (one level more than the body). How about that?

```suggestion
    member _.LongMethodWithLotsOfParameters
            (aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
             aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
             aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse) =
        // ... the body of the method
```

knocte: I think double indentation would be a very special, hard to follow manually, rule. Another downside of your approach @cmeeren is that it has more chances of generating bigger diffs when adding parameters.

cmeeren: > I think double indentation would be a very special, hard to follow manually, rule.

Hard to remember, you mean? I have to admit that I'm coming at this from the angle of making Fantomas avoid vanity alignment and thus have a tooling mindset, but IMHO it's still much easier to do and maintain manually than the current vanity alignment recommendation, so I'm not sure I buy the argument.

> Another downside of your approach @cmeeren is that it has more chances of generating bigger diffs when adding parameters.

Are you talking about the fact that the parameter lines are not "equal" in that the last one has a parenthesis? If so, then 1) I don't see how that creates big diffs, and 2) certainly not bigger than existing similar recommendations like records:

```f#
type MyRecord =
    { A: int;
      B: string;
      C: byte [] }
```

---

In any case, I can live with your suggestion. But even that may have the problem @cartermp mentioned above. I'll make a comment on the relevant code to show what I mean.

docs/fsharp/style-guide/formatting.md
cmeeren: @knocte Here's what I mean (/cc @cartermp). If you have a sizable return type, even with your original suggestion, the body is not separated well from the parameter list.

docs/fsharp/style-guide/formatting.md
knocte: maybe `re-assignments` is better? to distinguish from normal constant assignments

docs/fsharp/style-guide/formatting.md
knocte: wrt diffs, I know it's not very common but you could add parameters to the front, and with the proposed approach there would be only n number of diff lines for n new parameters, while you would get n+1 with your approach.

> Hard to remember, you mean?

Yeah my aim is to always use fantomas as well, but I don't want fantomas to dictate a format that is hard to remember or apply manually, otherwise you will always depend on it. I mean, ideally fantomas should dictate a format that is easy to remember so that you can progressively learn it and come out naturally at some point (and your CI builds start to be green from the start, when fantomas is run in your PRs or feature branches).

docs/fsharp/style-guide/formatting.md
cmeeren: > constant assignments

I may very well be wrong, but AFAIK in F# `let x = 2` is a declaration(?), not an assignment, and only `<-` is an assignment.

knocte: the fact that in f# you cannot declare without assigning doesn't make it not be an assignment IMO, that's why I didn't prefer this word (some people call it "symbol binding" too), but anyway we're bikeshedding here...

docs/fsharp/style-guide/formatting.md
knocte: IMO you cannot really treat the `<-` operator in the same way as the `->` because they are completely different. In general, for the latter you find big amounts of code after it, so it's normal that most uses of it get separated with an EOL and add an indentation level. However, for the re-assignment operator I think it's a bad practice if you add a lot of statements because it kills readability.

docs/fsharp/style-guide/formatting.md
cmeeren: Yep, I'm comfortable with "mutable setters". Maybe "mutable assignments" is the best option?

```suggestion
### Formatting pipeline operators or mutable assignments
```

As you say, we're bike-shedding here.

docs/fsharp/style-guide/formatting.md
cmeeren: > In general, for the latter you find big amounts of code after it

I agree.

> so it's normal that most uses of it get separated with an EOL and add an indentation level

Your suggestion also has an EOL and add an indentation level. I just want to put the operator on the "left" side of the EOL.

> However, for the re-assignment operator I think it's a bad practice if you add a lot of statements because it kills readability.

I don't find that.

Everything in moderation, of course. My second example above with a separate `let` binding is pushing things, though I wouldn't rule it out outright. `<-` is similar to `=`, and it's no problem having complex expressions when declaring a value using `=`, and it may in many cases be preferable to declaring temporary values before the assignment. As for my first example, I certainly find the following perfectly readable and much prefer it to having `<-` force three extra spaces of indentation.

```f#
ctx.Response.Headers.[HeaderNames.ContentType] <-
    some stuff
    |> thatSpans
    |> severalLines
    |> andMustBeIndented
```

---

Apart from `<-` on the second line forcing extra indentation and thus seeming like a kind of "vanity indentation" whose removal is the point of this PR, I think my main issue is that formatting should be consistent with the initial `=` assignment, which of course is always on the first line. Even though it's valid syntax, I have never seen anyone write

```f#
let someValue
  = Constants.jsonApiMediaType
    |> StringValues
```

I think `<-` should be treated like `=` to be consistent with assignments and re-assignments:

```f#
let mutable someValue =
    Constants.jsonApiMediaType
    |> StringValues

someValue <- 
    Constants.jsonApiMediaType
    |> StringValues
```

cmeeren: I mentioned above that it's more difficult to do manually. You have to indent more manually in the IDE.

Here's my suggestion. The IDE indents automatically.

new

Here's the original PR suggestion. You have to indent manually both on the `<-` line and on the next line.

old

cartermp: I think I would prefer the suggestion to place `<-` on the same line largely due to the symmetry of `=` also almost always being on that same line.

docs/fsharp/style-guide/formatting.md
cartermp: Sure, but this circumstance feels incredibly rare though.

docs/fsharp/style-guide/formatting.md
cartermp: yeah, i'd prefer no double indentation here.

cmeeren: Let's keep the current PR suggestion, then. 🙂 Though note that `// ... the body of the method` is indented incorrectly and should be fixed.

docs/fsharp/style-guide/formatting.md
```diff
@@ -100,42 +100,52 @@ If you have a long function definition, place the parameters on new lines and in
...
-                                        (aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse)
-                                        (aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse)
-                                        =
-        // ... the body of the method follows
+    let LongFunctionWithLotsOfParameters
+        (aVeryLongParam: AVeryLongTypeThatYouNeedToUse)
+        (aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse)
+        (aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse)
+        =
+            // ... the body of the method follows
```
cmeeren, line 108:
```suggestion
        // ... the body of the method follows
```

docs/fsharp/style-guide/formatting.md
cmeeren, line 121:
```suggestion
        // ... the body of the method
```

docs/fsharp/style-guide/formatting.md
cmeeren, line 129: The body placeholder is indented incorrectly.

```suggestion
    // ... the body of the class follows
```

docs/fsharp/style-guide/formatting.md
cmeeren, line 142: The body placeholder is indented incorrectly.

```suggestion
        // ... the body of the method
```

docs/fsharp/style-guide/formatting.md
```diff
@@ -100,42 +100,52 @@ If you have a long function definition, place the parameters on new lines and in
...
+            aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
+            aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse
+        ) : AReturnType =
+            // ... the body of the method
+    member _.LongMethodWithLotsOfCurrifiedParams
+        (aVeryLongParam: AVeryLongTypeThatYouNeedToUse)
+        (aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse)
+        (aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse)
+        =
+            // ... the body of the method
```
cmeeren, line 148: The body placeholder is indented incorrectly.

```suggestion
        // ... the body of the method
```

docs/fsharp/style-guide/formatting.md
cmeeren: Though are mutable setters/assignments related to pipeline operators in terms of formatting? It doesn't seem that way to me, so perhaps it should be its own section.

docs/fsharp/style-guide/formatting.md
cmeeren: Please disregard this as per [this comment](https://github.com/dotnet/docs/pull/21690/files#r529190027)

docs/fsharp/style-guide/formatting.md
cmeeren: Let's keep the original suggestion then to be consistent with [this](https://github.com/dotnet/docs/pull/21690/files#r529190027)

docs/fsharp/style-guide/formatting.md
knocte: yeah @cmeeren convinced me on this one

cmeeren:
An open question: What about `do` and `do!` keywords? (And possibly others.)

In the spirit of this PR, perhaps there should be symmetry between `let`/`let!` and `do`/`do!`. I'm showing 2 spaces indentation below to make the difference more clear.

```f#
// Yes
async {
  let! foo = 
    fooBarBaz
    |> loremIpsumDolorSitAmet
    |> theBrownFoxJumpedOverTheLazyDog
  do!
    fooBarBaz
    |> loremIpsumDolorSitAmet
    |> theBrownFoxJumpedOverTheLazyDog
}

// No
async {
  let! foo =
    fooBarBaz
    |> loremIpsumDolorSitAmet
    |> theBrownFoxJumpedOverTheLazyDog
  do! fooBarBaz
      |> loremIpsumDolorSitAmet
      |> theBrownFoxJumpedOverTheLazyDog
}
```

Here's with 4 space indentation and `do` in a class.

```f#
// Yes
type Foo () =
    let foo = 
        fooBarBaz
        |> loremIpsumDolorSitAmet
        |> theBrownFoxJumpedOverTheLazyDog
    do
        fooBarBaz
        |> loremIpsumDolorSitAmet
        |> theBrownFoxJumpedOverTheLazyDog

// No
type Foo () =
    let foo =
        fooBarBaz
        |> loremIpsumDolorSitAmet
        |> theBrownFoxJumpedOverTheLazyDog
    do fooBarBaz
       |> loremIpsumDolorSitAmet
       |> theBrownFoxJumpedOverTheLazyDog
```

Note that if the `do` block contains multiple statements, then of course it should be indented in the `// Yes` fashion above anyway; I think everyone agrees on that. So using the `// Yes` variants above is more consistent with that, too.

The only downside is that with 4 spaces of indentation, the it may look a bit weird? I'm using 2 spaces anyway, so I'll leave others to decide that.

cartermp:
That is good guidance for sure. Would gladly accept a PR that added it in (or if it were added here).

cmeeren:
I'm more than happy to have this added here. @knocte freel free to do it; I don't need my name on any commits.

knocte:
> @knocte freel free to do it; I don't need my name on any commits.

Sorry I'm quite busy these days. You could create a PR on my fork and if I merge it, it would show up here :)

docs/fsharp/style-guide/formatting.md
knocte: Thing is, my opinion about the double-indentation on the parameters is not the same as the extra indentation for the body. The latter feels visually better because it adds one extra level after the parameters (so, parameters -> +1, body -> +2).

knocte: (while you were suggesting `parameters -> +2; body -> +1` before)

cmeeren: Are you saying that if the parameters are indented, the body should be doubly indented? If you intend that to be part of the PR, that should be made explicit.

Personally I think the body should be kept at one level of indentation. The `=` on a separate line (with or without type annotation) creates sufficient separation. @cartermp any opinions?

cmeeren:
PR created: https://github.com/knocte/dotnet-docs/pull/1

After this is merged, the only outstanding issue for me is the indentation of the body placeholder comments. We already have a discussion of that in another comment. I will ping both of you there.

docs/fsharp/style-guide/formatting.md
cmeeren: @knocte @cartermp The four last comments above relate to the double body indentation.

I currently stand by my view that double-indenting the body is unnecessary (and AFAIK without precedent elsewhere in the style guide), and that there is already sufficient separation between the parameters and the body.

@knocte If you agree, or if @cartermp agrees and decides on keeping bodies single-indented, it should be simple to just go through my suggestions here in this PR and commit them.

Alternatively, if you both agree that bodies should in fact be doubly indented in these cases, ignore my change suggestions, but please make it explicit somewhere that this should be done and why.

cartermp (commented):
Once this passes CI I will merge it in. Lots of good improvements here, thanks!

cartermp (approved):

cmeeren:
@cartermp @knocte

@nojaf asked for some clarification in https://github.com/fsprojects/fantomas/issues/657#issuecomment-747536973. I can provide a PR in case the docs need updating; I just want to discuss it first, and while this PR is merged, it seems a good place since it relates only to changes made here.

I have provided a summary at the bottom of how things look with my recommendations. Note that while this is my preference (IMHO it looks balanced and symmetrical), there are no religious feelings involved.

## Point 1: Member indentation for records with internal/private constructors

I feel fairly certain about this, but I'm putting it here just to be safe.

This would be my interpretation:

```f#
type MyType =
    internal
        { Field1: ..
          ..
          Field2: .. }

    static member Create ..
```

As opposed to:

```f#
type MyType =
    internal
        { Field1: ..
          ..
          FieldN: .. }

        static member Create ..
```

In the latter example, members are indented relative to the constructor, not the type declaration. Since the constructor is `internal`, it has one extra level of indentation. Indenting members an extra level too has no benefits, and breaks consistency with normal class members (or members in records whose constructor is not `internal`/`private`).

I personally see no need for a clarification on this point in the docs. Let me know if you disagree.

## Point 2: Indentation of parameters in long parameter lists

This wasn't asked by @nojaf but was something I saw while replying.

This PR was merged with the following example:

```f#
    member _.LongMethodWithLotsOfParameters
        (
            aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
            aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
            aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse
        ) : AReturnType =
        // ... the body of the method
```

The return type makes it look a bit lopsided, with an almost empty line at the top but not at the bottom. It becomes even more clear with one parameter:

```f#
    member _.LongMethodWithLotsOfParameters
        (
            aVeryLongParam: AVeryLongTypeThatYouNeedToUse
        ) : AReturnType =
        // ... the body of the method
```

Granted, it looks more symmetrical without a return type:

```f#
    member _.LongMethodWithLotsOfParameters
        (
            aVeryLongParam: AVeryLongTypeThatYouNeedToUse
        ) :
        // ... the body of the method
```

I can live with this, but I'd still like to suggest two ways that both address different aspects of this and could be written into the guide (both of them or either one) - just give the word and I can create a PR.

### 1. Getting rid of the blank line at the top

I suggest we format the parameter list like a record type definition, with 1 space between `(` and the first parameter and between the last parameter and `)` (which is at the same line as the last parameter instead of the line with `=`). I like this consistency with the record formatting, and it looks uncluttered and symmetrical.

```f#
    member _.LongMethodWithLotsOfParameters
        ( aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
          aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
          aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse )
        : AReturnType =
        // ... the body of the method

    member _.LongMethodWithLotsOfParameters
        ( aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
          aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
          aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse )
        =
        // ... the body of the method
```

For completeness, here is an "in-between" solution that keeps `)` on the same line as `=` (and the return type, if any). I like the above suggestion best, but find both better than the current suggestion.

```f#
    member _.LongMethodWithLotsOfParameters
        ( aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
          aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
          aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse
        ) : AReturnType =
        // ... the body of the method

    member _.LongMethodWithLotsOfParameters
        ( aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
          aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
          aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse
        ) =
        // ... the body of the method
```

### 2. Special-case the one-parameter case

Format it like this:

```f#
    member _.LongMethodWithLotsOfParameters
        (aVeryLongParam: AVeryLongTypeThatYouNeedToUse)
        : AReturnType =
        // ... the body of the method

    member _.LongMethodWithLotsOfParameters
        (aVeryLongParam: AVeryLongTypeThatYouNeedToUse)
        =
        // ... the body of the method
```

Note that this isn't critical if point 1 is implemented. With point 1 but without point 2, it looks like:

```f#
    member _.LongMethodWithLotsOfParameters
        ( aVeryLongParam: AVeryLongTypeThatYouNeedToUse )
        : AReturnType =
        // ... the body of the method

    member _.LongMethodWithLotsOfParameters
        ( aVeryLongParam: AVeryLongTypeThatYouNeedToUse )
        =
        // ... the body of the method
```

Though the extra space after `(` and before `)` is unnecessary, I don't find that it is problematic. If implementing point 1, consider if the special-casing of the 1-parameter case is worth the gain.

## Summary of my recommendation

```f#
type MyType =
    internal
        { Field1: int
          Field2: string
          Field3: bool }

    member _.LongMethodWithLotsOfParameters
        ( aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
          aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
          aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse )
        : AReturnType =
        // ... the body of the method

    member _.LongMethodWithLotsOfParameters
        ( aVeryLongParam: AVeryLongTypeThatYouNeedToUse,
          aSecondVeryLongParam: AVeryLongTypeThatYouNeedToUse,
          aThirdVeryLongParam: AVeryLongTypeThatYouNeedToUse )
        =
        // ... the body of the method

    member _.LongMethodWithLotsOfParameters
        (aVeryLongParam: AVeryLongTypeThatYouNeedToUse)  // Or with spaces by the parentheses
        : AReturnType =
        // ... the body of the method

    member _.LongMethodWithLotsOfParameters
        (aVeryLongParam: AVeryLongTypeThatYouNeedToUse)  // Or with spaces by the parentheses
        =
        // ... the body of the method
```

knocte:
> I personally see no need for a clarification on this point in the docs. Let me know if you disagree.

I agree with the first snippet yeah. Maybe we don't need an extra clarification but there's no harm in adding an extra code snippet.

> Granted, it looks more symmetrical without a return type:

Nit: I think the snippet you posted following this sentence has a small typo: should be `=` instead of `:` right?

> I can live with this

Then let's live with it :) 
I mean, TBH I'm not convinced about your proposals because the single-space additions are hard to specify, remember, and follow, IMO. I get your point that it maps better with the way records are styled, but:
1) You're only talking about the short way of styling records, there's a longer way (which is the one that I exclusively use myself FYI, and in fantomas I enable this via fsharp_multiline_block_brackets_on_same_column=true.
2) You're focusing too much on members that received non-currified parameters, which is not idiomatic in F#. In fact, if we focus more on currified parameters, we don't have this problem, and in fact for the 1-parameter case it would look like exactly one of the suggestions that you hint at the bottom of your comment:

```
    member _.LongMethodWithLotsOfParameters
        (aVeryLongParam: AVeryLongTypeThatYouNeedToUse)
        : AReturnType =
        // ... the body of the method

    member _.LongMethodWithLotsOfParameters
        (aVeryLongParam: AVeryLongTypeThatYouNeedToUse)
        =
        // ... the body of the method
```

AFAIU, this is how one-parameter members/lets should be formatted according to what the guide says at the current moment, because a one-parameter function signature is not a tuple.

cmeeren:
Fair enough. :)

cmeeren:
Just to be clear, @nojaf wants @cartermp to confirm before implementing this in Fantomas.

PR #21701, opened 2020-11-23
Fix wording

ebwinters:
## Summary

The original wording of this phrase is slightly confusing, and the answer to the question comes in the second part of the sentence instead of the beginning

### DIFF
--- a/docs/core/tools/dotnet-tool-restore.md
+++ b/docs/core/tools/dotnet-tool-restore.md
@@ -9,7 +9,7 @@ ms.date: 02/14/2020
 
 ## Name
 
-`dotnet tool restore` - Installs the .NET local tools onto your machine that are in scope for the current directory.
+`dotnet tool restore` - Installs the .NET local tools that are in scope for the current directory.
 
 ## Synopsis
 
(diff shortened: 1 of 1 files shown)


docs/core/tools/dotnet-tool-restore.md
```diff
@@ -9,7 +9,7 @@ ms.date: 02/14/2020
 
 ## Name
 
-`dotnet tool restore` - Installs on your machine the .NET local tools that are in scope for the current directory.
+`dotnet tool restore` - Installs the .NET local tools onto your machine that are in scope for the current directory.
```
adegeo, line 12: Does getting rid of machine completely make it even clearer? What do you think?
```suggestion
`dotnet tool restore` - Installs the .NET local tools that are in scope for the current directory.
```

adegeo (approved):
Thanks for helping out! I left one suggestion, let me know what you think.

docs/core/tools/dotnet-tool-restore.md
ebwinters: Makes more sense, yeah

PR #21727, opened 2020-11-25
Newtonsoft's ReferenceLoopHandling is not supported on 5.0

jozkee:
As https://github.com/dotnet/runtime/issues/40099#issuecomment-733693576 points out, we currently enumerate `ReferenceLoopHandling` as having a STJ equivalent option, which is currently not true; it will be added once https://github.com/dotnet/runtime/issues/40099 closes.


tdykstra (approved):
Thanks for catching and fixing this!

PR #21728, opened 2020-11-26
Fix PowerShell minimum version test

DXPetti:
Updated the PowerShell minimum version test to include the necessary `\` between HKLM: and SOFTWARE paths. Without, the test will fail regardless if the key is present or not due to path being incorrect

## Summary

Describe your changes here.


gewarren (approved):
@DXPetti Thank you!

PR #20476, opened 2020-09-05
Minor change in the output option description

keenthinker:
## Summary

Removed a dot character from the default output directory description for FDE.

PR #21749, opened 2020-11-29
Fix text that was causing translation error

tdykstra:
Fixes https://github.com/dotnet/docs.zh-cn/issues/662

### DIFF
--- a/docs/core/tutorials/debugging-with-visual-studio-code.md
+++ b/docs/core/tutorials/debugging-with-visual-studio-code.md
@@ -41,15 +41,7 @@ The breakpoint is located after a `Console.ReadLine` method call. The **Debug Co
 
 1. Open *.vscode/launch.json*.
 
-1. Change the `console` setting to `integratedTerminal`.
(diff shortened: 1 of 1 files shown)


IEvangelist (approved):

Youssef1313:
@tdykstra @IEvangelist Was this intended to be merged into live directly? The change doesn't exist in current `main`.

IEvangelist:
> @tdykstra @IEvangelist Was this intended to be merged into live directly? The change doesn't exist in current `main`.

Hi @Youssef1313 - that is a great observation. I'm not actually sure, @tdykstra do you know - if not we should sync this?

Youssef1313:
Pinging @tdykstra (there is https://github.com/dotnet/docs/pull/20476 as well).

tdykstra:
Yes this was accidentally merged directly to live-- @BillWagner I think the best thing to do here is revert the PR and redo the change in a PR to main -- do you have a better idea?

PR #21752, opened 2020-11-29
Use --counters option instead of the default argument option for dotnet-counters

sywhang:
dotnet-counters now prefers the --counters option over the counter_list argument to be specified. This PR changes references to the outdated option in the tutorial docs.


tdykstra (approved):
LGM

PR #21766, opened 2020-12-01
Trim redirects (round 1)

gewarren:
I wrote a tool to check link clicks on each source URL in the redirection file. This PR looks at the first 500 redirect entries and trims any entry that has zero page views for the last 90 days. Also attaching a tab-separated values file showing link clicks.

The diff doesn't seem to be any good, but any redirect source_path links with 0 hits were deleted from the redirection file (295 entries).

[clicks.txt](https://github.com/dotnet/docs/files/5620288/clicks.txt)


gewarren:
Closing in favor of #21772.

PR #21772, opened 2020-12-01
Trim redirects

gewarren:
I wrote a tool to check link clicks on each source URL in the redirection file. This PR removes any redirect entry that has zero page views for the last **90** days. Also attaching [an Excel file showing page views](https://github.com/dotnet/docs/files/5624797/page-views.xlsx) over the last 90 days for each redirect URL (sorted by page views descending).

The first commit reformats the file as-is (i.e. no deletions) such then when the data is serialized after removing the dead entries, it doesn't appear that the entire file has changed (since the serialization pretty printing differs to how the file was originally formatted). **To view the diff of which entries were removed, please look at the second and third commits.**

This PR removes 1518 redirects, which is roughly half of the redirects in the file (3152).


tdykstra (approved):
LGTM based on the first-50 commit where it's easy to see what was removed; the diff for the big commit gets confused so it's harder to read.

xtqqczze:
@gewarren What is the rationale for trimming these redirects? Why break the URLs if you avoid it?

gewarren:
> @gewarren What is the rationale for trimming these redirects? Why break the URLs if you avoid it?

Our redirection file (and other repos) got so big that GitHub couldn't display it anymore. If we never remove redirects, the file grows indefinitely. Obviously we don't want to remove any redirects that are actually getting used.

xtqqczze:
> > @gewarren What is the rationale for trimming these redirects? Why break the URLs if you avoid it?
> 
> Our redirection file (and other repos) got so big that GitHub couldn't display it anymore. If we never remove redirects, the file grows indefinitely. Obviously we don't want to remove any redirects that are actually getting used.

In my opinion, every 404 is a bad user experience and URLs, especially _documentation URLs_ should be **permanent**.

So it seems you have a technical problem with a large number of redirects, but a broken link is a _user experience_ issue.

Are there any other solutions that don't burden users with a technical issue of the repository?

PR #21779, opened 2020-12-01
Remove zero page view redirects

gewarren:
I wrote a tool to check link clicks on each source URL in the redirection file. This PR removes any redirect entry that has zero page views for the last **90** days. Also attaching an Excel file showing page views over the last 90 days for each redirect URL (sorted by page views descending).

[dotnet-90d-redirect-views.txt](https://github.com/dotnet/docs/files/5625610/dotnet-90d-redirect-views.txt)

The first commit reformats the file as-is (i.e. no deletions) such then when the data is serialized after removing the dead entries, it doesn't appear that the entire file has changed (since the serialization pretty printing differs to how the file was originally formatted). **To view the diff of which entries were removed, please look at the second and third commits.**

This PR removes 1521 redirects, which is roughly half of the redirects in the file (3152).

Contributes to #21720.


tdykstra (approved):
LGTM, again based on the first-100 commit, as the big one gets confused.

gewarren:
Thanks Tom!

PR #21780, opened 2020-12-01
Update ghal with new rules

adegeo:
- Moves the trigger logic from issue.opened to issue.labeled
  - Checks for /prod and /tech labels prior to processing.
  - Automatically adds not triaged when no org label is present
  - Automatically removes not triaged when an org label is present.
- Added .NET Core label for fundamentals area

### Requirements 
- [x] 1.16.4 of labeler AZFunction.

### DIFF
--- a/.ghal.rules.json
+++ b/.ghal.rules.json
@@ -1,172 +1,196 @@
 {
   "version": 3,
-  "configRevision": 2,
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @adegeo 

Let's :shipit: and test it.

PR #21810, opened 2020-12-03
C# value types reference: mention struct constraint

pkulikov:
With enum constraint available, it may be not obvious that the `struct` constraint allows both structure and enum types, but not only structure types.


BillWagner (approved):
Thanks again @pkulikov 

PR #21556, opened 2020-11-13
Add NET to preprocessor symbols

pavel-mikula-sonarsource:
## Summary

Add missing `NET` preprocessor symbol.

Fixes #21520

PR #21814, opened 2020-12-03
Added Missing Symbol for .NET X

Silvenga:
Adds the missing .NET preprocessor symbol for .NET 5, but versionless - `NET`.

(someone should double check this, this symbol is undocumented 😆)


gewarren:
@Silvenga Thanks for contributing. This is the same as #21556, which hasn't been merged yet.

PR #21828, opened 2020-12-04
Document IncludeAllContentForSelfExtract

NN---:


adegeo:
@LakshanF Can you shed some light on `IncludeAllContentForSelfExtract`? How does it affect the running of the app? I want to make sure this is accurate and also perhaps clarify some other parts of the article.

LakshanF:
cc @agocke as well

NN---:
Sure.

This flag makes single-file application working as it did in .NET Core 3.
It first extracts the files to the disk and then loads assemblies from the files in the disk.
By default without this flag, .NET 5.0 doesn't extract anything and uses everything in process memory.

This flag appears in few issues and some articles in the internet but doesn't appear in documentation.
It is essential for licenses which require dynamic loading if you want your application to use other license.

agocke:
LGTM -- we should also probably document `IncludeNativeLibrariesForSelfExtract`

NN---:
It is documented

https://docs.microsoft.com/en-us/dotnet/core/deploying/single-file

agocke:
Ah, in the "other considerations." In that case, let's not create a new section here and put those together. I certainly wouldn't want to encourage use of `IncludeAllContent` vs `IncludeNativeLibraries`

NN---:
Updated

docs/core/deploying/single-file.md
```diff
@@ -51,6 +51,8 @@ To fix these errors, _mscordbi_ needs to be copied next to the executable. _msco
 
 Single-file doesn't bundle native libraries by default. On Linux, we prelink the runtime into the bundle and only application native libraries are deployed to the same directory as the single-file app. On Windows, we prelink only the hosting code and both the runtime and application native libraries are deployed to the same directory as the single-file app. This is to ensure a good debugging experience, which requires native files to be excluded from the single file. There is an option to set a flag, `IncludeNativeLibrariesForSelfExtract`, to include native libraries in the single file bundle, but these files will be extracted to a temporary directory in the client machine when the single file application is run.
 
+Specifying `IncludeAllContentForSelfExtract` will extract all files before running the executable. This preserves original .NET Core single-file application behavior.
```
agocke, line 54:
```suggestion
Specifying `IncludeAllContentForSelfExtract` will extract all files before running the executable. This preserves the original .NET Core single-file deployment behavior.
```

agocke (approved):
LGTM

adegeo:
@NN--- Just noting that your original suggestion was deleted and replaced by a different one?

NN---:
Yes. I updated according to the review comment.

agocke:
@adegeo Can we merge this?

adegeo (approved):

PR #21844, opened 2020-12-06
Update nested-types.md

cenksener:
Delegate type can inside in Class type. Not only class, struct or interface. Delegates can be nested type.

We should update documentation.

### DIFF
--- a/docs/csharp/programming-guide/classes-and-structs/nested-types.md
+++ b/docs/csharp/programming-guide/classes-and-structs/nested-types.md
@@ -8,7 +8,7 @@ ms.assetid: f2e1b315-e3d1-48ce-977f-7bae0960ba99
 ---
 # Nested Types (C# Programming Guide)
 
-A type defined within a [class](../../language-reference/keywords/class.md), [struct](../../language-reference/builtin-types/struct.md), [delegate](../../language-reference/builtin-types/reference-types.md) or [interface](../../language-reference/keywords/interface.md) is called a nested type. For example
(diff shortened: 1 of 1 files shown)


docs/csharp/programming-guide/classes-and-structs/nested-types.md
```diff
@@ -8,7 +8,7 @@ ms.assetid: f2e1b315-e3d1-48ce-977f-7bae0960ba99
 ---
 # Nested Types (C# Programming Guide)
 
-A type defined within a [class](../../language-reference/keywords/class.md), [struct](../../language-reference/builtin-types/struct.md), or [interface](../../language-reference/keywords/interface.md) is called a nested type. For example
+A type defined within a [class](../../language-reference/keywords/class.md), [struct](../../language-reference/builtin-types/struct.md), [delegate](../../language-reference/builtin-types/reference-types.md) or [interface](../../language-reference/keywords/interface.md) is called a nested type. For example
```
BillWagner, line 11: We should add the anchor to this link:

```suggestion
A type defined within a [class](../../language-reference/keywords/class.md), [struct](../../language-reference/builtin-types/struct.md), [delegate](../../language-reference/builtin-types/reference-types.md#the-delegate-type) or [interface](../../language-reference/keywords/interface.md) is called a nested type. For example
```

BillWagner (changes requested):
Thanks for spotting this and fixing it @cenksener I like the change.

I made one suggestion so that the link includes the anchor to the `delegate` type. Once you accept that suggestion, this is ready to merge.

Thanks again!

BillWagner (approved):

PR #21862, opened 2020-12-07
Document Microsoft.Diagnostics.NETCore.Client

sywhang:
## Summary

This documents the Microsoft.Diagnostics.NETCore.Client library.

Part of https://github.com/dotnet/diagnostics/issues/515.


docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,459 @@
...
+            source.Process();
+        }
+        catch (Exception) {}
+    }
+}
+```
+
+### Trigger a CPU trace for given number of seconds
+
+This sample shows an example where we trigger an EventPipe session for certain period of time, with the default CLR trace keyword as well as the sample profiler, and read from the stream that gets created as a result and write the bytes out to a file. Essentially this is what `dotnet-trace` uses internally to write a trace file.
```
hoyosjs, line 121:
```suggestion
This sample shows how to trigger an EventPipe session for certain period of time with the default CLR trace keyword as well as the sample profiler. Afterward, it reads the output stream and writes the bytes out to a file. Essentially this is what [`dotnet-trace`](dotnet-trace.md) uses internally to write a trace file.
```

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,459 @@
...
+    {
+        Task copyTask = Task.Run(async () =>
+        {
+            using (FileStream fs = new FileStream(traceName, FileMode.Create, FileAccess.Write))
+            {
+                await traceSession.EventStream.CopyToAsync(fs);
+            }
+        });
+
+        copyTask.Wait(duration * 1000);
```
hoyosjs, line 149: Why is this `copyTask.Wait`? As long as the session is not gone, this will always block for the full duration. You sent the call to a background thread, and then effectively blocked this thread. Is this the best example that we can give here? A Task.WhenAny + Task.Delay may be better, and we should add something to observe the exceptions in `copyTask`.

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,459 @@
...
+```cs
+using Microsoft.Diagnostics.NETCore.Client;
+using System.Diagnostics;
+using System.Linq;
+
+public static void PrintProcessStatus()
+{
+    var processes = DiagnosticsClient.GetPublishedProcesses()
+        .Select(Process.GetProcessById)
+        .Where(process => process != null);
```
hoyosjs, line 168: When would process be null? Also, slightly better
```suggestion
        .Where(process => process is not null);
```

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,459 @@
...
+
+```csharp
+public IDictionary<string, string> Arguments { get; }
+```
+
+An `IDictionary` of key-value pair string representing optional arguments to be passed to EventSource representing the given `EventPipeProvider`.
+
+#### Remarks
+
+This class is immutable, as EventPipe does not allow a provider's configuration to be modified during an EventPipe session (as of .NET Core 3.1).
```
hoyosjs, line 384:
```suggestion
This class is immutable as EventPipe does not allow a provider's configuration to be modified during an EventPipe session as of .NET Core 3.1.
```

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,459 @@
...
+
+```csharp
+public class EventPipeSession : IDisposable
+{
+    public Stream EventStream { get; }
+    public void Stop();
+}
+```
+
+This class represents an ongoing EventPipe session that has been started. It is immutable and acts as a handle to an EventPipe session of the given runtime.
```
hoyosjs, line 396:
```suggestion
This class represents an ongoing EventPipe session. It is immutable and acts as a handle to an EventPipe session of the given runtime.
```

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,459 @@
...
+
+#### ServerNotAvailableException
+
+```csharp
+public class ServerNotAvailableException : DiagnosticsClientException
+```
+
+This may be thrown when the runtime is not available for diagnostics IPC commands, such as early during runtime startup before the runtime is ready for diagnostics commands, or when the runtime is shutting down.
+
+#### ServerErrorException
```
hoyosjs, line 453: Missing `UnsupportedCommandException`. Also, should we document `DiagnosticsClientException` generically as the exceptions that are thrown from the library for it's operations?

hoyosjs (approved):

docs/core/diagnostics/eventpipe.md
```diff
@@ -64,7 +64,7 @@ This is the easiest way to use EventPipe to trace your application. To learn mor
 
 ## Trace using environment variables
 
-The preferred mechanism for using EventPipe is to use `dotnet-trace` or the `Microsoft.Diagnostics.NETCore.Client` library.
+The preferred mechanism for using EventPipe is to use [`dotnet-trace`](dotnet-trace.md) or the [`Microsoft.Diagnostics.NETCore.Client`](microsoft-diagnostics-netcore-client.md) library.
 
 However, you can use the following environment variables to set up an EventPipe session on an app and have it write the trace directly to a file. To stop tracing, exit the application.
```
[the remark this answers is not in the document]
sywhang: Yup

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,459 @@
+# Microsoft.Diagnostics.NETCore.Client
+
+**This article applies to: ✔️** .NET Core 3.0 SDK and later versions
```
josalem, line 3: Should this message apply to the version of apps the library can _target_ or the version of apps the library can _run_ on?

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,459 @@
+# Microsoft.Diagnostics.NETCore.Client
```
adegeo, line 1: Missing the metadata yaml header.

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,459 @@
...
+
+**This article applies to: ✔️** .NET Core 3.0 SDK and later versions
+
+Microsoft.Diagnostics.NETCore.Client (also known as the Diagnostics Client library) is a managed library that lets you interact with .NET Core runtime (CoreCLR) for various diagnostics related tasks, such as tracing via [EventPipe](eventpipe.md), requesting a dump, or attaching an ICorProfiler. This library is the backing library behind many diagnostics tools such as [`dotnet-counters`](dotnet-counters.md), [`dotnet-trace`](dotnet-trace.md), [`dotnet-gcdump`](dotnet-gcdump.md), and [`dotnet-dump`](dotnet-dump.md). Using this library, you can write your own diagnostics tools customized for your particular scenario.
+
+## Install
+
+You can acquire [Microsoft.Diagnostics.NETCore.Client](https://www.nuget.org/packages/Microsoft.Diagnostics.NETCore.Client/) by adding a `PackageReference` to your project. The package is hosted on `NuGet.org`.
```
adegeo, line 9: Remove trailing space.
```suggestion
You can acquire [Microsoft.Diagnostics.NETCore.Client](https://www.nuget.org/packages/Microsoft.Diagnostics.NETCore.Client/) by adding a `PackageReference` to your project. The package is hosted on `NuGet.org`.
```

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,459 @@
...
+
+**This article applies to: ✔️** .NET Core 3.0 SDK and later versions
+
+Microsoft.Diagnostics.NETCore.Client (also known as the Diagnostics Client library) is a managed library that lets you interact with .NET Core runtime (CoreCLR) for various diagnostics related tasks, such as tracing via [EventPipe](eventpipe.md), requesting a dump, or attaching an ICorProfiler. This library is the backing library behind many diagnostics tools such as [`dotnet-counters`](dotnet-counters.md), [`dotnet-trace`](dotnet-trace.md), [`dotnet-gcdump`](dotnet-gcdump.md), and [`dotnet-dump`](dotnet-dump.md). Using this library, you can write your own diagnostics tools customized for your particular scenario.
+
+## Install
+
+You can acquire [Microsoft.Diagnostics.NETCore.Client](https://www.nuget.org/packages/Microsoft.Diagnostics.NETCore.Client/) by adding a `PackageReference` to your project. The package is hosted on `NuGet.org`. 
+
+## Use Microsoft.Diagnostics.NETCore.Client to write your own diagnostic tools
```
adegeo, line 11: This header is too long for the TOC area

```suggestion
## Write your own diagnostic tool
```

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,459 @@
...
+
+This snippet shows how to start an EventPipe session using the [.NET runtime provider](../../fundamentals/diagnostics/runtime-events.md) with the GC keyword at informational level and how to use the `EventPipeEventSource` class provided by the [TraceEvent library](https://www.nuget.org/packages/Microsoft.Diagnostics.Tracing.TraceEvent/) to parse the incoming events and print their names to the console in real time.
+
+```cs
+using Microsoft.Diagnostics.NETCore.Client;
+using Microsoft.Diagnostics.Tracing.Parsers;
+using System.Collections.Generic;
+
+public void PrintRuntimeGCEvents(int processId)
+{
```
adegeo, line 25: All of the code in this article is pretty big, it should be moved and referenced as a snippet. See the paragraph that starts with **or code snippets, create a subfolder called snippets inside the folder that contains your article** on https://docs.microsoft.com/en-us/contribute/dotnet/dotnet-contribute#make-a-contribution-to-net-docs

You can also look at this article as an example:

- Article: https://github.com/dotnet/docs/blob/master/docs/core/tutorials/library-with-visual-studio.md
- Snippets folder for article: https://github.com/dotnet/docs/tree/master/docs/core/tutorials/snippets/library-with-visual-studio

adegeo (dismissed):
A few things need to be addressed. Further @dotnet/docs is this the appropriate place for this article? It feels like a lot of this content should be at the root of the library namespace in the api docs repo and then a supporting article here. There is quite a bit of code reference here. Unsure, thought I would bring it up.

gewarren:
To me it seems like the API part should be automatically generated as part of the .NET API ref docs. Is there a reason we're not doing that? I think it would fall under the [platform extensions moniker](https://docs.microsoft.com/en-us/dotnet/api/?term=diagnostics&view=dotnet-plat-ext-5.0).

hoyosjs:
> To me it seems like the [API part](https://review.docs.microsoft.com/en-us/dotnet/core/diagnostics/microsoft-diagnostics-netcore-client?branch=pr-en-us-21862#api-description) should be automatically generated as part of the .NET API ref docs. Is there a reason we're not doing that? I think it would fall under the [platform extensions moniker](https://docs.microsoft.com/en-us/dotnet/api/?term=diagnostics&view=dotnet-plat-ext-5.0).

I do agree to some extent that it would be better if this was created as a part reference, part API docs. However, there's some problems to some extent. This gets created in a separate repo, and the versioning of the library and shipping cadence is independent of the .NET product, so the extensions might couple us in a way it's not desirable (we can't version separately) and is not quite true (this library can target several runtime versions simultaneously). 

- Check this in if possible as is. This would improve discoverability compared to what we have right now.
- Define a place to have this fall under docs. We should then have a way to publish the XML generated from the docstrings. Is there a place or someone I can sync up on this?
- Delete the API portion that was checked in, point to the docs as needed, and leave only the reference section.

sywhang:
I would ideally have this be part of the diagnostics documentation because the doc's main purpose is not to be an API reference but to serve as guide for anyone trying to create their custom diagnostic tool using the library. As Juan mentioned this library isn't part of the core BCL APIs or rest of the .NET stack and it ships on its own cadence. 

If this isn't an appropriate place for it I could theoretically split it up to API docs, tutorial, and a brief intro on it in the main page but I don't think that's a good experience for someone who wants to use this library since what they're seeking for is mostly:

1. code they can copy & paste and modify to fit their needs
2. reference to APIs itself in case they run into an issue

gewarren:
> Define a place to have this fall under docs. We should then have a way to publish the XML generated from the docstrings. Is there a place or someone I can sync up on this?

I'd check with @safern about this. I know we just published some extensions that don't have the exact same cadence either. For example: https://docs.microsoft.com/en-us/dotnet/api/system.binarydata?view=dotnet-plat-ext-5.0

hoyosjs:
Those only get updated in the case of a release. Basically, you expect a matching reference for that package in 5.0 or whatever release. This is completely outside the libraries shipped with the .NET product.

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -1,244 +1,64 @@
...
-        }
-        catch (Exception e)
-        {
-            Console.WriteLine("Error encountered while processing events");
-            Console.WriteLine(e.ToString());
-        }
-    }
-}
-```
+:::code language="csharp" source="snippets/Microsoft.Diagnostics.NETCore.Client/PrintRuntimeGCEvents.cs":::
```
adegeo, line 25: Put these in a `snippets/Microsoft.Diagnostics.NETCore.Client/csharp` folder. This way we can add other languages such as `vb`.

docs/core/diagnostics/snippets/Microsoft.Diagnostics.NETCore.Client/csharp/TraceProcessForDuration.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            Task copyTask = Task.Run(async () =>
+            {
+                using (FileStream fs = new FileStream(traceName, FileMode.Create, FileAccess.Write))
+                {
+                    await traceSession.EventStream.CopyToAsync(fs);
+                }
+            });
+
+            Task.WhenAny(copyTask, Task.Delay(duration * 1000));
```
hoyosjs, line 30: This needs an await, otherwise it's going to exit almost immediately. This method needs to be async at that point. Then you need to check which task returned. If it was copy, we ended the session somehow. Otherwise, we need to give time for the copy to finish by awaiting that one (probably with a timeout) after stopping the session, Otherwise the session will get disposed and the stream will get closed and then the async copy will give you corrupted traces.

The file should probably be created outside the session's using statement to handle failure cases.

safern:
> @gewarren I do agree to some extent that it would be better if this was created as a part reference, part API docs. However, there's some problems to some extent. This gets created in a separate repo, and the versioning of the library and shipping cadence is independent of the .NET product, so the extensions might couple us in a way it's not desirable (we can't version separately) and is not quite true (this library can target several runtime versions simultaneously).

@hoyosjs we currently support packages from multiple repos to contribute to the docs drop. The infrastructure to get the packages lives here: https://github.com/dotnet/api-catalog-infra -- please let me know if you have any questions regarding that repo. The end goal is to have a drop generated per commit on that repo, so even if your release is on a different cadence, since this uses darc, we would get a new drop whenever you ship a new package and darc knows about it.

hoyosjs:
@safern, this library doesn't fit under such libraries. The versioning is on the 0.1 range and needs a package reference. We also don't want to generate docs for builds on master until we decide to publish a package.

@gewarren @adegeo Is there any way to get this merged so users can start getting some documentation on the API? The API is small enough that reference + API docs fits in ~500 lines. The surface area of the library is small to the point that it will be very rare that they need using one function without needing another one in the documentation (basically, it's very rare you'll target just one API).

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,287 @@
...
+
+:::code language="csharp" source="snippets/Microsoft.Diagnostics.NETCore.Client/csharp/PrintEventsLive.cs":::
+
+### Attach an ICorProfiler profiler
+
+This sample shows how to attach an ICorProfiler to a process via profiler attach.
+
+:::code language="csharp" source="snippets/Microsoft.Diagnostics.NETCore.Client/csharp/ProfilerAttach.cs":::
+
+## API Description
```
gewarren, line 63: I'd suggest moving the API description to a separate file, and then adding the following metadata at the top:

```
ms.topic: reference
```

Also if you can standardize to the [API reference wording](https://github.com/dotnet/dotnet-api-docs/wiki), that would be good. For example, a get-only property description starts with "Gets ..." and a read-write property starts with "Gets or sets ...".

adegeo: I agree.

Where does the source for the library live? Is it /// commented?

docs/core/diagnostics/snippets/Microsoft.Diagnostics.NETCore.Client/csharp/TraceProcessForDuration.cs
IEvangelist, line 30:
```suggestion
            Task.WhenAny(copyTask, Task.Delay(TimeSpan.FromMilliseconds(duration * 1000)));
```

Use `TimeSpan` overload of `Task.Delay`

docs/core/diagnostics/snippets/Microsoft.Diagnostics.NETCore.Client/csharp/TriggerDumpOnCpuUsage.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        var providers = new List<EventPipeProvider>()
+        {
+            new EventPipeProvider(
+                "System.Runtime",
+                EventLevel.Informational,
+                (long)ClrTraceEventParser.Keywords.None,
+                new Dictionary<string, string>() {
+                    { "EventCounterIntervalSec", "1" }
+                }
```
IEvangelist, line 21:
```suggestion
                new Dictionary<string, string>
                {
                    ["EventCounterIntervalSec"] = "1"
                }
```

docs/core/diagnostics/snippets/Microsoft.Diagnostics.NETCore.Client/csharp/TriggerDumpOnCpuUsage.cs
```diff
@@ -0,0 +1,51 @@
...
+            )
+        };
+        var client = new DiagnosticsClient(processId);
+        using(var session = client.StartEventPipeSession(providers))
```
IEvangelist, line 25:
```suggestion
        using (var session = client.StartEventPipeSession(providers))
```

docs/core/diagnostics/snippets/Microsoft.Diagnostics.NETCore.Client/csharp/TriggerDumpOnCpuUsage.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            var source = new EventPipeEventSource(session.EventStream);
+            source.Dynamic.All += (TraceEvent obj) =>
+            {
+                if (obj.EventName.Equals("EventCounters"))
+                {
+                    IDictionary<string, object> payloadVal = (IDictionary<string, object>)(obj.PayloadValue(0));
+                    IDictionary<string, object> payloadFields = (IDictionary<string, object>)(payloadVal["Payload"]);
```
IEvangelist, line 33:
```suggestion
                    var payloadVal = (IDictionary<string, object>)(obj.PayloadValue(0));
                    var payloadFields = (IDictionary<string, object>)(payloadVal["Payload"]);
```

docs/core/diagnostics/snippets/Microsoft.Diagnostics.NETCore.Client/csharp/PrintRuntimeGCEvents.cs
```diff
@@ -0,0 +1,39 @@
...
+        };
+
+        DiagnosticsClient client = new DiagnosticsClient(processId);
+        using (EventPipeSession session = client.StartEventPipeSession(providers, false))
+        {
+            var source = new EventPipeEventSource(session.EventStream);
+
+            source.Clr.All += (TraceEvent obj) => {
+                Console.WriteLine(obj.ToString());
+            };
```
IEvangelist, line 26:
```suggestion
            source.Clr.All += (TraceEvent obj) => Console.WriteLine(obj.ToString());
```

docs/core/diagnostics/snippets/Microsoft.Diagnostics.NETCore.Client/csharp/PrintRuntimeGCEvents.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    public static void PrintRuntimeGCEvents(int processId)
+    {
+        List<EventPipeProvider> providers = new List<EventPipeProvider>()
+        {
+            new EventPipeProvider("Microsoft-Windows-DotNETRuntime",
+                EventLevel.Informational, (long)ClrTraceEventParser.Keywords.GC)
+        };
+
+        DiagnosticsClient client = new DiagnosticsClient(processId);
```
IEvangelist, line 19:
```suggestion
        var client = new DiagnosticsClient(processId);
```

docs/core/diagnostics/snippets/Microsoft.Diagnostics.NETCore.Client/csharp/PrintRuntimeGCEvents.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.Diagnostics.Tracing.Parsers;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Tracing;
+
+public class RuntimeGCEventsPrinter
+{
+    public static void PrintRuntimeGCEvents(int processId)
+    {
+        List<EventPipeProvider> providers = new List<EventPipeProvider>()
```
IEvangelist, line 13:
```suggestion
        var providers = new List<EventPipeProvider>()
```

Use `var` where the type is obvious.

docs/core/diagnostics/snippets/Microsoft.Diagnostics.NETCore.Client/csharp/PrintEventsLive.cs
```diff
@@ -0,0 +1,55 @@
...
+        using (var session = client.StartEventPipeSession(providers, false))
+        {
+
+            Task streamTask = Task.Run(() =>
+            {
+                var source = new EventPipeEventSource(session.EventStream);
+                source.Clr.All += (TraceEvent obj) =>
+                {
+                    Console.WriteLine(obj.EventName);
+                };
```
IEvangelist, line 29:
```suggestion
                source.Clr.All += (TraceEvent obj) => Console.WriteLine(obj.EventName);
```

docs/core/diagnostics/snippets/Microsoft.Diagnostics.NETCore.Client/csharp/PrintEventsLive.cs
```diff
@@ -0,0 +1,55 @@
...
+                    Console.WriteLine(e.ToString());
+                }
+            });
+
+            Task inputTask = Task.Run(() =>
+            {
+                Console.WriteLine("Press Enter to exit");
+                while (Console.ReadKey().Key != ConsoleKey.Enter)
+                {
+                    Task.Delay(100);
```
IEvangelist, line 47:
```suggestion
                    Task.Delay(TimeSpan.FromMilliseconds(100));
```

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,287 @@
...
+
+    public static bool operator !=(Provider left, Provider right);
+}
+```
+
+```csharp
+public EventPipeProvider(string name,
+                         EventLevel eventLevel,
+                         long keywords = 0,
+                         IDictionary<string, string> arguments = null)
```
IEvangelist, line 173:
```suggestion
public EventPipeProvider(
    string name,
    EventLevel eventLevel,
    long keywords = 0,
    IDictionary<string, string> arguments = null)
```

Formatting

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,287 @@
...
+* `profilerPath` : Path to the ICorProfiler dll to be attached.
+* `additionalData` : Optional additional data that can be passed to the runtime during profiler attach.
+
+```csharp
+public static IEnumerable<int> GetPublishedProcesses();
+```
+
+Get an `IEnumerable` of process IDs of all the active .NET processes that can be attached to.
+
+#### class EventPipeProvider
```
IEvangelist, line 138:
```suggestion
#### EventPipeProvider class
```

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,287 @@
...
+```csharp
+public DiagnosticsClient(int processId);
+```
+
+Creates a new instance of `DiagnosticsClient` for a compatible .NET process running with process ID of `processId`.
+
+`processID` : Process ID of the target application.
+
+```csharp
+public EventPipeSession StartEventPipeSession(IEnumerable<EventPipeProvider> providers, bool requestRundown=true, int circularBufferMB=256)
```
IEvangelist, line 89:
```suggestion
public EventPipeSession StartEventPipeSession(
    IEnumerable<EventPipeProvider> providers,
    bool requestRundown = true,
    int circularBufferMB = 256)
```

Formatting

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,287 @@
...
+
+This sample shows how to attach an ICorProfiler to a process via profiler attach.
+
+:::code language="csharp" source="snippets/Microsoft.Diagnostics.NETCore.Client/csharp/ProfilerAttach.cs":::
+
+## API Description
+
+This section describes the APIs of the library.
+
+### class DiagnosticsClient
```
IEvangelist, line 67:
```suggestion
### DiagnosticsClient class
```

docs/core/diagnostics/microsoft-diagnostics-netcore-client.md
```diff
@@ -0,0 +1,287 @@
...
+
+```cs
+public DiagnosticsClient
+{
+    public DiagnosticsClient(int processId);
+    public EventPipeSession StartEventPipeSession(IEnumerable<EventPipeProvider> providers, bool requestRundown=true, int circularBufferMB=256);
+    public void WriteDump(DumpType dumpType, string dumpPath, bool logDumpGeneration=false);
+    public void AttachProfiler(TimeSpan attachTimeout, Guid profilerGuid, string profilerPath, byte[] additionalData=null);
+    public static IEnumerable<int> GetPublishedProcesses();
+}
```
IEvangelist, line 77:
```suggestion
public DiagnosticsClient
{
    public DiagnosticsClient(int processId);
    
    public EventPipeSession StartEventPipeSession(
        IEnumerable<EventPipeProvider> providers,
        bool requestRundown = true,
        int circularBufferMB = 256);
    
    public void WriteDump(
        DumpType dumpType,
        string dumpPath,
        bool logDumpGeneration = false);
    
    public void AttachProfiler(
        TimeSpan attachTimeout,
        Guid profilerGuid,
        string profilerPath,
        byte[] additionalData = null);
    
    public static IEnumerable<int> GetPublishedProcesses();
}
```

gewarren:
@adegeo Can you rereview so we can get this merged?

PR #21863, opened 2020-12-07
Update .NET Core title suffix

gewarren:
I left the suffixes as-is for porting and migration, but open to feedback about that.


tdykstra (approved):
Leaving Core on migration and porting sounds good to me.

PR #21881, opened 2020-12-08
update tag position

hujiehaha:
in this chapter, the code snippet is missing a right curly brackets, so this pr change the tag position to the right line.
[https://docs.microsoft.com/en-us/dotnet/csharp/tour-of-csharp/program-building-blocks#method-overloading](https://docs.microsoft.com/en-us/dotnet/csharp/tour-of-csharp/program-building-blocks#method-overloading)


BillWagner (approved):
Thanks for spotting this and submitting the fix @hujiecs 

PR #21932, opened 2020-12-09
Updating BC notify workflow

timheuer:
## Summary
The bcnotify.yaml workflow exists to notify people of a breaking-change labeled item.  Unfortunately GitHub's api fires on 'any labeled' event.  

Adding the label condition check in the workflow itself so that this workflow only executes if the labeled event is the one with the `breaking-change` label itself.


IEvangelist (approved):

gewarren:
@timheuer How do people subscribe to this?

PR #21934, opened 2020-12-09
Add Dictionary DOS attack warning

### LINKED ISSUE #21724: Provide security guidance on JSON (de)serializing dictionaries with non-string keys
In .NET 5.0, we added support for (de)serializing dictionaries with non-string keys. Deserializing to a `Dictionary<TKey, TValue>` where `TKey` is typed as anything other than string could introduce a security vulnerability in the consuming application. We should mention this wherever we document this feature, e.g.: https://docs.microsoft.com/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to?pivots=dotnet-5-0#dictionary-with-non-string-key.

cc @tdykstra @Jozkee

tdykstra:
Fixes #21724

### DIFF
--- a/docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
+++ b/docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
@@ -4,7 +4,7 @@ description: "Learn how to migrate from Newtonsoft.Json to System.Text.Json. Inc
 author: tdykstra
 ms.author: tdykstra
 no-loc: [System.Text.Json, Newtonsoft.Json]
-ms.date: 11/30/2020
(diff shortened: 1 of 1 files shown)


IEvangelist (approved):

GrabYourPitchforks:
I don't know if the URL https://github.com/dotnet/runtime/issues/4761 is the best link to include in the documentation. That issue describes a technical change we might make to Dictionary in a future release, but it's not really a concise statement of "here's why this can be dangerous" that's easily digestible for a general audience. I imagine a typical developer won't know how that link is relevant to the issue.

That said, I can't think offhand of what a better link might be. Maybe it's a gap in our docs we need to fill in?

PR #21944, opened 2020-12-10
CA5399 & CA5400: Note that target framework upgrade may be necessary

dotpaul:
## Summary

CheckCertificateRevocationList isn't available in all .NET versions, so note that developers may need to upgrade.


gewarren (approved):
@dotpaul Thanks!

PR #21946, opened 2020-12-10
Linux installer updates

adegeo:
## Summary

Ping @omajid @RheaAyase @dleeapho @dagood 

This PR is to play around with other styles of presenting the Linux install pages. This is just a step in improving the Linux docs. I focused on redesigning Fedora and while we can continue to improve it after the holidays, do you have any quick thoughts on this new structure? I have a bunch of other minor fixes in there too for the other pages, but the major change is moving snap/manual instructions into their own pages.

- Improve fedora structure.
- Add manual/script page
- Add snap page
- Update Alpine to point to other pages (remove includes)
- Remove snap/manual includes from each distro page
- Delete snap/manual includes


docs/core/install/linux-scripted-manual.md
```diff
@@ -0,0 +1,151 @@
+---
+title: Manually install .NET on Linux - .NET
+description: Demonstrates how to install the .NET SDK and the .NET Runtime without a package manager on Linux. Use the install script or manually extract the binaries.
+author: adegeo
+ms.author: adegeo
+ms.date: 12/16/2020
```
BillWagner, line 6: Commenting here, but it applies to all the articles you've changed. I'd update the date to today:

```suggestion
ms.date: 1/06/2021
```

BillWagner (approved):
This LGTM @adegeo. It's a significant improvement over what we currently have in place.

You can :shipit: when ready.

PR #21948, opened 2020-12-10
Update known counters outdated description

sywhang:
## Summary

We deprecated the " / sec" or "/ min" descriptions from dotnet-counters because it was confusing to the user as interpolating the value can often be misleading. For instance gen 2 GC / min often is "60" when it's reported per second the value gets interpolated to be per min and makes it "60". This can be misleading because it makes it seem like the value means 60 gen 2 GCs happened in the last minute. 

This updates the descriptions for these "rate" counters in the official docs.

### DIFF
--- a/docs/core/diagnostics/includes/available-counters.md
+++ b/docs/core/diagnostics/includes/available-counters.md
@@ -9,7 +9,7 @@ The following counters are published as part of .NET runtime, and are maintained
 | Counter | Description |
 |--|--|
 | :::no-loc text="% Time in GC since last GC"::: (`time-in-gc`) | The percent of time in GC since the last GC |
-| :::no-loc text="Allocation Rate"::: (`alloc-rate`) | The rate of allocation in bytes |
+| :::no-loc text="Allocation Rate"::: (`alloc-rate`) | The number of bytes allocated per update interval |
 | :::no-loc text="CPU Usage"::: (`cpu-usage`) | The percent of CPU usage of the process |
 | :::no-loc text="Exception Count"::: (`exception-count`) | The number of exceptions that have occurred |
 | :::no-loc text="GC Heap Size"::: (`gc-heap-size`) | The number of bytes thought to be allocated based on <xref:System.GC.GetTotalMemory(System.Boolean)?displayProperty=nameWithType> |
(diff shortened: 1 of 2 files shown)


docs/core/diagnostics/includes/available-counters.md
```diff
@@ -62,12 +62,12 @@ The following counters are published as part of the [ASP.NET Core Kestrel web se
 | Counter | Description |
 |--|--|
 | :::no-loc text="Connection Queue Length"::: (`connection-queue-length`) | The current length of the connection queue |
-| :::no-loc text="Connection Rate"::: (`connections-per-second`) | The number of connections per second to the web server |
+| :::no-loc text="Connection Rate"::: (`connections-per-second`) | The number of connections per update interval to the web server |
```
tdykstra, line 65: The name `connections-per-second` is now out of sync with the description.

docs/core/diagnostics/includes/available-counters.md
```diff
@@ -62,12 +62,12 @@ The following counters are published as part of the [ASP.NET Core Kestrel web se
...
 | :::no-loc text="Current Connections"::: (`current-connections`) | The current number of active connections to the web server |
 | :::no-loc text="Current TLS Handshakes"::: (`current-tls-handshakes`) | The current number of TLS handshakes |
 | :::no-loc text="Current Upgraded Requests (WebSockets)"::: (`current-upgraded-requests`) | The current number of upgraded requests (WebSockets) |
 | :::no-loc text="Failed TLS Handshakes"::: (`failed-tls-handshakes`) | The total number of failed TLS handshakes |
 | :::no-loc text="Request Queue Length"::: (`request-queue-length`) | The current length of the request queue |
-| :::no-loc text="TLS Handshake Rate"::: (`tls-handshakes-per-second`) | The number of TLS handshakes per second |
+| :::no-loc text="TLS Handshake Rate"::: (`tls-handshakes-per-second`) | The number of TLS handshakes per update interval |
```
tdykstra, line 71: The name `tls-handshakes-per-second` is now out of sync with the description

docs/core/diagnostics/includes/available-counters.md
sywhang: Unfortunately, the names of counters are not documentation. They're implementation details. We could change it, but this will cause backward compat issues.

docs/core/diagnostics/includes/available-counters.md
sywhang: Same as above.

josalem (approved):
I think it is worth calling out the name discrepency between the counter and the actual unit of time used by the tool.

sywhang:
> I think it is worth calling out the name discrepency between the counter and the actual unit of time used by the tool.

Potentially. I need to ping the ASP.NET Core folks and ask if they have any plans to change this. I'll follow up accordingly.

PR #21951, opened 2020-12-11
Fix documentation ID format links

Evangelink:
## Summary

Fix documentation ID format links


gewarren (approved):
@Evangelink Thank you for updating these links.

PR #21959, opened 2020-12-11
Update version info for breaking change

### LINKED ISSUE #21952: Misleading page : change affects all versions of .NET, not just .NET 5
The page is misleading. It states that the change only affects .NET 5, but in fact it also affect any version of .NET framework if the SDK for .NET 5 has been installed (regardless of whether any other SDKs are also installed).

I have a .NET 4.5.2 solution with a few WPF and Winforms projects including one that was explicitly set to `<OutputType>Exe</OutputType>` to make sure the console was displayed. Upon updating Visual Studio (without any code or config change), it started to break (no console shown).

I would have expected the rule rewriting the `OutputType` to be guarded and check that .NET 5 is part of the targeted frameworks. Or this page to be found in a more reachable section of the documentation: it took me a while to find it because I wasn't expected a change in .NET 5 to affect .NET Framework.

**Steps to reproduce**

Create a new winforms app. Change the target framework to `net452`. Set the output type to `Exe` explicitly --> no console.

Related issue: https://github.com/dotnet/sdk/issues/13331

---
#### Document Details

* ID: d8aa9669-0b6a-0429-23da-ebd62c5e6b34
* Version Independent ID: 5471792d-993c-61ce-b472-1e11b00b7adc
* Content: [Breaking change: OutputType set to WinExe for WPF and WinForms apps - .NET](https://docs.microsoft.com/en-us/dotnet/core/compatibility/windows-forms/5.0/automatically-infer-winexe-output-type)
* Content Source: [docs/core/compatibility/windows-forms/5.0/automatically-infer-winexe-output-type.md](https://github.com/dotnet/docs/blob/master/docs/core/compatibility/windows-forms/5.0/automatically-infer-winexe-output-type.md)
* GitHub Login: @gewarren

gewarren:
Fixes #21952.

### DIFF
--- a/docs/core/compatibility/windows-forms/5.0/automatically-infer-winexe-output-type.md
+++ b/docs/core/compatibility/windows-forms/5.0/automatically-infer-winexe-output-type.md
@@ -9,7 +9,7 @@ ms.date: 09/18/2020
 
 ## Change description
 
-In previous versions of .NET, the value that's specified for `OutputType` in the project file is used. For example:
+In previous versions of the .NET SDK, the value that's specified for `OutputType` in the project file is used. For example:
 
 ```xml
 <PropertyGroup>
@@ -17,7 +17,7 @@ In previous versions of .NET, the value that's specified for `OutputType` in the
 </PropertyGroup>
 ```
 
-Starting in .NET 5.0, `OutputType` is automatically set to `WinExe` for WPF and Windows Forms apps. For example:
+Starting in the 5.0.1 version of the .NET SDK, `OutputType` is automatically set to `WinExe` for WPF and Windows Forms apps that target any framework version, including .NET Framework. For example:
 
 ```xml
 <PropertyGroup>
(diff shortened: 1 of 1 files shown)


tdykstra (approved):

dsplaisted:
@gewarren The .NET SDK follows a different versioning scheme than the .NET Runtime: https://docs.microsoft.com/en-us/dotnet/core/versions/

This was introduced in .NET SDK 5.0.100, not 5.0.1 (which doesn't exist).

gewarren:
@dsplaisted Thanks, I'll update it. So 5.0.100 is the first version of the SDK that shipped with .NET 5.0, not 5.0.000?

dsplaisted:
> @dsplaisted Thanks, I'll update it. So 5.0.100 is the first version of the SDK that shipped with .NET 5.0, not 5.0.000?

Yes, correct

PR #21974, opened 2020-12-13
Correction of a delegate related information

polatayd:
## Summary

"delegate" cannot be an outer type. A type cannot be defined "within" a delegate. Therefore, the statement "A type defined within a class, struct, delegate or interface is called a nested type." is not fully correct and "delegate" should be removed from this statement.


BillWagner (approved):
Thanks for spotting this and making the change @polatayd 

Issue #21927, opened 2020-12-09
Clarification for formatting of multi-line lambdas as the only argument


samarsha:
Fantomas recently changed its default formatting style (fsprojects/fantomas#1221) from:

```fsharp
xs
|> List.map (fun x ->
    // foo
    printfn "%s" x)
```

to

```fsharp
xs
|> List.map
    (fun x ->
        // foo
        printfn "%s" x)
```

@nojaf says this is to match the [Formatting function parameter application](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-function-parameter-application) section of the style guide, which uses an example of a call to `List.iter` without a piped second argument:

```fsharp
    List.iter
        (fun elem ->
            printfn $"%d{a + elem}")
        list1
```

In my opinion, it seems like the style guide is saying the lambda should start on the next line and be indented only if there is more than one argument applied to the function directly (not counting a pipe). There are examples of the lambda starting on the same line as the function name in other sections of the guide, but it's not stated explicitly.

Could the style guide clarify what the expected formatting is for a multi-line lambda as a single argument to a function? Thanks!

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 8c09996f-24ac-cf39-ef43-d7909cbc2b89
* Version Independent ID: 1efdd523-8552-6421-e149-74f76826ef8f
* Content: [F# code formatting guidelines](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-function-parameter-application)
* Content Source: [docs/fsharp/style-guide/formatting.md](https://github.com/dotnet/docs/blob/master/docs/fsharp/style-guide/formatting.md)
* Product: **dotnet-fsharp**
* GitHub Login: @cartermp
* Microsoft Alias: **phcart**

nojaf:
//cc @drhumlen

cartermp:
Thanks for the issue, it's an interesting one. I opened https://github.com/dotnet/docs/pull/21978 - care to give it a look?

hvester:
I think further clarification is needed for this. Is the following recommended default, valid alternative or is it discouraged?
```fsharp
let printListWithOffsetPiped a list1 =
    list1
    |> List.iter (fun elem ->
        printfn $"%d{a + elem}")
```
Currently the documentation says:
[...]

PR #21978, opened 2020-12-13
Clarify parameter application formatting

cartermp:
For https://github.com/dotnet/docs/issues/21927

### DIFF
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -832,7 +832,7 @@ with
 
 ## Formatting function parameter application
 
-In general, most function parameter application is done on the same line:
+In general, most arguments are provided on the same line:
 
 ```fsharp
 let x = sprintf "\t%s - %i\n\r" x.IngredientName x.Quantity
@@ -849,7 +849,7 @@ let printListWithOffsetPiped a list1 =
     |> List.iter (fun elem -> printfn $"%d{a + elem}")
 ```
 
-However, you may wish to apply parameters to a function on a new line, as a matter of readability or because the list of arguments or the argument names are too long. In that case, indent with one scope:
+However, you may wish to pass arguments to a function on a new line, as a matter of readability or because the list of arguments or the argument names are too long. In that case, indent with one scope:
 
 ```fsharp
 
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -582,10 +582,11 @@ type MyRecord =
 
 let foo a =
     a
-    |> Option.map (fun x ->
-        {
-            MyField = x
-        })
+    |> Option.map
+        (fun x ->
+            {
+                MyField = x
+            })
 ```
 
 The same rules apply for list and array elements.
@@ -796,10 +797,11 @@ Pattern matching of anonymous functions, starting by `function`, should generall
 
 ```fsharp
 lambdaList
-|> List.map (function
-    | Abs(x, body) -> 1 + sizeLambda 0 body
-    | App(lam1, lam2) -> sizeLambda (sizeLambda 0 lam1) lam2
-    | Var v -> 1)
+|> List.map
+    (function
+        | Abs(x, body) -> 1 + sizeLambda 0 body
+        | App(lam1, lam2) -> sizeLambda (sizeLambda 0 lam1) lam2
+        | Var v -> 1)
 ```
 
 Pattern matching in functions defined by `let` or `let rec` should be indented four spaces after starting of `let`, even if `function` keyword is used:
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -832,11 +832,27 @@ with
 
 ## Formatting function parameter application
 
-In general, most function parameter application is done on the same line.
-
-If you wish to apply parameters to a function on a new line, indent them by one scope.
+In general, most function parameter application is done on the same line:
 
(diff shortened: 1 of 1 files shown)


samarsha:
Thanks for the clarification! There are examples in the style guide that don't follow this format, such as:

* The lambda to `Option.map` here: https://github.com/dotnet/docs/blob/299e34e1020ffdc96dba176ce1f9d046d7d38676/docs/fsharp/style-guide/formatting.md#formatting-records
* The `function` to `List.map` here: https://github.com/dotnet/docs/blob/299e34e1020ffdc96dba176ce1f9d046d7d38676/docs/fsharp/style-guide/formatting.md#pattern-matching-constructs

Should those be updated?

docs/fsharp/style-guide/formatting.md
```diff
@@ -832,11 +832,27 @@ with
 
 ## Formatting function parameter application
 
-In general, most function parameter application is done on the same line.
+In general, most function parameter application is done on the same line:
```
IEvangelist, line 835: I'm not really sure this line makes sense - perhaps it's my TypeScript and C# background or unfamiliarity with F#, but the term "function parameter application" to me is just "argument", right?

```suggestion
In general, most arguments are provided on the same line:
```

docs/fsharp/style-guide/formatting.md
```diff
@@ -832,11 +832,27 @@ with
...
+
+When pipelines are concerned, the same is typically also true, where a curried function is applied as an argument on the same line:
+
+```
+let printListWithOffsetPiped a list1 =
+    list1
+    |> List.iter (fun elem -> printfn $"%d{a + elem}")
+```
+
+However, you may wish to apply parameters to a function on a new line, as a matter of readability or because the list of arguments or the argument names are too long. In that case, indent with one scope:
```
IEvangelist, line 852:
```suggestion
However, you may wish to pass arguments to a function on a new line, as a matter of readability or because the list of arguments or the argument names are too long. In that case, indent with one scope:
```

Again, perhaps this is an F# term that is getting me a bit mixed up. But when I read "apply parameters" I think that I'm building out a function, which doesn't seem to be the case. Instead, it seems as though we're providing arguments to a function - thus, we "provide arguments" or "pass arguments" - not "apply parameters, right?

IEvangelist (approved):
I'm good with these changes, as long as the correct terms are used - I made a few suggestions and asked a few questions.

docs/fsharp/style-guide/formatting.md
cartermp: It's technically more correct to call it parameter application, but I think this is fine for an explanation

cartermp:
@samarsha added changes to those.

drhumlen:
Hmm.. Are sure about this? I like the clarification :D but I still _like_ the former style more. Our codebase gained a significant amount of extra lines and indentation after formatting with this new (Fantomas) rule. Maybe I just need to get used to it 🤔 but the old style was much more dense/terse, whereas now it's more "airy". And while a little air is nice, too much air makes the code harder to navigate imo. One strategy is ofc to replace inline lambdas with a "more proper" `let` - but then you lose the type inference on the parameters.

Thoughts? Has anyone looked at the before & after this new style change on a larger code base? Maybe I just have to adjust to more airy code. 😄

cartermp:
Note that the guidance does not state anything hard and fast, just that if lines are getting too long you should indent.

drhumlen:
Before vs after this style rule
```fsharp
let names =
    people
    |> List.filter (fun person -> person.Age >= 18)
    |> List.map (fun person ->
        let petNames =
            pets
            |> List.choose (fun pet ->
                if pet.IsAlive then Some pet.Name else None)
        "{person.Name} has pets: {pets}")
```
vs
```fsharp
let names =
    people
    |> List.filter (fun person -> person.Age >= 18)
    |> List.map
        (fun person ->
            let petNames =
                pets
                |> List.choose
                    (fun pet ->
                        if pet.IsAlive then
                            Some pet.Name
                        else
                            None)

            "{person.Name} has pets: {pets}")
```

cartermp:
Sure, but that's also not what I said. There is no hard and fast rule here. The width chosen by fantomas before a line break is just a particular width, which can change and/or be configured. It is orthogonal to guidance about placing things on new lines when lines start to get long.

drhumlen:
Fantomas isn't splitting it into 2 lines because of the width, but because it's multi-lined lambda. The current interpretation is that: all multi-lined lambdas must be placed on the next line. So since `(fun person ->` stretches over multiple lines here, it goes on the next line in the second example.

But if I follow you, you're saying that's _not_ a rule, and that my first code example is just as valid as the latter? 

Which means we can change it back to the old behaviour in Fantomas 🤔  ?  Given that most people prefer the former style is better ofc 😄 Maybe I'm the odd one out here.

✨ Summoning   @nojaf ✨

cartermp:
The documentation clearly says this:

> For lambda expressions, you may also want to consider placing the body of a lambda expression on a new line, indented by one scope, **if it is long enough**:

There is certainly plenty of room for interpretation here, but I stand by my statements in this thread.

nojaf:
Hey all, I don't really have a stake in this.
My two cents:
- The new rule is more consistent across arguments passed to functions.
  Meaning putting the lambda on the next line is exactly the same behaviour as all the other [SynExpr ](https://fsharp.github.io/FSharp.Compiler.Service/reference/fsharp-compiler-syntaxtree-synexpr.html) expressions, which I like.
- All change is hard, in time you will learn to live with this.

PR #21990, opened 2020-12-14
Update snippets.5000 for new custom host setting

adegeo:
## Summary

- Add logic to check for "custom" host in snippets.5000.json file.
- "dotnet" host will use the default value.
- Throw error if snippets.5000.json file present but nothing used from it.


IEvangelist (approved):

PR #21999, opened 2020-12-14
Fix partial read calls

tdykstra:
Follow-up to #21991


BillWagner (approved):
This LGTM @tdykstra 

You can :shipit: when ready

PR #22000, opened 2020-12-14
Move syslib warnings into subfolder

gewarren:
To keep things neat and tidy, as I anticipate more of these coming along.


BillWagner (approved):
This LGTM @gewarren 

I assume you looked at traffic and felt that redirections weren't necessary? If not, consider adding them.

gewarren:
> I assume you looked at traffic and felt that redirections weren't necessary? If not, consider adding them.

Good catch!

PR #22003, opened 2020-12-15
Design-time build breaking change

### LINKED ISSUE #20372: Design Time Builds will only return top-level packages references by default
## Design Time Builds will only return top-level packages references by default

Historically, the `RunResolvePackageDependencies` target would create the following MSBuild items containing information from the NuGet assets file:

- PackageDefinitions
- PackageDependencies
- TargetDefinitions
- FileDefinitions
- FileDependencies

This data was used to populate the Dependencies node in Solution Explorer in Visual Studio.  However, it could be a large amount of data, and most of it was not generally used.  So now, most of these items are not generated by default, and Visual Studio reads the information for the Dependencies node directly from the assets file.  Find out more here: https://github.com/dotnet/sdk/pull/11358

### Version introduced

.NET SDK 3.1.400 and Visual studio 16.7.

### Reason for change
This was done to improve project load performance inside of Visual Studio.  Previously, all package references would be loaded which would involve loading a lot of references that a customer would never view.

### Recommended action
If you have MSBuild logic that depends on these items being created, set the  `EmitLegacyAssetsFileItems` property to `true` in your project file.  This will revert to the previous behavior where these items are all created.

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #20372.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/3.1?branch=pr-en-us-22003#design-time-builds-only-return-top-level-package-references).


dsplaisted (approved):
Looks good, though I kind of thought we had already documented this.

BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready

gewarren:
Thanks for the reviews!

PR #22027, opened 2020-12-16
Remove duplicated command

evolvedlight:
The creation of the template pack is done above, this looks like it was accidentally copied here

## Summary

Describe your changes here.


evolvedlight:
But "dotnet pack" is also duplicated - why is this? Can I remove that one too?

tdykstra (approved):
Thanks @evolvedlight !

tdykstra:
>But "dotnet pack" is also duplicated - why is this? Can I remove that one too?

Sorry, missed this additional comment earlier.  Yes you can remove that too.  These are relics of a mass change some time ago to split up code blocks in order to use dotnetcli language identifier wherever possible.

PR #22055, opened 2020-12-17
Fix multi-targeting csproj example

rgwood:
## Summary

The current example after "**Warns** if the project is multi-targeted" does not actually show multi-targeting. Guessing this was just a copy/paste error.


BillWagner (approved):
Thanks for submitting this and fixing it @rgwood 

I'll :shipit: now.

Issue #22040, opened 2020-12-16
Describe solution for classes derived from disposable class


Mikejo5000:
Moved issue to .NET
Old issue: [https://github.com/MicrosoftDocs/visualstudio-docs/issues/4342](https://github.com/MicrosoftDocs/visualstudio-docs/issues/4342), opened by @jameslear

Dispose is one of the more arcane methods for newbies, especially when it is overloaded.  System.Windows.Forms.Form is a common example.  Any chance we could see a sample implementation using a disposable base class with an overloaded Dispose?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: d118a61c-9b0d-6046-79e8-3ddff8282c7f
* Version Independent ID: 58b6286d-20d7-76ac-bab4-4078b8e71767
* Content: [CA1001: Types that own disposable fields should be disposable (code analysis) - .NET](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1001)
* Content Source: [docs/fundamentals/code-analysis/quality-rules/ca1001.md](https://github.com/dotnet/docs/blob/master/docs/fundamentals/code-analysis/quality-rules/ca1001.md)
* Product: **dotnet-fundamentals**
* GitHub Login: @gewarren
* Microsoft Alias: **gewarren**

gewarren:
@jameslear [Here is an example](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#implement-the-dispose-pattern-for-a-derived-class). I've also added a "see also" link to the CA1001 reference page.

PR #22091, opened 2020-12-19
Fix ups to CA1001 and CA1062

gewarren:
Contributes to #22040.


IEvangelist (approved):

PR #22133, opened 2020-12-23
Remove extra underscore in configuration-providers.md

skrysm:
## Summary

I think there's a typo in <https://docs.microsoft.com/en-us/dotnet/core/extensions/configuration-providers#environment-variable-configuration-provider>. Instead of `CustomPrefix__SecretKey` it should read `CustomPrefix_SecretKey` (only one underscore). Or am I interpreting this wrong?

### DIFF
--- a/docs/core/extensions/configuration-providers.md
+++ b/docs/core/extensions/configuration-providers.md
@@ -191,9 +191,9 @@ The prefix is stripped off when the configuration key-value pairs are read.
 The following commands test the custom prefix:
 
 ```dotnetcli
-set CustomPrefix_SecretKey="Secret key with CustomPrefix_ environment"
-set CustomPrefix_TransientFaultHandlingOptions__Enabled=true
-set CustomPrefix_TransientFaultHandlingOptions__AutoRetryDelay=00:00:21
+set CustomPrefix__SecretKey="Secret key with CustomPrefix_ environment"
+set CustomPrefix__TransientFaultHandlingOptions__Enabled=true
+set CustomPrefix__TransientFaultHandlingOptions__AutoRetryDelay=00:00:21
 
 dotnet run
 ```
--- a/docs/core/extensions/configuration-providers.md
+++ b/docs/core/extensions/configuration-providers.md
@@ -191,7 +191,7 @@ The prefix is stripped off when the configuration key-value pairs are read.
 The following commands test the custom prefix:
 
 ```dotnetcli
-set CustomPrefix__SecretKey="Secret key with CustomPrefix_ environment"
+set CustomPrefix_SecretKey="Secret key with CustomPrefix_ environment"
 set CustomPrefix_TransientFaultHandlingOptions__Enabled=true
 set CustomPrefix_TransientFaultHandlingOptions__AutoRetryDelay=00:00:21
 


docs/core/extensions/configuration-providers.md
```diff
@@ -191,7 +191,7 @@ The prefix is stripped off when the configuration key-value pairs are read.
 The following commands test the custom prefix:
 
 ```dotnetcli
-set CustomPrefix__SecretKey="Secret key with CustomPrefix_ environment"
+set CustomPrefix_SecretKey="Secret key with CustomPrefix_ environment"
 set CustomPrefix_TransientFaultHandlingOptions__Enabled=true
 set CustomPrefix_TransientFaultHandlingOptions__AutoRetryDelay=00:00:21
```
IEvangelist, line 196:
```suggestion
set CustomPrefix__TransientFaultHandlingOptions__Enabled=true
set CustomPrefix__TransientFaultHandlingOptions__AutoRetryDelay=00:00:21
```

docs/core/extensions/configuration-providers.md
```diff
@@ -191,7 +191,7 @@ The prefix is stripped off when the configuration key-value pairs are read.
 The following commands test the custom prefix:
 
 ```dotnetcli
-set CustomPrefix__SecretKey="Secret key with CustomPrefix_ environment"
+set CustomPrefix_SecretKey="Secret key with CustomPrefix_ environment"
```
IEvangelist, line 194:
```suggestion
set CustomPrefix__SecretKey="Secret key with CustomPrefix_ environment"
```

IEvangelist (commented):
Ah, good catch. But the others are actually incorrect. The double underscore is how they should all be.

IEvangelist:
If you add the environment variables:

```
setx CustomPrefix__SecretKey "Secret key with CustomPrefix_ environment"
setx CustomPrefix__TransientFaultHandlingOptions__Enabled true
setx CustomPrefix__TransientFaultHandlingOptions__AutoRetryDelay 00:00:21
```

And run it with this code below you'll see the values.

```csharp
using System.Threading.Tasks;
using Microsoft.Extensions.Configuration;
using Microsoft.Extensions.Hosting;

namespace ConsoleEnvironment.Example
{
    class Program
    {
        static async Task Main(string[] args)
        {
            using IHost host = CreateHostBuilder(args).Build();

            // Application code should start here.
            var config = host.Services.GetService(typeof(IConfiguration));
            if (config is IConfiguration configuration)
            {
                static void Write(string message) => System.Console.WriteLine(message);

                Write(configuration["CustomPrefix:SecretKey"]);
                Write(configuration["CustomPrefix:TransientFaultHandlingOptions:Enabled"]);
                Write(configuration["CustomPrefix:TransientFaultHandlingOptions:AutoRetryDelay"]);
            }

            await host.RunAsync();
        }

        static IHostBuilder CreateHostBuilder(string[] args) =>
            Host.CreateDefaultBuilder(args)
                .ConfigureAppConfiguration((hostingContext, configuration) =>
                    configuration.AddEnvironmentVariables(
                        prefix: "CustomPrefix_"));
    }
}
```

IEvangelist (approved):

skrysmanski:
@IEvangelist I think you misinterpreted how this should work.

Your code should have been written like this:

```c#
Write(configuration["SecretKey"]);
Write(configuration["TransientFaultHandlingOptions:Enabled"]);
Write(configuration["TransientFaultHandlingOptions:AutoRetryDelay"]);
```

The point here is that the prefix is *only* used for the environment variables to make an environment variable named `CustomPrefix_SecretKey` overwrite an environment variable named `SecretKey`.

Or am I interpreting this wrong?

PR #22138, opened 2020-12-23
Removed init keyword from property

dutts:
## Summary

Documentation introduces this snippet by talking about language features available from 7.2. The init accessor isn't available until 9.0. While this is mentioned later it's confusing.

### DIFF
--- a/docs/csharp/language-reference/builtin-types/snippets/shared/StructType.cs
+++ b/docs/csharp/language-reference/builtin-types/snippets/shared/StructType.cs
@@ -59,8 +59,8 @@ namespace builtin_types
                 Y = y;
             }
 
-            public double X { get; init; }
-            public double Y { get; init; }
+            public double X { get; }
+            public double Y { get; }
 
             public override string ToString() => $"({X}, {Y})";
         }


BillWagner:
Thanks for pointing this out and making the update @dutts 

I think a better fix would be to explain the `init` syntax that's currently used in the example. Thoughts?

/cc @pkulikov

pkulikov:
Possible alternative solution is to put the example last in the section, after init accessors are mentioned. (Basically, that's in line with the suggestion of @BillWagner)

@dutts @BillWagner what do you think? If you agree, then let's close this PR and I will submit a PR with the article update.

dutts:
@BillWagner @pkulikov I the suggestion of moving the example to after the `init` accessors have been mentioned would flow better.

pkulikov:
@dutts your reply misses the most important part - the verb :) however it looks you agree with the alternative solution?

dutts:
> @dutts your reply misses the most important part - the verb :) however it looks you agree with the alternative solution?

lol, yes that's correct =)

pkulikov:
> @BillWagner @pkulikov I think the suggestion of moving the example to after the `init` accessors have been mentioned would flow better.

cool, thank you @dutts 
I'll open the PR for that later this week then.

pkulikov:
> I'll close this and wait for the updated PR.

@BillWagner an updated PR: #22230

PR #22145, opened 2020-12-24
fixed command typo

shunjid:
## Summary

While reading the documentation, I found issues with commands for regression and recommendation section.

- Replaced `classification` with `regression` in regression section
- Replaced `classification` with `recommendation` in recommendation section


luisquintanilla (approved):
Thanks for these changes @shunjid

PR #22159, opened 2020-12-26
FS0037

brettrowberry:
## Summary

There's kind of a lot going on here.

- Added compiler error FS0037
- FS0052 .md file now uses code text instead of a pull quote like every other F# example
- FS0052 F# script uses multiline quotes like every other example
- The F# compiler messages index page now says how to contribute
- `create-new-fsharp-compiler-message.fsx` now has a shebang, includes path from root of repo for convenience in calling, uses interpolated strings (now that Ionide 5 is out!), prevents insertion of duplicate F# compiler error numbers into the table of contents, F# script code example comment uses multiline quotes

Related to #16177

I am happy to workshop any of these changes, especially since some have been debated in prior PRs, like
- how to format the error messages in the `.md` files and 
- whether to use multiline comments in the `.fsx` files.
 
I just want a good, consistent result.


cartermp (approved):
Thanks! I like the changes.

PR #22186, opened 2020-12-30
Update console-webapiclient.md

adaggarwal:
## Summary

Before `dotnet run`, the user should be inside the directory which contains the csproj otherwise it'll error out.


BillWagner (approved):
Thanks for spotting this and making the fix @adaggarwal We appreciate it.

I'll :shipit: now.

PR #22205, opened 2021-01-03
Dapr for .NET Developers - Chapter 8: Bindings

EdwinVW:
Chapter 8 of the Dapr for .NET devs ebook about bindings.


amolenk:
This is chapter 8 about the Bindings building block.

@nishanil / @IEvangelist please review. Thanks!

docs/architecture/dapr-for-net-developers/bindings-buildingblock.md
```diff
@@ -0,0 +1,235 @@
...
+```
+
+Note that the HTTP port is the same as used by the Dapr sidecar (in this case, the default Dapr HTTP port `3500`).
+
+The structure of the payload (that is, message sent) will vary per binding. In the example above, the payload contains a `data` element with a message. Bindings to other types of external resources can be different, especially for the metadata that is sent. Each payload must also contain an `operation` field, that defines the operation the binding will execute. The above example specifies a  `create` operation that creates the SMS message. Common operations include:
+
+- create
+- get
+- delete
+- list
```
IEvangelist, line 101: What about `update`, is that a thing? I initially thought of CRUD when I read this list, but noticed that `update` isn't called out. Also, assuming that `create` is idempotent - it cannot be an update?

docs/architecture/dapr-for-net-developers/bindings-buildingblock.md
```diff
@@ -0,0 +1,235 @@
...
+{
+    var string message = CreateEmailBody(notification);
+    var metadata = new Dictionary<string,string> { 
+      { emailFrom = "eShopOn@dapr.io" },
+      { emailTo = notification.UserName },
+      { subject = $"Your eShopOnDapr order #{notification.Order.Id}" }
+    };
+
+    return _daprClient.InvokeBindingAsync("sendmail", "create", message, metadata, cancellationToken);
+}
```
IEvangelist, line 214:
```suggestion
public Task Handle(OrderStartedDomainEvent notification, CancellationToken cancellationToken)
{
    var string message = CreateEmailBody(notification);
    var metadata = new Dictionary<string, string>
    { 
        ["emailFrom"] = "eShopOn@dapr.io",
        ["emailTo"] = notification.UserName,
        ["subject"] = $"Your eShopOnDapr order #{notification.Order.Id}"
    };

    return _daprClient.InvokeBindingAsync("sendmail", "create", message, metadata, cancellationToken);
}
```

Fix compiler error, the collection initializer for a dictionary doesn't work this way - this code does not compile. Also, where does the, `emailFrom`, `emailTo`, and `subject` values comes from? Or are they the string literals for the metadata property names?

IEvangelist (changes requested):
There are several compile errors with the C# snippets include, and a few other minor nits.

EdwinVW:
> There are several compile errors with the C# snippets include, and a few other minor nits.

@IEvangelist I'll review your suggestions a.s.a.p. 
Thanks!

docs/architecture/dapr-for-net-developers/bindings-buildingblock.md
```diff
@@ -0,0 +1,235 @@
...
+    public ActionResult Post(JsonElement data)
+    {
+        // handle tweet
+        //...
+
+        // acknowledge message
+        return Ok();
+    }
+  
+    //...
```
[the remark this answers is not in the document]
EdwinVW: Yes, great suggestion. I will change that.

docs/architecture/dapr-for-net-developers/bindings-buildingblock.md
EdwinVW: Yes, that's possible. There is a default list of operation constants (see [definition](https://github.com/dapr/components-contrib/blob/aba184f9425b4e1950ad8d8e60d6912f37827d65/bindings/requests.go#L16)), but a binding can add its own operations to the list (in the end it's just a string).

I will add a remark about that in the text.

docs/architecture/dapr-for-net-developers/bindings-buildingblock.md
EdwinVW: Yes, the email fields are metadata property names.

EdwinVW:
@IEvangelist I've applied your suggestions. Please review.

IEvangelist (approved):

PR #22210, opened 2021-01-04
Corrected wording about record specific allowances to options pattern

### LINKED ISSUE #22134: Records don't fullfill any requirement of option classes
The "Options pattern" description contains this part:

> Create the following `TransientFaultHandlingOptions` class:
> 
> ```c#
> public record TransientFaultHandlingOptions(
>     bool Enabled = false, TimeSpan AutoRetryDelay = default);
> ```
>
> An options class:
>
> * Must be non-abstract with a public parameterless constructor
> * Contain public read-write properties to bind (fields are not bound)

Both requirements are **not true** for the specified record example:

> Must be non-abstract with a public parameterless constructor

The `TransientFaultHandlingOptions` does *not* have a parameterless constructor.

> Contain public read-write properties to bind (fields are not bound)

The properties of `TransientFaultHandlingOptions` are read-only - or `init`-only (not read-write).

I haven't checked but I'm assuming that records are supported none the less. It's just that the description doesn't match the reality.

I'm guessing the following is true for the options pattern:
* It allows for either parameterless constructors *or* for constructors where *all* parameters have default values.
* Properties can be read-write or read-initonly (which, to my understanding, is the same when accessing the property through reflection).

---
#### Document Details

* Content: [Options pattern in .NET](https://docs.microsoft.com/en-us/dotnet/core/extensions/options)
* Content Source: [docs/core/extensions/options.md](https://github.com/dotnet/docs/blob/master/docs/core/extensions/options.md)
* GitHub Login: @IEvangelist
* Microsoft Alias: **dapine**

IEvangelist:
## Summary

Corrected wording about record specific allowances to options pattern

Fixes #22134

## Preview

✔️ [Options pattern in .NET](https://review.docs.microsoft.com/en-us/dotnet/core/extensions/options?branch=pr-en-us-22210)

### DIFF
--- a/docs/core/extensions/options.md
+++ b/docs/core/extensions/options.md
@@ -17,7 +17,7 @@ Options also provide a mechanism to validate configuration data. For more inform
 
 ## Bind hierarchical configuration
 
-The preferred way to read related configuration values is using the options pattern. The options pattern is possible through the <xref:Microsoft.Extensions.Options.IOptions%601> interface, where the generic type parameter `T` is constrained to `class`. The `IOptions<T>` can later be provided through dependency injection. For more information, see [Dependency injection in .NET](dependency-injection.md).
+The preferred way to read related configuration values is using the options pattern. The options pattern is possible through the <xref:Microsoft.Extensions.Options.IOptions%601> interface, where the generic type parameter `TOptions` is constrained to `class`. The `IOptions<TOptions>` can later be provided through dependency injection. For more information, see [Dependency injection in .NET](dependency-injection.md).
 
 For example, to read the following configuration values:
 
@@ -27,6 +27,7 @@ Create the following `TransientFaultHandlingOptions` class:
 
 :::code language="csharp" source="snippets/configuration/console-json/TransientFaultHandlingOptions.cs" range="5-9":::
 
+<span id="options-class"></span>
 When using the options pattern, an options class:
 
 - Must be non-abstract with a public parameterless constructor
@@ -39,9 +40,6 @@ The following code:
 
 :::code language="csharp" source="snippets/configuration/console-json/Program.cs" range="31-38":::
 
-> [!NOTE]
-> The <xref:Microsoft.Extensions.Configuration.ConfigurationBinder.Bind(Microsoft.Extensions.Configuration.IConfiguration,System.Object)?displayProperty=nameWithType> extension method allows for any `object` to be bound and is **not** constrained to a `class`.
-
 In the preceding code, changes to the JSON configuration file after the app has started are read.
 
 [`ConfigurationBinder.Get<T>`](xref:Microsoft.Extensions.Configuration.ConfigurationBinder.Get%2A) binds and returns the specified type. `ConfigurationBinder.Get<T>` may be more convenient than using `ConfigurationBinder.Bind`. The following code shows how to use `ConfigurationBinder.Get<T>` with the `TransientFaultHandlingOptions` class:
@@ -51,13 +49,16 @@ IConfigurationRoot configurationRoot = configuration.Build();
 
 var options =
     configurationRoot.GetSection(nameof(TransientFaultHandlingOptions))
-        .Get<TransientFaultHandlingOptions>();
+                     .Get<TransientFaultHandlingOptions>();
 
 Console.WriteLine($"TransientFaultHandlingOptions.Enabled={options.Enabled}");
 Console.WriteLine($"TransientFaultHandlingOptions.AutoRetryDelay={options.AutoRetryDelay}");
 ```
 
 In the preceding code, changes to the JSON configuration file after the app has started are read.
+
+> [!IMPORTANT]
+> The <xref:Microsoft.Extensions.Configuration.ConfigurationBinder> class exposes several APIs, such as `.Bind(object instance)` and `.Get<T>()` that are ***not*** contained to `class`. When using any of the [Options interfaces](#options-interfaces), you must adhere to aforementioned [options class constraints](#options-class).
 
 An alternative approach when using the options pattern is to bind the `"TransientFaultHandlingOptions"` section and add it to the [dependency injection service container](dependency-injection.md). In the following code, `TransientFaultHandlingOptions` is added to the service container with <xref:Microsoft.Extensions.DependencyInjection.OptionsConfigurationServiceCollectionExtensions.Configure%2A> and bound to configuration:
 
--- a/docs/core/extensions/options.md
+++ b/docs/core/extensions/options.md
@@ -3,7 +3,7 @@ title: Options pattern in .NET
 author: IEvangelist
 description: Learn how to use the options pattern to represent groups of related settings in .NET apps.
 ms.author: dapine
-ms.date: 12/04/2020
+ms.date: 01/04/2021
 ---
 
 # Options pattern in .NET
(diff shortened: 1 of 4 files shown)


docs/core/extensions/options.md
```diff
@@ -27,8 +27,8 @@ Create the following `TransientFaultHandlingOptions` class:
 
 An options class:
```
skrysmanski, line 28: @IEvangelist  I think, if you want to reword the list below like you did, you also need to reword this paragraph.

docs/core/extensions/options.md
```diff
@@ -27,8 +27,8 @@ Create the following `TransientFaultHandlingOptions` class:
...
 
-* Must be non-abstract with a public parameterless constructor
-* Contain public read-write properties to bind (fields are ***not*** bound)
+- It allows for either parameterless constructors, or for constructors where all parameters have default values
+- Properties can be read-write, or read-init only
```
BillWagner, line 31: Throughout the C# docs, we're standardizing on the phrase "init only" for these properties:

```suggestion
- Properties can be read-write, or init only
```

BillWagner: Your edit removes the requirement that options classes must be non-abstract. Was that intentional?

IEvangelist: Sort of, I mean they allow records - perhaps it should be a bit more explicit?

docs/core/extensions/options.md
```diff
@@ -27,8 +27,8 @@ Create the following `TransientFaultHandlingOptions` class:
 
 An options class:
 
-- It allows for either parameterless constructors, or for constructors where all parameters have default values
-- Properties can be read-write, or read-init only
+- Must be a non-abstract class with either a public parameterless constructor, or a public constructor where all parameters have default values.
```
skrysmanski, line 30: You added "class" here. Is this correct? Or can structs also be used for options?

docs/core/extensions/options.md
```diff
@@ -27,8 +27,8 @@ Create the following `TransientFaultHandlingOptions` class:
...
+- All bindable properties must be read-write, or init only. Fields are **not** bound.
```
skrysmanski, line 31: You changed "public" to "bindable". Was this intentional? Are any other visibilities except "public" supported here?

docs/core/extensions/options.md
BillWagner: Also, records.  Positional records won't work because of the constructor requirement.

docs/core/extensions/options.md
BillWagner: My take: I think the most precise term would be "accessible". In practice, I think that means public properties. I suppose an option type could have additional properties that are protected or internal, but aren't bound. That's true, but not very helpful.

skrysmanski: @BillWagner So would you prefer "accessible" or "public"? I personally would tend to "public" because that's what developers will write (and "accessible" is a little muddy in my opinion because it depends on the "point-of-view"; e.g. private members are accessible from within a class but they're not what this paragraph is getting at).

docs/core/extensions/options.md
skrysmanski: >  Positional records won't work because of the constructor requirement.

Are you sure about this? Aren't positional records something like this (which is the example given in the docs):

```c#
public record TransientFaultHandlingOptions(bool Enabled = false, TimeSpan AutoRetryDelay = default);
```

IEvangelist: Positional records do actually work, I have tested that locally. What I have not done was exhaustively tested all variants of access modifier and type, i.e.; `struct`, `readonly struct`, `tuple`, `interface`, etc. I'm assuming that there is some generic type parameter constraint, but I haven't dug into that yet. What we have thus far is technically still accurate, just might be missing something.

docs/core/extensions/options.md
IEvangelist: There actually seems to be an option to bind non public too, https://github.com/dotnet/runtime/blob/master/src/libraries/Microsoft.Extensions.Configuration.Binder/src/BinderOptions.cs.

docs/core/extensions/options.md
```diff
@@ -27,6 +27,7 @@ Create the following `TransientFaultHandlingOptions` class:
 
 :::code language="csharp" source="snippets/configuration/console-json/TransientFaultHandlingOptions.cs" range="5-9":::
 
+<span id="options-class"></span>
```
skrysmanski, line 30: Was this line intentional?

docs/core/extensions/options.md
```diff
@@ -17,18 +17,20 @@ Options also provide a mechanism to validate configuration data. For more inform
...
-:::code language="csharp" source="snippets/configuration/console-json/TransientFaultHandlingOptions.cs" range="5-6":::
+:::code language="csharp" source="snippets/configuration/console-json/TransientFaultHandlingOptions.cs" range="5-9":::
 
-An options class:
+When using the options pattern, an options class:
 
-- Must be a non-abstract class with either a public parameterless constructor, or a public constructor where all parameters have default values.
-- All bindable properties must be read-write, or init only. Fields are **not** bound.
+- Must be non-abstract with a public parameterless constructor
+- Contain public read-write properties to bind (fields are ***not*** bound)
```
skrysmanski, line 34: Now the description is exactly as it was in the beginning. Does this mean that records are not supported?

docs/core/extensions/options.md
```diff
@@ -46,14 +49,17 @@ IConfigurationRoot configurationRoot = configuration.Build();
...
+                     .Get<TransientFaultHandlingOptions>();
 
 Console.WriteLine($"TransientFaultHandlingOptions.Enabled={options.Enabled}");
 Console.WriteLine($"TransientFaultHandlingOptions.AutoRetryDelay={options.AutoRetryDelay}");
 ```
 
 In the preceding code, changes to the JSON configuration file after the app has started are read.
 
+> [!IMPORTANT]
+> The <xref:Microsoft.Extensions.Configuration.ConfigurationBinder> class exposes several APIs, such as `.Bind(object instance)` and `.Get<T>()` that are ***not*** contained to `class`. When using any of the [Options interfaces](#options-interfaces), you must adhere to aforementioned [options class constraints](#options-class).
```
skrysmanski, line 61: "contained"? Or "constrained"?

docs/core/extensions/options.md
IEvangelist: Yes, this allows for linking to this section without any visible indicators.

docs/core/extensions/options.md
IEvangelist: Records are not supported for the options pattern, that is correct. I changed the type to be a regular class instead.

docs/core/extensions/options.md
IEvangelist, line 61:
```suggestion
> The <xref:Microsoft.Extensions.Configuration.ConfigurationBinder> class exposes several APIs, such as `.Bind(object instance)` and `.Get<T>()` that are ***not*** constrained to `class`. When using any of the [Options interfaces](#options-interfaces), you must adhere to aforementioned [options class constraints](#options-class).
```

IEvangelist: Good catch

adegeo: Did you actually apply this suggestion? It doesn't seem like it.

BillWagner (approved):
Thanks @IEvangelist 

PR #22212, opened 2021-01-04
Added Elsa workflow

### LINKED ISSUE #123: Topics not in the TOC and blockquote
As I'm starting to validate the files I have in my cloned core-docs folder, I'm finding topics that are there but not in the TOC. Is that by design? Should I even be looking at those?
For example:
https://github.com/dotnet/core-docs/blob/master/docs/concepts/assembly-format.rst
https://github.com/dotnet/core-docs/blob/master/docs/concepts/class-libraries.rst

Also, for the first topic above, when converted to MD, the second paragraph is using a blockquote (">"). Is that the intended result? Or do we want to just indent the paragraph?

Alexplose:
Added Elsa workflow as an alternative to CoreWF


IEvangelist:
Fixes: https://github.com/dotnet/docs.fr-fr/issues/123

IEvangelist (approved):

PR #22215, opened 2021-01-04
improved clarity #22149

### LINKED ISSUE #22149: Classes vs class fields
> Record fields differ from classes in that they are automatically exposed as properties, and they are used in the creation and copying of records.

I think it is obvious that record fields and _classes_ are different.  Instead, I think the intention here is to contrast record fields with _class fields_.

If this is the intention, then I can create a PR to replace "classes" with "class fields".

TysonMN:
## Summary

replaced "classes" with "class fields"

Fixes #22149


cartermp (commented):
Thanks!

cartermp (approved):

PR #22221, opened 2021-01-04
Update windows install page for switches

adegeo:
## Summary

Per an email thread, adding this information.


docs/core/install/windows.md
```diff
@@ -264,12 +264,32 @@ While Visual Studio Code doesn't come with an automated .NET Core installer like
...
+Prevents any UI and prompts from displaying.
+
+- `norestart`\
+Suppresses any attempts to restart.
+
+```console
+dotnet-sdk-3.1.301-win-x64.exe /install /quiet /norestart
+```
+
+For more information, see [Standard Installer Command-Line Options](/windows/win32/msi/standard-installer-command-line-options).
```
joeloff, line 288: Should we point them to the MSI reference? The commandline options available are what the WiX Engine supports, not MSI.

jamshedd: Looks great

docs/core/install/windows.md
```diff
@@ -264,12 +264,32 @@ While Visual Studio Code doesn't come with an automated .NET Core installer like
...
+
+> [!TIP]
+> If the installer returns exit code 3010, it indicates that a reboot is required.
+
```
jamshedd, line 292: Return code 0 is success (no reboot required), 3010 is success (reboot required), and anything else is generally an installation failure.

docs/core/install/windows.md
adegeo: I thought about pointing them here for the MSI switches: https://docs.microsoft.com/en-us/windows/win32/msi/command-line-options but then I thought it may be confusing since it clearly indicates `msiexec.exe` and the page I did link to is more generic.

joeloff: Although those commands look similar to what the .NET bundles use, both pages apply to ```msiexec```.

adegeo: @joeloff I thought WiX is just a wrapper to configure MSIs during install? Do you have a specific alternative page in mind? Or do you just suggest deleting the link all together?

IEvangelist (approved):

PR #22222, opened 2021-01-04
Suppress false positive markdown violation

adegeo:
## Summary

This file is causing some linter errors on various PRs. This error is a false positive though and this PR disables that specific error check.

### DIFF
--- a/docs/standard/base-types/regular-expression-example-scanning-for-hrefs.md
+++ b/docs/standard/base-types/regular-expression-example-scanning-for-hrefs.md
@@ -33,7 +33,9 @@ The following example searches an input string and displays all the href="…" v
  [!code-vb[RegularExpressions.Examples.HREF#2](../../../samples/snippets/visualbasic/VS_Snippets_CLR/RegularExpressions.Examples.HREF/vb/example.vb#2)]  
   
  The regular expression pattern `href\s*=\s*(?:["'](?<1>[^"']*)["']|(?<1>\S+))` is interpreted as shown in the following table.  
-  
+
+<!-- markdownlint-disable MD011 -->
+
 |Pattern|Description|  
 |-------------|-----------------|  
 |`href`|Match the literal string "href". The match is case-insensitive.|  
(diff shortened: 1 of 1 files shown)


docs/standard/base-types/regular-expression-example-scanning-for-hrefs.md
```diff
@@ -33,7 +33,9 @@ The following example searches an input string and displays all the href="…" v
  [!code-vb[RegularExpressions.Examples.HREF#2](../../../samples/snippets/visualbasic/VS_Snippets_CLR/RegularExpressions.Examples.HREF/vb/example.vb#2)]  
   
  The regular expression pattern `href\s*=\s*(?:["'](?<1>[^"']*)["']|(?<1>\S+))` is interpreted as shown in the following table.  
-  
+
+<!-- markdownlint-disable MD011 -->
```
Youssef1313, line 37: The table doesn't render correctly, so this isn't a false positive.

image


One possible fix is to use `<code></code>` instead of backticks, and use the encoding of the vertical bar. Not sure if there is a better solution.

IEvangelist: It does render correctly. See the live site:

image

IEvangelist (approved):
Looks good to me @adegeo, :shipit: when you're ready.

docs/standard/base-types/regular-expression-example-scanning-for-hrefs.md
IEvangelist: I agree that it would be nice to have it render correctly in both, but my primary concern is the docs site. Instead of using `<code></code>`, are there any obvious or alternative approaches you have in mind that do not break the inline HTML concern/linting rules?

Youssef1313: @IEvangelist Happy new year to you too!

I'd probably ask @DavidAnson for help on this. I opened https://github.com/DavidAnson/markdownlint/issues/363 asking for help on this.

Meanwhile, you can merge to fix the docs site, which is more important than GitHub-view.

DavidAnson:
I fixed something that looks very much like this issue last month: https://github.com/dotnet/docs/pull/21977

Was that reverted for some reason?

DavidAnson:
Doesn't seem reverted and still renders correctly on GitHub (and real site), it seems: https://github.com/dotnet/docs/blob/master/docs/standard/base-types/regular-expression-example-scanning-for-hrefs.md

Where isn't this working?

Youssef1313:
@DavidAnson Thanks!

@adegeo It looks like you only need to merge master into your branch in the PR where markdownlint is failing

PR #22229, opened 2021-01-05
Update coding conventions wording and string samples

cartermp:
Some of the samples here won't compile, whoops! Also clarify some wording.


IEvangelist (approved):

PR #22240, opened 2021-01-06
Be explicit about methods

### LINKED ISSUE #22037: Examples and explanation needed
Moved this issue to .NET
Old issue: [https://github.com/MicrosoftDocs/visualstudio-docs/issues/5755](https://github.com/MicrosoftDocs/visualstudio-docs/issues/5755), opened by @HugoRoss

I think this whole rule needs some examples of how and why such ambiguities may occur and a good explanation why it contradicts everything that was preached the last 30 odd years. It was always the rule to convert to lower-case as there are more lower-case letters than upper-case ones and therefore some information might get lost when converting to upper-case.

Example: The German lower-case "ß" sign is converted to "SS" in all 8-bit codepages and one could not determine whether the word "MASSE" is the upper-case of "Maße" (dimensions) or "Masse" (ground/mass/matter), therefore to avoid such an ambiguity it was recommended to always convert to lower-case.

Please add some examples why in Unicode the opposite might happen and why converting to upper-case is no problem (e.g. if Unicode would define more upper-case letters than lower-case ones and it is guaranteed that every lower-case letter has an upper-case letter defined).

---
#### Document Details

* ID: b3894a92-71f2-1eb3-6384-54e0d034570a
* Version Independent ID: 06a39aec-f6d4-5ccb-152b-521e8af3c4bb
* Content: [CA1308: Normalize strings to uppercase (code analysis) - .NET](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1308)
* Content Source: [docs/fundamentals/code-analysis/quality-rules/ca1308.md](https://github.com/dotnet/docs/blob/master/docs/fundamentals/code-analysis/quality-rules/ca1308.md)
* GitHub Login: @gewarren

gewarren:


docs/standard/base-types/best-practices-strings.md
```diff
@@ -230,7 +230,7 @@ Default interpretation: <xref:System.StringComparison.CurrentCulture?displayProp
...
+The <xref:System.String.ToUpperInvariant%2A?displayProperty=nameWithType> and <xref:System.String.ToLowerInvariant%2A?displayProperty=nameWithType> methods are also available. <xref:System.String.ToUpperInvariant%2A> is the standard way to normalize case. (If you normalize to lowercase, some characters, for example, the Turkish letter İ, can't be converted back to their uppercase representation correctly.) Comparisons made using <xref:System.StringComparison.OrdinalIgnoreCase?displayProperty=nameWithType> are behaviorally the composition of two calls: calling <xref:System.String.ToUpperInvariant%2A> on both string arguments, and doing a comparison using <xref:System.StringComparison.Ordinal?displayProperty=nameWithType>.
```
IEvangelist, line 233:
```suggestion
The <xref:System.String.ToUpperInvariant%2A?displayProperty=nameWithType> and <xref:System.String.ToLowerInvariant%2A?displayProperty=nameWithType> methods are also available. <xref:System.String.ToUpperInvariant%2A> is the standard way to normalize case. If you normalize to lowercase, some characters, for example, the Turkish letter İ, can't be converted back to their uppercase representation correctly. Comparisons made using <xref:System.StringComparison.OrdinalIgnoreCase?displayProperty=nameWithType> are behaviorally the composition of two calls: calling <xref:System.String.ToUpperInvariant%2A> on both string arguments, and doing a comparison using <xref:System.StringComparison.Ordinal?displayProperty=nameWithType>.
```

I think the parenthesis here is unnecessary emphasis.

IEvangelist (approved):

docs/standard/base-types/best-practices-strings.md
gewarren: I tried that out (roundtripping İ by first converting to lowercase), and as long as you use the "Invariant" overload of the ToUpper/ToLower methods, it doesn't actually matter which one you use. So I removed my addition.

adegeo (approved):

PR #22253, opened 2021-01-06
Fix 22180

### LINKED ISSUE #22180: Incorrect example output on page.
The commented output of the second example of continuation with child tasks (detached) is incorrect.
The example shows the following output: (which seems to be just copied by mistake from the previous example)

// The example displays the similar output:
//     Running antecedent task 1...
//     Launching attached child tasks...
//     Finished launching detached child tasks...
//        Attached child task #2 running
//        Attached child task #3 running
//        Attached child task #1 running
//        Attached child task #4 running
//        Attached child task #5 running
//     Executing continuation of Task 1

**While the correct output should be:**

Running antecedent task 1...
Launching attached child tasks...
Finished launching detached child tasks...
Executing continuation of Task 1
   Attached child task #1 running
   Attached child task #4 running
   Attached child task #2 running
   Attached child task #3 running
   Attached child task #5 running

---
#### Document Details

* ID: be073b14-6d0b-60be-7dff-4674bcb6c674
* Version Independent ID: a5fa3e78-73de-3360-1253-00545a2b4ef2
* Content: [Chaining tasks using continuation tasks](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/chaining-tasks-by-using-continuation-tasks)
* Content Source: [docs/standard/parallel-programming/chaining-tasks-by-using-continuation-tasks.md](https://github.com/dotnet/docs/blob/master/docs/standard/parallel-programming/chaining-tasks-by-using-continuation-tasks.md)
* GitHub Login: @IEvangelist
* Microsoft Alias: **dapine**

IEvangelist:
## Summary

Verified the output was [incorrect here](https://dotnetfiddle.net/5WHAgn).

Fixes #22180

## Preview

✔️  [Chaining tasks using continuation tasks](https://review.docs.microsoft.com/en-us/dotnet/standard/parallel-programming/chaining-tasks-by-using-continuation-tasks?branch=pr-en-us-22253)

### DIFF
--- a/docs/standard/parallel-programming/snippets/cs/detached1.cs
+++ b/docs/standard/parallel-programming/snippets/cs/detached1.cs
@@ -38,9 +38,9 @@ public class DetachedExample
 //     Running antecedent task 1...
 //     Launching attached child tasks...
 //     Finished launching detached child tasks...
+//     Executing continuation of Task 1
+//        Attached child task #1 running
+//        Attached child task #5 running
 //        Attached child task #2 running
 //        Attached child task #3 running
-//        Attached child task #1 running
 //        Attached child task #4 running
-//        Attached child task #5 running
-//     Executing continuation of Task 1
--- a/docs/core/diagnostics/dotnet-counters.md
+++ b/docs/core/diagnostics/dotnet-counters.md
@@ -279,7 +279,7 @@ dotnet-counters monitor [-h|--help] [-p|--process-id] [-n|--name] [--diagnostic-
   > dotnet-counters list
 
   Showing well-known counters for .NET (Core) version 3.1 only. Specific processes may support additional counters.
-  System.Runtime              
+  System.Runtime
       cpu-usage                          The percent of process' CPU usage relative to all of the system CPU resources [0-100]
       working-set                        Amount of working set used by the process (MB)
       gc-heap-size                       Total heap size reported by the GC (MB)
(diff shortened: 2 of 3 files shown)


docs/standard/parallel-programming/snippets/cs/detached1.cs
```diff
@@ -38,9 +38,9 @@ public static async Task Main()
...
 //     Finished launching detached child tasks...
+//     Executing continuation of Task 1
+//        Attached child task #1 running
+//        Attached child task #5 running
 //        Attached child task #2 running
 //        Attached child task #3 running
-//        Attached child task #1 running
 //        Attached child task #4 running
-//        Attached child task #5 running
-//     Executing continuation of Task 1
```
Youssef1313, line 46: The output order is non-deterministic. Each run can produce a different output.
That's why the comment before the output says that it produces "**similar**" output.
Instead of changing the order, which doesn't solve the issue, I'd emphasize the fact that the order is non-deterministic, possibly with technical reasoning about why this is happening.

BillWagner:
@IEvangelist 

The comment by @Youssef1313 is correct. The order of tasks finishing is not guaranteed in this sample. This change isn't necessary, and the related issue can be closed.

docs/standard/parallel-programming/snippets/cs/detached1.cs
IEvangelist: > @IEvangelist David Pine FTE
> 
> The comment by @Youssef1313 is correct. The order of tasks finishing is not guaranteed in this sample. This change isn't necessary, and the related issue can be closed.

Yes, the numbered tasks are non-deterministic. This fix was to re-order the "Executing continuation of Task 1" which is guaranteed before the numbered tasks.  It was incorrectly at the bottom before - which is wrong.

BillWagner: Got it. Thanks.

BillWagner (approved):
This LGTM @IEvangelist 

You can :shipit: when ready.

PR #22262, opened 2021-01-07
change version to v1.0

GoEddie:
## Summary

1.0 was a major release and there was so much in it, it is worth including at least that version in the doc


luisquintanilla (approved):
Thanks @GoEddie

PR #22275, opened 2021-01-08
Elaborate on why WriteIndented is not suitable for production

perlun:
## Summary

Added a short sentence about why `WriteIntended` might be a bad idea for production usage.


tdykstra (approved):
Thanks, @perlun !

PR #22297, opened 2021-01-11
Update the guide to target .NET 5

vitek-karas:
3.0 is out of support.

The respective sample is updated to .NET 5 in https://github.com/dotnet/samples/pull/4096.

### DIFF
--- a/docs/core/tutorials/creating-app-with-plugin-support.md
+++ b/docs/core/tutorials/creating-app-with-plugin-support.md
@@ -17,7 +17,10 @@ This tutorial shows you how to create a custom <xref:System.Runtime.Loader.Assem
 
 ## Prerequisites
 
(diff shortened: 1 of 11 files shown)


vitek-karas:
@elinor-fung what's the maintenance mechanism for the snippets for this guide: https://github.com/dotnet/docs/tree/master/samples/snippets/core/tutorials/creating-app-with-plugin-support/csharp. That seems to be a full copy of the sample project. Is there a way to refer to the samples repo directly instead? Should I update the snippets with the latest from samples manually?

tdykstra:
> Should I update the snippets with the latest from samples manually?

Yes -- We used to do cross-repo references for code, but it causes problems whenever you update the doc or code, so we deliberately moved away from that.  The duplication is unfortunate but is better than the alternative.

tdykstra (approved):
Looks good, thanks!

PR #22309, opened 2021-01-12
Updated reference to CA1052

paulomorgado:
## Summary

Updated reference to CA1052 to use the correct description of the rule.


gewarren (approved):
@paulomorgado Thank you!

PR #22350, opened 2021-01-13
Replaced usage of ApplicationException with Exception

### LINKED ISSUE #21865: Samples show ApplicationException
The current sample has several instances where it throws `ApplicationException` however the current guidance on [ApplicationException](https://docs.microsoft.com/dotnet/api/system.applicationexception) is to avoid throwing `ApplicationException` directly.

> You should not throw an ApplicationException exception in your code, and you should not catch an ApplicationException exception unless you intend to re-throw the original exception.

---
#### Document Details

* ID: 5bec519d-b37f-92e2-2526-ad80dfcd82a1
* Version Independent ID: 9576bafa-e207-448d-c947-508d574c7d4e
* Content: [How to: Hook Up a Delegate Using Reflection](https://docs.microsoft.com/en-us/dotnet/framework/reflection-and-codedom/how-to-hook-up-a-delegate-using-reflection)
* Content Source: [docs/framework/reflection-and-codedom/how-to-hook-up-a-delegate-using-reflection.md](https://github.com/dotnet/docs/blob/master/docs/framework/reflection-and-codedom/how-to-hook-up-a-delegate-using-reflection.md)
* GitHub Login: @adegeo
* Microsoft Alias: **adegeo**

TheEskhaton:
## Summary

Changed the thrown exception type to be Exception instead of ApplicationException which gives off a warning as per #21865

Fixes #21865

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR/HookUpDelegate/cs/source.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR/HookUpDelegate/cs/source.cs
@@ -96,7 +96,7 @@ class Example
         //<Snippet9>
         Type returnType = GetDelegateReturnType(tDelegate);
         if (returnType != typeof(void))
-            throw new ApplicationException("Delegate has a return type.");
+            throw new Exception("Delegate has a return type.");
 
         DynamicMethod handler =
             new DynamicMethod("",
@@ -149,11 +149,11 @@ class Example
     private Type[] GetDelegateParameterTypes(Type d)
     {
         if (d.BaseType != typeof(MulticastDelegate))
-            throw new ApplicationException("Not a delegate.");
+            throw new Exception("Not a delegate.");
 
         MethodInfo invoke = d.GetMethod("Invoke");
         if (invoke == null)
-            throw new ApplicationException("Not a delegate.");
+            throw new Exception("Not a delegate.");
 
         ParameterInfo[] parameters = invoke.GetParameters();
         Type[] typeParameters = new Type[parameters.Length];
@@ -167,11 +167,11 @@ class Example
     private Type GetDelegateReturnType(Type d)
     {
         if (d.BaseType != typeof(MulticastDelegate))
-            throw new ApplicationException("Not a delegate.");
+            throw new Exception("Not a delegate.");
 
         MethodInfo invoke = d.GetMethod("Invoke");
         if (invoke == null)
-            throw new ApplicationException("Not a delegate.");
+            throw new Exception("Not a delegate.");
 
         return invoke.ReturnType;
     }
--- a/samples/snippets/csharp/VS_Snippets_CLR/HookUpDelegate/cs/source.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR/HookUpDelegate/cs/source.cs
@@ -96,7 +96,7 @@ class Example
         //<Snippet9>
         Type returnType = GetDelegateReturnType(tDelegate);
         if (returnType != typeof(void))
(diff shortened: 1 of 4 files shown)


TheEskhaton:
That's a pretty unexpected build error.

I assume the files in questions were never build with this particular build process so this was never figured out before?

Is there anything I can do about this?

samples/snippets/csharp/VS_Snippets_CLR/HookUpDelegate/cs/source.cs
```diff
@@ -96,7 +96,7 @@ private void HookUpDelegate()
         //<Snippet9>
         Type returnType = GetDelegateReturnType(tDelegate);
         if (returnType != typeof(void))
-            throw new ApplicationException("Delegate has a return type.");
+            throw new Exception("Delegate has a return type.");
```
Youssef1313, line 99: I think it's also a bad practice to throw the general exception type. Probably `InvalidOperationException` is more suitable?

Youssef1313:
> That's a pretty unexpected build error.
>
> I assume the files in question were never built with this particular build process so this was never figured out before?
>
> Is there anything I can do about this?



@TheEskhaton Yes, the CI build system for code files is pretty new in this repository. The fix is simply to add a `csproj` and `vbproj` files.

samples/snippets/csharp/VS_Snippets_CLR/HookUpDelegate/cs/source.cs
TheEskhaton: You are right, maybe even an `ArgumentException` since most of the checks are done on the argument, or properties on the argument?

BillWagner:
Thanks for putting this together @TheEskhaton 

I agree with both comments from @Youssef1313. Thanks for adding the csproj and vbproj files. That helps us improve.

Can you also accept the suggestion to change the type of exception being thrown?

Background on the CI process:

When our samples were migrated from an internal system, the build artifacts were not included. The samples previously used an internal build tool.  Because of that situation, our CI script ensures that code changes result in a project being built. That way, whenever code is changed, any missing projects are flagged for updates. It's a trigger that helps us improve.

BillWagner:
@TheEskhaton 

The most recent build failure indicates that these samples must be built on Windows. To fix this, add a file named *snippets.5000.json* in the project directory. The file should have the following contents:

```json
{
    "host": "visualstudio"
}
```

Then, our CI build will use Visual Studio instead of the .NET CLI.

samples/snippets/csharp/VS_Snippets_CLR/HookUpDelegate/cs/source.cs
BillWagner: I like the `ArgumentException` here.

TheEskhaton:
@BillWagner 

I've managed to fix the build issue by making the projects target net48 and added a reference to the System.Windows.Forms assembly into the project files. 

Is this sufficient or do you want to have them target net5.0 and include the _snippets.5000.json _ as you've mentioned?

The Exception type has been changed to `ArgumentException`.

BillWagner:
@TheEskhaton 

> Is this sufficient or do you want to have them target net5.0 and include the snippets.5000.json as you've mentioned?

That's fine as well. Thank you.

TheEskhaton:
Thanks! Then this is ready for a merge as far as I'm concerned.

BillWagner (approved):
thanks for doing all this @TheEskhaton 

We appreciate it!

I'll :shipit: now.

Again, thanks for helping us.

PR #22356, opened 2021-01-13
Clarify how timestamp and MVID work in determinism

jaredpar:
Had a few customer questions over the years about how the timestamp and MVID fields work in deterministic builds. This is my attempt to update and keep it rather succinct.

## Summary

Describe your changes here.

### DIFF
--- a/docs/csharp/language-reference/compiler-options/deterministic-compiler-option.md
+++ b/docs/csharp/language-reference/compiler-options/deterministic-compiler-option.md
@@ -21,7 +21,7 @@ Causes the compiler to produce an assembly whose byte-for-byte output is identic
 
 ## Remarks
 
-By default, compiler output from a given set of inputs is unique, since the compiler adds a timestamp and a GUID that is generated from random numbers. You use the `-deterministic` option to produce a *deterministic assembly*, one whose binary content is identical across compilations as long as the input remains the same.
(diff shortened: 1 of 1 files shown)


docs/csharp/language-reference/compiler-options/deterministic-compiler-option.md
```diff
@@ -21,7 +21,7 @@ Causes the compiler to produce an assembly whose byte-for-byte output is identic
 
 ## Remarks
 
-By default, compiler output from a given set of inputs is unique, since the compiler adds a timestamp and a GUID that is generated from random numbers. You use the `-deterministic` option to produce a *deterministic assembly*, one whose binary content is identical across compilations as long as the input remains the same.
+By default, compiler output from a given set of inputs is unique, since the compiler adds a timestamp and a MVID that is generated from random numbers. You use the `-deterministic` option to produce a *deterministic assembly*, one whose binary content is identical across compilations as long as the input remains the same. In such a build the timestamp and MVID fields will be replaced with values derived from a hash of all the compilation inputs.
```
Youssef1313, line 24:
```suggestion
By default, compiler output from a given set of inputs is unique, since the compiler adds a timestamp and a MVID that is generated from random numbers. You use the `-deterministic` option to produce a *deterministic assembly*, one whose binary content is identical across compilations as long as the input remains the same. In such a build, the timestamp and MVID fields will be replaced with values derived from a hash of all the compilation inputs.
```

BillWagner (approved):
Thanks @jaredpar (And thanks for the review @Youssef1313 )

After accepting this change, I'll :shipit:

PR #22374, opened 2021-01-15
Remove manual install of .NET from snippets5000

adegeo:


IEvangelist (approved):
We should make this a secret value so we can configure it without code changes:
`${{ secrets.DOTNET_DO_INSTALL }}`

PR #22432, opened 2021-01-20
Note about non-escaping code points

### LINKED ISSUE #22147: Can't serializ the '\u3000' when using with UnicodeRanges.All
The '\u3000' seems can't be serialized when using with UnicodeRanges.All.
BTW, I have tested the [UnicodeRanges.CjkSymbolsandPunctuation](https://docs.microsoft.com/en-us/dotnet/api/system.text.unicode.unicoderanges.cjksymbolsandpunctuation?view=net-5.0) and the '\u3000' can't be serialized too.
QQ截图20201224162108

---
#### Document Details

* ID: fbfcc376-0c18-5aab-f6b8-b2c6ef282acc
* Version Independent ID: 79bcd93d-81f8-5f02-3958-d364606265cd
* Content: [How to customize character encoding with System.Text.Json](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-character-encoding)
* Content Source: [docs/standard/serialization/system-text-json-character-encoding.md](https://github.com/dotnet/docs/blob/master/docs/standard/serialization/system-text-json-character-encoding.md)
* GitHub Login: @tdykstra

tdykstra:
Fixes #22147

[Internal review URL](https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-character-encoding?branch=pr-en-us-22432#block-lists)

### DIFF
--- a/docs/standard/serialization/system-text-json-character-encoding.md
+++ b/docs/standard/serialization/system-text-json-character-encoding.md
@@ -1,7 +1,7 @@
 ---
 title: How to customize character encoding with System.Text.Json
 description: "Learn how to customize character encoding while serializing to and deserializing from JSON in .NET."
-ms.date: 01/21/2021
+ms.date: 01/22/2021
 no-loc: [System.Text.Json, Newtonsoft.Json]
 helpviewer_keywords:
   - "JSON serialization"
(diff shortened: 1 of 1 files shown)


GrabYourPitchforks:
It's not just whitespace. There are certain code points across various categories that are always escaped, even if you ask politely not to escape them. For example, the ampersand (`'&'`) character will always be escaped by the HTML encoder, even though the ampersand falls under the BasicLatin block and all the escapers are initialized with BasicLatin by default.

In order for a code point to pass through unescaped, __all__ of the following criteria must be met.

1. The encoder must have been initialized with a `TextEncoderSettings` or `UnicodeRange[]` which allows the code point. Example: an encoder initialized with only BasicLatin will disallow `'é'`, which is not in the BasicLatin block.

2. The code point must not be in the global block list. We've never really documented this global block list because its members are subject to change. But it includes things like private-use characters, control characters, undefined code points, and certain members of other Unicode categories. (This is where the U+3000 IDEOGRAPHIC SPACE is being rejected.)

3. The code point must not be in the encoder's own block list. For example: the HTML encoder blocks `'<'` and `'&'`. The JSON encoder blocks `'\'`. The URL encoder blocks `'%'`. And so on.

Is there a good way to describe this within the document?

GrabYourPitchforks:
Latest changes LGTM. Only remaining feedback is that the block list is an implementation detail. It has changed in every release of netcore. Users shouldn't take a dependency on a character being a member of (or not being a member of) the global block list.

gewarren (approved):

tdykstra:
Thanks for the review, @gewarren !

PR #22436, opened 2021-01-20
Recategorize ms.topic for standard/basetypes

tdykstra:
Content performance project for Q3

Recategorizing default from for standard/base-types folder from conceptual to how-to, overriding non-how-to docs at file level.


BillWagner (approved):
This all LGTM @tdykstra 

I like the way you've made decisions between conceptual and reference. This will help us clarify those types.

adegeo (approved):
LGTM too.

PR #22445, opened 2021-01-22
5.0 version of inferred types converter

### LINKED ISSUE #22401: Code-Snippet of "Deserialize inferred types" does not work anymore
Since *dotnet5* your code-snippet of `ObjectToInferredTypesConverter`-class does not work in some cases. In the past I used this converter for type inference of untyped Json-elements, like in the following class:

```csharp
class MyObjectToSerialize
{
   public int AnyTypedProperty { get; set; }
   public object UntypedJsonProperty { get; set; }  // e.g. "UntypedJsonProperty":{"A":true,"B":42}
}
```

Now if I serialize an object, the `Write`-method of the `ObjectToInferredTypesConverter` is called for any `JsonElement`s. Therefore it is no longer possible to have no implementation (or just throw an exception as in your code-snippet) for the `Write`-method.

```csharp
 var bytes = JsonSerializer.SerializeToUtf8Bytes(myObject, myJsonSerializerOptions);
```

I understand that your example is only intended for the purpose of deserialization, but this also worked for the purpose of serialization quite fine in *dotnet core 3.1*. I think that serialization of `JsonElement`s in older `System.Text.Json`-implementations was done using the `ObjectDefaultConverter`?

---
#### Document Details

* Content: [How to write custom converters for JSON serialization - .NET](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?pivots=dotnet-5-0#deserialize-inferred-types-to-object-properties)
* GitHub Login: @tdykstra

tdykstra:
Fixes #22401

[Internal review URL](https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?branch=pr-en-us-22445&pivots=dotnet-5-0#deserialize-inferred-types-to-object-properties)

### DIFF
--- a/docs/standard/serialization/system-text-json-converters-how-to.md
+++ b/docs/standard/serialization/system-text-json-converters-how-to.md
@@ -1,7 +1,7 @@
 ---
 title: "How to write custom converters for JSON serialization - .NET"
 description: "Learn how to create custom converters for the JSON serialization classes that are provided in the System.Text.Json namespace."
-ms.date: 12/14/2020
+ms.date: 01/21/2021
 no-loc: [System.Text.Json, Newtonsoft.Json]
 zone_pivot_groups: dotnet-version
 helpviewer_keywords: 
@@ -236,6 +236,16 @@ For scenarios that require type inference, the following code shows a custom con
 * Strings to `string`
 * Everything else to `JsonElement`
 
+::: zone pivot="dotnet-core-3-1"
+
+:::code language="csharp" source="snippets/system-text-json-how-to-5-0/csharp/CustomConverterInferredTypesToObject.cs":::
+
+The example shows the converter code and a `WeatherForecast` class with `object` properties. The `Main` method deserializes a JSON string into a `WeatherForecast` instance, first without using the converter, and then using the converter. The console output shows that without the converter the run time type for the `Date` property is `JsonElement`; with the converter, the run time type is `DateTime`.
+
+:::zone-end
+
+::: zone pivot="dotnet-core-3-1"
+
 :::code language="csharp" source="snippets/system-text-json-how-to/csharp/ObjectToInferredTypesConverter.cs":::
 
 The following code registers the converter:
@@ -255,6 +265,8 @@ The following example of JSON to deserialize contains values that will be deseri
   "Summary": "Hot",
 }
 ```
+
+:::zone-end
 
 Without the custom converter, deserialization puts a `JsonElement` in each property.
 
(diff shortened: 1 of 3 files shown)


docs/standard/serialization/system-text-json-converters-how-to.md
```diff
@@ -236,6 +236,16 @@ For scenarios that require type inference, the following code shows a custom con
...
+::: zone pivot="dotnet-5-0"
+
+:::code language="csharp" source="snippets/system-text-json-how-to-5-0/csharp/CustomConverterInferredTypesToObject.cs":::
+
+The example shows the converter code and a `WeatherForecast` class with `object` properties. The `Main` method deserializes a JSON string into a `WeatherForecast` instance, first without using the converter, and then using the converter. The console output shows that without the converter the run time type for the `Date` property is `JsonElement`; with the converter, the run time type is `DateTime`.
+
+:::zone-end
+
+::: zone pivot="dotnet-core-3-1"
+
```
layomia, line 248: Is there any reason for the difference in writing style between the 3.1 and 5.0 pivots? In 3.1, we break up the code sample and our explanation into chunks:  "The following code registers the converter:", "Here's an example with `object` properties:", and so on; whereas in 5.0, there's just one mass of code, followed by some text. 

It's not a critical crucial difference, just something that struck me as I compared the two pivots. We could make them consistent, or leave as-is (up to you).

docs/standard/serialization/system-text-json-converters-how-to.md
```diff
@@ -266,12 +266,12 @@ The following example of JSON to deserialize contains values that will be deseri
 }
 ```
 
-:::zone-end
-
 Without the custom converter, deserialization puts a `JsonElement` in each property.
 
 The [unit tests folder](https://github.com/dotnet/runtime/blob/81bf79fd9aa75305e55abe2f7e9ef3f60624a3a1/src/libraries/System.Text.Json/tests/Serialization/) in the `System.Text.Json.Serialization` namespace has more examples of custom converters that handle deserialization to `object` properties.
 
+:::zone-end
```
layomia, line 273: This removes the link to the tests in the 5.0 pivots. Why do we have it this way?

docs/standard/serialization/system-text-json-converters-how-to.md
tdykstra: After publishing the 3.1 STJ docs there were some comments complaining about lack of context for the code snippets, so I took a different approach with the snippets created for the 5.0 updates -- each is a complete program that you can copy/paste into a project to run the code. This approach results in longer code examples but if it proves to be more helpful I can gradually change existing snippets created for 3.1 to follow the same pattern. Since the how-to was split up from one big doc into many smaller ones, it's more practical to have bigger code snippets.

docs/standard/serialization/system-text-json-converters-how-to.md
tdykstra: The permalink points to the code as it was for 3.1, I need to make a new link for 5.0 and copy the paragraph with the link.

luisquintanilla (approved):
LGTM

PR #22446, opened 2021-01-22
Do not capitalize keyword

syohex:
## Summary

Do not capitalize keyword. `Internal` -> `internal`


BillWagner (approved):
Thanks for spotting this and making the fix @syohex We appreciate it!

Thanks again.

PR #22466, opened 2021-01-23
Unexpected error caused by additional argument.

jjgorton:
Currently this snippet (:::code language="csharp" source="./snippets/object-oriented-programming/Program.cs" ID="TestLineOfCredit":::)  will result in an unexpected error.  These changes allow for that error to remain for the reader's benefit while also pointing out the solution to arrive at the "expected" error.

Alternatively, the above code snippet could simply be edited to remove the 3rd argument. (see: https://github.com/dotnet/docs/pull/22467)

PR #22467, opened 2021-01-23
Unexpected error caused by additional argument.

jjgorton:
## Summary

Describe your changes here.

This fixes an issue where the tutorial purposefully allows for an error for educational reasons.  However the snippet referenced actually results in another error other than the expected one.  

Simply removing the 3rd argument for the constructor at this point in the tutorial allows the reader to continue smoothely. 

IMPORTANT:  this is an alternative solution to https://github.com/dotnet/docs/pull/22466


BillWagner (approved):
Thanks for submitting both possible solutions @jjgorton 

I do like this one better, and I'll :shipit: now.

Issue #20787, opened 2020-09-25
Suggestion for clarification w.r.t using a discard to suppress an exception


RexJaeschke:
Hi, I'm Rex Jaeschke, editor of the C# Standard, under contract to MS. I was confused by what this meant, so I asked Mads Torgersen of the C# Team @ MS to help clarify. Here's what he sent to me:

"No, discards have no exception-suppressing capabilities! Here’s what happens in the example:

- We Task.Run a lambda that ends by throwing an exception
- The result of the Task is therefore the exception
- If you await the exception, then that throws the exception in your code
- To deliberately avoid that exception, you might want to simply drop the Task on the floor and not await it
- But if you do, the compiler helpfully warns you saying hey, didn’t you forget to deal with this task?
- To show the compiler that you’re dealing with the Task, all you have to do is use it somehow, e.g. by assigning it to a variable.
- But now you have this dummy variable containing a Task that you specifically don’t want to touch again.
- Better to use a discard! Now you can’t accidentally touch the Task later, and you don’t have the variable name cluttering up your declaration space."

You might want to borrow some of his words to help improve this page.

Rex
rex@RexJaeschke.com

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 8047b417-55af-c6df-9333-f1e1c428e594
* Version Independent ID: 8bcc16bb-c2d8-0b6d-af20-6958964efe9e
* Content: [Discards - C# Guide](https://docs.microsoft.com/en-us/dotnet/csharp/discards#a-standalone-discard)
* Content Source: [docs/csharp/discards.md](https://github.com/dotnet/docs/blob/master/docs/csharp/discards.md)
* Product: **dotnet-csharp**
* Technology: **csharp-fundamentals**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Thanks @RexJaeschke 
I'll add this to our list to update.

BillWagner:
When this is fixed, explain that the behavior exhibited is different when a debugger is attached. (With the debugger, the debugger forces the app to break when the exception is thrown and not caught. It will be discarded without a debugger attached).

/cc @kendaleiv

PR #22484, opened 2021-01-25
Increase iterations in example

kendaleiv:
Prior to this change I was able to unexpectedly encounter the `InvalidOperationException` because the loop completed.

This example is viewable at https://docs.microsoft.com/en-us/dotnet/csharp/discards#a-standalone-discard.

### DIFF
--- a/samples/snippets/csharp/programming-guide/discards/standalone-discard1.cs
+++ b/samples/snippets/csharp/programming-guide/discards/standalone-discard1.cs
@@ -12,7 +12,7 @@ public class Example
    {
       Console.WriteLine("About to launch a task...");
       _ = Task.Run(() => { var iterations = 0;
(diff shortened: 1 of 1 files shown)


BillWagner:
Thanks for updating this @kendaleiv 

I think I want to close this, because the fix would be a better explanation on what the discard does for this sample.  See #20787 for a discussion on what the discard adds to this code.

What I would recommend instead (which will address that issue) is to update the explanation with that explanation.

I'm curious where you observed the exception. Were you running your sample with a debugger attached?

kendaleiv:
Ran with Visual Studio 2019 with the debugger attached, it encountered the `InvalidOperationException` before the 5 second wait was done.

BillWagner:
Thanks @kendaleiv 

I'll keep this closed, but add a note that I should explain that behavior when I fix #20787.

PR #22502, opened 2021-01-26
Update toc.yml (Typo "Azure Blob storage"→"Azure Blob Storage")

hyoshioka0128:
## Summary
Typo "Azure Blob storage"→"Azure Blob Storage"
https://docs.microsoft.com/en-us/dotnet/fsharp/using-fsharp-on-azure/blob-storage


cartermp (approved):
Thanks!

PR #22509, opened 2021-01-26
Typo in docker image version

jonnychipz:
## Summary

I believe that there is a typo error in the docker image version number as identified version 5.01 does not exist and this should read 5.0 as modified.


nishanil (approved):
Great catch @jonnychipz. Thank you. LGTM

PR #22514, opened 2021-01-26
Restore section that was deleted

tdykstra:
Follow-up to #21108

[Internal review URL](https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-ignore-properties?branch=pr-en-us-22514&pivots=dotnet-core-3-1#ignore-null-when-deserializing)

### DIFF
--- a/docs/standard/serialization/system-text-json-ignore-properties.md
+++ b/docs/standard/serialization/system-text-json-ignore-properties.md
@@ -133,7 +133,7 @@ By default, if a property in JSON is null, the corresponding property in the tar
 
 For example, suppose the following code represents your target object:
 
-[!code-csharp[](snippets/system-text-json-how-to/csharp/WeatherForecast.cs?name=SnippetWFWithDefault)]
+[!code-csharp[](snippets/system-text-json-how-to/csharp/WeatherForecast.cs?name=WFWithDefault)]
 
 And suppose the following JSON is deserialized:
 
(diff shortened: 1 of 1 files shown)


docs/standard/serialization/system-text-json-ignore-properties.md
```diff
@@ -30,6 +30,8 @@ When serializing C# objects to JavaScript Object Notation (JSON), by default, al
 * [All null-value properties](#ignore-all-null-value-properties)
 ::: zone-end
 
+This article also includes a section about how to [ignore null when deserializing](#ignore-null-when-deserializing)
```
tdykstra, line 33:
```suggestion
This article also includes a section about how to [ignore null when deserializing](#ignore-null-when-deserializing).
```

Jozkee:
I wonder if this section was intentionally removed given that `JsonSerializerOptions.IgnoreNullValues` will become obsolete. 
cc @layomia.

tdykstra:
> I think I found the reason it was removed, see https://github.com/dotnet/docs/pull/21108#discussion_r517720104. 

Yes that's it, thanks. I'll leave this PR open until your question about obsoletion is resolved, but I'll change it back to draft.

layomia:
@jozkee hmm I don't think the edge case of avoiding `null` overwriting instantiated properties on deserialization for the ignore-reference feature is worth keeping the `IgnoreNullProperty` alive (non-obsolete). 

I just don't like the idea of having two advertised ways of doing the same thing (`IgnoreNullValues` and `DefaultIgnoreCondition`). If it is important to support the above scenario, then we can consider extending `JsonIgnoreCondition` to include values that affect deserialization. 

---

@tdykstra, what motivated opening this PR? Just want to be sure I'm not missing any info, otherwise, I say we close this PR.

tdykstra:
I'm fine with closing it.  @jozkee asked me where the section went, and I forgot the story behind the deletion so I investigated, found it was removed in the 5.0 PR, thought I had removed it accidentally, and rushed (rather too quickly, it turns out) to put it back.

PR #22529, opened 2021-01-27
Update net-container-os-targets.md

cwilhit:
## Summary

We should not be encouraging users to incorporate arbitrary system .dlls from host machines during Windows container image build process. It opens security and patch-ability risks for the user, and non-redistributable .dlls should not be redistributed.


nishanil (approved):
Thanks for the update.

splitt3r:
What is the preferred way to solve the problem now? What about replacing the link instead of simply removing it? Is there any documentation to point to instead?

cwilhit:
@splitt3r I'd encourage you to look at the Windows container [docs](https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/container-base-images#guidelines). For convenience, pulling from the guidelines section, we recommend:

> Is the Windows Server Core container image missing a dependency your app needs? If the answer to this question is yes, you should attempt to target Windows. This image is much larger than the other base images, but it carries many of the core Windows libraries (such as the GDI library).

PR #22537, opened 2021-01-27
Address issues with discards article

### LINKED ISSUE #13239: Stand alone discard example seems confusing to me
Hi,

I'm not able to understand the whole point of this code example for the standalone discard feature: 

```
using System;
using System.Threading.Tasks;

public class Example
{
   public static void Main()
   {
      ExecuteAsyncMethods().Wait();
   }

   private static async Task ExecuteAsyncMethods()
   {    
      Console.WriteLine("About to launch a task...");
      _ = Task.Run(() => { var iterations = 0;  
                           for (int ctr = 0; ctr < int.MaxValue; ctr++)
                              iterations++;
                           Console.WriteLine("Completed looping operation...");
                           throw new InvalidOperationException();
                         });
      await Task.Delay(5000);                        
      Console.WriteLine("Exiting after 5 second delay");
   }
}
// The example displays output like the following:
//       About to launch a task...
//       Completed looping operation...
//       Exiting after 5 second delay
```

I can't understand the usefulness of the discard assignment, in the sense that based on my understanding the code behavior doesn't change even if the discard assignment is removed. The only difference is that, when the discard assignment is removed, a compiler warning pops up because inside the async function `ExecuteAsyncMethods` there is a task which is not awaited. **Apart from that, I can't see any difference in the runtime behavior**. 

The following sentence which is used to explain the example it's even more confusing: 

> The following example uses a standalone discard to ignore the Task object returned by an asynchronous operation. **This has the effect of suppressing the exception that the operation throws as it is about to complete**.

Based on my knowledge of tasks and exceptions the exception raised inside `Task.Run` is ignored by the CLR because the task is not awaited in any way, so the exception is abandoned and the process doesn't crash. This is **not** related with the assignment to a discard expression, instead **it's the default behavior of tasks since .NET 4.5**. 

---
#### Document Details

* ID: 8047b417-55af-c6df-9333-f1e1c428e594
* Version Independent ID: 8bcc16bb-c2d8-0b6d-af20-6958964efe9e
* Content: [Discards - C# Guide](https://docs.microsoft.com/en-us/dotnet/csharp/discards#feedback)
* Content Source: [docs/csharp/discards.md](https://github.com/dotnet/docs/blob/master/docs/csharp/discards.md)
* Product: **dotnet-csharp**
* GitHub Login: @rpetrusha
* Microsoft Alias: **ronpet**

### LINKED ISSUE #17198: Add argument validation use case to discard documentation
I think it would be helpful to add another use case to the standalone discard documentation for simplified argument validation.

From:
```csharp
if (arg is null)
{
    throw new ArgumentNullException(nameof(arg));
}
```

To:
```csharp
_ = arg ?? throw new ArgumentNullException(nameof(arg));
```

---
#### Document Details

* ID: 8047b417-55af-c6df-9333-f1e1c428e594
* Version Independent ID: 8bcc16bb-c2d8-0b6d-af20-6958964efe9e
* Content: [Discards - C# Guide](https://docs.microsoft.com/en-us/dotnet/csharp/discards)
* Content Source: [docs/csharp/discards.md](https://github.com/dotnet/docs/blob/master/docs/csharp/discards.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

### LINKED ISSUE #19880: Missing example section in Discards article
The section on pattern matching mentions two uses of the discard, but only the use as a default condition in `switch` is actually presented with the usage with `is` missing.

---
#### Document Details

* ID: 8047b417-55af-c6df-9333-f1e1c428e594
* Version Independent ID: 8bcc16bb-c2d8-0b6d-af20-6958964efe9e
* Content: [Discards - C# Guide](https://docs.microsoft.com/en-us/dotnet/csharp/discards)
* Content Source: [docs/csharp/discards.md](https://github.com/dotnet/docs/blob/master/docs/csharp/discards.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

### LINKED ISSUE #20787: Suggestion for clarification w.r.t using a discard to suppress an exception

BillWagner:
Fixes #13239
Fixes #20787

Clarify the effects of assigning the result of `Task.Run` to a discard. 

Fixes #17198

Add the null check as a s;tandalone example for using a discard.

Fixes #19880

Explain that using a discard with `is` would be rare, and update the H2 to match.

Finally, do an edit pass.


docs/csharp/discards.md
```diff
@@ -6,74 +6,75 @@ ms.date: 09/22/2020
...
 
-In C# 7.0 and later, discards are supported in assignments in the following contexts:
-
-- Tuple and object [deconstruction](deconstruct.md).
-- Pattern matching with [is](language-reference/keywords/is.md) and [switch](language-reference/keywords/switch.md).
-- Calls to methods with `out` parameters.
-- A standalone `_` when no `_` is in scope.
-
-Beginning with C# 9.0, you can use discards to specify unused input parameters of a lambda expression. For more information, see the [Input parameters of a lambda expression](language-reference/operators/lambda-expressions.md#input-parameters-of-a-lambda-expression) section of the [Lambda expressions](language-reference/operators/lambda-expressions.md) article.
+Discards are available starting with C# 7.0. Beginning with C# 9.0, you can use discards to specify unused input parameters of a lambda expression. For more information, see the [Input parameters of a lambda expression](language-reference/operators/lambda-expressions.md#input-parameters-of-a-lambda-expression) section of the [Lambda expressions](language-reference/operators/lambda-expressions.md) article.
```
IEvangelist, line 19:
```suggestion
Beginning with C# 9.0, you can use discards to specify unused input parameters of a lambda expression. For more information, see the [Input parameters of a lambda expression](language-reference/operators/lambda-expressions.md#input-parameters-of-a-lambda-expression) section of the [Lambda expressions](language-reference/operators/lambda-expressions.md) article.
```

Remove redundant statement, as it was in the previous section.

docs/csharp/snippets/discards/discard-pattern2.cs
```diff
@@ -0,0 +1,42 @@
...
+                    Console.WriteLine("Some object type without format information");
+                    break;
+            }
+        }
+        // The example displays the following output:
+        //    en-US object
+        //    System.Globalization.DateTimeFormatInfo object
+        //    System.Globalization.NumberFormatInfo object
+        //    Some object type without format information
+        //    A null object reference: Its use could result in a NullReferenceException
```
IEvangelist, line 39:
```suggestion
        static void ProvidesFormatInfo(object obj) =>
            Console.WriteLine(obj switch
            {
                IFormatProvider fmt => $"{fmt.GetType()} object",
                null => "A null object reference: Its use could result in a NullReferenceException",
                _ => "Some object type without format information"
            });
        // The example displays the following output:
        //    System.Globalization.CultureInfo object
        //    System.Globalization.DateTimeFormatInfo object
        //    System.Globalization.NumberFormatInfo object
        //    Some object type without format information
        //    A null object reference: Its use could result in a NullReferenceException
```

Not sure if this is better, but when I saw this it made me think of a switch expression use case. Also, the example output doesn't seem correct. [See here](https://dotnetfiddle.net/h7GTEj).

docs/csharp/snippets/discards/standalone-discard1.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    public static async Task TaskDiscard()
+    {
+        await ExecuteAsyncMethods();
+    }
+
+    // <ArgNullCheck>
+    public static void Method(string arg)
+    {
+        _ = arg ?? throw new ArgumentNullException(paramName: nameof(arg), message: "arg can't be null");
```
IEvangelist, line 15: I really like this!

IEvangelist (approved):

docs/csharp/snippets/discards/discard-pattern2.cs
BillWagner: Hmm. Output matches for me.

PR #22543, opened 2021-01-28
fix issues with EventCounter sample due to race in EventListener

### LINKED ISSUE #21506: The SimpleEventListener example has a race?
The `SimpleEventListener` example doesn't appear to work for me. It seems like at construction, the `OnEventSourceCreated` method can get called prior to `_intervalSec` being assigned, which sets the `EventCounterIntervalSec` argument to 0 instead of the intended value.

---
#### Document Details

* ID: 05179b70-8fe7-5b1a-ca2a-4ba0297f90c8
* Version Independent ID: 19d45d3b-026c-33e6-99b9-b50458f5bcac
* Content: [EventCounters in .NET Core](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/event-counters)
* Content Source: [docs/core/diagnostics/event-counters.md](https://github.com/dotnet/docs/blob/master/docs/core/diagnostics/event-counters.md)
* Product: **dotnet-core**
* GitHub Login: @sywhang
* Microsoft Alias: **suwhang**

sywhang:
## Summary

EventListener's OnEventWritten callback can race with the constructor of a subclass of EventListener because the base constructor registers the instance as callable and the child constructor runs after the base ctor. This code snippet relies on the ctor have to finished running, which makes it vulnerable to the described race condition causing a NRE. 

This fixes the race condition in the sample code.

Fixes #21506.

### DIFF
--- a/docs/core/diagnostics/event-counters.md
+++ b/docs/core/diagnostics/event-counters.md
@@ -191,11 +191,11 @@ You can consume the counter values via the <xref:System.Diagnostics.Tracing.Even
 
 First, the <xref:System.Diagnostics.Tracing.EventSource> that produces the counter value needs to be enabled. Override the <xref:System.Diagnostics.Tracing.EventListener.OnEventSourceCreated%2A?displayProperty=nameWithType> method to get a notification when an <xref:System.Diagnostics.Tracing.EventSource> is created, and if this is the correct <xref:System.Diagnostics.Tracing.EventSource> with your EventCounters, then you can call <xref:System.Diagnostics.Tracing.EventListener.EnableEvents%2A?displayProperty=nameWithType> on it. Here is an example override:
 
-:::code language="csharp" source="snippets/EventCounters/SimpleEventListener.cs" range="16-27":::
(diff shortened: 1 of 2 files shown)


tdykstra (changes requested):
The new code is now out of sync with the text that introduces lines 16-27 of the current code:

> Override the EventListener.OnEventSourceCreated method to get a notification when an EventSource is created, and if this is the correct EventSource with your EventCounters, then you can call EventListener.EnableEvents on it. Here is an example override:

The following line and text that introduces it needs to be updated:

```
:::code language="csharp" source="snippets/EventCounters/SimpleEventListener.cs" range="16-27":::
```

cc @IEvangelist

sywhang:
Thanks @tdykstra . Thinking about it, I think it would be better to just show a simpler example where the `intervalSec` argument is not configurable via the ctor, since this doc is focused on EventCounters and it wouldn't be immediately obvious why a race condition exists in EventListener. I'll be following up with a PR that updates the docs on EventListener that details this behavior since that'd be a better place to document this behavior and show example.

tdykstra (approved):
Looks good, thanks!

PR #22554, opened 2021-01-28
combine redundant articles

BillWagner:
The get started and the tour of C# have a lot of duplicated content. Combine them.

-Add redirects (and replace one or two redirects from a retired destination)
- Update inbound links.


.openpublishing.redirection.json
```diff
@@ -787,7 +787,7 @@
         },
         {
             "source_path": "docs/csharp/getting-started/additional-resources.md",
-            "redirect_url": "/dotnet/csharp/getting-started/index"
+            "redirect_url": "/dotnet/csharp/tour-of-csharp/index"
```
adegeo, line 790: nit: The others aren't using /index on the redirect while this one is.

docs/csharp/misc/sorry-we-don-t-have-specifics-on-this-csharp-error.md
```diff
@@ -748,7 +749,7 @@ We're sorry, but we don't have any specific information about this error. You ca
 Provides a QA website for questions about and issues with Visual C#.  
 
  Also see:  
- [Additional Resources for Visual C# Programmers](../getting-started/index.md)  
+ [Additional Resources for Visual C# Programmers](../tour-of-csharp/index.md)
```
adegeo, line 752: nit: This section seems odd. Since you're editing the link, might as well as move it 3 lines down to the **See also** section

adegeo (approved):

docs/csharp/tour-of-csharp/types.md
```diff
@@ -5,6 +5,8 @@ ms.date: 08/06/2020
 ---
 # Types and members
 
+As an object-oriented language, C# supports the concepts of encapsulation, inheritance, and polymorphism. A class may inherit directly from one parent class, but it may implement any number of interfaces. Methods that override virtual methods in a parent class require the `override` keyword as a way to avoid accidental redefinition. In C#, a struct is like a lightweight class; it's a stack-allocated type that can implement interfaces but doesn't support inheritance. C# also provides records, which are class types whose purpose is primarily storing data values.
```
tdykstra, line 8:
```suggestion
As an object-oriented language, C# supports the concepts of encapsulation, inheritance, and polymorphism. A class may inherit directly from one parent class, and it may implement any number of interfaces. Methods that override virtual methods in a parent class require the `override` keyword as a way to avoid accidental redefinition. In C#, a struct is like a lightweight class; it's a stack-allocated type that can implement interfaces but doesn't support inheritance. C# also provides records, which are class types whose purpose is primarily storing data values.
```

tdykstra (approved):
This looks great, I just noted a couple opportunities to improve clarity and reduce redundancy.

PR #22565, opened 2021-01-29
Update recursive-functions-the-rec-keyword.md

laenas:
## Summary

Typos in the code samples for non-recursive fibonacci causes an off-by-one issue in the fibonacci sequence.


cartermp (approved):
Thanks!

PR #22580, opened 2021-01-30
Updating MessageBox to MsgBox

Daniel-kenobi:
in this session you should be taught to rename the "MessageBox" function to "MsgBox", but in the C # version this did not happen. Now it's fixed!


BillWagner (approved):
Thanks for making this change @Daniel-kenobi 

PR #22590, opened 2021-01-31
Cross platform P/Invoke

saul:
## Summary

Adds an article detailing the order in which DLLs are searched on Windows, Linux and macOS. I haven't seen any pre-existing docs for this and it's very useful information when writing cross-platform P/Invoke code. See [dllimport.cpp](https://github.com/dotnet/runtime/blob/2c4e9e0e05a6d53c451b6ff9bb1eeb384daf3414/src/coreclr/vm/dllimport.cpp#L5996-L6094) for runtime source.


BillWagner:
Thanks @saul 

While I wait for @jkoritzinsky to weigh in, can you add a project file (*.csproj) where you added the new file? You'll need to specify a single entry point, so the compiler doesn't complain about multiple `Main` methods. That way, we ensure that everything compiles cleanly.

(If it's not too much to ask, these should be separate projects as they are all standalone programs, but I realize that's a lot to ask in expanding this PR).

jkoritzinsky (approved):

BillWagner:
Thanks again @saul 

I'll :shipit: now.

PR #22600, opened 2021-02-01
add solution files

BillWagner:
If a solution file changes, we should build some code.

Also, change the message from ".NET Core" to ".NET"


IEvangelist (approved):

Note, @adegeo added solutions to the [*.gitignore* file](https://github.com/dotnet/docs/blob/master/.gitignore#L46). So to make changes the user will have to force updates.

PR #22602, opened 2021-02-01
Parameter name matching for parameterized ctors

### LINKED ISSUE #22488: [JsonSerializer] Provide doc showing that serialization attributes are honored when deserializing property/field via constructor parameter
I am trying to convert from NewtonSoft over to System.Text.Json and all went well until I hit this.

I have many classes like this (because the JSON I'm parsing isn't the best) where I want to "rename" the property in the JSON to something meaningful in the C# code. (This class is pretty much unmodified from the conversion):

```c#
    /// <summary>
    /// Incorporates the cloudiness of the weather.
    /// </summary>
    internal class Cloudiness
    {
        /// <summary>
        /// Initializes a new instance of the <see cref="Cloudiness"/> class.
        /// </summary>
        /// <param name="all ">The percentage of cloud cover.</param>
        [JsonConstructor]
        public Cloudiness(double all)
        {
            this.Precentage = all;
        }

        /// <summary>
        /// Gets the percentage of cloud cover.
        /// </summary>
        [JsonPropertyName(name: "all")]
        public double Precentage { get; }
    }
```

Except this simple test fails:
```c#
var testString = JsonSerializer.Serialize(new Cloudiness(2));
var testModel = JsonSerializer.Deserialize<Cloudiness>(testString);
```
Each parameter in constructor 'Void .ctor(Double)' on type 'Cloudiness' must bind to an object property or field on deserialization. Each parameter name must match with a property or field on the object. The match can be case-insensitive.

Is there a way to have System.Text.Json map the constructor arguments according to the desired "rename" of the property via the JsonPropertyName attribute? (I really don't want to have to write a bunch of converters for this :()

---
#### Document Details

* ID: 31f52750-1f3a-3cba-9790-bf36a5ae30a5
* Version Independent ID: cf9fe1d5-1079-025c-73cb-668cc9cbd520
* Content: [How to use immutable types and non-public accessors with System.Text.Json](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-immutability)
* Content Source: [docs/standard/serialization/system-text-json-immutability.md](https://github.com/dotnet/docs/blob/master/docs/standard/serialization/system-text-json-immutability.md)
* GitHub Login: @tdykstra

tdykstra:
Fixes #22488

[Internal review URL](https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-immutability?branch=pr-en-us-22602)

### DIFF
--- a/docs/standard/serialization/system-text-json-immutability.md
+++ b/docs/standard/serialization/system-text-json-immutability.md
@@ -24,7 +24,7 @@ This article shows how to use immutable types, parameterized constructors, and n
 
 The parameter names of a parameterized constructor must match the property names. Matching is case-insensitive, and the constructor parameter must match the actual property name even if you use [[JsonPropertyName]](xref:System.Text.Json.Serialization.JsonPropertyNameAttribute) to rename a property. In the following example, the name for the `TemperatureC` property is changed to `celsius` in the JSON, but the constructor parameter is still named `temperatureC`:
 
-:::code language="csharp" source="snippets/system-text-json-how-to-5-0/csharp/ImmutableTypesCtorParms.cs" highlight="13":::
+:::code language="csharp" source="snippets/system-text-json-how-to-5-0/csharp/ImmutableTypesCtorParms.cs" highlight="10,14-16":::
 
 Records in C# 9 are also supported, as shown in the following example:
 
--- a/docs/standard/serialization/system-text-json-immutability.md
+++ b/docs/standard/serialization/system-text-json-immutability.md
@@ -18,7 +18,7 @@ This article shows how to use immutable types, parameterized constructors, and n
 ## Immutable types and Records
 
 ::: zone pivot="dotnet-5-0"
-`System.Text.Json` can use a parameterized constructor, which makes it possible to deserialize an immutable class or struct. For a class, if the only constructor is a parameterized one, that constructor will be used. For a struct, or a class with multiple constructors, specify the one to use by applying the [[JsonConstructor]](xref:System.Text.Json.Serialization.JsonConstructorAttribute) attribute. When the attribute is not used, a public parameterless constructor is always used if present. The attribute can only be used with public constructors. The following example uses the `[JsonConstructor]` attribute:
+`System.Text.Json` can use a public parameterized constructor, which makes it possible to deserialize an immutable class or struct. For a class, if the only constructor is a parameterized one, that constructor will be used. For a struct, or a class with multiple constructors, specify the one to use by applying the [[JsonConstructor]](xref:System.Text.Json.Serialization.JsonConstructorAttribute) attribute. When the attribute is not used, a public parameterless constructor is always used if present. The attribute can only be used with public constructors. The following example uses the `[JsonConstructor]` attribute:
 
 :::code language="csharp" source="snippets/system-text-json-how-to-5-0/csharp/ImmutableTypes.cs" highlight="13":::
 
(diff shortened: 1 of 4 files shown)


docs/standard/serialization/system-text-json-immutability.md
```diff
@@ -13,15 +13,19 @@ helpviewer_keywords:
 
 # How to use immutable types and non-public accessors with System.Text.Json
 
-In this article, you will learn how to use immutable types, such as Records, with the `System.Text.Json` namespace.
+This article shows how to use immutable types, parameterized constructors, and non-public accessors with the `System.Text.Json` namespace.
```
layomia, line 16: Why drop the `record` term here?

layomia (approved):
LGTM

docs/standard/serialization/system-text-json-immutability.md
```diff
@@ -13,15 +13,19 @@ helpviewer_keywords:
...
 
 :::code language="csharp" source="snippets/system-text-json-how-to-5-0/csharp/ImmutableTypes.cs" highlight="13":::
 
+The parameter names of a parameterized constructor must match the property names. Matching is case-insensitive, and the constructor parameter must match the actual property name even if you use [[JsonPropertyName]](xref:System.Text.Json.Serialization.JsonPropertyNameAttribute) to rename a property. In the following example, the name for the `TemperatureC` property is changed to `celsius` in the JSON, but the constructor parameter is still named `temperatureC`:
```
layomia, line 25: We don't need to provide more samples, but we should also mention that the other applicable serialization attributes (`JsonConverterAttribute`, `JsonIgnoreAttribute`, `JsonIncludeAttribute`, and `JsonNumberHandlingAttribute`) are also honored. I'm sure this is the case, but opened https://github.com/dotnet/runtime/issues/47855 to mop up any surprises.

docs/standard/serialization/system-text-json-immutability.md
```diff
@@ -13,15 +13,19 @@ helpviewer_keywords:
...
+`System.Text.Json` can use a parameterized constructor, which makes it possible to deserialize an immutable class or struct. For a class, if the only constructor is a parameterized one, that constructor will be used. For a struct, or a class with multiple constructors, specify the one to use by applying the [[JsonConstructor]](xref:System.Text.Json.Serialization.JsonConstructorAttribute) attribute. When the attribute is not used, a public parameterless constructor is always used if present. The attribute can only be used with public constructors. The following example uses the `[JsonConstructor]` attribute:
```
layomia, line 21: Consider adding the `public` qualifier here and in other locations where it wouldn't be awkward.

```suggestion
`System.Text.Json` can use a public parameterized constructor, which makes it possible to deserialize an immutable class or struct. For a class, if the only constructor is a parameterized one, that constructor will be used. For a struct, or a class with multiple constructors, specify the one to use by applying the [[JsonConstructor]](xref:System.Text.Json.Serialization.JsonConstructorAttribute) attribute. When the attribute is not used, a public parameterless constructor is always used if present. The attribute can only be used with public constructors. The following example uses the `[JsonConstructor]` attribute:
```

docs/standard/serialization/system-text-json-immutability.md
tdykstra: I omitted "such as Records" because this intro is shown for both 3.1 and 5.0, and "such as Records" didn't seem important enough to warrant making separate 3.1 and 5.0 intro sentences.

PR #22606, opened 2021-02-02
Fix APIScan metadata; make assembly visible

gewarren:
For APIScan.

*Hide whitespace changes in diff.*


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

PR #22625, opened 2021-02-03
Wrong order in code snippet example.

kristofferborgstroem:
"acc" and "elem" should be switched in sumListBack. The "acc" actually refers to the element in question and vice versa. Misleading, since the folder function for fold and foldback have different types.

## Summary

Switched acc and elem


cartermp (approved):
thanks!

PR #22643, opened 2021-02-03
Correct deconstruct rules

### LINKED ISSUE #22026: Impossible overloads in tuple deconstruction
The second half of section `Deconstructing user-defined types` contains two paragraphs and code example about warnings related to overloading `Deconstruct` methods.

While it's possible to compile the example, it's impossible to use this for Tuple deconstruction. Having 
```C#
    public void SomeMethod(Person person)
    {
        string a, b, c;         
		int i;
        decimal d;

        (a, b, c, i) = person;
        (a, b, c, d) = person;
    }
```
produces `CS0121	The call is ambiguous between the following methods or properties: 'Person.Deconstruct(out string, out string, out string, out int)' and 'Person.Deconstruct(out string, out string, out string, out decimal)'`

According to https://github.com/dotnet/csharplang/issues/1092 and https://github.com/dotnet/csharplang/issues/1998 it also looks that it's confirmed and intentional to distinguish the overload only by it's arity.

Therefore paragraph

>Because you can overload the Deconstruct method..

and paragraph

>The overloaded Deconstruct method in the following example...

and the following example should be removed from the documentation. Also new paragraph explaining this arity-only based logic should be added. Discouraging users from adding overloads with same arity.

---
#### Document Details

* ID: ec91dfa8-265f-a220-fe14-14047375ee93
* Version Independent ID: b2afe72f-5b30-8938-7b64-9f45056df36a
* Content: [Deconstructing tuples and other types](https://docs.microsoft.com/en-us/dotnet/csharp/deconstruct)
* Content Source: [docs/csharp/deconstruct.md](https://github.com/dotnet/docs/blob/master/docs/csharp/deconstruct.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixes #22026

Deconstruct methods must have a different number of arguments.


IEvangelist (approved):
The only thing that I miss from the original text is the "unambiguous" wording, something to call out that there is the potential for ambiguity. Otherwise, looks good to me, :shipit: when you're ready.

BillWagner:
I did update the wording this morning. I think it reads more clear now.

PR #22666, opened 2021-02-05
Adjust description

wuyuansushen:
Describe nothing


BillWagner:
Thanks for submitting this @wuyuansushen 

I'm going to close it. The syntax for combining delegates define this `+` operator in terms of delegates. For that reason, the original text is more correct.

PR #22682, opened 2021-02-05
update description metadata - batch 8

v-kents:
@billwagner @tdykstra

Updates for missing/incorrect description metadata.


docs/visual-basic/misc/bc30112.md
```diff
@@ -1,4 +1,5 @@
 ---
+description: "Learn more about: <typename>' is a namespace and cannot be used as an expression"
```
tdykstra, line 2: When the first word is surrounded by apostrophes, the first apostrophe is getting dropped from the description text.

tdykstra (changes requested):

docs/visual-basic/misc/bc30112.md
tdykstra: Sorry, I should have been clearer -- I only marked one instance in the comments but there are a number of them, the same fix needs to be applied to all of them.  Same clarification applies to batch 9.

v-kents: Yup, noticed that after I fixed the one. Just pushed another ~400. I've requested an update to the tool that does this scan. I'm wondering if it would be better to close off these two and re-run once the new version is available? (unless I got them all in this one)

tdykstra: >wondering if it would be better to close off these two and re-run once the new version is available? 

That makes sense to me unless it will be a long time before the fix is available.

v-kents: Hopefully Monday, but it might mean re-running from master. I'll see what Gene has to say.

v-kents:
OK, all the missing single quotes fixed on this one (that I could find)

docs/visual-basic/misc/bc30112.md
v-kents: Fixed all I could find (both ends) in this PR

tdykstra (approved):
LGTM

PR #22703, opened 2021-02-06
Fix typo in implementing-disposeasync.md

ericmutta:
Fixed a typo in the following sentence:

> Both `IAsyncDisposable` instances have **there** `DisposeAsync()` methods awaited, thus performing its asynchronous clean up operation. 

...and then tweaked the sentence for clarity so it reads:

> Both `IAsyncDisposable` instances have their `DisposeAsync()` method awaited, so each instance performs its asynchronous clean up operation.


gewarren (approved):
@ericmutta Thank you!

ericmutta:
Thanks for following up @gewarren :+1:

PR #22705, opened 2021-02-06
Fix typo in csharp-version-history.md

ericmutta:
The following sentence:

> Using generics is better than create `ListInt` that derives from...

Should read:

> Using generics is better than creating a `ListInt` type which derives from...


BillWagner (approved):
Thanks for submitting this fix @ericmutta 

ericmutta:
Thanks for following up @BillWagner :+1:

PR #22723, opened 2021-02-08
Clarify WinExe property breaking change

adegeo:
## Summary

Clarify property usage. If the property is omitted, it's not automatically set to `WinExe` as the description implies. This clarifies that it only applies when it's present. You still get a library output and not an exe output if you omit the property.

### DIFF
--- a/docs/core/compatibility/windows-forms/5.0/automatically-infer-winexe-output-type.md
+++ b/docs/core/compatibility/windows-forms/5.0/automatically-infer-winexe-output-type.md
@@ -17,7 +17,7 @@ In previous versions of the .NET SDK, the value that's specified for `OutputType
 </PropertyGroup>
 ```
 
-Starting in the 5.0.100 version of the .NET SDK, when the `OutputType` property is present, `OutputType` is automatically set to `WinExe` for WPF and Windows Forms apps that target any framework version, including .NET Framework. For example:
+Starting in the 5.0.100 version of the .NET SDK, when `OutputType` is set to `Exe`, it is automatically changed to `WinExe` for WPF and Windows Forms apps that target any framework version, including .NET Framework.
 
 ```xml
 <PropertyGroup>
--- a/docs/core/compatibility/windows-forms/5.0/automatically-infer-winexe-output-type.md
+++ b/docs/core/compatibility/windows-forms/5.0/automatically-infer-winexe-output-type.md
@@ -25,6 +25,8 @@ Starting in the 5.0.100 version of the .NET SDK, when `OutputType` is set to `Ex
 </PropertyGroup>
 ```
 
+ If `OutputType` is not specified in the project file, it defaults to `Library` and that value doesn't change.
+
 ## Reason for change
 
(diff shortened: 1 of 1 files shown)


gewarren:
@marcpopMSFT Is this correct?

marcpopMSFT:
That is correct.  We will [override](https://github.com/dotnet/sdk/blob/91c1326f478598c13095bf8bd5ee06246633730c/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets#L25) OutputType=_Exe_ to _WinExe_ for WPF and Winforms applications.  However, if OutputType is not specified in the project file, it will default to _Library_ and we won't change that value.

docs/core/compatibility/windows-forms/5.0/automatically-infer-winexe-output-type.md
```diff
@@ -17,7 +17,7 @@ In previous versions of the .NET SDK, the value that's specified for `OutputType
 </PropertyGroup>
 ```
 
-Starting in the 5.0.100 version of the .NET SDK, `OutputType` is automatically set to `WinExe` for WPF and Windows Forms apps that target any framework version, including .NET Framework. For example:
+Starting in the 5.0.100 version of the .NET SDK, when the `OutputType` property is present, `OutputType` is automatically set to `WinExe` for WPF and Windows Forms apps that target any framework version, including .NET Framework. For example:
```
gewarren, line 20:
```suggestion
Starting in the 5.0.100 version of the .NET SDK, when `OutputType` is set to `Exe`, it is automatically changed to `WinExe` for WPF and Windows Forms apps that target any framework version, including .NET Framework.
```

gewarren (approved):
Can you also add this after line 27?

```
 If `OutputType` is not specified in the project file, it defaults to `Library` and that value doesn't change.
```

adegeo:
> Can you also add this after line 27?
> 
> ```
>  If `OutputType` is not specified in the project file, it defaults to `Library` and that value doesn't change.
> ```

Is this still needed? Now that we clarified it only changes EXE to WINEXE, I don't know that `Library` is relevant.

gewarren:
> > Can you also add this after line 27?
> > ```
> >  If `OutputType` is not specified in the project file, it defaults to `Library` and that value doesn't change.
> > ```
> 
> Is this still needed? Now that we clarified it only changes EXE to WINEXE, I don't know that `Library` is relevant.

I think it's useful.

Kryptos-FR:
Thanks for updating the documentation and fixing #21952.

With that said, I have a general remark regarding the documentation: the date displayed on the page is hard-coded and hence is never updated. I always expected that after any update, it would also be reflected in the page. A lot of times, I visited some pages waiting for an update to occur, but since that date never changed, I had no idea.

I would suggest to either automatically update it based on the last commit date, or to set a rule that except for typo fixes, any addition or removal should also update the date.

For the current page, the date is still 09/18/2020, while this pull-request does present a significant change, considering how many people were confused by the new default behavior.

Issue #22731, opened 2021-02-09
Blazor still appears as "éblouissant" in french docs…


vintzl:
See my past issues on this subject…
- https://docs.microsoft.com/fr-fr/aspnet/core/blazor/state-management?view=aspnetcore-5.0&pivots=webassembly
- https://docs.microsoft.com/fr-fr/dotnet/core/compatibility/aspnet-core/5.0/blazor-browser-support-updated

Youssef1313:
Well, there are two issues here:

1. For the first link, it belongs to [another repo](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/state-management.md), and the issue is that "Blazor" is already in no-loc metadata. @JasonCard 

2. For the second link, it belongs to this repo and is missing no-loc metadata. <https://github.com/dotnet/docs/blob/master/docs/core/compatibility/aspnet-core/5.0/blazor-browser-support-updated.md>

Youssef1313:
Just figured out that the article is relying on an "include" in the ASP.NET docs repo, which doesn't have no-loc.

So for both links, it's the same reason.

However, there is one occurrence that I don't know how to lock its localization:

https://github.com/dotnet/AspNetCore.Docs/blob/c8600c840f492164308c14dd12e7f7ec389cb50d/aspnetcore/zone-pivot-groups.yml#L13-L20

JasonCard:
@Youssef1313 Regarding the text coming from zone-pivot-group.yml, that file should not have been localized, so it's been reverted back to English.

gewarren:
Can we close this issue now?

Youssef1313:
@gewarren I think yes. The remaining issues are tracked in asp docs repo.

PR #22732, opened 2021-02-09
Prevent Blazor from being localized

Youssef1313:
Contributes to #22731

### DIFF
--- a/docs/core/compatibility/aspnet-core/5.0/blazor-browser-support-updated.md
+++ b/docs/core/compatibility/aspnet-core/5.0/blazor-browser-support-updated.md
@@ -4,6 +4,7 @@ description: "Learn about the breaking change in ASP.NET Core 5.0 titled Blazor:
 author: scottaddie
 ms.author: scaddie
 ms.date: 10/01/2020
+no-loc: [Blazor, Blazor WebAssembly, Blazor Server]
 ---
 # Blazor: Updated browser support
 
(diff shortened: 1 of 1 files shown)


guardrex:
FYI: Strings must be quoted.

docs/core/compatibility/aspnet-core/5.0/blazor-browser-support-updated.md
```diff
@@ -4,6 +4,7 @@ description: "Learn about the breaking change in ASP.NET Core 5.0 titled Blazor:
 author: scottaddie
 ms.author: scaddie
 ms.date: 10/01/2020
+no-loc: [Blazor, Blazor WebAssembly, Blazor Server]
```
Youssef1313, line 7:
```suggestion
no-loc: [Blazor, "Blazor WebAssembly", "Blazor Server"]
```

guardrex:
Thanks @Youssef1313 ... I'm going to ask Jason if we need to do this *everywhere* in the ASP.NET Core docs. I think the answer will be that we should. If so, I'm going to open an issue and address it everywhere at once (over there, not here on this doc set).

gewarren:
> FYI: Strings (with spaces) must be quoted.

According to [this spec](https://review.docs.microsoft.com/en-us/new-hope/specs/markdown/ignore-localization?branch=master#implementations), the quotes aren't needed. But they don't hurt.

guardrex:
It was recommended by @JasonCard, so we adopted it. I perhaps mistakenly thought that he was saying that it's a requirement when it actually isn't. My bad.

guardrex:
Ok ... I see it. It was originally a misunderstanding of mine based on earlier remarks, but @JasonCard said that it seems happy either way ...

https://github.com/dotnet/AspNetCore.Docs/issues/19462#issuecomment-672104607

gewarren (approved):

PR #22771, opened 2021-02-10
Fix inconsistency in access modifiers for code sample

### LINKED ISSUE #22765: Inconsistency between ActivitySource example and OpenTelemetry example
Hi,

I'm trying to figure out how to store the (single) ActivitySource in my application, and I'm a bit confused by the inconsistency between the ActivitySource example and "best practices":

```csharp
internal static ActivitySource source = new ActivitySource("MyCompany.MyComponent.SourceName", "v1");
```
and best practices: 
> Create the ActivitySource once and store it in a static variable and use that instance as long as needed.

and the OpenTelemetry example:

```csharp
public class Program
{
    private static readonly ActivitySource MyActivitySource = new ActivitySource(
        "MyCompany.MyProduct.MyLibrary");

    public static void Main()
}
```

Why the difference in access modifiers (`internal static` vs `private static readonly`)?

If this is because the example only uses ActivitySource in the Main-function which is a purely artificial example, consider the example as the starting point for most beginners to ActivitySource/OpenTelemetry and change it to be consistent. Perhaps defining it to be `internal static` immediately would be best? (I'm not primarily a C# developer, som I actually don't know)

(I will submit this issue to opentelemetry-dotnet as well, depending on the outcome)

---
#### Document Details

* ID: 05233ce0-ba2f-6b67-c321-ce630c650968
* Version Independent ID: 18c7fa3f-5b50-1b53-68ba-4dbd9fd89396
* Content: [Distributed tracing - .NET](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/distributed-tracing)
* Content Source: [docs/core/diagnostics/distributed-tracing.md](https://github.com/dotnet/docs/blob/master/docs/core/diagnostics/distributed-tracing.md)
* GitHub Login: @sywhang
* Microsoft Alias: **suwhang**

sywhang:
## Summary

The two code samples for `ActivitySource` can be confusing because one is marked as `private` and the other is marked as `internal`. 

Fix #22765

### DIFF
--- a/docs/core/diagnostics/distributed-tracing.md
+++ b/docs/core/diagnostics/distributed-tracing.md
@@ -73,7 +73,7 @@ Applications and libraries can easily publish tracing data by simply using the <
 The first step to publish tracing data is to create an instance of the ActivitySource class. The ActivitySource is the class that provides APIs to create and start Activity objects and to register ActivityListener objects to listen to the Activity events.
 
 ```csharp
-    internal static ActivitySource source = new ActivitySource("MyCompany.MyComponent.SourceName", "v1");
+    private static ActivitySource source = new ActivitySource("MyCompany.MyComponent.SourceName", "v1");
 ```
 
 #### Best Practices


IEvangelist (approved):

PR #22779, opened 2021-02-11
Fix typo in nullable-analysis.md

ericmutta:
The sentence:

> You add the MemberNotNullAttribute to a method declaration **to fields** that are initialized to a non-null value in the method.

Should read:

> You add the MemberNotNullAttribute to a method declaration **and specify the fields** that are initialized to a non-null value in the method


BillWagner (approved):
Thanks for this fix @ericmutta 

That is more clear.

I'll :shipit: now.

ericmutta:
Always a pleasure @BillWagner many thanks for following up!

PR #22819, opened 2021-02-12
Set topic type to reference for WCF diagnostics

gewarren:
These may get archived soon, but just in case they don't...


tdykstra (approved):
LGTM

PR #22838, opened 2021-02-14
Fix xrefs; use bold for DebugView instead of code

zspitz:
## Summary

1. Fix xrefs to only display the type name where appropriate (and for the section of the VB document describing numeric literals, prefer the full type name).
2. Change `DebugView` to **DebugView**.


BillWagner (approved):
This LGTM @zspitz 
Thanks again for all your contributions.

I'll :shipit: now.

PR #22840, opened 2021-02-15
Added explanation of finally not getting executed

### LINKED ISSUE #20274: Finally is executed in C# 8.0
[The article gives an example of the finally block not getting executed when an exception is not handled. Is this still true for C# 8.0. I wrote an example to see and the finally block is getting hit. If so, could this page be updated to indicate this.]

---
#### Document Details

* ID: 3cd59701-079e-fdda-0c47-290f24f9eb4e
* Version Independent ID: a6f2953b-0da7-19fd-e121-aaeb65f60b17
* Content: [try-finally - C# Reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/try-finally)
* Content Source: [docs/csharp/language-reference/keywords/try-finally.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/keywords/try-finally.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

thekeviv:
## Summary

Added the extra explanation as suggested in the issue comments.

Fixes #20274


docs/csharp/language-reference/keywords/try-finally.md
```diff
@@ -14,7 +14,7 @@ ms.assetid: c27623fb-7261-4464-862c-7a369d3c8f0a
...
 
-Within a handled exception, the associated `finally` block is guaranteed to be run. However, if the exception is unhandled, execution of the `finally` block is dependent on how the exception unwind operation is triggered. That, in turn, is dependent on how your computer is set up.
+Within a handled exception, the associated `finally` block is guaranteed to be run. However, if the exception is unhandled, execution of the `finally` block is dependent on how the exception unwind operation is triggered. That, in turn, is dependent on how your computer is set up. The only cases where `finally` clauses don't run involve a program being immediately stopped. An example of this would be when `InvalidProgramException` gets thrown because of the IL statements being corrupt. On most operating systems, reasonable resource cleanup will take place as part of stopping and unloading the process.
```
Youssef1313, line 17: If a `try/finally` statement is written without `catch`. Does finally block run if the try throws an exception?

If not, this should also be added as a case.

BillWagner: In a `try`/`finally`, yes, the finally runs if the `try` throws an exception

BillWagner, line 17: Updating code fence to an xref:

```suggestion
Within a handled exception, the associated `finally` block is guaranteed to be run. However, if the exception is unhandled, execution of the `finally` block is dependent on how the exception unwind operation is triggered. That, in turn, is dependent on how your computer is set up. The only cases where `finally` clauses don't run involve a program being immediately stopped. An example of this would be when <xref:System.InvalidProgramException> gets thrown because of the IL statements being corrupt. On most operating systems, reasonable resource cleanup will take place as part of stopping and unloading the process.
```

BillWagner (changes requested):
Thank you for making this change @thekeviv 

I had one formatting suggestion. Once you accept that change, I'll :shipit:

BillWagner (approved):

BillWagner:
Thanks again @thekeviv 

I'll :shipiit: now.

PR #22845, opened 2021-02-15
Update deploy-with-cli.md

explorest:
## Summary

The "**--self-contained false**" switch  is actually added here to prevent creation of self contained deployment ( which, in addition to the application and its dependencies, also  includes the dotnet runtime and libraries ).  Self contained deployment  is the default when an '**-r <RID>**' switch is used with the "**dotnet publish**" command.

### DIFF
--- a/docs/core/deploying/deploy-with-cli.md
+++ b/docs/core/deploying/deploy-with-cli.md
@@ -134,7 +134,7 @@ For .NET 2.1, you must use the following switches with the `dotnet publish` comm
   This switch uses an identifier (RID) to specify the target platform. For a list of runtime identifiers, see [Runtime Identifier (RID) catalog](../rid-catalog.md).
 
 - `--self-contained false`
-  This switch tells the .NET Core SDK to not create a self-contained deployment (SCD), which is the default when -r switch is used. 
+  This switch disables the default behavior of the `-r` switch, which is to create a self-contained deployment (SCD). This switch creates an FDE.
 
 | Publish Mode                   | SDK Version | Command                                                     |
 |--------------------------------|-------------|-------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


docs/core/deploying/deploy-with-cli.md
```diff
@@ -134,7 +134,7 @@ For .NET 2.1, you must use the following switches with the `dotnet publish` comm
   This switch uses an identifier (RID) to specify the target platform. For a list of runtime identifiers, see [Runtime Identifier (RID) catalog](../rid-catalog.md).
 
 - `--self-contained false`
-  This switch tells the .NET Core SDK to create an executable as an FDE.
+  This switch tells the .NET Core SDK to not create a self-contained deployment (SCD), which is the default when -r switch is used.
```
adegeo, line 137: I like this idea. However, I think we should promote FDE here. Mentioning only SCD in the FDE section doesn't seem clear enough.

How does this sound?

```suggestion
  This switch disables the default behavior of the `-r` switch which is to create a self-contained deployment (SCD). Instead, an FDE will be created.
```

adegeo (changes requested):
There is an extra space at the end of your modification that needs to be removed. However, I have a suggestion for you to consider.

docs/core/deploying/deploy-with-cli.md
explorest: Cool. Its perfect.

explorest: Oh sorry -- I'm completely new to pull requests. It was my very first pull request ever.  I thought  I was done after commenting and your suggestion would be applied automatically. Today I checked and it occurred to me that i should commit the suggestion. So I did.  I don't know what else to do.

adegeo:
@explorest Yep that's all you have to do, commit the suggestion. Unfortunately we're having an issue with our publishing system so we're not getting any build reports back and can't merge this until then. Rest assured that your change will make it in once this problem has been fixed. Thanks again!

adegeo (approved):

BillWagner:

PR #22848, opened 2021-02-16
Remove Activity bullet from logging and tracing doc

sywhang:
## Summary

Now that we've added Distributed Tracing doc and have a link to it from this page, we should remove this to avoid redundancy / confusion.


noahfalk (approved):
👍

PR #22863, opened 2021-02-17
Update field naming convention

### LINKED ISSUE #22822: Fields should not be named camelCase because it overlaps with local variable names
Solution: To either prefix with a _fieldName or _PropertyName so that it matches the property name.

If you name a field in camel case, you have to fill in 3 details:
```
private int myVar;
public int MyProperty
{
get { return **myVar**; }
set { myVar = value; }
}
```
With the following convention, only two are required"
```
private int _MyProperty;
public int MyProperty
{
get { return _MyProperty; }
set { _MyProperty = value; }
}
 ```
Furthermore, if you end up having to change the name, you can do a simple search and replace and replace everywhere rather than having to be case sensitive and search twice. Also without at the minimum a underscore in front, a field could match the local variable (inside a method) with the same name.

Referring @MarkMichaelis to this as well as this is reflected in his naming conventions.

---
#### Document Details

* ID: 1df47593-8586-e54e-fb74-cf3a96262d40
* Version Independent ID: b464cd68-4713-c848-e9a7-30705d17e6f7
* Content: [Fields - C# Programming Guide](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/fields)
* Content Source: [docs/csharp/programming-guide/classes-and-structs/fields.md](https://github.com/dotnet/docs/blob/master/docs/csharp/programming-guide/classes-and-structs/fields.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

### PR DESCRIPTION
**Pull request by BenjaminMichaelis**
## Summary

- Updated field names to follow coding guidelines.
- Put `_date` next to its wrapping property.
- Placed public field below the public property as it shouldn't generally be used.

Fixes #22822

### DIFF
--- a/docs/csharp/programming-guide/classes-and-structs/snippets/fields/Program.cs
+++ b/docs/csharp/programming-guide/classes-and-structs/snippets/fields/Program.cs
@@ -3,25 +3,23 @@
 //<Snippet1>
 public class CalendarEntry
 {
-    // private field
-    private DateTime date;
 
-    // public field (Generally not recommended.)
-    public string day;
+    // private field (Located near wrapping "Date" property).
+    private DateTime _date;
 
-    // Public property exposes date field safely.
+    // Public property exposes _date field safely.
     public DateTime Date
     {
         get
         {
-            return date;
+            return _date;
         }
         set
         {
             // Set some reasonable boundaries for likely birth dates.
             if (value.Year > 1900 && value.Year <= DateTime.Today.Year)
             {
-                date = value;
+                _date = value;
             }
             else
             {
@@ -30,7 +28,10 @@ public class CalendarEntry
         }
     }
 
-    // Public method also exposes date field safely.
+    // public field (Generally not recommended).
+    public string Day;
+
+    // Public method also exposes _date field safely.
     // Example call: birthday.SetDate("1975, 6, 30");
     public void SetDate(string dateString)
     {
@@ -39,7 +40,7 @@ public class CalendarEntry
         // Set some reasonable boundaries for likely birth dates.
         if (dt.Year > 1900 && dt.Year <= DateTime.Today.Year)
         {
-            date = dt;
+            _date = dt;
         }
         else
         {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by Youssef1313 — CHANGES_REQUESTED** [review]
Please update <https://github.com/BenjaminMichaelis/docs/blob/BenjaminMichaelis-Issue%2322822/docs/csharp/programming-guide/classes-and-structs/fields.md> as well. Thanks!

File: docs/csharp/programming-guide/classes-and-structs/snippets/fields/Program.cs [hunk 1]
```diff
@@ -3,25 +3,23 @@
 //<Snippet1>
 public class CalendarEntry
 {
-    // private field
-    private DateTime date;
 
-    // public field (Generally not recommended.)
```
BillWagner, line 9: This line should stay in the source. While not recommended, it is allowed. The comment does indicate that it's not the recommended practice.

You could / should change the name to "Day" to follow the conventions.

BillWagner (changes requested):
Thanks for putting this together @BenjaminMichaelis 

I had one change in the code for you to make. Also, as @Youssef1313 mentioned, before I merge this, you need to update the corresponding text to match the new variable names.

You can add commits to this branch, and that updates the PR.

docs/csharp/programming-guide/classes-and-structs/snippets/fields/Program.cs
BenjaminMichaelis: This public field has already been renamed to "Day" but I moved it below the property so it still exists. I moved it away from the top because I thought that it would be poor to put an example that is not recommended up top, as well as then it allows the private field to be by its wrapping property which I think could be good practice, and if that happens, the public property would have to be the first example seen if it is still up top, but I can add it back to the top again, and move the private field above the public field like originally. Does that sounds good to move it back to the original format ordering or keep it as is? I will update the .md as well.

BillWagner: Thanks for clarifying @BenjaminMichaelis.  I missed it looking at the diffs. The declaration can stay where it is.

BillWagner (approved):

BillWagner:
Thanks for making the  additional updates @BenjaminMichaelis 

I'll :shipit: now.

You should see the changes live on our site in the next day or two, on our regular publishing cycle.

PR #22870, opened 2021-02-17
Fixing newing-up StreamWriter example

ghost:
## Summary

The example code is missing `StreamWriter` after `new` 

### DIFF
--- a/docs/csharp/programming-guide/file-system/snippets/write-text/StreamWriterTwo.cs
+++ b/docs/csharp/programming-guide/file-system/snippets/write-text/StreamWriterTwo.cs
@@ -5,7 +5,7 @@ class StreamWriterTwo
 {
     public static async Task ExampleAsync()
     {
-        using StreamWriter file = new("WriteLines2.txt", append: true);
+        using StreamWriter file = new StreamWriter("WriteLines2.txt", append: true);
         await file.WriteLineAsync("Fourth line");
     }
 }


docs/csharp/programming-guide/file-system/snippets/write-text/StreamWriterTwo.cs
```diff
@@ -5,7 +5,7 @@ class StreamWriterTwo
 {
     public static async Task ExampleAsync()
     {
-        using StreamWriter file = new("WriteLines2.txt", append: true);
+        using StreamWriter file = new StreamWriter("WriteLines2.txt", append: true);
```
Youssef1313, line 8: It's not missing. See https://github.com/dotnet/docs/issues/22810.

@BillWagner @IEvangelist Should we switch to the following instead:

```suggestion
        using var file = new StreamWriter("WriteLines2.txt", append: true);
```

The article focus isn't about C# 9 features.

IEvangelist: I do not believe there is any reason to change this. In fact, I do not believe that we should accept this PR. If we start marching everything backwards we'd end up eventually dropping the using declaration, then `async / await`, and even then `var`. C# 9 is out, and GA - we need to move forward. What do you think @BillWagner?

davisben85: Ah, thank you for the info.  I agree with the suggested change also, to provide an example that is _more_ compatible.

BillWagner: I agree. I think we should close this.

This is an interesting point overall. We should continue showing the newest syntax, to help readers learn it from our samples. 

That said, we also need to show some of the long-standing alternatives that readers will see in existing codebases.  

So, my conclusion is that our samples mirror that: Many are old enough to show older syntax. Newer ones (or those that have been updated) show newer variants. This is one of those newer variants.

davisben85: Bill - if providing some contextual documentation is possible, that would be great.  Even though C# 9 "is out", not everyone can immediately start using it for various reasons ( I am limited to what Unity3d supports ), so examples or even a comment in the C# samples that indicate "this is new in C# X" might be helpful.


Thanks for the time on this one, I didn't mean to cause a commotion!

PR #22884, opened 2021-02-17
Prevent localization by using inline code

Youssef1313:
Addresses https://github.com/dotnet/docs/pull/22874#issuecomment-780782715


tdykstra (approved):
Looks like this method works without breaking links, thanks for making the change.

PR #22777, opened 2021-02-11
Remove myget feeds from nuget.config

gewarren:
See https://docs.opensource.microsoft.com/tools/cg/security-alerts/nuget-multifeed-configuration.html?q=multifeed.

PR #22938, opened 2021-02-22
Add a note to dotnet nuget add source

Youssef1313:
FYI @gewarren

cc @mmitche if you want to add more details.

### DIFF
--- a/docs/core/tools/dotnet-nuget-add-source.md
+++ b/docs/core/tools/dotnet-nuget-add-source.md
@@ -25,6 +25,9 @@ dotnet nuget add source -h|--help
 
 The `dotnet nuget add source` command adds a new package source to your NuGet configuration files.
 
+> [IMPORTANT]
+> Avoid having the public `nuget.org` feed along with other feeds. For more information, see [Dependency confusion attack](https://www.msn.com/money/other/microsoft-warns-enterprises-of-new-dependency-confusion-attack-technique/ar-BB1dynJl).
+
 ## Arguments
 
 - **`PACKAGE_SOURCE_PATH`**
--- a/docs/core/tools/dotnet-nuget-add-source.md
+++ b/docs/core/tools/dotnet-nuget-add-source.md
@@ -26,7 +26,7 @@ dotnet nuget add source -h|--help
 The `dotnet nuget add source` command adds a new package source to your NuGet configuration files.
 
 > [IMPORTANT]
-> Avoid having the public `nuget.org` feed along with other feeds. For more information, see [Dependency confusion attack](https://www.msn.com/money/other/microsoft-warns-enterprises-of-new-dependency-confusion-attack-technique/ar-BB1dynJl).
+> Avoid having the public `nuget.org` feed along with other feeds. For more information, see [Dependency confusion attack](https://www.msn.com/en-us/money/other/microsoft-warns-enterprises-of-new-dependency-confusion-attack-technique/ar-BB1dynJl).
 
 ## Arguments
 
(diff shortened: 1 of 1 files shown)


docs/core/tools/dotnet-nuget-add-source.md
```diff
@@ -25,6 +25,9 @@ dotnet nuget add source -h|--help
 
 The `dotnet nuget add source` command adds a new package source to your NuGet configuration files.
 
+> [IMPORTANT]
+> Avoid having the public `nuget.org` feed along with other feeds. For more information, see [Dependency confusion attack](https://www.msn.com/money/other/microsoft-warns-enterprises-of-new-dependency-confusion-attack-technique/ar-BB1dynJl).
```
IEvangelist, line 29: This link is invalid, doesn't resolve for me. I did find related articles, I'm not sure this is directly related though.

IEvangelist (changes requested):
I'd like to have @blowdart chime in here for the correct course of action - thank you.

I wonder if linking to something like this is more appropriate? https://docs.microsoft.com/nuget/concepts/security-best-practices

docs/core/tools/dotnet-nuget-add-source.md
Youssef1313: @IEvangelist See also https://github.com/dotnet/docs/pull/22777.

blowdart:
The intent is good but linking to news articles is bad. Instead, consider something like, wrapped in the warning chrome?

Care should be taken when adding multiple feeds not to introduce a [dependency confusion attack](https://aka.ms/pkg-sec-wp).

blowdart:
That works, I leave it to @IEvangelist to chime in on wording :)

docs/core/tools/dotnet-nuget-add-source.md
```diff
@@ -25,6 +25,9 @@ dotnet nuget add source -h|--help
 
 The `dotnet nuget add source` command adds a new package source to your NuGet configuration files.
 
+> [!WARNING]
+> Be careful when adding multiple feeds not to introduce a [dependency confusion attack](https://aka.ms/pkg-sec-wp).
```
IEvangelist, line 29:
```suggestion
> [!WARNING]
> When adding multiple NuGet sources, be careful not to introduce a [dependency confusion attack](https://aka.ms/pkg-sec-wp).
```

IEvangelist (approved):

docs/core/tools/dotnet-nuget-add-source.md
Youssef1313: @IEvangelist Should it be multiple package sources instead to match the name used in [nuget.config](https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file#packagesources)?

IEvangelist: Yeah, that is fine. I updated the suggestion...

mmitche (approved):

docs/core/tools/dotnet-nuget-add-source.md
```diff
@@ -25,6 +25,9 @@ dotnet nuget add source -h|--help
 
 The `dotnet nuget add source` command adds a new package source to your NuGet configuration files.
 
+> [!WARNING]
+> When adding multiple package sources, be careful not to introduce a [dependency confusion attack](https://aka.ms/pkg-sec-wp).
```
tdykstra, line 29:
```suggestion
> When adding multiple package sources, be careful not to introduce a [dependency confusion vulnerability](https://aka.ms/pkg-sec-wp).
```

tdykstra (approved):
Thanks, @Youssef1313 !

PR #22943, opened 2021-02-22
Update TransientFaultHandlingOptions.cs

doterik:
According to the documentation:

- Consider the `TransientFaultHandlingOptions` **record type** defined


IEvangelist:
Hi 👋 @doterik
Thank you so much for posting this issue. We really appreciate it.

The doc is actually incorrect and should be updated, this was a part of a larger issue that was addressed a while back. Initially it was assumed that `record` types were valid for the options pattern, they are not. Where do you see this in the docs?

PR #22993, opened 2021-02-24
Fix misleading link in system-text-json-converters-how-to.md

ericmutta:
The following code in the [Registration sample - Converters collection](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?pivots=dotnet-5-0#registration-sample---converters-collection) section does NOT compile:

```csharp
var serializeOptions = new JsonSerializerOptions
{
    WriteIndented = true,
    Converters =
    {
        new DateTimeOffsetConverter()
    }
};

jsonString = JsonSerializer.Serialize(weatherForecast, serializeOptions);
```

The problem is that the [JsonSerializerOptions.Converters](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions.converters?view=net-5.0) property is of type `IList<JsonConverter>` and we are trying to add an instance of `DateTimeOffsetConverter` which inherits from `TypeConverter` instead of the expected `JsonConverter`. The compiler reports the error as follows:

image

While filing an issue for the bug, I noticed that the `DateTimeOffsetConverter` class being referenced in the sample code is not in fact the standard `System.ComponentModel.DateTimeOffsetConverter` but the custom version created in the [Sample basic converter](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?pivots=dotnet-5-0#sample-basic-converter) section. So I simply changed the link to point to the custom implementation instead of `System.ComponentModel.DateTimeOffsetConverter`.

### DIFF
--- a/docs/standard/serialization/system-text-json-converters-how-to.md
+++ b/docs/standard/serialization/system-text-json-converters-how-to.md
@@ -142,7 +142,7 @@ You can throw other exceptions as needed, but they don't automatically include J
 
 ## Registration sample - Converters collection
 
-Here's an example that makes the <xref:System.ComponentModel.DateTimeOffsetConverter> the default for properties of type <xref:System.DateTimeOffset>:
+Here's an example that makes the [DateTimeOffsetConverter](#sample-basic-converter) the default for properties of type <xref:System.DateTimeOffset>:
 
 :::code language="csharp" source="snippets/system-text-json-how-to/csharp/RegisterConverterWithConvertersCollection.cs" id="Serialize":::
 


IEvangelist (approved):
I'll let @tdykstra verify this -- but I believe that your observations are correct. Thank you @ericmutta, but to be clear the code does compile, right?

ericmutta:
Thanks for following up @IEvangelist! Yes, the code does compile if you take the whole sample including the custom `DateTimeOffsetConverter` implementation. However if you just copy the snippet in isolation, the compiler resolves `DateTimeOffsetConverter`  to the built-in `System.ComponentModel.DateTimeOffsetConverter` and then the code fails to compile.

tdykstra (approved):
thanks for fixing this @ericmutta. I'll change the name of the converter to eliminate ambiguity.

ericmutta:
Thanks @tdykstra - removing the ambiguity completely is the way to go :+1:

tdykstra:
Closing and reopening to trigger build

PR #22996, opened 2021-02-24
clarify defensive copies

### LINKED ISSUE #21691: Mention how to mitigate performance issues with mutable structs and in parameter modifier
The Documentation says that using `in` parameters with mutable structs may lead to defensive copy's that decrease performance compared to a `by-value` approach.

However depending on the code a big struct may need to be modified often. Following advice would lead to performance decrease in those cases.

From what I'm understand if the method with the `in` parameter only calls `readonly` methods/properties and access only filed's on the struct it will not create a defensive copy. But it is not stated that explicitly.

>However, the compiler can't know if any member method modifies the state of the struct. To ensure that the object isn't modified, the compiler creates a copy and calls member references using that copy. Any modifications are to that defensive copy.

This sounds to my like it will always copy the value if its not `readonly struct`, but profiling suggests otherwise.

I would like to have a list of does and don'ts if a immutable object is not an option in the documentation like following:

**I'm don't know if the below is correct**

The compiler will generate a defensive copy if any:
 * A method or property is called that is not marked with `readonly`
 * ...?

The compiler does not generate a defensive copy if you only:
 * access fields of the struct
 * call only methods and properties that are marked with readonly
 * pass the parameter to an method using an `in` parameter and that method will abides this rules
 * copy the struct to a local value

---
#### Document Details

* Content: [Write safe and efficient C# code](https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#apply-the-in-modifier-to-readonly-struct-parameters-larger-than-systemintptrsize)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

### LINKED ISSUE #22955: A concrete example of the usage could be added
At first glance this seems to be the same as passing by value without the `in` keyword. To illustrate the difference, an example such as in the top comment of this post could be added. https://stackoverflow.com/questions/52820372/why-would-one-ever-use-the-in-parameter-modifier-in-c

---
#### Document Details

* ID: 6b1434b0-5459-7569-47f7-46bf33fc3953
* Version Independent ID: 8a4ae81b-1ae5-cbe6-d58b-df455414e245
* Content: [in parameter modifier - C# Reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/in-parameter-modifier)
* Content Source: [docs/csharp/language-reference/keywords/in-parameter-modifier.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/keywords/in-parameter-modifier.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixes #21691

- Clarify the rules for defensive copies

Fixes #22955

- Provide link to conceptual doc with examples.

### DIFF
--- a/docs/csharp/language-reference/keywords/in-parameter-modifier.md
+++ b/docs/csharp/language-reference/keywords/in-parameter-modifier.md
@@ -9,7 +9,7 @@ helpviewer_keywords:
 
 # in parameter modifier (C# Reference)
 
-The `in` keyword causes arguments to be passed by reference. It makes the formal parameter an alias for the argument, which must be a variable. In other words, any operation on the parameter is made on the argument. It is like the [ref](ref.md) or [out](out-parameter-modifier.md) keywords, except that `in` arguments cannot be modified by the called method. Whereas `ref` arguments may be modified, `out` arguments must be modified by the called method, and those modifications are observable in the calling context.
+The `in` keyword causes arguments to be passed by reference but ensures the argument is not modified. It makes the formal parameter an alias for the argument, which must be a variable. In other words, any operation on the parameter is made on the argument. It is like the [ref](ref.md) or [out](out-parameter-modifier.md) keywords, except that `in` arguments cannot be modified by the called method. Whereas `ref` arguments may be modified, `out` arguments must be modified by the called method, and those modifications are observable in the calling context.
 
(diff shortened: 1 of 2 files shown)


IEvangelist (approved):

PR #23022, opened 2021-02-26
Change class to record

iozcelik:
According the documentation, it would be record.
"Consider the TransientFaultHandlingOptions record type defined as follows:"

## Summary

Describe your changes here.

### DIFF
--- a/docs/core/extensions/snippets/configuration/console-json/TransientFaultHandlingOptions.cs
+++ b/docs/core/extensions/snippets/configuration/console-json/TransientFaultHandlingOptions.cs
@@ -2,7 +2,7 @@
 
 namespace ConsoleJson.Example
 {
-    public class TransientFaultHandlingOptions
+    public record TransientFaultHandlingOptions
     {
         public bool Enabled { get; set; }
         public TimeSpan AutoRetryDelay { get; set; }


docs/core/extensions/snippets/configuration/console-json/TransientFaultHandlingOptions.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace ConsoleJson.Example
 {
-    public class TransientFaultHandlingOptions
+    public record TransientFaultHandlingOptions
```
Youssef1313, line 5: Looks like the code is correct but the docs wording should be changed to "class".

@IEvangelist Can you please take a look since this is related to https://github.com/dotnet/docs/pull/22210?

Youssef1313: Actually there are two different articles using the same code snippet. I'm not sure if both should be classes or not.

IEvangelist: All options pattern should be shown using a `class`, the `record` will only work with rare-undesired workarounds and partially. It was a mistake to use `record` with this.

Youssef1313: Article 1:

https://github.com/dotnet/docs/blob/master/docs/core/extensions/options.md
https://docs.microsoft.com/en-us/dotnet/core/extensions/options

Article 2:

https://github.com/dotnet/docs/blob/master/docs/core/extensions/configuration-providers.md
https://docs.microsoft.com/en-us/dotnet/core/extensions/configuration-providers

gewarren:
Close/reopen for build.

PR #23034, opened 2021-02-28
Update linux-snap.md

phillipsj:
## Summary

Global tools produce an error when installed using the snap because it isn't to locate the DOTNET_ROOT. I am including this change to highlight this others.


adegeo:
Thank you for this suggestion. Do you know if the snaps can just set that for you automatically? Is there some other configuration through a snap profile or something that can set this up? I don't know linux well enough to answer these questions. @rbhanda perhaps you can answer?

@phillipsj I don't think this should be a section pointing out the error, this should be a section just about configuring the system for global tools when you use snap for dotnet.

rbhanda:
Hi @adegeo I could not find anything in snap documentation to set this up automatically.

adegeo (approved):

adegeo:
Thanks @phillipsj I'll get this merged now.

PR #23039, opened 2021-02-28
Discards article: Fix the section about the discard pattern

pkulikov:
Fixed two issues:
- Discard pattern is not supported by the `is` operator (`if (o is _) { }` doesn't compile; neither `case _` within the `switch` statement compiles). So, only by the `switch` expression. Also added the link to it.
- Example has been updated but its description - not. Changed the mention of `is` operator to `switch` expression, which is used in the example.

### DIFF
--- a/docs/csharp/discards.md
+++ b/docs/csharp/discards.md
@@ -36,7 +36,7 @@ For more information on deconstructing user-defined types with discards, see [De
 
 ## Pattern matching with `switch`
 
-The *discard pattern* can be used in pattern matching with the [switch expression](language-reference/operators/switch-expression.md). Every expression always matches the discard pattern.
+The *discard pattern* can be used in pattern matching with the [switch expression](language-reference/operators/switch-expression.md). Every expression, including `null`,  always matches the discard pattern.
 
 The following example defines a `ProvidesFormatInfo` method that uses a `switch` expression to determine whether an object provides an <xref:System.IFormatProvider> implementation and tests whether the object is `null`. It also uses the discard pattern to handle non-null objects of any other type.
 
--- a/docs/csharp/discards.md
+++ b/docs/csharp/discards.md
@@ -36,9 +36,9 @@ For more information on deconstructing user-defined types with discards, see [De
 
 ## Pattern matching with `switch`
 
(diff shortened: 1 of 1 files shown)


docs/csharp/discards.md
```diff
@@ -36,9 +36,9 @@ For more information on deconstructing user-defined types with discards, see [De
 
 ## Pattern matching with `switch`
 
-The *discard pattern* can be used in pattern matching with the [switch](language-reference/keywords/switch.md) keyword. Every expression always matches the discard pattern. (It can be used with [is](language-reference/keywords/is.md) expressions. However, that use is rare because the discard can be removed without changing its meaning).
```
Youssef1313, line 39: > (It can be used with [is](language-reference/keywords/is.md) expressions. However, that use is rare because the discard can be removed without changing its meaning).

This removed sentence seems correct (but probably not clear/confusing). I believe it's referring to `if (obj is SomeType _)`.

pkulikov: It's not the discard pattern. It's the declaration pattern with a discard.

docs/csharp/discards.md
```diff
@@ -36,9 +36,9 @@ For more information on deconstructing user-defined types with discards, see [De
...
+The *discard pattern* can be used in pattern matching with the [switch expression](language-reference/operators/switch-expression.md). Every expression always matches the discard pattern.
```
BillWagner, line 39: I'd be more explicit to call out `null` here:

```suggestion
The *discard pattern* can be used in pattern matching with the [switch expression](language-reference/operators/switch-expression.md). Every expression, including `null`,  always matches the discard pattern.
```

BillWagner (commented):
Overall, this is great. I had one small nit to consider.

(Commenting so it stays in my queue to review again).

docs/csharp/discards.md
pkulikov: @BillWagner I agree with your suggestion, Thank you

BillWagner (approved):
This looks great @pkulikov 

I'll :shipit: once we've resolved our ongoing CI build issues.

BillWagner:

Issue #20584, opened 2020-09-11
Language reference: Patterns


BillWagner:
As of C# 9, the pattern expressions are robust enough to justify their own page in "expressions and operators". This should cover:

- type pattern
- constant pattern
- null pattern as a useful special case of the constant pattern
- range patterns (`<`, `>`, `<=`, `>=`)
- property pattern (`when` patterns)
- `and`, `or`, and `not` pattern

In addition, the `switch` and `is` pages should link to this new page. They can cover less of the pattern syntax which likely moves to this new page.

pkulikov:
@BillWagner what about the `swtich` statement? Does it support all those patterns as well? If yes, the separate TOC node (just under the "Language reference" node) for "Pattern matching" is possible with individual articles per pattern. Any such article would present three examples of usage, then (with `switch` and `is` expressions and with `switch` statement). Placing patterns under "Expressions and operators" doesn't play well with the `switch` statement supporting them as well.

Anyway, I like a lot that a lot of duplication from `is`/`switch` articles would be consolidated.

BillWagner:
This update should also cover the discard pattern (`_` as the switch arm in a `switch` expression).  See https://github.com/dotnet/docs/pull/21318/files#r516723654

pkulikov:
@BillWagner I'm on fence about if it should be one article or multiple. Multiple articles suggest the separate TOC item and more visible structure/navigation among various patterns; however, with that approach some article can be very small (for example, one about discard pattern). What do you think?

BillWagner:
@pkulikov I've been in that situation a few times myself.  

I'd recommend starting with one article.  If that gets too long, having all the content and examples in one place will often give us guidance on how best to break it up.

pkulikov:
@BillWagner thank you; let's go on with one article. 

Original idea was to put it under "Expression and operators". Is it still the good location or there is a better one?

Also, I like to submit small PRs. So, it will be a PR per pattern. How do you want it to be arranged? Submit PRs directly to the master (so, the article would grow live), or submit PRs to the dedicated branch, that will be merged into the master at the end (or interim).

BillWagner:
@pkulikov 

> Also, I like to submit small PRs. So, it will be a PR per pattern. How do you want it to be arranged? 

I think it would be better to publish them all at once. I'm concerned that each smaller PR might be less useful for readers.

[...]

PR #23045, opened 2021-03-01
C# patterns reference: constant and discard patterns

pkulikov:
Contributes to #20584

### DIFF
--- a/docs/csharp/language-reference/operators/patterns.md
+++ b/docs/csharp/language-reference/operators/patterns.md
@@ -1,7 +1,7 @@
 ---
 title: "Patterns - C# reference"
 description: "Learn about the patterns supported by C# switch and is expressions and C# switch statement - C# reference"
-ms.date: 02/18/2021
+ms.date: 03/02/2021
 helpviewer_keywords: 
   - "pattern matching [C#]"
 ---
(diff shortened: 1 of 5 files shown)


docs/csharp/language-reference/operators/patterns.md
```diff
@@ -41,6 +41,37 @@ Beginning with C# 9.0, for that purpose you can use the *type pattern* as the fo
...
+
+The compiler guarantees that no user-overloaded equality operator `==` is invoked when expression `x is null` is evaluated.
+
+## Discard pattern
+
+Beginning with C# 8.0, you use the *discard pattern* `_` to match any expression, including `null`, as the following example shows:
+
+:::code language="csharp" source="snippets/patterns/DiscardPattern.cs" id="BasicExample":::
+
+The discard pattern is useful in `switch` expressions. If none of a `switch` expression's patterns matches an input, the runtime throws an exception. To guarantee that a `switch` expression handles all possible values of its input as you want, provide an arm with the discard pattern. As a `switch` expression evaluates its arms in text order, the arm with the discard pattern must be last.
```
BillWagner, line 71: It might be useful to mention that the compiler emits a warning if the arms of a `switch` expression don't handle all possible inputs.

This one may be surprising for some readers when the switch variable is an `enum`.

BillWagner (commented):
I like this as well @pkulikov 

I had one suggestion, and then, this is ready.

It does spark questions on ensuring how we'll make sure the final version is "complete". I had several thoughts on features to add, but I think these are going into later PRs (relational patterns, property pattterns, recursive patterns and so on). I think that means simply ensuring one final review for completeness when you think all the relevant features are covered. LMK what you think of the one suggestion, and then I'll merge this.

docs/csharp/language-reference/operators/patterns.md
pkulikov: @BillWagner I think that mention should be not in this article, but in the `switch` expression article; that article even has the [dedicated section](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression#non-exhaustive-switch-expressions) for that. What do you think?

> This one may be surprising for some readers when the switch variable is an `enum`.

Yes, I've tested that scenario :). I think that the compiler message is very good and clear. For example, switching on `DayOfWeek` enum:

> Warning CS8524 The switch expression does not handle some values of its input type (it is not exhaustive) involving an unnamed enum value. For example, the pattern '(System.DayOfWeek)7' is not covered.

I would say such a warning is self-explanatory.

pkulikov:
@BillWagner 

> It does spark questions on ensuring how we'll make sure the final version is "complete". I had several thoughts on features to add, but I think these are going into later PRs

Yes, for sure there will be one final PR to "proof-read" the whole article. Maybe it will be the one that would add the introduction. As for the next PRs, the section to come are (1) relational patterns, (2) pattern combinators (and, or, not), (3) tuple patterns, (4) positional patterns, (5) property pattern, and (6) `var` pattern.

I like working against the separate branch as I can come back to existing sections and enhance them along the way (somehow it feels more natural than editing against the main branch).

So, with the next sections being added, I expect some links between sections/patterns. That would be implicit description of recursive patterns.

docs/csharp/language-reference/operators/patterns.md
pkulikov: I can update the switch expression article in the main branch accordingly

BillWagner:
@pkulikov 

>  Maybe it will be the one that would add the introduction

That's a great plan.

I'll merge this one now.

PR #23046, opened 2021-03-01
Numeric parsing precision breaking change

### LINKED ISSUE #22458: Update standard numeric format parsing to handle higher precision
## Update the parsing/formatting logic to allow for precision more than 99 digits

Currently, the standard numeric format parsing logic is limited to precision <= 99. Some numeric types have more precision, but `ToString(string format)` does not expose it correctly. With https://github.com/dotnet/runtime/issues/46874, we now support precision up to `int.MaxValue`. Precision > `int.MaxValue` will throw a `FormatException`. Since the change occurs in the parsing logic, this will affect all numeric types.

### Version introduced

.NET 6 Preview 2

### Old behavior

Did not support precision > 99. For ex: `32.ToString("C100")` will print `C132`. This occurred because the parse routine interpreted precision more than 99 as a custom numeric format string. Consequently formats such as `H100` would be interpreted as a custom format and print wrong values while `H99` would throw because it was (correctly) interpreted as an unsupported **standard** format 

### New behavior
The new rule is: A format modifier with any number of digits is interpreted as a standard format + precision. If the value doesn't fit into an `int`, we throw.
For ex:
`32.ToString("C100")` will print `$32.0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000`
Unsupported formats such as `32.ToString("H100")` will throw a FormatException.

### Reason for change

Fix unexpected behavior when using higher precision.

### Recommended action

If you are using formats such as `G142` today, you are technically encountering a bug. If you want to maintain the previous behavior (i.e. where 42.ToString("G999") was intentionally meant to return G999), you can explicitly single quote the first character (i.e. 42.ToString("'G'999")). This will work on .NET Framework, .NET Core, and .NET 5+

Format strings, like the following, would continue to be interpreted as "custom numeric format strings":

- start with any character that is not an ASCII alphabetical character (such as $ or è)
- start with an ASCII alphabetical character but which are not followed by an ASCII digit (such as A$)
- start with an ASCII alphabetical character, followed by an ASCII digit sequence, and then any character that is not an ASCII digit character (such as A55A).

### Category

- [ * ] Core .NET libraries

### Affected APIs

``` csharp
BigInteger.ToString(string format)
BigInteger.ToString(string format, IFormatProvider provider)
BigInteger.TryFormat()
// Integral Types
int/uint/byte/sbyte/short/ushort/long/ulong.ToString(string format)
int/uint/byte/sbyte/short/ushort/long/ulong.ToString(string format, IFormatProvider provider)
int/uint/byte/sbyte/short/ushort/long/ulong.TryFormat()
// Floating Types
half/single/double/decimal.ToString(string format)
half/single/double/decimal.ToString(string format, IFormatProvider provider)
half/single/double/decimal.TryFormat()
```

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #22458.

Also updates "Standard numeric format strings" to include the new precision limit and to add the `TryFormat` method (available since .NET Core).

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/numeric-format-parsing-handles-higher-precision?branch=pr-en-us-23046).

### DIFF
--- a/docs/core/compatibility/core-libraries/6.0/numeric-format-parsing-handles-higher-precision.md
+++ b/docs/core/compatibility/core-libraries/6.0/numeric-format-parsing-handles-higher-precision.md
@@ -5,11 +5,22 @@ ms.date: 02/26/2021
 ---
 # Standard numeric format parsing precision
 
-
+In .NET 6+, .NET supports precision up to <xref:System.Int32.MaxValue?displayProperty=nameWithType> when formatting numbers as strings using `ToString` and `TryFormat`.
 
 ## Change description
 
+In previous .NET versions, the standard numeric format parsing logic is limited to a precision of 99 or less. Some numeric types have more precision, but `ToString(string format)` does not expose it correctly. If you specify a precision greater than 99, for example, `32.ToString("C100")`, the format string is interpreted as a custom numeric format string instead of "currency with precision 100". In custom numeric format strings, characters are interpreted as character literals. In addition, a format string that contains an invalid format specifier (that is, the character at the beginning of the string) is interpreted differently depending on the precision value. `H99` throws a <xref:System.FormatException> for the invalid format specifier, while `H100` is interpreted as a custom numeric format string.
 
+Starting in .NET 6, .NET supports precision up to <xref:System.Int32.MaxValue?displayProperty=nameWithType>. A format string that consists of a valid format specifier with any number of digits is interpreted as a standard numeric format string with precision. If you specify a precision that's greater than <xref:System.Int32.MaxValue?displayProperty=nameWithType>, a <xref:System.FormatException> is thrown. This change was implemented in the parsing logic that affects all numeric types.
+
+The following table shows how the behavior for various format strings changed.
+
+| Format string | Previous behavior | .NET 6+ behavior |
+| - | - | - |
+| `C2` | Denotes currency with two decimal digits | Denotes currency with two decimal digits (no change) |
+| `C100` | Denotes custom numeric format string that prints "C100" | Denotes currency with 100 decimal digits |
+| `H99` | Throws <xref:System.FormatException> due to invalid standard format specifier "H" | Denotes custom numeric format string |
+| `H100` | Denotes custom numeric format string | Denotes custom numeric format string (no change) |
 
 ## Version introduced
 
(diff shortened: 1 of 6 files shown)


docs/core/compatibility/core-libraries/6.0/numeric-format-parsing-handles-higher-precision.md
```diff
@@ -0,0 +1,142 @@
...
+
+Starting in .NET 6, .NET supports precision up to <xref:System.Int32.MaxValue?displayProperty=nameWithType>. A format string that consists of a valid format specifier with any number of digits is interpreted as a standard numeric format string with precision. If you specify a precision that's greater than <xref:System.Int32.MaxValue?displayProperty=nameWithType>, a <xref:System.FormatException> is thrown. This change was implemented in the parsing logic that affects all numeric types.
+
+The following table shows how the behavior for various format strings changed.
+
+| Format string | Previous behavior | .NET 6+ behavior |
```
gewarren, line 18: @pgovind Can you check these examples?

docs/standard/base-types/standard-numeric-format-strings.md
```diff
@@ -20,11 +20,11 @@ helpviewer_keywords:
...
+- `xxx` is an optional integer called the *precision specifier*. The precision specifier affects the number of digits in the resulting string. In .NET 6 and later versions, the maximum precision value is <xref:System.Int32.MaxValue?displayProperty=nameWithType>. In previous .NET versions, the precision can range from 0 to 99. The precision specifier controls the number of digits in the string representation of a number. It does not round the number itself. To perform a rounding operation, use the <xref:System.Math.Ceiling%2A?displayProperty=nameWithType>, <xref:System.Math.Floor%2A?displayProperty=nameWithType>, or <xref:System.Math.Round%2A?displayProperty=nameWithType> method.
```
gewarren, line 27: Not sure if `xxx` is the best way to show the placeholder for the precision since it can be much more than 3 digits. Open to suggestions.

docs/core/compatibility/core-libraries/6.0/numeric-format-parsing-handles-higher-precision.md
```diff
@@ -0,0 +1,142 @@
+---
+title: "Breaking change: Standard numeric format parsing precision"
+description: Learn about the .NET 6.0 breaking change in core .NET libraries where standard numeric format parsing now handles higher precisions.
+ms.date: 02/26/2021
+---
+# Standard numeric format parsing precision
+
+In .NET 6+, .NET supports precision up to <xref:System.Int32.MaxValue?displayProperty=nameWithType> when formatting numbers as strings using `ToString` and `TryFormat`.
```
adegeo, line 8: This seems confusing. Considering MaxValue isn't a number of decimal places, but the highest value an Int32 can represent, it seems odd that it's being referenced. I may not be understanding something though. It may be the way the word precision is used.

docs/core/compatibility/core-libraries/6.0/numeric-format-parsing-handles-higher-precision.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Change description
+
+In previous .NET versions, the standard numeric format parsing logic is limited to a precision of 99 or less. Some numeric types have more precision, but `ToString(string format)` does not expose it correctly. If you specify a precision greater than 99, for example, `32.ToString("C100")`, the format string is interpreted as a custom numeric format string instead of "currency with precision 100". In custom numeric format strings, characters are interpreted as character literals. In addition, a format string that contains an invalid format specifier (that is, the character at the beginning of the string) is interpreted differently depending on the precision value. `H99` throws a <xref:System.FormatException> for the invalid format specifier, while `H100` is interpreted as a custom numeric format string.
```
adegeo, line 12: When this is talking about "precision of 99" is this 99 digits after the decimal?

docs/core/compatibility/core-libraries/6.0/numeric-format-parsing-handles-higher-precision.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Reason for change
+
+This change corrects unexpected behavior when using higher precision for numeric format parsing.
+
+## Recommended action
+
+In most cases, no action is necessary and the correct precision will be shown in the resulting strings.
+
+However, if you want to revert to the previous (incorrect) behavior where the alphabetic character in the format string is interpreted as a literal character when the precision is greater than 99, you can wrap the first character in single quotes or escape it with a backslash. For example, in previous .NET versions, `42.ToString("G999")` returns `G999`. To maintain that behavior, change the format string to `"'G'999"`. This will work on .NET Framework, .NET Core, and .NET 5+.
```
adegeo, line 37: It would also be helpful to show the backslash version to.

docs/standard/base-types/standard-numeric-format-strings.md
adegeo: Since these are placeholders, what about `A###`

adegeo (approved):
Looks good. Just some thoughts to consider.

docs/standard/base-types/standard-numeric-format-strings.md
gewarren: # might get confused with the digit placeholder `#` of a custom numeric format string: https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-numeric-format-strings

docs/core/compatibility/core-libraries/6.0/numeric-format-parsing-handles-higher-precision.md
pgovind: The last one `H100` needs to be changed. `H` is an invalid standard format specifier, so it would throw a `FormatException` (same as `H99`). Up to you if you want to keep both `H99` and `H100` or pick one :)

PR #23136, opened 2021-03-05
Remove manual HTML ext icons

### LINKED ISSUE #23135: Now that the platform adds an ext icon to links outside docs.microsoft.com - we need a bit of clean up
The platform now automatically styles links that are external (defined as any URL not containing "docs.microsoft.com"), and adds an EXT icon. We were doing this manually for a while due to the limitation of the platform -- now it's time to do the clean up work.

image

IEvangelist:
## Summary

Remove manual HTML ext icons

Fixes #23135


CamSoper (approved):
Thanks for catching this!

PR #23142, opened 2021-03-05
Add note about buffer size for EventPipe

sywhang:
## Summary

Add note about buffer size being too small & events get dropped.

### DIFF
--- a/docs/core/diagnostics/dotnet-trace.md
+++ b/docs/core/diagnostics/dotnet-trace.md
@@ -87,7 +87,7 @@ dotnet-trace collect [--buffersize <size>] [--clreventlevel <clreventlevel>] [--
   Sets the size of the in-memory circular buffer, in megabytes. Default 256 MB.
 
   > [!NOTE]
-  > If the target process writes events too frequently, it can overflow this buffer and some events might be dropped. If too many events are getting dropped, increase the buffer size to see if the number of dropped events reduces. If the number of dropped events does not decrease with a larger buffer size, it may be due to a slow reader blocking the target process' buffers to be flushed.
+  > If the target process writes events too frequently, it can overflow this buffer and some events might be dropped. If too many events are getting dropped, increase the buffer size to see if the number of dropped events reduces. If the number of dropped events does not decrease with a larger buffer size, it may be due to a slow reader preventing the target process' buffers from being flushed.
 
 - **`--clreventlevel <clreventlevel>`**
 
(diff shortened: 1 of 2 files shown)


docs/core/diagnostics/dotnet-trace.md
```diff
@@ -86,6 +86,9 @@ dotnet-trace collect [--buffersize <size>] [--clreventlevel <clreventlevel>] [--
 
   Sets the size of the in-memory circular buffer, in megabytes. Default 256 MB.
 
+  > [!NOTE]
+  > If the target process is writing events too frequently, it can overflow this buffer which results in some events to get dropped. When too many events are getting dropped, increasing the buffer size may help.
```
gewarren, line 90:
```suggestion
  > If the target process writes events too frequently, it can overflow this buffer and some events might be dropped. If too many events are getting dropped, increase the buffer size.
```

docs/core/diagnostics/eventpipe.md
```diff
@@ -74,6 +74,9 @@ However, you can use the following environment variables to set up an EventPipe
 
 * `COMPlus_EventPipeCircularMB`: A hexadecimal value that represents the size of EventPipe's internal buffer in megabytes. This configuration value is only used when EventPipe is configured to run via `COMPlus_EnableEventPipe`. The default buffer size is 1024MB which translates to this environment variable being set to `400`, since `0x400` == `1024`.
 
+  > [!NOTE]
+  > If the process is writing events too frequently, it can overflow this buffer which results in some events to get dropped. When too many events are getting dropped, increasing the buffer size may help.
```
gewarren, line 78:
```suggestion
  > If the target process writes events too frequently, it can overflow this buffer and some events might be dropped. If too many events are getting dropped, increase the buffer size.
```

docs/core/diagnostics/dotnet-trace.md
sywhang: Thanks for the feedback @gewarren - I had to reword it a little bit because the "may help" phrasing here was chosen deliberately because it doesn't always help. I've updated the PR to reflect the suggestions minus that part and made some clarifications to that.

docs/core/diagnostics/dotnet-trace.md
```diff
@@ -86,6 +86,9 @@ dotnet-trace collect [--buffersize <size>] [--clreventlevel <clreventlevel>] [--
 
   Sets the size of the in-memory circular buffer, in megabytes. Default 256 MB.
 
+  > [!NOTE]
+  > If the target process writes events too frequently, it can overflow this buffer and some events might be dropped. If too many events are getting dropped, increase the buffer size to see if the number of dropped events reduces. If the number of dropped events does not decrease with a larger buffer size, it may be due to a slow reader blocking the target process' buffers to be flushed.
```
josalem, line 90:
```suggestion
  > If the target process writes events too frequently, it can overflow this buffer and some events might be dropped. If too many events are getting dropped, increase the buffer size to see if the number of dropped events reduces. If the number of dropped events does not decrease with a larger buffer size, it may be due to a slow reader preventing the target process' buffers from being flushed.
```

josalem (approved):

PR #23201, opened 2021-03-06
Update tutorial to have an example of Environment.NewLine

BenjaminMichaelis:
## Summary

Having an example of Environment.NewLine I think could be good practice as it is a more versatile option compared to `/n` and is platform independent, and only downside is that it can't be used to define a constant, but that is a more rare scenario, so having an example of it in the tutorial allows new users to get used to a good option for newlines.

### DIFF
--- a/docs/core/tutorials/snippets/with-visual-studio/csharp/Program.cs
+++ b/docs/core/tutorials/snippets/with-visual-studio/csharp/Program.cs
@@ -10,11 +10,11 @@ namespace HelloWorld
         public static void Main(string[] args)
         {
             // <MainMethod>
-            Console.WriteLine("\nWhat is your name? ");
+            Console.WriteLine($"{Environment.NewLine}What is your name? ");
             var name = Console.ReadLine();
             var date = DateTime.Now;
             Console.WriteLine($"{Environment.NewLine}Hello, {name}, on {date:d} at {date:t}!");
-            Console.Write("\nPress any key to exit...");
+            Console.Write($"{Environment.NewLine}Press any key to exit...");
             Console.ReadKey(true);
             // </MainMethod>
         }
--- a/docs/core/tutorials/snippets/with-visual-studio/csharp/Program.cs
+++ b/docs/core/tutorials/snippets/with-visual-studio/csharp/Program.cs
@@ -13,7 +13,7 @@ namespace HelloWorld
             Console.WriteLine("\nWhat is your name? ");
             var name = Console.ReadLine();
             var date = DateTime.Now;
-            Console.WriteLine($"\nHello, {name}, on {date:d} at {date:t}!");
+            Console.WriteLine($"{Environment.NewLine}Hello, {name}, on {date:d} at {date:t}!");
             Console.Write("\nPress any key to exit...");
             Console.ReadKey(true);
             // </MainMethod>
--- a/docs/core/tutorials/snippets/with-visual-studio/vb/Program.vb
+++ b/docs/core/tutorials/snippets/with-visual-studio/vb/Program.vb
@@ -1,11 +1,11 @@
 ﻿Module Program
     Sub Main(args As String())
         ' <MainMethod>
-        Console.WriteLine(vbCrLf + "What is your name? ")
+        Console.WriteLine($"{Environment.NewLine}What is your name? ")
         Dim name = Console.ReadLine()
         Dim currentDate = DateTime.Now
         Console.WriteLine($"{Environment.NewLine}Hello, {name}, on {currentDate:d} at {currentDate:t}")
-        Console.Write(vbCrLf + "Press any key to exit... ")
+        Console.Write($"{Environment.NewLine}Press any key to exit... ")
         Console.ReadKey(True)
         ' </MainMethod>
     End Sub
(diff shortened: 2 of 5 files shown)


docs/core/tutorials/snippets/with-visual-studio/csharp/Program.cs
```diff
@@ -13,7 +13,7 @@ public static void Main(string[] args)
             Console.WriteLine("\nWhat is your name? ");
             var name = Console.ReadLine();
             var date = DateTime.Now;
-            Console.WriteLine($"\nHello, {name}, on {date:d} at {date:t}!");
+            Console.WriteLine($"{Environment.NewLine}Hello, {name}, on {date:d} at {date:t}!");
             Console.Write("\nPress any key to exit...");
```
tdykstra, line 17: For consistency we should change the other two instances of \n to Environment.NewLine also.

docs/core/tutorials/snippets/with-visual-studio/vb/Program.vb
```diff
@@ -4,7 +4,7 @@
         Console.WriteLine(vbCrLf + "What is your name? ")
         Dim name = Console.ReadLine()
         Dim currentDate = DateTime.Now
-        Console.WriteLine($"{vbCrLf}Hello, {name}, on {currentDate:d} at {currentDate:t}")
+        Console.WriteLine($"{Environment.NewLine}Hello, {name}, on {currentDate:d} at {currentDate:t}")
         Console.Write(vbCrLf + "Press any key to exit... ")
```
tdykstra, line 8: For consistency we should change the other two instances of vbCrLf to Environment.NewLine also.

docs/core/tutorials/with-visual-studio.md
```diff
@@ -103,6 +103,7 @@ Enhance the application to prompt the user for their name and display it along w
    This code displays a prompt in the console window and waits until the user enters a string followed by the <kbd>Enter</kbd> key. It stores this string in a variable named `name`. It also retrieves the value of the <xref:System.DateTime.Now?displayProperty=nameWithType> property, which contains the current local time, and assigns it to a variable named `date` (`currentDate` in Visual Basic). And it displays these values in the console window. Finally, it displays a prompt in the console window and calls the <xref:System.Console.ReadKey(System.Boolean)?displayProperty=nameWithType> method to wait for user input.
 
    The `\n` (or `vbCrLf` in the Visual Basic code) represents a newline character.
```
tdykstra, line 105: This line is no longer needed if we change all 3 instances of \n or vbCrLf to Environment.NewLine.

docs/core/tutorials/with-visual-studio.md
```diff
@@ -103,6 +103,7 @@ Enhance the application to prompt the user for their name and display it along w
...
+   The `Environment.NewLine` is another way to represent a newline character, but is platform independent and works in both C# and Visual Basic code. This is further described in the [Environment.NewLine](https://docs.microsoft.com/dotnet/api/system.environment.newline) document.
```
tdykstra, line 106:
```suggestion
   <xref:System.Environment.NewLine> represents a newline character.
```

tdykstra (changes requested):
@BenjaminMichaelis Thanks for this suggestion! For this introductory tutorial I think it would be better to consistently show the preferred way to represent a newline character, so if we make this change I would prefer to change every occurrence of \n and vbCrLf.  Also, if we make the change in this tutorial we need to make the corresponding change to explanatory text in the tutorials for [Visual Studio 2019](https://docs.microsoft.com/dotnet/core/tutorials/with-visual-studio) and [Visual Studio for Mac](https://docs.microsoft.com/dotnet/core/tutorials/with-visual-studio-mac).

docs/core/tutorials/with-visual-studio.md
BenjaminMichaelis: Is there another place you think I should add details in the tutorial about what `Environment.NewLine` is and what it does in a short summary, as I'm thinking if you are coming from something like C/C++ you may be used to seeing `\n` and not `Environment.NewLine` and might not know what the advantage to this way of representing a newline character vs `\n`.
As well as this, having a comment to say something like "We use `Environment.NewLine` instead of `\n` or `vbCrLf` for __this___ reason" could be helpful to guide usage towards using `Environment.NewLine` even if they see `\n` or `vbCrLf` elsewhere.

tdykstra: That makes sense, how about something closer to your original text, like this:

> <xref:System.Environment.NewLine> is a platform-independent and language-independent way to represent a line break. Alternatives are `\n` in C# and `vbCrLf` in Visual Basic.

For more details, the reader can click on the API ref link.

docs/core/tutorials/snippets/with-visual-studio/vb/Program.vb
```diff
@@ -1,11 +1,11 @@
 ﻿Module Program
     Sub Main(args As String())
         ' <MainMethod>
-        Console.WriteLine(vbCrLf + "What is your name? ")
```
BenjaminMichaelis, line 4: Removed string concatenation to opt for string interpolation instead

tdykstra (approved):
Thanks for this contribution, @BenjaminMichaelis !

tdykstra:
@BenjaminMichaelis It just occurred to me that there is no need for a line break in the first Console.WriteLine -- would you like to submit another PR to remove it?  If not, no problem, I can do it.

BenjaminMichaelis:
Good point! Updated in #23221

PR #23204, opened 2021-03-07
Update wcf-web-http-service-help-page.md

Niels-V:
This doc was claiming the config value to enable the helppage is  `enabledHelp` but it seems to be `helpEnabled`. See also this page with the correct config in it: https://docs.microsoft.com/en-us/dotnet/api/system.servicemodel.description.webhttpbehavior.helpenabled?view=netframework-4.6.1

## Summary

Fixed the wrong config attribute in the article to the correct one.


BillWagner (approved):
Thanks @Niels-V 

I apologize for the delay. I'll :shipit: now.

PR #23138, opened 2021-03-05
CA1034 rule mentioned in nested types description

BartoszKlonowski:
This pull request fixes #20784 

It mentiones the *CA1034* compiler warning to the nested types topic's description.

---

The *CA1034* has been mentioned as a subpoint of the access modifiers section, just like the *CS0628* compiler warning.
The reason for that is the fact that *CS1034* is all about the nested types accesses.

The additional change done to the nested types description is that the *CS1034* is linked within the "See also" section.

PR #23216, opened 2021-03-08
Update nested-types.md

Youssef1313:
Follow-up to #23138


BillWagner (approved):
Thanks @Youssef1313 

Great catch, and a quick fix.

I'll :shipit: now.

PR #23231, opened 2021-03-09
Remove multiple internal feeds

gewarren:
NuGet Security Analysis found 1 vulnerable package manifest in the repository. Visit https://aka.ms/nugetmultifeed for more details.

### DIFF
--- a/docs/csharp/roslyn-sdk/tutorials/snippets/how-to-write-csharp-analyzer-code-fix/NuGet.config
+++ b/docs/csharp/roslyn-sdk/tutorials/snippets/how-to-write-csharp-analyzer-code-fix/NuGet.config
@@ -3,7 +3,6 @@
   <packageSources>
     <clear />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
-    <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
   </packageSources>
   <disabledPackageSources>
     <clear />


docs/csharp/roslyn-sdk/tutorials/snippets/how-to-write-csharp-analyzer-code-fix/NuGet.config
```diff
@@ -3,7 +3,6 @@
   <packageSources>
     <clear />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
-    <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
```
Youssef1313, line 6: This is certainly a false positive. The removal of the package will break the sample.

Youssef1313:
@gewarren Yes it should, and it should fail.

docs/csharp/roslyn-sdk/tutorials/snippets/how-to-write-csharp-analyzer-code-fix/NuGet.config
gewarren: @Youssef1313 Can we configure it differently? It's blocking our publishing pipeline.

Youssef1313: There is currently no way. The appropriate fix should be from roslyn-sdk side when the package is published on nuget.org.

BillWagner (approved):
I'm good with this, once we resolve the build concerns.

IEvangelist:
> Should SNIPPETS 5000 have run here?

No, it is only configured to run on certain files. [See here](https://github.com/dotnet/docs/blob/main/.github/workflows/build-validation.yml#L9-L21) for the listing. Maye you should add a _nuget.config_ entry?

Youssef1313:
@IEvangelist I think it's not the only needed change. See https://github.com/dotnet/samples/pull/3746#issuecomment-661246784

IEvangelist:
> @IEvangelist David Pine FTE I think it's not the only needed change. See [dotnet/samples#3746 (comment)](https://github.com/dotnet/samples/pull/3746#issuecomment-661246784)

Hi @Youssef1313 - I'm well aware my friend. We might end up going with an exclusion listing (negation) rather than include paths, might be simpler to say when we don't want it running versus all the things that we'd like it to trigger on.

Youssef1313:
I agree it sounds better to go for an exclusion list. :tada:

gewarren:
PR is incorrect. Also, I was able to add a variable to the pipelines so the build doesn't fail.

PR #22797, opened 2021-02-11
Reimagine C# compiler options section

BillWagner:
Fixes #22659 (for this task)

**Currently open issues on the compiler options sections**

Fixes #20061 
Fixes #20162
Fixes #18371
Fixes #22460
Fixes #6773

**Closed as "won't fix" articles on the compiler options sections that were fixed in this PR**

Fixes #17370
Fixes #4192

***Guidance for reviewers:***

The work done here is to consolidate all the compiler options pages into groups. Instead of a single page for each option, this organization groups options by usage. In addition, the existing pages all used the "classic" syntax. The updates focus on the new MSBuild syntax, but provides the *csc* alternative for existing users and translation to the updated syntax.

This is a large PR, but it's focused in concept. One option is to review it commit by commit. Each of the early commits simplify one section. Later commits fix links to the files that were deleted.

Another option is to review the files by type of change:

- *deleted files*: These are the old option files. Content was moved into one of the combined files.
- *new files*: These are the new combined option files. Each has a theme. The main content change was to remove redundant content, and update to focus on the new MSBuild syntax.
- *update link destinations*: The changes in other areas of the .NET docs are links to C# compiler options. These have been updated to point to the updated file and anchor.

*Notes on changes*:

Most of the changes involved moving content, and trimming that content. In addition old options were translated to the new MSBuild options. There are a few exceptions to this: the `-bugreport` option is no longer supported. The `errorreport` option should be used instead. The `recurse` option is no longer supported, and was removed. Most links were updated to use the new MSBuild syntax. A couple places that weren't were for compiler errors where the order of the compiler options impacted the error. These links continued to use the old syntax.

[Review site link](https://review.docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/?branch=pr-en-us-22797) This goes to the index for the section with new content.

PR #23234, opened 2021-03-09
C# Programming guide: Top-level statements

### LINKED ISSUE #20586: Language reference: Top level statements
The page for `Main` should include the new valid entry points for a C# program. This should include the restrictions on top-level statements and multiple files. This should also include how the synthesized `Main` method responds to the presence of `await` expressions in top-level statements. It should also include how to access `args` and return an `int` value.

tdykstra:
Fixes #20586


docs/csharp/programming-guide/main-and-command-args/how-to-display-command-line-arguments.md
```diff
@@ -1,14 +1,14 @@
...
-ms.date: 07/20/2015
+ms.date: 03/08/2021
 helpviewer_keywords: 
   - "command-line arguments [C#], displaying"
 ms.assetid: b8479f2d-9e05-4d38-82da-2e61246e5437
 ---
 # How to display command-line arguments (C# Programming Guide)
 
-Arguments provided to an executable on the command line are accessible through an optional parameter to `Main`. The arguments are provided in the form of an array of strings. Each element of the array contains one argument. White-space between arguments is removed. For example, consider these command-line invocations of a fictitious executable:  
+Arguments provided to an executable on the command line are accessible in [top-level statements](top-level-statements.md) and through an optional parameter to `Main`. The arguments are provided in the form of an array of strings. Each element of the array contains one argument. White-space between arguments is removed. For example, consider these command-line invocations of a fictitious executable:
```
Youssef1313, line 11: Is "and" extra?

docs/csharp/programming-guide/main-and-command-args/index.md
```diff
@@ -21,6 +21,12 @@ There can only be one entry point in a C# program. If you have more than one cla
 
 [!code-csharp[csProgGuideMain#17](~/samples/snippets/csharp/VS_Snippets_VBCSharp/csProgGuideMain/CS/Class1.cs#17)]
 
+Starting in C# 9, you can omit the `Main` method, and write C# statements as if they were in the `Main` method, as in the following example:
+
+:::code language="csharp" source="snippets/top-level-statements-1/Program.cs":::
+
+For information about how to write application code without a `Main` method, see [Top-level statements](top-level-statements.md).
```
Youssef1313, line 28: "with an implicit" may be more accurate?

docs/csharp/programming-guide/main-and-command-args/snippets/top-level-statements-3/top-level-statements.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Folder Include="Properties\" />
+  </ItemGroup>
```
Youssef1313, line 10:
```suggestion

```

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
```diff
@@ -0,0 +1,90 @@
+---
+title: "Top-level statements - C# Programming Guide"
```
Youssef1313, line 2: The feature is also known as "simple programs" if I recall correctly. You may want to include this name for SEO purposes.

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
```diff
@@ -0,0 +1,90 @@
...
+description: Learn about top-level statements in C# 9 and later.
+ms.date: 03/08/2021
+helpviewer_keywords:
+  - "C# language, top-level statements"
+  - "C# language, Main method"
+---
+# Top-level statements (C# Programming Guide)
+
+Starting in C# 9, you don't have to explicitly include a `Main` method in a console application project. Instead, you can use the *top-level statements* feature to minimize the code you have to write. In this case, the compiler provides an implicit class and `Main` method entry point for the application.
```
Youssef1313, line 11: I personally lean towards using main without backticks instead of `Main`. Because the method the compiler generates isn't called `Main`, but `<Main>$`.

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
```diff
@@ -0,0 +1,90 @@
...
+
+:::code language="csharp" source="snippets/top-level-statements-4/Program.cs":::
+
+## Implicit `Main` method signature
+
+The compiler generates an implicit `Main` method with a signature that depends on what code is found within the top-level statements.
+
+| Top-level code contains| Implicit `Main` signature                    |
+|------------------------|----------------------------------------------|
+| `await` and `return`   | `static async Task<int> Main(string[] args)` |
```
Youssef1313, line 77: The name the compiler generates is `<Main>$`, not `Main`. If this shouldn't be pointed in this docs, I'd replace the second column with "Return type" and "is async" colums.

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
```diff
@@ -0,0 +1,90 @@
...
+| `await`                | `static async Task Main(string[] args)`      |
+| `return`               | `static int Main(string[] args)`             |
+| No `await` or `return` | `static void Main(string[] args)`            |
```
Youssef1313, line 80: Not sure if this should be "nor"?

docs/csharp/programming-guide/main-and-command-args/index.md
tdykstra: Agreed, that may be clearer, I'll change it.

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
tdykstra: Where did you see this name?

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
tdykstra: I think "implicit" clarifies the status of `Main`, but in this instance it will be clearer to repeat "implicit" directly before Main.

Youssef1313: I don't think "implicit" resolves this. It could be implicit and called `Main`, but the compiler implementation decided it's called `<Main>$`.

And actually, from a compiler-point of view, it's not implicit at all (see discussion at https://github.com/dotnet/roslyn/pull/48199), but it's okay to mention implicit for the purpose of this article I think?

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
tdykstra: Acrolinx doesn't like "nor" otherwise this could be neither ... nor.

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
tdykstra: The public messaging has been consistently to call this feature "top-level statements," and with "simple programs" showing up only in the Roslyn repo, I think using that as an alternate feature name would be unnecessarily confusing, but I'll follow the lead of the top-level statements tutorial and work in the phrase "simple programs" into the text.

tdykstra:
@Youssef1313 Thanks for the preliminary review (this is still a draft PR) -- what do you think of my rewrite of the **Implicit `Main` method signature** section intro text? I also am considering removing this section as it's non-necessary background information. I added it thinking it might help to understand one aspect of what's going on behind the scenes.

docs/csharp/programming-guide/classes-and-structs/methods.md
```diff
@@ -1,15 +1,17 @@
...
   - "methods [C#]"
   - "C# language, methods"
 ms.assetid: cc738f07-e8cd-4683-9585-9f40c0667c37
 ---
 # Methods (C# Programming Guide)
 
-A method is a code block that contains a series of statements. A program causes the statements to be executed by calling the method and specifying any required method arguments. In C#, every executed instruction is performed in the context of a method. The `Main` method is the entry point for every C# application and it's called by the common language runtime (CLR) when the program is started.
+A method is a code block that contains a series of statements. A program causes the statements to be executed by calling the method and specifying any required method arguments. In C#, every executed instruction is performed in the context of a method.
+
+The `Main` method is the entry point for every C# application and it's called by the common language runtime (CLR) when the program is started. In an application that uses [top-level statements](../main-and-command-args/top-level-statements.md), the `Main` method is implicit.
```
BillWagner, line 14: I'm fine with how you wrote this. Consider this wording. It feels more correct, but it's also quite a bit more wordy and may be a distraction in this article.

```suggestion
The `Main` method is the entry point for every C# application and it's called by the common language runtime (CLR) when the program is started. In an application that uses [top-level statements](../main-and-command-args/top-level-statements.md), the `Main` method is generated by the compiler and contains all top-level statements.
```

docs/csharp/programming-guide/main-and-command-args/main-return-values.md
```diff
@@ -17,7 +17,12 @@ It can also return an `int`:
 
  [!code-csharp[csProgGuideMain#13](~/samples/snippets/csharp/VS_Snippets_VBCSharp/csProgGuideMain/CS/Class3.cs#13)]
 
-If the return value from `Main` is not used, returning `void` allows for slightly simpler code. However, returning an integer enables the program to communicate status information to other programs or scripts that invoke the executable file. The return value from `Main` is treated as the exit code for the process. If `void` is returned from `Main`, the exit code will be implicitly `0`. The following example shows how the return value from `Main` can be accessed.
+If the return value from `Main` is not used, returning `void` allows for slightly simpler code. However, returning an integer enables the program to communicate status information to other programs or scripts that invoke the executable file.
```
BillWagner, line 20: Not related to your PR, but this article should include that `Main` can return `Task` or `Task<int>`.

If we don't add it now, we should make an issue for that.

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
BillWagner: "Top level statements" is the official name. The product team followed our lead here to avoid the use of "simple". The goal is that programs that use "top level statements" need not be "simple".  See [FeatherApp](https://github.com/featherhttp) for one example of possibilities.

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
```diff
@@ -0,0 +1,90 @@
...
+---
+# Top-level statements (C# Programming Guide)
+
+Starting in C# 9, you don't have to explicitly include a `Main` method in a console application project. Instead, you can use the *top-level statements* feature to minimize the code you have to write. In this case, the compiler implicitly provides a class and `Main` method entry point for the application.
+
+Here's a *Program.cs* file that is a complete C# program in C# 9:
+
+:::code language="csharp" source="snippets/top-level-statements-1/Program.cs":::
+
+Top-level statements let you write simple programs for small utilities such as Azure Functions and GitHub Actions. They also make it simpler for new C# programmers to get started learning and writing code.
```
BillWagner, line 17: This use of "simple programs" is fine because it does flow naturally with the concept. I like how you've done this Tom.

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
```diff
@@ -0,0 +1,90 @@
...
+
+A project can have any number of additional source code files that don't have top-level statements.
+
+## No other entry points
+
+You can write a `Main` method explicitly, but it can't function as an entry point. The compiler issues the following warning:
+
+> CS7022 The entry point of the program is global code; ignoring 'Main()' entry point.
+
+In a project with top-level statements, you can't use the [-main](../../language-reference/compiler-options/main-compiler-option.md) compiler option to select the entry point, even if the project has one or more `Main` methods.
```
BillWagner, line 35: Note for me, depending on how soon we merge this: I'm update the options section to use the new MSBuild properties rather than the `csc` spellings. I'll have to update his depending on the order of merging. See #22797 for the work in progress.

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
```diff
@@ -0,0 +1,90 @@
...
+
+To return an `int` value when the application ends, use the `return` statement as you would in a `Main` method that returns an `int`. For example:
+
+:::code language="csharp" source="snippets/top-level-statements-5/Program.cs":::
+
+## Implicit entry point method
+
+The compiler generates a method to serve as the program entry point for a project with top-level statements. The name of this method isn't actually `Main`, it's an implementation detail that your code can't reference directly. The signature of the method depends on whether the top-level statements contain the `await` keyword or the `return` statement. The following table shows what the method signature would look like, using the method name `Main` in the table for convenience.
+
+| Top-level code contains| Implicit `Main` signature                    |
```
BillWagner, line 75: I like this table.  See my previous comment about the return value of `Main`. Adding this table there would be perfect.

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
```diff
@@ -0,0 +1,90 @@
...
+|------------------------|----------------------------------------------|
+| `await` and `return`   | `static async Task<int> Main(string[] args)` |
+| `await`                | `static async Task Main(string[] args)`      |
+| `return`               | `static int Main(string[] args)`             |
+| No `await` or `return` | `static void Main(string[] args)`            |
+
+## C# language specification
+
+[!INCLUDE[CSharplangspec](~/includes/csharplangspec-md.md)]
+
```
BillWagner, line 85: I'd add a link to the [top level statements speclet](https://docs.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-9.0/top-level-statements). The link would be `~/_csharplang/proposals/csharp-9.0/top-level-statements.md`

BillWagner (approved):
This is excellent work @tdykstra 

I had a few small comments, and then this is ready to merge.

Thanks again.

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
```diff
@@ -0,0 +1,90 @@
...
+title: "Top-level statements - C# Programming Guide"
+description: Learn about top-level statements in C# 9 and later.
+ms.date: 03/08/2021
+helpviewer_keywords:
+  - "C# language, top-level statements"
+  - "C# language, Main method"
+---
+# Top-level statements (C# Programming Guide)
+
+Starting in C# 9, you don't have to explicitly include a `Main` method in a console application project. Instead, you can use the *top-level statements* feature to minimize the code you have to write. In this case, the compiler implicitly provides a class and `Main` method entry point for the application.
```
Youssef1313, line 11: Consider applying similar change to what @BillWagner suggested here. This aligns with the *IsImplicitlyDeclared* compiler API being false, and also aligns with the fact that these have `[CompilerGenerated]` attribute.

```suggestion
Starting in C# 9, you don't have to explicitly include a `Main` method in a console application project. Instead, you can use the *top-level statements* feature to minimize the code you have to write. In this case, the compiler generates a class and `Main` method entry point for the application.
```

docs/csharp/programming-guide/main-and-command-args/main-return-values.md
tdykstra: Adding it now.

docs/csharp/programming-guide/main-and-command-args/top-level-statements.md
tdykstra: Adding a slightly modified table to both the return value and command-line args docs.

PR #23277, opened 2021-03-10
C# patterns reference: relational patterns and pattern combinators

pkulikov:
Contributes to #20584

### DIFF
--- a/docs/csharp/language-reference/operators/patterns.md
+++ b/docs/csharp/language-reference/operators/patterns.md
@@ -98,7 +98,7 @@ Beginning with C# 9.0, you use the `not`, `and`, and `or` pattern combinators to
 
   :::code language="csharp" source="snippets/patterns/PatternCombinators.cs" id="NotPattern":::
 
-- *Conjunctive* `and` pattern that matches an expression when both patterns match the expression. For example, you can combine [relational patterns](#relational-patterns) to check if a value is at a certain range, as the following code shows:
+- *Conjunctive* `and` pattern that matches an expression when both patterns match the expression. For example, you can combine [relational patterns](#relational-patterns) to check if a value is in a certain range, as the following code shows:
 
   :::code language="csharp" source="snippets/patterns/PatternCombinators.cs" id="AndPattern":::
 
--- a/docs/csharp/language-reference/operators/patterns.md
+++ b/docs/csharp/language-reference/operators/patterns.md
@@ -1,7 +1,7 @@
 ---
 title: "Patterns - C# reference"
 description: "Learn about the patterns supported by C# switch and is expressions and C# switch statement - C# reference"
-ms.date: 03/02/2021
+ms.date: 03/10/2021
 helpviewer_keywords: 
   - "pattern matching [C#]"
 ---
@@ -62,6 +62,10 @@ Use the constant pattern to check for `null`, as the following example shows:
 
 The compiler guarantees that no user-overloaded equality operator `==` is invoked when expression `x is null` is evaluated.
 
+Beginning with C# 9.0, you can use the [negated](#pattern-combinators) `null` constant pattern to check for non-null, as the following example shows:
+
+:::code language="csharp" source="snippets/patterns/ConstantPattern.cs" id="NonNullCheck":::
+
 ## Discard pattern
 
 Beginning with C# 8.0, you use the *discard pattern* `_` to match any expression, including `null`, as the following example shows:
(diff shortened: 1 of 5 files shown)


docs/csharp/language-reference/operators/patterns.md
```diff
@@ -72,6 +76,40 @@ The discard pattern is useful in `switch` expressions. If none of a `switch` exp
 
 The discard pattern cannot be a pattern in an `is` expression or a `switch` statement.
 
+## Relational patterns
```
BillWagner, line 79: In this section, is it worth highlighting that the right side of the expression *must* be a constant? That restriction is the reason relational patterns are limited to the builtin numeric types. (Specifically, user defined types that define relational operators can't be used in relational patterns).

docs/csharp/language-reference/operators/patterns.md
```diff
@@ -72,6 +76,40 @@ The discard pattern is useful in `switch` expressions. If none of a `switch` exp
...
+
+Beginning with C# 9.0, you use a *relational pattern* to compare an expression result with a constant, as the following example shows:
+
+:::code language="csharp" source="snippets/patterns/RelationalPatterns.cs" id="BasicExample":::
+
+In a relational pattern, you can use any of the [relational operators](comparison-operators.md) `<`, `>`, `<=`, or `>=`. A constant to compare with can be of an [integral](../builtin-types/integral-numeric-types.md), [floating-point](../builtin-types/floating-point-numeric-types.md), [char](../builtin-types/char.md), or [enum](../builtin-types/enum.md) type.
```
BillWagner, line 85: You can also use the nullable value types that correspond:

```suggestion
In a relational pattern, you can use any of the [relational operators](comparison-operators.md) `<`, `>`, `<=`, or `>=`. A constant to compare with can be of an [integral](../builtin-types/integral-numeric-types.md), [floating-point](../builtin-types/floating-point-numeric-types.md), [char](../builtin-types/char.md), or [enum](../builtin-types/enum.md) type, or the corresponding nullable value type.
```

docs/csharp/language-reference/operators/patterns.md
```diff
@@ -72,6 +76,40 @@ The discard pattern is useful in `switch` expressions. If none of a `switch` exp
...
+
+To check if an expression value is at a certain range, match it against a [combined pattern](#pattern-combinators), as the following example shows:
+
+:::code language="csharp" source="snippets/patterns/RelationalPatterns.cs" id="WithCombinators":::
+
+A relational pattern doesn't match an expression if an expression result is `null` or it fails to convert to the type of a constant by a nullable or unboxing conversion.
```
BillWagner, line 91: See above 😄

docs/csharp/language-reference/operators/patterns.md
```diff
@@ -72,6 +76,40 @@ The discard pattern is useful in `switch` expressions. If none of a `switch` exp
...
+
+## Pattern combinators
+
+Beginning with C# 9.0, you use the `not`, `and`, and `or` pattern combinators to create the following patterns from other patterns:
+
+- *Negation* `not` pattern that matches an expression when the negated pattern doesn't match the expression. For example, you can negate the [constant](#constant-pattern) `null` pattern to check if an expression is non-null, as the following code shows:
+
+  :::code language="csharp" source="snippets/patterns/PatternCombinators.cs" id="NotPattern":::
+
+- *Conjunctive* `and` pattern that matches an expression when both patterns match the expression. For example, you can combine [relational patterns](#relational-patterns) to check if a value is at a certain range, as the following code shows:
```
BillWagner, line 101: I think "in" would be better than "at" here:

```suggestion
- *Conjunctive* `and` pattern that matches an expression when both patterns match the expression. For example, you can combine [relational patterns](#relational-patterns) to check if a value is in a certain range, as the following code shows:
```

docs/csharp/language-reference/operators/snippets/patterns/PatternCombinators.cs
```diff
@@ -0,0 +1,64 @@
...
+            Console.WriteLine(GetCalendarSeason(new DateTime(2021, 1, 19)));  // output: winter
+            Console.WriteLine(GetCalendarSeason(new DateTime(2021, 10, 9)));  // output: autumn
+            Console.WriteLine(GetCalendarSeason(new DateTime(2021, 5, 11)));  // output: spring
+
+            static string GetCalendarSeason(DateTime date) => date.Month switch
+            {
+                3 or 4 or 5 => "spring",
+                6 or 7 or 8 => "summer",
+                9 or 10 or 11 => "autumn",
+                12 or 1 or 2 => "winter",
```
BillWagner, line 54: I really like this example! well done.

docs/csharp/language-reference/operators/snippets/patterns/RelationalPatterns.cs
```diff
@@ -0,0 +1,48 @@
...
+            Console.WriteLine(GetCalendarSeason(new DateTime(2021, 3, 14)));  // output: spring
+            Console.WriteLine(GetCalendarSeason(new DateTime(2021, 7, 19)));  // output: summer
+            Console.WriteLine(GetCalendarSeason(new DateTime(2021, 2, 17)));  // output: winter
+
+            static string GetCalendarSeason(DateTime date) => date.Month switch
+            {
+                >= 3 and <= 5 => "spring",
+                >= 6 and <= 8 => "summer",
+                >= 9 and <= 11 => "autumn",
+                12 or (>= 1 and <= 2) => "winter",
```
BillWagner, line 42: Style concern where I don't have strong opinions.  I might consider this form instead:

```suggestion
                >= 3 and < 6 => "spring",
                >= 6 and < 9 => "summer",
                >= 9 and < 12 => "autumn",
                12 or (>= 1 and < 3) => "winter",
```

BillWagner (commented):
This is really well done @pkulikov 

I had a few small comments, many of which are optional. 

I also had one discussion question we should resolve, and then this is ready to merge.

Thanks again for taking this on.

docs/csharp/language-reference/operators/patterns.md
pkulikov: @BillWagner to confirm I understand you correctly: you suggest to explicitly mention that the right-hand side must be a constant? If yes, I agree and will update accordingly.

docs/csharp/language-reference/operators/patterns.md
pkulikov: Hm, constant at the right-hand side cannot be null. But can it be of a nullable type? This sentence is about the type of constant, not expression. I'll check that, thank you.

BillWagner: You're right. I misread this originally. Consider this update instead:

```suggestion
In a relational pattern, you can use any of the [relational operators](comparison-operators.md) `<`, `>`, `<=`, or `>=`. A constant to compare with can be of an [integral](../builtin-types/integral-numeric-types.md), [floating-point](../builtin-types/floating-point-numeric-types.md), [char](../builtin-types/char.md), or [enum](../builtin-types/enum.md) type. When the left side of the comparison is a nullable value type, the constant on the right side is converted to the corresponding nullable type.
```

pkulikov: That sounds too technical. Why to mention nullable value type, but not mention `object`, for example? And actually I  think it's not the constant that is converted, but the expression result is unwrapped to the underlying type.

BillWagner: You're right. I didn't like the way it sounded either.

I may be overthinking this, and leaving it exactly as you wrote it is best.

docs/csharp/language-reference/operators/patterns.md
pkulikov: @BillWagner I've updated the paragraph after the first example. Please check how it reads now.

pkulikov:
@BillWagner thank you for review and suggestions. I think I've addressed all the feedback.

BillWagner:
Thanks @pkulikov 

I'll :shipit:

PR #23330, opened 2021-03-15
Protected Friend

gewarren:
- Fixes capitalization
- Removes incorrect info about applying only at class level. `Protected Friend` can also be applied to members of classes according to https://github.com/dotnet/vblang/blob/19ec79a287fb79180b05a0ad20e8291e75fc63df/spec/types.md#class-members.


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

PR #23339, opened 2021-03-16
 Remove "input" as a qualifier to "argument"

### LINKED ISSUE #23327: Post condition attributes act on output arguments, not input
There are several instances of these summaries:

MaybeNullWhen: A non-nullable input argument may be null when the method returns the specified bool value.
NotNullWhen: A nullable input argument won't be null when the method returns the specified bool value.

I am fairly sure that should be **output** argument?

---
#### Document Details

* ID: 9ae1afa1-7c5f-862b-08f9-23cf5928953e
* Version Independent ID: 9ae2922b-5108-ed50-768b-30f08cc3fc30
* Content: [C# Reserved attributes: Nullable static analysis](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis)
* Content Source: [docs/csharp/language-reference/attributes/nullable-analysis.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/attributes/nullable-analysis.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixes #23327

A reader was concerned that these attributes could not be applied to arguments that might have the `out` qualifier. The word is redundant, so remove it from every location where "input" qualified either "parameter" or "argument". In other cases, where "input" was used without "argument", replace "input" with "argument". That's more correct and more consistent.

### DIFF
--- a/docs/csharp/language-reference/attributes/nullable-analysis.md
+++ b/docs/csharp/language-reference/attributes/nullable-analysis.md
@@ -30,12 +30,12 @@ Several attributes have been added to express additional information about the n
 
 The rules for your APIs are likely more complicated, as you saw with the `TryGetValue` API scenario. Many of your APIs have more complex rules for when variables can or can't be `null`. In these cases, you'll use one of the following attributes to express those rules:
 
-- [AllowNull](xref:System.Diagnostics.CodeAnalysis.AllowNullAttribute): A non-nullable input argument may be null.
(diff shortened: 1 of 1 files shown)


IEvangelist (approved):
They are arguments or parameters depending on the context. Not inputs, I like these changes...

PR #23368, opened 2021-03-17
Fix build warning

gewarren:


BillWagner (approved):
Thanks for spotting this and fixing it.  

I shouldn't have missed it in that last PR.

PR #23369, opened 2021-03-17
Empty CODEOWNERS file in live branch only

gewarren:
So reviewers don't get tagged on main -> live publish PRs.


BillWagner:
I like the idea, but is this the right fix? My concern is that whenever we update the CODEOWNERS in the main branch, we'll overwrite this one.

That happens with some regularity across our repos.

gewarren:
> I like the idea, but is this the right fix? My concern is that whenever we update the CODEOWNERS in the main branch, we'll overwrite this one.

That would only be an issue for manual publishes after CODEOWNERS is updated in main. The pipelines are configured to exclude the CODEOWNERS file from the merge.

BillWagner (approved):
This LGTM @gewarren 

Let's :shipit:

PR #23407, opened 2021-03-20
with expression: updated the link to the record reference

pkulikov:


BillWagner (approved):
Thanks for making this change @pkulikov 

It looks great, and I'll :shipit: now.

Issue #23356, opened 2021-03-17
"must" is imprecise, reason should be provided


goyzhang:
> To implement multiple event properties in a class, the class **must** internally store and maintain the delegate defined for each event. A typical approach is to implement a delegate collection that is indexed by an event key.

Why must? If it IS a must, then the reason should be provided. If the reason is just to reduce backing fields to improve performance, then the "must" should be changed to "suggested", with a reason explained.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e6a5f33d-52b9-fd2a-ecd0-35582d462f2c
* Version Independent ID: 80c4a287-3ad7-6642-ed89-d63cc71e860b
* Content: [How to: Handle Multiple Events Using Event Properties](https://docs.microsoft.com/en-us/dotnet/standard/events/how-to-handle-multiple-events-using-event-properties)
* Content Source: [docs/standard/events/how-to-handle-multiple-events-using-event-properties.md](https://github.com/dotnet/docs/blob/main/docs/standard/events/how-to-handle-multiple-events-using-event-properties.md)
* Product: **dotnet-fundamentals**
* GitHub Login: @IEvangelist
* Microsoft Alias: **dapine**

IEvangelist:
Hi 👋 @goyzhang
Thank you so much for posting this issue. We really appreciate it.

What other approach are you referring to for this? I'm a bit confused by your statements, but happy to update this accordingly.

goyzhang:
Hi @IEvangelist ,
My point is that the current description is **imprecise**. The **must** word is too strong.  It's better to change it to a suggestion like:

`To implement multiple event properties in a class, the class **should** internally store and maintain the delegate defined for each event. A typical approach is to implement a delegate collection that is indexed by an event key. That's because **[REASON]**`

Actually, before I read this page,  I always added multiple event properties without using an EventHandlerList. And It worked, so I wonder why I have to do this. As for the reason, I'm not an expert about this but I think I got my answer from Marc Gravell : https://stackoverflow.com/questions/11895971/what-are-the-advantages-of-adding-delegate-instances-to-component-events-eventh
You can refer to that.

IEvangelist:
Yes, Mark brings up a great point. I'm fine with "should" instead of "must" but would prefer some additional context as to why. I'm going to label this as up-for-grabs, and I encourage you to contribute a proposal for improving this. Is that something you'd be open to try?

[...]

PR #23414, opened 2021-03-22
Update how-to-handle-multiple-events-using-event-properties.md

goyzhang:
Improve the description for the issue

#23356


docs/standard/events/how-to-handle-multiple-events-using-event-properties.md
```diff
@@ -15,7 +15,7 @@ ms.assetid: 30047cba-e2fd-41c6-b9ca-2ad7a49003db
 ---
 # How to: Handle Multiple Events Using Event Properties
 
-To use event properties, you define the event properties in the class that raises the events, and then set the delegates for the event properties in classes that handle the events. To implement multiple event properties in a class, the class must internally store and maintain the delegate defined for each event. A typical approach is to implement a delegate collection that is indexed by an event key.  
+To use event properties, you define the event properties in the class that raises the events, and then set the delegates for the event properties in classes that handle the events. To implement multiple event properties in a class, the class should internally store and maintain the delegate defined for each event. A typical approach is to implement a delegate collection that is indexed by an event key. Because each field-like-event requires a reference backing-field. And when the number of events increases, a lot of fields will be created even they are not subscribed. This can introduce unnecessary allocations when only a few events are subscribed.
```
IEvangelist, line 18:
```suggestion
To use event properties, you define the event properties in the class that raises the events, and then set the delegates for the event properties in classes that handle the events. To implement multiple event properties in a class, the class should internally store and maintain the delegate defined for each event.

As the number of events increases, so too does the number of backing-field references. To help avoid unnecessary allocations an approach is to implement a delegate collection that is indexed by an event key.
```

IEvangelist (commented):
How do you feel about this wording instead?

docs/standard/events/how-to-handle-multiple-events-using-event-properties.md
goyzhang: "As the number of events increases" is imprecise, as you can see from the code example, even with self-defined add/remove accessors the number of events still increases anyway. The issue only happens when you rely on the complier's syntax sugar, a "field-like-event", which will generate a backing-field:
`public event EventHandler<EventArgs> MyEvent;`

IEvangelist: I updated my suggestion, please have another look - and let me know what you think.

goyzhang: HI, IMO it's generally good😊, but why "must" again? As you can see in you example `Progress`  and `AnotherProgress` generated backing-field, they still work but suffer from the issue when the number of events increases, but `ManualProgress` doesn't.

goyzhang: I've updated the doc according to your suggestion. Take a look

IEvangelist (approved):

Issue #22690, opened 2021-02-05
Switched definition of  Abstraction and Encapsulation


Rashair:
I think that definition of **Abstraction** and **Encapsulation** is switched.
This
```
Abstraction  means hiding the unnecessary details from type consumers.
Encapsulation means that a group of related properties, methods, and other members are treated as a single unit or object.
```
should be this
```
Abstraction  means that a group of related properties, methods, and other members are treated as a single unit or object.
Encapsulation means hiding the unnecessary details from type consumers.
```
Similarly in the summary.
 
---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ece1648c-e8a7-153d-a22b-350a619aa26d
* Version Independent ID: a0c9221a-cc55-4c31-4f54-0e5fbdcbce25
* Content: [Object-Oriented Programming (C#)](https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/intro-to-csharp/object-oriented-programming)
* Content Source: [docs/csharp/tutorials/intro-to-csharp/object-oriented-programming.md](https://github.com/dotnet/docs/blob/master/docs/csharp/tutorials/intro-to-csharp/object-oriented-programming.md)
* Product: **dotnet-csharp**
* Technology: **csharp-get-started**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

stankevichea:
@Rashair is right

BillWagner:
Thanks for bringing this to our attention @Rashair 

I've added to my plans for this month, while I'm working on other issues in this area.

See #22671 for related tasks.

Rashair:
Summary was not fixed :/

PR #23423, opened 2021-03-22
Update object-oriented-programming.md

VladimirtheGreatest:
Not sure if this is right but I am assuming that Encapsulation refers to the bundling of data with the methods that operate on that data while abstraction means hiding the unnecessary details from type consumers.

## Summary

Change the description of the principles of OOP.


BillWagner:
Thanks @VladimirtheGreatest 

I'm closing this based on the discussion in #22690. These definitions were reversed, and the consensus was that the current definitions are correct.

VladimirtheGreatest:
@BillWagner I think you are right, I was really confused there are so many resources about this topic and it is hard to find out which one is the correct one. Thanks for clarification.

PR #23426, opened 2021-03-22
Fix redirect for 2.0 - 2.1 information

adegeo:
## Summary

- Redirect had .md in the target (weird, build should have caught that right?)
- Changed to the breaking changes file the content was ported to.

Thanks @gewarren for spotting the orphaned file!


gewarren (approved):
Thanks for fixing!

PR #23427, opened 2021-03-22
Clarify which object are not properly disposed

### LINKED ISSUE #23400: The explanation in the section "Unacceptable pattern" is not correct.
The explanation of the section "Unacceptable pattern" is not clear and the comment in the code sample is not correct.

If an exception is thrown inside the `AnotherAsyncDisposable` constructor then **none** of the created objects would be properly disposed.

And it is not because of the "stacked usings" pattern itself. But because of the fact that the execution of the function will be terminated even before getting to the first `using`. So none of the constructed objects would be passed to `using` to be disposed.

Another acceptable pattern could look like this:

```cs
ExampleAsyncDisposable objOne;
AnotherAsyncDisposable objTwo;

await using ((objOne = new ExampleAsyncDisposable()).ConfigureAwait(false))
await using ((objTwo = new AnotherAsyncDisposable()).ConfigureAwait(false))
{
	// ...
}
```

But it looks ugly enough.

---
#### Document Details

* ID: 61a961dc-3ebf-2a96-a6eb-2e0b62ab909f
* Version Independent ID: d72276e1-9059-815f-dc14-a03a7544ab87
* Content: [Implement a DisposeAsync method](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync)
* Content Source: [docs/standard/garbage-collection/implementing-disposeasync.md](https://github.com/dotnet/docs/blob/main/docs/standard/garbage-collection/implementing-disposeasync.md)
* GitHub Login: @IEvangelist
* Microsoft Alias: **dapine**

IEvangelist:
## Summary

Corrected the wording in the sample source and the describing text. Both objects are not properly disposed of.

Fixes #23400

## Preview

✔️ [Implement a DisposeAsync method](https://review.docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync?branch=pr-en-us-23427)


docs/standard/garbage-collection/implementing-disposeasync.md
```diff
@@ -121,7 +121,7 @@ In the preceding example, each asynchronous clean up operation is implicitly sco
 
 ### Unacceptable pattern
 
-The highlighted lines in the following code show what it means to have "stacked usings". If an exception is thrown from the `AnotherAsyncDisposable` constructor, then `objOne` does not get properly disposed.
+The highlighted lines in the following code show what it means to have "stacked usings". If an exception is thrown from the `AnotherAsyncDisposable` constructor, both objects are not properly disposed of.
```
BillWagner, line 124: I'd state this a bit differently. The key point is that if `objTwo`'s constructor throws an exception, the constructor must clean up any resources allocated before it throws:

```suggestion
The highlighted lines in the following code show what it means to have "stacked usings". If an exception is thrown from the `AnotherAsyncDisposable` constructor, neither object is properly disposed of. The variable `objTwo` is never assigned because the constructor did not complete successfully. As a result, the constructor for `AnotherAsyncDisposable` is responsible for disposing any resources allocated before it throws an exception. 
```

BillWagner (approved):
This LGTM @IEvangelist 

I have one clarifying suggestion. Then, this is ready to :shipit:

docs/standard/garbage-collection/implementing-disposeasync.md
IEvangelist: Yes, I do like that - thank you!

PR #23431, opened 2021-03-22
Add a value for Deterministic and ProduceOnlyReferenceAssembly examples

akoeplinger:
This makes it consistent with the example for Optimize.


BillWagner (approved):
Good catch @akoeplinger Thank you for making this update.

Thanks again.

Issue #23408, opened 2021-03-20
Update doc to reflect that as of C# 9 partial methods can have accessibility modifiers and return types other than void


maxkoshevoi:
In C# 9 all restrictions are removed from `partial` methods: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/extending-partial-methods

This needs to be reflected on this page.

I can update it if it's ok.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 450a248b-ab3f-ac3f-d3a5-b1e74b3355ba
* Version Independent ID: 5e05ac12-7d29-d144-a49a-e54765343c20
* Content: [Partial Classes and Methods - C# Programming Guide](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/partial-classes-and-methods)
* Content Source: [docs/csharp/programming-guide/classes-and-structs/partial-classes-and-methods.md](https://github.com/dotnet/docs/blob/main/docs/csharp/programming-guide/classes-and-structs/partial-classes-and-methods.md)
* Product: **dotnet-csharp**
* Technology: **csharp-fundamentals**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

Youssef1313:
@maxkoshevoi I believe you can open a PR to update that if you want.

maxkoshevoi:
@Youssef1313 Does issue needs to be approved, before I can open a PR?

Youssef1313:
While I'm not a Microsoft employee, I believe the team will want to have such an enhancement. But you can wait for @BillWagner or @IEvangelist to confirm.

IEvangelist:
Hi @maxkoshevoi &and @Youssef1313)

Yes, this needs to be updated to correctly reflect the changes you called attention to. We would be thrilled to review a PR, and we appreciate it. Please reference this issue number in the PR, and note that I've updated this issue with the up-for-grabs label. Thank you.

BillWagner:
See also #23083 and #23084 for more instances that need to be updated.

tdykstra:
This will also address #20578

PR #23435, opened 2021-03-22
Update partial method specification

### LINKED ISSUE #20578: Language reference: Partial methods
The language reference for `partial` needs to describe the new rules, and describe how the compiler determine a "new" or "old" partial method type because the new style must have an implementation.

### LINKED ISSUE #23083: Out of date description of partial method requirements.
C# 9.0 changed the semantics of partial methods, this page needs to be updated accordingly. New semantics: https://github.com/dotnet/csharplang/blob/master/proposals/csharp-9.0/extending-partial-methods.md

---
#### Document Details

* ID: b4eedbc9-36be-6c79-fcb4-d5fe6998e5b1
* Version Independent ID: 85bf7eb0-b476-94b6-e10a-d0f42fee407b
* Content: [partial method - C# Reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/partial-method)
* Content Source: [docs/csharp/language-reference/keywords/partial-method.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/keywords/partial-method.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

maxkoshevoi:
## Summary

Updated `partial` method specification according to changes made in C# 9

Fixes #20578, #23083, #23408

### DIFF
--- a/docs/csharp/language-reference/keywords/partial-method.md
+++ b/docs/csharp/language-reference/keywords/partial-method.md
@@ -1,7 +1,7 @@
 ---
 description: "partial method - C# Reference"
 title: "partial method - C# Reference"
-ms.date: 07/20/2015
+ms.date: 03/22/2021
 f1_keywords: 
   - "partialmethod_CSharpKeyword"
 helpviewer_keywords: 
@@ -14,9 +14,15 @@ A partial method has its signature defined in one part of a partial type, and it
 
 - Signatures in both parts of the partial type must match.
 
-- The method must return void.
+Partial method must have an implementation part in following cases:
 
-- No access modifiers are allowed. Partial methods are implicitly private.
+- It has any accessibility modifiers (even [private](../../language-reference/keywords/private.md)).
+
+- It has any return type other than [void](../../language-reference/builtin-types/void.md).
+
+- It has [out](../../language-reference/keywords/out-parameter-modifier.md) parameters.
+
+- It has a [virtual](../../language-reference/keywords/virtual.md), [override](../../language-reference/keywords/override.md), [sealed](../../language-reference/keywords/sealed.md), [new](../../language-reference/keywords/new-modifier.md), or [extern](../../language-reference/keywords/extern.md) modifier.
 
 The following example shows a partial method defined in two parts of a partial class:
 
--- a/docs/csharp/language-reference/keywords/partial-method.md
+++ b/docs/csharp/language-reference/keywords/partial-method.md
@@ -1,7 +1,7 @@
 ---
 description: "partial method - C# Reference"
 title: "partial method - C# Reference"
-ms.date: 03/22/2021
+ms.date: 03/23/2021
 f1_keywords: 
   - "partialmethod_CSharpKeyword"
 helpviewer_keywords: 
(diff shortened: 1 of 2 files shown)


docs/csharp/language-reference/keywords/partial-method.md
```diff
@@ -14,9 +14,15 @@ A partial method has its signature defined in one part of a partial type, and it
 
 - Signatures in both parts of the partial type must match.
 
-- The method must return void.
+Partial method must have an implementation part in following cases:
```
BillWagner, line 17: nit:

```suggestion
Partial method must have an implementation part in the following cases:
```

docs/csharp/language-reference/keywords/partial-method.md
```diff
@@ -14,9 +14,15 @@ A partial method has its signature defined in one part of a partial type, and it
...
 
-- No access modifiers are allowed. Partial methods are implicitly private.
+- It has any accessibility modifiers (even [private](../../language-reference/keywords/private.md)).
+
+- It has any return type other than [void](../../language-reference/builtin-types/void.md).
+
+- It has [out](../../language-reference/keywords/out-parameter-modifier.md) parameters.
+
+- It has a [virtual](../../language-reference/keywords/virtual.md), [override](../../language-reference/keywords/override.md), [sealed](../../language-reference/keywords/sealed.md), [new](../../language-reference/keywords/new-modifier.md), or [extern](../../language-reference/keywords/extern.md) modifier.
```
BillWagner, line 25: Here, and in the following change noted, I think it would be easier to define this by stating when a `partial` method *doesn't* need to have an implementation.  Namely, a partial method that returns `void`, and doesn't include any modifiers (including the default `private` access modifier) doesn't require an implementation. 

I like transposing it this way for two reasons: First, it's a simpler definition. Second, it provides a clear mental model to how the new features don't represent a breaking change.

docs/csharp/programming-guide/classes-and-structs/partial-classes-and-methods.md
```diff
@@ -137,29 +137,35 @@ A partial method declaration consists of two parts: the definition, and the impl
...
-- Partial methods cannot be [extern](../../language-reference/keywords/extern.md), because the presence of the body determines whether they are defining or implementing.
+- Partial method signatures in both parts of the partial type must match.
 
 - Partial methods can have [static](../../language-reference/keywords/static.md) and [unsafe](../../language-reference/keywords/unsafe.md) modifiers.
 
 - Partial methods can be generic. Constraints are put on the defining partial method declaration, and may optionally be repeated on the implementing one. Parameter and type parameter names do not have to be the same in the implementing declaration as in the defining one.
 
 - You can make a [delegate](../../language-reference/builtin-types/reference-types.md) to a partial method that has been defined and implemented, but not to a partial method that has only been defined.
 
+Partial method must have an implementation part in following cases:
```
BillWagner, line 159: See my previous comment.

BillWagner (changes requested):
Thanks for making this update @maxkoshevoi We appreciate it.

I have some comments for you to consider, and then this is ready to merge. 

Thanks again!

docs/csharp/language-reference/keywords/partial-method.md
maxkoshevoi: A partial method doesn't need to have an implementation if it returns `void`, doesn't have any `out` parameters, doesn't include any modifiers (including the default private access modifier) and doesn't have `extern` modifier. 

So condition for "optional implementation" is also not that simple.

BillWagner: @maxkoshevoi 

> So condition for "optional implementation" is also not that simple.

It's not that simple, but it is simpler than what's written. And, my second point still applies. Effectively, partial methods don't need an implementation if they follow the rules in place before C# 9.0

maxkoshevoi: @BillWagner Yes, I see your point, but don't quite get, how you want to phrase it.

Something like:

> Implementation part is optional, if all of the conditions are met:
>
> - Partial method doesn't have any accessibility modifiers (even [private](../../language-reference/keywords/private.md)).
> - Partial method returns [void](../../language-reference/builtin-types/void.md).
> - Partial method doesn't have any [out](../../language-reference/keywords/out-parameter-modifier.md) parameters.
> - Partial method doesn't have any of the following modifiers: [virtual](../../language-reference/keywords/virtual.md), [override](../../language-reference/keywords/override.md), [sealed](../../language-reference/keywords/sealed.md), [new](../../language-reference/keywords/new-modifier.md), or [extern](../../language-reference/keywords/extern.md).

Or do you want to duplicate conditions like this:

>  A partial method doesn't need to have an implementation if it returns void, doesn't have any out parameters, doesn't include any modifiers (including the default private access modifier) and doesn't have extern modifier.
>
> Partial method must have an implementation part in the following cases:
>
> It has any accessibility modifiers
> ...

BillWagner: Thanks @maxkoshevoi.  I think a mix of the two reads best. Let me know your thoughts.  Something like:

> A `partial` method doesn't need to have an implementation if it returns `void`, doesn't have any `out` parameters, doesn't include any modifiers (including the default `private` access modifier) and doesn't have the `extern` modifier.
>
> Any method that doesn't conform to all those restrictions (for example, `virtual public` methods), must provide an implementation.

maxkoshevoi: @BillWagner Looks good! I wanted to add those restrictions as a list for better readability. Updated PR. What do you think about correct version?

PS: Also added example of using partial method with source generator (since those changes were made mainly for that reason)

BillWagner:
Thanks for making these updates @maxkoshevoi I like the final result.  I'll :shipit: now.

BillWagner (approved):

PR #23466, opened 2021-03-23
Update best-practices.md for cross-platform C long

AaronRobinsonMSFT:
Add section about cross-platform concerns. Provide example for the C long issue.

/cc @jkoritzinsky @elinor-fung @danmoseley

### DIFF
--- a/docs/standard/native-interop/best-practices.md
+++ b/docs/standard/native-interop/best-practices.md
@@ -169,29 +169,29 @@ Here is a list of data types commonly used in Windows APIs and which C# types to
 
 The following types are the same size on 32-bit and 64-bit Windows, despite their names.
 
-| Width | Windows          | C (Windows)          | C#       | Alternative                          |
-|:------|:-----------------|:---------------------|:---------|:-------------------------------------|
-| 32    | `BOOL`           | `int`                | `int`    | `bool`                               |
-| 8     | `BOOLEAN`        | `unsigned char`      | `byte`   | `[MarshalAs(UnmanagedType.U1)] bool` |
-| 8     | `BYTE`           | `unsigned char`      | `byte`   |                                      |
-| 8     | `CHAR`           | `char`               | `sbyte`  |                                      |
-| 8     | `UCHAR`          | `unsigned char`      | `byte`   |                                      |
-| 16    | `SHORT`          | `short`              | `short`  |                                      |
-| 16    | `CSHORT`         | `short`              | `short`  |                                      |
-| 16    | `USHORT`         | `unsigned short`     | `ushort` |                                      |
-| 16    | `WORD`           | `unsigned short`     | `ushort` |                                      |
-| 16    | `ATOM`           | `unsigned short`     | `ushort` |                                      |
-| 32    | `INT`            | `int`                | `int`    |                                      |
-| 32    | `LONG`           | `long`               | `int`    |  See `CLong` and `CULong` below.     |
-| 32    | `ULONG`          | `unsigned long`      | `uint`   |  See `CLong` and `CULong` below.     |
-| 32    | `DWORD`          | `unsigned long`      | `uint`   |  See `CLong` and `CULong` below.     |
-| 64    | `QWORD`          | `long long`          | `long`   |                                      |
-| 64    | `LARGE_INTEGER`  | `long long`          | `long`   |                                      |
-| 64    | `LONGLONG`       | `long long`          | `long`   |                                      |
-| 64    | `ULONGLONG`      | `unsigned long long` | `ulong`  |                                      |
-| 64    | `ULARGE_INTEGER` | `unsigned long long` | `ulong`  |                                      |
-| 32    | `HRESULT`        | `int`                | `int`    |                                      |
-| 32    | `NTSTATUS`       | `int`                | `int`    |                                      |
+| Width | Windows          | C#       | Alternative                          |
+|:------|:-----------------|:---------|:-------------------------------------|
+| 32    | `BOOL`           | `int`    | `bool`                               |
+| 8     | `BOOLEAN`        | `byte`   | `[MarshalAs(UnmanagedType.U1)] bool` |
+| 8     | `BYTE`           | `byte`   |                                      |
+| 8     | `CHAR`           | `sbyte`  |                                      |
+| 8     | `UCHAR`          | `byte`   |                                      |
+| 16    | `SHORT`          | `short`  |                                      |
+| 16    | `CSHORT`         | `short`  |                                      |
+| 16    | `USHORT`         | `ushort` |                                      |
+| 16    | `WORD`           | `ushort` |                                      |
+| 16    | `ATOM`           | `ushort` |                                      |
+| 32    | `INT`            | `int`    |                                      |
+| 32    | `LONG`           | `int`    |  See `CLong` and `CULong` below.     |
+| 32    | `ULONG`          | `uint`   |  See `CLong` and `CULong` below.     |
+| 32    | `DWORD`          | `uint`   |                                      |
+| 64    | `QWORD`          | `long`   |                                      |
+| 64    | `LARGE_INTEGER`  | `long`   |                                      |
+| 64    | `LONGLONG`       | `long`   |                                      |
+| 64    | `ULONGLONG`      | `ulong`  |                                      |
+| 64    | `ULARGE_INTEGER` | `ulong`  |                                      |
+| 32    | `HRESULT`        | `int`    |                                      |
+| 32    | `NTSTATUS`       | `int`    |                                      |
 
 The following types, being pointers, do follow the width of the platform. Use `IntPtr`/`UIntPtr` for these.
 
--- a/docs/standard/native-interop/best-practices.md
+++ b/docs/standard/native-interop/best-practices.md
@@ -280,7 +280,7 @@ extern static CLong Function(CLong a);
 nint result = Function(new CLong(10)).Value;
 ```
 
-When targeting .NET 5 and prior .NET versions, you should declare separate Windows and non-Windows signatures to handle the problem.
+When targeting .NET 5 and earlier .NET versions, you should declare separate Windows and non-Windows signatures to handle the problem.
 
 ```csharp
 static readonly bool IsWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
--- a/docs/standard/native-interop/best-practices.md
+++ b/docs/standard/native-interop/best-practices.md
@@ -295,14 +295,14 @@ extern static int FunctionWindows(int a);
 extern static nint FunctionUnix(nint a);
 
 // Usage
-int result;
+nint result;
 if (IsWindows)
 {
-    result = (int)FunctionWindows(10);
+    result = FunctionWindows(10);
 }
 else
 {
-    result = (int)FunctionUnix(10);
+    result = FunctionUnix(10);
 }
 ```
 
--- a/docs/standard/native-interop/best-practices.md
+++ b/docs/standard/native-interop/best-practices.md
@@ -259,7 +259,7 @@ There are types in the C/C++ language that have latitude in how they are defined
 
 C/C++ `long` and C# `long` are not the same types. Using C# `long` to interop with C/C++ `long` is almost never correct.
 
-The `long` keyword in C/C++ is defined to have ["at least 32"](https://en.cppreference.com/w/c/language/arithmetic_types) bits. This means there is a minimum required but platforms are not limited to that. The table below illustrates the differences in provided bits for the C/C++ `long` data type between platforms.
+The `long` type in C/C++ is defined to have ["at least 32"](https://en.cppreference.com/w/c/language/arithmetic_types) bits. This means there is a minimum number of required bits, but platforms can choose to use more bits if desired. The following table illustrates the differences in provided bits for the C/C++ `long` data type between platforms.
 
 |             | 32-bit | 64-bit |
 |:------------|:-------|:-------|
--- a/docs/standard/native-interop/best-practices.md
+++ b/docs/standard/native-interop/best-practices.md
@@ -251,6 +251,59 @@ finally
 }
 ```
 
+## Windows and non-Windows data type considerations
+
+There are types in the C language that have latitude in how they are defined. When writing cross-platform interop, cases can arise where platforms differ and can cause issues if not considered.
+
+### C `long`
+
+The `long` in C is defined to have ["at least 32"](https://en.cppreference.com/w/c/language/arithmetic_types) bits. This means there is a minimum required but platforms are not limited to that. The table below illustrates the differences in provided bits for the C `long` data type between platforms.
+
+|             | 32-bit | 64-bit |
+|:------------|:-------|:-------|
+| Windows     | 32     | 32     |
+| macOS/\*nix | 32     | 64     |
+
+These differences make authoring cross-platform P/Invokes difficult when the native function is defined to use `long` on all platforms. It is recommended for .NET 5 and prior .NET version to declare a Windows and a non-Windows signature to handle the problem.
+
+```csharp
+    // Cross platform C function
+    // void Function(long a);
+#if WINDOWS
+    [DllImport("NativeLib", EntryPoint = "Function")]
+    extern static void Function32(int a);
+    [DllImport("NativeLib", EntryPoint = "Function")]
+    extern static void Function64(int a);
+#else
+    [DllImport("NativeLib", EntryPoint = "Function")]
+    extern static void Function32(int a);
+    [DllImport("NativeLib", EntryPoint = "Function")]
+    extern static void Function64(long a);
+#endif
+    
+// Usage
+if (IntPtr.Size == 4)
+{
+    Function32(a);
+}
+else
+{
+    Function64(a);
+}
+```
+
+In .NET 6 a new type has been introduced to help ameliorate this issue - [`CLong` and `CULong`](https://github.com/dotnet/runtime/issues/13788). In .NET 6 the following is now possible.
+
+```csharp
+// Cross platform C function
+// void Function(long a);
+[DllImport("NativeLib")]
+extern static void Function(CLong a);
+    
+// Usage
+Function(a);
+```
+
 ## Structs
 
 Managed structs are created on the stack and aren't removed until the method returns. By definition then, they are "pinned" (it won't get moved by the GC). You can also simply take the address in unsafe code blocks if native code won't use the pointer past the end of the current method.
--- a/docs/standard/native-interop/best-practices.md
+++ b/docs/standard/native-interop/best-practices.md
@@ -266,7 +266,7 @@ The `long` keyword in C is defined to have ["at least 32"](https://en.cppreferen
 | Windows     | 32     | 32     |
 | macOS/\*nix | 32     | 64     |
 
-These differences make authoring cross-platform P/Invokes difficult when the native function is defined to use `long` on all platforms. 
+These differences make authoring cross-platform P/Invokes difficult when the native function is defined to use `long` on all platforms.
 
 In .NET 6 a new type has been introduced to help ameliorate this issue - [`CLong` and `CULong`](https://github.com/dotnet/runtime/issues/13788). In .NET 6 the following is now possible. The example is for `CLong`, but you can use `CULong` to abstract C's `unsigned long` in a similar way.
 
--- a/docs/standard/native-interop/best-practices.md
+++ b/docs/standard/native-interop/best-practices.md
@@ -277,7 +277,7 @@ In .NET 6 a new type has been introduced to help ameliorate this issue - [`CLong
 extern static CLong Function(CLong a);
     
 // Usage
-int result = Function(new CLong(10)).Value;
+nint result = Function(new CLong(10)).Value;
 ```
 
 When targeting .NET 5 and prior .NET versions, you should declare separate Windows and non-Windows signatures to handle the problem.
--- a/docs/standard/native-interop/best-practices.md
+++ b/docs/standard/native-interop/best-practices.md
@@ -256,7 +256,7 @@ finally
 There are types in the C language that have latitude in how they are defined. When writing cross-platform interop, cases can arise where platforms differ and can cause issues if not considered.
 
 ### C `long`
-
+C `long` and C# `long` are not the same types. Using C# `long` to interop with C `long` is almost never correct.
 The `long` keyword in C is defined to have ["at least 32"](https://en.cppreference.com/w/c/language/arithmetic_types) bits. This means there is a minimum required but platforms are not limited to that. The table below illustrates the differences in provided bits for the C `long` data type between platforms.
 
 |             | 32-bit | 64-bit |
--- a/docs/standard/native-interop/best-practices.md
+++ b/docs/standard/native-interop/best-practices.md
@@ -182,9 +182,9 @@ The following types are the same size on 32-bit and 64-bit Windows, despite thei
 | 16    | `WORD`           | `unsigned short`     | `ushort` |                                      |
 | 16    | `ATOM`           | `unsigned short`     | `ushort` |                                      |
 | 32    | `INT`            | `int`                | `int`    |                                      |
-| 32    | `LONG`           | `long`               | `int`    |                                      |
-| 32    | `ULONG`          | `unsigned long`      | `uint`   |                                      |
-| 32    | `DWORD`          | `unsigned long`      | `uint`   |                                      |
+| 32    | `LONG`           | `long`               | `int`    |  See `CLong` and `CULong` below.     |
+| 32    | `ULONG`          | `unsigned long`      | `uint`   |  See `CLong` and `CULong` below.     |
+| 32    | `DWORD`          | `unsigned long`      | `uint`   |  See `CLong` and `CULong` below.     |
 | 64    | `QWORD`          | `long long`          | `long`   |                                      |
 | 64    | `LARGE_INTEGER`  | `long long`          | `long`   |                                      |
 | 64    | `LONGLONG`       | `long long`          | `long`   |                                      |
@@ -251,12 +251,14 @@ finally
 }
 ```
 
-## Windows and non-Windows data type considerations
+## Cross-platform data type considerations
 
 There are types in the C language that have latitude in how they are defined. When writing cross-platform interop, cases can arise where platforms differ and can cause issues if not considered.
 
 ### C `long`
+
 C `long` and C# `long` are not the same types. Using C# `long` to interop with C `long` is almost never correct.
+
 The `long` keyword in C is defined to have ["at least 32"](https://en.cppreference.com/w/c/language/arithmetic_types) bits. This means there is a minimum required but platforms are not limited to that. The table below illustrates the differences in provided bits for the C `long` data type between platforms.
 
 |             | 32-bit | 64-bit |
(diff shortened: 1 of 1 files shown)


docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,59 @@ finally
...
+### C `long`
+
+The `long` in C is defined to have ["at least 32"](https://en.cppreference.com/w/c/language/arithmetic_types) bits. This means there is a minimum required but platforms are not limited to that. The table below illustrates the differences in provided bits for the C `long` data type between platforms.
+
+|             | 32-bit | 64-bit |
+|:------------|:-------|:-------|
+| Windows     | 32     | 32     |
+| macOS/\*nix | 32     | 64     |
+
+These differences make authoring cross-platform P/Invokes difficult when the native function is defined to use `long` on all platforms. It is recommended for .NET 5 and prior .NET version to declare a Windows and a non-Windows signature to handle the problem.
```
danmoseley, line 267:
```suggestion
These differences make authoring cross-platform P/Invokes difficult when the native function is defined to use `long` on all platforms. When targeting .NET 5 and prior .NET versions, you should declare separate Windows and non-Windows signatures to handle the problem.
```

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,59 @@ finally
...
+
+```csharp
+    // Cross platform C function
+    // void Function(long a);
+#if WINDOWS
```
danmoseley, line 272: Maybe extract the identical 32 bit one out of the #if?

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,59 @@ finally
...
+    [DllImport("NativeLib", EntryPoint = "Function")]
+    extern static void Function64(int a);
+#else
+    [DllImport("NativeLib", EntryPoint = "Function")]
+    extern static void Function32(int a);
+    [DllImport("NativeLib", EntryPoint = "Function")]
+    extern static void Function64(long a);
+#endif
+    
+// Usage
```
danmoseley, line 284: This assumes you are compiling your IL for anycpu rather than x64/x86. I assume that's the most common practice (?)

AaronRobinsonMSFT: > I assume that's the most common practice (?)

Yes. I don't see the non-IL build's as common anymore outside of the runtime. I don't know who would define that best-practice but it is what we recommend.

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,59 @@ finally
...
+{
+    Function32(a);
+}
+else
+{
+    Function64(a);
+}
+```
+
+In .NET 6 a new type has been introduced to help ameliorate this issue - [`CLong` and `CULong`](https://github.com/dotnet/runtime/issues/13788). In .NET 6 the following is now possible.
```
danmoseley, line 294: Ideally we'd have a doc to link to. @carlossanlop there are XML comments in the files [eg](https://github.com/dotnet/runtime/blob/1ee59da9f6104c611b137c9d14add04becefdf14/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/CLong.cs#L17-L25) should I expect that there's a page on docs.microsoft.com that will appear for these?

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,59 @@ finally
...
+```csharp
+// Cross platform C function
+// void Function(long a);
+[DllImport("NativeLib")]
+extern static void Function(CLong a);
+    
+// Usage
+Function(a);
+```
+
```
danmoseley, line 305: "You can use `CULong` to abstract `ulong` in a similar way." ?

danmoseley: Should this also mention `NFloat` ?

danmoseley (approved):
thanks!

docs/standard/native-interop/best-practices.md
AaronRobinsonMSFT: Yes. Example or a comment?

AaronRobinsonMSFT: Oh. Misread the suggestion :)

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,58 @@ finally
...
+[DllImport("NativeLib", EntryPoint = "Function")]
+extern static void Function32(int a);
+
+#if WINDOWS
+    [DllImport("NativeLib", EntryPoint = "Function")]
+    extern static void Function64(int a);
+#else
+    [DllImport("NativeLib", EntryPoint = "Function")]
+    extern static void Function64(long a);
+#endif
```
jkoritzinsky, line 281: An alternative sample could be to use `int` on Windows and `IntPtr` on non-Windows and not split the code below on IntPtr.Size.

danmoseley: That's shorter, but perhaps less clearer. Does this suggest we're missing an `nlong`/`nulong` ?

danmoseley:
cc @tannergooding

docs/standard/native-interop/best-practices.md
jkoritzinsky: That's what `CLong` and `CULong` are. This example is for before those types existed.

AaronRobinsonMSFT: >That's shorter, but IMO less clear.

I agree. This is a generalized pattern and people free to optimize it with that suggestion. I want to explicitly demonstrate the sizes. I think `IntPtr` is hiding too much here and is implying the `size_t` semantics here which this follows but that is incidental.

jkoritzinsky: I understand the desire to be explicit. Let's go with the example you made.

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,58 @@ finally
 }
 ```
 
+## Windows and non-Windows data type considerations
+
+There are types in the C language that have latitude in how they are defined. When writing cross-platform interop, cases can arise where platforms differ and can cause issues if not considered.
+
+### C `long`
+
+The `long` in C is defined to have ["at least 32"](https://en.cppreference.com/w/c/language/arithmetic_types) bits. This means there is a minimum required but platforms are not limited to that. The table below illustrates the differences in provided bits for the C `long` data type between platforms.
```
tannergooding, line 260:
```suggestion
The `long` keyword in C is defined to have ["at least 32"](https://en.cppreference.com/w/c/language/arithmetic_types) bits. This means there is a minimum required but platforms are not limited to that. The table below illustrates the differences in provided bits for the C `long` data type between platforms.
```

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,58 @@ finally
...
+// void Function(long a);
+[DllImport("NativeLib", EntryPoint = "Function")]
+extern static void Function32(int a);
+
+#if WINDOWS
+    [DllImport("NativeLib", EntryPoint = "Function")]
+    extern static void Function64(int a);
+#else
+    [DllImport("NativeLib", EntryPoint = "Function")]
+    extern static void Function64(long a);
```
tannergooding, line 280: nit: C# `long` is only correct on 64-bit Unix. The correct type here is `IntPtr` or `nint` (and `UIntPtr`/`nuint` for `unsigned long`)

tannergooding: I see you actually discussed this in one of the collapsed threads.

I don't agree that it is incidental, the Unix ABI explicitly defines `unsigned long` to be the same size as `size_t` and therefore the same size as a pointer given the other constraints set in place.

AaronRobinsonMSFT: >nit: C# long is only correct on 64-bit Unix. 

It is also on macOS if I recall? I assume you are using Unix for all non-Windows platforms. That is the reason for the `64` suffix.

>The correct type here is IntPtr or nint (and UIntPtr/nuint for unsigned long)

I don't see it as the correct type. That is the pointer size and `long` isn't defined as being related to pointer size it just happens to align with that. It is a fine optimization for sure but as an example I prefer stating these sizes explicitly.

AaronRobinsonMSFT: > the Unix ABI explicitly defines unsigned long to be the same size as size_t and therefore the same size as a pointer given the other constraints set in place.

A link to this would be interesting. I was not aware that was an explicit decision.

AaronRobinsonMSFT: @tannergooding From what I am seeing this is the opposite - from https://uclibc.org/docs/psABI-x86_64.pdf.

> The type size_t is defined as unsigned long

AaronRobinsonMSFT: I guess it doesn't matter in principle - they are the same which we all agree. I don't see this as something that is required though because this is defined by a platform so it can be what they wish. I do accept it will work but as mentioned I prefer to keep this explicit and users are free to optimize as desired.

tannergooding: > A link to this would be interesting. I was not aware that was an explicit decision.

https://gitlab.com/x86-psABIs/i386-ABI and https://gitlab.com/x86-psABIs/x86-64-ABI

Specifically: https://gitlab.com/x86-psABIs/x86-64-ABI/-/blob/master/x86-64-ABI/low-level-sys-info.tex#L261 and https://gitlab.com/x86-psABIs/i386-ABI/-/blob/hjl/master/low-level-sys-info.tex#L151

I would think the preferred pattern here is actually:
```
private static readonly bool IsWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);

[DllImport("NativeLib", EntryPoint = "Function")]
extern static void FunctionWindows(int a);

[DllImport("NativeLib", EntryPoint = "Function")]
extern static void FunctionUnix(nint a);

if (IsWindows)
{
    FunctionWindows(a);
}
else
{
    FunctionUnix(a);
}
```

This results in the irrelevant path being optimized away as dead code and matches what we effectively do for `CLong`/`CULong` internally.
You might theoretically want an assert or check for `IsUnix`, but that isn't necessary given .NET Framework or .NET Core.

AaronRobinsonMSFT: Alright. I've been convinced. I will update the example. Thanks @tannergooding 

/cc @danmoseley @jkoritzinsky

danmoseley: I think this discussion was worthwhile because we're trying to establish a pattern for everyone. thanks.

tannergooding (approved):

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,56 @@ finally
...
+{
+    FunctionWindows(a);
+}
+else
+{
+    FunctionUnix(a);
+}
+```
+
+In .NET 6 a new type has been introduced to help ameliorate this issue - [`CLong` and `CULong`](https://github.com/dotnet/runtime/issues/13788). In .NET 6 the following is now possible. The example is for `CLong`, but you can use `CULong` to abstract C's `unsigned long` in a similar way.
```
jkotas, line 292: Can this link to CLong and CULong type documentation instead of the dotnet/runtime issue?

danmoseley: Doesn't seem to exist - I asked @carlossanlop above when it appears.

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,56 @@ finally
 }
 ```
```
jkotas, line 252: Also mention CLong as alternative option for `long` in the "Common Windows data types" table, and/or link from that table to this paragraph?

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,56 @@ finally
...
 
+## Windows and non-Windows data type considerations
+
+There are types in the C language that have latitude in how they are defined. When writing cross-platform interop, cases can arise where platforms differ and can cause issues if not considered.
+
+### C `long`
+
```
jkotas, line 259:
```suggestion
C `long` and C# `long` are not the same types. Using C# `long` to interop with C `long` is almost never correct.
```
This is very common newbie mistake. I think it is worth highlighting.

docs/standard/native-interop/best-practices.md
jkotas, line 292: I would put the .NET 6+ solution first. My general rule for this sort of documentation is to optimize for future and put the simple modern way of doing stuff first, so that the people that are not the latest version of the platform do not need to read the whole history.

docs/standard/native-interop/best-practices.md
danmoseley: BTW in practice does this apply to C++ as well (since many times folks are writing C in code they consider C++)

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,56 @@ finally
...
+In .NET 6 a new type has been introduced to help ameliorate this issue - [`CLong` and `CULong`](https://github.com/dotnet/runtime/issues/13788). In .NET 6 the following is now possible. The example is for `CLong`, but you can use `CULong` to abstract C's `unsigned long` in a similar way.
+
+```csharp
+// Cross platform C function
+// void Function(long a);
+[DllImport("NativeLib")]
+extern static void Function(CLong a);
+    
+// Usage
+Function(a);
```
jkotas, line 301:
```suggestion
// long Function(long a);
[DllImport("NativeLib")]
extern static long Function(CLong a);
    
int result = Function(new CLong(42)).Value;
```
Assuming type of `a` is same as in the other example, `Function(a)` won't actually compile. Also, it may be nice to demonstrate both to and from conversions.

docs/standard/native-interop/best-practices.md
AaronRobinsonMSFT: >BTW in practice does this apply to C++ as well (since many times folks are writing C in code they consider C++)

Yep. Exact same behavior.

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,56 @@ finally
 }
 ```
 
+## Windows and non-Windows data type considerations
```
jkotas, line 254:
```suggestion
## Cross-platform data type considerations
```

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,56 @@ finally
...
+// Cross platform C function
+// void Function(long a);
+[DllImport("NativeLib")]
+extern static void Function(CLong a);
+    
+// Usage
+Function(a);
+```
+
 ## Structs
```
jkotas, line 309: The doc jumps back and forth between general guidelines and very specific Windows stuff. I think overall flow of this document can be improved. It can be done in a different PR. 

General guidelines - strings, structs, fixed buffers, etc.
Cross-platform data type considerations
Platform-specific - Common Windows data types, Formerly built-in supported types

It may be even fine to split this into multiple docs. I see that we have [Writing cross platform P/Invoke code](https://docs.microsoft.com/en-us/dotnet/standard/native-interop/cross-platform). Should the "Cross-platform data type considerations" topic be there?

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,61 @@ finally
...
+C `long` and C# `long` are not the same types. Using C# `long` to interop with C `long` is almost never correct.
+
+The `long` keyword in C is defined to have ["at least 32"](https://en.cppreference.com/w/c/language/arithmetic_types) bits. This means there is a minimum required but platforms are not limited to that. The table below illustrates the differences in provided bits for the C `long` data type between platforms.
+
+|             | 32-bit | 64-bit |
+|:------------|:-------|:-------|
+| Windows     | 32     | 32     |
+| macOS/\*nix | 32     | 64     |
+
+These differences make authoring cross-platform P/Invokes difficult when the native function is defined to use `long` on all platforms.
```
jkotas, line 269: Nit: It is not difficult in .NET 6+ anymore

docs/standard/native-interop/best-practices.md
AaronRobinsonMSFT: Yes. This is part of the exercise being done with the interop-doc - what is a logical flow for learning about interop and best to walk users through writing interop solutions.

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,61 @@ finally
...
+In .NET 6 a new type has been introduced to help ameliorate this issue - [`CLong` and `CULong`](https://github.com/dotnet/runtime/issues/13788). In .NET 6 the following is now possible. The example is for `CLong`, but you can use `CULong` to abstract C's `unsigned long` in a similar way.
+
+```csharp
+// Cross platform C function
+// long Function(long a);
+[DllImport("NativeLib")]
+extern static CLong Function(CLong a);
+    
+// Usage
+int result = Function(new CLong(10)).Value;
```
jkotas, line 280:
```suggestion
nint result = Function(new CLong(10)).Value;
```
Correcting my own mistake

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,61 @@ finally
...
+// Usage
+int result;
+if (IsWindows)
+{
+    result = (int)FunctionWindows(10);
+}
+else
+{
+    result = (int)FunctionUnix(10);
+}
```
jkotas, line 306:
```suggestion
nint result;
if (IsWindows)
{
    result = FunctionWindows(10);
}
else
{
    result = FunctionUnix(10);
}
```

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,61 @@ finally
...
+The `long` keyword in C is defined to have ["at least 32"](https://en.cppreference.com/w/c/language/arithmetic_types) bits. This means there is a minimum required but platforms are not limited to that. The table below illustrates the differences in provided bits for the C `long` data type between platforms.
+
+|             | 32-bit | 64-bit |
+|:------------|:-------|:-------|
+| Windows     | 32     | 32     |
+| macOS/\*nix | 32     | 64     |
+
+These differences can make authoring cross-platform P/Invokes difficult when the native function is defined to use `long` on all platforms.
+
+In .NET 6 a new type has been introduced to help ameliorate this issue - [`CLong` and `CULong`](https://github.com/dotnet/runtime/issues/13788). In .NET 6 the following is now possible. The example is for `CLong`, but you can use `CULong` to abstract C's `unsigned long` in a similar way.
```
jkotas, line 271:
```suggestion
In .NET 6 and later versions, use [`CLong` and `CULong`](https://github.com/dotnet/runtime/issues/13788) types for interop with C `long` and `ulong` data types. The example is for `CLong`, but you can use `CULong` to abstract C's `unsigned long` in a similar way.
```
I think that this is the style usually used for documentation. "In .NET 6 the following is now possible" is what we would say in a blog post.

docs/standard/native-interop/best-practices.md
```diff
@@ -182,9 +182,9 @@ The following types are the same size on 32-bit and 64-bit Windows, despite thei
 | 16    | `WORD`           | `unsigned short`     | `ushort` |                                      |
 | 16    | `ATOM`           | `unsigned short`     | `ushort` |                                      |
 | 32    | `INT`            | `int`                | `int`    |                                      |
-| 32    | `LONG`           | `long`               | `int`    |                                      |
-| 32    | `ULONG`          | `unsigned long`      | `uint`   |                                      |
-| 32    | `DWORD`          | `unsigned long`      | `uint`   |                                      |
+| 32    | `LONG`           | `long`               | `int`    |  See `CLong` and `CULong` below.     |
+| 32    | `ULONG`          | `unsigned long`      | `uint`   |  See `CLong` and `CULong` below.     |
+| 32    | `DWORD`          | `unsigned long`      | `uint`   |  See `CLong` and `CULong` below.     |
```
tannergooding, line 187: What about `HRESULT` and `NTSTATUS` below? They are also Windows `long` and therefore on Windows only should be `int`/`uint`.

However, if that code is ported one-to-one to Windows (like was essentially done for https://devblogs.microsoft.com/directx/directx-heart-linux/, https://github.com/microsoft/DirectX-Headers/blob/main/include/wsl), they start requiring `CLong` and `CULong`.

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,61 @@ finally
...
+// Cross platform C function
+// long Function(long a);
+[DllImport("NativeLib")]
+extern static CLong Function(CLong a);
+    
+// Usage
+int result = Function(new CLong(10)).Value;
+```
+
+When targeting .NET 5 and prior .NET versions, you should declare separate Windows and non-Windows signatures to handle the problem.
```
jkotas, line 283:
```suggestion
When targeting .NET 5 and earlier .NET versions, you should declare separate Windows and non-Windows signatures to handle the problem.
```
I believe that earlier/later is what the documentation usually uses when refering to versions.

docs/standard/native-interop/best-practices.md
AaronRobinsonMSFT: They actually did the right thing there. This should really be `unsigned int`. A 64-bit `HRESULT` is going to be a nightmare to convert.

https://github.com/microsoft/DirectX-Headers/blob/5b5b57aee33e8e38eff0242221893a43ac6ea90d/include/wsl/winadapter.h#L24

https://github.com/microsoft/DirectX-Headers/blob/5b5b57aee33e8e38eff0242221893a43ac6ea90d/include/wsl/winadapter.h#L185

docs/standard/native-interop/best-practices.md
```diff
@@ -251,56 +251,61 @@ finally
...
+int result;
 if (IsWindows)
 {
-    FunctionWindows(a);
+    result = (int)FunctionWindows(10);
 }
 else
 {
-    FunctionUnix(a);
+    result = (int)FunctionUnix(10);
```
tannergooding: nit: This is potentially problematic as it may incorrectly truncate the return value on `Unix`. There are many Unix APIs that do restrict the value to 32-bits even when returning `long` on a 64-bit platform, but not all of them do.

docs/standard/native-interop/best-practices.md
jkotas: That depends on how you choose to build your Windows emulator. NTSTATUS is 32-bit in CoreCLR PAL that number of projects copied Windows emulator layer from.

docs/standard/native-interop/best-practices.md
```diff
@@ -182,16 +182,16 @@ The following types are the same size on 32-bit and 64-bit Windows, despite thei
...
+| 32    | `ULONG`          | `unsigned long`      | `uint`   |  See `CLong` and `CULong` below.     |
+| 32    | `DWORD`          | `unsigned long`      | `uint`   |  See `CLong` and `CULong` below.     |
 | 64    | `QWORD`          | `long long`          | `long`   |                                      |
 | 64    | `LARGE_INTEGER`  | `long long`          | `long`   |                                      |
 | 64    | `LONGLONG`       | `long long`          | `long`   |                                      |
 | 64    | `ULONGLONG`      | `unsigned long long` | `ulong`  |                                      |
 | 64    | `ULARGE_INTEGER` | `unsigned long long` | `ulong`  |                                      |
-| 32    | `HRESULT`        | `long`               | `int`    |                                      |
-| 32    | `NTSTATUS`       | `long`               | `int`    |                                      |
+| 32    | `HRESULT`        | `int`                | `int`    |                                      |
```
jkotas, line 193: This is technically incorrect since this column is called `C (Windows)` and HRESULT is really `long` on Windows. It may be best to delete the `C (Windows)` column. I do not think that it is helping.

docs/standard/native-interop/best-practices.md
```diff
@@ -182,16 +182,16 @@ The following types are the same size on 32-bit and 64-bit Windows, despite thei
 | 16    | `WORD`           | `unsigned short`     | `ushort` |                                      |
 | 16    | `ATOM`           | `unsigned short`     | `ushort` |                                      |
 | 32    | `INT`            | `int`                | `int`    |                                      |
-| 32    | `LONG`           | `long`               | `int`    |                                      |
-| 32    | `ULONG`          | `unsigned long`      | `uint`   |                                      |
-| 32    | `DWORD`          | `unsigned long`      | `uint`   |                                      |
+| 32    | `LONG`           | `long`               | `int`    |  See `CLong` and `CULong` below.     |
+| 32    | `ULONG`          | `unsigned long`      | `uint`   |  See `CLong` and `CULong` below.     |
+| 32    | `DWORD`          | `unsigned long`      | `uint`   |  See `CLong` and `CULong` below.     |
```
jkotas, line 187:
```suggestion
| 32    | `DWORD`          | `unsigned long`      | `uint`   |                                      |
```
I think this should be on the same plan as `HRESULT`.

docs/standard/native-interop/best-practices.md
tannergooding: I don't think the change of `HRESULT`/`NTSTATUS` in the table to `int` is correct. They are fundamentally the `long` type in C and on Unix systems, without a typedef fixup, will be 32-64 bits.

Yes, the Windows intent is that it is 32-bits, same as the Windows intent is that DWORD is 32-bits. But, in practice it only matters what is used to compile on Unix which may still be `long` and which may not be retyped.

It is 100% a nightmare and it likewise causes issues between `Unix` and `Windows` because on `Windows` you can't represent a full point but the Unix intent might have been that it is a pointer.

I did miss that WSL had `typedef'd` `LONG` as `int32_t`; ut they've not handled other cases. `HWND` for example is a proper `void*` on Windows, but an `int` in the WSL header. `wchar_t` is a `uint16_t` on Windows and a `uint32_t` on Unix. `LPARAM; and `LONG_PTR` is a pointer on Windows, while they've typedef'd to `int64_t` 😄

jkotas: Yes, building Windows emulation layers is fun and everybody does it slightly differently dependending on what makes the most sense for their specific component.

jkotas (approved):

docs/standard/native-interop/best-practices.md
carlossanlop: This public type was [introduced on Jan 12 2021](https://github.com/dotnet/runtime/pull/46401). We have not yet created any ".NET 6.0 Preview" entry in the [Version dropdown of MS Docs](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices?view=net-5.0).

@AaronRobinsonMSFT I suggest you create an issue and add it to your backlog, so you can update this document and have the link point to the .NET 6.0 MS Docs API page of the `CLong` type, after we create the first Preview dropdown item.

docs/standard/native-interop/best-practices.md
carlossanlop: [Answered](https://github.com/dotnet/docs/pull/23466#discussion_r600978192).

docs/standard/native-interop/best-practices.md
danmoseley: Right I just meant, will folks writing C++ read this and skip over? Perhaps it should say C/C++. Whatver you think.

AaronRobinsonMSFT: Great suggestion. Will update.

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,61 @@ finally
...
 
+## Cross-platform data type considerations
+
+There are types in the C/C++ language that have latitude in how they are defined. When writing cross-platform interop, cases can arise where platforms differ and can cause issues if not considered.
+
+### C/C++ `long`
+
+C/C++ `long` and C# `long` are not the same types. Using C# `long` to interop with C/C++ `long` is almost never correct.
+
+The `long` keyword in C/C++ is defined to have ["at least 32"](https://en.cppreference.com/w/c/language/arithmetic_types) bits. This means there is a minimum required but platforms are not limited to that. The table below illustrates the differences in provided bits for the C/C++ `long` data type between platforms.
```
gewarren, line 262:
```suggestion
The `long` type in C/C++ is defined to have ["at least 32"](https://en.cppreference.com/w/c/language/arithmetic_types) bits. This means there is a minimum number of required bits, but platforms can choose to use more bits if desired. The following table illustrates the differences in provided bits for the C/C++ `long` data type between platforms.
```

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,61 @@ finally
...
+
+|             | 32-bit | 64-bit |
+|:------------|:-------|:-------|
+| Windows     | 32     | 32     |
+| macOS/\*nix | 32     | 64     |
+
+These differences can make authoring cross-platform P/Invokes difficult when the native function is defined to use `long` on all platforms.
+
+In .NET 6 and later versions, use [`CLong` and `CULong`](https://github.com/dotnet/runtime/issues/13788) types for interop with C/C++ `long` and `unsigned long` data types. The example is for `CLong`, but you can use `CULong` to abstract `unsigned long` in a similar way.
```
gewarren, line 271:
```suggestion
In .NET 6 and later versions, use the [`CLong` and `CULong`](https://github.com/dotnet/runtime/issues/13788) types for interop with C/C++ `long` and `unsigned long` data types. The following example is for `CLong`, but you can use `CULong` to abstract `unsigned long` in a similar way.
```

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,61 @@ finally
...
+// Cross platform C function
+// long Function(long a);
+[DllImport("NativeLib")]
+extern static CLong Function(CLong a);
+    
+// Usage
+nint result = Function(new CLong(10)).Value;
+```
+
+When targeting .NET 5 and earlier .NET versions, you should declare separate Windows and non-Windows signatures to handle the problem.
```
gewarren, line 283:
```suggestion
When targeting .NET 5 and earlier versions, you should declare separate Windows and non-Windows signatures to handle the problem.
```

docs/standard/native-interop/best-practices.md
```diff
@@ -251,6 +251,61 @@ finally
...
+
+There are types in the C/C++ language that have latitude in how they are defined. When writing cross-platform interop, cases can arise where platforms differ and can cause issues if not considered.
+
+### C/C++ `long`
+
+C/C++ `long` and C# `long` are not the same types. Using C# `long` to interop with C/C++ `long` is almost never correct.
+
+The `long` keyword in C/C++ is defined to have ["at least 32"](https://en.cppreference.com/w/c/language/arithmetic_types) bits. This means there is a minimum required but platforms are not limited to that. The table below illustrates the differences in provided bits for the C/C++ `long` data type between platforms.
+
+|             | 32-bit | 64-bit |
```
gewarren, line 264:
```suggestion
| Platform | 32-bit | 64-bit |
```

docs/standard/native-interop/best-practices.md
```diff
@@ -169,29 +169,29 @@ Here is a list of data types commonly used in Windows APIs and which C# types to
...
+| 8     | `BYTE`           | `byte`   |                                      |
+| 8     | `CHAR`           | `sbyte`  |                                      |
+| 8     | `UCHAR`          | `byte`   |                                      |
+| 16    | `SHORT`          | `short`  |                                      |
+| 16    | `CSHORT`         | `short`  |                                      |
+| 16    | `USHORT`         | `ushort` |                                      |
+| 16    | `WORD`           | `ushort` |                                      |
+| 16    | `ATOM`           | `ushort` |                                      |
+| 32    | `INT`            | `int`    |                                      |
+| 32    | `LONG`           | `int`    |  See `CLong` and `CULong` below.     |
```
gewarren, line 185:
```suggestion
| 32    | `LONG`           | `int`    |  See [`CLong` and `CULong`](#cc-long).     |
```

docs/standard/native-interop/best-practices.md
```diff
@@ -169,29 +169,29 @@ Here is a list of data types commonly used in Windows APIs and which C# types to
...
+| 32    | `ULONG`          | `uint`   |  See `CLong` and `CULong` below.     |
```
gewarren, line 186:
```suggestion
| 32    | `ULONG`          | `uint`   |  See [`CLong` and `CULong`](#cc-long).      |
```

gewarren (approved):
LGTM!

PR #23477, opened 2021-03-24
Update subscribe-events.md

kovandzhiev:
There is not need to check `raiseProductPriceChangedEvent` twice.

## Summary

When we have first if we no longer need to `raiseProductPriceChangedEvent` to check for second time.


adegeo (approved):
Thanks for spotting that code error!

PR #23480, opened 2021-03-24
Fix compiler option redirects

gewarren:
Fixes redirects introduced by https://github.com/dotnet/docs/pull/22797.


BillWagner (approved):
Thanks @gewarren 

I'll :shipit: now.

PR #23486, opened 2021-03-25
Include dynamic type in list of exceptions

dlech:
Strictly speaking, `dynamic` is not an alias for `System.Object` in the same way that `nint` is not an alias for `System.IntPtr`.

## Summary

Describe your changes here.


BillWagner (approved):
Thanks for making this change @dlech We appreciate it.

PR #23500, opened 2021-03-26
Update how-to-write-csharp-analyzer-code-fix.md

KevinSWork:
Build Unit Tests doesn't remove the test methods created by the template, which will no longer pass based on the changes to the analyzer and code fix.  Then, the tutorial gives a single test method to add but asks the user to run "these two tests" -- fixed to refer to only the single, added test.

## Summary

Describe your changes here.


BillWagner (approved):
Thanks for making this update @KevinSWork 

I'll :shipit: now.

PR #23511, opened 2021-03-26
Remove text that was causing loc issues

tdykstra:
Fixes https://github.com/dotnet/docs.pl-pl/issues/206


gewarren (approved):
It would be good to understand why "(zero seconds)" caused a loc issue?

tdykstra:
Because zero isn't zero seconds, it's a number that gets interpreted to mean "use the default", which is 300 seconds.

gewarren:
Oops, this should be reverted. The base branch was live.

BillWagner:
> Oops, this should be reverted. The base branch was live.

Or just cherry-pick it into master.

PR #23516, opened 2021-03-26
Remove text that was causing loc issue

tdykstra:
Fixes https://github.com/dotnet/docs.pl-pl/issues/206


gewarren (approved):

tdykstra:
Thanks for the reviews and for catching the mistake, @gewarren !

PR #23520, opened 2021-03-28
Update dependency-injection.md

xp44mm:
it repeats with line 170th in same file.


IEvangelist (approved):

PR #23522, opened 2021-03-28
C# patterns reference: revised the discard pattern section

pkulikov:
Contributes to #20584

### DIFF
--- a/docs/csharp/language-reference/operators/patterns.md
+++ b/docs/csharp/language-reference/operators/patterns.md
@@ -172,7 +172,7 @@ Beginning with C# 8.0, you use a *discard pattern* `_` to match any expression, 
 
 :::code language="csharp" source="snippets/patterns/DiscardPattern.cs" id="BasicExample":::
 
-In the preceding example, a discard pattern is used to handle `null` and any integer value that doesn't have the corresponding member of the <xref:System.DayOfWeek> enumeration. That guarantees that a `switch` expression in the example handles all possible values of its input. If you don't use a discard pattern in a `switch` expression and none of the expression's patterns matches an input, the runtime [throws an exception](switch-expression.md#non-exhaustive-switch-expressions).The compiler generates a warning if a `switch` expression doesn't handle all possible input values. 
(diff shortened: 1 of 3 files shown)


docs/csharp/language-reference/operators/patterns.md
```diff
@@ -176,6 +166,16 @@ In the preceding example, pattern `var (x, y)` is equivalent to a [positional pa
 
 In a `var` pattern, the type of a declared variable is the compile-time type of the expression that is matched against the pattern.
 
+## Discard pattern
+
+Beginning with C# 8.0, you use a *discard pattern* `_` to match any expression, including `null`, as the following example shows:
+
+:::code language="csharp" source="snippets/patterns/DiscardPattern.cs" id="BasicExample":::
+
+In the preceding example, a discard pattern is used to handle `null` and any integer value that doesn't have the corresponding member of the <xref:System.DayOfWeek> enumeration. That guarantees that a `switch` expression in the example handles all possible values of its input. If you don't use a discard pattern in a `switch` expression and none of the expression's patterns matches an input, the runtime [throws an exception](switch-expression.md#non-exhaustive-switch-expressions).
```
BillWagner, line 175: I'd add that the compiler generates a warning in this situation:

```suggestion
In the preceding example, a discard pattern is used to handle `null` and any integer value that doesn't have the corresponding member of the <xref:System.DayOfWeek> enumeration. That guarantees that a `switch` expression in the example handles all possible values of its input. The compiler generates a warning if a `switch` expression doesn't handle all possible values of the input type. If you don't use a discard pattern in a `switch` expression and none of the expression's patterns matches an input, the runtime [throws an exception](switch-expression.md#non-exhaustive-switch-expressions).
```

BillWagner (changes requested):
Thanks again @pkulikov 

I had one comment that we should include information on the compiler warning when a `switch` expression doesn't handle all input values.

My suggestion feels a bit stilted, so feel free to word differently.

docs/csharp/language-reference/operators/patterns.md
pkulikov: @BillWagner why do you want to include that sentence here, in the patterns reference? How would it help a reader to learn about a discard pattern?

pkulikov: I will definitely include that sentence into the article about a switch expression when I come to updating it. However, I doubt if it's helpful in this article.

pkulikov: @BillWagner thank you for more context; I've updated the text.

BillWagner: Thanks @pkulikov 

There's one markdown lint error (extra blank space). Once that's fixed, I'll approve and merge.

BillWagner: and of course, you added the commit while I was typing. Merging.

BillWagner (approved):

PR #23523, opened 2021-03-28
Porting Feedback from MR

ardalis:
## Summary

Updates and fixes to Porting eBook based on MR feedback.


ardalis:
@mjrousos

docs/architecture/porting-existing-aspnet-apps/example-migration-eshop.md
```diff
@@ -320,7 +319,7 @@ Then replace the content of the footer in the layout with the following code:
 <section class="col-sm-6">
     <img class="esh-app-footer-text hidden-xs" src="~/images/main_footer_text.png" width="335" height="26" alt="footer text image" />
     <br />
-<small>@eShopPorted.Startup.MachineName - @eShopPorted.Startup.StartTime.ToString() UTC</small>
+<small>@Environment.MachineName - @eShopPorted.Startup.StartTime.ToString() UTC</small>
```
mjrousos, line 322: Maybe use `Process.GetCurrentProcess().StartTime` to avoid needing to store a static in Startup.cs?

In other scenarios where a method like that isn't available, a more DI-friendly architecture might be to have a stateful service registered as a singleton and request that, as needed, to set or update state. That will allow views to not depend directly on startup.

docs/architecture/porting-existing-aspnet-apps/understand-update-dependencies.md
```diff
@@ -41,7 +41,7 @@ Analyze your use of third-party NuGet packages and determine if any of them don'
...
+In some cases, no version of a given package works with .NET Core. In that case, teams have a couple options. They can continue depending on the .NET Framework version, but this has limitations. The app will may only run on Windows, and the team may want to run Portability Analyzer on the package's binaries to see if there are any issues likely to be encountered. Certainly the team will want to test thoroughly, since if .NET Framework packages are used that reference APIs not available in .NET Core, a runtime exception will occur. The other option is to find a different package or, if the required package is open source, upgrade it to .NET Standard or .NET Core themselves.
```
mjrousos, line 44: You have both "will" and "may" here now; can probably lose the "will"

mjrousos (approved):
Looks good. A couple suggestions inline.

docs/architecture/porting-existing-aspnet-apps/example-migration-eshop.md
ardalis: That's a good idea - I'll use `Process.GetCurrentProcess().StartTime`, thanks!

docs/architecture/porting-existing-aspnet-apps/understand-update-dependencies.md
ardalis: The difference is because the exception only *will* be thrown if they reference APIs not available in Core. The app *may* run in non-Windows environments if this isn't the case. Does that make sense?

ardalis:
@IEvangelist can you take a quick look? Probably good to 🚢 at this point.

docs/architecture/porting-existing-aspnet-apps/configuration-differences.md
```diff
@@ -54,7 +54,7 @@ public class TestModel : PageModel
 
 **Figure 2-2.** Accessing configuration values with `IConfiguration`.
 
-Using the options pattern, settings access is similar but is strongly typed and more specific to the setting(s) needed by the consuming class, as Figure 2-3 demonstrates.
+Using the [options pattern](/aspnet/core/fundamentals/configuration/options), settings access is similar but is strongly typed and more specific to the setting(s) needed by the consuming class, as Figure 2-3 demonstrates.
```
IEvangelist, line 57:
```suggestion
Using the [options pattern](/dotnet/core/extensions/options), settings access is similar but is strongly typed and more specific to the setting(s) needed by the consuming class, as Figure 2-3 demonstrates.
```

Most (if not all) of the extensions / fundamentals content was intentionally moved from ASP.NET docs to .NET docs instead, as it is applicable to all .NET apps - not just ASP.NET. With that said, I think it is appropriate to use the .NET docs like for the options pattern here instead.

docs/architecture/porting-existing-aspnet-apps/understand-update-dependencies.md
IEvangelist, line 44:
```suggestion
In some cases, no version of a given package works with .NET Core. In that case, teams have a couple options. They can continue depending on the .NET Framework version, but this has limitations. The app may only run on Windows, and the team may want to run Portability Analyzer on the package's binaries to see if there are any issues likely to be encountered. Certainly the team will want to test thoroughly, since if .NET Framework packages are used that reference APIs not available in .NET Core, a runtime exception will occur. The other option is to find a different package or, if the required package is open source, upgrade it to .NET Standard or .NET Core themselves.
```

Remove `"will"` as @mjrousos suggested

IEvangelist (commented):
Hi @ardalis - this is awesome. A few minor things, when these are addressed - I'll :shipit:

docs/architecture/porting-existing-aspnet-apps/understand-update-dependencies.md
ardalis: I was looking at the wrong instance of "will". Sounds good.

ardalis:
Should be all set now!

nishanil (approved):
LGTM!

PR #23535, opened 2021-03-29
Document the correct way to apply an F# attribute to a .NET module.

teo-tsirpanis:
Using `[<module: MyAttribute>]` fails with a mysterious compiler error, and the correct way is not apparent. This PR documents that it is ```[<``module``: MyAttribute>]```.


cartermp (approved):
Thanks!

PR #23556, opened 2021-03-30
Document that bash is required for the install script

danmoseley:
Unless and until https://github.com/dotnet/install-scripts/issues/173 is fixed.

cc @mthalman @bozturkMSFT

### DIFF
--- a/docs/core/install/linux-scripted-manual.md
+++ b/docs/core/install/linux-scripted-manual.md
@@ -77,6 +77,12 @@ For .NET apps that use the *System.Drawing.Common* assembly, you'll also need th
 
 The [dotnet-install scripts](../tools/dotnet-install-script.md) are used for automation and non-admin installs of the **SDK** and **Runtime**. You can download the script from <https://dot.net/v1/dotnet-install.sh>.
 
+The script currently requires bash. On some distros this needs to be installed. On Alpine, for example, you would install bash with:
+
+```bash
+apk add bash
+```
+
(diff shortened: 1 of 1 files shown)


docs/core/install/linux-scripted-manual.md
```diff
@@ -77,6 +77,12 @@ For .NET apps that use the *System.Drawing.Common* assembly, you'll also need th
 
 The [dotnet-install scripts](../tools/dotnet-install-script.md) are used for automation and non-admin installs of the **SDK** and **Runtime**. You can download the script from <https://dot.net/v1/dotnet-install.sh>.
 
+The script currently requires bash. On some distros this needs to be installed. On Alpine, for example, you would install bash with:
+
+```bash
+apk add bash
+```
+
```
adegeo, line 85: For simplicity, I would just make this generic.

```suggestion
>![IMPORTANT]
>Bash is required to run the script.
```

danmoseley: Note that in the distro specific topics we helpfully share the command to use, eg https://docs.microsoft.com/en-us/dotnet/core/install/linux-alpine#dependencies

But I think we can assume folks know (or can figure out) how to install it.

adegeo (approved):

adegeo:
@danmoseley Are you happy with me merging this without the other's review?

bekir-ozturk (approved):
Thanks a lot @danmoseley. You are always one step ahead of us 🙂

danmoseley:
great @adegeo I think we can merge.

mthalman (approved):

PR #23586, opened 2021-04-01
Lesser -> greater

### LINKED ISSUE #23558: Mismatch in HeapHardLimitPercent documentation
Heap limit percent - https://docs.microsoft.com/en-us/dotnet/core/run-time-config/garbage-collector#heap-limit-percent
This link says - "**The default value, which only applies in certain cases, is the lesser of 20 MB or 75% of the memory limit on the container. The default value applies if:**
Specifies the allowable GC heap usage as a percentage of the total physical memory.

If System.GC.HeapHardLimit is also set, this setting is ignored.

This setting only applies to 64-bit computers.

If the process is running inside a container that has a specified memory limit, the percentage is calculated as a percentage of that memory limit.

This setting is ignored if the Per-object-heap limits are configured.

**The default value, which only applies in certain cases, is the lesser of 20 MB or 75% of the memory limit on the container. The default value applies if:**

But the below links says max of the both values - max (20mb, 75% of the memory limit on the container)
https://devblogs.microsoft.com/dotnet/running-with-server-gc-in-a-small-container-scenario-part-1-hard-limit-for-the-gc-heap/

When a process is running inside a container, and if you have specified a memory limit on the container, the process is not supposed to use more than that much memory. However, as we mentioned the GC heap usage is not the only memory contender in the process so by default we specify this as the hard limit for the GC heap usage inside a container

max (20mb, 75% of the memory limit on the container)

Can we please clarify what is the default when not specified ? 

---
#### Document Details

* ID: e544aba5-cfe6-1790-023e-bec738182ea8
* Version Independent ID: 8392a21a-8404-d352-ffb1-d6373bbbdd1b
* Content: [Garbage collector config settings - .NET](https://docs.microsoft.com/en-us/dotnet/core/run-time-config/garbage-collector#heap-limit-percent)
* Content Source: [docs/core/run-time-config/garbage-collector.md](https://github.com/dotnet/docs/blob/main/docs/core/run-time-config/garbage-collector.md)
* GitHub Login: @gewarren

gewarren:
Fixes #23558

### DIFF
--- a/docs/core/run-time-config/garbage-collector.md
+++ b/docs/core/run-time-config/garbage-collector.md
@@ -282,7 +282,7 @@ Example:
 - This setting only applies to 64-bit computers.
 - If the process is running inside a container that has a specified memory limit, the percentage is calculated as a percentage of that memory limit.
 - This setting is ignored if the [Per-object-heap limits](#per-object-heap-limits) are configured.
-- The default value, which only applies in certain cases, is the lesser of 20 MB or 75% of the memory limit on the container. The default value applies if:
+- The default value, which only applies in certain cases, is the greater of 20 MB or 75% of the memory limit on the container. The default value applies if:
 
   - The process is running inside a container that has a specified memory limit.
   - [System.GC.HeapHardLimit](#heap-limit) is not set.


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

Issue #22021, opened 2020-12-16
[WinForms] API validate input parameters and throw ANE


RussKie:
## Selected API validate input parameters and throw `ArgumentNullException`.

Selected API now validate input parameters and throw `ArgumentNullException` where previously a `NullReferenceException` could be thrown, if APIs were invoked with `null` input arguments.
Throwing `ArgumentException` is inline with the behaviours in .NET Runtime, and provides developers with a better debug experience clearly communicating which argument caused the exception.

### Old behavior

The listed API would throw `NullReferenceException` if invoked with `null` input arguments.

### New behavior

The listed API would throw `ArgumentNullException` if invoked with `null` input arguments.

### Reason for change

Throwing `ArgumentNullException` is inline with the behaviours in .NET Runtime, and provides developers with a better debug experience clearly communicating which argument caused the exception.

### Recommended action

- Review and, if necessary, update the code to prevent passing `null` input arguments in the listed API.
- If the code is handling `NullReferenceException` replace (or supplement) it with handling `ArgumentNullException`.

### Category

- Windows Forms

#### Affected APIs

* [TreeNodeCollection](https://docs.microsoft.com/dotnet/api/system.windows.forms.treenodecollection)
	- [x] [TreeNodeCollection.Item[Int32] Property](https://docs.microsoft.com/dotnet/api/system.windows.forms.treenodecollection.item#System_Windows_Forms_TreeNodeCollection_Item_System_Int32_)

#### Affected APIs

* [DrawTreeNodeEventArgs](https://docs.microsoft.com/dotnet/api/system.windows.forms.DrawTreeNodeEventArgs)
	- [x] [DrawTreeNodeEventArgs(Graphics, TreeNode, Rectangle, TreeNodeStates) Constructor](https://docs.microsoft.com/dotnet/api/system.windows.forms.drawtreenodeeventargs.-ctor) when `graphics` is `null`

#### Affected APIs

* [DataGridViewRowStateChangedEventArgs](https://docs.microsoft.com/dotnet/api/system.windows.forms.DataGridViewRowStateChangedEventArgs)
	- [x] [DataGridViewRowStateChangedEventArgs(DataGridViewRow, DataGridViewElementStates) Constructor](https://docs.microsoft.com/dotnet/api/system.windows.forms.datagridviewrowstatechangedeventargs.-ctor) when `dataGridViewRow` is `null`

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
[...]

PR #23588, opened 2021-04-01
Add ANE breaking change

gewarren:
Contributes to #22021


gewarren:
The preview is giving a 404 for some reason, but this PR just adds one line, so I don't think we need to check it.

RussKie (approved):

PR #23593, opened 2021-04-02
Fix grammar

arwn:
Change 'of' to 'with'.

### DIFF
--- a/docs/csharp/tour-of-csharp/program-building-blocks.md
+++ b/docs/csharp/tour-of-csharp/program-building-blocks.md
@@ -14,7 +14,7 @@ The members of a `class` are either ***static members*** or ***instance members*
 The following list provides an overview of the kinds of members a class can contain.
 
 - **Constants**: Constant values associated with the class
-- **Fields**:  Variables that are associated of the class
+- **Fields**:  Variables that are associated with the class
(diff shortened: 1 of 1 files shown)


docs/csharp/tour-of-csharp/program-building-blocks.md
```diff
@@ -14,7 +14,7 @@ The members of a `class` are either ***static members*** or ***instance members*
 The following list provides an overview of the kinds of members a class can contain.
 
 - **Constants**: Constant values associated with the class
-- **Fields**:  Variables that are associated of the class
+- **Fields**:  Variables that are associated with the class
```
Youssef1313, line 17: The grammar change looks good. But I think the sentence was (and still) incorrect.

Fields are associated with *objects* of the class.
Static fields are associated with the *class*.

@BillWagner @IEvangelist Thoughts?

BillWagner: @Youssef1313 

I agree. If you look at the surrounding text, only the **Indexers** description uses "instance of the class". I'll approve and merge this, and we should create a new PR for that change.

BillWagner (approved):
Thanks @arwn 

This LGTM, I'll :shipit: now.

PR #23601, opened 2021-04-02
Invalid path chars breaking change

### LINKED ISSUE #20004: System.IO APIs throw exceptions for invalid paths only when used, not when normalizing
## System.IO APIs throw exceptions for invalid paths only when used, not when normalizing

See https://docs.microsoft.com/en-us/archive/blogs/jeremykuhne/system-io-in-net-core-2-1-sneak-peek
See https://github.com/dotnet/runtime/issues/38487#issuecomment-650665737

I'll let a product team member writes this in a more organized way.

### Version introduced

.NET Core 2.1

### Old behavior

System.IO APIs used to throw exceptions for invalid paths when normalizing a path.

### New behavior

System.IO APIs are no longer throwing exceptions when normalizing a path. They only throw exceptions when *using* the path.

### Reason for change

Leaving this to the product team to write.

### Recommended action

Leaving this to the product team to write.

### Category

- Core .NET libraries

### Affected APIs

I don't have a complete list of the affected APIs.

### Note

The API reference (for example, for [FileInfo ctor](https://docs.microsoft.com/en-us/dotnet/api/system.io.fileinfo.-ctor?view=netcore-3.1)) is outdated. It states that it will throw an exception for invalid paths.

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #20004.

[Internal preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/2.1?branch=pr-en-us-23601#path-apis-dont-throw-an-exception-for-invalid-characters).


tdykstra (approved):
LGTM

includes/core-changes/corefx/2.1/path-apis-dont-throw-exception-for-invalid-paths.md
```diff
@@ -0,0 +1,53 @@
+### Path APIs don't throw an exception for invalid characters
+
+<xref:System.IO.Path?displayProperty=fullName> APIs no longer validate path characters and throw an <xref:System.ArgumentException> if an invalid character is found.
```
Youssef1313, line 3: System.IO.FileInfo was also affected (https://github.com/dotnet/runtime/issues/38487#issuecomment-650665737). I'm not sure what other System.IO APIs were changed.

gewarren: I saw that, but I think it's because it calls GetFilePath.

Youssef1313: @gewarren Should still be documented (I think). This is an implementation detail that API consumers aren't usually aware of.

carlossanlop (approved):

includes/core-changes/corefx/2.1/path-apis-dont-throw-exception-for-invalid-paths.md
gewarren: Okay, I see that the [FileInfo constructor docs](https://docs.microsoft.com/en-us/dotnet/api/system.io.fileinfo.-ctor?view=net-5.0#System_IO_FileInfo__ctor_System_String_) do mention the exception thrown for invalid path characters, so I think you're right. I've updated that exception (and a whole bunch of others that I think are also affected) in https://github.com/dotnet/dotnet-api-docs/pull/6608. I'm wondering if all of those APIs should be mentioned here.

PR #23605, opened 2021-04-02
Add status check for build warnings

gewarren:
This workflow calls a GitHub action that creates a new status check with state=failure IFF there are OPS build warnings.

### DIFF
--- a/.github/workflows/check-for-build-warnings.yml
+++ b/.github/workflows/check-for-build-warnings.yml
@@ -1,0 +1,10 @@
+on: [pull_request]
+
+jobs:
+  status_checker_job:
+    runs-on: ubuntu-latest
+    name: Checks the OPS status check for build warnings
+    steps:
+    - uses: dotnet/samples/.github/actions/status-checker@main
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/check-for-build-warnings.yml
+++ b/.github/workflows/check-for-build-warnings.yml
@@ -1,4 +1,12 @@
-on: [pull_request]
+on: 
+  pull_request:
+    types: [opened, synchronize, reopened]
+  workflow_dispatch:
+    inputs:
+      reason:
+        description: 'The reason for running the workflow'
+        required: true
+        default: 'Manual run'
 
 jobs:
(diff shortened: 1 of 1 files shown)


gewarren:
You can see that the new workflow/check ran successfully even in this PR.

.github/workflows/check-for-build-warnings.yml
```diff
@@ -0,0 +1,10 @@
+on: [pull_request]
+
+jobs:
+  status_checker_job:
+    runs-on: ubuntu-latest
+    name: Checks the OPS status check for build warnings
+    steps:
+    - uses: dotnet/samples/.github/actions/status-checker@main
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
```
Youssef1313, line 10: Do we want it to run on all branches?
For example, see https://github.com/dotnet/docs/pull/23568 which doesn't target `main` or `live`, do we want to run this check in such PRs?

.github/workflows/check-for-build-warnings.yml
```diff
@@ -0,0 +1,10 @@
+on: [pull_request]
```
BillWagner, line 1: We've started making the following on other actions:

```suggestion
on: 
  pull_request:
    types: [opened, synchronize, reopened]
  workflow_dispatch:
    inputs:
      reason:
        description: 'The reason for running the workflow'
        required: true
        default: 'Manual run'
```

The first change runs the workflow only when a PR is opened, the branch is updated, or the PR is re-opened. That prevents it running when labels are added, or other PR events are triggered.

The second lets us run it manually (mostly for testing). The extra arguments require the person starting the manual process to state a reason for the run.

.github/workflows/check-for-build-warnings.yml
BillWagner: I think so. Let's see how much noise it generates.

BillWagner (approved):
This looks great @gewarren 

You can :shipit: when ready.

.github/workflows/check-for-build-warnings.yml
gewarren: Yes, I'd say we want this to run on all branches.

PR #23619, opened 2021-04-04
document GC committed counter

sywhang:
## Summary

This documents the GC committed bytes counter that was added via https://github.com/dotnet/runtime/pull/50604.


IEvangelist (approved):

PR #23627, opened 2021-04-05
fix typo

mweeden2:
There is an extra "the" on line 37:
"In the case of the ref assignment operator, the both of its operands must be of the same type."
should be
"In the case of the ref assignment operator, both of its operands must be of the same type."

## Summary

Describe your changes here.


BillWagner (approved):
Thanks @mweeden2 for spotting this and fixing it. We appreciate it.

I'll :shipit: now. You should see the change on the live site tomorrow on our regular publishing cycle.

PR #23631, opened 2021-04-05
var keyword reference: mention a var pattern

pkulikov:


BillWagner (approved):
This LGTM @pkulikov 

I'll :shipit: now.

PR #23634, opened 2021-04-05
Give workflow a name

gewarren:
Give workflow a name so it doesn't appear.

Old look:

image

New look:

image

### DIFF
--- a/.github/workflows/check-for-build-warnings.yml
+++ b/.github/workflows/check-for-build-warnings.yml
@@ -3,12 +3,6 @@ name: 'Status checker'
 on: 
   pull_request:
     types: [opened, synchronize, reopened]
-  workflow_dispatch:
-    inputs:
-      reason:
-        description: 'The reason for running the workflow'
-        required: true
-        default: 'Manual run'
 
 jobs:
   status_checker_job:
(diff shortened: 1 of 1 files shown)


.github/workflows/check-for-build-warnings.yml
```diff
@@ -10,8 +12,8 @@ on:
```
gewarren, line 12: @BillWagner For the manual run to be useful, we'd have to pass in a PR number. Do you want me to change the action to take a PR number as input?

https://github.com/dotnet/docs/runs/2273417489?check_suite_focus=true

BillWagner: Ping @IEvangelist who taught me this trick on the manual run.

BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

.github/workflows/check-for-build-warnings.yml
IEvangelist: @gewarren yes, you'd have to update the action to take an optional PR number. Then you'd essentially [coalesce on the payload](https://github.com/dotnet/samples/blob/main/.github/actions/status-checker/src/status-checker.ts#L10), as that would be `null` when manually ran.

gewarren: Thanks, I think it's best to just remove the manual trigger for now. We can add it later if we decide it's needed.

PR #23645, opened 2021-04-06
Moved record from reserved to contextual keywords

pkulikov:
The following code compiles:
```csharp
public static class Program
{
    public record A;

    public static void Main()
    {
        var record = new A();
    }
}
```

So, `record` can be a variable's name and it's not a reserved keyword, but contextual.


BillWagner (approved):
Thanks @pkulikov 

I'll :shipit: now.

PR #23646, opened 2021-04-06
Type-testing is operator: simplified the section about pattern matching

pkulikov:
Removed the information that is already presented in another article.


BillWagner (approved):
This LGTM @pkulikov 

I'll :shipit: now.

PR #23697, opened 2021-04-07
Add method naming rule

### LINKED ISSUE #21412: Method naming style
It seems like default naming style for class methods is Pascal Case, it should be mentioned here https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/naming-rules#default-naming-styles

---
#### Document Details

* ID: 8cfcd87a-a475-5fcf-2344-2b3a706a4b6d
* Version Independent ID: 95aa3bbf-7621-2064-f9d0-55105d55ee7e
* Content: [Code style naming rules - .NET](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/naming-rules)
* Content Source: [docs/fundamentals/code-analysis/style-rules/naming-rules.md](https://github.com/dotnet/docs/blob/master/docs/fundamentals/code-analysis/style-rules/naming-rules.md)
* Product: **dotnet**
* GitHub Login: @gewarren

adegeo:
## Summary

Add reference to `methods`.

Fixes #21412

### DIFF
--- a/docs/fundamentals/code-analysis/style-rules/naming-rules.md
+++ b/docs/fundamentals/code-analysis/style-rules/naming-rules.md
@@ -129,9 +129,9 @@ The order in which naming rules are defined in an EditorConfig file doesn't matt
 
 If you don't specify any custom naming rules, the following default styles are used:
 
-- For classes, structs, enumerations, properties, methods, and events with `public`, `private`, `internal`, `protected`, or `protected_internal` accessibility, the default naming style is Pascal case.
+- For classes, structs, enumerations, properties, methods, and events with `public`, `private`, `internal`, `protected`, `private_protected`, or `protected_internal` accessibility, the default naming style is Pascal case.
 
-- For interfaces with `public`, `private`, `internal`, `protected`, or `protected_internal` accessibility, the default naming style is Pascal case with a required prefix of **I**.
+- For interfaces with `public`, `private`, `internal`, `protected`, `private_protected`, or `protected_internal` accessibility, the default naming style is Pascal case with a required prefix of **I**.
 
 ## <a name="rule-id-ide1006-naming-rule-violation"></a>Code Rule ID: `IDE1006 (Naming rule violation)`
 
--- a/docs/fundamentals/code-analysis/style-rules/naming-rules.md
+++ b/docs/fundamentals/code-analysis/style-rules/naming-rules.md
@@ -129,9 +129,9 @@ The order in which naming rules are defined in an EditorConfig file doesn't matt
 
 If you don't specify any custom naming rules, the following default styles are used:
 
-- For classes, structs, enumerations, properties, methods, and events with `public`, `private`, `internal`, `protected`, `private_protected`, or `protected_internal` accessibility, the default naming style is Pascal case.
+- For classes, structs, enumerations, properties, methods, and events with any accessibility, the default naming style is Pascal case.
 
-- For interfaces with `public`, `private`, `internal`, `protected`, `private_protected`, or `protected_internal` accessibility, the default naming style is Pascal case with a required prefix of **I**.
+- For interfaces with any accessibility, the default naming style is Pascal case with a required prefix of **I**.
 
 ## <a name="rule-id-ide1006-naming-rule-violation"></a>Code Rule ID: `IDE1006 (Naming rule violation)`
 
--- a/docs/fundamentals/code-analysis/style-rules/naming-rules.md
+++ b/docs/fundamentals/code-analysis/style-rules/naming-rules.md
@@ -129,7 +129,7 @@ The order in which naming rules are defined in an EditorConfig file doesn't matt
 
 If you don't specify any custom naming rules, the following default styles are used:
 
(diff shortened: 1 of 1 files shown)


gewarren (approved):

docs/fundamentals/code-analysis/style-rules/naming-rules.md
```diff
@@ -129,7 +129,7 @@ The order in which naming rules are defined in an EditorConfig file doesn't matt
 
 If you don't specify any custom naming rules, the following default styles are used:
 
-- For classes, structs, enumerations, properties, and events with `public`, `private`, `internal`, `protected`, or `protected_internal` accessibility, the default naming style is Pascal case.
+- For classes, structs, enumerations, properties, methods, and events with `public`, `private`, `internal`, `protected`, or `protected_internal` accessibility, the default naming style is Pascal case.
 
 - For interfaces with `public`, `private`, `internal`, `protected`, or `protected_internal` accessibility, the default naming style is Pascal case with a required prefix of **I**.
```
Youssef1313, line 134: Should these include `private_protected` as well? @mavasani to confirm.

adegeo: Yep. I just tested in visual studio

gewarren: Maybe we should just list the accessibilities it doesn't apply to, since the list is getting so long.

adegeo: @gewarren I don't think that's a good idea actually. I think it's better to be explicit and verbose. If someone is looking for a specific keyword they'll want to see it called out rather than omitted.

gewarren: Actually, these 6 accessibilities [appear to be *all* the accessibilities](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/accessibility-levels). So I'm really not sure why we even need to mention them here. It leaves you thinking, "so what doesn't this apply to?".

gewarren: Maybe if you don't specify any accessibility, the rule doesn't apply? In that case, we could say something like "For methods that have an accessibility level specified, the default naming rule is Pascal case."

Youssef1313: > Maybe if you don't specify any accessibility, the rule doesn't apply?

The rule will still apply. For example, a class with no declared accessibility in source will have an effective accessibility of "internal".

adegeo: That is true, it's all of the accessibilities, but it's not applying to `field` or `variable`. Maybe calling those two out specifically would be useful? @BillWagner

docs/fundamentals/code-analysis/style-rules/naming-rules.md
```diff
@@ -129,9 +129,9 @@ The order in which naming rules are defined in an EditorConfig file doesn't matt
 
 If you don't specify any custom naming rules, the following default styles are used:
 
-- For classes, structs, enumerations, properties, and events with `public`, `private`, `internal`, `protected`, or `protected_internal` accessibility, the default naming style is Pascal case.
+- For classes, structs, enumerations, properties, methods, and events with `public`, `private`, `internal`, `protected`, `private_protected`, or `protected_internal` accessibility, the default naming style is Pascal case.
```
gewarren, line 132:
```suggestion
- For classes, structs, enumerations, properties, methods, and events with any accessibility, the default naming style is Pascal case.
```

docs/fundamentals/code-analysis/style-rules/naming-rules.md
```diff
@@ -129,9 +129,9 @@ The order in which naming rules are defined in an EditorConfig file doesn't matt
...
 
-- For interfaces with `public`, `private`, `internal`, `protected`, or `protected_internal` accessibility, the default naming style is Pascal case with a required prefix of **I**.
+- For interfaces with `public`, `private`, `internal`, `protected`, `private_protected`, or `protected_internal` accessibility, the default naming style is Pascal case with a required prefix of **I**.
```
gewarren, line 134:
```suggestion
- For interfaces with any accessibility, the default naming style is Pascal case with a required prefix of **I**.
```

gewarren (changes requested):

docs/fundamentals/code-analysis/style-rules/naming-rules.md
BillWagner: I like @gewarren 's suggestion. When I read this, you've enumerated *all* the possible access levels. It's simpler to just say that.
 (Historically, I bet that at one time `private` used a different standard).

adegeo:
@gewarren This fell off the radar 😄 I talked with Bill and he agreed your approach is best. So I accepted your changes. Please approve. Thanks!

gewarren (approved):

PR #23716, opened 2021-04-08
Update standard-numeric-format-strings.md

### LINKED ISSUE #21884: Update examples for .NET 5 on Windows
Some of the example output in the table might need to be updated due to the NLS -> ICU change for globalization in .NET 5.

See https://github.com/dotnet/runtime/issues/44587#issuecomment-726364372 as an example.

---
#### Document Details

* ID: 0ae38fce-fc98-9420-6b0b-ceee4e885f2d
* Version Independent ID: f23035e0-f271-b48c-f9a9-8582be80b4ad
* Content: [Standard numeric format strings](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings)
* Content Source: [docs/standard/base-types/standard-numeric-format-strings.md](https://github.com/dotnet/docs/blob/master/docs/standard/base-types/standard-numeric-format-strings.md)
* GitHub Login: @adegeo
* Microsoft Alias: **adegeo**

adegeo:
## Summary

- Added a note detailing how cultural formats may differ.

Fixes #21884

@gewarren according to the linked issue in the runtime repo, the team was unable to reproduce the problem. I too cannot reproduce it natively unless I start tweaking cultural settings, than the results may be different between versions. Hopefully this note will clarify this scenario. I linked to the ICU article too.

What do you think I should do with the note? Move it near the top of the article or put it within each section? I think it should be at the top, but people may just be looking at specific format strings instead of the article overall.

### DIFF
--- a/docs/standard/base-types/standard-numeric-format-strings.md
+++ b/docs/standard/base-types/standard-numeric-format-strings.md
@@ -50,6 +50,8 @@ Standard numeric format strings are supported by:
 
 The following table describes the standard numeric format specifiers and displays sample output produced by each format specifier. See the [Notes](#notes) section for additional information about using standard numeric format strings, and the [Code example](#code-example) section for a comprehensive illustration of their use.
 
+> The result of a formatted string for a specific culture might differ from the following examples. Operating system settings, user settings, environment variables, and the .NET version you're using can all affect the format. For example, starting with .NET 5, .NET tries to unify cultural formats across platforms. For more information, see [.NET globalization and ICU](../globalization-localization/globalization-icu.md).
+
 |Format specifier|Name|Description|Examples|
 |----------------------|----------|-----------------|--------------|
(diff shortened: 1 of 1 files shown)


docs/standard/base-types/standard-numeric-format-strings.md
```diff
@@ -252,6 +252,9 @@ The following example formats assorted floating-point values with the number for
 [!code-csharp[Formatting.Numeric.Standard#6](../../../samples/snippets/csharp/VS_Snippets_CLR/Formatting.Numeric.Standard/cs/Standard.cs#6)]
 [!code-vb[Formatting.Numeric.Standard#6](../../../samples/snippets/visualbasic/VS_Snippets_CLR/Formatting.Numeric.Standard/vb/Standard.vb#6)]
 
+> [!IMPORTANT]
+> The format for a specific culture may change over time. Operating system settings, user settings, environment variables, and changes to .NET itself, may affect the format of a specific culture. Starting with .NET 5.0, .NET tries to unify cultural formats across platforms. For more information, see [.NET globalization and ICU](../globalization-localization/globalization-icu.md).
```
gewarren, line 256:
```suggestion
> The format you see for a specific culture might differ from the following examples. Operating system settings, user settings, environment variables, and the .NET version you're using can all affect the format. For example, starting with .NET 5, .NET tries to unify cultural formats across platforms. For more information, see [.NET globalization and ICU](../globalization-localization/globalization-icu.md).
```

gewarren (approved):
I think I would put the note right above the table (between the intro text for the "Standard format specifiers" section and the table).

docs/standard/base-types/standard-numeric-format-strings.md
adegeo: What about adding **result**, as it's not the format that changes, but the result.

```suggestion
> The result of a formatted string for a specific culture might differ from the following examples. Operating system settings, user settings, environment variables, and the .NET version you're using can all affect the format. For example, starting with .NET 5, .NET tries to unify cultural formats across platforms. For more information, see [.NET globalization and ICU](../globalization-localization/globalization-icu.md).
```

gewarren: Maybe just "The formatted string for a specific culture...".

adegeo: I still feels like that implies the format string (the "N" you pass to .ToString) is what could change, but it's really the output. OK Thanks for the feedback, I'll fix it up 😄

PR #23735, opened 2021-04-09
Add not null example

BillWagner:
See https://github.com/dotnet/csharplang/discussions/4623#discussioncomment-584988

This is a better example.

### DIFF
--- a/docs/csharp/language-reference/attributes/snippets/NullableAttributes.cs
+++ b/docs/csharp/language-reference/attributes/snippets/NullableAttributes.cs
@@ -53,7 +53,7 @@ namespace attributes
         {
             if (value is null) throw new ArgumentNullException(valueExpression);
         }
-        // </ThrownWhenNull>
+        // </ThrowWhenNull>
 
         // <ExtractComponent>
         string GetTopLevelDomainFromFullUrl(string url)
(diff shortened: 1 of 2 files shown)


docs/csharp/language-reference/attributes/snippets/NullableAttributes.cs
```diff
@@ -0,0 +1,196 @@
...
+
+        // <ExtractComponentIfNotNull>
+        [return: NotNullIfNotNull("url")]
+        string? GetTopLevelDomainFromFullUrl(string? url)
+        // </ExtractComponentIfNotNull>
+        {
+            if (url is null)
+            {
+                return null;
+            } else
```
Youssef1313, line 190:
```suggestion
            }
            else
```

docs/csharp/language-reference/attributes/snippets/NullableAttributes.cs
```diff
@@ -0,0 +1,197 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace attributes
+{
+#nullable disable
+    public class NullableOblivious
+    {
+        // <TryGetExample>
```
CamSoper, line 13: I had no idea this was an allowable way of doing code snippets. I'd always used regions.

CamSoper (approved):

PR #23743, opened 2021-04-10
Note that integral literal values are positive

### LINKED ISSUE #23736: Clarify that hexadecimal literals are assumed to be positive values.
The documentation on numeric literals states the following:

> The type of an integer literal is determined by its suffix as follows:
  If the literal has no suffix, its type is the first of the following types in which its value can be represented: int, uint, long, ulong.

Now consider the literal `0x80000000`. This literal does not have a suffix and the bit pattern fits into a 32-bit integer so this documentation may imply to some readers that this will be treated as a signed integer. (In other languages, PowerShell for instance, this kind of literal is indeed interpreted to be the same as System.Int32.MinValue.)

In C# the apparent rule is that hexadecimal literals are always assumed to be positive values, meaning the positive value 0x80000000 must be interpreted as an unsigned int since it is too large to fit into a signed integer.

This doesn't seem to be clearly documented and it would be good to make this more explicit.

---#### Document Details⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.** ID: 3a22464d-b60c-36c0-e4e8-b209a33b0219* Version Independent ID: cb500323-82a3-09ea-0468-dd16c9e13675* Content: [Integral numeric types - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types)* Content Source: [docs/csharp/language-reference/builtin-types/integral-numeric-types.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/builtin-types/integral-numeric-types.md)* Product: **dotnet-csharp*** Technology: **csharp-language-reference*** GitHub Login: @BillWagner* Microsoft Alias: **wiwagn**

pkulikov:
Fixes #23736

/cc: @Treit

### DIFF
--- a/docs/csharp/language-reference/builtin-types/integral-numeric-types.md
+++ b/docs/csharp/language-reference/builtin-types/integral-numeric-types.md
@@ -1,7 +1,7 @@
 ---
 title: "Integral numeric types - C# reference"
 description: "Learn the range, storage size, and uses for each of the integral numeric types."
-ms.date: 03/17/2021
+ms.date: 04/10/2021
 f1_keywords:
   - "byte_CSharpKeyword"
   - "sbyte_CSharpKeyword"
@@ -55,8 +55,6 @@ System.Int32 b = 123;
 
 The `nint` and `nuint` types in the last two rows of the table are native-sized integers. They are represented internally by the indicated .NET types, but in each case the keyword and the .NET type are not interchangeable. The compiler provides operations and conversions for `nint` and `nuint` as integer types that it doesn't provide for the pointer types `System.IntPtr` and `System.UIntPtr`. For more information, see [`nint` and `nuint` types](nint-nuint.md).
 
-For information about the native-sized integer types, see [`nint` and `nuint`](nint-nuint.md).
-
 The default value of each integral type is zero, `0`. Each of the integral types except the native-sized types has `MinValue` and `MaxValue` constants that provide the minimum and maximum value of that type.
 
 Use the <xref:System.Numerics.BigInteger?displayProperty=nameWithType> structure to represent a signed integer with no upper or lower bounds.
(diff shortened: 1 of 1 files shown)


docs/csharp/language-reference/builtin-types/integral-numeric-types.md
```diff
@@ -55,8 +55,6 @@ System.Int32 b = 123;
 
 The `nint` and `nuint` types in the last two rows of the table are native-sized integers. They are represented internally by the indicated .NET types, but in each case the keyword and the .NET type are not interchangeable. The compiler provides operations and conversions for `nint` and `nuint` as integer types that it doesn't provide for the pointer types `System.IntPtr` and `System.UIntPtr`. For more information, see [`nint` and `nuint` types](nint-nuint.md).
 
-For information about the native-sized integer types, see [`nint` and `nuint`](nint-nuint.md).
```
pkulikov, line 58: This duplicates the preceding paragraph.

docs/csharp/language-reference/builtin-types/integral-numeric-types.md
```diff
@@ -82,6 +80,10 @@ The preceding example also shows the use of `_` as a *digit separator*, which is
 The type of an integer literal is determined by its suffix as follows:
 
 - If the literal has no suffix, its type is the first of the following types in which its value can be represented: `int`, `uint`, `long`, `ulong`.
+
+  > [!NOTE]
+  > Literals are interpreted as positive values. For example, the literal `0xFF_FF_FF_FF` represents the number `4294967295` of the `uint` type, though it has the same bit representation as the number `-1` of the `int` type. If you need a value of a certain type, cast a literal to that type. Use the `unchecked` operator, if a literal value cannot be represented in the target type. For example, `unchecked((int)0xFF_FF_FF_FF)` produces `-1`.
```
Treit, line 85: This is good, although perhaps to be perfectly accurate this should say "Literals are interpreted as positive values, unless preceded by a minus sign." ?

pkulikov: To be perfectly accurate, a minus sign is not the part of a literal. It's the unary minus operator. So, first, a literal is interpreted and then the minus operator is applied to that value. (There are two exceptions to that logic, but it's really the spec material.)

Treit: Ok, I see, thanks!

Treit (approved):

BillWagner (approved):

I'll :shipit: now.

PR #23734, opened 2021-04-09
Remove pull_request trigger

gewarren:
To avoid duplicate status checks on PRs.

image

PR #23744, opened 2021-04-10
Run markdownlint in PR instead of push

Youssef1313:


MSDN-WhiteKnight:
Why is this change needed? How this actually works, will the check run on commits pushed after PR is opened (Github Actions docs are not clear on this matter)?

Youssef1313:
> Why is this change needed? How this actually works, will the check run on commits pushed after PR is opened (Github Actions docs are not clear on this matter)?

I don't know the "detailed" description of both options. But currently the workflow is only run on commits in main branch (i.e, it is run *after* the PR is merged)

BillWagner:
Adding @gewarren @IEvangelist (We're all learning the nuances of when these run).

We should test this to verify. My understanding is that `push` runs when commits are pushed. If you add a branch filter, it will only run when commits are pushed to that branch. I think it should run on any branch when you push commits, including those on your fork.

IEvangelist:
Hi @Youssef1313 - thanks for this contribution, but I do not think we need this.

This change will have the markdown linter ran on PRs only, whereas now it runs on [any push to any branch](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#on) on this repo.

I do not think this change is necessary, the only difference is that with this change it wouldn't run the linter when we merge PRs to main (in that way it might be slightly more efficient), but it's not really really something that we're ever currently held back by.

On the other hand (currently as is) this does serve as a safety net to help ensure that in the rare case that PRs are merged without all checks passing - that we'd more than likely notice linter errors, as the status badge would flip to failing. Again, I think we can leave this as it is without this change.

Youssef1313:
I think the linter error should be detected before the PR is merged. Without this change, it will only be detected *after* it is merged, which isn't ideal behavior IMO.

IEvangelist:
> I think the linter error should be detected before the PR is merged. Without this change, it will only be detected _after_ it is merged, which isn't ideal behavior IMO.

Interesting, I was under the impression that `push` was inclusive of `pull_request`. I think we should have both in that case.

.github/workflows/markdownlint.yml
```diff
@@ -1,7 +1,7 @@
 name: Markdownlint
 
 on:
-  push:
+  pull_request:
```
IEvangelist, line 4:
```suggestion
  [push, pull_request]:
```

.github/workflows/markdownlint.yml
```diff
@@ -1,7 +1,7 @@
 name: Markdownlint
 
 on:
```
IEvangelist, line 3:
```suggestion
on: [push, pull_request]
```

.github/workflows/markdownlint.yml
IEvangelist, line 4:
```suggestion
```

Youssef1313 (commented):
@IEvangelist This will cause PRs that are created from branches within the repository (i.e, not forks) have the action run twice (since they are both push & PR). @gewarren wanted to avoid this duplication in https://github.com/dotnet/docs/pull/23734.

gewarren:
> I think the linter error should be detected before the PR is merged. Without this change, it will only be detected _after_ it is merged, which isn't ideal behavior IMO.

That's not true. The push-triggered check appears in the status checks section of a PR (as does a pull_request-triggered workflow). That's why I [deleted the pull_request part of this workflow](https://github.com/dotnet/docs/pull/23734) on Friday.

Youssef1313:
@gewarren This is only true when the branch of the PR within the repository (i.e, not a fork). See https://github.com/dotnet/docs/pull/23758 for example. It doesn't have markdownlint run.

gewarren:
> Why is this change needed? How this actually works, will the check run on commits pushed after PR is opened (Github Actions docs are not clear on this matter)?

Yes, if the trigger is pull_request, the action is run on each commit in a pull request, which is great.

gewarren (approved):
I think this is a good change as-is.

MSDN-WhiteKnight:
I think ideal setup should be like that:

```
on: 
  push:
    branches:
      - main
      - live
  pull_request:
...
```
so we run checks on push in selected branches, and on pull request for every branch. That way there would be no duplicate checks when PR branch is on main repository. But maybe it's not worth the complexity, as everything that ends up in main comes from PRs anyway.

gewarren:
> so we run checks on push in selected branches, and on pull request for every branch. That way there would be no duplicate checks when PR branch is on main repository. But maybe it's not worth the complexity, as everything that ends up in main comes from PRs anyway.

I like it. Or even better:

```
on: 
  push:
    branches:
      - main
      - live
  pull_request:
    branches-ignore:
      - main
      - live
...
```

Youssef1313:
> ```
>  pull_request:
>    branches-ignore:
>      - main
>      - live
```

I don't think any PRs are ever opened from `main` or `live`. Even the `Merge main into live` PRs are opened from a snapshot branch called `publish-some_id`.

**Comment by gewarren** [comment]
> > ```
> >  pull_request:
> >    branches-ignore:
> >      - main
> >      - live
> > ```
> 
> I don't think any PRs are ever opened from `main` or `live`. Even the `Merge main into live` PRs are opened from a snapshot branch called `publish-some_id`.

I was under the impression this is for the base branch. From https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags:

For a pull_request event, only branches and tags on the base are evaluated.

**Comment by gewarren** [comment]
I see, my proposal would end up with no markdownlint on PRs to main from a fork branch. In that case, I think what @MSDN-WhiteKnight proposed is correct (but I would remove "live" from the push branches since anything going to live has already gone through gates).

```
on: 
  push:
    branches:
      - main
  pull_request:
...
```

I guess the only downside is you don't get markdownlint checks before you create the PR (which is fine with me).

PR #23745, opened 2021-04-10
Conditional operator: moved the example to the beginning of the article

pkulikov:
This PR has a story behind.

Once, a colleague of mine was confused with some C# doc article and very emotionally expressed that confusion to me. I asked to show the article and explain what was wrong. Turned out that the article about the conditional operator (and it was revised by me long time ago, so I was confused too :)) is not very beginner-friendly. Looks like providing a simple example  at the very beginning of the article would make it clearer for the readers who for the first time learn about the conditional operator. So, this PR does that: moves the example usage from the middle of the article to its beginning. Also, I've tried to simplify the example itself.


BillWagner (approved):
I like this change @pkulikov 

Also, I appreciated reading the story behind it. It's a good reminder to make sure all our content is clear for new C# developers.

PR #23752, opened 2021-04-11
Fix example of CA1034

Flash0ver:
No warning is reported when the enclosing type is not externally visible

## Summary

Changed the _accessibility level_ of the enclosing type from `internal` to `public` in order to match the *Cause* of rule _CA1034_:
> An externally visible type contains an externally visible type declaration. . . .

With the current version of the example, _warning CA1034_ is not reported using _.NET SDK 5.0.201_.

See also [Stack Overflow](https://stackoverflow.com/q/67021887/10167996), where this issue was also discussed.


gewarren (approved):
Thank you @Flash0ver . From the description of the rule, this change looks correct to me.

PR #23766, opened 2021-04-13
Document new linker options and trimming libraries

### LINKED ISSUE #16140: Make _TrimmerDefaultAction public
In .NET 6, we are introducing `[assembly: AssemblyMetadata("IsTrimmable", "True")]` as a way to opt a library into trimming. MSBuild metadata can also be used to override this per-assembly.

With this change, the linker has two global knobs to control trimming:
- `--trim-mode`: the trimming granularity for assemblies marked `IsTrimmable`
  - The SDK sets this to `copyused` by default, but it is configurable via the property `TrimMode`
- `--action`: the trimming granularity for assemblies *not* marked `IsTrimmable`
  - The SDK sets this to `copy` by default (so that assemblies which aren't marked `IsTrimmable` are kept).

We should expose a public MSBuild option that maps to the global `--action` (maybe `TrimmerDefaultAction`), which allows developers to enable member-level trimming for the entire application by setting:
```xml
<TrimMode>link</TrimMode>
<TrimmerDefaultAction>link</TrimmerDefaultAction>
```

When we discussed full-app trimming earlier, we were hesitant to make it *too* easy (see the notes under [`TrimAllAssemblies`](https://github.com/mono/linker/blob/main/docs/design/trimmed-assemblies.md#trimallassemblies-global-opt-in)).

On the other hand, the option is already there (introduced in https://github.com/dotnet/sdk/pull/16094), so the real question is just whether it should be exposed to the user (and if so, what the name should be). Note that xamarin-android already has some user-facing knobs to do exactly this (`<AndroidLinkMode>Full</AndroidLinkMode>`).

@marek-safar @vitek-karas @samsp-msft @eerhardt

### LINKED ISSUE #23740: TrimmerRootDescriptor  is .net 5 specific option
`TrimmerRootDescriptor` does not work on .net core 3.1, that option is completely ignored.

It should be mentioned in documentation, otherwise developer is wondering why it does not work.

---
#### Document Details

* ID: 811fc2ce-1f05-5be2-d033-7fce35e49525
* Version Independent ID: fca1c728-4716-8c28-89e1-02eb947bf602
* Content: [Trimming options - .NET](https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming-options)
* Content Source: [docs/core/deploying/trimming-options.md](https://github.com/dotnet/docs/blob/main/docs/core/deploying/trimming-options.md)
* GitHub Login: @sbomer
* Microsoft Alias: **svbomer**

sbomer:
This adds documents for the new linker options and instructions for trimming libraries. The "Trimming libraries" document will be the target of the aka.ms link in [IL2104](https://github.com/mono/linker/blob/main/docs/error-codes.md#il2104-assembly-assembly-produced-trim-warnings-for-more-information-see-httpsakamsdotnet-illinklibraries) and should (hopefully) nicely complement the [blog post](https://github.com/mono/linker/pull/1946) @agocke. For now this mostly uses similar examples to the blog post, but over time I think this would be a good place to add more involved examples that could be useful to developers.

For trimming libraries I think we should fix https://github.com/dotnet/sdk/issues/16140 to get the behavior described in the doc.

This also fixes https://github.com/dotnet/docs/issues/23740.

### Internal preview

✔️ [Prepare libraries for trimming](https://review.docs.microsoft.com/en-us/dotnet/core/deploying/prepare-libraries-for-trimming?branch=pr-en-us-23766)
✔️ [Trimming options](https://review.docs.microsoft.com/en-us/dotnet/core/deploying/trimming-options?branch=pr-en-us-23766)


docs/core/deploying/trimming-libraries.md
```diff
@@ -0,0 +1,161 @@
...
+author: sbomer
+ms.author: svbomer
+ms.date: 08/25/2020
+---
+
+# Trimming libraries
+
+## Trim warnings for libraries
+
+`PublishTrimmed` will produce trim analysis [warnings](trimming-options.md#analysis-warnings) for patterns that are not statically understood to be compatible with trimming, including patterns in your code and in dependencies. You may see warnings like `warning IL2104: Assembly 'SomeAssembly' produced trim warnings`.
```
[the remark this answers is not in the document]
sbomer: Yeah, it was supposed to give some context on where warnings like IL2104 come from. I reorganized it a bit to make it more clear, PTAL

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,163 @@
...
+dotnet publish -c Release
+```
+
+- Publishing a self-contained app ensures that the library is analyzed in a context where its dependencies are available, so that you are alerted if your library uses any code from dependencies that is incompatible with trimming.
+
+- `TrimmerRootAssembly` ensures that every part of the library is analyzed. This is necessary in case the library has `[AssemblyMetadata("IsTrimmable", "True")]`, which would otherwise let trimming remove the unused library without analyzing it.
+
+- `<TrimmerDefaultAction>link</TrimmerDefaultAction>` ensures that only used parts of dependencies are analyzed. Without this option, you would see warnings originating from _any_ part of a dependency that doesn't set `[AssemblyMetadata("IsTrimmable", "True")]`, including parts that are unused by your library.
+
+- `<SuppressTrimAnalysisWarnings>link</SuppressTrimAnalysisWarnings>` is implied by `PublishTrimmed` in the .NET SDK. It is only needed if the app targets form factors like Android where the trim analysis warnings are disabled by default. Consult the documentation for your SDK.
```
tlakollo, line 67: Nit: this should be <SuppressTrimAnalysisWarnings>false<SuppressTrimAnalysisWarnings>

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,163 @@
...
+    {
+        foreach (var method in type.GetMethods())
+        {
+            // ...
+        }
+    }
+}
+```
+
+Here, `Foo` is calling a method which takes a `Type` argument that is annotated with a [`DynamicallyAccessedMembers`](https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.dynamicallyaccessedmembersattribute?view=net-5.0) requirement. The requirement states that the value passed in as this argument must represent a type whose public methods are available. In this case, you can fix this by adding the same requirement to the parameter of `Foo`. Like above, once you have bubbled up such warnings to public APIs, you are done.
```
tlakollo, line 129: Maybe we should put another piece of code with the code changes for this and the other examples a before/after to be more explicit.
Also, I'm not comfortable with using "Like above, once you have bubbled ..." If we update the document we might point to something else, better to put something like "Like with RequiresUnreferencedCode, once you have bubbled ..."

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,163 @@
...
+
+    [RequiresUnreferencedCode("Bar does something incompatible with trimming.")]
+    static void Bar()
+    {
+    }
+}
+```
+
+This means the library calls a method which has explicitly been annotated as incompatible with trimming, using [`RequiresUnreferencedCodeAttribute`](
+https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.requiresunreferencedcodeattribute?view=net-5.0). To get rid of the warning, consider whether `Foo` needs to call `Bar` to do its job. If so, annotate the caller `Foo` with `RequiresUnreferencedCode` as well; this will "bubble up" the warning so that callers of `Foo` get a warning instead. Once you have "bubbled up" the attribute all the way to public APIs (so that these warnings are produced only for public methods, if at all), you are done. Apps which call your library will now get warnings if they call those public APIs, but these will no longer produce warnings like `IL2104: Assembly 'MyLibrary' produced trim warnings`.
```
tlakollo, line 100: Is this the right place to include the DynamicDependency, Descriptor options to solve the RequiresUnreferencedCode warning and be able to use UnconditionalSuppressMessage?

docs/core/deploying/preparing-libraries-for-trimming.md
sbomer: Good catch, thanks!

docs/core/deploying/preparing-libraries-for-trimming.md
sbomer: Yes, I think either in this doc or one next to it would be a good place for these. I might do it in a separate PR if I don't get to it soon.

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,188 @@
...
+
+# Preparing libraries for trimming
+
+## Trim warnings in apps
+
+When publishing an app, `PublishTrimmed` will produce trim analysis [warnings](trimming-options.md#analysis-warnings) (in .NET 6+) for patterns that are not statically understood to be compatible with trimming, including patterns in your code and in dependencies.
+
+You will encounter detailed warnings originating from your own code and `ProjectReference` dependencies. You may also see warnings like `warning IL2104: Assembly 'SomeAssembly' produced trim warnings` for `PackageReference` libraries. This means that the library contained patterns which are not guaranteed to work in the context of the trimmed app, and may result in a broken app. Consider contacting the author to see if the library can be annotated for trimming.
+
+To resolve warnings originating from the app code, jump ahead to the instructions on [resolving trim warnings](resolving-trim-warnings.md). If you are interested in making your own `ProjectReference` libraries trim friendly, follow the instructions to [enable library trim warnings](enabling-library-trim-warnings).
```
eerhardt, line 17: Both links in this line 404 for me.

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,188 @@
...
+
+If your app only uses parts of a library that are compatible with trimming, consider [enabling trimming](trimming-options.md#trim-additional-assemblies) of this library if it is not already being trimmed. This will only produce warnings if your app uses problematic parts of the library. (You can also [show detailed warnings](trimming-options.md#showing-detailed-warnings) for the library to see which parts of it are problematic.)
+
+## Enabling library trim warnings
+
+These instructions show how to enable and resolve static analysis warnings to prepare a library for trimming. Follow these steps if you are authoring a library and either want to proactively make your library trimmable, or have been contacted by app authors who encountered trim warnings from your library.
+
+Use the .NET 6 SDK for the best experience. It is possible to [show analysis warnings](trimming-options.md#analysis-warnings) in .NET 5, but this will show detailed warnings for every library, including framework libraries. These instructions assume you are using the .NET 6 SDK.
```
eerhardt, line 25:
```suggestion
Use the .NET 6 SDK for the best experience. It is possible to [show analysis warnings](trimming-options.md#analysis-warnings) in .NET 5, but not all framework assemblies were annotated correctly in .NET 5. Also, this will show detailed warnings for every library, including framework libraries. These instructions assume you are using the .NET 6 SDK.
```

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,188 @@
...
+
+- `<SuppressTrimAnalysisWarnings>false</SuppressTrimAnalysisWarnings>` is implied by `PublishTrimmed` in the .NET SDK. It is only needed if the app targets form factors like Android where the trim analysis warnings are disabled by default. Consult the documentation for your SDK.
+
+You can also follow the same pattern for multiple libraries, adding them all to the same project as `ProjectReference` and `TrimmerRootAssembly` item to see trim analysis warnings for more than one library at a time, but note that this will warn about dependencies if _any_ of the root libraries use a trim-unfriendly API in a dependency. To see warnings that have to do with only a particular library, reference that library only.
+
+Note that the analysis results depend on the implementation details of your dependencies. If you update to a new version of a dependency, this may introduce analysis warnings if the new version added non-understood reflection patterns, even if there were no API changes. In other words, introducing trim analysis warnings to a library is a breaking change when the library is used with `PublishTrimmed`.
+
+## Resolving trim warnings
```
eerhardt, line 73: How do you feel about combining @agocke's guidance blog here? https://github.com/mono/linker/pull/1946

There are good things about both descriptions. I'd rather have 1 really good one that combines the best of both, than have 2 different ones with different information.

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,188 @@
...
+    public static void Foo(Type type)
+    {
+        // warning IL2067: MyLibrary.Foo(Type): 'type' argument does not satisfy
+        // 'DynamicallyAccessedMemberTypes.PublicMethods' in call to 'MyLibrary.UseMethods(Type)'.
+        // The parameter 't' of method 'MyLibrary.Foo(Type)' does not have matching annotations.
+        UseMethods(type);
+    }
+
+    static void UseMethods(
+        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)]
```
eerhardt, line 129: I think it is better to start with a method that straight uses Reflection, teach how to annotate that method with `DynamicallyAccessedMembers`. Then the next step is to show other methods calling that first method, and how to annotate them.

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,188 @@
...
+    {
+        MyLibrary.type = typeof(System.Tuple);
+    }
+```
+
+In this case the trim analysis will simply keep public methods of `System.Tuple`, and will not produce further warnings.
+
+### Recommendations
+
+In general, try to avoid reflection if possible, and when using reflection limit it in scope so that it is reachable only from a small part of the library. For example, avoid using non-understood patterns in places like static constructors that will result in the warning propagating to all members of the class.
```
eerhardt, line 184:
```suggestion
In general, try to avoid reflection if possible. When using reflection, limit it in scope so that it is reachable only from a small part of the library. For example, avoid using non-understood patterns in places like static constructors that will result in the warning propagating to all members of the class.
```

docs/core/deploying/trimming-options.md
```diff
@@ -9,27 +9,47 @@ ms.date: 08/25/2020
...
+Trimming with `PublishTrimmed` was introduced in .NET Core 3.0. The other options are available only in .NET 5 and above.
+
 ## Enable trimming
 
 - `<PublishTrimmed>true</PublishTrimmed>`
 
    Enable trimming during publish, with the default settings defined by the SDK.
 
-When using `Microsoft.NET.Sdk`, this will perform assembly-level trimming of the framework assemblies from the netcoreapp runtime pack. Application code and non-framework libraries are not trimmed. Other SDKs may define different defaults.
+When using `Microsoft.NET.Sdk`, this will trim the framework assemblies from the netcoreapp runtime pack. Application code and non-framework libraries are not trimmed. Other SDKs may define different defaults.
```
eerhardt, line 20: Not just `framework assemblies`, right? Any assembly marked with `IsTrimmable`, which all framework assemblies are annotated with.

docs/core/deploying/trimming-options.md
```diff
@@ -9,27 +9,47 @@ ms.date: 08/25/2020
...
+- `<TrimMode>link</TrimMode>`
+
+    Enable member-level trimming, which removes unused members from types. This is the default in .NET 6+.
+
 - `<TrimMode>copyused</TrimMode>`
 
    Enable assembly-level trimming, which will keep an entire assembly if any part of it is used (in a statically understood way).
 
-- `<TrimMode>link</TrimMode>`
+Assemblies with `<IsTrimmable>true</IsTrimmable>` metadata but no explicit `TrimMode` will use the global `TrimMode`. The default `TrimMode` for `Microsoft.NET.Sdk` is `link` in .NET 6+, and `copyused` in previous versions.
```
eerhardt, line 34: > Assemblies with `<IsTrimmable>true</IsTrimmable>` metadata

What exactly does this mean? What assemblies? `TrimmableAssembly`? `ManagedAssemblyToLink`?

docs/core/deploying/trimming-options.md
```diff
@@ -114,6 +148,16 @@ Individual [warning codes](https://github.com/mono/linker/blob/master/docs/error
 
     Don't treat ILLink warnings as errors. This may be useful to avoid turning trim analysis warnings into errors when treating compiler warnings as errors globally.
 
+## Showing detailed warnings
+
+In .NET 6+, trim analysis will produce at most one warning for each assembly that comes from a `PackageReference`, indicating that the assembly's internals are not compatible with trimming. You can also show individual warnings for all assemblies:
+
+- `<TrimmerSingleWarn>false</TrimmerSingleWarn>`
+
+    Show detailed warnings for all assemblies, instead of collapsing them to a single warning per assembly.
```
eerhardt, line 157:
```suggestion
    Show all detailed warnings, instead of collapsing them to a single warning per assembly.
```

docs/core/deploying/preparing-libraries-for-trimming.md
eerhardt: When we add this, we should have a "warning" or "advanced" banner on it. It is definitely a "this is something you can do to make things work, but it is really only for people who know precisely what they are doing".

docs/core/deploying/preparing-libraries-for-trimming.md
sbomer: I think it makes sense to have both a fixed announcement blog post and docs that we can expand over time. At the moment there's a lot of overlap, but I'd like to keep adding to these instructions so that they read more like a reference for authors who really have to dive into the annotations. @agocke thoughts?

I already see one discrepancy, which is that these docs don't yet mention UnconditionalSuppressMessage (I will add it). I'm also happy to take any other suggestions so that the docs contain a superset of the info in the blog post.

docs/core/deploying/trimming-options.md
sbomer: Clarified this

docs/core/deploying/trimming-options.md
sbomer: Clarified this in the section before the `TrimMode` bullets

docs/core/deploying/preparing-libraries-for-trimming.md
eerhardt: I'm fine with 2 posts, as long as the official docs (i.e. this document) is better than the blog. The case I don't want to have happen is the blog to be really good, but the official docs being worse.

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,188 @@
...
+
+If your app only uses parts of a library that are compatible with trimming, consider [enabling trimming](trimming-options.md#trim-additional-assemblies) of this library if it is not already being trimmed. This will only produce warnings if your app uses problematic parts of the library. (You can also [show detailed warnings](trimming-options.md#showing-detailed-warnings) for the library to see which parts of it are problematic.)
+
+## Enabling library trim warnings
+
+These instructions show how to enable and resolve static analysis warnings to prepare a library for trimming. Follow these steps if you are authoring a library and either want to proactively make your library trimmable, or have been contacted by app authors who encountered trim warnings from your library.
+
+Use the .NET 6 SDK for the best experience. It is possible to [show analysis warnings](trimming-options.md#analysis-warnings) in .NET 5, but not all framework assemblies were annotated correctly in .NET 5. Also, this will show detailed warnings for every library, including framework libraries. These instructions assume you are using the .NET 6 SDK.
```
vitek-karas, line 25: I would just state that users should use .NET 6 - and not go into the differences in behavior (there are way too many of them).

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Incomplete Roslyn analyzer
+
+During development, you may set `<EnableTrimAnalyzer>true</EnableTrimAnalyzer>` (in .NET 6+) in your library project to get a _limited_ set of warnings from the Roslyn analyzer. This analyzer is incomplete and should only be used as a convenience. It is important to follow the next steps to ensure that your library is compatible with trimming.
```
vitek-karas, line 29: This makes us look pretty bad - to soften this a bit I would add that we're working on improving the analyzer.

vitek-karas: In fact I would probably change this section. (Rewrite in real english please :wink:)

### Enable analysis across toolchain

You should set `<EnableTrimAnalyzer>true</EnableTrimAnalyzer>` in your library project. This will not have any effect on the output, but it will enable trim analysis across the entire toolchain.

It will enable the Roslyn analyzer for trimming, which is currently not completely, but it's useful for the cases where it works. It provides immediate response from within IDE or build. It doesn't not report all trim analysis warnings and doesn't cover all patterns though. The analyzer should improve over time to provide wider coverage of the trim incompatible patterns.

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,188 @@
...
+    {
+        foreach (var method in type.GetMethods())
+        {
+            // ...
+        }
+    }
+}
+```
+
+Here, `Foo` is calling a method which takes a `Type` argument that is annotated with a [`DynamicallyAccessedMembers`](https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.dynamicallyaccessedmembersattribute?view=net-5.0) requirement. The requirement states that the value passed in as this argument must represent a type whose public methods are available. In this case, you can fix this by adding the same requirement to the parameter of `Foo`.
```
vitek-karas, line 140: I would stay away from `Foo` - I know that at some point policheck wasn't happy about it. It also doesn't look "professional", or it looks "lazy". Basically just like we should not use `Foo` or similar simple name in tests - we should make the names descriptive of what they do.

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,188 @@
...
+    {
+        MyLibrary.type = typeof(System.Tuple);
+    }
+```
+
+In this case the trim analysis will simply keep public methods of `System.Tuple`, and will not produce further warnings.
+
+### Recommendations
+
+In general, try to avoid reflection if possible. When using reflection, limit it in scope so that it is reachable only from a small part of the library. For example, avoid using non-understood patterns in places like static constructors that will result in the warning propagating to all members of the class.
```
vitek-karas, line 184: I would probably also mention virtual/interface methods as bad places to annotate (as it spreads very wide).

vitek-karas (approved):

docs/core/deploying/preparing-libraries-for-trimming.md
agocke: I think this is good for long-term documentation. This looks more detailed, which I think is better for docs, and mine is more narrative, which may be useful for people just starting. I don't see any reason not to keep both up-to-date since the scenarios will be relevant long-term.

docs/core/deploying/preparing-libraries-for-trimming.md
sbomer: Reworded, hopefully this is close to what you had in mind.

docs/core/deploying/preparing-libraries-for-trimming.md
sbomer: Added a new section with UnconditionalSuppressMessage. I'm having some trouble coming up with examples for DynamicDependency that aren't super contrived. Do you have any suggestions?

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Showing all warnings
+
+To show all analysis warnings for your library, including warnings about dependencies, create a project like the following that references your library, and publish it with `PublishTrimmed`.
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
```
marek-safar, line 41: The description talks about library but output is exe

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,245 @@
...
+    <TargetFramework>net6.0</TargetFramework>
+    <!-- Use a RID of your choice. -->
+    <RuntimeIdentifier>linux-x64</RuntimeIdentifier>
+    <PublishTrimmed>true</PublishTrimmed>
+    <TrimmerDefaultAction>link</TrimmerDefaultAction>
```
marek-safar, line 46: Is this necessary to have it here? Maybe just mention it bellow

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,245 @@
...
+    <!-- Optional, for non-default workloads -->
```
marek-safar, line 47: I don't understand what non-default workloads means

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,245 @@
...
+    <SuppressTrimAnalysisWarnings>false</SuppressTrimAnalysisWarnings>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="path/to/MyLibrary.csproj" />
+    <TrimmerRootAssembly Include="MyLibrary" />
```
marek-safar, line 53: Isn't this the default behavior ?

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -22,11 +22,13 @@ If your app only uses parts of a library that are compatible with trimming, cons
...
 
-Use the .NET 6 SDK for the best experience. It is possible to [show analysis warnings](trimming-options.md#analysis-warnings) in .NET 5, but not all framework assemblies were annotated correctly in .NET 5. Also, this will show detailed warnings for every library, including framework libraries. These instructions assume you are using the .NET 6 SDK.
+Ensure you are using the .NET 6 SDK for these steps. They will not work correctly in previous versions.
 
-### Incomplete Roslyn analyzer
+## Enable Roslyn analyzer
 
-During development, you may set `<EnableTrimAnalyzer>true</EnableTrimAnalyzer>` (in .NET 6+) in your library project to get a _limited_ set of warnings from the Roslyn analyzer. This analyzer is incomplete and should only be used as a convenience. It is important to follow the next steps to ensure that your library is compatible with trimming.
+During development, set `<EnableTrimAnalyzer>true</EnableTrimAnalyzer>` (in .NET 6+) in your library project. This will not have any effect on the output, but it will enable trim analysis during build via a Roslyn analyzer.
```
vitek-karas, line 29: Why only during development?
If I want my library to be trim compatible - I should set this property... as simple as that.

docs/core/deploying/preparing-libraries-for-trimming.md
sbomer: Clarified by improving the description before the sample

docs/core/deploying/preparing-libraries-for-trimming.md
sbomer: It is, unless the library is trimmable. Added a comment inline to make it more clear

docs/core/deploying/preparing-libraries-for-trimming.md
sbomer: Fixed to describe what it does

docs/core/deploying/preparing-libraries-for-trimming.md
sbomer: I think it's pretty important - added a comment to the sample

docs/core/deploying/preparing-libraries-for-trimming.md
eerhardt: > Do you have any suggestions?

One example is when serialization doesn't work. 
* https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/src/libraries/System.Runtime.Serialization.Xml/tests/TrimmingTests/EndToEndTest.cs#L13-L15
* https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/src/libraries/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs#L177-L182

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,247 @@
...
+ms.date: 08/25/2020
+---
+
+# Preparing libraries for trimming
+
+## Trim warnings in apps
+
+When publishing an app, `PublishTrimmed` will produce trim analysis [warnings](trimming-options.md#analysis-warnings) (in .NET 6+) for patterns that are not statically understood to be compatible with trimming, including patterns in your code and in dependencies.
+
+You will encounter detailed warnings originating from your own code and `ProjectReference` dependencies. You may also see warnings like `warning IL2104: Assembly 'SomeAssembly' produced trim warnings` for `PackageReference` libraries. This means that the library contained patterns which are not guaranteed to work in the context of the trimmed app, and may result in a broken app. Consider contacting the author to see if the library can be annotated for trimming.
```
IEvangelist, line 15:
```suggestion
You will encounter detailed warnings originating from your own code and `ProjectReference` dependencies. You may also see warnings like `warning IL2104: Assembly 'SomeAssembly' produced trim warnings` for `PackageReference` libraries. This warning means that the library contained patterns which are not guaranteed to work in the context of the trimmed app, and may result in a broken app. Consider contacting the author to see if the library can be annotated for trimming.
```

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,247 @@
...
+
+To resolve warnings originating from the app code, jump ahead to the instructions on [resolving trim warnings](#resolving-trim-warnings). If you are interested in making your own `ProjectReference` libraries trim friendly, follow the instructions to [enable library trim warnings](#enabling-library-trim-warnings).
```
IEvangelist, line 17:
```suggestion
To resolve warnings originating from the app code, see [resolving trim warnings](#resolve-trim-warnings). If you are interested in making your own `ProjectReference` libraries trim friendly, follow the instructions to [enable library trim warnings](#enable-library-trim-warnings).
```

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,247 @@
...
+
+- `<SuppressTrimAnalysisWarnings>false</SuppressTrimAnalysisWarnings>` is implied by `PublishTrimmed` in the .NET SDK. It is only needed if the app targets form factors like Android where the trim analysis warnings are disabled by default. Consult the documentation for your SDK.
+
+You can also follow the same pattern for multiple libraries, adding them all to the same project as `ProjectReference` and `TrimmerRootAssembly` item to see trim analysis warnings for more than one library at a time, but note that this will warn about dependencies if _any_ of the root libraries use a trim-unfriendly API in a dependency. To see warnings that have to do with only a particular library, reference that library only.
+
+Note that the analysis results depend on the implementation details of your dependencies. If you update to a new version of a dependency, this may introduce analysis warnings if the new version added non-understood reflection patterns, even if there were no API changes. In other words, introducing trim analysis warnings to a library is a breaking change when the library is used with `PublishTrimmed`.
```
IEvangelist, line 75:
```suggestion
> [!NOTE]
> The analysis results depend on the implementation details of your dependencies. If you update to a new version of a dependency, this may introduce analysis warnings if the new version added non-understood reflection patterns, even if there were no API changes. In other words, introducing trim analysis warnings to a library is a breaking change when the library is used with `PublishTrimmed`.
```

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Resolving trim warnings
```
IEvangelist, line 77:
```suggestion
## Resolve trim warnings
```

IEvangelist (changes requested):
This is awesome, well done. I do have some suggestions I'd love to see incorporated into this to make it even better. Thanks

docs/core/deploying/preparing-libraries-for-trimming.md
```diff
@@ -0,0 +1,247 @@
+---
+title: Preparing libraries for trimming
+description: Learn how to prepare libraries for trimming.
+author: sbomer
+ms.author: svbomer
+ms.date: 08/25/2020
+---
+
+# Preparing libraries for trimming
+
```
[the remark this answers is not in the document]
sbomer: Thanks, added a short intro!

docs/core/deploying/prepare-libraries-for-trimming.md
```diff
@@ -0,0 +1,250 @@
...
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <!-- Use a RID of your choice. -->
+    <RuntimeIdentifier>linux-x64</RuntimeIdentifier>
+    <PublishTrimmed>true</PublishTrimmed>
+    <!-- Prevent warnings from unused code in dependencies -->
+    <TrimmerDefaultAction>link</TrimmerDefaultAction>
+    <!-- Enable warnings if the SDK you are using disables them -->
+    <SuppressTrimAnalysisWarnings>false</SuppressTrimAnalysisWarnings>
```
marek-safar, line 53: I'm wondering if this should be done automatically, anytime when someone has `<TrimmerDefaultAction>link</TrimmerDefaultAction>`

docs/core/deploying/prepare-libraries-for-trimming.md
```diff
@@ -0,0 +1,250 @@
...
+
+Ensure you are using the .NET 6 SDK for these steps. They will not work correctly in previous versions.
+
+## Enable Roslyn analyzer
+
+Set `<EnableTrimAnalyzer>true</EnableTrimAnalyzer>` (in .NET 6+) in your library project. This will not have any effect on the output, but it will enable trim analysis during build via a Roslyn analyzer.
+
+The Roslyn analyzer is useful for a fast feedback cycle with IDE integration, but is currently incomplete. It doesn't cover all trim analysis warnings, but the set of patterns it understands will improve over time to give more complete coverage. The Roslyn analyzer also isn't able to analyze the implementations of reference assemblies that you depend on. It is important to follow the next steps to ensure that your library is fully compatible with trimming.
+
+### Show all warnings
```
marek-safar, line 35: This section is not really only about showing all warnings but about trimming everything, right? Should this section cover how to do it for a library and not only for an app?

docs/core/deploying/prepare-libraries-for-trimming.md
```diff
@@ -0,0 +1,250 @@
...
+
+    [RequiresUnreferencedCode("DynamicBehavior is incompatible with trimming.")]
+    static void DynamicBehavior()
+    {
+    }
+}
+```
+
+This means the library calls a method which has explicitly been annotated as incompatible with trimming, using [`RequiresUnreferencedCodeAttribute`](
+https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.requiresunreferencedcodeattribute?view=net-5.0). To get rid of the warning, consider whether `Method` needs to call `DynamicBehavior` to do its job. If so, annotate the caller `Method` with `RequiresUnreferencedCode` as well; this will "bubble up" the warning so that callers of `Method` get a warning instead:
```
IEvangelist, line 107:
```suggestion
https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.requiresunreferencedcodeattribute?view=net-5.0&preserve-view=true). To get rid of the warning, consider whether `Method` needs to call `DynamicBehavior` to do its job. If so, annotate the caller `Method` with `RequiresUnreferencedCode` as well; this will "bubble up" the warning so that callers of `Method` get a warning instead:
```

docs/core/deploying/prepare-libraries-for-trimming.md
```diff
@@ -0,0 +1,250 @@
...
+        // The parameter 't' of method 'MyLibrary.UseMethods(Type)' does not have matching annotations.
+        foreach (var method in type.GetMethods())
+        {
+            // ...
+        }
+    }
+}
+```
+
+Here, `UseMethods` is calling a reflection method which has a [`DynamicallyAccessedMembers`](https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.dynamicallyaccessedmembersattribute?view=net-5.0) requirement. The requirement states that the type's public methods are available. In this case, you can fix this by adding the same requirement to the parameter of `UseMethods`.
```
IEvangelist, line 140:
```suggestion
Here, `UseMethods` is calling a reflection method which has a [`DynamicallyAccessedMembers`](https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.dynamicallyaccessedmembersattribute?view=net-5.0&preserve-view=true) requirement. The requirement states that the type's public methods are available. In this case, you can fix this by adding the same requirement to the parameter of `UseMethods`.
```

docs/core/deploying/prepare-libraries-for-trimming.md
```diff
@@ -0,0 +1,250 @@
...
+
+> [!WARNING]
+> These techniques might break your code if used incorrectly.
+
+When suppressing warnings, you are responsible for guaranteeing the trim compatibility of your code based on invariants that you know to be true by inspection. Be very careful with these annotations, because if they are incorrect, or if invariants of your code change, they might end up hiding real issues.
+
+### UnconditionalSuppressMessage
+
+If the intent of your code can't be expressed with the annotations, but you know that the warning doesn't represent a real issue at runtime, you can suppress the warnings using [`UnconditionalSuppressMessageAttribute`](https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.unconditionalsuppressmessageattribute?view=net-5.0). This is similar to `SuppressMessageAttribute`, but it is persisted in IL and respected during trim analysis. For example:
```
IEvangelist, line 203:
```suggestion
If the intent of your code can't be expressed with the annotations, but you know that the warning doesn't represent a real issue at runtime, you can suppress the warnings using [`UnconditionalSuppressMessageAttribute`](https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.unconditionalsuppressmessageattribute?view=net-5.0&preserve-view=true). This is similar to `SuppressMessageAttribute`, but it is persisted in IL and respected during trim analysis. For example:
```

IEvangelist (approved):
LGTM, we'll :shipit: if everyone else is good with this - thank you 🙏

docs/core/deploying/prepare-libraries-for-trimming.md
sbomer: It is about showing library warnings - I added a sentence to clarify why the step of creating an app project is necessary.

docs/core/deploying/prepare-libraries-for-trimming.md
sbomer: Removed, since this will be done automatically in https://github.com/dotnet/sdk/pull/16865.

docs/core/deploying/preparing-libraries-for-trimming.md
sbomer: Thanks! I will add an example for DynamicDependency in a follow-up.

gewarren:
Close/reopen to test status check.

Issue #23694, opened 2021-04-07
Examine reports to determine course of action


BillWagner:
There are two reports we'll use to determine actions for different areas of our content:

- [Content performance](https://aka.ms/contentperformancedashboard)
- [Low / No page views](https://aka.ms/docs-low-no-pv)

Possible actions to recommend may be:

- **Correcting metadata**: The default article type when content was migrated was "article" or "conceptual". That may not be accurate in all cases where it is used.
- **Archiving outdated content**: Some low page view content may need to be archived as it only applies to outdated versions.
- **Edit pass**: Some content may need a general update.

Top level folders to examine:

- [ ] Architecture
- [ ] Azure
- [ ] core
- [ ] csharp
- [ ] devops
- [ ] framework
- [ ] fsharp
- [ ] fundamentals
- [ ] iot
- [ ] standard
- [ ] visual-basic

tdykstra:
* docs/core folder: no articles in low/no PV report
  Updated ms.topic to reference for reference articles.

* docs/fundamentals folder: no articles in low/no PV report
  Updated ms.topic to reference for reference articles.

* docs/standard folder:  1 article in low/no PV report: https://docs.microsoft.com/en-us/dotnet/standard/data/xml/xslt-transformations-over-different-stores
 
  Recommend we archive all docs about the obsolete-since-Fx-2.0 XslTransform class: everything in TOC section headed by https://docs.microsoft.com/en-us/dotnet/standard/data/xml/xslt-transformations-with-the-xsltransform-class

BillWagner:
See #24002 for a proposed plan for the low PV articles in compiler messages.

tdykstra:
The low/no-PV report has 20 articles in the wcf folder aside from the wcf/diagnostics folder, which we're recommending for archival. Aside from diagnostics, I don't see any logical groupings that it makes sense to separate from the rest of the wcf documentation for archival. There are 11 in wcf/feature-details, but that's just a few out of a folder that has hundreds.  I'm thinking we should wait until we can archive the whole wcf folder, rather than pick and choose 20 individual docs.

tdykstra:
The low/no-PV report has 68 error reference articles in the framework/windows-workflow-foundation folder, and there would be many more if we raised the 5-PV limit for considering an article to be low-PV. But we decided against archiving error reference articles. We could archive with indexing enabled, but that would eliminate the main benefit of archiving that makes the effort involved in archiving worthwhile.

PR #23802, opened 2021-04-14
Update ms.topic for fundamentals folder

tdykstra:
Contributes to #23694

Makes reference the default topic type for subfolders under fundamentals, changes a few articles from conceptual to reference and one to how-to.


IEvangelist (approved):

PR #23809, opened 2021-04-15
C# reference: revised the switch expression article

pkulikov:
Moved focus from patterns to a `switch` expression itself.

### DIFF
--- a/docs/csharp/language-reference/operators/switch-expression.md
+++ b/docs/csharp/language-reference/operators/switch-expression.md
@@ -18,7 +18,7 @@ The following example demonstrates a `switch` expression, which converts values 
 
 The preceding example shows the basic elements of a `switch` expression:
 
-- The *input expression* that is the expression on the left-hand side of the `switch` keyword. At the preceding example, it's the `direction` method parameter.
+- An expression followed by the `switch` keyword. In the preceding example, it's the `direction` method parameter.
 - The *`switch` expression arms*, separated by commas. Each `switch` expression arm contains a *pattern*, an optional [*case guard*](#case-guards), the `=>` token, and an *expression*.
 
 At the preceding example, a `switch` expression uses the following patterns:
--- a/docs/csharp/language-reference/operators/switch-expression.md
+++ b/docs/csharp/language-reference/operators/switch-expression.md
@@ -1,54 +1,60 @@
 ---
 title: "switch expression - C# reference"
-description: Learn how to use the C# switch expression for pattern matching and other data introspection
-ms.date: 01/14/2021
+description: Learn about the C# switch expression that provides switch-like semantics based on pattern matching
+ms.date: 04/16/2021
+f1_keywords:
+  - "switch_CSharpKeyword"
+helpviewer_keywords:
+  - "switch expression [C#]"
+  - "pattern matching [C#]"
 ---
 # switch expression (C# reference)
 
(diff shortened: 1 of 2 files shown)


docs/csharp/language-reference/operators/switch-expression.md
```diff
@@ -1,54 +1,60 @@
...
-
-The `switch` expression provides for `switch`-like semantics in an expression context. It provides a concise syntax when the switch arms produce a value. The following example shows the structure of a switch expression. It translates values from an `enum` representing visual directions in an online map to the corresponding cardinal direction:
+The following example demonstrates a `switch` expression, which converts values of an [`enum`](../builtin-types/enum.md) representing visual directions in an online map to the corresponding cardinal directions:
 
 :::code language="csharp" source="snippets/shared/SwitchExpressions.cs" id="SnippetBasicStructure":::
 
-The preceding sample shows the basic elements of a switch expression:
+The preceding example shows the basic elements of a `switch` expression:
 
-- The *range expression*: The preceding example uses the variable `direction` as the range expression.
```
pkulikov, line 18: That name, "range expression", has been always confusing to me. Why?

My guess, that comes from the [spec proposal](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/patterns#switch-expression). Particularly, from the following spec:
```csharp
switch_expression
    : range_expression 'switch' '{' '}'
    | range_expression 'switch' '{' switch_expression_arms ','? '}'
    ;
```

However, in the text, the proposal doesn't use "range expression" term but the following (emphasis mine):
> At runtime, the result of the switch_expression is the value of the expression of the first switch_expression_arm for which the expression on **the left-hand-side of the switch_expression** matches 

I think `range_expression` comes from the [System.Range spec](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/ranges#systemrange) and is used to indicate the precedence of a `switch` expression. So, no need to use the same term in the docs. I think the "input expression" term is clearer.

docs/csharp/language-reference/operators/switch-expression.md
```diff
@@ -1,54 +1,60 @@
...
-- The *switch expression arms*: Each switch expression arm contains a *pattern*, an optional *case guard*, the `=>` token, and an *expression*.
+- The *input expression* that is the expression on the left-hand side of the `switch` keyword. At the preceding example, it's the `direction` method parameter.
```
BillWagner, line 21: 🎉 I get to reference the open source work on the standard.  The grammar for the [switch statement](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/statements.md#1383-the-switch-statement) uses the term "expression" instead of `range_expression` or `input expression`. I'm thinking the committee will likely follow that norm, so I'll make the following suggestion:

```suggestion
- An expression followed by the `switch` keyword. In the preceding example, it's the `direction` method parameter.
```

BillWagner (changes requested):
Overall, this looks great @pkulikov 

I had one suggestion, and then I'll :shipit:

docs/csharp/language-reference/operators/switch-expression.md
pkulikov: @BillWagner cool! Thank you!

BillWagner (approved):

PR #23811, opened 2021-04-15
Update ms.topic

tdykstra:
Contributes to #23694


IEvangelist (approved):

PR #23827, opened 2021-04-17
Replaced "solution folder" with "project folder"

ferenczy:
## Summary

Replaced "_solution folder_" with "_project folder_", because projects are placed in project folders, not solution folders, by default.

### DIFF
--- a/docs/core/tools/dotnet-sln.md
+++ b/docs/core/tools/dotnet-sln.md
@@ -94,7 +94,7 @@ dotnet sln add [-h|--help]
 
 - **`--in-root`**
 
-  Places the projects in the root of the solution, rather than creating a solution folder. Available since .NET Core 3.0 SDK.
+  Places the projects in the root of the solution, rather than creating a project folder. Available since .NET Core 3.0 SDK.
 
 - **`-s|--solution-folder <PATH>`**
 


docs/core/tools/dotnet-sln.md
```diff
@@ -94,7 +94,7 @@ dotnet sln add [-h|--help]
 
 - **`--in-root`**
 
-  Places the projects in the root of the solution, rather than creating a solution folder. Available since .NET Core 3.0 SDK.
+  Places the projects in the root of the solution, rather than creating a project folder. Available since .NET Core 3.0 SDK.
```
Youssef1313, line 97: I'm not sure whether this change is correct or not. But in case "project folder" is the correct term to use, it should probably get fixed in dotnet/sdk too.

https://github.com/dotnet/sdk/blob/88b1517c37c035df41532d22785a213d8243cd33/src/Cli/dotnet/commands/dotnet-sln/LocalizableStrings.resx#L165

ferenczy: I might be wrong. I'm not sure what that option is actually doing, I didn't find any difference when I have added a project to a solution file with and without that option.

It would probably deserve a better explanation here, so let's keep this PR open until someone who actually knows how it's supposed to work is able to review it and possibly explain.

tdykstra: "solution folder" is correct here -- see https://github.com/dotnet/sdk/issues/9971 and https://github.com/dotnet/cli/pull/10873 -- .

However, I also see that there seems to be no difference between including and omitting --in-root. Both are different from what you get with --solution-folder, which does put a project in a solution folder. 

@wli3 Is `--in-root` the default if `--solution-folder` is not used?

ferenczy: @tdykstra I see. In such a case, we can close this PR.

I would just ask you, could you explain to me the actual meaning of the `--in-root` option's description, please?

> Places the projects in the root of the solution, rather than creating a solution folder. Available since .NET Core 3.0 SDK.

It's talking about placing the projects and creating a solution folder, but when I'm executing the `dotnet sln add` command, a solution directory, a solution file and one or more projects (directories containing `csproj` files) already exist, otherwise the command fails.

Let's say, I have already created a solution and a single project using the dotnet CLI, so we have the following directory hierarchy - the `my-solution` directory containing the solution file `my-solution.sln` and the `my-project` directory with a project file `my-project.csproj`:

    ./my-solution/
    |
    + my-solution.sln
    +- my-project/
       |
       + my-project.csproj

Now, I'm going to add the project to the solution using the `dotnet sln add project/project.csproj` command.

Here are my questions:

First, what is meant by the "_projects_" in "_Places the projects in the root of the solution_"? The command `dotnet sln add --in-root <project-file>` just adds a reference of an existing project to an existing solution file, it's not actually creating anything, so what actually is it supposed to place in the solution's root?

Second, when NOT using the option `--in-root`, the `dotnet sln add <project-file>` should be "_creating a solution folder_" (instead of "_Places the projects in the root of the solution_"). What does the "_creating a solution folder_" mean? Or what's meant by the "_solution folder_"? For me, the "solution folder" is the `my-solution` directory. But it already exists, it already contains the solution file `my-solution.sln` and I'm executing the `dotnet` command from inside of that directory. If it wouldn't exist, I'm not able to add any project reference to it. So I don't understand, what "_solution folder_" the `dotnet sln add` command is supposed to create.

@tdykstra Since you know that the "_solution folder_" is correct in the description, you know the actual meaning, I guess? 🙂

tdykstra: >First, what is meant by the "projects" in "Places the projects in the root of the solution"?

This is referring to the references to projects in the solution file.

>Second, when NOT using the option --in-root, the dotnet sln add <project-file> should be "creating a solution folder" (instead of "Places the projects in the root of the solution"). What does the "creating a solution folder" mean? Or what's meant by the "solution folder"?

"Solution folder" refers to a virtual folder that appears in the Visual Studio Solution Explorer -- see https://docs.microsoft.com/en-us/visualstudio/ide/solutions-and-projects-in-visual-studio?view=vs-2019#solution-folder

>we can close this PR.

Closing the PR, but feel free to continue the discussion if needed. I'll open a PR to add a link to the Visual Studio documentation on solution folder.

Issue #16122, opened 2019-12-05
unnecessary addition of .exe


summershoe:
Needed to comment out line 59 of PipeClient to run the program. It appears to be adding an unnecessary '.exe' to the end of the file path?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b86675e7-4bfb-707d-fe36-c02da2023342
* Version Independent ID: ee821843-1cad-0455-f278-1194b88dd0b0
* Content: [How to: Use Named Pipes for Network Interprocess Communication](https://docs.microsoft.com/en-us/dotnet/standard/io/how-to-use-named-pipes-for-network-interprocess-communication#feedback)
* Content Source: [docs/standard/io/how-to-use-named-pipes-for-network-interprocess-communication.md](https://github.com/dotnet/docs/blob/master/docs/standard/io/how-to-use-named-pipes-for-network-interprocess-communication.md)
* Product: **dotnet**
* Technology: **dotnet-standard**
* GitHub Login: @mairaw
* Microsoft Alias: **mairaw**

gewarren:
`currentProcessName = Path.ChangeExtension(currentProcessName, ".exe");`

gewarren:
This is an easy fix. Reopening.

PR #23833, opened 2021-04-18
move samples and create vbproj and sln files

nemrism:
## Summary

As discussed previously, this PR aims to create a solution for the VB samples of the "How to: Use Named Pipes for Network Interprocess Communication" page

Contributes to #16122


BillWagner (approved):
Thanks @nemrism 

These look great, and I'll :shipit: now.

PR #23845, opened 2021-04-20
Move virtual keyword verbiage

### LINKED ISSUE #23732: Commentary mismatched with rule
Current:

> ❌ DISALLOWED: Removing the virtual keyword from a member
> 
> While this often is not a breaking change because the C# compiler tends to emit callvirt Intermediate Language (IL) instructions to call non-virtual methods (callvirt performs a null check, while a normal call doesn't), this behavior is not invariable for several reasons:
> 
> C# is not the only language that .NET targets.
> 
> The C# compiler increasingly tries to optimize callvirt to a normal call whenever the target method is non-virtual and is probably not null (such as a method accessed through the ?. null propagation operator).
> 
> Making a method virtual means that the consumer code would often end up calling it non-virtually.
> 
> ❌ DISALLOWED: Adding the virtual keyword to a member
> 

Suggested:

> ❌ DISALLOWED: Removing the virtual keyword from a member
> 
> ❌ DISALLOWED: Adding the virtual keyword to a member
> While this often is not a breaking change because the C# compiler tends to emit callvirt Intermediate Language (IL) instructions to call non-virtual methods (callvirt performs a null check, while a normal call doesn't), this behavior is not invariable for several reasons:
> 
> C# is not the only language that .NET targets.
> 
> The C# compiler increasingly tries to optimize callvirt to a normal call whenever the target method is non-virtual and is probably not null (such as a method accessed through the ?. null propagation operator).
> 
> Making a method virtual means that the consumer code would often end up calling it non-virtually.

---
#### Document Details

* GitHub Login: @gewarren

gewarren:
Fixes #23732

### DIFF
--- a/docs/core/compatibility/index.md
+++ b/docs/core/compatibility/index.md
@@ -145,14 +145,15 @@ Changes in this category modify the public surface area of a type. Most of the c
 
 - ❌ **DISALLOWED: Removing the [virtual](../../csharp/language-reference/keywords/virtual.md) keyword from a member**
 
+- ❌ **DISALLOWED: Adding the [virtual](../../csharp/language-reference/keywords/virtual.md) keyword to a member**
+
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

PR #23855, opened 2021-04-20
Add System.Half

### LINKED ISSUE #23030: Need to add Half to this page
https://docs.microsoft.com/en-us/dotnet/api/system.half?view=net-5.0

---
#### Document Details

* ID: 5796f0df-3d6f-30e1-ed1a-2169dc27dc38
* Version Independent ID: 56cab56d-07c8-885b-ac7a-b4e90e5409df
* Content: [.NET class library overview](https://docs.microsoft.com/en-us/dotnet/standard/class-library-overview#system-namespace)
* Content Source: [docs/standard/class-library-overview.md](https://github.com/dotnet/docs/blob/main/docs/standard/class-library-overview.md)
* GitHub Login: @gewarren

gewarren:
Fixes #23030. (Hide whitespace changes.)

See https://github.com/dotnet/runtime/pull/37630 and https://github.com/dotnet/runtime/issues/936.

[Internal preview](https://review.docs.microsoft.com/en-us/dotnet/standard/numerics?branch=pr-en-us-23855#floating-point-types).


docs/standard/numerics.md
```diff
@@ -14,63 +14,68 @@ ms.assetid: dfebc18e-acde-4510-9fa7-9a0f4aa3bd11
...
-|<xref:System.Decimal?displayProperty=nameWithType>|16|±1.0 x 10<sup>-28</sup> to ±7.9228 x 10<sup>28</sup>|28-29 digits|  
-  
-Both <xref:System.Single> and <xref:System.Double> types support special values that represent not-a-number and infinity. For example, the <xref:System.Double> type provides the following values: <xref:System.Double.NaN?displayProperty=nameWithType>, <xref:System.Double.NegativeInfinity?displayProperty=nameWithType>, and <xref:System.Double.PositiveInfinity?displayProperty=nameWithType>. You use the <xref:System.Double.IsNaN%2A?displayProperty=nameWithType>, <xref:System.Double.IsInfinity%2A?displayProperty=nameWithType>, <xref:System.Double.IsPositiveInfinity%2A?displayProperty=nameWithType>, and <xref:System.Double.IsNegativeInfinity%2A?displayProperty=nameWithType> methods to test for these special values.
+.NET includes the following primitive floating-point types:
+
+|Type|Size (in bytes)|Approximate range|Precision|
+|----------|--------|---------------------|--------------------|
+|<xref:System.Half?displayProperty=nameWithType>|2|||
```
gewarren, line 55: @pgovind Can you help me add the approximate range and precision values?

docs/standard/base-types/standard-numeric-format-strings.md
```diff
@@ -347,15 +347,15 @@ Formatting is influenced by the properties of the current <xref:System.Globaliza
```
gewarren, line 347: @pgovind Is there a special format specifier to round-trip a Half value like there is for Double and Single (see line 288)?

gewarren: Also, I noticed that Half has properties whereas Single and Double have fields. Is that intentional?

docs/standard/numerics.md
IEvangelist: I was going to suggest add the appropriate range too:

- [min](https://review.docs.microsoft.com/en-us/dotnet/api/system.half.minvalue?view=net-5.0&branch=main#System_Half_MinValue)
- [max](https://review.docs.microsoft.com/en-us/dotnet/api/system.half.maxvalue?view=net-5.0&branch=main#System_Half_MaxValue)

docs/standard/base-types/custom-numeric-format-strings.md
```diff
@@ -254,7 +254,7 @@ The following example uses both approaches to include reserved characters in a c
 
 ### Floating-Point infinities and NaN
 
-Regardless of the format string, if the value of a <xref:System.Single> or <xref:System.Double> floating-point type is positive infinity, negative infinity, or not a number (NaN), the formatted string is the value of the respective <xref:System.Globalization.NumberFormatInfo.PositiveInfinitySymbol%2A>, <xref:System.Globalization.NumberFormatInfo.NegativeInfinitySymbol%2A>, or <xref:System.Globalization.NumberFormatInfo.NaNSymbol%2A> property specified by the currently applicable <xref:System.Globalization.NumberFormatInfo> object.
+Regardless of the format string, if the value of a <xref:System.Half>,<xref:System.Single>, or <xref:System.Double> floating-point type is positive infinity, negative infinity, or not a number (NaN), the formatted string is the value of the respective <xref:System.Globalization.NumberFormatInfo.PositiveInfinitySymbol%2A>, <xref:System.Globalization.NumberFormatInfo.NegativeInfinitySymbol%2A>, or <xref:System.Globalization.NumberFormatInfo.NaNSymbol%2A> property specified by the currently applicable <xref:System.Globalization.NumberFormatInfo> object.
```
adegeo, line 257: missing space
```suggestion
Regardless of the format string, if the value of a <xref:System.Half>, <xref:System.Single>, or <xref:System.Double> floating-point type is positive infinity, negative infinity, or not a number (NaN), the formatted string is the value of the respective <xref:System.Globalization.NumberFormatInfo.PositiveInfinitySymbol%2A>, <xref:System.Globalization.NumberFormatInfo.NegativeInfinitySymbol%2A>, or <xref:System.Globalization.NumberFormatInfo.NaNSymbol%2A> property specified by the currently applicable <xref:System.Globalization.NumberFormatInfo> object.
```

docs/standard/language-independence.md
```diff
@@ -295,6 +295,7 @@ CLS-compliant type | Description
 [Int16](xref:System.Int16) | 16-bit signed integer
 [Int32](xref:System.Int32) | 32-bit signed integer
 [Int64](xref:System.Int64) | 64-bit signed integer
+[Half](xref:System.Half) | Half-precision floating-point value
```
adegeo, line 298: It may be worth pointing out what is said in the API docs:

>An IEEE 754 compliant `float16` type.

docs/standard/language-independence-and-language-independent-components.md
```diff
@@ -172,6 +172,7 @@ The .NET [common type system](base-types/common-type-system.md) includes a numbe
 |<xref:System.Int16>|16-bit signed integer|
 |<xref:System.Int32>|32-bit signed integer|
 |<xref:System.Int64>|64-bit signed integer|
+|<xref:System.Half>|Half-precision floating-point value|
```
adegeo, line 175: Same note here as before

docs/standard/base-types/standard-numeric-format-strings.md
tannergooding: This is intentional. `double` and `single` are runtime primitives and have special encoding in metadata for constants. `Half` is functionally a "user-defined" type and so cannot declare constants today.

docs/standard/class-library-overview.md
```diff
@@ -38,68 +38,69 @@ ms.assetid: 7e4c5921-955d-4b06-8709-101873acf157
...
+||<xref:System.UInt64>|A 64-bit unsigned integer.<br /><br /> Not CLS-compliant.|**ULong**|**ulong**|**unsigned __int64**|**uint64**|
+|Floating point|<xref:System.Half>|A half-precision (16-bit) floating-point number.|||||
+||<xref:System.Single>|A single-precision (32-bit) floating-point number.|**Single**|**float**|**float**|**float32**<br> or<br>**single**|
+||<xref:System.Double>|A double-precision (64-bit) floating-point number.|**Double**|**double**|**double**|**float**<br> or <br> **double**|
+|Logical|<xref:System.Boolean>|A Boolean value (true or false).|**Boolean**|**bool**|**bool**|**bool**|
+|Other|<xref:System.Char>|A Unicode (16-bit) character.|**Char**|**char**|**wchar_t**|**char**|
+||<xref:System.Decimal>|A decimal  (128-bit) value.|**Decimal**|**decimal**|**Decimal**|**decimal**|
+||<xref:System.IntPtr>|A signed integer whose size depends on the underlying platform (a 32-bit value on a 32-bit platform and a 64-bit value on a 64-bit platform).|**IntPtr**<br /><br /> No built-in type.|**IntPtr**<br /><br /> No built-in type.|**IntPtr**<br /><br /> No built-in type.|**unativeint**|
```
tannergooding, line 95: This is unrelated to the current PR, but `IntPtr` and `UIntPtr` now have built-in types on C#, these are `nint` and `nuint`, respectively.

docs/standard/class-library-overview.md
```diff
@@ -38,68 +38,69 @@ ms.assetid: 7e4c5921-955d-4b06-8709-101873acf157
...
+||<xref:System.SByte>|An 8-bit signed integer.<br /><br /> Not CLS-compliant.|**SByte**|**sbyte**|**char**<br /> -or-<br /> **signed** **char**|**sbyte**|
+||<xref:System.Int16>|A 16-bit signed integer.|**Short**|**short**|**short**|**int16**|
+||<xref:System.Int32>|A 32-bit signed integer.|**Integer**|**int**|**int**<br /><br /> -or-<br /><br /> **long**|**int**|
+||<xref:System.Int64>|A 64-bit signed integer.|**Long**|**long**|**__int64**|**int64**|
+||<xref:System.UInt16>|A 16-bit unsigned integer.<br /><br /> Not CLS-compliant.|**UShort**|**ushort**|**unsigned short**|**uint16**|
+||<xref:System.UInt32>|A 32-bit unsigned integer.<br /><br /> Not CLS-compliant.|**UInteger**|**uint**|**unsigned int**<br /> -or-<br /> **unsigned long**|**uint32**|
+||<xref:System.UInt64>|A 64-bit unsigned integer.<br /><br /> Not CLS-compliant.|**ULong**|**ulong**|**unsigned __int64**|**uint64**|
+|Floating point|<xref:System.Half>|A half-precision (16-bit) floating-point number.|||||
+||<xref:System.Single>|A single-precision (32-bit) floating-point number.|**Single**|**float**|**float**|**float32**<br> or<br>**single**|
```
tannergooding, line 90: Should `Half` be listed here as well? It isn't currently a language primitive but it is a "fundamental type" in that it is part of the ABI for various platforms and may become a language primitive in the future.

The same would be true if we added `Int128` and `UInt128` types in the future, for example.

docs/standard/language-independence-and-language-independent-components.md
tannergooding: The formal IEEE 754 name is `binary16` for `half`, `binary32` for single, and `binary64` for `double`.

docs/standard/numerics.md
```diff
@@ -14,63 +14,68 @@ ms.assetid: dfebc18e-acde-4510-9fa7-9a0f4aa3bd11
...
+|Type|Signed/Unsigned|Size (in bytes)|Minimum value|Maximum value|
+|----------|----------------------|--------------------|-------------------|-------------------|
+|<xref:System.Byte?displayProperty=nameWithType>|Unsigned|1|0|255|
+|<xref:System.Int16?displayProperty=nameWithType>|Signed|2|-32,768|32,767|
+|<xref:System.Int32?displayProperty=nameWithType>|Signed|4|-2,147,483,648|2,147,483,647|
+|<xref:System.Int64?displayProperty=nameWithType>|Signed|8|-9,223,372,036,854,775,808|9,223,372,036,854,775,807|
+|<xref:System.SByte?displayProperty=nameWithType>|Signed|1|-128|127|
+|<xref:System.UInt16?displayProperty=nameWithType>|Unsigned|2|0|65,535|
+|<xref:System.UInt32?displayProperty=nameWithType>|Unsigned|4|0|4,294,967,295|
+|<xref:System.UInt64?displayProperty=nameWithType>|Unsigned|8|0|18,446,744,073,709,551,615|
```
tannergooding, line 36: Not related to this PR, but should `IntPtr` (`nint`) and `UIntPtr` (`nuint`) be listed here as well?

They are fundamental integer types, but have different ranges when running on 32-bit vs 64-bit

docs/standard/numerics.md
```diff
@@ -14,63 +14,68 @@ ms.assetid: dfebc18e-acde-4510-9fa7-9a0f4aa3bd11
...
-|<xref:System.Decimal?displayProperty=nameWithType>|16|±1.0 x 10<sup>-28</sup> to ±7.9228 x 10<sup>28</sup>|28-29 digits|  
-  
-Both <xref:System.Single> and <xref:System.Double> types support special values that represent not-a-number and infinity. For example, the <xref:System.Double> type provides the following values: <xref:System.Double.NaN?displayProperty=nameWithType>, <xref:System.Double.NegativeInfinity?displayProperty=nameWithType>, and <xref:System.Double.PositiveInfinity?displayProperty=nameWithType>. You use the <xref:System.Double.IsNaN%2A?displayProperty=nameWithType>, <xref:System.Double.IsInfinity%2A?displayProperty=nameWithType>, <xref:System.Double.IsPositiveInfinity%2A?displayProperty=nameWithType>, and <xref:System.Double.IsNegativeInfinity%2A?displayProperty=nameWithType> methods to test for these special values.
+.NET includes the following primitive floating-point types:
+
+|Type|Size (in bytes)|Approximate range|Precision|Notes|
+|----------|--------|---------------------|--------------------|
+|<xref:System.Half?displayProperty=nameWithType>|2|||Introduced in .NET 5|
```
tannergooding, line 55: The exact range for `Half` is `+/-65504`.

The number of digits, mirroring what is used in the other columns is `~3-5 digits`. However, that column is very misleading in my opinion and may give users a false sense that any value with 6 significant digits can be exactly represented.

In practice, the lower limit here is that if a string with at most `n` significant digits (`3`, `6`, `15` for `half`, `float`, and `double`, respectively) is converted to a floating-point value and then back to a string with the same number of digits, the strings should match. That is, `float.Parse(s).ToString("G6") == s` should be `true` for any `s` that contains no more than `6` significant digits.
* This isn't guaranteed by the IEEE 754 spec and may depend on the formatting algorithm used

While the upper limit here is that if a floating-point value is converted to a string with that many digits (`5`, `9`, `17` for `half`, `float`, and `double`, respectively), and then converted back to a floating-point value, the result should match the original floating-point value. That is, `float.Parse(x.ToString("G9")).Equals(x)` should be `true` for any `float`.
* This is guaranteed by the IEEE 754 spec

docs/standard/language-independence-and-language-independent-components.md
adegeo: @tannergooding we should probably update the API docs, as it says float16: https://docs.microsoft.com/en-us/dotnet/api/system.half?view=net-5.0

docs/standard/base-types/standard-numeric-format-strings.md
pgovind: > Is there a special format specifier to round-trip a Half value like there is for Double and Single (see line 288)

`G5` for Half

pgovind (approved):
Other reviews have called out everything I noticed too. Approving assuming the comments will be resolved! Overall LGTM!

docs/standard/base-types/standard-numeric-format-strings.md
tannergooding: Since `Half` is only supported on .NET Core, `R` is the right thing to use here and will always do the correct thing.

tannergooding: We only need to list `G9` and `G17` for float/double because `R` had bugs on full framework.

docs/standard/numerics.md
pgovind: Called out elsewhere already. Just for complete-ness: The range is `+/-65504`

docs/standard/base-types/standard-numeric-format-strings.md
pgovind: Ah that explains why. I was wondering what cases `R` gave wrong results for.

docs/standard/numerics.md
gewarren: Thanks. For System.Single, this table shows the low end of the range as 1.5x10<sup>-45</sup>. But the [API ref docs show the MinValue](https://docs.microsoft.com/en-us/dotnet/api/system.single.minvalue?view=net-5.0) as -3.40282347E+38. Which one is correct?

pgovind: `MinValue` is `-3.40282347E+38`. However, there is some nuance here about normal and subnormal numbers. @tannergooding, we should decide what we want to docs to state the range as (applies to Double too)

tannergooding: One is giving the minimum value and the other is giving the smallest representable magnitude (that is the smallest +/- value which is not zero)

pgovind: Indeed, but the way the range is specified in the docs right now is misleading IMO. Does it make more sense here to just state the range as `+-3.40282347E+38`. The documentation for `Epsilon` at https://docs.microsoft.com/en-us/dotnet/api/system.single.epsilon?view=net-5.0 is quite good already.

tannergooding: > Indeed, but the way the range is specified in the docs right now is misleading IMO

Agreed, it can be confusing depending on how you read it and how familiar you are with IEEE 754 floating-point.

pgovind: Cool. I think we should just state the range here as `+-3.4E+38` for Single, `+-1.7E+308` for Double and `+-7.9E+28` for Decimal.

docs/standard/class-library-overview.md
gewarren: It's there already on line 89.

docs/standard/numerics.md
gewarren: I will plan to remove that column then. Thanks!

docs/standard/class-library-overview.md
tannergooding: Completely missed it 😅

gewarren:
Thanks for the great reviews!

PR #23884, opened 2021-04-22
Moved and revised the is operator article

pkulikov:
Removed the pattern matching details and added the links to the patterns article.
And, at last (!), moved the article to the operators section of the language reference.

PR #23920, opened 2021-04-24
Corrected runtime to run-time

pkulikov:
Thanks to @gewarren for the review of #23884 - I've checked all the articles in the operators folder and changed "runtime" to "run-time" where the adjective is used.


IEvangelist (approved):

PR #23927, opened 2021-04-26
Fix typo in date format name

mbikovitsky:
## Summary

This fixes a typo in the format name in the first paragraph: "ISO 8601:-2019" --> "ISO 8601-1:2019".


IEvangelist (approved):
Looks good to me, we'll :shipit: - thank you 🙏 - You are correct, in accordance with the official [ISO standard website](https://www.iso.org/standard/70907.html), this should be fixed.

PR #23941, opened 2021-04-26
C# operators: updated placeholder descriptions in metadata

pkulikov:
Mostly, update those descriptions that repeat an article title.

### DIFF
--- a/docs/csharp/language-reference/operators/null-coalescing-operator.md
+++ b/docs/csharp/language-reference/operators/null-coalescing-operator.md
@@ -1,6 +1,6 @@
 ---
 title: "?? and ??= operators - C# reference"
-description: "Learn about the C# null-coalescing operators ?? and ??="
+description: "Learn about ?? and ??= which are the C# null-coalescing operators."
 ms.date: 09/10/2019
 f1_keywords: 
   - "??_CSharpKeyword"
--- a/docs/csharp/language-reference/operators/null-coalescing-operator.md
+++ b/docs/csharp/language-reference/operators/null-coalescing-operator.md
@@ -1,6 +1,6 @@
 ---
-description: "?? and ??= operators - C# reference"
 title: "?? and ??= operators - C# reference"
+description: "Learn about the C# null-coalescing operators ?? and ??="
 ms.date: 09/10/2019
 f1_keywords: 
   - "??_CSharpKeyword"
--- a/docs/csharp/language-reference/operators/patterns.md
+++ b/docs/csharp/language-reference/operators/patterns.md
@@ -1,6 +1,6 @@
 ---
 title: "Patterns - C# reference"
-description: "Learn about the patterns supported by C# pattern matching expressions and statements - C# reference"
+description: "Learn about the patterns supported by C# pattern matching expressions and statements"
 ms.date: 04/05/2021
 f1_keywords: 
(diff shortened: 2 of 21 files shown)


docs/csharp/language-reference/operators/patterns.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: "Patterns - C# reference"
-description: "Learn about the patterns supported by C# pattern matching expressions and statements - C# reference"
+description: "Learn about the patterns supported by C# pattern matching expressions and statements"
```
Youssef1313, line 3: What are the guidelines about description? Should they end in a period or not?
Currently the PR uses both.

pkulikov: I don't know. I've noticed some have a period, others - not. I've decided to ignore that for now. However, if anyone knows what should be, I'll update the articles in this folder. @gewarren do you know by chance?

BillWagner: I looked this morning. The contributor guide doesn't say one way or the other. I think we've been adding the period, as they are typically complete sentences. (That's the standard we follow in bullet lists: The items should either all be sentences, or all not be sentences, and the punctuation should match).

docs/csharp/language-reference/operators/null-coalescing-operator.md
```diff
@@ -1,6 +1,6 @@
 ---
-description: "?? and ??= operators - C# reference"
 title: "?? and ??= operators - C# reference"
+description: "Learn about the C# null-coalescing operators ?? and ??="
```
BillWagner, line 3: This should add a period, but that looks awkward, so I rewrote the sentence a bit:

```suggestion
description: "Learn about ?? and ??= which are the C# null-coalescing operators."
```

docs/csharp/language-reference/operators/patterns.md
BillWagner, line 3:
```suggestion
description: "Learn about the patterns supported by C# pattern matching expressions and statements."
```

docs/csharp/language-reference/operators/true-false-operators.md
```diff
@@ -1,6 +1,6 @@
 ---
-description: "true and false operators - C# reference"
 title: "true and false operators - C# reference"
+description: "Learn about the C# true and false operators"
```
BillWagner, line 3: same:

```suggestion
description: "Learn about the C# true and false operators."
```

BillWagner (commented):
I think the sentences (with a period) are correct for descriptions.  

Commenting so this stays in my queue.

docs/csharp/language-reference/operators/patterns.md
pkulikov: Thank you @BillWagner I've applied your suggestions and also checked other operator articles to ensure descriptions end with a period.

gewarren: I don't think it really matters for search engine purposes, but I'd say it looks better with a period since it's a complete sentence. From the SEO cheat sheet:

image

BillWagner (commented):
Thanks again @pkulikov 

This is ready to :shipit:

BillWagner (approved):

PR #23949, opened 2021-04-27
Update master to main and .NET Core to .NET

Youssef1313:


IEvangelist (approved):

PR #23956, opened 2021-04-27
Use HYPHEN-MINUS instead of EN DASH in dotnet nuget command example

0xced:
Use `-` (U+002D HYPHEN-MINUS) instead of `–` (U+2013 EN DASH) in the `dotnet nuget locals all -l` example to be copy/paste friendly.


tdykstra (approved):
Thanks for the correction @0xced !

PR #23961, opened 2021-04-27
C# Metadata sweep

BillWagner:
Review article metadata and update where appropriate.

The main changes where that the XML Doc articles were all tagged as "conceptual". The entire folder was changed to "reference"
The entire folder for the introductory tutorials was added to docfx.json
Finally, a number of tutorials and how-tos were update in many folders. These weren't updated using folder based metadata, because there wasn't a great glob pattern to use.


docs/csharp/programming-guide/xmldoc/recommended-tags-for-documentation-comments.md
```diff
@@ -14,21 +15,82 @@ The C# compiler processes documentation comments in your code and formats them a
...
-|[\<include>](./include.md)*|[\<remarks>](./remarks.md)|[\<typeparamref>](./typeparamref.md)|  
-|[\<list>](./list.md)|[\<inheritdoc>](./inheritdoc.md)|[\<returns>](./returns.md)|
-  
+> Documentation comments cannot be applied to a namespace.
+
+The compiler will process any tag that is valid XML. The following tags provide generally used functionality in user documentation.
+
+## Tags
+
+:::row:::
```
tdykstra, line 24: I'm not sure this syntax is better than Markdown syntax, as the column and row positions are still effectively hard-coded and you have to do a lot of shuffling to add items and there's a lot more syntax text to wade through compared to pipes separating columns.  And when rendered, it spaces out the columns all the way across the page although that isn't needed for this content. 
It does eliminate the heading build suggestion which may become a warning eventually, so  it has that going for it. Looks like this one already needs some reshuffling, as it has 4 columns in the first row and 3 columns thereafter.

tdykstra (approved):
LGTM, a couple of questions/comments.

docs/csharp/programming-guide/xmldoc/recommended-tags-for-documentation-comments.md
BillWagner: Yeah, I find the syntax verbose. One feature I really like about it (in instances like this) is that it's not rendering as an HTML table, but as a CSS grid.  This way it adapts to screen width (phone, tablet, etc). I do really like that feature, especially when it's not a table, but rather a desire for multi-column output.

docfx.json
```diff
@@ -195,14 +195,16 @@
...
+        "docs/csharp/language-reference/compiler-messages/*.md": "error-reference",
         "docs/csharp/linq/*.md" : "how-to",
-        "docs/csharp/misc/**/*.md": "error-reference",
+        "docs/csharp/misc/*.md": "error-reference",
         "docs/csharp/programming-guide/**" : "conceptual",
+        "docs/csharp/programming-guide/xmldoc/**" : "reference",
         "docs/csharp/roslyn-sdk/get-started/*.md" : "tutorial",
         "docs/csharp/roslyn-sdk/tutorials/*.md" : "tutorial",
-        "docs/csharp/tour-of-csharp/**" : "overview",
+        "docs/csharp/tour-of-csharp/**" : "conceptual",
```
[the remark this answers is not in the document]
BillWagner: Good point. I updated this.

PR #23968, opened 2021-04-28
add misc compiler errors and warnings

BillWagner:
Update the message TOC, update the breadcrumbs file

Found while looking at the low PV report. Most of them are in the language reference / compiler messages area.

[Internal review site](https://review.docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/?branch=pr-en-us-23968)

### DIFF
--- a/docs/csharp/language-reference/compiler-messages/toc.yml
+++ b/docs/csharp/language-reference/compiler-messages/toc.yml
@@ -1879,7 +1879,7 @@ items:
     href: cs1610.md
   - name: CS1712
     href: ../../misc/cs1712.md
--name: Level 5 warning messages
+- name: Level 5 warning messages
   items:
   - name: CS8892
     href: cs8892.md
--- a/docs/csharp/language-reference/compiler-messages/toc.yml
+++ b/docs/csharp/language-reference/compiler-messages/toc.yml
@@ -3,218 +3,1550 @@
   items:
   - name: Compiler error CS0001
     href: cs0001.md
+  - name: Compiler Error CS0003
+    href: ../../misc/cs0003.md
+  - name: Compiler Error CS0004
+    href: ../../misc/cs0004.md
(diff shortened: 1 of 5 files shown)


docs/csharp/language-reference/compiler-messages/toc.yml
```diff
@@ -3,218 +3,1550 @@
   items:
   - name: Compiler error CS0001
     href: cs0001.md
+  - name: Compiler Error CS0003
```
gewarren, line 6:
```suggestion
  - name: Compiler error CS0003
```

docs/csharp/language-reference/compiler-messages/toc.yml
```diff
@@ -3,218 +3,1550 @@
   items:
```
gewarren, line 3: Maybe have two top level nodes: Errors and Warnings. Then nest the individual messages underneath, omitting either "Compiler error" or "Compiler warning" from the displayed text. As is, it's one very long list, and a lot of repeated words.

BillWagner: Great suggestion!

docs/csharp/language-reference/compiler-messages/toc.yml
BillWagner: replaced in all locations (and warnings)

BillWagner:
Thanks @gewarren for the suggestion.  It's a big improvement.

This should be ready for final review.

gewarren (approved):
Looks much better! If you want to take it a step further, https://review.docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/?branch=pr-en-us-23968 should also be updated:

- The title should be "C# compiler messages"
- It's very VS focused and assumes you're using VS
- It should recommend putting the compiler message number in the TOC filter to find the topic

PR #23982, opened 2021-04-29
Simplify example

### LINKED ISSUE #23981: improve nullability code sample
the doc sample code is currently
public static bool IsValid(Person? person)
{
    return person != null && !string.IsNullOrEmpty(person.Name);
}
but should [simplify] be
public static bool IsValid(Person? person)
{
    return !string.IsNullOrEmpty(person?.Name);
}
or even [better]
public static bool IsValid(Person? person) => !string.IsNullOrEmpty(person?.Name);

---
#### Document Details

* ID: 0471f98c-f5cd-00d2-a705-396682f73bcb
* Version Independent ID: a24b862d-c462-a040-8470-9771c74c4e6f
* Content: [! (null-forgiving) operator - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving)
* Content Source: [docs/csharp/language-reference/operators/null-forgiving.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/operators/null-forgiving.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

Youssef1313:
Fixes https://github.com/dotnet/docs/issues/23981

### DIFF
--- a/docs/csharp/language-reference/operators/snippets/shared/NullForgivingOperator.cs
+++ b/docs/csharp/language-reference/operators/snippets/shared/NullForgivingOperator.cs
@@ -39,7 +39,7 @@ namespace operators
         }
 
         public static bool IsValid(Person? person)
-            => !string.IsNullOrEmpty(person?.Name);
+            => person is not null && person.Name is not null;
         // </SnippetUseNullForgiving>
 
         public static Person? Find(string name) => null;
@@ -58,7 +58,7 @@ namespace operators
         }
 
         public static bool IsValid([NotNullWhen(true)] Person? person)
-            => !string.IsNullOrEmpty(person?.Name);
+            => person is not null && person.Name is not null;
         // </SnippetUseAttribute>
 
         public static Person? Find(string name) => null;
--- a/docs/csharp/language-reference/operators/snippets/shared/NullForgivingOperator.cs
+++ b/docs/csharp/language-reference/operators/snippets/shared/NullForgivingOperator.cs
@@ -39,9 +39,7 @@ namespace operators
         }
 
         public static bool IsValid(Person? person)
-        {
-            return person != null && !string.IsNullOrEmpty(person.Name);
-        }
+            => !string.IsNullOrEmpty(person?.Name);
         // </SnippetUseNullForgiving>
 
         public static Person? Find(string name) => null;
@@ -60,9 +58,7 @@ namespace operators
         }
 
         public static bool IsValid([NotNullWhen(true)] Person? person)
-        {
-            return person != null && !string.IsNullOrEmpty(person.Name);
-        }
+            => !string.IsNullOrEmpty(person?.Name);
         // </SnippetUseAttribute>
 
         public static Person? Find(string name) => null;


pkulikov:
It's a good refactoring in general. However, I'm afraid it may make understanding the containing article more difficult. See the article for the context:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving

Now, it's clear from the first look that if the `IsValid` method returns true, its argument is non-null. With the introduced change, a reader should spend more time to reason about the example. Not sure if that's necessary.

docs/csharp/language-reference/operators/snippets/shared/NullForgivingOperator.cs
```diff
@@ -39,9 +39,7 @@ public static void Main()
         }
 
         public static bool IsValid(Person? person)
-        {
-            return person != null && !string.IsNullOrEmpty(person.Name);
-        }
+            => !string.IsNullOrEmpty(person?.Name);
```
IEvangelist, line 42:
```suggestion
            => person is { Name: { Length: > 0 } };

```

docs/csharp/language-reference/operators/snippets/shared/NullForgivingOperator.cs
```diff
@@ -60,9 +58,7 @@ public static void Main()
         }
 
         public static bool IsValid([NotNullWhen(true)] Person? person)
-        {
-            return person != null && !string.IsNullOrEmpty(person.Name);
-        }
+            => !string.IsNullOrEmpty(person?.Name);
```
IEvangelist, line 61:
```suggestion
            => person is { Name: { Length: > 0 } };
```

docs/csharp/language-reference/operators/snippets/shared/NullForgivingOperator.cs
BillWagner, line 42: I'm curious about this change instead:

```csharp
public static bool IsValid(Person? person) => person is {​​​​​​​ Name: {​​​​​​​ Length: > 0 }​​​​​​​ }​​​​​​​;
```

Thoughts?

pkulikov: It's a good example of pattern matching that in the context of this article may be a bit overwhelming (like I noticed above)

pkulikov: What about `person is not null && person.Name is not null`
It's not refactoring, though, but would fit the article

BillWagner: I like that version from @pkulikov It is semantically different, but does highlight the concepts for the article better than the existing code, or the other suggestions.

Youssef1313: > What about `person is not null && person.Name is not null`

Looks good. Actually, even `person is not null` would be enough for the article. But I went for your suggestion.

BillWagner (approved):
Thanks for updating this, and for the discussion @Youssef1313 @pkulikov @IEvangelist 

This is ready to merge.

PR #24013, opened 2021-05-02
Update metrics.md

SebastianKirsch123:
Expanded the section on R2 to explain why R2 can be negative because mathematically inclined readers will recognize that a negative square is a mathematical impossibility. Added a reference with more in-depth explanations in the list.

## Summary

Describe your changes here.


luisquintanilla (approved):
Looks great. Thanks for this contribution @SebastianKirsch123

PR #24026, opened 2021-05-04
Removed unneeded comma

hjerpbakk:
## Summary

Removed unneeded comma from code example.


Youssef1313 (commented):
For the context of the article, it doesn't matter much.

But generally I like including it, so in the future, adding new properties won't require modifying the last existing property, resulting in a cleaner diff.

tdykstra (approved):
@Sankra Thanks for making this change. Although the code compiles with the comma, I too prefer to omit it when it's not needed.

PR #24036, opened 2021-05-04
Update branches-and-loops.yml

echokrist:
Proposed grammatical change in the else conditional Console.WriteLine string for the && operator.

## Summary

Change the string in the Console.Writeline for the && operator instruction in order to use the word "And" instead of "Or". This ensures that there are no confusions as to how the forementioned operator work.

### DIFF
--- a/docs/csharp/tour-of-csharp/tutorials/branches-and-loops.yml
+++ b/docs/csharp/tour-of-csharp/tutorials/branches-and-loops.yml
@@ -113,7 +113,7 @@ items:
     else
     {
         Console.WriteLine("The answer is not greater than 10");
-        Console.WriteLine("Or the first number is not equal to the second");
+        Console.WriteLine("And the first number is not equal to the second");
     }
     ```
 


docs/csharp/tour-of-csharp/tutorials/branches-and-loops.yml
```diff
@@ -113,7 +113,7 @@ items:
     else
     {
         Console.WriteLine("The answer is not greater than 10");
-        Console.WriteLine("Or the first number is not equal to the second");
+        Console.WriteLine("And the first number is not equal to the second");
```
Youssef1313, line 116: The original statement is correct. In the `else` branch, it means that the `if` condition evaluates to false. When you have `A && B` and it evaluates to false, then either `A` or `B` or both evaluates to false. (This originates from De Morgan's law)

KNA0NE: Thank you for the feedback. You're right, the changes proposed will be revoked.

PR #24040, opened 2021-05-04
Add details about file system polling option as an alert

### LINKED ISSUE #23931: Add file watcher details with Docker support
Add details to existing .NET docker articles, and options pattern with .NET articles that describes watching the file system.

See https://github.com/dotnet/runtime/issues/51866#issuecomment-826936347, and [ASP.NET Core docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/file-providers?view=aspnetcore-5.0#watch-for-changes).

IEvangelist:
## Summary

Add details about file system polling option as an alert.

Fixes #23931


BillWagner (approved):
This LGTM @IEvangelist 

You can :shipit: when ready.

PR #24051, opened 2021-05-05
Update channel-credentials.md

stebet:
## Summary

Adding details for a workaround needed for PEM loaded client certificates on Windows.

### DIFF
--- a/docs/architecture/grpc-for-wcf-developers/channel-credentials.md
+++ b/docs/architecture/grpc-for-wcf-developers/channel-credentials.md
@@ -100,7 +100,7 @@ class Program
         // Assume path to a client .pfx file and password are passed from command line
         // On Windows this would probably be a reference to the Certificate Store
         var cert = new X509Certificate2(args[0], args[1]);
-
+        
         var handler = new HttpClientHandler();
         handler.ClientCertificates.Add(cert);
         var httpClient = new HttpClient(handler);
@@ -116,6 +116,17 @@ class Program
     }
 }
 ```
+
+> [!NOTE]
+> Due to an internal Windows bug as [documented here](https://github.com/dotnet/runtime/issues/23749#issuecomment-388231655) you will need to apply the following a workaround if the certificate is created from certificate and private key PEM data.
+> ```csharp
+> X509Certificate2 cert = X509Certificate2.CreateFromPem(certificatePem, rsaPrivateKeyPem));
+> if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+> {
+>     var originalCert = cert;
+>     cert = new X509Certificate2(cert.Export(X509ContentType.Pkcs12));
+>     originalCert.Dispose();
+> }
 
 ## Combine ChannelCredentials and CallCredentials
 
--- a/docs/architecture/grpc-for-wcf-developers/channel-credentials.md
+++ b/docs/architecture/grpc-for-wcf-developers/channel-credentials.md
@@ -100,7 +100,7 @@ class Program
         // Assume path to a client .pfx file and password are passed from command line
         // On Windows this would probably be a reference to the Certificate Store
         var cert = new X509Certificate2(args[0], args[1]);
-        
+
         var handler = new HttpClientHandler();
         handler.ClientCertificates.Add(cert);
         var httpClient = new HttpClient(handler);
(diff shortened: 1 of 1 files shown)


adegeo:
@stebet Hi! Thank you for the submission. Did you test this code? I spotted a code error in your submission. After reading the thread I'm not sure this workaround applies. The thread states that when the certificate is imported with in-memory key, it fails. Is the code prior to your submission doing that? It looks like it's using files passed on program args.

stebet:
> @stebet Hi! Thank you for the submission. Did you test this code? I spotted a code error in your submission. After reading the thread I'm not sure this workaround applies. The thread states that when the certificate is imported with in-memory key, it fails. Is the code prior to your submission doing that? It looks like it's using files passed on program args.

That's why I mentioned specifically in the note that if you are loading the client cert via PEM strings instead of files (PFX + password), you'll need to apply this workaround? The example above the note show only one of several ways to use a client certificate and after hitting the issue myself after having followed this method (despite loading the cert differently which is just one line in the code) and scratching my head for several trying to figure out why this didn't work, I felt a note here might be helpful.

Maybe adding another example that shows the PEM way + workaround as an alternative instead of the note would be clearer?

This is in my opinion pretty important info for anyone using client certificate authentication in gRPC to be aware of.

stebet:
Maybe @blowdart or @JamesNK have some ideas/opinions here?

docs/architecture/grpc-for-wcf-developers/channel-credentials.md
```diff
@@ -117,6 +117,19 @@ class Program
 }
 ```
 
+> [!NOTE]
+> Due to an internal Windows bug as [documented here](https://github.com/dotnet/runtime/issues/23749#issuecomment-388231655), you'll need to apply the following a workaround if the certificate is created from certificate and private key PEM data.
+> 
+> ```csharp
+> X509Certificate2 cert = X509Certificate2.CreateFromPem(certificatePem, rsaPrivateKeyPem));
```
adegeo, line 124: fix code error
```suggestion
> X509Certificate2 cert = X509Certificate2.CreateFromPem(certificatePem, rsaPrivateKeyPem);
```

adegeo:
@stebet sorry for the delay. I'm just curious if the note intro should be modified to explain exactly why this code would be needed. It sounds like it's always needed. (I'm not super familiar with cert management though) The intro doesn't specify it's in-memory keys such as a string etc..

stebet:
> @stebet sorry for the delay. I'm just curious if the note intro should be modified to explain exactly why this code would be needed. It sounds like it's always needed. (I'm not super familiar with cert management though) The intro doesn't specify it's in-memory keys such as a string etc..

No need to be sorry :)

Changing the intro a bit makes sense. I'll reword and make a few changes to explain it better and why it's needed.

stebet:
Got delayed a bit with other tasks. I'll take a look at this after the weekend.

stebet:
@adegeo I added another example section that hopefully clarifies the example and note a bit better.

docs/architecture/grpc-for-wcf-developers/channel-credentials.md
```diff
@@ -92,6 +92,7 @@ public class Startup
 
 With the `Grpc.Net.Client` package, you configure certificates on an <xref:System.Net.Http.HttpClient> instance that is provided to the `GrpcChannel` used for the connection.
 
+### Loading a client certificate from a .PFX file
```
IEvangelist, line 95:
```suggestion
### Load a client certificate from a .PFX file

```

docs/architecture/grpc-for-wcf-developers/channel-credentials.md
```diff
@@ -117,6 +118,45 @@ class Program
 }
 ```
 
+### Loading a client certificate from certificate and private key .PEM files
```
IEvangelist, line 121:
```suggestion
### Load a client certificate from certificate and private key .PEM files

```

docs/architecture/grpc-for-wcf-developers/channel-credentials.md
```diff
@@ -117,6 +118,45 @@ class Program
...
+    static async Task Main(string[] args)
+    {
+        // Assume path to a certificate and private key .pem files are passed from command line
+        string certificatePem = File.ReadAllText(args[0]);
+        string privateKeyPem = File.ReadAllText(args[1]);
+        var cert = X509Certificate2.CreateFromPem(certificatePem, privateKeyPem);
+
+        var handler = new HttpClientHandler();
+        handler.ClientCertificates.Add(cert);
+        var httpClient = new HttpClient(handler);
```
IEvangelist, line 134:
```suggestion
        using HttpClient httpClient = new(handler);
```

adegeo (approved):
Looks good! Thank you for being so patient!

PR #24056, opened 2021-05-05
Add note about default implementations

### LINKED ISSUE #24019: Is adding a member to an interface allowed if it has a default implementation?
> DISALLOWED: Adding a member to an interface

The purpose of the "default implementations in interfaces" feature was to allow for adding new members to existing interfaces, right? If so, I'm curious whether this rule should be changed/restated to account for that feature.

Even if there is a reason why that can't work, given the feature's existence it seems worth having some commentary in this document explaining why this remains a breaking change.

---
#### Document Details

* GitHub Login: @gewarren

gewarren:
Fixes #24019

### DIFF
--- a/docs/core/compatibility/index.md
+++ b/docs/core/compatibility/index.md
@@ -121,6 +121,8 @@ Changes in this category modify the public surface area of a type. Most of the c
 
 - ❌ **DISALLOWED: Adding a member to an interface**
 
+  Unless you provide a define an implementation, adding a new member to an interface is a breaking change. For more information about default implementations, see [Update interfaces with default interface methods in C# 8.0](../../csharp/whats-new/tutorials/default-interface-methods-versions.md).
+
 - ❌ **DISALLOWED: Changing the value of a public constant or enumeration member**
 
 - ❌ **DISALLOWED: Changing the type of a property, field, parameter, or return value**
--- a/docs/core/compatibility/index.md
+++ b/docs/core/compatibility/index.md
@@ -119,7 +119,7 @@ Changes in this category modify the public surface area of a type. Most of the c
 
    This includes removing or renaming a getter or setter from a property, as well as renaming or removing enumeration members.
 
-- ❌ **DISALLOWED: Adding a member to an interface**
+- ❓ **REQUIRES JUDGMENT: Adding a member to an interface**
 
   Unless you provide an implementation, adding a new member to an interface is a breaking change. For more information about default implementations, see [Update interfaces with default interface methods in C# 8.0](../../csharp/whats-new/tutorials/default-interface-methods-versions.md).
 
--- a/docs/core/compatibility/index.md
+++ b/docs/core/compatibility/index.md
@@ -121,7 +121,7 @@ Changes in this category modify the public surface area of a type. Most of the c
 
 - ❌ **DISALLOWED: Adding a member to an interface**
 
-  Unless you provide a define an implementation, adding a new member to an interface is a breaking change. For more information about default implementations, see [Update interfaces with default interface methods in C# 8.0](../../csharp/whats-new/tutorials/default-interface-methods-versions.md).
+  Unless you provide an implementation, adding a new member to an interface is a breaking change. For more information about default implementations, see [Update interfaces with default interface methods in C# 8.0](../../csharp/whats-new/tutorials/default-interface-methods-versions.md).
 
 - ❌ **DISALLOWED: Changing the value of a public constant or enumeration member**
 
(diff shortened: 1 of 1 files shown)


gewarren:
@BillWagner Should we have someone from the product team review this? If so, who would you recommend?

docs/core/compatibility/index.md
```diff
@@ -121,6 +121,8 @@ Changes in this category modify the public surface area of a type. Most of the c
 
 - ❌ **DISALLOWED: Adding a member to an interface**
 
+  Unless you provide a define an implementation, adding a new member to an interface is a breaking change. For more information about default implementations, see [Update interfaces with default interface methods in C# 8.0](../../csharp/whats-new/tutorials/default-interface-methods-versions.md).
```
IEvangelist, line 124:
```suggestion
  Unless you provide an implementation, adding a new member to an interface is a breaking change. For more information about default implementations, see [Update interfaces with default interface methods in C# 8.0](../../csharp/whats-new/tutorials/default-interface-methods-versions.md).
```

IEvangelist (approved):

docs/core/compatibility/index.md
```diff
@@ -121,6 +121,8 @@ Changes in this category modify the public surface area of a type. Most of the c
 
 - ❌ **DISALLOWED: Adding a member to an interface**
 
+  Unless you provide an implementation, adding a new member to an interface is a breaking change. For more information about default implementations, see [Update interfaces with default interface methods in C# 8.0](../../csharp/whats-new/tutorials/default-interface-methods-versions.md).
```
gewarren, line 124: I'm also leaning towards changing it from DISALLOWED to REQUIRES JUDGMENT.

IEvangelist: How about "DO" and "AVIOD", do being the recommended course of action and avoid being something that you'd try to avoid doing but using your better judgment.

gewarren: REQUIRES JUDGMENT is already in use in this doc. I don't think we should introduce new terms at this point.

IEvangelist: Perhaps replace the existing terms? I thought we had other docs that used the DO and AVOID terminology.

docs/core/compatibility/index.md
```diff
@@ -121,6 +121,8 @@ Changes in this category modify the public surface area of a type. Most of the c
 
 - ❌ **DISALLOWED: Adding a member to an interface**
```
gewarren, line 122:
```suggestion
- ❓ **REQUIRES JUDGMENT: Adding a member to an interface**
```

docs/core/compatibility/index.md
gewarren: It looks like the [design guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/) use that terminology.

gewarren:
@terrajobst or @cecilphillip Could you review/approve this change in the breaking change guidelines?

docs/core/compatibility/index.md
terrajobst: @bartonjs can comment on the official wording we use in the API design guidelines.

Long story short, adding a default interface member (DIM) is a breaking change:

1. Not all languages support DIMs
2. There can still be cases where consumers of an interface can end up in a shared diamond where the runtime can't decide which default interface member to invoke.

bartonjs: I get to quote my own annotation from the book.  I feel famous or something.

> JEREMY BARTON
> C# 8.0 added a feature, Default Interface Methods,
which allows creating new methods on an interface without necessarily
resulting in compile failures in downstream assemblies. While this feature
can be useful for adding a simplified overload for an existing method, it
usually can’t introduce a new concept without a compile-time failure or a
runtime exception. Because new concepts can’t be introduced to an interface
and Just Work, we still consider the interface to be effectively fixed once
it ships.
There isn’t any guidance for Default Interface Methods in the third edition
because we haven’t yet learned where things really go wrong. The best
proto-guideline we have is “DO NOT use Default Interface Methods to provide
an implementation for a member from a base interface,” so as to avoid
the “diamond problem.”

If the question is less about DIMs and more about our guidance words, the spectrum is `DO, CONSIDER, AVOID, DO NOT`.  In the core libraries our answer right now is "DO NOT" because we don't have the experience for when to say otherwise.  (Every time it comes up we have a reason why adding a DIM won't work, so we've still never added one)

gewarren: Thank you @terrajobst and @bartonjs for the prompt replies!

@IEvangelist I'm not going to change the wording in this PR for a couple reasons: 1) I'm not sure when to map "requires judgment" to "consider" and when to map it to "avoid", and 2) If we change "allowed" to "do", we end up with statements like "✔️ DO: Move a type from one assembly to another". Which makes it sound like we are encouraging you to move types? Anyway, feel free to create a new PR or issue if you feel strongly about this.

PR #24088, opened 2021-05-07
Fix note for passing properties to dotnet build

Youssef1313:
Using `/p:....` will be interpreted as if it's a project file name and will show an error. `-p` worked correctly for me.


IEvangelist (approved):

PR #24114, opened 2021-05-10
Updated GetAll to work with example code

maxs-rose:
## Summary

With the example implementation of CardType GetAll now works as it will search for Properties and not Fields


adegeo (approved):
#23726 broke this code. Does anyone know why they were originally implemented as fields and not properties? That seems deliberate. Otherwise, it seems this PR is good. The sample code though needs to be fixed to match https://github.com/dotnet-architecture/eShopOnContainers/blob/dev/src/Services/Ordering/Ordering.Domain/AggregatesModel/BuyerAggregate/CardType.cs

IEvangelist:
This was fixed in a different way, elsewhere. Closing this, see https://github.com/dotnet/docs/pull/24236/

PR #24208, opened 2021-05-12
Documented TimeSpan marshalling

Const-me:
Tested in .NET 5 on Win64. Not sure about other platforms, please specify.

### DIFF
--- a/docs/standard/native-interop/type-marshaling.md
+++ b/docs/standard/native-interop/type-marshaling.md
@@ -42,6 +42,7 @@ This first table describes the mappings for various types for whom the marshalin
 | `decimal` | COM `DECIMAL` struct |
 | .NET Delegate | Native function pointer |
 | `System.DateTime` | Win32 `DATE` type |
+| `System.TimeSpan` | `int64_t` |
 | `System.Guid` | Win32 `GUID` type |
 
 A few categories of marshaling have different defaults if you're marshaling as a parameter or structure.


jkoritzinsky:
The intent of the `type-marshalling.md` doc is to only include types that have specially-handled well-defined marshalling rules in the .NET Runtime.

TimeSpan is not one of these types. It ends up marshalling as a struct that contains an int64_t since that is the equivalent native layout to its managed layout, but that is not guaranteed by the runtime and could change in the future.

Const-me:
> It ends up marshalling as a struct that contains an int64_t since that is the equivalent native layout to its managed layout, but that is not guaranteed by the runtime and could change in the future.

How likely it is to change? Have you seen this comment? https://source.dot.net/#System.Private.CoreLib/TimeSpan.cs,55 Has it ever changed since .NET framework 1.0 released in 2002?

Anyway, if you really want to keep it unspecified (e.g. planning to change RAM layout in the near future), then what's the recommended way to pass time intervals to native code? That structure is available since .NET Framework 1.0, and used all over the software. I know how to call `TimeSpan.FromTicks` but that's not the best usability, and in some cases can be slower.

jkoritzinsky:
> > It ends up marshalling as a struct that contains an int64_t since that is the equivalent native layout to its managed layout, but that is not guaranteed by the runtime and could change in the future.
> 
> How likely it is to change? Have you seen this comment? [source.dot.net/#System.Private.CoreLib/TimeSpan.cs,55](https://source.dot.net/#System.Private.CoreLib/TimeSpan.cs,55) Has it ever changed since .NET framework 1.0 released in 2002?

I do not know how likely it is to change, but I do not want to add documentation for something we do not guarantee.

That comment you point to is a reference to the BinarySerializer serialization mechanism, which was marked Obsolete in .NET 5.0.

> Anyway, if you really want to keep it unspecified (e.g. planning to change RAM layout in the near future), then what's the recommended way to pass time intervals to native code? That structure is available since .NET Framework 1.0, and used all over the software. I know how to call `TimeSpan.FromTicks` but that's not the best usability, and in some cases can be slower.

I'd suggest calling `TimeSpan.Ticks` to get the ticks to pass down to use `TimeSpan.FromTicks` or `new TimeSpan(ticks)` to construct a `TimeSpan` from the ticks value (both of these options will likely have equivalent performance in release builds).

Const-me:
> I do not know how likely it is to change, but I do not want to add documentation for something we do not guarantee.

If that thing has not changed for 20 years and unlikely to change in the future, maybe it should be considered? Is there another repository to discuss these guarantees?

> I'd suggest calling `TimeSpan.Ticks` to get the ticks to pass down to use `TimeSpan.FromTicks` or `new TimeSpan(ticks)` to construct a `TimeSpan` from the ticks value (both of these options will likely have equivalent performance in release builds).

It depends. To marshal an array of them, you gonna need to allocate or borrow some memory, and run some code to copy the values. Pretty sure the runtime ain't gonna optimize it away.

For marshalling single values it does not harm performance, but it does harm usability. That's boilerplate code for no good reason. Multimedia libraries and many other things deal with timestamps a lot.

AaronRobinsonMSFT:
>then what's the recommended way to pass time intervals to native code?

That is a great question. The issue here is that `TimeSpan` is implemented as a Windows-ism and has no official alignment with any other platform surface area. This is alluded to in the `TimeSpan` [Remarks](https://docs.microsoft.com/dotnet/api/system.timespan#remarks):

>A tick is equal to 100 nanoseconds, or one ten-millionth of a second.

The 100 nanoseconds is arbitrary and was chosen to align with Windows' [`FILETIME`](https://docs.microsoft.com/windows/win32/api/minwinbase/ns-minwinbase-filetime) type. Forever defining this as a 64-bit integer is possible but if we do then we are also restricted on how it should be stored because we've stated it will be marshalled as a 64-bit number.

@jkoritzinsky's point is that that is not something we want to guarantee nor define for this type. Previously in .NET, the marshalling/layout of types was trivial to define because we were aligning ourselves with a single underlying platform (i.e., Windows) but that is no longer true. This means that how some types are marshalled should be done explicitly so expectations are clear and differences aren't lost.

> I'd suggest calling TimeSpan.Ticks to get the ticks to pass down to use TimeSpan.FromTicks or new TimeSpan(ticks) to construct a TimeSpan from the ticks value (both of these options will likely have equivalent performance in release builds).

Agree. The `Ticks` property has a clearly defined semantic and size as opposed to the abstract notion that is intended by the `TimeSpan` type itself.

> It depends. To marshal an array of them, you gonna need to allocate or borrow some memory, and run some code to copy the values. Pretty sure the runtime ain't gonna optimize it away.

This is a fair point. Is there a common array size that is a concern? How often is this marshalling of the `TimeSpan` type going to occur in practice? @jkoritzinsky Is this something we could optimize in the `DllImport` source generator?

> For marshalling single values it does not harm performance, but it does harm usability.

Usability isn't the high order bit for Interop. The primary focus is about accuracy and predictability of being a platform agnostic run time. Interop is an advanced scenario but we do attempt to make it as simple as possible without boxing ourselves into a corner.

> That's boilerplate code for no good reason.

It is for explicit intention and is incredibly important for Interop. For example, we recently [introduced types](https://docs.microsoft.com/dotnet/standard/native-interop/best-practices#cross-platform-data-type-considerations) to avoid unrelated cross-platform issues.

> Multimedia libraries and many other things deal with timestamps a lot.

I think that is a perfect example of why being explicit is so important. On Windows most APIs defer to the canonical 100s of nanoseconds but I'm not sure that is true in all cases. Do you know if Linux or macOS multimedia APIs assume 100s of nanoseconds?

Const-me:
@AaronRobinsonMSFT  It already works this way, and unlikely to change in the future. Processors have converged to just 2 major architectures, Intel and ARM. All of them natively support 64-bit integers. None of them natively supports 128-bit scalars. With the current 64-bit ticks at 10 MHz, both range and resolution are good enough for 99% of applications.

Usability is important, and native interop is one uniquely strong area of .NET runtime. Remove native interop from .NET and some people will pick Java/Scala/Kotlin or JavaScript/TypeScript instead, because ecosystem and job market.

> How often is this marshalling of the TimeSpan type going to occur in practice?

Personally, I only needed to pass individual values.

> Do you know if Linux or macOS multimedia APIs assume 100s of nanoseconds?

I never saw int64 10MHz ticks outside Windows.

On Linux, different APIs handle `struct timeval` microseconds or `struct timespec` nanoseconds. Neither of them is `int64_t`, they are structures of two integers the first one being seconds. V4L2 [uses](https://github.com/Const-me/Vrmac/blob/1.2/VrmacVideo/IO/Kernel/LinuxTime.cs#L30-L31) [both](https://github.com/Const-me/Vrmac/blob/1.2/VrmacVideo/IO/Kernel/LinuxTime.cs#L60-L61) in different structures.

On iOS, some parts use `struct CMTime` a rational number int64 / int32. Other parts use `TimeInterval` which is seconds in FP64 floating point representation.

GStreamer (a cross-platform multimedia framework, supports pretty much all platforms in existence) uses `int64_t` nanoseconds in most places.

If you wanna improve time precision in .NET, consider `int64_t` nanoseconds. The range is 100x smaller than the current version but still reasonable, about ±292 years.

AaronRobinsonMSFT:
> Processors have converged to just 2 major architectures, Intel and ARM. All of them natively support 64-bit integers.

The argument is the interpretation of those 64-bits and how we store data in a `TimeSpan` type. The default marshalling of that type is the issue here and its size is not set in stone nor required to be the same on all platforms.

>Usability is important, and native interop is one uniquely strong area of .NET runtime.

Agreed, but not to the detriment of leading people down a path that isn't obvious. Interop is best done explicitly when ambiguity is possible &ndash; see the `CLong` and `CULong` example. It sounds like the current desire is for a 64-bit interpretation of the `TimeSpan` based on 100s of nanoseconds so `Ticks` is the most appropriate option.

>If you wanna improve time precision in .NET, consider int64_t nanoseconds. The range is 100x smaller than the current version but still reasonable, about ±292 years.

I am not attempting to make a precision debate but rather how we store data in the `TimeSpan` type. Declaring that the `TimeSpan` type will always be marshalled as `int64_t` is the concern because we could change it. Right now the `Ticks` property exposes an explicit sized integer type that is independent of the `TimeSpan` type and therefore the best way to work with a stable contract. Declaring that the `TimeSpan` type will always be marshaled as a `int64_t` is not something we are inclined to do because it is based on a Windows platform type and we don't normally document the intrinsic size of types.

@jkoritzinsky statement about the purposes of this document are for special cases. The `TimeSpan` right now is backed by a 64-bit integer and that is fine to rely upon and even assert if desired. It is not a guarantee we are willing to stand behind forever though and declare. Using the `TimeSpan.Ticks` is the best way forward if an explicit size contract is desired.

adegeo:
@Const-me Thank you for taking the opportunity to contribute to docs! However, based on the discussion here, it sounds like this change isn't appropriate for this specific doc, so I'm going to close the PR. You can discuss design on https://github.com/dotnet/runtime/

Cheers!

PR #24219, opened 2021-05-13
Add missing null-conditional operator

duracellko:
## Summary

Example of `IAsyncDisposable` implementation has missing null-conditional operator. The `DisposeAsync` method may fail, when called 2 times. Also other methods in example use null-conditional operator, when calling `Dispose`.


IEvangelist (approved):

PR #24232, opened 2021-05-14
Acrolinx 5/13

gewarren:


docs/standard/datetime/performing-arithmetic-operations.md
```diff
@@ -42,11 +42,11 @@ More generally, the <xref:System.DateTime.Kind%2A?displayProperty=nameWithType>
...
+A <xref:System.DateTimeOffset> value includes not only a date and time, but also an offset that unambiguously defines that date and time relative to UTC. This offset makes it possible to define equality differently than for <xref:System.DateTime> values. Whereas <xref:System.DateTime> values are equal if they have the same date and time value, <xref:System.DateTimeOffset> values are equal if they both refer to the same point in time. when used in comparisons and in most arithmetic operations that determine the interval between two dates and times, a <xref:System.DateTimeOffset> value is more accurate and less in need of interpretation. The following example, which is the <xref:System.DateTimeOffset> equivalent to the previous example that compared local and UTC <xref:System.DateTimeOffset> values, illustrates this difference in behavior.
```
tdykstra, line 49:
```suggestion
A <xref:System.DateTimeOffset> value includes not only a date and time, but also an offset that unambiguously defines that date and time relative to UTC. This offset makes it possible to define equality differently than for <xref:System.DateTime> values. Whereas <xref:System.DateTime> values are equal if they have the same date and time value, <xref:System.DateTimeOffset> values are equal if they both refer to the same point in time. When used in comparisons and in most arithmetic operations that determine the interval between two dates and times, a <xref:System.DateTimeOffset> value is more accurate and less in need of interpretation. The following example, which is the <xref:System.DateTimeOffset> equivalent to the previous example that compared local and UTC <xref:System.DateTimeOffset> values, illustrates this difference in behavior.
```

tdykstra (approved):
LGTM, one suggestion

gewarren:
Thanks Tom!

PR #24054, opened 2021-05-05
Added naming convention details for C# code

IEvangelist:
## Summary

Added naming convention details for C# code.

Fixes #23835

### Internal preview

✔️ [C# Coding Conventions](https://review.docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions?branch=pr-en-us-24054)

PR #24237, opened 2021-05-14
Post merge review and customer feedback

IEvangelist:
## Summary

Related to #24054, and addressing a post merge review from @gewarren.

Also, added more explicit details about the flexibility of the guidelines, directly addresses customer feedback:

- https://github.com/dotnet/docs/pull/24054#issuecomment-840972518
- https://github.com/dotnet/docs/pull/24054#issuecomment-840983748

### DIFF
--- a/docs/csharp/programming-guide/inside-a-program/coding-conventions.md
+++ b/docs/csharp/programming-guide/inside-a-program/coding-conventions.md
@@ -1,7 +1,7 @@
 ---
 title: "C# Coding Conventions - C# Programming Guide"
 description: Learn about coding conventions in C#. Coding conventions create a consistent look to the code and facilitate copying, changing, and maintaining the code.
-ms.date: 05/05/2021
+ms.date: 05/14/2021
 helpviewer_keywords:
   - "coding conventions, C#"
   - "Visual C#, coding conventions"
@@ -27,7 +27,7 @@ There are several naming conventions to consider when writing C# code.
 
 ### Pascal case
 
-You should use pascal casing ("PascalCasing") when naming a `class`, `record`, or `struct`.
+Use pascal casing ("PascalCasing") when naming a `class`, `record`, or `struct`.
 
 ```csharp
 public class DataService
@@ -57,7 +57,7 @@ public interface IWorkerQueue
 }
 ```
 
-When naming `public` members of types, such as; fields, properties, events, methods, and local functions use pascal casing.
+When naming `public` members of types, such as fields, properties, events, methods, and local functions, use pascal casing.
 
 ```csharp
 public class ExampleEvents
@@ -81,7 +81,7 @@ public class ExampleEvents
 }
 ```
 
-When writing positional-records, use pascal casing for parameters as they're the public properties of the record.
+When writing positional records, use pascal casing for parameters as they're the public properties of the record.
 
 ```csharp
 public record PhysicalAddress(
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @IEvangelist 

Once @gewarren agrees, you can :shipit:

docs/csharp/programming-guide/inside-a-program/coding-conventions.md
```diff
@@ -19,15 +19,16 @@ Coding conventions serve the following purposes:
 > - They facilitate copying, changing, and maintaining the code.
 > - They demonstrate C# best practices.
 
-The guidelines in this article are used by Microsoft to develop samples and documentation.
+> [!IMPORTANT]
+> The guidelines in this article are used by Microsoft to develop samples and documentation. They were adopted from the [.NET Runtime, C# Coding Style](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md) guidelines. You can use them, or adapt them to your needs. The primary objectives are consistency and readability within your project, team, organization, or company source code.
```
SIkebe, line 23: >The guidelines in this article are used by Microsoft to develop samples and documentation

Hmm,... even the very basic example code in C# reference docs does not follow the newly added convention `s_`. It's not so popular outside runtime repo.
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/static#example---static-initialization

I think it's too late to add these new guidelines after 20 years have passed since C# was born.

SIkebe: Consistency in official docs is so important. It needs a lot of work to correct these examples. Is it really worth to change at this time?

IEvangelist: Hi @SIkebe,

It's never too late to update guidelines. It is natural for them to morph over time - especially with a language that continues to evolve. Following the .NET runtime as a basis is a solid foundation, since it's open source and in the view of the public. Even within the runtime code base, you can find violations and inconsistencies - that's okay. There are bound to be inconsistencies, but those too (when and where applicable) could be addressed. Guidelines are a set of suggestions, not something that is mandated. You're free to ignore them.

gewarren (approved):
Thanks for following up on this. I did start the review before the previous PR was merged.

docs/csharp/programming-guide/inside-a-program/coding-conventions.md
SIkebe: >It is natural for them to morph over time

Yeah, I agree if there is a decent need to do so, or strong feedback from users. Could you point to the discussion about this newly added convention? Users want to know "why" this happened at this time.

And how about @nanasi880 's point?
https://github.com/dotnet/docs/pull/24054#issuecomment-841094612

mayuki:
@IEvangelist 
@BillWagner 
Thank you for your quick response to my feedback!

I’m sorry I should have explained to you more clearly.

First of all, I think this document exists in a position to be regarded as an introduction to C#.

In that regard, I am concerned that many users will not read the notes on the page, but will just read the title and assume that this is a C# standard coding convention, and if they do read it, they may assume that this looks like a convention adopted (or recommended) by .NET Team or Microsoft.

If it is for internal samples or documentation, it might be better to move the location of the documentation.

In fact, unfortunately, in Japanese C# community, this has been mistaken for a standard C# coding convention, which has caused some confusion.😢

Also, as others are concerned, naming conventions such as s_ are not general in new projects today.

.NET Runtime repository is not general, as some code has a long history.

This is a document that could be considered a standard for C# at this point, so it would be nice to be cautious about areas that could be controversial.

docs/csharp/programming-guide/inside-a-program/coding-conventions.md
nanasi880: > I think it's too late to add these new guidelines after 20 years have passed since C# was born.
I agree with this.

It's not that I don't like the change, but the reason for the change is unclear, and as I pointed out, this is a change that Microsoft has explicitly denied.

It seems that static variables have an 's_' prefix and thread-static variables have 't_', what about AsyncLocal for example, and ThreadLocal?

Also, what if there are more of these in the future?
What if fiber is born? Is 'f_' the preposition that fiber local storage has?

We believe that history has proven that distinguishing these things by prepositions is a bad idea.
This is exactly what we call Hungarian notation, and we don't think it's an idea we should adopt, at least not actively now.

PR #24240, opened 2021-05-14
Enable recommendations

gewarren:


docfx.json
```diff
@@ -476,7 +477,11 @@
         "docs/standard/security/**.md": "dotnet-security",
         "docs/visual-basic/language-reference/error-messages/**/**.md": "vb-diagnostics",
         "docs/visual-basic/misc/**/**.md": "vb-diagnostics"
-      },
+      },
```
Youssef1313, line 480:
```suggestion
      },
```

docfx.json
```diff
@@ -476,7 +477,11 @@
         "docs/standard/security/**.md": "dotnet-security",
         "docs/visual-basic/language-reference/error-messages/**/**.md": "vb-diagnostics",
         "docs/visual-basic/misc/**/**.md": "vb-diagnostics"
-      },
+      },    
+     "recommendations": { 
+       "architecture/**/**.md": false,
+       "standard/design-guidelines/**/**.md": false
```
Youssef1313, line 483:
```suggestion
       "standard/design-guidelines/**/**.md": false
```

Youssef1313 (commented):
I'm curious what this setting is for?

gewarren:
> I'm curious what this setting is for?

It enables a section at the end of the article called "Recommended content" that suggests articles based on an algorithm.

docfx.json
```diff
@@ -476,7 +477,11 @@
         "docs/standard/security/**.md": "dotnet-security",
         "docs/visual-basic/language-reference/error-messages/**/**.md": "vb-diagnostics",
         "docs/visual-basic/misc/**/**.md": "vb-diagnostics"
-      },
+      },    
+     "recommendations": { 
+       "architecture/**/**.md": false,
+       "standard/design-guidelines/**/**.md": false  
+     },
```
BillWagner, line 484: We should add the language standards here:

```suggestion
        "_csharplang/spec/*.md": false,
        "_vblang/spec/*.md": false
     },
```

BillWagner (approved):
I had one additional suggestion, and then it's good.

docfx.json
gewarren: They are marked as "language-reference", so they will automatically be excluded. It probably doesn't hurt to be explicit, but then are there any other reference folders we want to explicitly exclude?

```
      "ms.topic": {
        "_csharplang/**/*.md": "language-reference",
        "_vblang/spec/*.md": "language-reference",
```

BillWagner: Those are the only ones I'd want to be explicit on. If recommendations eventually runs on reference content, I can see value in having it in other areas, but I'd still want these off.

gewarren: Since those cross-repo refs map to the csharp/language-reference and visual-basic/reference/language-specification folders, do we use the source path or the destination path here? I would guess the destination.

```
        "src": "_csharplang/spec",
        "dest": "csharp/language-reference/language-specification",

        "src": "_vblang/spec",
        "dest": "visual-basic/reference/language-specification",
```

BillWagner: I couldn't tell from the docs on the feature. You're likely right.

gewarren: Actually, I think I'm wrong based on other metadata you've set on those paths. Your initial suggestion was correct.

image

docfx.json
```diff
@@ -132,22 +137,17 @@
...
+          "folder_relative_path_in_docset": "docs/standard/modernize-with-azure-and-containers/"
+        },
+        "docs/architecture/serverless/": {
+          "folder_relative_path_in_docset": "docs/standard/serverless-architecture/"
+        }
+      }
     },
     "fileMetadata": {
       "feedback_system": {
+        "_csharplang/**/*.md": "None",
```
gewarren, line 149: @BillWagner Did you want to turn feedback off for these folders too?

BillWagner: No. I'm thinking that our docs often have more details and tutorials to learn about new features before these new features are integrated into the standard. Those recommendations could be useful.

pkulikov:
> It enables a section at the end of the article called "Recommended content" that suggests articles based on an algorithm.

@gewarren @BillWagner what is the objective of that algorithm?

Currently, it may lead to the waste of time of people who is not familiar with the language or a feature. For example, consider the reference article about the null-forgiving operator https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving:
image

I know a bit about the language and I don't understand why those four articles are recommended. If I were a novice, I would think those articles provide additional context (or examples) to understand the null-forgiving operator. I'm afraid it's not the case now. 

I've checked a number of articles in the Operators section of the C# language reference and usually the recommended content is rather misleading. At its current stage, the algorithm looks to be more useful to the authors of the docs to get the list of *candidates* for the **See also** section. Let's hope the algorithm will improve.

BillWagner:
@pkulikov Thanks for the thoughtful review.  I've forwarded these comments to the development team working on the recommended content.

IEvangelist:
> @pkulikov Thanks for the thoughtful review. I've forwarded these comments to the development team working on the recommended content.

I'd like to echo @BillWagner's appreciation of you voicing these observations and concerns, @pkulikov - much appreciated. 🙏🏽

sudeepku:
@pkulikov  Thanks for your feedback. I run the content recommendations feature on docs.microsoft.com platform. This is a new capability which we enabled on .NET content recently. We will be retraining the underlying ML models going forward. We are also considering re-phrasing the verbiage to make it less confusing. Appreciate your patience as we work on these changes and more to make the recommendations better.  Thanks @BillWagner, @IEvangelist for sharing this feedback with our product team.

pkulikov:
@BillWagner @IEvangelist @sudeepku thank you for your actions and replies! Interesting to see how this feature will evolve.

PR #24279, opened 2021-05-18
Add two no-loc words globally

gewarren:
See https://github.com/dotnet/docs.pl-pl/pull/211.

### DIFF
--- a/docfx.json
+++ b/docfx.json
@@ -121,7 +121,7 @@
       "ms.devlang": "dotnet",
       "ms.prod":"dotnet",
       "ms.topic": "conceptual",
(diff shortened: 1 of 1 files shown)


tdykstra (commented):
It would be nice to be able to do that, but the contributor guide says we can't.

>The no-loc metadata is not supported as global metadata in docfx.json file. The localization pipeline doesn't read the docfx.json file, so the no-loc metadata must be added into each individual source file.

-- https://review.docs.microsoft.com/en-us/help/contribute/markdown-reference?branch=master#non-localized-strings

gewarren:
Thanks Tom. Seemed a bit too good to be true.

PR #24283, opened 2021-05-19
Add MSBuild properties collected to list of telemetry data points

dsplaisted:
## Summary

Add MSBuild properties collected to list of telemetry data points.

Code to collect these is here: https://github.com/dotnet/sdk/blob/481c149f88d826ab456b7ed4401a14f4142ff8e9/src/Cli/dotnet/commands/dotnet-msbuild/MSBuildLogger.cs#L88-L92


tdykstra (approved):
LGTM

PR #24291, opened 2021-05-19
Rename JsonDocument how-to article

### LINKED ISSUE #24126: JSON deserialization behavior regarding private/protected/internal set properties
The page mentions that
> Deserialization to immutable objects or read-only properties isn't supported.

This is of course a very useful information but I think it could use more precision.

The deserialization works even if the class only has a `private` parameterless constructor, so I was expecting that maybe a property with a `private`/`protected`/`internal set` accessor would be populated.

It seems however that it is not the case according to my (basic) tests: the only `set` accessor accessibility that allows a `public` property to be populated is "public" (no keyword, e.g. `{ get; set; }`). If the property has `private`/`protected`/`internal set`, it doesn't seem to be populated.

What's misleading here is that the text says "read-only properties" aren't supported, but I find it hard to consider that a `public` property with a `get` accessor and a `private`/`protected`/`internal set` accessor is read-only. What are you thoughts about this?

---
#### Document Details

* Content Source: [docs/standard/serialization/system-text-json-how-to.md](https://github.com/dotnet/docs/blob/main/docs/standard/serialization/system-text-json-how-to.md)
* GitHub Login: @tdykstra

### LINKED ISSUE #24251: Suggestion: document how to deserialize json or object into a a dynamic json object (JsonDocument)
For example, given arbitrary object `obj`, I could deserialize into a "json-enabled" `JsonDocument` like this:

```
JsonDocument objJson = JsonDocument.Parse(JsonSerializer.SerializeToUtf8Bytes(obj));
```

or

```
JsonElement objJson = JsonSerializer.Deserialize<JsonElement>(JsonSerializer.SerializeToUtf8Bytes(obj));
```

I suggest to document how to achieve this scenario (maybe there is a better way than my examples above), as well as how to obtain the `JsonDocument` from a `json` string.

In other words, currently the documentation explains only the following deserialization scenario:
`json -> typed/custom POCO`
but I also want to see:
`json -> dynamic/json POCO`
`object -> dynamic/json POCO`

I see the `JsonDocument` is mentioned here:
[How to write custom serializers and deserializers with System.Text.Json](https://docs.microsoft.com/en-us/dotnet/standard/serialization/write-custom-serializer-deserializer)
but it is unclear this is where I have to look for it.

Related issue: 
- #24252 Suggestion: document how to work with non-read-only equivalent of JsonDocument 

---
#### Document Details

* Content: [How to serialize and deserialize JSON using C# - .NET](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to?pivots=dotnet-5-0#how-to-read-json-as-net-objects-deserialize)
* Content Source: [docs/standard/serialization/system-text-json-how-to.md](https://github.com/dotnet/docs/blob/main/docs/standard/serialization/system-text-json-how-to.md)
* GitHub Login: @tdykstra

tdykstra:
Fixes #24251 
Fixes #24126


IEvangelist (approved):

PR #24302, opened 2021-05-20
do not deploy extension on build

Youssef1313:
See https://github.com/dotnet/samples/pull/4594.


BillWagner (approved):
Thanks for this fix @Youssef1313 

I'll :shipit: now.

PR #24309, opened 2021-05-20
updated reverseName and structReverseName examples to return expected…

lc9er:
… values.

## Summary

Describe your changes here.

Working through the examples in the 'Performance' section of the coding conventions docs. I noticed that both examples `reverseName` and `structReverseName` returned `val it : Name = Name "System.String[]"` and `val it : SName = SName "System.String[]"`, respectively, rather than the reversed string.  
  
With the updated code, each function returns a reversed Name/SName string value. 

Still very new to F#, so apologies if I missed the point of the example.

### DIFF
--- a/docs/fsharp/style-guide/conventions.md
+++ b/docs/fsharp/style-guide/conventions.md
@@ -513,13 +513,15 @@ The previous observations about performance with struct tuples and records also 
     let reverseName (Name s) =
         s.ToCharArray()
         |> Array.rev
-        |> string
+        |> Array.map Char.ToString
+        |> String.concat ""
         |> Name
 
     let structReverseName (SName s) =
         s.ToCharArray()
         |> Array.rev
-        |> string
+        |> Array.map Char.ToString
+        |> String.concat ""
         |> SName
 ```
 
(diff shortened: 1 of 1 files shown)


docs/fsharp/style-guide/conventions.md
```diff
@@ -513,13 +513,15 @@ The previous observations about performance with struct tuples and records also
     let reverseName (Name s) =
         s.ToCharArray()
         |> Array.rev
-        |> string
+        |> Array.map Char.ToString
+        |> String.concat ""
```
cartermp, line 517: This is certainly confusing, but it should actually be `System.String` and not `string`.

Explanation:

Pipelining into `System.String` calls the `String()` constructor, passing in the reverse `char[]`, which then creates a new string.

Pipelining into `string` passes the `char[]` to a function that calls `.ToString()` on that array. This produces what you saw.

lc9er:
Ah. Ok. This makes sense. Thanks for taking the time to explain.

cartermp:
@lc9er do you still want to proceed with making the changes? In this case, just changing `string` to `System.String` will do the trick.

lc9er:
Should I resubmit? This is a new process for me, so sorry for my ignorance
here.

cartermp:
@lc9er No worries, you can edit the same file in this branch and it should be good to go.

lc9er:
> 
> 
> @lc9er No worries, you can edit the same file in this branch and it should be good to go.

Alright. I *think* I made the change as needed. If not, please feel free to make the change in whatever is the easiest for you.

cartermp (approved):
Thanks!

PR #24314, opened 2021-05-20
Turn off selected recommendations

BillWagner:
Turn of "recommended content" generation in two cases:

- When the article is part of a series. These are identified by finding the "step-by-step" div  for previous and next actions.
- When the article has a clear "Next actions" at the close. We're directing the reader to a single next step, so why distract them?

Where a single folder was affected, the folder was configured in docfx.json. For most of the tutorials, these were mixed in folders with other content, so the change is in the file itself.

I tagged several people for review. You only need to check the content you're responsible for, not the entire PR.


BillWagner:
/cc @luisquintanilla @cjgronlund 

I didn't make these same changes in the ML.NET and .NET for Apache Spark folders. Let me know if I should add them to this PR.

gewarren (approved):

tdykstra (approved):

PR #24315, opened 2021-05-20
Clarify volatile docs

GrabYourPitchforks:
Based on feedback received via email, clarify the C# _volatile_ keyword docs by stating explicitly that writes are not published immediately.

/cc @kouvel and @mangod9 as _System.Threading_ area owners, since they likely have opinions on this.

### DIFF
--- a/docs/csharp/language-reference/keywords/volatile.md
+++ b/docs/csharp/language-reference/keywords/volatile.md
@@ -13,6 +13,9 @@ ms.assetid: 78089bc7-7b38-4cfd-9e49-87ac036af009
 
 The `volatile` keyword indicates that a field might be modified by multiple threads that are executing at the same time. The compiler, the runtime system, and even hardware may rearrange reads and writes to memory locations for performance reasons. Fields that are declared `volatile` are not subject to these optimizations. Adding the `volatile` modifier ensures that all threads will observe volatile writes performed by any other thread in the order in which they were performed. There is no guarantee of a single total ordering of volatile writes as seen from all threads of execution.
 
+> [!NOTE]
+> When writing to a field marked `volatile`, the `volatile` keyword only controls the order in which writes are published. It does not guarantee that these writes are published immediately.
+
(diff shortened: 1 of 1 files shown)


docs/csharp/language-reference/keywords/volatile.md
```diff
@@ -13,6 +13,9 @@ ms.assetid: 78089bc7-7b38-4cfd-9e49-87ac036af009
 
 The `volatile` keyword indicates that a field might be modified by multiple threads that are executing at the same time. The compiler, the runtime system, and even hardware may rearrange reads and writes to memory locations for performance reasons. Fields that are declared `volatile` are not subject to these optimizations. Adding the `volatile` modifier ensures that all threads will observe volatile writes performed by any other thread in the order in which they were performed. There is no guarantee of a single total ordering of volatile writes as seen from all threads of execution.
 
+> [!NOTE]
+> When writing to a field marked `volatile`, the `volatile` keyword only controls the order in which writes are published. It does not guarantee that these writes are published immediately.
```
kouvel, line 17:
```suggestion
> When writing to a field marked `volatile`, the `volatile` keyword controls the order in which writes are performed. It does not guarantee that these writes are visible to other threads immediately.
```

Thought this might clarify what published means in the two cases. Also changed `only controls` to `controls` since there are other effects. The docs probably could use more clarity, it doesn't look like they're fully accurate either, filed https://github.com/dotnet/docs/issues/24318.

kouvel (approved):
LGTM, just a suggestion

docs/csharp/language-reference/keywords/volatile.md
GrabYourPitchforks: I took this suggestion but switched some words around for better flow - thanks!

mangod9 (approved):

paulomorgado (approved):

BillWagner (approved):
Thanks @GrabYourPitchforks 

I'll :shipit: now.

PR #24334, opened 2021-05-22
Update config.yml

mohini1597:
## Summary

Describe your changes here.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-blank_issues_enabled: true
+blank_issues_enabled: false
 contact_links:
   - name: docs.microsoft.com site feedback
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,4 +1,4 @@
-blank_issues_enabled: true
+blank_issues_enabled: false
```
Youssef1313, line 1: Why do we want to disable blank issues? It's sometimes useful if the issue doesn't match one of the templates.

IEvangelist:
Hi @mohini1597 - thank you for posting this pull request. But without additional context or reason, I'm going to close this. It's easy to create an issue from a template, and then clear it out - but we do not want to lead people with an empty issue. Happy to discuss this further, but for now we're going to close this.

PR #24337, opened 2021-05-23
Fix link to .NET samples JSON project

konrad-jamrozik:
## Summary

As in PR title.


konrad-jamrozik:
My bad, I think the link is actually OK.

PR #24360, opened 2021-05-24
fix some typos in the Docker configuration for the dapr-for-dotnet-developers getting started example

FacioRatio:
## Summary

Fixes "connection refused" error that occurs in Index.cshtml.cs of DaprFrontEnd when trying to invoke "weatherforecast" method of "daprbackend"


IEvangelist (approved):

IEvangelist:
@Youssef1313 - is this a false negative on the **links verifier**?

Youssef1313:
@IEvangelist The action run here is using the old version before the fix from https://github.com/dotnet/docs/pull/24372.

Ignoring the failure should be okay, subsequent PRs should succeed properly.

In case you don't want to merge with a failing check, merging the current main into the PR branch should do the work.

PR #24372, opened 2021-05-25
Fix links verifier failure

Youssef1313:


Youssef1313:
@gewarren The workflow failed due to me having a `global.json` with a specific version, and the official docker image has updated the version. I updated the version in my `global.json` temporarily and added a fix to 0.0.9.

This is expected to fail again though if the docker image is updated again *before* I release a version with pinned SDK version in my Dockerfile.

Sorry for the inconvenience here. I'll try to get things more stable soon!

tdykstra (approved):

PR #24374, opened 2021-05-25
Update create-dotnet-github-action.md

Youssef1313:
I'm not sure if this information is useful enough. But it's a mistake I did.

### DIFF
--- a/docs/devops/create-dotnet-github-action.md
+++ b/docs/devops/create-dotnet-github-action.md
@@ -106,6 +106,9 @@ The preceding *Dockerfile* steps include:
 > [!TIP]
 > The MCR in `mcr.microsoft.com` stands for "Microsoft Container Registry", and is Microsoft's syndicated container catalog from the official Docker hub. For more information, see [Microsoft syndicates container catalog](https://azure.microsoft.com/blog/microsoft-syndicates-container-catalog/).
 
+> [!CAUTION]
+> If you use a *global.json* file to pin the SDK version (for example, 5.0.300), you should use `mcr.microsoft.com/dotnet/sdk:5.0.300` to prevent breaking the GitHub Action when a new 5.0.x version is released.
+
 ## Define action inputs and outputs
 
 In the [Explore the app](#explore-the-app) section, you learned about the `ActionInputs` class. This object represents the inputs for the GitHub Action. For GitHub to recognize that the repository is a GitHub Action, you need to have an *action.yml* file at the root of the repository.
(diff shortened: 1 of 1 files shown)


docs/devops/create-dotnet-github-action.md
```diff
@@ -106,6 +106,9 @@ The preceding *Dockerfile* steps include:
 > [!TIP]
 > The MCR in `mcr.microsoft.com` stands for "Microsoft Container Registry", and is Microsoft's syndicated container catalog from the official Docker hub. For more information, see [Microsoft syndicates container catalog](https://azure.microsoft.com/blog/microsoft-syndicates-container-catalog/).
 
+> [!CAUTION]
+> If you use a *global.json* file to pin the SDK version (for example, 5.0.300), you should use `mcr.microsoft.com/dotnet/sdk:5.0.300` to prevent breaking the GitHub Action when a new 5.0.x version is released.
```
IEvangelist, line 110:
```suggestion
> If you use a *global.json* file to pin the SDK version (for example, `5.0.300`), you should use `mcr.microsoft.com/dotnet/sdk:5.0.300` to prevent breaking the GitHub Action when a new 5.0.x version is released.
```

IEvangelist (approved):
Ah, interesting find. Yeah, I try to avoid using the _global.json_ as much as possible :)

docs/devops/create-dotnet-github-action.md
```diff
@@ -106,6 +106,9 @@ The preceding *Dockerfile* steps include:
 > [!TIP]
 > The MCR in `mcr.microsoft.com` stands for "Microsoft Container Registry", and is Microsoft's syndicated container catalog from the official Docker hub. For more information, see [Microsoft syndicates container catalog](https://azure.microsoft.com/blog/microsoft-syndicates-container-catalog/).
 
+> [!CAUTION]
+> If you use a *global.json* file to pin the SDK version (for example, `5.0.300`), you should use `mcr.microsoft.com/dotnet/sdk:5.0.300` to prevent breaking the GitHub Action when a new 5.0.x version is released.
+
```
CamSoper, line 111:
```suggestion
> [!CAUTION]
> If you use a *global.json* file to pin the SDK version, you should explicitly refer to that version in your *Dockerfile*. For example, if you've used *global.json* to pin SDK version `5.0.300`, your *Dockerfile* should use `mcr.microsoft.com/dotnet/sdk:5.0.300`. This prevents breaking the GitHub Action when a new minor revision is released.
```

Seemed a little *too* specific as originally written. How does this look?

CamSoper (approved):

PR #24418, opened 2021-05-27
Executable ref breaking change

### LINKED ISSUE #24231: Generate errors when referencing an Executable project with different SelfContained value
## Generate errors when referencing an Executable project with different SelfContained value

Generally, an executable project references library projects, not other executable projects.  When an executable project does reference another executable project, it may be in order to use APIs that are defined in the executable project.  However, some customers want to reference an executable project from another executable project so that both apps will end up in and be runnable from the same output folder.  In .NET SDK 5.0.200 (with https://github.com/dotnet/sdk/pull/14488), we made this second scenario work better by copying additional files that are required for the referenced executable to run.

However, this scenario still does not work if a self-contained executable references a non self-contained executable, or vice versa.  Because of how the apphost works, both apps won't be launchable.  As such, we added an errors (NETSDK1150, NETSDK1151) when we detect this situation.  
https://github.com/dotnet/sdk/issues/15117
https://github.com/dotnet/sdk/pull/15134

We later added exclusions for this error for Blazor and for test scenarios that we identified as incorrectly triggering this error as both projects were not meant to be launchable by the apphost.

### Version introduced

5.0.300-preview3

### Old behavior
Customers could reference a self-contained exe project from a non self-contained exe project without a build error, but both apps would not be runnable.

### New behavior
If an executable project references another executable project and the SelfContained values don't match, an error will be generated.

### Reason for change
Prevent customers from getting into a state where they expect to be able to launch both applications and cannot.

### Recommended action
If the reason for the project reference is not to produce a runnable version of the referenced project in the output folder, customers can set a property to avoid this error check:
`<ValidateExecutableReferencesMatchSelfContained>false<ValidateExecutableReferencesMatchSelfContained>`

### Category

- [ ] ASP.NET Core
- [ ] C#
- [ ] Code analysis
- [ ] Core .NET libraries
- [ ] Cryptography
- [ ] Data
- [ ] Debugger
- [ ] Deployment
- [ ] Globalization
- [ ] Interop
- [ ] JIT
- [ ] LINQ
- [ ] MSBuild
- [ ] Networking
- [ ] Printing
- [ ] Security
- [ ] Serialization
- [x] SDK
- [ ] Visual Basic
- [ ] Windows Forms
- [ ] XML, XSLT

### Affected APIs
 "Not detectable via API analysis"

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #24231

[Internal preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/5.0/referencing-executable-generates-error?branch=pr-en-us-24418).

### DIFF
--- a/docs/core/compatibility/5.0.md
+++ b/docs/core/compatibility/5.0.md
@@ -130,6 +130,7 @@ If you're migrating an app to .NET 5, the breaking changes listed here might aff
 
 ## SDK
 
+- [Error generated when executable project references mismatched executable](sdk/5.0/referencing-executable-generates-error.md)
 - [OutputType set to WinExe](sdk/5.0/automatically-infer-winexe-output-type.md)
 - [WinForms and WPF apps use Microsoft.NET.Sdk](sdk/5.0/sdk-and-target-framework-change.md)
 
--- a/docs/core/compatibility/sdk/5.0/referencing-executable-generates-error.md
+++ b/docs/core/compatibility/sdk/5.0/referencing-executable-generates-error.md
@@ -1,0 +1,46 @@
+---
+title: "Breaking change: Error generated when executable references executable"
+description: Learn about the breaking change in .NET 5 where compile-time errors are generated when an executable references an executable whose SelfContained value is different.
+ms.date: 05/26/2021
+---
+# Error generated when executable project references mismatched executable
+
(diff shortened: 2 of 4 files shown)


BillWagner (approved):
This LGTM.

PR #24452, opened 2021-05-29
Typo "NuGet.config files"→"NuGet.Config files"

hyoshioka0128:
https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-update-source
#PingMSFTDocs

## Summary

Describe your changes here.


IEvangelist:
According to the reference docs from NuGet, the config file is either _NuGet.Config_ or _NuGet.config_. See [nuget.config reference](https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file). I do not think this PR is necessary.

PR #24464, opened 2021-06-01
Clarify finally block description

AdarshPala:
## Summary

I added a "not" to the description.


docs/csharp/fundamentals/exceptions/index.md
```diff
@@ -29,7 +29,7 @@ Exceptions have the following properties:
 - If a `catch` block defines an exception variable, you can use it to obtain more information about the type of exception that occurred.
 - Exceptions can be explicitly generated by a program by using the `throw` keyword.
 - Exception objects contain detailed information about the error, such as the state of the call stack and a text description of the error.
-- Code in a `finally` block is executed even if an exception is thrown. Use a `finally` block to release resources, for example to close any streams or files that were opened in the `try` block.
+- Code in a `finally` block is executed even if an exception is not thrown. Use a `finally` block to release resources, for example to close any streams or files that were opened in the `try` block.
```
AdarshPala, line 32:
```suggestion
- Code in a `finally` block is executed regardless of if an exception is thrown. Use a `finally` block to release resources, for example to close any streams or files that were opened in the `try` block.
```
Here's an alternate description if needed. Now that I think about it the original description also works; feel free to close this. Sorry for the confusion!

BillWagner (approved):
Thanks for making this change @AdarshPala 

I like the alternative as the most clear. I've accepted that suggestion, and I'll :shipit: now.

Thanks for helping us improve.

PR #24465, opened 2021-06-01
Update F# on Azure Docs - Queue Storage

zedy-wj:
1.Updating F# doc to the new SDKs - Queue Storage

2.Updated files : 

- i.docs/fsharp/using-fsharp-on-azure/queue-storage.md
- ii.samples/snippets/fsharp/azure/queue-storage.fsx

@jongio for notification.

### DIFF
--- a/docs/fsharp/using-fsharp-on-azure/queue-storage.md
+++ b/docs/fsharp/using-fsharp-on-azure/queue-storage.md
@@ -52,7 +52,7 @@ Using Azure Configuration Manager is optional. You can also use an API such as t
 
 The `QueueClient` class enables you to retrieve queues stored in Queue storage. Here's one way to create the service client:
 
-[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L19-L20)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L19-L19)]
 
 Now you are ready to write code that reads data from and writes data to Queue storage.
 
@@ -60,7 +60,7 @@ Now you are ready to write code that reads data from and writes data to Queue st
 
 This example shows how to create a queue if it doesn't already exist:
 
-[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L26-L27)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L25-L25)]
 
 ## Insert a message into a queue
 
@@ -69,20 +69,20 @@ Message. Next, call the `SendMessage` method. A
 Message can be created from string (in UTF-8
 format), like this:
 
-[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L33-L34)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L31-L31)]
 
 ## Peek at the next message
 
 You can peek at the message in the front of a queue, without removing it
 from the queue, by calling the `PeekMessage` method.
 
-[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L40-L41)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L37-L38)]
 
 ## Get the next message for processing
 
 You can retrieve the message at the front of a queue for processing by calling the `ReceiveMessage` method.
 
-[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L47-L48)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L44-L44)]
 
 You later indicate successful processing of the message by using `DeleteMessage`.
 
@@ -101,7 +101,7 @@ you would keep a retry count as well, and if the message is retried more
 than some number of times, you would delete it. This protects against a message
 that triggers an application error each time it is processed.
 
-[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L54-L59)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L50-L54)]
 
 ## De-queue the next message
 
@@ -116,13 +116,13 @@ software failure, another instance of your code can get the same message
 and try again. Your code calls `DeleteMessage` right after the message
 has been processed.
 
-[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L65-L67)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L60-L61)]
 
 ## Use Async workflows with common Queue storage APIs
 
 This example shows how to use an async workflow with common Queue storage APIs.
 
-[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L73-L82)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L67-L76)]
 
 ## Additional options for de-queuing messages
 
@@ -136,20 +136,20 @@ each message. The 5 minutes starts for all messages at the same
 time, so after 5 minutes have passed since the call to `ReceiveMessages`, any
 messages that have not been deleted will become visible again.
 
-[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L88-L90)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L82-L84)]
 
 ## Get the queue length
 
 You can get an estimate of the number of messages in a queue. The `GetProperties` method asks the Queue service to retrieve the queue attributes, including the message count. The `ApproximateMessagesCount` property returns the last value retrieved by the `GetProperties` method, without calling the Queue service.
 
-[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L96-L97)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L90-L91)]
 
 ## Delete a queue
 
 To delete a queue and all the messages contained in it, call the
 `Delete` method on the queue object.
 
-[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L103-L103)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L97-L97)]
 
 ## Next steps
 
--- a/docs/fsharp/using-fsharp-on-azure/queue-storage.md
+++ b/docs/fsharp/using-fsharp-on-azure/queue-storage.md
@@ -154,7 +154,7 @@ To delete a queue and all the messages contained in it, call the
 
 ## Note
 
-If you're migrating from the old libraries, they Base64 encoded messages by default but the new libraries do not because it's more performant. 
+If you're migrating from the old libraries, they Base64 encoded messages by default but the new libraries do not because it's more performant.
 See [MessageEncoding](https://docs.microsoft.com/dotnet/api/azure.storage.queues.queueclientoptions.messageencoding?view=azure-dotnet#Azure_Storage_Queues_QueueClientOptions_MessageEncoding) for how to set that up.
 
 ## Next steps
--- a/docs/fsharp/using-fsharp-on-azure/queue-storage.md
+++ b/docs/fsharp/using-fsharp-on-azure/queue-storage.md
@@ -30,7 +30,7 @@ Next, use a [package manager](package-management.md) such as [Paket](https://fsp
 
 Add the following `open` statements to the top of the `queues.fsx` file:
 
-[!code-fsharp[QueueStorage](~/samples/snippets/fsharp/azure/queue-storage.fsx#L1-L3)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L1-L3)]
 
 ### Get your connection string
 
--- a/docs/fsharp/using-fsharp-on-azure/queue-storage.md
+++ b/docs/fsharp/using-fsharp-on-azure/queue-storage.md
@@ -152,9 +152,13 @@ To delete a queue and all the messages contained in it, call the
 
 [!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L99-L99)]
 
+## Note
+
+If you're migrating from the old libraries, they Base64 encoded messages by default but the new libraries do not because it's more performant. 
+See [MessageEncoding](https://docs.microsoft.com/en-us/dotnet/api/azure.storage.queues.queueclientoptions.messageencoding?view=azure-dotnet#Azure_Storage_Queues_QueueClientOptions_MessageEncoding) for how to set that up.
+
 ## Next steps
 
-If you're migrating from the old libraries, they Base64 encoded messages by default but the new libraries do not because it's more performant. See [MessageEncoding](https://docs.microsoft.com/en-us/dotnet/api/azure.storage.queues.queueclientoptions.messageencoding?view=azure-dotnet#Azure_Storage_Queues_QueueClientOptions_MessageEncoding) for how to set that up.
 Now that you've learned the basics of Queue storage, follow these links to learn about more complex storage tasks.
 
 - [Azure Storage APIs for .NET](/dotnet/api/overview/azure/storage)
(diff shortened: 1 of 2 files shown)


docs/fsharp/using-fsharp-on-azure/queue-storage.md
```diff
@@ -30,67 +30,53 @@ Next, use a [package manager](package-management.md) such as [Paket](https://fsp
...
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L19-L20)]
 
 Now you are ready to write code that reads data from and writes data to Queue storage.
 
-## Create a queue
-
-This example shows how to create a queue if it doesn't already exist:
-
-[!code-fsharp[QueueStorage](~/samples/snippets/fsharp/azure/queue-storage.fsx#L32-L36)]
-
```
jongio, line 72: Why was this removed?

docs/fsharp/using-fsharp-on-azure/queue-storage.md
```diff
@@ -124,40 +110,40 @@ software failure, another instance of your code can get the same message
...
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L68-L77)]
 
 ## Additional options for de-queuing messages
 
 There are two ways you can customize message retrieval from a queue.
 First, you can get a batch of messages (up to 32). Second, you can set a
 longer or shorter invisibility timeout, allowing your code more or less
 time to fully process each message. The following code example uses
-`GetMessages` to get 20 messages in one call and then processes
+`ReceiveMessages` to get 20 messages in one call and then processes
```
jongio, line 127: Is this still true that it will get 20 messages?

docs/fsharp/using-fsharp-on-azure/queue-storage.md
```diff
@@ -124,40 +110,40 @@ software failure, another instance of your code can get the same message
...
 each message. It also sets the invisibility timeout to five minutes for
 each message. The 5 minutes starts for all messages at the same
-time, so after 5 minutes have passed since the call to `GetMessages`, any
+time, so after 5 minutes have passed since the call to `ReceiveMessages`, any
```
jongio, line 130: Is the default timeout still 5 min?

docs/fsharp/using-fsharp-on-azure/queue-storage.md
```diff
@@ -30,67 +30,53 @@ Next, use a [package manager](package-management.md) such as [Paket](https://fsp
 
 Add the following `open` statements to the top of the `queues.fsx` file:
 
-[!code-fsharp[QueueStorage](~/samples/snippets/fsharp/azure/queue-storage.fsx#L1-L3)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L1-L2)]
```
jongio, line 33: Why were all the paths changed from ~/ to ../../../?

jongio (changes requested):

samples/snippets/fsharp/azure/queue-storage.fsx
```diff
@@ -12,102 +11,88 @@ let storageConnString = "..." // fill this in from your storage account
...
+queueClient.CreateIfNotExists()
 
 //
 // Insert a message into a queue.
 //
 
 // Create a message and add it to the queue.
-let message = new CloudQueueMessage("Hello, World")
-queue.AddMessage(message)
+let message = "Hello, World"
```
jsquire, line 27: Given that this is a primitive now, it may be better to collapse 27 and 28 into a single line.    The comment on L26 should also be adjusted.

Is this restricted to `string` or is there a more complex flow that we should also consider demonstrating?

samples/snippets/fsharp/azure/queue-storage.fsx
```diff
@@ -12,102 +11,88 @@ let storageConnString = "..." // fill this in from your storage account
...
 
 //
 // Peek at the next message.
 //
 
 // Peek at the next message.
-let peekedMessage = queue.PeekMessage()
-let msgAsString = peekedMessage.AsString
+let peekedMessage = queueClient.PeekMessage()
+let msgToString = peekedMessage.ToString()
```
jsquire, line 36:
```suggestion
let messageContents = peekedMessage.ToString()
```

samples/snippets/fsharp/azure/queue-storage.fsx
```diff
@@ -12,102 +11,88 @@ let storageConnString = "..." // fill this in from your storage account
...
 //
 // Use Async-Await pattern with common Queue storage APIs.
 //
 
 async {
-    let! exists = queue.CreateIfNotExistsAsync() |> Async.AwaitTask
+    let! exists = queueClient.CreateIfNotExistsAsync() |> Async.AwaitTask
 
-    let! retrieved = queue.GetMessageAsync() |> Async.AwaitTask
+    let! receipt = queueClient.ReceiveMessageAsync() |> Async.AwaitTask
```
jsquire, line 70: I'm not sure why we're using different nomenclature for a received message:

- `receipt` here
- `retrieved` on L43
- `peekedMessage` on L35
- `message` for a string on L27
- `msg` on L82

We should consider normalizing terminology unless we've got this many types to represent the same concept.

samples/snippets/fsharp/azure/queue-storage.fsx
```diff
@@ -12,102 +11,88 @@ let storageConnString = "..." // fill this in from your storage account
...
-queue.FetchAttributes()
-let count = queue.ApproximateMessageCount.GetValueOrDefault()
+let properties = queueClient.GetProperties().Value
+let count = properties.ApproximateMessagesCount
 
 //
 // Delete a queue.
 //
 
 // Delete the queue.
```
jsquire, line 97:
```suggestion
```

This comment doesn't add any context to the block on L93-95,  may as well delete it.

samples/snippets/fsharp/azure/queue-storage.fsx
```diff
@@ -12,102 +11,88 @@ let storageConnString = "..." // fill this in from your storage account
...
 //
 // Change the contents of a retrieved message.
 //
 
 // Update the message contents and set a new timeout.
-retrieved.SetMessageContent("Updated contents.")
-queue.UpdateMessage(retrieved,
-    TimeSpan.FromSeconds(60.0),
-    MessageUpdateFields.Content ||| MessageUpdateFields.Visibility)
+queueClient.UpdateMessage(
```
jsquire, line 50: If the call on L43 fails, what happens when someone tries to reference the `Value` property?   If seems like we should discuss and cover it as a failure scenario.

docs/fsharp/using-fsharp-on-azure/queue-storage.md
jsquire: Looks that way from the code snippet.

samples/snippets/fsharp/azure/queue-storage.fsx
```diff
@@ -12,102 +11,88 @@ let storageConnString = "..." // fill this in from your storage account
...
-//
-// Create a queue.
-//
-
-// Retrieve a reference to a container.
-let queue = queueClient.GetQueueReference("myqueue")
-
-// Create the queue if it doesn't already exist
-queue.CreateIfNotExists()
+let queueClient = new QueueClient(storageConnString, "myqueue");
```
cartermp, line 19:
```suggestion
let queueClient = QueueClient(storageConnString, "myqueue");
```

Unless this is disposable, you can omit the `new`

zedy-wj: Thanks for your guidance!

samples/snippets/fsharp/azure/queue-storage.fsx
zedy-wj: Thanks for your guidance! I have adjust it.

Parameter `message` in the method `AddMessage` is only support two types `string` and `BinaryData` now.

docs/fsharp/using-fsharp-on-azure/queue-storage.md
zedy-wj: Because all the paths ~/ could not link to code page.

docs/fsharp/using-fsharp-on-azure/queue-storage.md
zedy-wj: Put it together with `Create the Queue Service client.`. It has been token back now.

docs/fsharp/using-fsharp-on-azure/queue-storage.md
zedy-wj: Yes, it was set [here](https://github.com/zedy-wj/docs/blob/queue-storage/samples/snippets/fsharp/azure/queue-storage.fsx#L88).

docs/fsharp/using-fsharp-on-azure/queue-storage.md
zedy-wj: Yes, it was set [here](https://github.com/zedy-wj/docs/blob/queue-storage/samples/snippets/fsharp/azure/queue-storage.fsx#L88).

samples/snippets/fsharp/azure/queue-storage.fsx
```diff
@@ -12,102 +11,93 @@ let storageConnString = "..." // fill this in from your storage account
...
-
-// Parse the connection string and return a reference to the storage account.
-let storageAccount = CloudStorageAccount.Parse(storageConnString)
 
 //
 // Create the Queue Service client.
 //
 
-let queueClient = storageAccount.CreateCloudQueueClient()
+// Instantiate a QueueClient which will be used to create and manipulate the queue
```
jsquire, line 19: It seems that the first comment after the 3-line header block for each section just repeats what the header mentions.   We should probably consider collapsing them; they're pretty redundant.

samples/snippets/fsharp/azure/queue-storage.fsx
```diff
@@ -12,102 +11,93 @@ let storageConnString = "..." // fill this in from your storage account
...
+    "Updated contents.",
+    TimeSpan.FromSeconds(60.0))
 
 //
 // De-queue the next message, indicating successful processing
 //
 
 // Process the message in less than 30 seconds, and then delete the message.
-queue.DeleteMessage(retrieved)
+let delMessage = queueClient.ReceiveMessage().Value
```
jsquire, line 66:
```suggestion
let deleteMessage = queueClient.ReceiveMessage().Value
```

jsquire (approved):
LGTM

zedy-wj:
Complete all the changes according to comments. Thanks for your guidance!

docs/fsharp/using-fsharp-on-azure/queue-storage.md
```diff
@@ -30,67 +30,59 @@ Next, use a [package manager](package-management.md) such as [Paket](https://fsp
...
-To parse the connection string, use:
-
-[!code-fsharp[QueueStorage](~/samples/snippets/fsharp/azure/queue-storage.fsx#L19-L20)]
-
-This will return a `CloudStorageAccount`.
-
 ### Create the Queue service client
 
-The `CloudQueueClient` class enables you to retrieve queues stored in Queue storage. Here's one way to create the service client:
+The `QueueClient` class enables you to retrieve queues stored in Queue storage. Here's one way to create the service client:
```
tg-msft, line 53: There's a separate type called `QueueServiceClient` so let's drop the word "service".

```suggestion
The `QueueClient` class enables you to retrieve queues stored in Queue storage. Here's one way to create the client:
```

samples/snippets/fsharp/azure/queue-storage.fsx
```diff
@@ -12,102 +11,87 @@ let storageConnString = "..." // fill this in from your storage account
...
 
 //
 // Peek at the next message.
 //
 
-// Peek at the next message.
-let peekedMessage = queue.PeekMessage()
-let msgAsString = peekedMessage.AsString
+let peekedMessage = queueClient.PeekMessage()
+let messageContents = peekedMessage.ToString()
```
tg-msft, line 38: Have you tested this?  We don't override `ToString` so I don't think you'll get anything good out here.  At a minimum, I think you want `peekedMessage.Value.Body.ToString()`.

samples/snippets/fsharp/azure/queue-storage.fsx
```diff
@@ -12,102 +11,87 @@ let storageConnString = "..." // fill this in from your storage account
...
+let properties = queueClient.GetProperties().Value
+let count = properties.ApproximateMessagesCount
 
 //
 // Delete a queue.
 //
 
-// Delete the queue.
-queue.Delete()
+queueClient.Delete()
```
tg-msft, line 97: Prefer `DeleteIfExists` over `Delete`.

docs/fsharp/using-fsharp-on-azure/queue-storage.md
```diff
@@ -30,67 +30,59 @@ Next, use a [package manager](package-management.md) such as [Paket](https://fsp
...
 
 ## Insert a message into a queue
 
 To insert a message into an existing queue, first create a new
-`CloudQueueMessage`. Next, call the `AddMessage` method. A
-`CloudQueueMessage` can be created from either a string (in UTF-8
-format) or a `byte` array, like this:
+Message. Next, call the `SendMessage` method. A
+Message can be created from string (in UTF-8
+format), like this:
```
tg-msft, line 70: It might be worth showing off different examples with [BinaryData](https://docs.microsoft.com/en-us/dotnet/api/overview/azure/system.memory.data-readme) because JSON is the most common customer use case for messages.

docs/fsharp/using-fsharp-on-azure/queue-storage.md
```diff
@@ -124,40 +116,40 @@ software failure, another instance of your code can get the same message
 and try again. Your code calls `DeleteMessage` right after the message
 has been processed.
 
-[!code-fsharp[QueueStorage](~/samples/snippets/fsharp/azure/queue-storage.fsx#L75-L76)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L60-L61)]
 
 ## Use Async workflows with common Queue storage APIs
 
 This example shows how to use an async workflow with common Queue storage APIs.
```
tg-msft, line 123: We should add a comment that all of the Queue methods we've shown so far have `Async` alternatives if customers would prefer the more performant option that doesn't block.

docs/fsharp/using-fsharp-on-azure/queue-storage.md
```diff
@@ -124,40 +116,40 @@ software failure, another instance of your code can get the same message
...
+time, so after 5 minutes have passed since the call to `ReceiveMessages`, any
 messages that have not been deleted will become visible again.
 
-[!code-fsharp[QueueStorage](~/samples/snippets/fsharp/azure/queue-storage.fsx#L97-L99)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L82-L84)]
 
 ## Get the queue length
 
-You can get an estimate of the number of messages in a queue. The `FetchAttributes` method asks the Queue service to retrieve the queue attributes, including the message count. The `ApproximateMessageCount` property returns the last value retrieved by the `FetchAttributes` method, without calling the Queue service.
+You can get an estimate of the number of messages in a queue. The `GetProperties` method asks the Queue service to retrieve the queue attributes, including the message count. The `ApproximateMessagesCount` property returns the last value retrieved by the `GetProperties` method, without calling the Queue service.
```
tg-msft, line 143: That "without calling the service" bit is no longer relevant in the latest client libraries.  (We still don't call the service, but NONE of our models will ever call the service.)

```suggestion
You can get an estimate of the number of messages in a queue. The `GetProperties` method asks the Queue service to retrieve the queue attributes, including the message count. The `ApproximateMessagesCount` property returns the last value retrieved by the `GetProperties` method.
```

docs/fsharp/using-fsharp-on-azure/queue-storage.md
```diff
@@ -124,40 +116,40 @@ software failure, another instance of your code can get the same message
...
 
 ## Delete a queue
 
 To delete a queue and all the messages contained in it, call the
 `Delete` method on the queue object.
 
-[!code-fsharp[QueueStorage](~/samples/snippets/fsharp/azure/queue-storage.fsx#L112-L113)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L97-L97)]
 
 ## Next steps
```
tg-msft, line 154: We should add a blurb that if you're migrating from the old libraries, they Base64 encoded messages by default but the new libraries do not because it's more performant.  See https://docs.microsoft.com/en-us/dotnet/api/azure.storage.queues.queueclientoptions.messageencoding?view=azure-dotnet#Azure_Storage_Queues_QueueClientOptions_MessageEncoding for how to set that up.

This is the most common problem people have when migrating to the new Queues library.

zedy-wj:
Hi, @tg-msft Thanks for your review! Updated according to your comments.

BillWagner (approved):
All comments have been addressed.  

docs/fsharp/using-fsharp-on-azure/queue-storage.md
```diff
@@ -109,60 +101,65 @@ you would keep a retry count as well, and if the message is retried more
...
 
 To delete a queue and all the messages contained in it, call the
 `Delete` method on the queue object.
 
-[!code-fsharp[QueueStorage](~/samples/snippets/fsharp/azure/queue-storage.fsx#L112-L113)]
+[!code-fsharp[QueueStorage](../../../samples/snippets/fsharp/azure/queue-storage.fsx#L99-L99)]
+
+## Note
+
+If you're migrating from the old libraries, they Base64 encoded messages by default but the new libraries do not because it's more performant.
```
BillWagner, line 157:
```suggestion
If you're migrating from the old libraries, they Base64 encoded messages by default but the new libraries do not because it's more performant.
```

PR #24475, opened 2021-06-01
Fix missing command argument

nycdotnet:
## Summary

The docs are missing the argument that tells it to try to upgrade.

Without this, it returns 

```
Required command was not provided.
Unrecognized command or argument '.\whateverproject.csproj'
```


BillWagner (approved):
Thanks for spotting this and fixing it @nycdotnet 

PR #24495, opened 2021-06-02
Add the upgrade command to the code sample

SibeeshVenu:
It never worked without providing the `upgrade` command to the existing code block.

## Summary

This is a simple fix for the code block given. 


BillWagner (approved):
Thanks for spotting this and fixing it @SibeeshVenu 

I'll :shipit: now.

PR #24494, opened 2021-06-02
change range to ID

BillWagner:
Fixes #24456

The snippet tags incorrectly used the "range" argument instead of the "ID" argument.

[Internal preview link](https://review.docs.microsoft.com/en-us/dotnet/csharp/fundamentals/types/namespaces?branch=pr-en-us-24494)

PR #24497, opened 2021-06-02
Fix incorrect snippet

Youssef1313:
Follow-up #24494


BillWagner (approved):
Thanks @Youssef1313 

I'll :shipit: now.

PR #24512, opened 2021-06-03
Fix text and diagram about surrogate code points

### LINKED ISSUE #24262: A Bit of Incorrect Information about Surrogate Pairs
In [this paragraph](https://docs.microsoft.com/en-us/dotnet/standard/base-types/character-encoding-introduction#surrogate-pairs), it states:
>The high surrogate code point is called "high" because it's used to calculate the higher-order 11 bits of the full 21-bit code point range. The low surrogate code point is used to calculate the lower-order 10 bits.  

However, there are only 20-bits because the surrogates can each only provide ten.  Of the 16 bits, the first 5 are always `11011` (which designate it as in the surrogate range), the next bit is 1 for the high surrogate code points and 0 for the low surrogate code points, and that only leaves 10 remaining bits for the actual data.

I couldn't figure out from where the numbers 21 or 11 came, so I [finally turned to Wikipedia](https://en.wikipedia.org/wiki/UTF-16#Code_points_from_U+010000_to_U+10FFFF), and they cleared things up (just a bit).

---
#### Document Details

* ID: a6cf7421-fbd2-0ca0-7385-4df65531fd66
* Version Independent ID: d53a0764-9f27-16b3-c40a-9627dd1bcc0d
* Content: [Introduction to character encoding in .NET](https://docs.microsoft.com/en-us/dotnet/standard/base-types/character-encoding-introduction)
* Content Source: [docs/standard/base-types/character-encoding-introduction.md](https://github.com/dotnet/docs/blob/main/docs/standard/base-types/character-encoding-introduction.md)
* GitHub Login: @adegeo
* Microsoft Alias: **adegeo**

tdykstra:
Fixes #24262

### DIFF
--- a/docs/standard/base-types/character-encoding-introduction.md
+++ b/docs/standard/base-types/character-encoding-introduction.md
@@ -148,7 +148,7 @@ code point = 65,536 +
   (low surrogate code point - 56,320)
 ```
 
-A *high* surrogate code point doesn't have a higher number value than a *low* surrogate code point. The high surrogate code point is called "high" because it's used to calculate the higher-order 11 bits of the full 21-bit code point range. The low surrogate code point is used to calculate the lower-order 10 bits.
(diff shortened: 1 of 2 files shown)


luisquintanilla (approved):
@tdykstra LGTM. :shipit:

PR #24518, opened 2021-06-04
CA2016 correct category

roshek:
## Summary

Small change to fix Category of CA2016 which seems to be incorrectly set to Performance instead of Reliability.


gewarren (approved):
Thank you @Roshek

PR #24527, opened 2021-06-05
Add information about type aliases for struct tuples (F#)

ForNeVeR:
Today, I was confused by type alias syntax around struct tuples, and found that the documentation could be improved by saying that the outer parentheses are required.


cartermp (approved):
Thanks! This is indeed required.

PR #24546, opened 2021-06-07
Add docs-verifier

Youssef1313:
[First release](https://github.com/MSDocsBuild/build-verifier/releases/tag/v0.0.1) of the docs-verifier (the extended action of markdown-links-verifier).

PR #24556, opened 2021-06-08
add redirection for missing error

BillWagner:
See #24546 for background.


Youssef1313 (commented):
We probably need to rename `cs2003.md` in this PR to see the effect of the validation when #24546 is merged.

BillWagner:
> We probably need to rename cs2003.md in this PR to see the effect of the validation when #24546 is merged.

Now that I see your latest commit, I'm wondering if we need this PR at all. Did I miss something?

I'm now seeing that you renamed "cs0003.md" to test, and then renamed it back. That would make this unnecessary.

Youssef1313:
@BillWagner In case you want to double check that the validation works, I think we can use this PR as a text after #24546 is merged.

BillWagner:
@Youssef1313 @gewarren 

This PR ran a few tests on the new workflow. It works well!  

Closing this PR.

PR #24564, opened 2021-06-08
Use Android API level instead of OS number in example

akoeplinger:
Android versions the APIs using a "level" concept which doesn't follow the version numbering scheme of the OS release.
We're using the API level for the SupportedOSPlatform annotations.


gewarren (approved):
Thank you @akoeplinger

PR #24618, opened 2021-06-09
Convert labeler to new format

adegeo:
Converts the old labeler tool config file into the new tool rule file.


BillWagner (approved):
This LGTM @adegeo

PR #24620, opened 2021-06-09
update guidance on finalizers

### LINKED ISSUE #17463: Finalizer is not reliably called regardless of .NET Core or .NET Framework
According to the page:
> In .NET Framework applications (but not in .NET Core applications), finalizers are also called when the program exits.

However, given the following code, it appears the finalizer doesn't reliably get called before the program shuts down regardless of whether .NET Core (netcoreapp3.1) or .NET Framework (net46) is used.  Adding the Collect() triggers the finalizer in both cases.

```c#
    class Program
    {
        static void Main(string[]? args)
        {
            DoSomething();
            // System.GC.Collect();
        }

        static void DoSomething()
        {
            Thing thing = new Thing();
        }
    }

    class Thing
    {
        ~Thing()
        {
            Console.WriteLine("Executing finalizer...");
        }
    }
```

Admittedly, there is likely some "timing" explanation but seemingly it isn't working as described.

---
#### Document Details

* ID: 26b6bf23-9cd5-a7c5-dc6a-f9b911d784ff
* Version Independent ID: 9f040450-27b7-7e4a-b908-c3695f6aaad1
* Content: [Finalizers - C# Programming Guide](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/destructors#feedback)
* Content Source: [docs/csharp/programming-guide/classes-and-structs/destructors.md](https://github.com/dotnet/docs/blob/master/docs/csharp/programming-guide/classes-and-structs/destructors.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixes #17463

This PR does the following:

- Uses the language from the C# standard to define if finalizers are called.
- copy edit the article. Mostly removing any use of "destructor" which is no longer the preferred term.
- Update samples so they don't show the practice of calling `GC.Collect()` and `GC.WaitForPendingFinalizers()`, which won't guarantee finalizers are called on .NET Core and .NET 5 and later.
- Link to `System.AppDomain.ProcessExit` as a way to be notified of app shutdown so cleanup can be performed.

### DIFF
--- a/docs/csharp/programming-guide/classes-and-structs/destructors.md
+++ b/docs/csharp/programming-guide/classes-and-structs/destructors.md
@@ -9,9 +9,7 @@ helpviewer_keywords:
 ---
 # Finalizers (C# Programming Guide)
 
-Finalizers (historically referred to as **destructors**) are used to perform any necessary final clean-up when a class instance is being collected by the garbage collector.
-
-In most cases, you can avoid writing a finalizer by using the  <xref:System.Runtime.InteropServices.SafeHandle?displayProperty=fullName> or derived classes to wrap any unmanaged handle.
+Finalizers (historically referred to as **destructors**) are used to perform any necessary final clean-up when a class instance is being collected by the garbage collector. In most cases, you can avoid writing a finalizer by using the  <xref:System.Runtime.InteropServices.SafeHandle?displayProperty=fullName> or derived classes to wrap any unmanaged handle.
 
 ## Remarks  
 
@@ -57,7 +55,7 @@ Whether or not finalizers are run as part of application termination is implemen
 > [!NOTE]
 > The .NET Framework implementation makes every reasonable effort to call finalizers for all of its objects that have not yet been garbage collected, unless such cleanup has been suppressed (by a call to the library method `GC.SuppressFinalize`, for example).
   
- If you need to perform cleanup reliably when an application exist, register a handler for the <xref:System.AppDomain.ProcessExit?displayProperty=fullName> event. That handler would ensure <xref:System.IDisposable.Dispose?displayProperty=nameWithType> has been called for all objects that require cleanup before application exit. Because you can't call *Finalize* directly, and you can't guarantee the garbage collector calls all finalizers before exit, you must use `Dispose` to ensure resources are freed.
+ If you need to perform cleanup reliably when an application exist, register a handler for the <xref:System.AppDomain.ProcessExit?displayProperty=fullName> event. That handler would ensure <xref:System.IDisposable.Dispose?displayProperty=nameWithType> (or, <xref:System.IAsyncDisposable.DisposeAsync?displayProperty=nameWithType>) has been called for all objects that require cleanup before application exit. Because you can't call *Finalize* directly, and you can't guarantee the garbage collector calls all finalizers before exit, you must use `Dispose` or `DisposeAsync` to ensure resources are freed.
 
 ## Using finalizers to release resources
 
@@ -71,6 +69,7 @@ For more information about cleaning up resources, see the following articles:
 
 - [Cleaning Up Unmanaged Resources](../../../standard/garbage-collection/unmanaged.md)
 - [Implementing a Dispose Method](../../../standard/garbage-collection/implementing-dispose.md)
+- [Implementing a Dispose Method](../../../standard/garbage-collection/implementing-disposeasync.md)
 - [using Statement](../../language-reference/keywords/using-statement.md)
 
 ## Example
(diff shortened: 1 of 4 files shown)


docs/csharp/programming-guide/classes-and-structs/destructors.md
```diff
@@ -1,85 +1,83 @@
...
   - "finalizers [C#]"
-ms.assetid: 1ae6e46d-a4b1-4a49-abe5-b97f53d9e049
 ---
 # Finalizers (C# Programming Guide)
 
-Finalizers (which are also called **destructors**) are used to perform any necessary final clean-up when a class instance is being collected by the garbage collector.  
-  
+Finalizers (historically referred to as **destructors**) are used to perform any necessary final clean-up when a class instance is being collected by the garbage collector.
+
+In most cases, you can avoid writing a finalizer by using the  <xref:System.Runtime.InteropServices.SafeHandle?displayProperty=fullName> or derived classes to wrap any unmanaged handle.
```
IEvangelist, line 14: The intro isn't a paragraph, it's two separate sentences. Do you think it makes sense to converge this a bit? Also, since we're calling out `SafeHandle` as an alternative approach, does it make sense to include that as an example?

docs/csharp/programming-guide/classes-and-structs/destructors.md
```diff
@@ -1,85 +1,83 @@
...
-- [Cleaning Up Unmanaged Resources](../../../standard/garbage-collection/unmanaged.md)  
-  
-- [Implementing a Dispose Method](../../../standard/garbage-collection/implementing-dispose.md)  
-  
-- [using Statement](../../language-reference/keywords/using-statement.md)  
+> [!NOTE]
+> The .NET Framework implementation makes every reasonable effort to call finalizers for all of its objects that have not yet been garbage collected, unless such cleanup has been suppressed (by a call to the library method `GC.SuppressFinalize`, for example).
   
-## Example  
+ If you need to perform cleanup reliably when an application exist, register a handler for the <xref:System.AppDomain.ProcessExit?displayProperty=fullName> event. That handler would ensure <xref:System.IDisposable.Dispose?displayProperty=nameWithType> has been called for all objects that require cleanup before application exit. Because you can't call *Finalize* directly, and you can't guarantee the garbage collector calls all finalizers before exit, you must use `Dispose` to ensure resources are freed.
```
IEvangelist, line 60: Does this also handle `IAsyncDisposable.DisposeAsync`, if not we should add that explicitly?

docs/csharp/programming-guide/classes-and-structs/destructors.md
```diff
@@ -1,85 +1,83 @@
...
+
+## Explicit release of resources
+
+If your application is using an expensive external resource, we also recommend that you provide a way to explicitly release the resource before the garbage collector frees the object. To release the resource, implement a `Dispose` method from the <xref:System.IDisposable> interface that performs the necessary cleanup for the object. This can considerably improve the performance of the application. Even with this explicit control over resources, the finalizer becomes a safeguard to clean up resources if the call to the `Dispose` method fails.
+
+For more information about cleaning up resources, see the following articles:  
+
+- [Cleaning Up Unmanaged Resources](../../../standard/garbage-collection/unmanaged.md)
+- [Implementing a Dispose Method](../../../standard/garbage-collection/implementing-dispose.md)
```
IEvangelist, line 73: We might want to also add _implementing-disposeasync.md_, if this is also applicable to `DisposeAsync`

IEvangelist (approved):
A few questions and comments

docs/csharp/programming-guide/classes-and-structs/destructors.md
BillWagner: I didn't add an example, because those explode quickly. `SafeHandle` has several derived classes for Files, network pipes, memory handles, and more. Depending on which kind of resource you're holding, you'd use different derived classes.

I think pointing readers to that class will help them find the correct derived class for their use.

docs/csharp/programming-guide/classes-and-structs/snippets/destructors/expr-bodied-destructor.cs
```diff
@@ -13,9 +13,10 @@ class Program
...
    {
-      var destroyer = new Destroyer();
-      destroyer = null;
-      GC.Collect();
-      Console.WriteLine("Exiting...");
-   }
+        var destroyer = new Destroyer();
+        destroyer = null;
+        GC.Collect();
+        GC.WaitForPendingFinalizers();
```
sharwell, line 19: 💡 `GC.GetTotalMemory` typically performs this operation in a manner that better meets user expectations.

```suggestion
        GC.GetTotalMemory(forceFullCollection: true);
```

PR #24665, opened 2021-06-14
Lowering the degree of recommendations

WhiteBlackGoose:
The conventions described in the document were taken from those from the runtime, but I think it might make sense to accept other common notations which make sense too.

I'm sorry if I missed the point of the document, but I think many developers would agree with me on lowering the level of those recommendations.

### DIFF
--- a/docs/csharp/fundamentals/coding-style/coding-conventions.md
+++ b/docs/csharp/fundamentals/coding-style/coding-conventions.md
@@ -96,7 +96,7 @@ For more information on positional records, see [Positional syntax for property 
 
 ### Camel case
 
-Use camel casing ("camelCasing") when naming `private` or `internal` fields, and prefix them with `_`.
+Use camel casing ("camelCasing") when naming `private` or `internal` fields. It is also common to prefix them with `_`.
 
 ```csharp
 public class DataService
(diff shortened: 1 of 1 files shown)


BillWagner:
ping @IEvangelist for your thoughts on this PR.

IEvangelist:
Thanks for pinging me @BillWagner.

Hi @WhiteBlackGoose - and thanks for posting this PR. I do not believe that this is a necessary or desirable. These conventions are what we strive to employ with in our docs and samples. There is an important notice immediately following the intro paragraph.

> The guidelines in this article are used by Microsoft to develop samples and documentation. They were adopted from the .NET Runtime, C# Coding Style guidelines. You can use them, or adapt them to your needs. The primary objectives are consistency and readability within your project, team, organization, or company source code.

The article details what we strive to do, you're free to take the conventions but we need to leave this language as is. As such, I'm going to close this PR. If you believe we should continue to discuss this, let's do so. Thank you

WhiteBlackGoose:
Hm. Yes, I saw that notice and I understand the reason why those conventions are as they are (because they were taken from the conventions used in the runtime). But that doesn't explain why they have to *remain* such, ignoring the common conventions used by .NET devs today.

It's more of a question, of course I don't insist on reopening the PR or anything. I'm curious.

PR #24677, opened 2021-06-15
If message is not null, get to the CustomMethod execution

Ioana37:
## Summary

If the logEntry.Exception is null, the method will simply exist without logging anything. Given that the normal case would be to have no exception, but to have a log message, I think it would be more valuable as a code sample if the execution could get to the "CustomLogicGoesHere" method. 
A similar code snippet is also present in the SystemdConsoleFormatter implementation (https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging.Console/src/SystemdConsoleFormatter.cs).

### DIFF
--- a/docs/core/extensions/snippets/logging/console-formatter-custom/CustomFormatter.cs
+++ b/docs/core/extensions/snippets/logging/console-formatter-custom/CustomFormatter.cs
@@ -26,7 +26,7 @@ namespace Console.ExampleFormatters.Custom
             IExternalScopeProvider scopeProvider,
             TextWriter textWriter)
         {
-            if (logEntry.Exception is null)
+            if (logEntry.Exception == null && message == null)
             {
                 return;
             }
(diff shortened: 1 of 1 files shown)


docs/core/extensions/snippets/logging/console-formatter-custom/CustomFormatter.cs
```diff
@@ -26,7 +26,7 @@ public CustomFormatter(IOptionsMonitor<CustomOptions> options)
             IExternalScopeProvider scopeProvider,
             TextWriter textWriter)
         {
-            if (logEntry.Exception is null)
+            if (logEntry.Exception == null && message == null)
```
IEvangelist, line 29:
```suggestion
            if (logEntry.Exception is null && message is null)
```

IEvangelist (approved):

IEvangelist:
Originally, when I read just the snippet change - it made sense, but now re-reading this -- it's invalid to check the `message` variable for `null` as it's not yet declared. I'm going to close this PR. Thanks

Ioana37:
Thanks for spotting the issue! I've fixed it and created another PR #24761. Please review it whenever convenient :)

PR #24680, opened 2021-06-15
Extra 1 character (typo)

jgscherber:
## Summary

There was an extra one (1) on the end of the line (typo). It might be a broken reference, but I didn't see anywhere below that it might have been linking to.


gewarren (approved):
Thank you @jgscherber

PR #24738, opened 2021-06-18
Update .repoman.yml config for metadata field

adegeo:
@gewarren


gewarren (approved):
Thanks, I had been noticing that the label wasn't applied either. It's not getting added to the project anymore either.

PR #24740, opened 2021-06-18
Recommend setting publish options in project file

sbomer:
The `PublishTrimmed` option also disables trim-incompatible features, and turns on a Roslyn analyzer for trim compatibility, so it should be set in the project file to ensure these settings apply during a `dotnet build`.

Similarly, `PublishSingleFile` enables a single-file analyzer so it should also be set in the project file.

Addresses part of https://github.com/mono/linker/issues/2085.

### DIFF
--- a/docs/core/deploying/single-file.md
+++ b/docs/core/deploying/single-file.md
@@ -136,29 +136,27 @@ These properties have the following functions:
 
 ## Publish a single file app - CLI
 
-Publish a single file application using the [dotnet publish](../tools/dotnet-publish.md) command. Set the following properties in your project file, then publish the app using `dotnet publish`.
+Publish a single file application using the [dotnet publish](../tools/dotnet-publish.md) command.
 
-- Publish for a specific runtime: `<RuntimeIdentifier>win-x64</RuntimeIdentifier>`
-- Publish as a single-file: `<PublishSingleFile>true</PublishSingleFile>`
+01. Add `<PublishSingleFile>true</PublishSingleFile>` to your project file.
 
-The following example publishes an app for Windows as a self-contained single file application.
+    This will produce a single-file app on self-contained publish. It also shows single-file compatibility warnings during build.
 
-```xml
-<PropertyGroup>
-    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
-    <PublishSingleFile>true</PublishSingleFile>
-</PropertyGroup>
-```
+    ```xml
+    <PropertyGroup>
+        <PublishSingleFile>true</PublishSingleFile>
+    </PropertyGroup>
+    ```
 
-The following example publishes an app for Linux as a framework dependent single file application.
+01. Publish the app as for a specific runtime identifier using `dotnet publish -r <RID>`
 
-```xml
-<PropertyGroup>
-    <RuntimeIdentifier>linux-x64</RuntimeIdentifier>
-    <PublishSingleFile>true</PublishSingleFile>
-    <SelfContained>false</SelfContained>
-</PropertyGroup>
-```
+    The following example publishes the app for Windows as a self-contained single file application.
+
+    `dotnet publish -r win-x64`
+
+    The following example publishes the app for Linux as a framework dependent single file application.
+
+    `dotnet publish -r linux-x64 --self-contained false`
 
 `<PublishSingleFile>` should be set in the project file to enable single-file analysis during build, but it is also possible to pass these options as `dotnet publish` arguments:
 
--- a/docs/core/deploying/single-file.md
+++ b/docs/core/deploying/single-file.md
@@ -125,7 +125,7 @@ These properties have the following functions:
 
 * `PublishSingleFile` - Enables single-file publishing. Also enables single-file warnings during `dotnet build`.
 * `SelfContained` - Determines whether the app will be self-contained or framework-dependent.
-* `RuntimeIdentifier` - Specifies the [OS and CPU type](../rid-catalog.md) you are targeting.
+* `RuntimeIdentifier` - Specifies the [OS and CPU type](../rid-catalog.md) you are targeting. Also sets `<SelfContained>true</SelfContained>` by default.
 * `PublishTrimmed` - Enables use of [assembly trimming](trim-self-contained.md), which is only supported for self-contained apps.
 * `PublishReadyToRun` - Enables [ahead-of-time (AOT) compilation](ready-to-run.md).
 
@@ -136,9 +136,9 @@ These properties have the following functions:
 
 ## Publish a single file app - CLI
 
-Publish a single file application using the [dotnet publish](../tools/dotnet-publish.md) command. When you publish your app, set the following properties:
+Publish a single file application using the [dotnet publish](../tools/dotnet-publish.md) command. Set the following properties in your project file, then publish the app using `dotnet publish`.
 
-- Publish for a specific runtime: `-r win-x64`
+- Publish for a specific runtime: `<RuntimeIdentifier>win-x64</RuntimeIdentifier>`
 - Publish as a single-file: `<PublishSingleFile>true</PublishSingleFile>`
 
 The following example publishes an app for Windows as a self-contained single file application.
@@ -147,7 +147,6 @@ The following example publishes an app for Windows as a self-contained single fi
 <PropertyGroup>
     <RuntimeIdentifier>win-x64</RuntimeIdentifier>
     <PublishSingleFile>true</PublishSingleFile>
-    <SelfContained>true</SelfContained>
 </PropertyGroup>
 ```
 
@@ -155,10 +154,16 @@ The following example publishes an app for Linux as a framework dependent single
 
 ```xml
 <PropertyGroup>
-    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
+    <RuntimeIdentifier>linux-x64</RuntimeIdentifier>
     <PublishSingleFile>true</PublishSingleFile>
     <SelfContained>false</SelfContained>
 </PropertyGroup>
+```
+
+`<PublishSingleFile>` should be set in the project file to enable single-file analysis during build, but it is also possible to pass these options as `dotnet publish` arguments:
+
+```dotnetcli
+dotnet publish -r linux-x64 -p:PublishSingleFile=true --self-contained false
 ```
 
 For more information, see [Publish .NET Core apps with .NET Core CLI](deploy-with-cli.md).
--- a/docs/core/deploying/single-file.md
+++ b/docs/core/deploying/single-file.md
@@ -123,7 +123,7 @@ Here's a sample project file that specifies single-file publishing:
 
 These properties have the following functions:
 
-* `PublishSingleFile` - Enables single-file publishing.
+* `PublishSingleFile` - Enables single-file publishing. Also enables single-file warnings during `dotnet build`.
 * `SelfContained` - Determines whether the app will be self-contained or framework-dependent.
 * `RuntimeIdentifier` - Specifies the [OS and CPU type](../rid-catalog.md) you are targeting.
 * `PublishTrimmed` - Enables use of [assembly trimming](trim-self-contained.md), which is only supported for self-contained apps.
@@ -139,18 +139,26 @@ These properties have the following functions:
 Publish a single file application using the [dotnet publish](../tools/dotnet-publish.md) command. When you publish your app, set the following properties:
 
 - Publish for a specific runtime: `-r win-x64`
-- Publish as a single-file: `-p:PublishSingleFile=true`
+- Publish as a single-file: `<PublishSingleFile>true</PublishSingleFile>`
 
 The following example publishes an app for Windows as a self-contained single file application.
 
-```dotnetcli
-dotnet publish -r win-x64 -p:PublishSingleFile=true --self-contained true
+```xml
+<PropertyGroup>
+    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
+    <PublishSingleFile>true</PublishSingleFile>
+    <SelfContained>true</SelfContained>
+</PropertyGroup>
 ```
 
 The following example publishes an app for Linux as a framework dependent single file application.
 
-```dotnetcli
-dotnet publish -r linux-x64 -p:PublishSingleFile=true --self-contained false
+```xml
+<PropertyGroup>
+    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
+    <PublishSingleFile>true</PublishSingleFile>
+    <SelfContained>false</SelfContained>
+</PropertyGroup>
 ```
 
 For more information, see [Publish .NET Core apps with .NET Core CLI](deploy-with-cli.md).
@@ -158,6 +166,8 @@ For more information, see [Publish .NET Core apps with .NET Core CLI](deploy-wit
 ## Publish a single file app - Visual Studio
 
 Visual Studio creates reusable publishing profiles that control how your application is published.
+
+01. Add `<PublishSingleFile>true</PublishSingleFile>` to your project file.
 
 01. On the **Solution Explorer** pane, right-click on the project you want to publish. Select **Publish**.
 
--- a/docs/core/deploying/trim-self-contained.md
+++ b/docs/core/deploying/trim-self-contained.md
@@ -71,25 +71,30 @@ If we're using .Net 5.0, we can update our project file to include the following
 
 ## Trim your app - CLI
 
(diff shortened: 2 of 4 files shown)


docs/core/deploying/trimming-options.md
```diff
@@ -16,10 +16,14 @@ Trimming with `PublishTrimmed` was introduced in .NET Core 3.0. The other option
...
 - `<PublishTrimmed>true</PublishTrimmed>`
 
-   Enable trimming during publish, with the default settings defined by the SDK.
+   Enable trimming during publish. This also turns off trim-incompatible features and enables [trim analysis](#roslyn-analyzer) during build.
+
+Place this in the project file to ensure that the setting applies during `dotnet build`, not just `dotnet publish`.
 
 This will trim any assemblies which have been configured for trimming. With `Microsoft.NET.Sdk` in .NET 6, this includes the any assemblies with `[AssemblyMetadata("IsTrimmable", "True")]`, which is the case for framework assemblies. In .NET 5, framework assemblies from the netcoreapp runtime pack are configured for trimming via `<IsTrimmable>` MSBuild metadata. Other SDKs may define different defaults.
 
+Starting in .NET 6, this setting also enables the trim compatibility [Roslyn analyzer](#roslyn-analyzer), and disables [features](#trimming-framework-library-features) which are incompatible with trimming.
```
LakshanF, line 25: Should we update the features list to include the new ones like BuiltInComInteropSupport, EnableCppCLIHostActivation etc.?

LakshanF (approved):

docs/core/deploying/trimming-options.md
sbomer: I don't think we should document the ones that are already disabled by default when linking - we don't want to encourage people to enable the unsupported ones.

But it looks like `UseNativeHttpHandler` is missing from the list. @eerhardt do you think we should document that one? It looks to me like `UseNativeHttpHandler=true` will prevent some (but not all) references to `HttpClientHandler` - I'm not sure whether this feature was designed for trimming in particular.

docs/core/tools/dotnet-publish.md
```diff
@@ -153,13 +153,13 @@ For more information, see the following resources:
 
   For more information about single-file publishing, see the [single-file bundler design document](https://github.com/dotnet/designs/blob/main/accepted/2020/single-file/design.md).
 
-  We recommend that you specify this option in a publish profile rather than on the command line. For more information, see [MSBuild](#msbuild).
+  We recommend that you specify this option in the project file rather than on the command line. Setting it in the project file enables a Roslyn analyzer that provides single-file compatibility warnings during `dotnet build`. For more information, see [MSBuild](#msbuild).
 
 - **`-p:PublishTrimmed=true`**
```
agocke, line 158:
```suggestion
- **`<PublishTrimmed>true</PublishTrimmed>`**
```

agocke: Same for `PublishSingleFile` above

sbomer: Unfortunately this document is talking about options for the `dotnet publish` command. Should we maybe just remove these sections from this file? It's strange that they're documented as publish options when really they're just passed to MSBuild. And these same docs then recommend _not_ using the command-line arg version.

agocke: Removing them seems best, although the unfortunate benefit is that if you specific `PublishTrimmed` on the command line then it applies to all projects, and would turn on the analyzer for all of them. I suppose we also need to recommend adding `EnableTrimmingAnalyzer` in every project.

docs/core/deploying/single-file.md
```diff
@@ -139,18 +139,26 @@ These properties have the following functions:
...
 
 The following example publishes an app for Windows as a self-contained single file application.
 
-```dotnetcli
-dotnet publish -r win-x64 -p:PublishSingleFile=true --self-contained true
+```xml
+<PropertyGroup>
+    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
+    <PublishSingleFile>true</PublishSingleFile>
+    <SelfContained>true</SelfContained>
```
eerhardt, line 150: You don't need to set `SelfContained=true` explicitly. It is defaulted that way when you set `RuntimeIdentifier=foo`.

https://github.com/dotnet/sdk/blob/c53c8ac81e1bcb8118c31bc4f5622396c72e9959/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets#L107-L115

docs/core/deploying/single-file.md
```diff
@@ -139,18 +139,26 @@ These properties have the following functions:
 Publish a single file application using the [dotnet publish](../tools/dotnet-publish.md) command. When you publish your app, set the following properties:
 
 - Publish for a specific runtime: `-r win-x64`
```
eerhardt, line 141: It's a little odd that we are mixing `-r` (the command line option) here and the project setting for `PublishSingleFile` in the next line. I think it would make sense to be consistent here.

docs/core/deploying/single-file.md
```diff
@@ -139,18 +139,26 @@ These properties have the following functions:
...
-- Publish as a single-file: `-p:PublishSingleFile=true`
+- Publish as a single-file: `<PublishSingleFile>true</PublishSingleFile>`
 
 The following example publishes an app for Windows as a self-contained single file application.
 
-```dotnetcli
-dotnet publish -r win-x64 -p:PublishSingleFile=true --self-contained true
```
eerhardt, line 147: Maybe we should list both the XML project settings and the command line options, and then explain the advantages to setting the XML project settings?

My concern here is that new developers could be confused that they can't set these options on the command line (especially the `RuntimeIdentifier`).

docs/core/deploying/trimming-options.md
eerhardt: I do find it a bit confusing that we are explicitly saying `disables features [link] which are incompatible with trimming`, but then the link doesn't list any features which are incompatible with trimming.

Maybe we should have 2 sections? Features that can be toggled when trimming, and features that are not compatible with trimming?

> But it looks like UseNativeHttpHandler is missing from the list.

@EgorBo @akoeplinger @steveisok @marek-safar - should we be documenting `UseNativeHttpHandler` here? I believe it is only valid for Android and iOS apps.

docs/core/deploying/trimming-options.md
```diff
@@ -16,10 +16,14 @@ Trimming with `PublishTrimmed` was introduced in .NET Core 3.0. The other option
 
 - `<PublishTrimmed>true</PublishTrimmed>`
 
-   Enable trimming during publish, with the default settings defined by the SDK.
+   Enable trimming during publish. This also turns off trim-incompatible features and enables [trim analysis](#roslyn-analyzer) during build.
+
+Place this in the project file to ensure that the setting applies during `dotnet build`, not just `dotnet publish`.
 
 This will trim any assemblies which have been configured for trimming. With `Microsoft.NET.Sdk` in .NET 6, this includes the any assemblies with `[AssemblyMetadata("IsTrimmable", "True")]`, which is the case for framework assemblies. In .NET 5, framework assemblies from the netcoreapp runtime pack are configured for trimming via `<IsTrimmable>` MSBuild metadata. Other SDKs may define different defaults.
```
tdykstra, line 23:
```suggestion
This will trim any assemblies that have been configured for trimming. With `Microsoft.NET.Sdk` in .NET 6, this includes the any assemblies with `[AssemblyMetadata("IsTrimmable", "True")]`, which is the case for framework assemblies. In .NET 5, framework assemblies from the netcoreapp runtime pack are configured for trimming via `<IsTrimmable>` MSBuild metadata. Other SDKs may define different defaults.
```

docs/core/deploying/trimming-options.md
```diff
@@ -16,10 +16,14 @@ Trimming with `PublishTrimmed` was introduced in .NET Core 3.0. The other option
...
 
+Starting in .NET 6, this setting also enables the trim compatibility [Roslyn analyzer](#roslyn-analyzer), and disables [features](#trimming-framework-library-features) which are incompatible with trimming.
+
 ## Trimming granularity
 
 The following granularity settings control how aggressively unused IL is discarded. This can be set as a property affecting all trimmer input assemblies, or as metadata on an [individual assembly](#trimmed-assemblies) which overrides the property setting.
```
tdykstra, line 29:
```suggestion
The following granularity settings control how aggressively unused IL is discarded. This can be set as a property affecting all trimmer input assemblies, or as metadata on an [individual assembly](#trimmed-assemblies), which overrides the property setting.
```

tdykstra (approved):
LGTM, a couple suggestions

docs/core/tools/dotnet-publish.md
sbomer: I removed PublishTrimmed, PublishSingleFile, and PublishReadyToRun. @tdykstra does this seem ok to you?

docs/core/deploying/trimming-options.md
sbomer: I added  a section for the trim-incompatible features, and also added a few more switches to the optional features section, including UseNativeHttpHandler.

docs/core/deploying/trim-self-contained.md
```diff
@@ -73,8 +73,8 @@ If we're using .Net 5.0, we can update our project file to include the following
 
 Trim your application using the [dotnet publish](../tools/dotnet-publish.md) command. When you publish your app, set the following properties:
```
eerhardt, line 74: Should this line be updated? It makes it sound like I should only change my project when I pubilsh.

```suggestion
Trim your application using the [dotnet publish](../tools/dotnet-publish.md) command. Set the following properties in your project file, then publish the app using `dotnet publish`.
```

docs/core/deploying/trim-self-contained.md
```diff
@@ -73,8 +73,8 @@ If we're using .Net 5.0, we can update our project file to include the following
...
 
-- Publish as a self-contained app for a specific runtime: `-r win-x64`
-- Enable trimming: `/p:PublishTrimmed=true`
+- Publish as a self-contained app for a specific runtime: `<RuntimeIdentifier>win-x64</RuntimeIdentifier>`
```
eerhardt, line 76: Generally, we've been telling people not to hard-code RIDs into their project file, and instead pass them on the command line / publish profile.

cc @dsplaisted 

Are we telling people to put the RID in the project now because without it, `PublishTrimmed` breaks `dotnet build`? If yes, maybe we should move the validation to the `publish` command that ensures you are publishing SelfContained if `PublishTrimmed=true`.

docs/core/tools/dotnet-publish.md
```diff
@@ -139,28 +138,6 @@ For more information, see the following resources:
 
     If you specify a relative path when publishing a solution, each project's output goes into a separate folder relative to the project file location. If you specify an absolute path when publishing a solution, all publish output for all projects goes into the specified folder.
 
-- **`-p:PublishReadyToRun=true`**
```
eerhardt, line 142: Are we concerned that removing these options here decreases the chances of people finding them and learning about what they do? I think it is discoverable to list these publish options in the "dotnet publish" document, so people discover them.

docs/core/deploying/trimming-options.md
```diff
@@ -197,8 +205,43 @@ Several feature areas of the framework libraries come with linker directives tha
...
+
+   Use of custom resource types is not supported. ResourceManager code paths that use reflection for custom resource types is trimmed.
+
+- `<EnableCppCLIHostActivation>`
+
+    C++/CLI host activation is disabled.
+
+- `<EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization>`
+
+   [`DesigntimeLicenceContextSerializer`](https://docs.microsoft.com/dotnet/api/system.componentmodel.design.designtimelicensecontextserializer?view=net-5.0) use of BinaryFormatter serialization is dissabeld.
```
eerhardt, line 243:
```suggestion
   [`DesigntimeLicenceContextSerializer`](https://docs.microsoft.com/dotnet/api/system.componentmodel.design.designtimelicensecontextserializer?view=net-5.0) use of BinaryFormatter serialization is disabled.
```

docs/core/deploying/trimming-options.md
```diff
@@ -197,8 +205,43 @@ Several feature areas of the framework libraries come with linker directives tha
...
+
+- `<CustomResourceTypesSupport>`
+
+   Use of custom resource types is not supported. ResourceManager code paths that use reflection for custom resource types is trimmed.
+
+- `<EnableCppCLIHostActivation>`
+
+    C++/CLI host activation is disabled.
+
+- `<EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization>`
```
eerhardt, line 241: Hmm, it is interesting that `EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization` is disabled by default when you trim, but `EnableUnsafeBinaryFormatterSerialization` isn't.

Should `EnableUnsafeBinaryFormatterSerialization` also be set to `false` by default when you trim?

cc @GrabYourPitchforks @pgovind

docs/core/deploying/trim-self-contained.md
sbomer: No, `PublishTrimmed` doesn't break build - I was just changing this for consistency with the line below to address https://github.com/dotnet/docs/pull/24740#discussion_r655498139.  I'll change it back to recommending `-r win-x64` and `<PublishTrimmed>true</PublishTrimmed>` but being more explicit about what should go in the project file.

docs/core/tools/dotnet-publish.md
sbomer: I'm slightly concerned, but removed it based on the discussion in https://github.com/dotnet/docs/pull/24740#discussion_r654626080. We do have dedicated documentation for these properties. Maybe the best thing would be to link to the important publish MSBuild options from the publish options doc, without mentioning them as command-line arguments? cc @agocke

eerhardt: > link to the important publish MSBuild options from the publish options doc, without mentioning them as command-line arguments?

I like that approach. That way they are still discoverable.

docs/core/deploying/trimming-options.md
eerhardt: I've re-opened https://github.com/dotnet/sdk/issues/14475 for this issue.

docs/core/tools/dotnet-publish.md
```diff
@@ -57,6 +56,30 @@ dotnet publish -p:PublishProfile=FolderProfile
 
 The preceding example uses the *FolderProfile.pubxml* file that is found in the *\<project_folder>/Properties/PublishProfiles* folder. If you specify a path and file extension when setting the `PublishProfile` property, they are ignored. MSBuild by default looks in the *Properties/PublishProfiles* folder and assumes the *pubxml* file extension. To specify the path and filename including extension, set the `PublishProfileFullPath` property instead of the `PublishProfile` property.
 
+The following MSBuild properties change the output of `dotnet publish`.
+
+- `PublishReadyToRun`
+
+  Compiles application assemblies as ReadyToRun (R2R) format. R2R is a form of ahead-of-time (AOT) compilation. For more information, see [ReadyToRun images](../deploying/ready-to-run.md). Available since .NET Core 3.0 SDK.
+
+  To see warnings about missing dependencies that could cause runtime failures, use `PublishReadyToRunShowWarning`.
```
eerhardt, line 65: `PublishReadyToRunShowWarning=true` ?

docs/core/tools/dotnet-publish.md
```diff
@@ -57,6 +56,30 @@ dotnet publish -p:PublishProfile=FolderProfile
...
+
+  We recommend that you specify this option in a publish profile rather than on the command line.
```
eerhardt, line 67: "this option" is a bit ambiguous between `PublishReadyToRun` and `PublishReadyToRunShowWarning` here.

docs/core/tools/dotnet-publish.md
```diff
@@ -57,6 +56,30 @@ dotnet publish -p:PublishProfile=FolderProfile
...
+
+- `PublishSingleFile`
+
+  Packages the app into a platform-specific single-file executable. The executable is self-extracting and contains all dependencies (including native) that are required to run the app. When the app is first run, the application is extracted to a directory based on the app name and build identifier. Startup is faster when the application is run again. The application doesn't need to extract itself a second time unless a new version is used. Available since .NET Core 3.0 SDK.
```
eerhardt, line 71: > When the app is first run, the application is extracted to a directory based on the app name and build identifier

I know this was already there, but this is no longer the case for 5.0+.

eerhardt (approved):
Looks great. Thanks for making these changes, @sbomer!

PR #24754, opened 2021-06-21
Update publish-subscribe.md

DarqueWarrior:
Corrected typo in the IEventBus interface.

## Summary

TEvent was not used in the definition of the Subscribe method. Based on the code TEvent was supposed to be T. You could also change the other to T's to TEvent as well.


IEvangelist (approved):

PR #24761, opened 2021-06-21
If message not null, execute CustomMethod

Ioana37:
If the logEntry.Exception is null, the method will simply exit without logging anything. Given that the normal case would be to have no exception, but to have a log message, I think it would be more valuable as a code sample if the execution could get to the "CustomLogicGoesHere" method.
A similar code snippet is also present in the SystemdConsoleFormatter implementation (https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging.Console/src/SystemdConsoleFormatter.cs).

## Summary

Describe your changes here.

### DIFF
--- a/docs/core/extensions/snippets/logging/console-formatter-custom/CustomFormatter.cs
+++ b/docs/core/extensions/snippets/logging/console-formatter-custom/CustomFormatter.cs
@@ -29,12 +29,12 @@ namespace Console.ExampleFormatters.Custom
             string message =
                 logEntry.Formatter(
                     logEntry.State, logEntry.Exception);
-    
+
             if (message == null)
             {
                 return;
             }
-            
+
             CustomLogicGoesHere(textWriter);
             textWriter.WriteLine(message);
         }
--- a/docs/core/extensions/snippets/logging/console-formatter-custom/CustomFormatter.cs
+++ b/docs/core/extensions/snippets/logging/console-formatter-custom/CustomFormatter.cs
@@ -25,9 +25,12 @@ namespace Console.ExampleFormatters.Custom
             in LogEntry<TState> logEntry,
             IExternalScopeProvider scopeProvider,
             TextWriter textWriter)
-        {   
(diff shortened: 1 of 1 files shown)


docs/core/extensions/snippets/logging/console-formatter-custom/CustomFormatter.cs
```diff
@@ -25,17 +25,9 @@ public CustomFormatter(IOptionsMonitor<CustomOptions> options)
...
-            }
-
-            string message =
-                logEntry.Formatter(
-                    logEntry.State, logEntry.Exception);
-            
-            if (message == null)
+        {   
+            string message = logEntry.Formatter(logEntry.State, logEntry.Exception);
+            if (logEntry.Exception == null && message == null)
```
IEvangelist, line 30: I disagree with this proposed change.

If the `logEntry.Exception` is `null`, we should not be calling:

```csharp
logEntry.Formatter(logEntry.State, logEntry.Exception);
```

1. The evaluation is unnecessary as the method will `return`
2. It's added overhead (even though minimal)
3. It knowingly is passing `null` to the `.Formatter` method

IEvangelist (changes requested):
Thoughts on this?

```csharp
public override void Write<TState>(
    in LogEntry<TState> logEntry,
    IExternalScopeProvider scopeProvider,
    TextWriter textWriter)
{
    string message =
        logEntry.Formatter(
            logEntry.State, logEntry.Exception);
    
    if (message == null)
    {
        return;
    }
    
    CustomLogicGoesHere(textWriter);
    textWriter.WriteLine(message);
}
```

docs/core/extensions/snippets/logging/console-formatter-custom/CustomFormatter.cs
Ioana37: Looks good! I have committed the suggested changes. Thanks for the review :)

IEvangelist (approved):

PR #24775, opened 2021-06-22
Update property.md

### LINKED ISSUE #24193: Structs and records as indexer parameter type
**Help us make content visible**

I found the following guideline for [Indexed Property Design](https://github.com/dotnet/docs/blob/main/docs/standard/design-guidelines/property.md).

The guideline states the following: AVOID indexers with parameter types other than System.Int32, System.Int64, System.String, System.Object, or an enum.

So my question is what about structures and records? Should those types also be avoided or if not, should they be added to the list in the guideline?

IEvangelist:
## Summary

Updated the indexer parameter type avoidance recommendations.

Fixes #24193

[Internal preview](https://review.docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property?branch=pr-en-us-24775)

### DIFF
--- a/docs/standard/design-guidelines/property.md
+++ b/docs/standard/design-guidelines/property.md
@@ -1,7 +1,7 @@
 ---
 description: "Learn more about: Property Design"
 title: "Property Design"
-ms.date: "10/22/2008"
+ms.date: "6/22/2021"
 helpviewer_keywords:
   - "member design guidelines, properties"
   - "properties [.NET Framework], design guidelines"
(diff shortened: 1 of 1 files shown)


Youssef1313 (commented):
If I recall correctly, this is a copyrighted content that PRs used to always be refused in (even if PR was correct). Has that changed?

IEvangelist:
> If I recall correctly, this is a copyrighted content that PRs used to always be refused in (even if PR was correct). Has that changed?

You're probably right, I do see copyright details at the bottom of the article. I'll talk this over with the team today, maybe it's best to archive all these guides as they're extremely dated - and just point to the book itself.

PR #24786, opened 2021-06-23
Swap param order when using ArgumentException

Foda:
## Summary

The order of the parameters is wrong for the use of `ArgumentException` . The message should be first, followed by the name of the parameter, [as documented here](https://docs.microsoft.com/en-us/dotnet/api/system.argumentexception.-ctor?view=net-5.0#System_ArgumentException__ctor_System_String_System_String_).

Oddly enough, [the order used for the constructor](https://docs.microsoft.com/en-us/dotnet/api/system.argumentnullexception.-ctor?view=net-5.0#System_ArgumentNullException__ctor_System_String_System_String_) for `ArgumentNullException` is the opposite of `ArgumentException`!


BillWagner (approved):
Thanks for spotting this and fixing it @Foda 

I'll :shipit: now.

PR #24789, opened 2021-06-23
Update implement-value-objects.md

vkhorikov:
Additional reference for the base value object class.

### DIFF
--- a/docs/architecture/microservices/microservice-ddd-cqrs-patterns/implement-value-objects.md
+++ b/docs/architecture/microservices/microservice-ddd-cqrs-patterns/implement-value-objects.md
@@ -305,11 +305,11 @@ public class Address
 - **Complex types and/or value objects**. Discussion in the EF Core GitHub repo (Issues tab) \
   <https://github.com/dotnet/efcore/issues/246>
 
-- **Value Object: a better implementation** An article describing the base value object class in detail. \
-  <https://enterprisecraftsmanship.com/posts/value-object-better-implementation>
-
 - **ValueObject.cs.** Base value object class in eShopOnContainers. \
   <https://github.com/dotnet-architecture/eShopOnContainers/blob/dev/src/Services/Ordering/Ordering.Domain/SeedWork/ValueObject.cs>
+
+- **ValueObject.cs.** Base value object class in CSharpFunctionalExtensions. \
+  <https://github.com/vkhorikov/CSharpFunctionalExtensions/blob/master/CSharpFunctionalExtensions/ValueObject/ValueObject.cs>
 
 - **Address class.** Sample value object class in eShopOnContainers. \
   <https://github.com/dotnet-architecture/eShopOnContainers/blob/dev/src/Services/Ordering/Ordering.Domain/AggregatesModel/OrderAggregate/Address.cs>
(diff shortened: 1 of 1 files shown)


IEvangelist:
Hi @vkhorikov - I cannot accept this PR. While this implementation of a Value Object seems like a good one, the following reasons are standing in the way:

- The site that it links to immediately starts pushing for subscription, and sign up
- The implementation differs nearly entirely from the one covered in the article, thus deemphasizing it's relevancy dramatically

I would be okay with adding a link to the source code itself, like some of the other examples there?

vkhorikov:
> The implementation differs nearly entirely from the one covered in the article, thus deemphasizing it's relevancy dramatically

Do you mean the difference between the article and the [implementation in the CSharpFunctionalExtensions library](https://github.com/vkhorikov/CSharpFunctionalExtensions/blob/master/CSharpFunctionalExtensions/ValueObject/ValueObject.cs#L62)? I wouldn't say it's drastically different. Over the years, there were added some minor improvements, such as:

* Check for lazy loaded proxies
* Hash code caching
* `IComparable` implementation in base class

But fair enough, a link to https://github.com/vkhorikov/CSharpFunctionalExtensions/blob/master/CSharpFunctionalExtensions/ValueObject/ValueObject.cs#L62 is good too.

> But I'd want it listed under the one the others as they take precedence based on relevancy -- is that fair, and does that make sense?

How is it less relevant? AFAIK, eShopOnContainers's version was merged years after the article and it's not as complete as the version from CSharpFunctionalExtensions (e.g. it doesn't have `IComparable` implementation).

IEvangelist:
> How is it less relevant? AFAIK, eShopOnContainers's version was merged years after the article and it's not as complete as the version from CSharpFunctionalExtensions (e.g. it doesn't have `IComparable` implementation).

The article is part of eShopOnContainers's, the implementation exists in eShopOnContainers's, your implementation is not - though it is more complete, and probably better. It might make sense to have PR into the eShopOnContainers's repo and propose updating the impl to more closely align with yours -- what do you think?

vkhorikov:
Ok, I see what you mean. The ordering doesn't matter (to me), feel free to put it where it makes the most sense. Though, could you make the link https://github.com/vkhorikov/CSharpFunctionalExtensions/blob/master/CSharpFunctionalExtensions/ValueObject/ValueObject.cs (without the reference to a specific line)? I'll be moving the `ValueObject<T>` version to a separate file, the fact that both version live in the same file has been the source of confusion. I can make an update to this PR if you'd like.

IEvangelist:
> Ok, I see what you mean. The ordering doesn't matter (to me), feel free to put it where it makes the most sense. Though, could you make the link https://github.com/vkhorikov/CSharpFunctionalExtensions/blob/master/CSharpFunctionalExtensions/ValueObject/ValueObject.cs (without the reference to a specific line)? I'll be moving the `ValueObject<T>` version to a separate file, the fact that both version live in the same file has been the source of confusion. I can make an update to this PR if you'd like.

Yes, please update the PR. Thank you

vkhorikov:
@IEvangelist Updated the PR. Please take a look.

IEvangelist (approved):

vkhorikov:
Thanks much!

PR #24790, opened 2021-06-23
Update eShopOnContainers images to latest SPA theme

mvelosop:
## Summary

Minor changes for updating to latest eShopOnContainers SPA looks

(This is part of the PO to update the SPA application)


Youssef1313 (changes requested):
Image size (both old and new) is too large.

I converted it to `jpg` (same resolution), and it's reduced from 1.67MB to only 170KB. I recommend using the jpg version:

eshoponcontainers-sample-app-screenshot

IEvangelist (approved):

PR #24793, opened 2021-06-23
Enable CA2208 as error

Youssef1313:
This should prevent mistakes like the one fixed in https://github.com/dotnet/docs/pull/24786 when a sample is being edited.


Youssef1313:
@adegeo Snippet5000 should run when changing editorconfig as it may cause extra errors.

IEvangelist (approved):

adegeo:
@Youssef1313 We don't have a repo-wide version of this yet. We would probably do a one time report versus auto run. Having this auto run on the repo when this file changes would probably cause a lot of noise. Since we just fix projects as we work on them, things get caught eventually.

PR #24797, opened 2021-06-23
Update MinimalEventCounterSource.cs

VisualBean:
## Summary
`StopWatch.TotalMiliseconds` outputs a long - Which is what you would typically use in this case.
Also `WriteEvent` has a specific overload for `(int, int64, string)`, which is significantly faster than the `(int, Params Object[])` args one that was used in this snippet.


IEvangelist (approved):

PR #24799, opened 2021-06-23
Argument coercion breaking change

### LINKED ISSUE #24342: Add coercion for call arguments according to ECMA-335.
## Add coercion for call arguments according to ECMA-335.

ECMA-335 documents which implicit conversions are supports for call arguments, add checking for it (Table III.9: Signature Matching). 

### Version introduced

.NET 6.0

### Old behavior

It was leading to undefined behavior on some platforms, for example, passing long as int register on x86 was leaving the register undefined. Some conversions were working without issues, for example, the same long->int on 64-bit platform.

### New behavior

if implicit conversion is not allowed then the Jit throws `InvalidProgramException`.
There are 2 cases that are still tolerable:
- int8->nint on 64-bit platform, because it is used often and does not lead to bad code.
- byref->nint.

### Reason for change
That was creating silent bad codegen on some platforms including arm64 apple.

### Recommended action

If your app started to throw `InvalidProgramException` after 6.0 update because of this issue fix the method that is causing this by using an explicit conversion for the argument or fix the callee declaration.

### Category

- [ ] JIT

### Affected APIs

Not detectable via API analysis

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #24342.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/jit/6.0/coerce-call-arguments-ecma-335?branch=pr-en-us-24799).

### DIFF
--- a/docs/core/compatibility/6.0.md
+++ b/docs/core/compatibility/6.0.md
@@ -43,7 +43,7 @@ If you're migrating an app to .NET 6, the breaking changes listed here might aff
 
 ## JIT compiler
 
-- [Coerce call arguments according to ECMA-33](jit/6.0/coerce-call-arguments-ecma-335.md)
+- [Coerce call arguments according to ECMA-335](jit/6.0/coerce-call-arguments-ecma-335.md)
 
 ## Networking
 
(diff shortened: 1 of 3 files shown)


sandreenko (approved):
LGTM, thanks @gewarren

PR #24807, opened 2021-06-23
Echo dates, and use hardcoded config args

IEvangelist:
Echo dates, and use hardcoded config args


BillWagner (approved):
LGTM, for now

BillWagner:
Actually, I'm fine hardcoding it. This YML is in this repo, it will never run in a different repo.

PR #24669, opened 2021-06-14
Provide workaround for lack of DataTable support

tdykstra:
Fixes #21366 

Will change snippet link to a permalink in a new PR after this PR is merged.

Internal review URLs
* https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?branch=pr-en-us-24669
* https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to?branch=pr-en-us-24669
* https://review.docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-supported-collection-types?branch=pr-en-us-24669

Code file:
* https://github.com/tdykstra/dotnet-docs/blob/datatable/docs/standard/serialization/snippets/system-text-json-how-to/csharp/RoundtripDataTable.cs

PR #24858, opened 2021-06-25
Do something other than if/else if/else

IEvangelist:
## Summary

Attn: @tdykstra - what about this?

Related to #24669

### DIFF
--- a/docs/standard/serialization/snippets/system-text-json-how-to/csharp/RoundtripDataTable.cs
+++ b/docs/standard/serialization/snippets/system-text-json-how-to/csharp/RoundtripDataTable.cs
@@ -97,7 +97,7 @@ namespace RoundtripDataTable
                             DateTime value => key => writer.WriteString(key, value),
                             Guid value => key => writer.WriteString(key, value),
 
-                            { } value => key => writer.WriteString(key, value)
+                            _ => key => writer.WriteString(key, row[column].ToString())
                         };
                 }
                 jsonWriter.WriteEndObject();
--- a/docs/standard/serialization/snippets/system-text-json-how-to/csharp/RoundtripDataTable.cs
+++ b/docs/standard/serialization/snippets/system-text-json-how-to/csharp/RoundtripDataTable.cs
@@ -16,7 +16,7 @@ namespace RoundtripDataTable
 
     public static class Program
     {
-        public static List<WeatherForecast> weatherForecasts = new List<WeatherForecast>()
+        public static List<WeatherForecast> weatherForecasts = new()
         {
             new WeatherForecast { Date = new DateTime(2019, 8, 1), TemperatureCelsius = 25, Summary = "Hot" },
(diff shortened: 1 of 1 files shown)


docs/standard/serialization/snippets/system-text-json-how-to/csharp/RoundtripDataTable.cs
```diff
@@ -72,66 +72,33 @@ public override void Write(Utf8JsonWriter jsonWriter, DataTable value, JsonSeria
...
+                            _ when column.DataType == typeof(ushort) => key => writer.WriteNumber(key, (ushort)row[column]),
+                            _ when column.DataType == typeof(uint) => key => writer.WriteNumber(key, (uint)row[column]),
+                            _ when column.DataType == typeof(ulong) => key => writer.WriteNumber(key, (ulong)row[column]),
+
+                            // strings
+                            _ when column.DataType == typeof(DateTime) => key => writer.WriteString(key, (DateTime)row[column]),
+                            _ when column.DataType == typeof(Guid) => key => writer.WriteString(key, (Guid)row[column]),
+
+                            _ => key => writer.WriteString(key, row[column].ToString())
+                        };
```
Youssef1313, line 101: Would switching over `row[column]` with each switch arm being simplified to, for example, `bool b`, then use the "b" instead of a cast work? If it works, I think it will look nicer?

Youssef1313: Note: I wrote this directly in GitHub, so I may have an error.

```suggestion
                        DataRow row, DataColumn column, Utf8JsonWriter writer) => row[column] switch
                        {
                            // bool
                            bool value => key => writer.WriteBoolean(key, value),

                            // numbers
                            byte value => key => writer.WriteNumber(key, value),
                            sbyte value => key => writer.WriteNumber(key, value),
                            decimal value => key => writer.WriteNumber(key, value),
                            double value => key => writer.WriteNumber(key, value),
                            float value => key => writer.WriteNumber(key, value),
                            short value => key => writer.WriteNumber(key, value),
                            int value => key => writer.WriteNumber(key, value),
                            ushort value => key => writer.WriteNumber(key, value),
                            uint value => key => writer.WriteNumber(key, value),
                            ulong value => key => writer.WriteNumber(key, value),

                            // strings
                            DateTime value => key => writer.WriteString(key, value),
                            Guid value => key => writer.WriteString(key, value),

                            { } value => key => writer.WriteString(key, value)
                        };
```

tdykstra: @Youssef1313 Your code works correctly, but the last line needs to change to 

```
_ => key => writer.WriteString(key, row[column].ToString())
```

tdykstra (approved):
Looks good, compiles and runs.

PR #24867, opened 2021-06-25
Update version text

tdykstra:
Commit message details:

> This clarifies that --version displays the SDK version, not the muxur version (the muxor is the actual dotnet.exe) after discovering that some very smart people were confused about this.


luisquintanilla (approved):
Great find! LGTM

PR #24872, opened 2021-06-25
update to version 1.1.3

BillWagner:
This version of the what's new tool doesn't require start and end date. It used the previous month's first and last date.


.github/workflows/dependencies/run-dotnet-whatsnew.sh
```diff
@@ -7,9 +7,6 @@ set -e
 #     $2 is the <github.repository>
 #     $3 is the savedir
 
-declare -r STARTDATE=$(date "+%F" -d "$(date +'%Y-%m-01') -1 month");
-declare -r ENDDATE=$(date "+%F" -d "$STARTDATE +1 month -1 day");
-
 echo "From $STARTDATE to $ENDDATE"
```
scottaddie, line 10: Will this echo statement fail since the variables are no longer defined?

BillWagner: Yes. Good catch.

scottaddie (approved):
:shipit:

IEvangelist (approved):

PR #24896, opened 2021-06-29
Add docs for DisableFrameworkReferenceAnalyzers

ericstj:
## Summary

Add docs for property added in https://github.com/dotnet/sdk/pull/18501.

Design doc: https://github.com/dotnet/designs/blob/main/accepted/2021/InboxSourceGenerators.md

### DIFF
--- a/docs/core/project-sdk/msbuild-props.md
+++ b/docs/core/project-sdk/msbuild-props.md
@@ -732,7 +732,7 @@ The `ValidateExecutableReferencesMatchSelfContained` property can be used to dis
 
 ### DisableFrameworkReferenceAnalyzers
 
-The `DisableFrameworkReferenceAnalyzers` property can be used to disable all analyzers and/or source generators that are resolved from `FrameworkReference` items.  Similar to [EnableNETAnalyzers](#enablenetanalyzers) which controls analyzers that come from the SDK, however this property can be used to disable analyzers that are part of the framework itself (for example, `net6.0`).
+The `DisableFrameworkReferenceAnalyzers` property can be used to disable all analyzers and source generators that are resolved from `FrameworkReference` items.  This property is similar to [EnableNETAnalyzers](#enablenetanalyzers), which controls analyzers that come from the SDK. However, `DisableFrameworkReferenceAnalyzers` disables analyzers that are part of the framework itself, for example, `net6.0`.
 
 ```xml
 <PropertyGroup>
--- a/docs/core/project-sdk/msbuild-props.md
+++ b/docs/core/project-sdk/msbuild-props.md
@@ -684,6 +684,7 @@ The following MSBuild properties are documented in this section:
 - [DisableImplicitFrameworkReferences](#disableimplicitframeworkreferences)
 - [Restore-related properties](#restore-related-properties)
 - [ValidateExecutableReferencesMatchSelfContained](#validateexecutablereferencesmatchselfcontained)
+- [DisableFrameworkReferenceAnalyzers](#disableframeworkreferenceanalyzers)
 
 ### AssetTargetFallback
 
@@ -726,6 +727,16 @@ The `ValidateExecutableReferencesMatchSelfContained` property can be used to dis
 ```xml
 <PropertyGroup>
   <ValidateExecutableReferencesMatchSelfContained>false</ValidateExecutableReferencesMatchSelfContained>
+</PropertyGroup>
+```
+
+### DisableFrameworkReferenceAnalyzers
+
+The `DisableFrameworkReferenceAnalyzers` property can be used to disable all analyzers and/or source generators that are resolved from `FrameworkReference` items.  Similar to [EnableNETAnalyzers](#enablenetanalyzers) which controls analyzers that come from the SDK, however this property can be used to disable analyzers that are part of the framework itself (for example, `net6.0`).
+
+```xml
+<PropertyGroup>
+  <DisableFrameworkReferenceAnalyzers>true</DisableFrameworkReferenceAnalyzers>
 </PropertyGroup>
 ```
 


docs/core/project-sdk/msbuild-props.md
```diff
@@ -729,6 +730,16 @@ The `ValidateExecutableReferencesMatchSelfContained` property can be used to dis
 </PropertyGroup>
 ```
 
+### DisableFrameworkReferenceAnalyzers
+
+The `DisableFrameworkReferenceAnalyzers` property can be used to disable all analyzers and/or source generators that are resolved from `FrameworkReference` items.  Similar to [EnableNETAnalyzers](#enablenetanalyzers) which controls analyzers that come from the SDK, however this property can be used to disable analyzers that are part of the framework itself (for example, `net6.0`).
```
dsplaisted, line 735: This makes me question, should we really have two different properties for this?  If someone set `EnableNETAnalyzers` to false, wouldn't they expect not to get the FrameworkReference-based analyzers.

ericstj: I was thinking we should have a separate property since these are different in source and the reasons for disabling them might be different (eg: source build on old frameworks).

dsplaisted: To me it's pretty much an implementation detail that we decided to put analyzers in FrameworkReferences.  Most developers don't even need to know what FrameworkReferences are, or the fact that we have implicit ones for the base Framework.

ericstj: @jmarolf what is your thinking about this?  Does it make sense to have different properties?

jmarolf: 1. I think we want to keep the Enable/Disable language consistent here. If we do have both I do not think it makes sense to have `EnableNETAnalyzers` and `DisableFrameworkReferenceAnalyzers`. It should either be `DisableNETAnalyzers` and `DisableFrameworkReferenceAnalyzers` or `EnableNETAnalyzers` and `EnableFrameworkReferenceAnalyzers`
2. I need to think about these cases. For this case what is the user trying to do? They are trying to disable framework analyzers but keep other ones? Why? 

```xml
<DisableFrameworkReferenceAnalyzers>true</DisableFrameworkReferenceAnalyzers>
<EnableNETAnalyzers>true</EnableNETAnalyzers>
```

alternatively, what is happening here?
```xml
<DisableFrameworkReferenceAnalyzers>false</DisableFrameworkReferenceAnalyzers>
<EnableNETAnalyzers>false</EnableNETAnalyzers>
```

I agree with @dsplaisted  that we should not add configuration complexity just to add it. It must enable a specific scenario and that scenario must happen "in the wild" for at least one customer. As it stands I cannot think of a situation this applies to and am learning towards just having one property. Alternatively, we could have both, but I do not think we need to document both.

ericstj: The motivation behind the differences between Enable / Disable was that I had originally looked at `DisableImplicitFrameworkReferences` for precedent here, and this new property was meant to parallel that rather than `EnableNETAnalyzers`

One of the reasons that motivated the separate property is that the framework analyzers are coming from ref-packs so they may be older and not source-built.  

I'm fine going either way on this, but we'll need to make product changes to address any change we make here.
@terrajobst @KathleenDollard can you weigh in?

jmarolf: > One of the reasons that motivated the separate property is that the framework analyzers are coming from ref-packs so they may be older and not source-built.

My assumption would be to ask the customer to update their ref-packs if the analyzers that are coming along are not to their liking. Or they can use and editorconfig file and do whatever they want.

The motivation for `EnableNETAnalyzers`  is that we know it has a real build impact and so being able to easily disable analyzers was going to be an important escape hatch for customers (some of whom are building and deploying for single-core 32-bit machines).

Would like more detail on the customer scenario where `DisableImplicitFrameworkReferences` solves a customer problem that is not better solved by them updating their build environment.

ericstj: I can sympathize with having an common escape hatch for disabling unnecessary analyzers.

> Or they can use and editorconfig file and do whatever they want.

Can this control source generators too?

We are thinking about these a bit differently than the other analyzers.  These will contain source generators which are meant to act like an extension of the framework itself, so removing them could cause customer code to stop compiling or have different results.  cc @RussKie

jmarolf: > Can this control source generators too?

`editorconfig` can control what messages are shown in a build. The engine sometimes takes this information and makes optimization decisions (all the diagnostic messages in this analyzer are supressed so we do not need to run it etc.) but at a high level editorconfig does not change what is run only how the user interacts with the results.

> We are thinking about these a bit differently than the other analyzers. These will contain source generators which are meant to act like an extension of the framework itself, so removing them could cause customer code to stop compiling or have different results

right, but then isn't `<DisableFrameworkReferenceAnalyzers>true</DisableFrameworkReferenceAnalyzers>` an anti-pattern? Whats the case where the customer needs to do this? Is doing this considered a supported scenario? are we testing with this option off and on?

ericstj: > What's the case where the customer needs to do this? Is doing this considered a supported scenario? are we testing with this option off and on?

I've been thinking about it like an escape hatch in case they cause a [problem](https://github.com/dotnet/runtime/issues/56702) for whatever reason.  Similar to how you described `EnableNETAnalyzers`, but with greater consequences since you might depend on them running.  I don't think we'd recommend that folks do this, but it'd be nice to have this property in our back pocket so that we don't have to tell folks to write targets.

jmarolf: makes sense, I just need to understand if the problems that would cause people to use this option are the same or very different from `EnableNETAnalyzers`. If we think these situations are going to be about performance then I think this is a distinction without difference. for https://github.com/dotnet/runtime/issues/56702 I assume they would want to turn off _all_ analyzers. Yes these analyzers are coming from different places but I fail to see how that matters to the end user.

RussKie: > for [dotnet/runtime#56702](https://github.com/dotnet/runtime/issues/56702) I assume they would want to turn off _all_ analyzers.

Why all though? It looks like a problem with the specific SG (is it trying to do too much?).

jmarolf: How is the user going to be able to determine that it is a source generator coming from a framework reference?  If the build is broken by turning off framework reference generators then its not less broken by turning off all analyzers.

fitdev: @jmarolf Just wanted to clarify that our case with https://github.com/dotnet/runtime/issues/56702 had to do with a specific framework SG, which we were not depending on or using at all in our code. Just because there is a problematic SG, especially the one that's not being relied upon by user code, it does _not_ make sense to turn of _all_ SGs, let alone all analyzers. Ideally a user should have granular control over which SGs are enabled on a per-project basis, without resorting to [targets trick](https://github.com/dotnet/runtime/issues/56702#issuecomment-892079703), perhaps with the super option (in IDE maybe) to disable all SGs.

jmarolf: Right, that makes sense. But that is not what this property does. It disables all source generators coming as framework references. Still do not understand the user case of disabling all framework source generators but not all source generators. If you are trying to disable source generators for your build you are doing build surgery and frankly, really need to know what you are doing. My position is that adding this property does not make this daunting task simpler or easier.

ericstj: All right, I think I'm in agreement.  This property provides limited value, and might even be worse than that.  Potentially there is an opportunity for another feature here, which is to provide a hook to disable certain / all source generators.  Let's discuss that as a new SDK issue.  I will close this PR and remove the property from the SDK.

ericstj: Just to follow up on a few questions.

> It looks like a problem with the specific SG (is it trying to do too much?).

It's hot path is too costly, so large SLNs pay that hot path on a lot of source and it becomes too expensive.  

> How is the user going to be able to determine that it is a source generator coming from a framework reference

I think we need some source generator diagnostics like https://github.com/dotnet/roslyn/blob/315c2e149ba7889b0937d872274c33fcbfe9af5f/docs/analyzers/Report%20Analyzer%20Format.md to help folks realize what's going on.  We're going to talk to roslyn folks about this.

PR #24910, opened 2021-06-30
Fix reportgenerator usage example.

Escoto:
Just used this example a minute ago and noticed a wrong formatting on the parameters.


BillWagner (approved):
Thanks for spotting this and submitting the fix @Escoto We appreciate it.

PR #24931, opened 2021-06-30
Fix confusion and lack of clarity in various documents

### LINKED ISSUE #23822: The authors of this documentary confused “preorder” with “postorder”.
[It appears that in the fifth paragraph of this article (How to iterate through a directory tree), that beginning with “When you manually walk a directory tree”, the authors of this documentary confused “preorder” with “postorder”. Clearly, in a directory tree, files correspond to information in it, while the organization by subfolders allows the information as a whole to be structured. At the directory tree, the files in each directory correspond to the tree’s root and the other hand, subdirectories to subtrees. Therefore, go through the folds in the preorder, according to the classic work of Computing, The Art of Computer Programming, Donald Knuth, Vol. 1 / Fundamental Algorithms, Sections 2.3.1 Traversing Binary Trees and 2.3.2 Binary Tree Representation of Trees, “Preorder traversal is: visit the root, traverse the left subtree, and finally traverse the right subtree” what IT IS JUST THE OPPOSITE to what was stated in that paragraph, i.e., “that the entire tree is traversed in the current folder before iterating in the files that are directly in the folder itself “. Analogously, we can say something similar concerning what the same paragraph stated for postorder.]

---
#### Document Details

* ID: 25601a0f-eb3c-bf18-04d1-f371f5dc4341
* Version Independent ID: ce7a72d0-6156-f1f8-56ee-9774d93d6d3d
* Content: [How to iterate through a directory tree - C# Programming Guide](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/file-system/how-to-iterate-through-a-directory-tree)
* Content Source: [docs/csharp/programming-guide/file-system/how-to-iterate-through-a-directory-tree.md](https://github.com/dotnet/docs/blob/main/docs/csharp/programming-guide/file-system/how-to-iterate-through-a-directory-tree.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

### LINKED ISSUE #24882: Object oriented programming|Object Initializers do not circumvent explicit constructor call
From [Object oriented programming/Object Initializers](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/object-oriented/#object-initializers)

> You can instantiate and initialize class or struct objects, and collections of objects, without explicitly calling their constructor.

Considering example like

```csharp
 var list = new List<int> { 3 };
```
it is obivous that constructor is explictly called by the `new` keyword. For reference see [new operator (C# reference)/Constructor invocation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/new-operator#constructor-invocation)

Likely there should be something about ability to set properties without qualifying the access on the corresponding variable/instance of a type.

---
#### Document Details

* ID: 2eb21e6e-7798-e557-be3f-f27ab9c93e05
* Version Independent ID: a526a9cc-938e-7528-b5b4-4879c5085152
* Content: [Classes, structs, and records](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/object-oriented/)
* Content Source: [docs/csharp/fundamentals/object-oriented/index.md](https://github.com/dotnet/docs/blob/main/docs/csharp/fundamentals/object-oriented/index.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

steveberdy:
### Edits to how-to-iterate-through-a-directory-tree.md (issue #23822)

Line 26 fixes the confusion between pre-order and post-order traversal.

### Edits to index.md (issue #24882)

Line 79 fixes the lack of clarity concerning the meaning of "explicitly calling a constructor".

Fixes #23822
Fixes #24882

### DIFF
--- a/docs/csharp/fundamentals/object-oriented/index.md
+++ b/docs/csharp/fundamentals/object-oriented/index.md
@@ -76,7 +76,7 @@ You can define part of a class, struct, or method in one code file and another p
   
 ## Object Initializers  
 
-You can instantiate and initialize class or struct objects, and collections of objects, without explicitly calling their constructor.
+You can instantiate and initialize class or struct objects, and collections of objects, without explicitly calling their constructor with an ending `()` after the type.
   
 ## Anonymous Types  
 
--- a/docs/csharp/programming-guide/file-system/how-to-iterate-through-a-directory-tree.md
+++ b/docs/csharp/programming-guide/file-system/how-to-iterate-through-a-directory-tree.md
@@ -23,7 +23,7 @@ root.GetDirectories("*.*", System.IO.SearchOption.AllDirectories);
   
  The weakness in this approach is that if any one of the subdirectories under the specified root causes a <xref:System.IO.DirectoryNotFoundException> or <xref:System.UnauthorizedAccessException>, the whole method fails and returns no directories. The same is true when you use the <xref:System.IO.DirectoryInfo.GetFiles%2A> method. If you have to handle these exceptions on specific subfolders, you must manually walk the directory tree, as shown in the following examples.  
   
- When you manually walk a directory tree, you can handle the subdirectories first (*pre-order traversal*), or the files first (*post-order traversal*). If you perform a pre-order traversal, you walk the whole tree under the current folder before iterating through the files that are directly in that folder itself. The examples later in this document perform post-order traversal, but you can easily modify them to perform pre-order traversal.  
+ When you manually walk a directory tree, you can handle the files first (*pre-order traversal*), or the subdirectories first (*post-order traversal*). If you perform a pre-order traversal, you visit files directly under that folder itself, and then walk the whole tree under the current folder. Post-order traversal is the other way around, walking the whole tree beneath before getting to the current folder's files. The examples later in this document perform pre-order traversal, but you can easily modify them to perform post-order traversal.  
   
  Another option is whether to use recursion or a stack-based traversal. The examples later in this document show both approaches.  
   


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/docs?pullRequest=24931) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/docs?pullRequest=24931) before we can accept your contribution.<br/><br/>:x: steveberdy [sign now](https://cla.dotnetfoundation.org/dotnet/docs?pullRequest=24931)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/docs?pullRequest=24931) it.</sub>

BillWagner (approved):
Thanks @steveberdy 

Uzivatel919:
@BillWagner, @steveberdy 

>You can instantiate and initialize class or struct objects, and collections of objects, without explicitly calling their constructor with an ending `()` after the type.

In fact `()` ending is called argument list. Nonetheless this syntactic elision (as I believe) is not true benefit of object initializers and in fact is there just to allow the object initialization syntax to be more readable.

Imagine some class like
```csharp
public class SomeClass
{
  public SomeClass ( int id ) => Id = id;
  public int Id { get; }
  public string ArbitraryData { get; set; }  
}
```

It cannot be instantiated without specifying arguments for constructor since it has no parameter-less one.

```csharp
new SomeClass(default) // <--- It is error to omit explicit argument specification.
{ 
  ArbitraryData = default
}
```

By taking this sample further we can model in class like 

```csharp
public class SomeClass
{
  public SomeClass () => Id = 9966; // <-- Note the explicit value set.
  public int Id { get; }
  public string ArbitraryData { get; set; }
}
```

`new SomeClass { ArbitraryData = "Test" }.Id  // Output: 9966`

From this sample it is obvious that parameter-less constructor is called by the virtue behind the `new` keyword logic. Thus _object initializer_ is not language feature for omitting empty constructor _argument list_ when creating object.

From [Docs/.NET/C# guide/Language reference/Operators/new operator (C# reference)/Constructor invocation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/new-operator#constructor-invocation)

>To create a new instance of a type, you typically invoke one of the constructors of that type using the `new` operator:

It is `new` operator what invokes constructor not _argument list_.

For exhausting reference see C# 5.0 language specification, chapter 7.6.10.2 Object initializers.

Lines bellow have the same meaning.

```csharp
new SomeClass () { ArbitraryData = "Test" }
new SomeClass { ArbitraryData = "Test" }
```

The same as do lines

```csharp
new SomeClass () {}
new SomeClass ();
new SomeClass {}
```

Object initializers can be employed in cases when model design is turned in way to avoid setting default values in constructor when only some values are set up in some execution path while another ones are set in another path upon object creation.

Of course some (people) can prefer to instantiate object through initializer because it is their personal preference over constructor means.

The same _babble_ can be found at [Docs/.NET/C# guide/Programming guide/Classes and structs/Object and Collection Initializers (C# Programming Guide)/Object initializers](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/object-and-collection-initializers#object-initializers)

> Object initializers let you assign values to any accessible fields or properties of an object at creation time without having to invoke a constructor followed by lines of assignment statements. 

Appropriate is something like:

1. For [Docs/.NET/C# guide/Fundamentals/Object oriented programming/Object Initializers](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/object-oriented/#object-initializers)

_You can instantiate and initialize class or struct objects, and collections of object using member initializer syntax or collection initializer syntax respectively. Empty constructor argument list can be omitted._

2. For [Docs/.NET/C# guide/Programming guide/Classes and structs/Object and Collection Initializers (C# Programming Guide)/Object initializers](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/object-and-collection-initializers#object-initializers)

_Object initializers let you assign values to any accessible fields or properties of an object at creation time using member initializer syntax. Empty constructor argument list can be omitted._

----

From Constructor (object-oriented programming), https://en.wikipedia.org/w/index.php?title=Constructor_(object-oriented_programming)&oldid=1020375404 (last visited July 1, 2021).

>In class-based object-oriented programming, a constructor (abbreviation: ctor) is a special type of subroutine called to create an object. It prepares the new object for use, often accepting arguments that the constructor uses to set required member variables.

By the need parameter-less constructor invocation can be omitted when it has empty body but this is/would be compiler detail. 

Interpreting object initializer as replace for or circumvention of constructor is wrong, bad. 🖤

----

To obtain even more of available sureness let inspect CIL.

```csharp
static class ObjectInitializerConstructorCallTests
{
  static void Test1 () { new WithEmptyCtor { }; }
  static void Test2 () { new WithoutEmptyCtor { }; }
  class WithEmptyCtor { public WithEmptyCtor () { } }
  class WithoutEmptyCtor { } 
}
```
```csharp
.method private hidebysig static 
void Test1 () cil managed 
{
	.maxstack 8
	IL_0000: newobj instance void ObjectInitializerConstructorCallTests/WithEmptyCtor::.ctor()
	IL_0005: pop
	IL_0006: ret
}

.method private hidebysig static 
void Test2 () cil managed 
{		
	.maxstack 8		
	IL_0000: newobj instance void ObjectInitializerConstructorCallTests/WithoutEmptyCtor::.ctor()
	IL_0005: pop		
	IL_0006: ret
}
```

One can see constructor on both operation sequences. See also [OpCodes.Newobj Field](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.opcodes.newobj?view=net-5.0).

steveberdy:
@Uzivatel919 thanks for the comment. Since this was just a one line summary in an introduction, we're not covering all ways to construct an object and initialize it. We aren't saying that `new SomeClass ()` is the only way to do this, but we're using it as an example to show that `()` is not needed when we want to both instantiate **AND** initialize a class or struct object.

Uzivatel919:
>thanks for the comment. Since this was just a one line summary in an introduction, we're not covering all ways to construct an object and initialize it. We aren't saying that new SomeClass () is the only way to do this, but we're using it as an example to show that () is not needed when we want to both instantiate AND initialize a class or struct object.

I understand your intentions and realize what briefing is about but part _without explicitly calling their constructor_ is misleading at minimum.

What is difference between `()` and `{}` notation? Semantically none. Its just syntax. 

What about

_You can instantiate and initialize class or struct objects, and collections of objects using respective syntax within one statement._

Instantiate refers to `new`, object construction, … and initialize refers to object initializer, nested object initializer and collection initializer. So there is no need to inform about syntactic sugar options.

Sweet sugar
```csharp
new SomeClass () {}
new SomeClass ();
new SomeClass {}
```

In fact merged text claim can render wrong also in future. If language become enriched with option to not call default parameter-less constructor by omitting parameter list (`()`). In such case there will be semantic conflict – _do one want to not call constructor or do one just use some initializer syntax(?)_.

BillWagner:
This did need some improvement.

See #24968. As it is an introductory article, I didn't going into all the analysis here, but the new wording should avoid the misconceptions discussed here.

steveberdy:
@BillWagner thanks for doing that!

PR #24938, opened 2021-07-01
Add note about thread safety

### LINKED ISSUE #20563: Please add documentation for thread-safety guarantees of JsonDocument (and related types)
### Description

Looking at the [documentation for `JsonDocument`](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsondocument) it is not obvious if it can be safely used by multiple threads or not.

A cursory glance at the code suggests that the parsing is done up-front and, therefore, it is safe - however we tripped over the caching that is done via `_lastIndexAndString` (i.e., if multiple threads call `GetString` on an element at the same time, then they may get back the wrong value).

I want to be clear that this is a **doc bug**: I'm fine with having a string cache (since I assume that it provides noticeable perf benefits), but I would have liked clear guidance that `JsonDocument` and uses of any `JsonElement` from the same parent document cannot be done safely from multiple threads (without calling `Clone` on each element).

---
#### Document Details

* Product: **dotnet**
* GitHub Login: @tdykstra

tdykstra:
Fixes #20563

### DIFF
--- a/docs/standard/serialization/system-text-json-overview.md
+++ b/docs/standard/serialization/system-text-json-overview.md
@@ -33,6 +33,29 @@ There are some limitations on what parts of the library that you can use from Vi
 
 For information about security threats that were considered when designing <xref:System.Text.Json.JsonSerializer>, and how they can be mitigated, see [`System.Text.Json` Threat Model](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/docs/ThreatModel.md).
 
+## Thread safety
+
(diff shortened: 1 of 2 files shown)


gewarren (approved):
LGTM module build warnings:

Line 0: [Warning] Missing metadata attribute: 'zone_pivot_groups'. To use zone pivots in your file, you must specify the valid pivot group that contains the pivot IDs you want to use.
Line 54: [Warning] Cross reference not found: 'System.Text.Json.Node.JsonNode'.

PR #24956, opened 2021-07-01
Info on when counters were first introduced

shirhatti:


shirhatti:
Still need to fill in the remaining counters
👀 @josalem

josalem:
There are some counters in that doc that specify version dependencies in the description of the counter, see line 110 in the doc. We should be consistent in how we specify this requirement. One other consideration: we should differentiate between AspNet v3.1 and .net3.1 since you can run AspNet 2.1/2.2 on top of .net3.1 right?

shirhatti:
> We should be consistent in how we specify this requirement.

Sure. I'll update everything to the new format

> One other consideration: we should differentiate between AspNet v3.1 and .net3.1 since you can run AspNet 2.1/2.2 on top of .net3.1 right?

I don't think it's worth going into that level in detail. Especially given that the only supported scenario is running ASP.NET Core 2.1 on .NET Fx. If you're running ASP.NET Core 2.1 on .NET Core 3.1, you're already in unsupported territory

josalem:
Works for me 😃

josalem (approved):

docs/core/diagnostics/available-counters.md
```diff
@@ -13,128 +13,128 @@ The .NET runtime and libraries implement and publish several [`EventCounter`](./
...
+| :::no-loc text="Failed TLS Handshakes"::: (`failed-tls-handshakes`) | The total number of failed TLS handshakes | .NET 5 |
+| :::no-loc text="Request Queue Length"::: (`request-queue-length`) | The current length of the request queue | .NET 5 |
+| :::no-loc text="TLS Handshake Rate"::: (`tls-handshakes-per-second`) | The number of TLS handshakes per update interval | .NET 5 |
+| :::no-loc text="Total Connections"::: (`total-connections`) | The total number of connections to the web server | .NET 5 |
+| :::no-loc text="Total TLS Handshakes"::: (`total-tls-handshakes`) | The total number of TLS handshakes with the web server | .NET 5 |
 
 ## "System.Net.Http" counters
 
-The following counters are published by the HTTP stack.  These counters are available on .NET 5 and later versions only.
+The following counters are published by the HTTP stack.
```
BillWagner, line 85:
```suggestion
The following counters are published by the HTTP stack.
```

BillWagner (approved):
This LGTM.

I'll :shipit: now.

PR #24965, opened 2021-07-02
Update dotnet-new.md - wrong exampe shown

Der-Kraken:
The example showed how to create a VB-project. But the headline described how to create a netstandard project.


IEvangelist (commented):
@Der-Kraken - I'm curious, why do you think the example is wrong? Specifying the `--framework` is optional.

tdykstra (approved):
@Der-Kraken Thanks for finding and fixing this!

tdykstra:
@IEvangelist The command as it was created a .NET class library, not a .NET Standard class library as the heading line promised.

PR #21258, opened 2020-10-28
Adding Azure Service Fabric to modernization table

masnider:
Added Service Fabric to the modernize w/ containers table as it was missing. 

## Summary

Adding Azure Service Fabric to this guidance table based on functionality and customer experience.

Related to https://github.com/dotnet/docs/issues/15055

PR #24966, opened 2021-07-02
Remove table, as it confuses customers

gkaleta:
Please remove this chart as it confused customers more than it helps. Also it is not up to date if you decide to keep it.

## Summary

Describe your changes here.


IEvangelist:
Hmmm, I'm not actually sure. @nishanil this is for you to decide, I think that @DavidCBerry13 recently added similar content to the .NET Azure area. What do you both think about consolidating / updating?

masnider:
+1  Since the table leaves out so much detail (and several products) we have also seen it confuse and disappoint customers.

One example, which caused a disruption with a critical partner (Accenture) https://github.com/dotnet/docs/issues/15055

A PR I raised to address that feedback was rejected: https://github.com/dotnet/docs/pull/21258

DavidCBerry13:
I'm good with removing the table.  I think the intent was good but the info has become dated and confusing for people.

IEvangelist:
Hey, @nishanil and @CamSoper -- please chime in here, what do you think?

CamSoper:
Yeah this is done to death elsewhere, and with pictures. Let's kill the table.

IEvangelist (approved):

masnider:
Thanks everyone!

PR #24968, opened 2021-07-02
Clarify description of object initializers

BillWagner:
The statement "without calling the constructor" could be easily misinterpreted as claiming that no constructor is executed. That is incorrect.


IEvangelist (approved):

PR #24974, opened 2021-07-02
promote the standard in TOC

### LINKED ISSUE #24959: Promote ECMA spec documents in TOC
The ECMA standard and upcoming proposal articles should move out from under the "Language Reference" node, to a sibling node in the C# TOC. That would more clearly separate the language reference, which describes the features up through C# 10 in a more conversational tone from the C# Standard, which is more normative and currently behind the shipping product.

BillWagner:
Fixes #24959

[Internal preview link](https://review.docs.microsoft.com/en-us/dotnet/csharp/?branch=pr-en-us-24974)


DamianEdwards:
Could we actually nest those two nodes under the "Language reference" node? That seems to be how the structure is intended as clicking on one shows "Language reference" as the parent in breadcrumbs:
image

BillWagner:
@DamianEdwards 

> Could we actually nest those two nodes under the "Language reference" node?

That's where they are now.  Would a better fix be updating the breadcrumbs to match the new TOC?

DamianEdwards:
@BillWagner sorry I'm still not quite following. This is what I see:
image

Those three nodes are all at the same level in the TOC.

BillWagner:
@DamianEdwards Take a look now. I did a bit of rearranging.

I could also add another level under these nodes as well:

Screen Shot 2021-07-02 at 4 48 17 PM

BillWagner:
@Youssef1313 This is another false positive on the build verifier.

Youssef1313:
@BillWagner Is this specific to `toc.yml` only (ie, other yml files will need redirection)?

BillWagner:
@Youssef1313 

>  Is this specific to toc.yml only (ie, other yml files will need redirection)?

This only applies to any toc.yml file.

tdykstra (approved):
The new structure looks good, makes sense to me.  One nit: the TOC is out of sync with the text of some docs, e.g. [C# 9 Records](https://review.docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/records?branch=pr-en-us-24974)  intro text calls it a "proposal" that "tracks the specification", while the TOC identifies it as a specification.

BillWagner:
closing and reopening for lint to be run

adegeo:
Should these be ordered descending? I would think we want the latest version at the top.

PR #24985, opened 2021-07-04
Clarify "string" vs "System.String" in C# programming guide

aaronfranke:
## Summary

`string` and `String` are only equivalent if you have `using System;` in your file. This PR improves the docs to clarify this.

### DIFF
--- a/docs/csharp/programming-guide/strings/index.md
+++ b/docs/csharp/programming-guide/strings/index.md
@@ -13,7 +13,7 @@ A string is an object of type <xref:System.String> whose value is text. Internal
   
 ## string vs. System.String  
 
- In C#, the `string` keyword is an alias for <xref:System.String>. Therefore, `String` and `string` are equivalent, and you can use whichever naming convention you prefer. The `String` class provides many methods for safely creating, manipulating, and comparing strings. In addition, the C# language overloads some operators to simplify common string operations. For more information about the keyword, see [string](../../language-reference/builtin-types/reference-types.md). For more information about the type and its methods, see <xref:System.String>.  
+ In C#, the `string` keyword is an alias for <xref:System.String>. Therefore, `String` and `string` are equivalent, regardless it is recommended to use the provided alias `string` as it works even without `using System;`. The `String` class provides many methods for safely creating, manipulating, and comparing strings. In addition, the C# language overloads some operators to simplify common string operations. For more information about the keyword, see [string](../../language-reference/builtin-types/reference-types.md). For more information about the type and its methods, see <xref:System.String>.
   
 ## Declaring and Initializing Strings  
 


docs/csharp/language-reference/builtin-types/reference-types.md
```diff
@@ -31,7 +31,7 @@ The `object` type is an alias for <xref:System.Object?displayProperty=nameWithTy
 
 ## The string type
 
-The `string` type represents a sequence of zero or more Unicode characters. `string` is an alias for <xref:System.String?displayProperty=nameWithType> in .NET.
+The `string` type represents a sequence of zero or more Unicode characters. `string` is an alias for <xref:System.String?displayProperty=nameWithType> in .NET. If you have `using System;` in your file, `string` and `String` are equivalent.
```
Youssef1313, line 34: Even if `using System;` is there. They may not be equivalent (https://blog.paranoidcoding.com/2019/04/08/string-vs-String-is-not-about-style.html):

```csharp
using System;

String x = "Hello, world!";
Console.WriteLine(x);

namespace System
{
    class String
    {
        public static implicit operator String(string s) => null;
    }
}
```

It's a corner case that I may *not* be worth mentioning in docs.

docs/csharp/programming-guide/strings/index.md
```diff
@@ -13,7 +13,9 @@ A string is an object of type <xref:System.String> whose value is text. Internal
   
 ## string vs. System.String  
 
- In C#, the `string` keyword is an alias for <xref:System.String>. Therefore, `String` and `string` are equivalent, and you can use whichever naming convention you prefer. The `String` class provides many methods for safely creating, manipulating, and comparing strings. In addition, the C# language overloads some operators to simplify common string operations. For more information about the keyword, see [string](../../language-reference/builtin-types/reference-types.md). For more information about the type and its methods, see <xref:System.String>.  
+ In C#, the `string` keyword is an alias for <xref:System.String>. If you have `using System;` in your file, `string` and `String` are equivalent, and you can use whichever naming convention you prefer. If you don't have `using System;`, your options are `string` and `System.String`, therefore `string` is more portable and is recommended.
```
Youssef1313, line 16: I'd always recommend `string` regardless `using System;` is there or not. Probably refer to [.NET runtime coding style document](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md).

> We use language keywords instead of BCL types (e.g. int, string, float instead of Int32, String, Single, etc) for both type references as well as method calls (e.g. int.Parse instead of Int32.Parse). See issue [#13976](https://github.com/dotnet/runtime/issues/13976) for examples.

aaronfranke: What should I shorten it to? Maybe just remove the comparison altogether?

> In C#, the `string` keyword is an alias for \<xref:System.String\>. Using `string` is more portable and is recommended.

Youssef1313: I think @BillWagner or @IEvangelist should be able to provide guidance on what information to include here.

MSDN-WhiteKnight:
>therefore `string` is more portable and is recommended

"Portable" is generally used when talking about different architectures or operating systems, i think it could be confusing here. Using language keywords is recommended not because it enables avoiding `using System;`, this directive is common anyway; but actually, it is because identifiers can be resolved to different symbols depending on what is imported into current scope. So we could say something like "using `string` is recommended, because it unambigously refers to the `System.String` type in any context".

docs/csharp/language-reference/builtin-types/reference-types.md
IEvangelist: I agree with @Youssef1313 here on this, and thank you for sharing @jaredpar post -- it is a very good concern to be aware of.

docs/csharp/programming-guide/strings/index.md
IEvangelist: Yes, we should ALWAYS ALWAYS be using the alias names. This is the guidance I'd suggest for any and all C# developers, new and old alike. The alias names are part of the language, and exist to simplify your code. Most of the time a C# file will be `using System;`, but if it's not needed -- it's nice to remove it. It actually cleans the code up. I also much rather prefer the elegant look of using the aliases as they're more distinct, and have consistent character casing with other C# specific keywords.

I was going to say exactly what @Youssef1313 did, here: https://github.com/dotnet/docs/pull/24985/files#r663539805

IEvangelist:
```suggestion
 In C#, the `string` keyword is an alias for <xref:System.String>. Therefore, `String` and `string` are equivalent, regardless it is recommended to use the provided alias `string`. The `String` class provides many methods for safely creating, manipulating, and comparing strings. In addition, the C# language overloads some operators to simplify common string operations. For more information about the keyword, see [string](../../language-reference/builtin-types/reference-types.md). For more information about the type and its methods, see <xref:System.String>.
```

docs/csharp/language-reference/builtin-types/reference-types.md
IEvangelist:
```suggestion
The `string` type represents a sequence of zero or more Unicode characters. `string` is an alias for <xref:System.String?displayProperty=nameWithType> in .NET.
```

I do not believe that this is accurate to say, nor do I think it is needed here.

IEvangelist (changes requested):

IEvangelist:
Hi @aaronfranke thank you very much for your PR, it's much appreciated. With that said, I've provided a review - and wanted to offer up a few thoughts. Looking forward to hearing back from you, thanks.

aaronfranke:
I updated the PR with your suggestions and also added a note about `using System;`.

BillWagner (approved):
Thanks for all the work and discussion @aaronfranke @IEvangelist @Youssef1313 

This LGTM. Let's :shipit:

IEvangelist (approved):

PR #24995, opened 2021-07-06
Create live-protection.yml

Youssef1313:
This avoids problems like https://github.com/dotnet/docs/pull/20476. But feel free to reject if you think it's an unneeded noise.

### DIFF
--- a/.github/workflows/live-protection.yml
+++ b/.github/workflows/live-protection.yml
@@ -12,5 +12,5 @@ jobs:
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: 'It looks like this PR may be opened on live branch by mistake. Unless this is a merge from main to live, PRs should normally be opened on main branch.'
+              body: 'It looks like this pull request may have been opened on the `live` branch by mistake. Unless this is a merge from `main` to `live`, PRs should be opened targeting the `main` branch.'
             })
--- a/.github/workflows/live-protection.yml
+++ b/.github/workflows/live-protection.yml
@@ -2,7 +2,7 @@ on: [pull_request_target]
 
 jobs:
   comment:
-    if: ${{ github.event.issue.user.login != 'cxwtool' && github.ref == 'refs/heads/live' }}
+    if: ${{ github.base_ref == 'refs/heads/live' && !(github.event.issue.user.login == 'cxwtool' || github.head_ref == 'refs/heads/main') }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/github-script@v4
--- a/.github/workflows/live-protection.yml
+++ b/.github/workflows/live-protection.yml
@@ -12,5 +12,5 @@ jobs:
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: 'It looks like this pull request may have been opened on the `live` branch by mistake. Unless this is a merge from `main` to `live`, PRs should be opened targeting the `main` branch.'
+              body: 'It looks like this pull request may have been opened on the `live` branch by mistake. In general, PRs should target the `main` branch.'
             })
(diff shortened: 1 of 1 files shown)


.github/workflows/live-protection.yml
```diff
@@ -0,0 +1,16 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v4
+        with:
+          script: |
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'It looks like this PR may be opened on live branch by mistake. Unless this is a merge from main to live, PRs should normally be opened on main branch.'
```
IEvangelist, line 15:
```suggestion
              body: 'It looks like this pull request may have been opened on the `live` branch by mistake. Unless this is a merge from `main` to `live`, PRs should be opened targeting the `main` branch.'
```

IEvangelist: I'm not sure if it accepts the backtick formatting.

.github/workflows/live-protection.yml
```diff
@@ -0,0 +1,16 @@
+on: [pull_request_target]
+
+jobs:
+  comment:
+    if: ${{ github.event.issue.user.login != 'cxwtool' && github.ref == 'refs/heads/live' }}
```
IEvangelist, line 5: Is there anyway to also check if the PR is coming from `main`, or coming from an admin user? Those two use cases would be helpful to prevent noise.

Youssef1313: I think even if it comes from an admin user, it's good to notify. See https://github.com/dotnet/docs/pull/21749, which I think may have been opened in `live` by mistake.

gewarren, line 5:
```suggestion
    if: ${{ && github.base_ref == 'refs/heads/live' && github.head_ref != 'refs/heads/main' }}
```

I think it's good to check that the source branch is main.

Youssef1313: @gewarren The tool creates a "snapshot" branch named `publish-ID`. So checking `main` doesn't seem correct.

Youssef1313: For the live check, I'm not sure if I should be using `base_ref` or `ref`, or if there are any differences.

gewarren: Oh right. How about:

```suggestion
    if: ${{ github.base_ref == 'refs/heads/live' && !(github.event.issue.user.login == 'cxwtool' || github.head_ref == 'refs/heads/main') }}
```

gewarren: > For the live check, I'm not sure if I should be using `base_ref` or `ref`, or if there are any differences.

`base_ref` seems to be more specific to pull requests, so I'd go with that. `ref` can be a branch or tag.

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

.github/workflows/live-protection.yml
```diff
@@ -0,0 +1,16 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v4
+        with:
+          script: |
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'It looks like this pull request may have been opened on the `live` branch by mistake. Unless this is a merge from `main` to `live`, PRs should be opened targeting the `main` branch.'
```
gewarren, line 15:
```suggestion
              body: 'It looks like this pull request may have been opened on the `live` branch by mistake. In general, PRs should target the `main` branch.'
```

gewarren (approved):
Slight wording suggestion.

gewarren:
@Youssef1313 Is there a way to move the "if" statement to the actual execution? The problem is that for most PRs (i.e. to main), this check is skipped due to the "if" requirements, which leaves the status checks section of a PR expanded by default. It would be nicer if it showed as "passed" so it didn't expand the status checks section.

PR #25007, opened 2021-07-06
Revert "Fix text that was causing translation error"

tdykstra:
Reverts dotnet/docs#21749

was accidentally merged directly to live


tdykstra:
Thanks @Youssef1313 for pinging me again on this, as I missed the earlier notices due to some errant Outlook rules.

PR #25037, opened 2021-07-08
Mention the unit option for While

pbiggar:
## Summary

Mention that the `While` construct in computation expressions can have another form.

Other expressions, like Combine, already mention that there are alternate forms allowed.

According to Expert F# 4.0, 4th edition, page 473, Table 16-2:

> Used to de-sugar `while ... do ...` within computation expressions. `M<'T>` may optionally be `M<unit>`


cartermp (approved):
Thanks!

PR #25041, opened 2021-07-08
fix h1 to match title

Benio101:


docs/standard/events/index.md
```diff
@@ -15,7 +15,7 @@ helpviewer_keywords:
   - "events [.NET]"
 ms.assetid: b6f65241-e0ad-4590-a99f-200ce741bb1f
 ---
-# Handle and raising events
+# Handling and raising events
```
Youssef1313, line 18: h1s are okay to not match titles, and in fact it might be better not to be exact match. Also, the docs team has been moving from using the "ing" forms.

Benio101: So how about *Handle and raise events* then?

Youssef1313: @Benio101 Oh I didn't catch the "ing" in "raising" earlier. Yeah "Handle and raise events" should be good.

cc @gewarren

gewarren, line 18:
```suggestion
# Handle and raise events
```

gewarren (approved):
Thank you @Benio101

PR #25064, opened 2021-07-09
Minor clarification

adegeo:
## Summary

@gewarren I forgot to get this commit in my last PR. I just noticed it when cleaning my branches. Here it is.

### DIFF
--- a/docs/core/versions/selection.md
+++ b/docs/core/versions/selection.md
@@ -141,7 +141,7 @@ The roll-forward behavior for an application can be configured in four different
 
 ### Precedence
 
-Roll forward behavior is set by the following order when your app is run, higher numbered items taking precedent over lower numbered items:
+Roll forward behavior is set by the following order when your app is run, higher numbered items taking precedence over lower numbered items:
 
 01. First the `*.runtimeconfig.json` config file is evaluated.
 01. Next, the `DOTNET_ROLL_FORWARD` environment variable is considered, overriding the previous check.
--- a/docs/core/versions/selection.md
+++ b/docs/core/versions/selection.md
@@ -141,11 +141,11 @@ The roll-forward behavior for an application can be configured in four different
 
 ### Precedence
 
-Roll forward behavior is considered in the following order:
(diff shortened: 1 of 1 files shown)


docs/core/versions/selection.md
```diff
@@ -141,11 +141,11 @@ The roll-forward behavior for an application can be configured in four different
 
 ### Precedence
 
-Roll forward behavior is considered in the following order:
+Roll forward behavior is considered in the following order when your app is run:
 
-01. `*.runtimeconfig.json` config file.
-01. The `DOTNET_ROLL_FORWARD` environment variable.
-01. Any `--roll-forward` parameter passed to the running application.
+01. First the `*.runtimeconfig.json` config file is evaluated.
```
gewarren, line 146: So the precedence is actually the opposite of this order. I think this might be confusing because you might think that item 1 takes highest precedence.

docs/core/versions/selection.md
```diff
@@ -141,11 +141,11 @@ The roll-forward behavior for an application can be configured in four different
 
 ### Precedence
 
-Roll forward behavior is considered in the following order:
+Roll forward behavior is set by the following order when your app is run, higher numbered items taking precedent over lower numbered items:
```
tdykstra, line 144:
```suggestion
Roll forward behavior is set by the following order when your app is run, higher numbered items taking precedence over lower numbered items:
```

docs/core/versions/selection.md
```diff
@@ -141,11 +141,11 @@ The roll-forward behavior for an application can be configured in four different
 
 ### Precedence
 
-Roll forward behavior is considered in the following order:
+Roll forward behavior is set by the following order when your app is run, higher numbered items taking precedence over lower numbered items:
```
tdykstra, line 144: Since you've built precedence info into the step descriptions themselves, you could also shorten the intro:
```suggestion
Roll forward behavior is determined by the following procedure when your app is run:
```

tdykstra (approved):
Looks great!

Issue #24814, opened 2021-06-23
Reinstate WF/WCF samples


gewarren:
Related to #24648. The WinZip self-extractor executable was signed with SHA-1 and the trusted root authority for that expired in May, so the download was removed. Let's put the samples on dotnet/samples and reinstate the conceptual docs.

- [x] Add samples to dotnet/samples repo
- [x] Reinstate conceptual docs (and update any links/file paths as needed)
- [x] Remove samples docs from dotnet/docs-archive-pr repo

PR #25067, opened 2021-07-10
Reinstate WF sample docs

gewarren:
Contributes to #24814


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when you and @HongGit  agree that it's ready.

PR #25107, opened 2021-07-13
Update Byte Array Interop Breaking Change

TanayParikh:
## Summary
- We backed out part of this breaking change via https://github.com/dotnet/aspnetcore/pull/33768

### DIFF
--- a/docs/core/compatibility/aspnet-core/6.0/byte-array-interop.md
+++ b/docs/core/compatibility/aspnet-core/6.0/byte-array-interop.md
@@ -13,27 +13,6 @@ Blazor now supports optimized byte-array interop, which avoids encoding and deco
 
 ASP.NET Core 6.0 Preview 6
 
-## Return byte array from JavaScript to .NET
-
-### Old behavior
-
-```typescript
-function someJSMethodReturningAByteArray() {
-    const data = new Uint8Array([ 1, 2, 3 ]);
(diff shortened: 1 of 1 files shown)


docs/core/compatibility/aspnet-core/6.0/byte-array-interop.md
```diff
@@ -71,7 +50,7 @@ In the preceding code example, you'd treat the incoming parameter in JavaScript
 
 ### Return byte array from JavaScript to .NET
 
-If .NET expects a `byte[]`, JavaScript must provide a `Uint8Array`. Previously, it was possible to provide a Base64-encoded array using `btoa`.
+If .NET expects a `byte[]`, JavaScript _should_ provide a `Uint8Array`. It's still possible to provide a Base64-encoded array using `btoa`, however that is less performant.
 
 For example, if you have the following code, then you must provide a `Uint8Array` from JavaScript that's _not_ Base64-encoded:
```
TanayParikh, line 55:
```suggestion
For example, if you have the following code, then you _should_ provide a `Uint8Array` from JavaScript that's _not_ Base64-encoded:
```

pranavkm (approved):

gewarren (approved):
Does line 10 need to be updated at all?

TanayParikh:
> Does line 10 need to be updated at all?

Nope, I think that's fine as-is.

PR #25108, opened 2021-07-13
dotnet run option breaking change

### LINKED ISSUE #24727: Deprecating `-p` for `--project` for `dotnet run`
This change is documented in [Spec for resolving '-p' in 'dotnet run' ](https://github.com/dotnet/designs/pull/229/files)

## Deprecate -p

`-p` will no longer be a valid abbreviation for `project`. In .NET 6, users will receive a warning. 

This warning will be from the CLI parser, so will not generally trigger WarningsAsErrors failures. However, if a wrapping MSBuild or CI process checks for the text "Warning" it will appear in that check. 

Also, if `--property`  is an option passed to the application being run, the ` -- ` syntax separator must be used.

### Version introduced

.NET Core 6.0 Preview 6

### Old behavior

* `-p` previously indicated `--project`.
* `--property` was not a valid option on `dotnet run` and was passed through to the underlying application, even if the `--` syntax separator was not used. 

### New behavior

* `-p` will retain this meaning for .NET 6, but will result in a warning that it is deprecated and to use the full `--project`. At some future point, we will remove the use of `-p` to indicate `--project`.
* `--property` is now a valid option on `dotnet run` and this option will not be passed through to the underlying application, unless the `--` syntax separator is used. 

### Reason for change

We are deprecating `-p` because of the close relationship `dotnet run` has with `dotnet build` and `dotnet publish`. This breaking change is the first step in aligning abbreviations for these commands. 

Embracing Xamarin/Maui results in a need to pass MSBuild properties to MSBuild during the build portion of `dotnet run`. We want the option specifying this to be the same for  `dotnet run`, `dotnet build` and `dotnet publish`. 

### Recommended action

If you encounter the new warning, please use `--project`. If you have a project argument that includes an `=` and use the `-p` abbreviation, it will be interpreted as `--property` and passed to MSBuild, so you will have to use `--project` to pass it to your application in all cases. 

If you have an application that has an option called `--property`, you will need to use the syntax separator to pass the the value to your application: 

```C#
dotnet run -- --property myPropertyValue
```

Please review any scripts that use `dotnet run` where you might overlook the warning if `-p` is used.  

If you have any scripts that are using “dotnet run” and process the output you could encounter a break.  `dotnet run` typically doesn’t output anything of its own if there are no errors, so you only get the output of the program that is being run.  If you have a script or other program wrapping “dotnet run” and parsing the output, the warning would be unexpected text and may cause a failure.

### Category

- [ ] ASP.NET Core
- [ ] C#
- [ ] Code analysis
- [ ] Core .NET libraries
- [ ] Cryptography
- [ ] Data
- [ ] Debugger
- [ ] Deployment
- [ ] Globalization
- [ ] Interop
- [ ] JIT
- [ ] LINQ
- [ ] MSBuild
- [ ] Networking
- [ ] Printing
- [x] SDK
- [ ] Security
- [ ] Serialization
- [ ] Visual Basic
- [ ] Windows Forms
- [ ] XML, XSLT

### Affected APIs

Not detectable via API analysis

---
#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #24727.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/6.0/deprecate-p-option-dotnet-run?branch=pr-en-us-25108).


docs/core/compatibility/sdk/6.0/deprecate-p-option-dotnet-run.md
```diff
@@ -0,0 +1,66 @@
...
+
+We are deprecating `-p` because of the close relationship `dotnet run` has with `dotnet build` and `dotnet publish`. This breaking change is the first step in aligning abbreviations for these commands.
+
+To embrace Xamarin (MAUI), we need to pass MSBuild properties to MSBuild during the build portion of `dotnet run`. The option that specifies passing properties should be the same for `dotnet run`, `dotnet build`, and `dotnet publish`.
+
+## Recommended action
+
+If you encounter the new warning, use `--project`. If you have a project argument that includes an `=` and you use the `-p` abbreviation, the option will be interpreted as `--property` and passed to MSBuild. In this case, use `--project` to pass the argument to your application in all cases.
+
+If your application has an option called `--property`, you'll need to use the syntax separator `--` to pass the the value to your application:
```
tdykstra, line 42:
```suggestion
If your application has an option called `--property`, you'll need to use the syntax separator `--` to pass the value to your application:
```

tdykstra (approved):
LGTM, one typo.

PR #25122, opened 2021-07-14
Correction to code sample

definedrisk:
## Summary

Should this none-deterministic call from Finalizer use Dispose(false) as it does for base class and none-derived classes?

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR_System/system.idisposable/cs/derived2.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR_System/system.idisposable/cs/derived2.cs
@@ -3,7 +3,7 @@
     // To detect redundant calls
     bool _disposed = false;
 
-    ~DerivedClassWithFinalizer() => this.Dispose(true);
+    ~DerivedClassWithFinalizer() => this.Dispose(false);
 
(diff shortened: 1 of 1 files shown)


IEvangelist:
Hi @definedrisk - great catch. I believe that you're correct that this should in fact be passing `false` not `true`. @bartonjs can you confirm quick?

```csharp
class DerivedClassWithFinalizer : BaseClassWithFinalizer
{
    // To detect redundant calls
    bool _disposed = false;

    ~DerivedClassWithFinalizer() => this.Dispose(true); // This should be false, right? 😬

    // Protected implementation of Dispose pattern.
    protected override void Dispose(bool disposing)
    {
        if (_disposed)
        {
            return;
        }

        if (disposing)
        {
            // TODO: dispose managed state (managed objects).
        }

        // TODO: free unmanaged resources (unmanaged objects) and override a finalizer below.
        // TODO: set large fields to null.
        _disposed = true;

        // Call the base class implementation.
        base.Dispose(disposing);
    }
}
```

[Reference](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#implement-the-dispose-pattern-for-a-derived-class)

IEvangelist (approved):

PR #25154, opened 2021-07-14
SQLite: Document Command Timeout connection string keyword

bricelam:


roji (approved):
FWIW in Npgsql and SqlClient the parameter is `Command Timeout` (`Default Timeout` not supported). Is `Default Timeout` intentional here?

bricelam:
Yes. It's the keyword used by **System**.Data.SQLite, and keeping things as compatible as possible between the two libraries helps users transition between them. As documented, `Command Timeout` is also a supported alias.

docs/standard/data/sqlite/connection-strings.md
```diff
@@ -74,7 +74,7 @@ A value indicating whether to enable foreign key constraints.
 There's no need enable foreign keys if, like in e_sqlite3, SQLITE_DEFAULT_FOREIGN_KEYS was used to compile the native
 SQLite library.
 
-### Recursive triggers
+### Recursive Triggers
```
IEvangelist, line 77:
```suggestion
### Recursive triggers
```

docs/standard/data/sqlite/connection-strings.md
```diff
@@ -13,7 +13,7 @@ keywords and values.
 
 The following connection string keywords can be used with Microsoft.Data.Sqlite:
 
-### Data source
+### Data Source
```
IEvangelist, line 16:
```suggestion
### Data source
```

Per our style guides, these were correct as is. Titles and heading should be in [sentence case](https://docs.microsoft.com/en-us/contribute/style-quick-start#other-style-and-voice-issues-to-watch-for). However, if these are proper names that can be capitalized, but I do not believe that these are.

docs/standard/data/sqlite/connection-strings.md
```diff
@@ -86,6 +86,15 @@ A value that indicates whether to enable recursive triggers.
 | True  | Sends `PRAGMA recursive_triggers` immediately after opening the connection. |
 | False | Doesn't send `PRAGMA recursive_triggers`. This is the default.              |
 
+### Default Timeout
```
IEvangelist, line 89:
```suggestion
### Default timeout
```

docs/standard/data/sqlite/connection-strings.md
```diff
@@ -86,6 +86,15 @@ A value that indicates whether to enable recursive triggers.
...
+
+The default timeout (in seconds) for executing commands. The default value is 30. *Command Timeout* is an alias of this keyword.
```
IEvangelist, line 91:
```suggestion
The default timeout (in seconds) for executing commands. The default value is 30. The*command timeout* is an alias of this keyword.
```

I believe that we're confusing the way to show these keywords, or aliases, or property names. If we're referring to it conceptually, we should call it "default timeout", or as it's property name `DefaultTimeout`. It shouldn't be *Default Timeout".

docs/standard/data/sqlite/connection-strings.md
```diff
@@ -86,6 +86,15 @@ A value that indicates whether to enable recursive triggers.
...
+
+This value can be overridden using <xref:Microsoft.Data.Sqlite.SqliteConnection.DefaultTimeout> which in turn can be overridden using <xref:Microsoft.Data.Sqlite.SqliteCommand.CommandTimeout>.
+
+> [!NOTE]
+> The Default Timeout keyword was added in version 6.0.
```
IEvangelist, line 96:
```suggestion
> The `DefaultTimeout` property was added in version 6.0.
```

Instead of an alert here at the end, you might want to open this section with this info. For example, "Starting with version 6.0, you can use the `DefaultTimeout` to ..."

IEvangelist (commented):
A few nits, otherwise this should be good to go..

bricelam:
@IEvangelist So... they keywords are case-sensitive and their whitespace is significant. How do you recommend reconciling this with the style guidelines?

docs/standard/data/sqlite/connection-strings.md
bricelam: But... they keywords are case-sensitive. I know this goes against casing guidelines, but I think it's important preserve the actual case.

docs/standard/data/sqlite/connection-strings.md
bricelam: No properties here. Connection strings only have keywords. Some keywords (like `Command Timeout`) are merely aliases for the canonical form of the keyword (in this case, `Default Timeout`)

docs/standard/data/sqlite/connection-strings.md
bricelam: Again, there are no properties in connection strings--only on .NET types.

Using `NOTE` is consistent with other keywords on this page, and with the EF Core docs. Given that we have three supported versions, I think this information is important to see while scanning.

docs/standard/data/sqlite/connection-strings.md
IEvangelist: Ok, in that case we need to add:

```markdown
no-loc: Command Timeout, Default Timeout, Data Source, etc. 
```

To the docs' metadata, it wasn't obvious that these were keywords.

docs/standard/data/sqlite/connection-strings.md
IEvangelist: Ah, good to know. I was unaware of that.

BillWagner:
Ping @IEvangelist Is this ready to merge?

docs/standard/data/sqlite/connection-strings.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Connection strings
-ms.date: 12/08/2020
+ms.date: 07/14/2021
```
IEvangelist, line 3:
```suggestion
ms.date: 07/14/2021
no-loc: Command Timeout,Default Timeout,Data Source,Recursive Triggers
```

PR #25174, opened 2021-07-15
Fix broken links - July

gewarren:
Contributes to #24848. (Hide whitespace changes in diff.)


docs/breadcrumb/toc.yml
```diff
@@ -464,7 +464,7 @@ items:
           topicHref: /dotnet/csharp/language-reference/proposals/csharp-8.0/index
         - name: C# 9.0 feature specifications
           tocHref: /dotnet/csharp/language-reference/proposals/csharp-9.0/
-          topicHref: /dotnet/csharp/language-reference/proposals/csharp-9.0/index
+          topicHref: /dotnet/csharp/language-reference/proposals/csharp-9.0/records
```
gewarren, line 467: @BillWagner I see that for C# 8, this is done through a redirection entry instead. LMK if you'd prefer that here.

IEvangelist (approved):

docs/breadcrumb/toc.yml
BillWagner: @gewarren The reason I added the redirect for "index.md" was so hackable URLs work.  It's nice to have, but I don't have strong feelings about it.

gewarren:
Thanks for reviewing David!

PR #25185, opened 2021-07-16
Update snippet in how-to-write-a-simple-parallel-for-loop

### LINKED ISSUE #25087: Use normal string[] args
the first code example makes explicit call to obtain the command-line argument instead of using the normal Main params
 https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/how-to-write-a-simple-parallel-for-loop#directory-size-example

happily the remaining code uses the familiar & standard approach [less brain cycles required!]

Would also be nice to use a local variable to hold that directory argument for arg[0] rather than repeated arg[1] stuff

thanks!

---
#### Document Details

* ID: 629911c2-f9cc-7deb-df8e-3012d3e6afd0
* Version Independent ID: 1603adc7-b23f-ff8f-4c18-7c3f84fbcba8
* Content: [How to: Write a Simple Parallel.For Loop](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/how-to-write-a-simple-parallel-for-loop)
* Content Source: [docs/standard/parallel-programming/how-to-write-a-simple-parallel-for-loop.md](https://github.com/dotnet/docs/blob/main/docs/standard/parallel-programming/how-to-write-a-simple-parallel-for-loop.md)
* GitHub Login: @IEvangelist
* Microsoft Alias: **dapine**

Naidile-P-N:
Changing Environment.GetCommandLineArgs() to Main(string[] args)

Fixes #25087

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR/conceptual.parallel.for/cs/for1.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR/conceptual.parallel.for/cs/for1.cs
@@ -6,11 +6,10 @@ using System.Threading.Tasks;
 
 public class Example
 {
-   public static void Main()
+   public static void Main(string[] args)
    {
       long totalSize = 0;
-
-      String[] args = Environment.GetCommandLineArgs();
+      
       if (args.Length == 1) {
          Console.WriteLine("There are no command line arguments.");
          return;
--- a/samples/snippets/csharp/VS_Snippets_CLR/conceptual.parallel.for/cs/for1.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR/conceptual.parallel.for/cs/for1.cs
@@ -10,22 +10,22 @@ public class Example
    {
       long totalSize = 0;
       
-      if (args.Length == 1) {
+      if (args.Length == 0) {
          Console.WriteLine("There are no command line arguments.");
          return;
       }
-      if (! Directory.Exists(args[1])) {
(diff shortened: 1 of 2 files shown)


samples/snippets/csharp/VS_Snippets_CLR/conceptual.parallel.for/cs/for1.cs
```diff
@@ -6,11 +6,10 @@
...
 public class Example
 {
-   public static void Main()
+   public static void Main(string[] args)
    {
       long totalSize = 0;
-
-      String[] args = Environment.GetCommandLineArgs();
+      
       if (args.Length == 1) {
```
IEvangelist, line 13: This `if` logic in now wrong, as the [`Environment.GetCommandLineArgs()`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.getcommandlineargs?view=net-5.0) automatically adds the first element to the array as the "The first element is the executable file name".

IEvangelist (changes requested):
The `if` check needs to be fixed, should be `if (args.Length == 0)`.

samples/snippets/csharp/VS_Snippets_CLR/conceptual.parallel.for/cs/for1.cs
Naidile-P-N: Thanks @IEvangelist. I have committed the changes.

IEvangelist (approved):

PR #25215, opened 2021-07-19
Update migrate-request-reply.md

guillermo-ruffino:
## Summary

I believe this is the intended keyword here, as explained in the text itself.


adegeo (approved):
Indeed, thank you for spotting that error!

PR #25221, opened 2021-07-19
Fixes #24673 doc conflates GetType(TypeName), Type.GetType, and Object.GetType

### LINKED ISSUE #24673: doc conflates GetType(TypeName), Type.GetType, and Object.GetType
The first part of the doc is correct - it discusses the *compile time* syntax "GetType(typename) ".

However, the doc is wrong, starting with "If you want to get the type object of an object variable, use the Type.GetType method.".

That sentence has two problems:
1. Should say "... use the Object.GetType method." This is not the same as Type.GetType, which has a different purpose.  
2. Despite having the same name, we are no longer talking about the "GetType Operator", which is the topic of this page. This sentence belongs at the very end of the section, maybe as part of "See Also". Must make it clear that we are now talking about something different!

----------
The section after that is also confused.

This part is correct (lead line and first bullet point):

    The GetType operator can be useful in the following circumstances:  

     *  You must access the metadata for a type at run time. ..."

HOWEVER, the second bullet point has NOTHING to do with the "GetType(TypeName)" operator.

    * You want to compare two object references to see if they refer to instances of the same type. ...

That is a use of "Object.GetType method", and belongs on page docs.microsoft.com/en-us/dotnet/api/system.object.gettype

If it is kept here, then again it should be part of "See Also" - to make it clear that we are no longer discussing "GetType operator".

Probably should have See Also link to both Type.GetType Method and Object.GetType Method.

---
#### Document Details

* ID: 1ccc42f8-1dbf-ac46-950a-faa9bd2c1dd8
* Version Independent ID: 9d6fd9b3-c88a-05cd-443f-c30c33170585
* Content: [GetType Operator - Visual Basic](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/operators/gettype-operator)
* Content Source: [docs/visual-basic/language-reference/operators/gettype-operator.md](https://github.com/dotnet/docs/blob/main/docs/visual-basic/language-reference/operators/gettype-operator.md)

eriawan:
## Summary

This PR update the doc based on the suggestion in the issue.

Fixes #24673

### DIFF
--- a/docs/visual-basic/language-reference/operators/gettype-operator.md
+++ b/docs/visual-basic/language-reference/operators/gettype-operator.md
@@ -41,13 +41,12 @@ GetType(typename)
   
 - Any enumeration defined by Visual Basic, the .NET Framework, or your application.  
   
- If you want to get the type object of an object variable, use the <xref:System.Type.GetType%2A?displayProperty=nameWithType> method.  
+ If you want to get the type object of an object variable, use the <xref:System.Object.GetType%2A?displayProperty=nameWithType> method.  
   
- The `GetType` operator can be useful in the following circumstances:  
+ The `GetType` operator can be useful in the following circumstance:  
   
-- You must access the metadata for a type at run time. The <xref:System.Type> object supplies metadata such as type members and deployment information. You need this, for example, to reflect over an assembly. For more information, see <xref:System.Reflection?displayProperty=nameWithType>.  
-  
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks for submitting this fix @eriawan 

We appreciate it.  I've reviewed the changes, and I'll :shipit: now.

PR #25291, opened 2021-07-21
Includes cleanup

gewarren:
My power was out for a few hours so I needed something I could do offline. I decided to replace INCLUDE files with their literal text and delete them.

*[Hide whitespace changes in diff](https://github.com/dotnet/docs/pull/25291/files?diff=unified&w=1).*


docs/framework/wcf/simplified-configuration.md
```diff
@@ -6,84 +6,84 @@ ms.topic: reference
...
-  
+Configuring Windows Communication Foundation (WCF) services can be a complex task. There are many different options and it is not always easy to determine what settings are required. While configuration files increase the flexibility of WCF services, they also are the source for many hard to find problems. .NET Framework 4.6.1 addresses these problems and provides a way to reduce the size and complexity of service configuration.
```
tdykstra, line 9: Wouldn't netfx-current be 4.8 actually, so if we replace it with a specific older version should we add "and later versions" (for netfx-current-short and netfx-current-long)?

includes/compiler-options.md
```diff
@@ -1,2 +1,2 @@
 
-Every compiler option is available in two forms: **-option** and **/option**. The documentation only shows the -option form.
+Every compiler option is available in two forms: **`-option`** and **`/option`**. The documentation only shows the `-option` form.
```
tdykstra, line 2: good catch 😄

tdykstra (approved):
LGTM

docs/framework/wcf/simplified-configuration.md
gewarren: True, those replacements will require a little more thought other than blanket find/replace, so I've left them unchanged for now (reverted my changes).

PR #25292, opened 2021-07-21
20.10 out of support

adegeo:
## Summary

Set 20.10 as out of support.

Contributes to dotnet/core#6452

@rbhanda


luisquintanilla (approved):
LGTM

PR #25302, opened 2021-07-22
Add Release .NET 6 label to breaking change issues

gewarren:


.repoman.yml
```diff
@@ -43,6 +43,17 @@ issues:
       pass:
         - labels-add: ["source incompatible"]
 
+    # Checks breaking change version and adds a label
+    - check:
+        - type: query
+          value: "contains(Issue.body, '.NET 6 Preview 7') == `true`
```
adegeo, line 49: I believe YAML needs special syntax to make mutliline strings. I'm also not sure if JMESPath handles multiline.

But why not just use a simple:

```
contains(Issue.body, '.NET 6 Preview') == `true`
```

gewarren: We can probably just use the simple form. I'll do another commit.

tdykstra (approved):
LGTM

gewarren:
Thanks Tom!

PR #25305, opened 2021-07-22
Invariant mode breaking change

### LINKED ISSUE #24849: Restrict Cultures Creation and full support casing in Globalization Invariant Mode
## Restrict Cultures Creation in Globalization Invariant Mode

- [Globalization Invariant Mode](https://github.com/dotnet/runtime/blob/main/docs/design/features/globalization-invariant-mode.md) has been supported in .NET and .NET Core which used in the app scenarios that don't need any globalization support. This mode is turned on by default on some Docker container like Alpine containers.
The behavior of the Globalization Invariant Mode used to allow any culture gets created if the culture name conform to [BCP-47](https://tools.ietf.org/search/bcp47). When creating any culture (e.g. `en-US`), the Invariant Culture data will be used instead of the real culture data. This breaking change is changing this behavior which will throw exception when creating any culture other than Invariant Culture.

- Globalization Invariant Mode was supporting case mapping for ASCII characters range only. We are enabling full case mapping support for Globalization Invariant Mode.

### Version introduced

.NET 6.0 

### Old behavior

- When Globalization Invariant Mode is turned on and the app creating a culture different than Invariant Culture, this creation operation succeeds but the returned culture will always use the Invariant Culture data at that time instead of the real culture data. 
- When doing any operations need character case mapping (e.g. string comparisons, string search, and upper casing or lower casing string), the case mapping was done only for ASCII characters.

### New behavior

- When Globalization Invariant Mode is turned on and the app creating a culture different than Invariant Culture, the culture creation operation will throw an exception. The new behavior will allow creating Invariant culture only.
- When doing any operations need character case mapping (e.g. string comparisons, string search, and upper casing or lower casing string), the case mapping will support all Unicode defined characters.

### Reason for change

- The reason of this change is to have easier way to diagnose the problems when the Globalization Invariant Mode is enabled. Many users run their applications in environment that enable the Globalization Invariant Mode while they are not aware about that. The applications experience the Globalization Invariant Mode behavior which can be unexpected and will be hard to the app owners to know what is wrong and how to diagnose the problem.
- Support full case mapping is to have a better experience in Globalization Invariant Mode and make it more usable. 

### Recommended action

For anyone using the Globalization Invariant Mode and want to get back the old behavior, can do that by setting the configuration switch `System.Globalization.PredefinedCulturesOnly` to `false` in the application configuration file or define the environment variable `DOTNET_SYSTEM_GLOBALIZATION_PREDEFINED_CULTURES_ONLY=false`.

### Category

- [ ] ASP.NET Core
- [ ] C#
- [ ] Code analysis
- [ ] Core .NET libraries
- [ ] Cryptography
- [ ] Data
- [ ] Debugger
- [ ] Deployment
- [x] Globalization
- [ ] Interop
- [ ] JIT
- [ ] LINQ
- [ ] MSBuild
- [ ] Networking
- [ ] Printing
- [ ] SDK
- [ ] Security
- [ ] Serialization
- [ ] Visual Basic
- [ ] Windows Forms
- [ ] XML, XSLT

### Affected APIs

- CultureInfo constructors
- CultureInfo.GetCultureInfo
- RegionInfo constructors 
- String comparisons and search APIs
- String upper casing and lower casing APIs.

#### Issue metadata

* Issue type: breaking-change

gewarren:
Fixes #24849 

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/globalization/6.0/culture-creation-invariant-mode?branch=pr-en-us-25305).


docs/core/compatibility/globalization/6.0/culture-creation-invariant-mode.md
```diff
@@ -0,0 +1,70 @@
...
+Environment variable:
+
+Set `DOTNET_SYSTEM_GLOBALIZATION_PREDEFINED_CULTURES_ONLY` to `false`.
+
+## Affected APIs
+
+- <xref:System.Globalization.CultureInfo.%23ctor%2A>
+- <xref:System.Globalization.CultureInfo.CreateSpecificCulture(System.String)?displayProperty=fullName>
+- <xref:System.Globalization.CultureInfo.GetCultureInfo%2A?displayProperty=fullName>
+- <xref:System.Globalization.CultureAndRegionInfoBuilder?displayProperty=fullName>
```
gewarren, line 64: Not sure if this one is correct. Also, are there other APIs I should add here?

docs/core/compatibility/globalization/6.0/culture-creation-invariant-mode.md
```diff
@@ -0,0 +1,70 @@
...
+  if ("Á".Equals("á", StringComparison.CurrentCultureIgnoreCase)) // Evaluates to true.
+  ```
+
+## Version introduced
+
+.NET 6 Preview 7
+
+## Reason for change
+
+The culture-creation change was introduced to more easily diagnosis culture-related problems. Some users are unaware that their apps are running in an environment where globalization-invariant mode is enabled. They may experience unexpected behaviors due to globalization-invariant mode and don't make the association, so it's hard to diagnose the issue.
```
tdykstra, line 43:
```suggestion
The culture-creation change was introduced to more easily diagnose culture-related problems. Some users are unaware that their apps are running in an environment where globalization-invariant mode is enabled. They may experience unexpected behaviors due to globalization-invariant mode and don't make the association, so it's hard to diagnose the issue.
```

tdykstra (approved):
LGTM, one typo

docs/core/compatibility/globalization/6.0/culture-creation-invariant-mode.md
```diff
@@ -0,0 +1,70 @@
...
+
+In most cases, no action is needed. However, if you desire the previous behavior, you can set a configuration switch or environment variable to allow creation of any culture in globalization-invariant mode.
+
+Application configuration file:
+
+Set `System.Globalization.PredefinedCulturesOnly` to `false`.
+
+Environment variable:
+
+Set `DOTNET_SYSTEM_GLOBALIZATION_PREDEFINED_CULTURES_ONLY` to `false`.
```
tarekgh, line 57: could you add, the config switch can be done too by setting the msbuild property 

```Xml
<PredefinedCulturesOnly>false</PredefinedCulturesOnly> 
```

in the csproj.

docs/core/compatibility/globalization/6.0/culture-creation-invariant-mode.md
tarekgh: Please remove this one. `CultureAndRegionInfoBuilder` is not supported in .NET Core in general.

tarekgh (approved):
LGTM. added a couple of minor comments.

gewarren:
Thanks for reviewing Tom and Tarek.

PR #25320, opened 2021-07-23
Whatsnew improved author check

BillWagner:
Upload a new 2.0.0 What's new package.

This requires a new secret, an OSPO KEY that access the REST API for the Microsoft Open Source Programs Office. That API provides a better way to determine if an issue author or PR author is a Microsoft employee. (used for reporting only).


tdykstra (approved):
LGTM

PR #25321, opened 2021-07-23
Dedupe example headings

gewarren:
Regex used for find: `## Example\s?\s?\n((.*\n)*?)## Example\s?\s?\n`

Contributes to #25322


Youssef1313 (commented):
If there are examples under different headings, they may get numbered sequentially while this shouldn't. Was it confirmed that this is not happening? (in those cases, if any, a descriptive heading for what the example does would solve the duplication without numbering incorrectly)

gewarren:
> If there are examples under different headings, they may get numbered sequentially while this shouldn't. Was it confirmed that this is not happening? (in those cases, if any, a descriptive heading for what the example does would solve the duplication without numbering incorrectly)

Do you mean if there are `Example` headings that aren't H2s? Adding descriptive headings would take a lot longer. For the most part, I was able to use RegEx to number these examples.

docs/csharp/programming-guide/types/boxing-and-unboxing.md
```diff
@@ -55,7 +55,7 @@ It is also possible to perform the boxing explicitly as in the following example
 
 This example converts an integer variable `i` to an object `o` by using boxing. Then, the value stored in the variable `i` is changed from `123` to `456`. The example shows that the original value type and the boxed object use separate memory locations, and therefore can store different values.
 
-## Example
+## Example 1
```
Youssef1313, line 58: @gewarren This is an example of what I meant. This example is under `## Boxing` heading (it should probably have been H3 to indicate its a sub-heading - but that's not the point). The second example is under `## Unboxing` which is an example for something *different*. I'm not sure if my point is valid, but that's how I think about numbered examples (that they are different example for *one specific* concept). Imagine if this was rewritten in proper structure, that would be:

```md
## Unboxing

Definition of unboxing here.

### Example 1

## Boxing

Definition

### Example 2
```

However, `Example 2` is the first example of Unboxing.

gewarren: I see what you mean. However, that's beyond the scope of this PR. As you say, these Example headings should really be H2s, as should the Description H2. This is the current in-article TOC which is definitely not great.

image

tdykstra: If we make the Example headings H3 (which makes sense to me), they shouldn't need numbers to differentiate them. Do we get duplicate heading warnings for H3 headings?

gewarren: Let me try it. The bookmarks still wouldn't function properly even if there are no warnings.

gewarren: No warnings for duplicate H3s.

tdykstra (approved):
If there are other instances where they should be under h2 headings, my preference would be to make them h3s without number suffixes. Otherwise this LGTM.

gewarren:
> If there are other instances where they should be under h2 headings, my preference would be to make them h3s without number suffixes. Otherwise this LGTM.

I checked ALL the previews and there were two more articles to fix (where the example headings should have been H3s), which I've now fixed.

gewarren:
Thanks for reviewing, Youssef and Tom!

Issue #25322, opened 2021-07-23
Clean our build


gewarren:
Example report: https://opbuilduserstoragepublic.blob.core.windows.net/report/2021%5C7%5C23%5Ca3bda507-3390-de91-8f7f-26f90f4e5fc8%5CCommit%5C202107232002450051-main%5Cworkflow_report.html?skoid=280a3845-ef4c-4943-95ab-56a392a37cc8&sktid=975f013f-7f24-47e8-a7d3-abc4752bf346&skt=2021-07-19T01%3A32%3A23Z&ske=2021-07-26T01%3A37%3A23Z&sks=b&skv=2020-04-08&sv=2020-04-08&st=2021-07-23T20%3A04%3A56Z&se=2021-08-23T20%3A09%3A56Z&sr=b&sp=r&sig=bRYfeOeBH5YwwUVf27HEeNLXEtfypTA%2FEHvgD710PTA%3D

gewarren:
I think all the warnings are gone now 🥳

PR #25323, opened 2021-07-24
Fix duplicate title etc. warnings

gewarren:
Contributes to #25322


tdykstra (approved):
LGTM. There's some inconsistency between use of parens vs. "in" -- "Title (Scope)" vs. "Title in Scope" -- but it's up to you whether you want to try to standardize on one approach.

gewarren:
> LGTM. There's some inconsistency between use of parens vs. "in" -- "Title (Scope)" vs. "Title in Scope" -- but it's up to you whether you want to try to standardize on one approach.

That was intentional. I used whichever format was different to the "duplicated" article.

gewarren:
Thanks for the review @tdykstra !

PR #25334, opened 2021-07-25
Fixed a Incomplete method call.

18-F-cali:
## Summary

Fixed Incomplete method name on line 88.
Intially
`ref int distance = plant`
changed to 
`ref int distance = Planet.GetEstimatedDistance();`
because the intent was demonstrating return a value by reference.


BillWagner (approved):
Thanks for spotting this and fixing it @18-F-cali 

PR #25348, opened 2021-07-26
Update nullable-operators.md

dsyme:
We've had commentary that F# has "too many operators"

In practice, users need to know very few of them, the reset are corner case, see https://github.com/dsyme/fsharp-presentations/blob/master/design-notes/on-teaching-operators.md

This is one of a number of edits I'll be making to help clarify when certain operators come into play.

### DIFF
--- a/docs/fsharp/language-reference/symbol-and-operator-reference/nullable-operators.md
+++ b/docs/fsharp/language-reference/symbol-and-operator-reference/nullable-operators.md
@@ -1,6 +1,6 @@
 ---
-title: Nullable Operators
-description: Learn about the nullable operators that are available in the F# programming language.
+title: Nullable Operators (Queries)
+description: Learn about the nullable operators that are available in the F# programming language, for use in query expressions.
 ms.date: 05/16/2016
 ---
 # Nullable Operators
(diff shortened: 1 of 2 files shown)


docs/fsharp/language-reference/symbol-and-operator-reference/nullable-operators.md
```diff
@@ -1,12 +1,16 @@
...
+description: Learn about the nullable operators that are available in the F# programming language, for use in query expressions.
 ms.date: 05/16/2016
 ---
-# Nullable Operators
+# Nullable Operators in Queries
 
 Nullable operators are binary arithmetic or comparison operators that work with nullable arithmetic types on one or both sides. Nullable types arise frequently when you work with data from sources such as databases that allow nulls in place of actual values. Nullable operators are used frequently in query expressions. In addition to nullable operators for arithmetic and comparison, conversion operators can be used to convert between nullable types. There are also nullable versions of certain query operators.
 
+> [!NOTE]
+> Nullable operators are only used in [query expressions](../query-expressions.md). If you do not use query expressions, you do not need to
```
gewarren, line 11:
```suggestion
> Nullable operators are only used in [query expressions](../query-expressions.md). If you don't use query expressions, you don't need to
```

gewarren (approved):
I left one suggestion. Using contractions is encouraged to make the voice less formal and more conversational.

cartermp:
Hmm, I struggle with this one. I understand the rationale, but the nullable operators work perfectly fine outside of the context of a query expression. In a way, it's somewhat weird that they're tucked away into that namespace, since they have orthogonal use cases.

dsyme:
I've added a factorization of the operator/symbol table.

Although the original table was ordered by ASCII character, based on feedback I've received today it's more useful reference material to split into logical sections so people have a chance of finding and understanding things

dsyme:
> Hmm, I struggle with this one. I understand the rationale, but the nullable operators work perfectly fine outside of the context of a query expression. In a way, it's somewhat weird that they're tucked away into that namespace, since they have orthogonal use cases.

I'll work out a way of addressing this in the wording.  Though in truth they're only present in the design for use in queries, at least I want to imply that that should be the case.

dsyme:
@cartermp This is now ready for another review.

There are pros and cons to splitting up the table, and we also need material that means people don't coming looking here early in learning, but I think it's better in balance, even for myself I would find this more useful.

PR #25360, opened 2021-07-27
Expand compatibility descriptions

gewarren:
Based on discussion with Mark Miller and Priya.

### DIFF
--- a/.github/ISSUE_TEMPLATE/breaking-change.yml
+++ b/.github/ISSUE_TEMPLATE/breaking-change.yml
@@ -45,8 +45,8 @@ body:
     attributes:
       label: Type of breaking change
       options:
-        - label: Binary incompatible (code must be recompiled to use the newer API version)
-        - label: Source incompatible (code must be changed and recompiled to use the newer API version)
+        - label: Binary incompatible (breaks existing binaries when loaded into the new runtime/component - for example, failure to load/execute or different run-time behavior)
+        - label: Source incompatible (breaks source code when targeting the new runtime/component/SDK - for example, compile errors or different run-time behavior)
     validations:
       required: true
   - type: textarea
--- a/.github/ISSUE_TEMPLATE/breaking-change.yml
+++ b/.github/ISSUE_TEMPLATE/breaking-change.yml
@@ -45,8 +45,8 @@ body:
     attributes:
       label: Type of breaking change
       options:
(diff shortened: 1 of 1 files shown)


PriyaPurkayastha (approved):

PriyaPurkayastha:
cc @marklio

.github/ISSUE_TEMPLATE/breaking-change.yml
```diff
@@ -45,8 +45,8 @@ body:
     attributes:
       label: Type of breaking change
       options:
-        - label: Binary incompatible (code must be recompiled to use the newer API version)
-        - label: Source incompatible (code must be changed and recompiled to use the newer API version)
+        - label: Binary incompatible (breaks existing binaries when loaded into the new runtime/component - for example, failure to load/execute or different run-time behavior)
```
marklio, line 48:
```suggestion
        - label: Binary incompatible (existing binaries may encounter a breaking change in behavior - for example, failure to load/execute or different run-time behavior)
```

.github/ISSUE_TEMPLATE/breaking-change.yml
```diff
@@ -45,8 +45,8 @@ body:
...
+        - label: Source incompatible (breaks source code when targeting the new runtime/component/SDK - for example, compile errors or different run-time behavior)
```
marklio, line 49:
```suggestion
        - label: Source incompatible (existing source code may encounter a breaking change in behavior when targeting the new runtime/component/SDK - for example, compile errors or different run-time behavior)
```

marklio (changes requested):
A couple of tweaks to the wording that make things a little more clear to me in this context. It feels like we should have a larger document with more complete definitions, and some examples. I'm requesting changes because the proposed wording seems to indicate that breaks are guaranteed. We rarely break things in this way. Typically, there is only a narrow breaking scenario.

marklio (approved):
I think this is a big improvement from a content perspective. I'm not entirely familiar with the template process to know whether this would be a breaking change to the template (would it make new issues different from existing ones and make it hard to to machine categorization, etc.)

gewarren:
> I think this is a big improvement from a content perspective. I'm not entirely familiar with the template process to know whether this would be a breaking change to the template (would it make new issues different from existing ones and make it hard to to machine categorization, etc.)

Thanks. We only just added these compatibility types, so I don't think any issues are actually labeled accordingly yet. A good time to make it right.

PR #25361, opened 2021-07-27
Changed discard example to show a discard

maxs-rose:
## Summary

Changed the discard example to show the use of a discard as previously the paragraph above talked about using a discard but didn't actually use one.
I have also added semantic names to the returned tuple for clarity.


BillWagner (approved):
Thanks for spotting this and submitting the fix @maxs-rose 

PR #25377, opened 2021-07-28
Clean up learning links on main F# docs page

dsyme:
We've been reviewing the various paths into F# learning

1. We'd like to highlight https://dotnet.microsoft.com/learn/fsharp as the curated learning page for F#. It needs work but Beth Massi is improving it and it is the primary learning link we should use

2. We need one "get started immediately" link.  Currently the best we have is https://dotnet.microsoft.com/learn/languages/fsharp-hello-world-tutorial/intro and indeed this is what's linked from the main F# landing page https://dotnet.microsoft.com/languages/fsharp.  So  let's use this here.

3. The Q&A forums are not yet appropriate to link due to low volume. We can revist this later. But suggest removing that for now

### DIFF
--- a/docs/fsharp/index.yml
+++ b/docs/fsharp/index.yml
@@ -9,15 +9,17 @@ metadata:
   ms.topic: landing-page # Required
   author: cartermp
   ms.author: phcart
-  ms.date: 08/29/2020
+  ms.date: 07/28/2021
 
 landingContent:
   - title: "Learn to program in F#"
     linkLists:
       - linkListType: get-started
         links:
-          - text: "What is F#?"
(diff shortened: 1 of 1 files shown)


dsyme:
cc @cartermp

gewarren (approved):
It's too bad that the first two links lead you away from this doc context, but it sounds like you've thought through this.

image

dsyme:
@gewarren I've tweaked it, could you check it again please?

The just-published beginner video series is included now 

The "functional programming" tutorial is now under "Tutorials" . This makes the Overview section thin (Tour of F#), but that's ok, it indicates where new material can slot in.

PR #25379, opened 2021-07-28
Adds the .NET library reference to the panel for F# libraries

dsyme:
Adds the .NET library reference to the panel for F# libraries

### DIFF
--- a/docs/fsharp/index.yml
+++ b/docs/fsharp/index.yml
@@ -106,4 +106,4 @@ landingContent:
           - text: "F# core library reference"
             url: https://fsharp.github.io/fsharp-core-docs/
           - text: ".NET library reference"
-            url: https://docs.microsoft.com/en-us/dotnet/api/?view=net-5.0
+            url: /dotnet/api/?view=net-5.0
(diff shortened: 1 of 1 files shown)


docs/fsharp/index.yml
```diff
@@ -86,18 +87,23 @@ landingContent:
...
 
-  - title: "F# Core Library reference"
+  - title: "F# library reference"
     linkLists:
       - linkListType: reference
         links:
-          - text: "FSharp.Core API reference"
-            url: https://fsharp.github.io/fsharp-core-docs/
+          - text: "F# core library reference"
+            url: https://aka.ms/fsharp-core-reference
```
gewarren, line 107: What's the reason to make this an aka.ms link? We generally avoid them in docs since it's so easy to update the docs - https://review.docs.microsoft.com/en-us/help/contribute/links-how-to?branch=master#fwlinks-and-akams-links.

dsyme: I see. I'll revert

I would expect us to move https://fsharp.github.io/fsharp-core-docs to either https://fsharp.github.io/docs or https://docs.fsharp.org/FSharp.Core at some point, but we can update later

gewarren (approved):
LGTM

PR #25435, opened 2021-08-03
Remove redundant content in formatting-rules.md

GitHubPang:
The *i* and *d* at the end of the two removed sentences are likely just the words *indent* and *do*, i.e., the start of the two sentences above for "Option values".

Seems like leftover text when porting this doc from an older format, so I removed them.


gewarren (approved):
Thanks for fixing that @GitHubPang !

PR #25443, opened 2021-08-03
❌ DO NOT prefix non-interface names with the letter I.

adamralph:
## Summary

It's (kind of) implied by _DO prefix interface names with the letter I, to indicate that the type is an interface_, but I think it's better to call it out explicitly, for clarity.

### DIFF
--- a/docs/standard/design-guidelines/names-of-classes-structs-and-interfaces.md
+++ b/docs/standard/design-guidelines/names-of-classes-structs-and-interfaces.md
@@ -36,6 +36,8 @@ The naming guidelines that follow apply to general type naming.
 
  For example, `IComponent` (descriptive noun), `ICustomAttributeProvider` (noun phrase), and `IPersistable` (adjective) are appropriate interface names. As with other type names, avoid abbreviations.
 
+ ❌ DO NOT prefix non-interface names with the letter I.
+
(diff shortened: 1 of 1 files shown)


Youssef1313 (changes requested):
This is unfortunately a copyrighted content where modifications are almost always refused :(

IEvangelist:
Hi @adamralph - thank you for posting this. I agree, and would normally approve and merge PRs like this - but @Youssef1313 is correct. This content is borrowed with permission from the publisher. The .NET Framework guidance like this is untouchable, we'll have to close this PR. Thank you regardless.

adamralph:
@Youssef1313 @IEvangelist I see. I didn't realise that this documentation was specific to .NET Framework.

Are there some other naming guidelines in existence for _all_ .NET code, which we _can_ contribute to?

adamralph:
In fact, if these kinds of things are already captured in the Roslyn analyzers which ship with the SDK, then I guess these guidelines are kind of redundant.

IEvangelist:
> In fact, if these kinds of things are already captured in the Roslyn analyzers which ship with the SDK, then I guess these guidelines are kind of redundant.

Right, that's kind of the point - this content was based originally on the Framework, they are more broad than that now. We don't even maintain them anymore. We've been discussing potentially removing them entirely.

PR #25460, opened 2021-08-04
Remove stmaclea

gewarren:
Also removed his GitHub account from the CODEOWNERS file, but left him as `author`.

(Looks like he's left Microsoft.)


IEvangelist (approved):

PR #25470, opened 2021-08-05
Update integral-numeric-types.md (integral types -> integral numeric types)

Zuishin:
Integral types contains the char type unlike integral numeric types.
So, I thought it is better to write "integral numeric types" if it does't contain char type.

## Summary

Describe your changes here.


IEvangelist:
Interesting. I don't think it's wrong to add this level of specificity - but I'd also like to point out that `char` also has a `.MinValue` and `.MaxValue`, i.e.; range. They also have a size, type and reserved keyword. So, it's not technically wrong to add them to this - is it?

- [Char.MinValue](https://docs.microsoft.com/dotnet/api/system.char.minvalue)
- [Char.MaxValue](https://docs.microsoft.com/dotnet/api/system.char.maxvalue)

BillWagner:
@Zuishin @IEvangelist 

I'd prefer a different change. I'll propose adding a line in the table for `char`, then then a note pointing to the page for `char`, explaining that some conversion rules for `char` are different than other integral types.

Does that sound reasonable?

BillWagner:
@Zuishin Do you want to update this PR with the requested changes?

BillWagner:
closing due to lack of response from author.

Issue #23813, opened 2021-04-16
Bar in Formatting try/with expressions


nojaf:
Hello,

In [Formatting try/with expressions](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-trywith-expressions), nothing is really mentioned about the bar in the with clauses.
Should it always be there?

You could omit it in some scenarios like:
```fsharp
try
    foo()
with ex -> recover ()
```

The current sample in the guide also has the clause under the with keyword, should that always be the case?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 8c09996f-24ac-cf39-ef43-d7909cbc2b89
* Version Independent ID: 1efdd523-8552-6421-e149-74f76826ef8f
* Content: [F# code formatting guidelines](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting)
* Content Source: [docs/fsharp/style-guide/formatting.md](https://github.com/dotnet/docs/blob/main/docs/fsharp/style-guide/formatting.md)
* Product: **dotnet-fsharp**
* GitHub Login: @cartermp
* Microsoft Alias: **phcart**

cartermp:
Yeah, it's not necessary in the single case, which offhandedly I'd say is the most common case.

I guess since it's a pattern match and we do have the `|` in each case, we should be consistent here.

PR #25498, opened 2021-08-06
Update F# formatting for exception patterns

dsyme:
This relates to  https://github.com/fsprojects/fantomas/issues/1854#issuecomment-894324296 and will need some discussion with @nojaf (since it messes the Fantomas formatting tool around to change these rules)

I don't understand why we would be recommending changing this code:
```fsharp
try
    persistState currentState
with ex ->
    printfn "Something went wrong: %A" ex
```
to
```fsharp
try
    persistState currentState
with
| ex ->
    printfn "Something went wrong: %A" ex
```

The second is surely harder to read.

For non-simple solitary clauses I don't mind using `|` in this case - it can help disambiguate the `:?` and allow further exception types to be added - but the simplest rule is "one clause, no bar".  So not this:
```fsharp
try
    persistState currentState
with
| :? SomeException as ex ->
    printfn "Something went wrong: %A" ex
```
but rather this:
```fsharp
try
    persistState currentState
with :? SomeException as ex ->
    printfn "Something went wrong: %A" ex
```

### DIFF
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -1012,21 +1012,22 @@ with
     printfn "A second that was a multiple of 3"
 ```
 
-Always add a `|` for each clause, even when only having a single clause.
+Add a `|` for each clause, except when there is only a single clause consisting of a simple value pattern:
 
 ```fsharp
 // OK
+try
+    persistState currentState
+with ex ->
+    printfn "Something went wrong: %A" ex
+
+// Not OK
 try
     persistState currentState
 with
 | ex ->
     printfn "Something went wrong: %A" ex
 
-// Not OK
-try
-    persistState currentState
-with ex ->
-    printfn "Something went wrong: %A" ex
 ```
 
 ## Formatting function parameter application
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -1012,13 +1012,19 @@ with
     printfn "A second that was a multiple of 3"
 ```
 
-Add a `|` for each clause, except when there is only a single clause consisting of a simple value pattern:
+Add a `|` for each clause, except when there is only a single clause:
 
 ```fsharp
 // OK
 try
     persistState currentState
 with ex ->
+    printfn "Something went wrong: %A" ex
+
+// OK
+try
+    persistState currentState
+with :? System.ApplicationException as ex ->
     printfn "Something went wrong: %A" ex
 
 // Not OK
@@ -1028,6 +1034,12 @@ with
 | ex ->
     printfn "Something went wrong: %A" ex
 
+// Not OK
+try
+    persistState currentState
+with
+| :? System.ApplicationException as ex ->
+    printfn "Something went wrong: %A" ex
 ```
 
 ## Formatting function parameter application


pbiggar:
Thanks for making this change, very much agree!

I reported the fantomas change, so forgive an opinion on the other change you mention:

```
try
    persistState currentState
with :? SomeException as ex ->
    printfn "Something went wrong: %A" ex
```

I think this is a better indentation than going onto multiple lines, for the same reason as the non exception case.

dsyme:
@pbiggar Yes to be honest I agree

gewarren (approved):
LGTM

nojaf:
Hey @dsyme, the original guidance was introduced in https://github.com/dotnet/docs/issues/23813.
I raised the question as there was no real example yet.
I proposed the syntax purely because it to have the same thing from the [GR style guide](https://github.com/G-Research/fsharp-formatting-conventions#formatting-trywith-expressions), to be honest.
My life is slightly easier when both guides give the same feedback 😄.

I'm ok with this change and the potential change it brings to Fantomas.
I try to uphold the following policy: this style guide is the default and using settings you can adhere to the GR one.

dsyme:
@nojaf Reading the GR style guid I suspect their "pattern matching on exceptions" may mean multiple-clause pattern matching.  They don't give single-clause examples.  I've asked for their input here.

https://github.com/G-Research/fsharp-formatting-conventions#formatting-trywith-expressions

Smaug123:
I personally strongly prefer the bar in all cases. I presume you agree that `match foo with MyThing x -> x` is not desirable (though F# does allow it)?

I even more strongly prefer `type Foo = | Blah of int`, because `type Foo = Blah` is ambiguous: is it a type alias, or a single-case DU? If you always use the pipe, there is no possible ambiguity.

And once you're always using the pipe for everything else, why not for this one remaining case of `try… with`? A foolish consistency may be the hobgoblin of small minds, but I find it much more aesthetically pleasing to have no arbitrary exceptions to the rules.

In general, I understand the G-Research formatting principles to be highly consistent and quite easily summarised: *prefer formatting which minimises the friction of extending the code*. This is why we have airy records with lots of whitespace, airy lists with lots of whitespace, a newline between the end of a lambda body and its closing paren, newlines between every function parameter and before the `=` (if it won't all fit on one line), mandatory `|` characters, preferred prefix syntax for generics, and so on: if you want to add something to them, you just need to insert a newline (or a comma, in the case of generics) and start typing. In particular, this isn't true of single-case matching without a pipe (or single-case DU definitions without a pipe), though it *is* true of single-case matching with a pipe and a newline.

Having said that, this might not be a hill we'd choose to die on. I'll run it past the rest of the team whether it's worth annoying Florian by insisting that we differ from the MS style guidelines.

dsyme:
@smaug123 - thanks for the feedback

> And once you're always using the pipe for everything else, why not for this one remaining case of try… with? 

I think because 

1. `try .. with...` is genuinely frequently used with a single identifier
2. The `with` pattern is partial matching and this even if there's a type test pattern there's still very often only one clause

For (1):

```fsharp
let x = expr // this happens

match expr with x ->  ...   // this never happens

try ... with x -> ... // this happens

```

Smaug123:
(And there is always the chance that I'm simply incorrect as an oracle for the GR style guide, of course!)

nojaf:
> Having said that, this might not be a hill we'd choose to die on. I'll run it past the rest of the team whether it's worth annoying Florian by insisting that we differ from the MS style guidelines.

Don't worry about that, I'm ok with a little schism from time to time. In this case, the impact is minimal.

PR #23394, opened 2021-03-19
Add missing profiler docs

davmason:
## Summary

Add profiler APIs from 5.0 to the docs

Issue #23215, opened 2021-03-08
[planning] Migration guide TOC design


StephenBonikowsky:
- Migration Guide
   - [ ] Overview ()

- Migration Investigation

- Pre-Migration

- Migration

- Post-Migration

Pilchie:
Note - Winforms shouldn't have significant challenges in migration that require re-writing.

KathleenDollard:
Can we have a section, probably under Pre-Migration Analyze and Asses that is something like "Visual Basic Considerations". I'd like a TOC level entry that helps VB folks make the decision. That would give me some context for words :)

Also, in the Analyze and Assess section, I don't see anything on LTS and servicing. There is a pretty massive difference between "As long as Windows lives" and 3 months or 1 year after next release. For some companies that will be a big factor in a decision.

StephenBonikowsky:
@KathleenDollard thanks for the feedback!
I added "Visual Basic Considerations".

There is some great info already out there in blogs and such (@jamshedd) regarding .NET releases, the cadence, the support story, etc...I added a place under General Info --> About .NET for that, and then under Migration --> Create a porting plan, I added a bullet point to address the question of what TFM should be targeted which could refernce back to the main page with all the release and servicing info. This would also be a great place to reference Immo's blog [The Future of .NET Standard](https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/)

PR #25508, opened 2021-08-07
Add orphans to TOC

gewarren:
Contributes to #25176.

Profiling articles were added this year in #23394.


docs/fundamentals/toc.yml
```diff
@@ -2544,6 +2544,8 @@ items:
   items:
   - name: Overview
     href: ../core/porting/index.md
+  - name: Port .NET Framework libraries to .NET
+    href: ../core/porting/libraries.md
```
gewarren, line 2548: @adegeo I assume this wasn't omitted purposefully?

BillWagner (approved):
This LGTM @gewarren 

You can :shipit: once you get the answer from @adegeo

docs/fundamentals/toc.yml
adegeo: Quite possibly. This area is in progress ([project 128](https://github.com/dotnet/docs/projects/128)) and @StephenBonikowsky is handling redesigning and rewriting all of this. It was probably removed when this was restructured and isn't added back in yet until it's rewritten to fit with the migration workflow.

gewarren: @StephenBonikowsky Should I remove [docs/core/porting/libraries.md](https://github.com/dotnet/docs/blob/main/docs/core/porting/libraries.md) from the table of contents? Should the article be deleted and redirected?

StephenBonikowsky: That file doesn't fit with the new TOC (https://github.com/dotnet/docs/issues/23215).
It's contents will be addressed in numerous other pages most of which don't exist yet.

I would delete it and if it needs to be re-directed maybe just point to the main Overview page (which itself will be getting a significant re-write)
https://docs.microsoft.com/en-us/dotnet/core/porting/

Issue #10746, opened 2019-02-26
Abridged Reference: Statements / selection and jump statements


BillWagner:
This will cover the following statements:

- `if` / `else`
- `switch`
- `break`
- `continue`
- `goto`
- `return`

Sub-tasks are:

- [ ] Create the new articles
- [ ] Search for existing issues on content being migrated and consolidated. This is a good opportunity to fix those.
- [ ] Redirect existing content to the new articles.
- [ ] Remove old articles.
- [ ] Update all links in the .NET Content

BillWagner:
reopening and assigning to @pkulikov

PR #25520, opened 2021-08-08
C# reference: Consolidate the selection statements articles

pkulikov:
Contributes to #10746 (I think it would be better to have two separate articles for selection and jump statements)


docs/csharp/language-reference/compiler-messages/cs0151.md
```diff
@@ -46,25 +46,3 @@ public class MyClass
    }
 }
 ```
-
-## Example of void method
```
pkulikov, line 50: This example doesn't fit the article anymore as it generates CS8119 error

docs/csharp/language-reference/statements/selection-statements.md
```diff
@@ -0,0 +1,107 @@
...
+  - "if keyword [C#]"
+  - "else keyword [C#]"
+  - "switch statement [C#]"
+  - "switch keyword [C#]"
+  - "case keyword [C#]"
+  - "default keyword [C#]"
+---
+# Selection statements (C# reference)
+
+The following statements select statements to execute out of a number of possible statements based on the value of an expression:
```
BillWagner, line 22: typo, I think:

```suggestion
The following statements select statements to execute one of a number of possible statements based on the value of an expression:
```

docs/csharp/language-reference/statements/selection-statements.md
```diff
@@ -0,0 +1,107 @@
...
+
+- An `if` statement without an `else` part executes its body only if a Boolean expression evaluates to `true`, as the following example shows:
+
+  :::code language="csharp" interactive="try-dotnet-method" source="snippets/selection-statements/IfStatement.cs" id="OnlyIf":::
+
+You can nest `if` statements to check multiple conditions, as the following example shows:
+
+:::code language="csharp" interactive="try-dotnet-method" source="snippets/selection-statements/IfStatement.cs" id="NestedIf":::
+
+In an expression context, you can use the [conditional operator `?:`](../operators/conditional-operator.md) to execute one of the two expressions based on the value of a Boolean expression.
```
BillWagner, line 43: nit: expressions are evaluated. statements are executed:

```suggestion
In an expression context, you can use the [conditional operator `?:`](../operators/conditional-operator.md) to evaluate one of the two expressions based on the value of a Boolean expression.
```

docs/csharp/language-reference/statements/selection-statements.md
```diff
@@ -0,0 +1,107 @@
...
+At the preceding example, the `switch` statement uses the following patterns:
+
+- A [relational pattern](../operators/patterns.md#relational-patterns): to compare an expression result with a constant.
+- A [constant pattern](../operators/patterns.md#constant-pattern): to test if an expression result equals a constant.
+
+For information about the patterns supported by the `switch` statement, see [Patterns](../operators/patterns.md).
+
+The preceding example also demonstrates the `default` case. The `default` case specifies statements to execute when a match expression doesn't match any other case pattern. If a match expression doesn't match any case pattern and there is no `default` case, control falls through a `switch` statement.
+
+A `switch` statement evaluates the first section whose case pattern matches a match expression and whose [case guard](#case-guards), if present, evaluates to `true`. A `switch` statement evaluates case patterns in text order from top to bottom. The compiler generates an error when a `switch` statement contains an unreachable case. That is a case that is already handled by an upper case or whose pattern is impossible to match.
```
BillWagner, line 60: Some slight rewording here, also italicizing the spec terms when first introduced.

 ```suggestion
A `switch` statement executes the *statement list* in the first *switch section* whose *case pattern* matches a match expression and whose [case guard](#case-guards), if present, evaluates to `true`. A `switch` statement evaluates case patterns in text order from top to bottom. The compiler generates an error when a `switch` statement contains an unreachable case. That is a case that is already handled by an upper case or whose pattern is impossible to match.
```

File: docs/csharp/language-reference/statements/selection-statements.md [hunk 5]
```diff
@@ -0,0 +1,107 @@
...
+
+> [!NOTE]
+> The `default` case can appear in any place within a `switch` statement. Regardless of its position, the `default` case is always evaluated last and only if all other case patterns aren't matched.
+
+You can specify multiple case patterns for one section of a `switch` statement, as the following example shows:
+
+:::code language="csharp" source="snippets/selection-statements/SwitchStatement.cs" id="MultipleCases":::
+
+Within a `switch` statement, control cannot fall through from one section to the next. As the examples in this section show, typically you use the `break` statement at the end of each section of a `switch` statement to pass control out of a `switch` statement. You can also use the [return](../keywords/return.md) and [throw](../keywords/throw.md) statements to pass control out of a `switch` statement. To imitate the fall-through behavior and pass control to other section of a `switch` statement, you can use the [`goto` statement](../keywords/goto.md).
```
BillWagner, line 69: section -> switch section:

```suggestion
Within a `switch` statement, control cannot fall through from one switch section to the next. As the examples in this section show, typically you use the `break` statement at the end of each switch section of a `switch` statement to pass control out of a `switch` statement. You can also use the [return](../keywords/return.md) and [throw](../keywords/throw.md) statements to pass control out of a `switch` statement. To imitate the fall-through behavior and pass control to other switch section of a `switch` statement, you can use the [`goto` statement](../keywords/goto.md).
```

docs/csharp/language-reference/statements/selection-statements.md
```diff
@@ -0,0 +1,107 @@
...
+
+In an expression context, you can use the [`switch` expression](../operators/switch-expression.md) to evaluate a single expression from a list of candidate expressions based on a pattern match with an expression.
+
+### Case guards
+
+A case pattern may be not expressive enough to specify the condition for the evaluation of the section of a `switch` statement. In such a case, you can use a case guard. That is an additional condition that must be satisfied together with a matched pattern. A case guard must be a Boolean expression. You specify a case guard after the `when` keyword that follows a pattern, as the following example shows:
```
BillWagner, line 75: italicize "case guard" where it's defined:

```suggestion
A case pattern may be not expressive enough to specify the condition for the evaluation of the section of a `switch` statement. In such a case, you can use a *case guard*. That is an additional condition that must be satisfied together with a matched pattern. A case guard must be a Boolean expression. You specify a case guard after the `when` keyword that follows a pattern, as the following example shows:
```

BillWagner (commented):
Thanks @pkulikov  This is really good as well.

I have a few small nits to address, and then this is ready to :shipit:

Commenting so it stays in my queue to review

docs/csharp/language-reference/statements/selection-statements.md
pkulikov: @BillWagner no, not a typo. The `switch` statement executes several statements (all those in one section). `if` statement always selects one, because technically the whole block `{...}` is considered as one embedded statement.

Should it be:
> ...select statements to execute FROM a number of possible statements...

?

pkulikov:
@BillWagner thank you for the review. I've applied all your suggestions except the first one. Please check my latest commit, in which I've made more changes based on your feedback.

docs/csharp/language-reference/statements/selection-statements.md
BillWagner: Thanks for clarifying.  I'll update the suggestion.

BillWagner (approved):
Thanks for putting all this together @pkulikov 

I'll :shipit: now.

PR #25552, opened 2021-08-10
Update version recommendation

### LINKED ISSUE #19070: ".NET Standard" page disagrees with "Cross-platform targeting"
https://docs.microsoft.com/en-us/dotnet/standard/net-standard "Which .NET Standard version to target":

> In general, we recommend you to target the lowest version of .NET Standard possible.

https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting ".NET Standard":

> DO start with including a netstandard2.0 target.
> AVOID including a netstandard1.x target.

Recommendation: rewrite the first page to suggest targeting .NET Standard 2.0, unless the reader need to support older versions, in which case they should multi-target 1.x versions of Standard too. Adding a link to how to accomplish multi-targeting would be helpful.

---
#### Document Details

* ID: 9eb8b774-c6f0-8235-6964-15219baf2ea7
* Version Independent ID: 6819a4a7-7527-d8c7-e132-767f7344e00e
* Content: [.NET Standard](https://docs.microsoft.com/en-us/dotnet/standard/net-standard)
* Content Source: [docs/standard/net-standard.md](https://github.com/dotnet/docs/blob/master/docs/standard/net-standard.md)
* Product: **dotnet**
* GitHub Login: @gewarren

tdykstra:
Fixes #19070

### DIFF
--- a/docs/standard/net-standard.md
+++ b/docs/standard/net-standard.md
@@ -1,7 +1,7 @@
 ---
 title: .NET Standard
 description: Learn about .NET Standard, its versions, and the .NET implementations that support it.
-ms.date: 10/05/2020
+ms.date: 08/10/2021
 ms.prod: dotnet
 ms.technology: dotnet-standard
 ms.custom: "updateeachrelease"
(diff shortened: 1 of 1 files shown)


docs/standard/net-standard.md
```diff
@@ -28,17 +28,9 @@ To find the highest version of .NET Standard that you can target, do the followi
 
 ### Which .NET Standard version to target
 
-When choosing a .NET Standard version to target, consider this trade-off:
+We recommend you target .NET Standard 2.0, unless you need to support an earlier version. Most general-purpose libraries should not need APIs outside of .NET Standard 2.0. .NET Standard 2.0 is supported by all modern platforms and is the recommended way to support multiple platforms with one target.
 
-- The higher the version, the more APIs are available to your library's code.
```
gewarren, line 33: These two bullets seem useful. Do you not think we should keep them?

gewarren (approved):

docs/standard/net-standard.md
tdykstra: I was thinking they are more misleading than helpful, as it's only in exceptional cases that you should consider 1.x anymore.

PR #25567, opened 2021-08-11
Update headers

jwood803:
Update headers in VoTT doc so they appear in the "In this article" section
image


luisquintanilla (approved):
LGTM

PR #25642, opened 2021-08-15
C# reference: Highlight links to the patterns reference

### LINKED ISSUE #25606: Patterns pointers need to be more visible in switch expression page
Was trying to figure out how to use [Type patterns](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/patterns#declaration-and-type-patterns) for the `switch` expression. I started of course at the switch expression document, but didn't see the obvious case I was looking for. It's instead in the `Patterns` page which is only called out earlier amongst a sample here:

```
For information about the patterns supported by a switch expression, see Patterns.
```

And at the end in the related pages. It's not obvious that there's a ton of other samples for different ways to use the switch expression from the existing docs where there's so much more on the Patterns page.

I almost feel like these two pages should be combined, but at the least, it'd be cool if it was more prominently called out all the ways that one can use a switch expression and point to the patterns page for more details.

---
#### Document Details

* ID: fd278c65-14fb-2c27-966d-6e14c74fbcaa
* Version Independent ID: e94a68f9-1a6c-a7e0-e1ee-4be1114bb8dc
* Content: [switch expression - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression)
* Content Source: [docs/csharp/language-reference/operators/switch-expression.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/operators/switch-expression.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

pkulikov:
Fixes #25606

/cc: @michael-hawker

### DIFF
--- a/docs/csharp/language-reference/operators/is.md
+++ b/docs/csharp/language-reference/operators/is.md
@@ -37,7 +37,8 @@ The `is` operator can be useful in the following scenarios:
 
   :::code language="csharp" source="snippets/shared/IsOperator.cs" id="NonNullCheck":::
 
-For the complete list of patterns supported by the `is` operator, see [Patterns](patterns.md).
+> [!NOTE]
+> For the complete list of patterns supported by the `is` operator, see [Patterns](patterns.md).
 
 ## C# language specification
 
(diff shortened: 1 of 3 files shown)


IEvangelist (approved):

PR #25656, opened 2021-08-16
Link to manual install docs from "install other versions"

### LINKED ISSUE #21878: Install on Linux docs are verbose
For example: https://docs.microsoft.com/en-us/dotnet/core/install/linux-fedora, but applies to all Linux flavors.

What users are looking for is:
```
$ sudo dnf install dotnet-sdk-5.0
```
This should be at the top of the page (2nd or 3rd paragrah).

Instructions across Fedora versions are very similar. There is no reason to split them up and duplicate the same instructions for each version. The repo file should use a placeholder, like: `baseurl=https://packages.microsoft.com/fedora/$releasever/prod/`.
Fedora versions that are no longer supported should be removed (vs _❌ Please note that this version of Fedora is no longer supported._).

Do these sections apply to Fedora:
* Remove preview versions
* Troubleshoot the package manager

These sections feels out of place in the installing document:
* Dependencies
* Next steps

These sections are duplicated across many pages, and provide an alternate non Fedora-specific way of installing .NET. They could live in their own doc(s):
* Snap
* Scripted install
* Manual install

cc @omajid @adegeo @tdykstra @RheaAyase @dagood

adegeo:
## Summary

- Added a paragraph about installing other versions.

Final update for #21878

Fixes #21878

### DIFF
--- a/docs/core/install/includes/package-manager-heading-hack-pkgname.md
+++ b/docs/core/install/includes/package-manager-heading-hack-pkgname.md
@@ -1,5 +1,7 @@
 
-The packages added to package manager feeds are named in a hackable format: `{product}-{type}-{version}`.
+All versions of .NET are available for download at <https://dotnet.microsoft.com/download/dotnet>, but require [manual installation](../linux-scripted-manual.md). You can try and use the package manager to install a different version of .NET. However, the requested version may not be available.
+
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @adegeo 

You can :shipit: when ready.

PR #25657, opened 2021-08-16
run-time config to runtime config

gewarren:
I've always struggled with how to hyphenate this. Santi provided the following explanation as to why it shouldn't be hyphenated:

"[We are configuring] how [the app] behaves at run time, however the feature is called `runtime switch` because are read by the host and then the runtime sets them and we fetch them from the runtime."


tdykstra (approved):
LGTM

gewarren:
Thanks Tom!

PR #25685, opened 2021-08-17
OutputType breaking change

### LINKED ISSUE #25636: [Breaking change]: OutputType no longer automatically set to WinExe for WPF and WinForms apps
### Description

In .NET 5, we [automatically converted](https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/5.0/automatically-infer-winexe-output-type) an `OutputType` of `Exe` to `WinExe` for WPF and Windows forms apps.  This was intended to simplify .NET MAUI apps (so that the OutputType wouldn't need to be conditioned on the TargetFramework).

However:

- This broke user expectations and frustrated some folks.  See https://github.com/dotnet/sdk/issues/16563 and other linked issues
- .NET MAUI apps are going to by default use WinUI, not Windows Forms or WPF, so this logic won't even apply to them

So we are reverting to the original behavior in .NET 6.

### Version

.NET 6 RC1

### Previous behavior

If a project targeted .NET 5 or higher, the `OutputType` was `Exe`, and `UseWindowsForms` or `UseWPF` was set to true, then the .NET SDK would change the `OutputType` to `WinExe`.

### New behavior

The `OutputType` is no longer changed from what is in the project file.

### Type of breaking change

### Reason for change

The [previous breaking change](https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/5.0/automatically-infer-winexe-output-type) was more impactful than expected and ended up not being the right logic for .NET Maui.

### Recommended action

If you relied on the fact that the `OutputType` was changed from `Exe` to `WinExe`, you will now need to explicitly set it to `WinExe`.

If you were impacted by the previous breaking change and had to set `DisableWinExeOutputInference` in order to disable the logic that was added in .NET 5, you can remove that property.

### Feature area

SDK

### Affected APIs

_No response_

gewarren:
Fixes #25636 

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/6.0/outputtype-not-set-automatically?branch=pr-en-us-25685).

### DIFF
--- a/docs/core/compatibility/sdk/6.0/outputtype-not-set-automatically.md
+++ b/docs/core/compatibility/sdk/6.0/outputtype-not-set-automatically.md
@@ -13,11 +13,17 @@ In .NET 5, a change was made to automatically change `OutputType` from `Exe` to 
 
 ## Previous behavior
 
-If a project targeted .NET 5 or higher, the `OutputType` was `Exe`, and `UseWindowsForms` or `UseWPF` was set to `true`, then the .NET SDK would change the `OutputType` to `WinExe`.
+If a project targeted .NET 5 or higher, `OutputType` was set to `Exe`, and `UseWindowsForms` or `UseWPF` was set to `true`, then the .NET SDK would change `OutputType` to `WinExe`.
 
 ## New behavior
 
 `OutputType` is no longer changed from what's in the project file.
+
+```xml
+<PropertyGroup>
+  <OutputType>Exe</OutputType>
+</PropertyGroup>
+```
 
 ## Change category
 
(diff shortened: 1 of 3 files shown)


docs/core/compatibility/sdk/6.0/outputtype-not-set-automatically.md
```diff
@@ -0,0 +1,51 @@
+---
+title: "Breaking change: `OutputType` no longer automatically set to WinExe"
+description: Learn about the breaking change in .NET 6 where `OutputType` for WPF and Windows Forms projects is no longer automatically set to `WinExe`.
+ms.date: 08/17/2021
+---
+# OutputType not automatically set to WinExe
```
dsplaisted, line 6: This may be a bit clearer about the scope of the change:

```suggestion
# OutputType not changed from Exe to WinExe for Windows Forms and WPF projects
```

dsplaisted (approved):

twsouthwick:
@gewarren Will this revert it for apps building on .NET 6 SDK but targeting .NET 5?

dsplaisted:
> @gewarren @dsplaisted Will this revert it for apps building on .NET 6 SDK but targeting .NET 5?

Yes, it will.  The templates for .NET 5 still used `WinExe`, so I don't know if anyone would have taken a dependency on the fact that we would change `Exe` to `WinExe`.

PR #25712, opened 2021-08-19
List multi-module assemblies as an unsupported technology

MichalStrehovsky:
I couldn't find this being spelled out anywhere and this seems like a good spot.

See e.g. https://github.com/dotnet/runtime/issues/10571 for details.


gewarren:
@jkoritzinsky Could you review this one?

jkoritzinsky (approved):

Issue #25486, opened 2021-08-05
Content health: Resolve build suggestions


gewarren:
Currently, we have 850 build suggestions in the dotnet/docs repo. See https://msit.powerbi.com/groups/9d83d204-82a9-4b36-98f2-a40099093839/reports/68fd9d81-2568-4c0a-841f-a2d2637732d0/ReportSectionee2c78cb596204c7b1e3. We need to resolve these suggestions to improve our overall content health score as part of our OKRs

Warnings are taken care of by #25322.

Track progress here: https://msit.powerbi.com/groups/9d83d204-82a9-4b36-98f2-a40099093839/reports/68fd9d81-2568-4c0a-841f-a2d2637732d0/ReportSectionee2c78cb596204c7b1e3

PR #25731, opened 2021-08-19
Add column headers (VB)

gewarren:
Contributes to #25486 

- Adds column headers
- Or converts tables into columns

[Preview link for columns](https://review.docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/keywords/?branch=pr-en-us-25731)
[Preview link for data matrix with code fencing](https://review.docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/operators/data-types-of-operator-results?branch=pr-en-us-25731#-and--operators)

### DIFF
--- a/docs/visual-basic/language-reference/attributes.md
+++ b/docs/visual-basic/language-reference/attributes.md
@@ -8,14 +8,14 @@ ms.assetid: 5deb2b8a-1afd-4dbd-8ee8-f093d74ad0eb
 ---
 # Attributes (Visual Basic)
 
-Visual Basic provides several attributes that allow objects to interoperate with unmanaged code and one attribute that enables module members to be accessed without the module name. The following table lists the attributes used by Visual Basic.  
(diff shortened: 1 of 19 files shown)


BillWagner (approved):
This LGTM @gewarren 

One note on the columns: @tdykstra had come up with a neat trick to use one row, and put line breaks between each element in the same column. The presentation was the same as the standard column layout, but it was a lot less verbose. You can see an example [here](https://github.com/dotnet/docs/blame/main/docs/csharp/language-reference/keywords/index.md#L20).

gewarren:
> This LGTM @gewarren
> 
> One note on the columns: @tdykstra had come up with a neat trick to use one row, and put line breaks between each element in the same column. The presentation was the same as the standard column layout, but it was a lot less verbose. You can see an example [here](https://github.com/dotnet/docs/blame/main/docs/csharp/language-reference/keywords/index.md#L20).

That's nifty! The only problem is that you have to read down then across instead of across then down.

PR #25733, opened 2021-08-19
Fix view links with suggestions

gewarren:
Add `preserve-view` part.

Contributes to #25486


tdykstra (commented):
Rather than view=false, they recommend removing the view parameter if you don't want view=true.

https://review.docs.microsoft.com/en-us/help/contribute/validation-ref/preserve-view-not-set?branch=master#resolution

> If it's not essential to link to a specific version of the target page, remove the view parameter from your link, such as https://docs.microsoft.com/sql/t-sql/functions/cast-and-convert-transact-sql.

tdykstra (approved):

PR #25392, opened 2021-07-29
Fixes warnings in .NET architecture ebooks

sughosneo:
Fixes: https://github.com/dotnet/docs/issues/25380

PR #25735, opened 2021-08-19
Fix suggestions in Architecture folder

gewarren:
Contributes to #25486 

- Add column headers
- Replace absolute docs.microsoft.com links

(Note: this PR doesn't address the warnings. Those are addressed by #25392.)


IEvangelist (approved):

PR #25743, opened 2021-08-20
Fix a (I think?) typo in class-invariance sample-code

ELLIOTTCABLE:
## Summary

Shouldn't this be attempting to assign to a *less* derived type, to demonstrate the invariance? The current code assigns to the same type twice.


BillWagner:
Thanks for noting this @ELLIOTTCABLE 

In this case, the code was correct as written.  The `new` expression used `SampleImplementation<Button>`, and the assignment used `SampleImplementation<Object>`. Note that the type argument changed from `Button` to `Object`. That generates an error because the type argument is invariant for `SampleImplementation<T>`. 

I'll close this.

ELLIOTTCABLE:
Oh my god, I'm blind.

Sorry to have wasted your time! 😩

PR #25744, opened 2021-08-20
Update nullable-migration-strategies.md

rbgnk:
The documentation used an example with a method called TryGetMessage and later referred back to it as TryGetValue. This updates the method's name so it's consistent.


BillWagner (approved):
This for spotting this and fixing it @rbgnk 

PR #25749, opened 2021-08-20
Final suggestions except for csharplang

gewarren:
Contributes to #25486


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready

PR #25751, opened 2021-08-20
Explain --version-suffix

### LINKED ISSUE #25742: Better explain the intricacies of the `dotnet pack --version-suffix`
I was hoping to use `dotnet pack --version-suffix rc` to easily give my project in a CI a release candidate suffix, without needing to know the project version beforehand.

Due to https://github.com/dotnet/sdk/issues/1131 this did not work as well as I'd hoped, in fact the `--version-suffix` proper isn't able to help at all in this scenario. From what I understand, you instead have to go low-level and [manually set](https://github.com/dotnet/sdk/issues/19713#issuecomment-902460324) the `VersionPrefix` and `VersionSuffix` properties.

It would be great to see the docs improved in this area, since it's a bit of a pitfall.

---
#### Document Details

* ID: a7e805a3-f2ec-b05a-d5c8-29f7a8c16a5f
* Version Independent ID: 4257ec12-df6f-ec52-aad3-012e5fce8133
* Content: [dotnet pack command - .NET CLI](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-pack)
* Content Source: [docs/core/tools/dotnet-pack.md](https://github.com/dotnet/docs/blob/main/docs/core/tools/dotnet-pack.md)
* Technology: **dotnet-cli**
* GitHub Login: @tdykstra

tdykstra:
Fixes #25742 

-- | -- | --
Content | docs/core/tools/dotnet-pack.md | https://review.docs.microsoft.com/en-us/dotnet/core/tools/dotnet-pack?branch=pr-en-us-25751

### DIFF
--- a/docs/core/tools/dotnet-pack.md
+++ b/docs/core/tools/dotnet-pack.md
@@ -112,6 +112,8 @@ Web projects aren't packable by default. To override the default behavior, add t
 
   Sets the serviceable flag in the package. For more information, see [.NET Blog: .NET Framework 4.5.1 Supports Microsoft Security Updates for .NET NuGet Libraries](https://aka.ms/nupkgservicing).
 
+[!INCLUDE [verbosity](../../../includes/cli-verbosity.md)]
+
 - **`--version-suffix <VERSION_SUFFIX>`**
 
   Defines the value for the `VersionSuffix` MSBuild property. The effect of this property on the package version depends on the values of the `Version` and `VersionPrefix` properties, as shown in the following table:
@@ -127,8 +129,6 @@ Web projects aren't packable by default. To override the default behavior, add t
   If you want to use `--version-suffix`, specify `VersionPrefix` and not `Version` in the project file. For example, if `VersionPrefix` is `0.1.2` and you pass `--version-suffix rc1` to `dotnet pack`, the package version will be `0.1.2-rc1`.
 
   If `Version` has a value and you pass `--version-suffix` to `dotnet pack`, the value specified for `--version-suffix` is ignored.
-
-[!INCLUDE [verbosity](../../../includes/cli-verbosity.md)]
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @tdykstra 

You can :shipit: once you and @dsplaisted are satisfied.

dsplaisted (approved):
This is great and a big improvement.

However, we also have documentation for MSBuild properties, don't we?  Should we also have this information there, and link to it from this CLI documentation?

tdykstra:
>we also have documentation for MSBuild properties

For packaging-related properties [our MS Build properties doc](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#package-properties) links to a [NuGet doc](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#pack-target). The NuGet doc doesn't explain all of the ways these properties interact -- I'll log an issue for that, but meanwhile at least until that gets addressed I could leave this explanation here and link to the NuGet doc for more information.

tdykstra:
Actually, the NuGet doc seems to be talking about something different from what we use with `dotnet pack` that is linked to in https://github.com/dotnet/sdk/issues/19713#issuecomment-902460324  That being the case we don't have a better place for this documentation right now. 

@gewarren should we have our own packaging-related documentation in https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#package-properties rather than linking out to NuGet docs?

gewarren:
> Actually, the NuGet doc seems to be talking about something different from what we use with `dotnet pack` ([dotnet/sdk#19713 (comment)](https://github.com/dotnet/sdk/issues/19713#issuecomment-902460324)). That being the case we don't have a better place for this documentation right now.

I'm wondering if the "MSBuild property" names at https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#pack-target are even correct? I.e. `PackageVersion`, `PackageVersionPrefix`, and `PackageVersionSuffix`. [This article](https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/) says that, for example, the MSBuild properties `Version`, `VersionPrefix`, and `VersionSuffix` are used to set the value of `PackageVersion`, which is only visible in the build output.

Issue #25672, opened 2021-08-17
Extended property patterns


BillWagner:
[Feature specification](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/extended-property-patterns.md)

 This would enable `.` in a property pattern to match on the property of a property of the object (of any depth)

BillWagner:
To address this issue, the following pages should be updated:

- [Language Reference - Patterns](https://docs.microsoft.com/dotnet/csharp/language-reference/operators/patterns): This page must be updated to include the new property pattern syntax as a new section.
- [Pattern matching tutorial in fundamentals](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching) New examples for property patterns should be added.
- [Data driven algorithms with pattern matching](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/tutorials/pattern-matching) look for where extended property patterns should be used.
-

PR #25763, opened 2021-08-22
C# 10 reference update: Extended property pattern

pkulikov:
Contributes to #25672 (only reference article update)
(I've incorporated the update into the existing section, as it looks like an enhancement of an existing property pattern.)


BillWagner (approved):
This LGTM @pkulikov 

Thanks for submitting this fix.

I'll :shipit: now.

PR #25775, opened 2021-08-23
Provide Explicit Clarity

mikelle-rogers:
Altered text in diagnostic-port discussion to create clarity.

## Summary

Changed the pronoun it to the application to ensure clarity.

Done as part of the DevDiv Onboarding Quest


tommcdon (approved):
Good catch!  Thanks!

IEvangelist (approved):

PR #25807, opened 2021-08-25
Update guidance in classes.md

dsyme:
The guidance for which kinds of types to use can be read as implying classes are **only** for interop.  This is not the case - when writing F# code you should use an appropriate mix of classes, unions and records (with structs for performance).

See https://twitter.com/OnurGumusDev/status/1430596304812531714 and my reply

cc @cartermp


BillWagner (approved):
Thanks @dsyme 

I'll :shipit: now.

(And I'll open a separate PR for the linting error)

PR #25813, opened 2021-08-25
Expand the explanation of SupportedOSPlatformVersion

mhutch:
Expanded the explanation of SupportedOSPlatformVersion and the OS version in the TFM.

There's nontrivial chance this has errors so I've added a number of technical reviewers who have worked with related specs and code.


dsplaisted (approved):

docs/standard/frameworks.md
```diff
@@ -80,16 +80,28 @@ Use these guidelines to determine which TFM to use in your app:
...
 
-For example, the following project file excerpt specifies that iOS 14 APIs are available to the app, but it can run on iOS 13 or later machines.
+When an OS-specific TFM does not specify the platform version explicitly, it has a implied value that can be inferred from the base TFM and platform name. For example, the default platform value for iOS in .NET 6.0 is `15.0`, which means that `net6.0-ios` is shorthand for the canonical `net6.0-ios15.0` TFM. The implied platform version for a newer base TFM may be higher, for example a future `net7.0-ios` TFM could map to `net7.0-ios16.0`. The shorthand form is intended for use in project files only, and is expanded to the canonical form by the .NET SDK's MSBuild targets before being passed to other tools such as NuGet.
```
gewarren, line 85:
```suggestion
When an OS-specific TFM doesn't specify the platform version explicitly, it has an implied value that can be inferred from the base TFM and platform name. For example, the default platform value for iOS in .NET 6 is `15.0`, which means that `net6.0-ios` is shorthand for the canonical `net6.0-ios15.0` TFM. The implied platform version for a newer base TFM may be higher, for example, a future `net7.0-ios` TFM could map to `net7.0-ios16.0`. The shorthand form is intended for use in project files only, and is expanded to the canonical form by the .NET SDK's MSBuild targets before being passed to other tools, such as NuGet.
```

docs/standard/frameworks.md
```diff
@@ -80,16 +80,28 @@ Use these guidelines to determine which TFM to use in your app:
...
+
+The .NET SDK is designed to be able to support newly released APIs for an individual platform without a new version of the base TFM. This enables you to access platform-specific functionality without waiting for a major release of .NET. You can gain access to these newly released APIs by incrementing the platform version in the TFM. For example, if the iOS platform added iOS 15.1 APIs in a .NET 6.0.x SDK update, you could access them by using the TFM `net6.0-ios15.1`.
+
+#### Supporting older OS versions
```
gewarren, line 89:
```suggestion
#### Support older OS versions
```

docs/standard/frameworks.md
```diff
@@ -80,16 +80,28 @@ Use these guidelines to determine which TFM to use in your app:
...
+
+Although a platform-specific app or library is compiled against APIs from a specific version of that OS, you can make it compatible with earlier OS versions by adding the `SupportedOSPlatformVersion` element to your project file. The `SupportedOSPlatformVersion` element indicates the minimum OS version required to run your app or library. If this minimum runtime OS version is not specified explicitly in the project, it defaults to the platform version from the TFM.
```
gewarren, line 91:
```suggestion
Although a platform-specific app or library is compiled against APIs from a specific version of that OS, you can make it compatible with earlier OS versions by adding the `SupportedOSPlatformVersion` property to your project file. The `SupportedOSPlatformVersion` property indicates the minimum OS version required to run your app or library. If you don't explicitly specify this minimum run-time OS version in the project, it defaults to the platform version from the TFM.
```

docs/standard/frameworks.md
```diff
@@ -80,16 +80,28 @@ Use these guidelines to determine which TFM to use in your app:
...
+
+To run correctly on an older OS version, you cannot call APIs that do not exist on that version of the OS. However, you can add guards around calls to newer APIs so they are only called when running on a version of the OS that supports them. This pattern allows you to design your app or library to support running on older OS versions while taking advantage of newer OS functionality when running on newer OS versions.
```
gewarren, line 93:
```suggestion
For your app to run correctly on an older OS version, it can't call APIs that don't exist on that version of the OS. However, you can add guards around calls to newer APIs so they are only called when running on a version of the OS that supports them. This pattern allows you to design your app or library to support running on older OS versions while taking advantage of newer OS functionality when running on newer OS versions.
```

docs/standard/frameworks.md
```diff
@@ -80,16 +80,28 @@ Use these guidelines to determine which TFM to use in your app:
...
+
+The `SupportedOSPlatformVersion` value (whether explicit or default) is used by the [platform compatibility analyzers](analyzers/platform-compat-analyzer.md), which detect and warn about unguarded calls to newer APIs. It is burned into the project's compiled assembly as a <xref:System.Runtime.Versioning.UnsupportedOSPlatformAttribute> assembly attribute so that the platform compatibility analyzers can detect unguarded calls to that assembly's APIs from projects with a lower `SupportedOSPlatformVersion` value. On some platforms it affects platform-specific app packaging and build processes, which is covered in the documentation for those platforms.
```
gewarren, line 95:
```suggestion
The `SupportedOSPlatformVersion` value (whether explicit or default) is used by the [platform compatibility analyzer](analyzers/platform-compat-analyzer.md), which detects and warns about unguarded calls to newer APIs. It's burned into the project's compiled assembly as an <xref:System.Runtime.Versioning.UnsupportedOSPlatformAttribute> assembly attribute, so that the platform compatibility analyzer can detect unguarded calls to that assembly's APIs from projects with a lower `SupportedOSPlatformVersion` value. On some platforms, the `SupportedOSPlatformVersion` value affects platform-specific app packaging and build processes, which is covered in the documentation for those platforms.
```

gewarren (approved):

PR #25818, opened 2021-08-26
Remove core clr hosting

vitek-karas:
Hosting .NET through `coreclr.h` has been obsolete since .NET Core 3.0 and should not be used.

All the versions which needed it (.NET Core 2.2 and older) are now out of support.

Having the `coreclr.h` hosting available in the docs and samples is confusing and may lead people to actually use it for newer runtimes which can lead to unexpected results.

This change removes the sample code and modifies the hosting guide to only mention one hosting API. I added a note about the older runtimes needing to use `coreclr.h` but intentionally didn't include any details on that topic.

It also updates the `hostfxr` sample with a fresh copy of the sample from the samples repo.

Corresponding changes to samples: https://github.com/dotnet/samples/pull/4786.


elinor-fung (commented):
Would it make sense to do a general `.NET Core runtime` -> `.NET runtime` conversion (except for where it is explicitly `.NET Core 3.0`) throughout this tutorial?

mateoatr (approved):

elinor-fung (approved):

Earthman100:
Thats what dot net is made from! no outdating that!

PR #25875, opened 2021-08-31
Remove package management file

zedy-wj:
1. Remove package management file, it is no longer used.


zedy-wj:
Hi, @jsquire - The ci checks failed after deleting package-management.md file, it seems like it will to verify the existence of file package-management.md in the container.
image

I don’t have permission to pull the image and modify the content, can you help to take a look?

jsquire:
This appears to be related to the docs tools and, apparently, a process for ensuring a redirect is in-place.    I don't have insight, but @jongio may.

jongio:
I would prefer that we keep the page and update it for general guidance on how to consume packages.  Even if it is a repeat of what is in each of the other docs.

zedy-wj:
@jsquire ,@jongio - We now have two thoughts to update the package-management.md file.

Case1 :  Just update the content of T1 in the file to T2, do not modify the package manager of paket and nuget.

Case2 : Remove paket and nuget, just introduce how to install the package like other docs.

Note: If case1 is used, paket and nuget are out of date, and more concise installation package methods have been updated in other docs. In the other hand, if we use case2, it does not reflect the concept of package manager, the file name and it's content are inconsistent, so we need to update it's file name at the same time.

Any suggestion or better idea on this?

jsquire:
I defer to @jongio for the decision.

That said, my personal thought is that we should either remove the package manager documentation or give it an overhaul to modernize it.  The focus here isn't T1 => T2, but rather capturing changes in the .NET ecosystem - particularly emphasizing using the `dotnet` CLI to replace legacy executables and such.

However, given the focus of the samples themselves are around using F# scripts, I question whether there's value in documenting the package manager, since it isn't used and is orthogonal to the samples.  I don't think that we want to be in the business of documenting a core ecosystem concern like package management in isolation; better to refer to a common set of documentation that isn't Azure SDK specific for these concepts.

jongio:
@CamSoper - Can you please review this and help us make a decision?

FYI @jongalloway

CamSoper:
All things being equal, I support the removal of the doc and we just redirect to a relevant place, like an overview page or something. The lint for redirects is just to ensure that we don't return 404s. It's much more polite to return a 301 that goes anywhere than a 404. 

You can make the required changes to `.openpublishing.redirection.json` in the root of this repo. The format is pretty self-explanatory, just try to put it in a spot where it makes sense for other people to find it later.

zedy-wj:
Updated, we prepared a new pull request : #25969, and will close this one.

PR #25880, opened 2021-08-31
Some improvements to the single-file publish doc

vitek-karas:
* Most importantly improve the description of how files are extracted to disk for all platforms.
* Removed the "temporary" work for files extraction as on Linux/MacOS there's nothing temporary about the default extraction directory
* Removed PublishTrimmed=true in the sample - in .NET 5 this is dangerous, and in .NET 6 it will typically lead to lot of warnings for most apps. The doc already mentions the possibility to use trimming with single-file anyway.
* Include changes in behavior in .NET 6 in the text (as we're starting to get lot of questions about .NET 6 single-file difference).

### DIFF
--- a/docs/core/deploying/single-file.md
+++ b/docs/core/deploying/single-file.md
@@ -7,7 +7,7 @@ ms.date: 08/10/2021
 ---
 # Single file deployment and executable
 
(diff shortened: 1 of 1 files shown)


IEvangelist (approved):

PR #25887, opened 2021-08-31
Add another example to dotnet-workload-install.md

LoopedBard3:
## Summary

Add an example showing the installation of multiple workloads with dotnet workload install.


tdykstra (approved):
Good idea to include a multiple-workload example, thanks!

PR #25913, opened 2021-09-02
F# can't produce default interface methods

dunkyl:
The remark section on the [page about F# interfaces](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/interfaces) seems to allude to a capability that I don't think it has:

> However, you can provide a default implementation by also including a separate definition of the member as a method together with the default keyword. Doing so is equivalent to creating a virtual method in a base class in other .NET languages. Such a virtual method can be overridden in classes that implement the interface.

I may be misinterpreting this; I was surprised to see the git blame for this paragraph is quite old. F#5 can use C# default interface methods, but it cannot provide them. 

Related: [https://github.com/fsharp/fslang-suggestions/issues/679](https://github.com/fsharp/fslang-suggestions/issues/679)

If the existing remark is in fact wrong, I would of course like feedback on the wording of the clarification (in PR).

### DIFF
--- a/docs/fsharp/language-reference/interfaces.md
+++ b/docs/fsharp/language-reference/interfaces.md
@@ -38,7 +38,7 @@ let class-name (argument-list) =
 
 ## Remarks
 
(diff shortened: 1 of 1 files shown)


KathleenDollard:
@dsyme For review

dsyme (dismissed):

dsyme (approved):

PR #25925, opened 2021-09-02
Azsdk overview page update

DavidCBerry13:
Updated the .NET Framework version from 4.6.1 to 4.7.2 because .NET Standard libraries have known issues working with 4.6.1
Updated the image with the new Azure logo


IEvangelist (approved):

Actually, wait... pump the breaks:

> Updated the .NET Framework version from 4.6.1 to 4.7.2 because .NET Standard libraries have known issues working with 4.6.1

What? .NET Standard 2.0 is supposed to be implemented by .NET 4.61.

https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-implementation-support

I know that it calls attention to this "2 The versions listed here represent the rules that NuGet uses to determine whether a given .NET Standard library is applicable. While NuGet considers .NET Framework 4.6.1 as supporting .NET Standard 1.5 through 2.0, there are several issues with consuming .NET Standard libraries that were built for those versions from .NET Framework 4.6.1 projects. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher." But what specific issues are you seeing, and could we add them here?

PR #25933, opened 2021-09-02
Add ComWrappers tutorial

AaronRobinsonMSFT:
## Summary

Add a tutorial on using the `ComWrappers` API introduced in .NET 5.

Source for tutorial is in PR at https://github.com/dotnet/samples/pull/4798.

/cc @gewarren @jkoritzinsky @elinor-fung @agocke @sbomer @vitek-karas

### DIFF
--- a/docs/fundamentals/toc.yml
+++ b/docs/fundamentals/toc.yml
@@ -2472,12 +2472,10 @@ items:
           href: ../standard/native-interop/runtime-callable-wrapper.md
         - name: COM-callable wrapper
           href: ../standard/native-interop/com-callable-wrapper.md
+        - name: Tutorial: Use the ComWrappers API
+          href: ../standard/native-interop/tutorial-comwrappers.md
       - name: Qualifying .NET types for COM interop
         href: ../standard/native-interop/qualify-net-types-for-interoperation.md
-      - name: Tutorials
-        items:
-        - name: Using the ComWrappers API
-          href: ../standard/native-interop/tutorial-comwrappers.md
     - name: Apply interop attributes
       href: ../standard/native-interop/apply-interop-attributes.md
     - name: Exceptions
--- a/docs/fundamentals/toc.yml
+++ b/docs/fundamentals/toc.yml
@@ -2472,7 +2472,7 @@ items:
           href: ../standard/native-interop/runtime-callable-wrapper.md
         - name: COM-callable wrapper
           href: ../standard/native-interop/com-callable-wrapper.md
-        - name: Tutorial: Use the ComWrappers API
+        - name: Tutorial - Use the ComWrappers API
           href: ../standard/native-interop/tutorial-comwrappers.md
       - name: Qualifying .NET types for COM interop
         href: ../standard/native-interop/qualify-net-types-for-interoperation.md
--- a/docs/fundamentals/toc.yml
+++ b/docs/fundamentals/toc.yml
@@ -2475,6 +2475,10 @@ items:
           href: ../standard/native-interop/com-callable-wrapper.md
       - name: Qualifying .NET types for COM interop
         href: ../standard/native-interop/qualify-net-types-for-interoperation.md
+      - name: Tutorials
+        items:
+        - name: Using the ComWrappers API
+          href: ../standard/native-interop/tutorial-comwrappers.md
     - name: Apply interop attributes
       href: ../standard/native-interop/apply-interop-attributes.md
     - name: Exceptions
--- a/docs/standard/native-interop/tutorial-comwrappers.md
+++ b/docs/standard/native-interop/tutorial-comwrappers.md
@@ -1,3 +1,15 @@
+---
+title: "Using the ComWrappers API"
+description: A tutorial on properly implementing the ComWrappers API.
+author: arobins
+ms.date: "02/09/2021"
+helpviewer_keywords:
+  - "COM interop, ComWrappers"
+  - "RCW"
+  - "CCW"
+  - "interoperation with unmanaged code, COM wrappers"
+  - "NativeAOT"
+---
 # Using the `ComWrappers` API
 
 In this tutorial we are going to see how to properly subclass the [`ComWrappers`][api_comwrappers] type to provide an optimized and AOT friendly COM interop solution. It is assumed the reader of this document is familiar COM, its architecture, and existing COM interop solutions.
@@ -375,22 +387,22 @@ Aside from the lifetime, type system, and functional features that are discussed
 
 <!-- Reusable links -->
 
-[api_allocatetypeassociatedmemory]:https://docs.microsoft.com/dotnet/api/system.runtime.compilerservices.runtimehelpers.allocatetypeassociatedmemory
-[api_comwrappers]:https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.comwrappers
-[api_dynamicinterfacecastableimplementation]:https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.dynamicinterfacecastableimplementationattribute
-[api_idynamicinterfacecastable]:https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.idynamicinterfacecastable
-[api_marshalqueryinterface]:https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.marshal.queryinterface
-[api_referencetracker]:https://docs.microsoft.com/windows/win32/api/windows.ui.xaml.hosting.referencetracker/
-[api_rogetagilereference]:https://docs.microsoft.com/windows/win32/api/combaseapi/nf-combaseapi-rogetagilereference
-[api_unmanagedcallersonly]:https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.unmanagedcallersonlyattribute
+[api_allocatetypeassociatedmemory]:/dotnet/api/system.runtime.compilerservices.runtimehelpers.allocatetypeassociatedmemory
+[api_comwrappers]:/dotnet/api/system.runtime.interopservices.comwrappers
+[api_dynamicinterfacecastableimplementation]:/dotnet/api/system.runtime.interopservices.dynamicinterfacecastableimplementationattribute
+[api_idynamicinterfacecastable]:/dotnet/api/system.runtime.interopservices.idynamicinterfacecastable
+[api_marshalqueryinterface]:/dotnet/api/system.runtime.interopservices.marshal.queryinterface
+[api_referencetracker]:/windows/win32/api/windows.ui.xaml.hosting.referencetracker/
+[api_rogetagilereference]:/windows/win32/api/combaseapi/nf-combaseapi-rogetagilereference
+[api_unmanagedcallersonly]:/dotnet/api/system.runtime.interopservices.unmanagedcallersonlyattribute
 
-[doc_comapartments]:https://docs.microsoft.com/windows/win32/com/processes--threads--and-apartments
-[doc_comsecurity]:https://docs.microsoft.com/windows/win32/com/security-in-com
-[doc_globalinterfacetable]:https://docs.microsoft.com/windows/win32/com/when-to-use-the-global-interface-table
-[doc_impliunknown]:https://docs.microsoft.com/windows/win32/com/using-and-implementing-iunknown
-[doc_nullable]:https://docs.microsoft.com//dotnet/csharp/nullable-references
-[doc_unloadability]:https://docs.microsoft.com/dotnet/standard/assembly/unloadability
-[doc_unsafekeyword]:https://docs.microsoft.com/dotnet/csharp/language-reference/keywords/unsafe
+[doc_comapartments]:/windows/win32/com/processes--threads--and-apartments
+[doc_comsecurity]:/windows/win32/com/security-in-com
+[doc_globalinterfacetable]:/windows/win32/com/when-to-use-the-global-interface-table
+[doc_impliunknown]:/windows/win32/com/using-and-implementing-iunknown
+[doc_nullable]:/dotnet/csharp/nullable-references
+[doc_unloadability]:/dotnet/standard/assembly/unloadability
+[doc_unsafekeyword]:/dotnet/csharp/language-reference/keywords/unsafe
 
 [repo_cswinrt]:https://github.com/microsoft/CsWinRT
 
--- a/docs/standard/native-interop/tutorial-comwrappers.md
+++ b/docs/standard/native-interop/tutorial-comwrappers.md
@@ -48,26 +48,26 @@ IDemoStoreType : public IUnknown
 };
 ```
 
-## Overview of the `ComWrappers` design.
+## Overview of the `ComWrappers` design
 
-The [`ComWrappers`][api_comwrappers] API was designed to provide the minimal runtime interaction needed to accomplish COM interop with the .NET Core runtime. This means that many of the niceties that exist with the build-in COM interop system are not present and must be built up from basic building blocks. The two primary responsibilities of the API are:
+The [`ComWrappers`][api_comwrappers] API was designed to provide the minimal runtime interaction needed to accomplish COM interop with the .NET 5+ runtime. This means that many of the niceties that exist with the build-in COM interop system are not present and must be built up from basic building blocks. The two primary responsibilities of the API are:
 
 - Enable efficient object identity (for example, mapping between an `IUnknown*` instance and a managed object).
 - Efficient Garbage Collector (GC) interaction.
 
 The above are accomplished by requiring wrapper creation and acquisition to go through the `ComWrappers` API.
 
-Since the `ComWrappers` API has so few responsibilities it stands to reason the majority of the interop work is to be handled by the consumer &ndash; this is true. However, the additional work is largely mechanical and is expected to be performed by a source generation solution. As an example, the [C#/WinRT tool chain][repo_cswinrt] is a source generator solution that is built on top of `ComWrappers` to provide WinRT interop.
+Since the `ComWrappers` API has so few responsibilities it stands to reason the majority of the interop work is to be handled by the consumer &ndash; this is true. However, the additional work is largely mechanical and is expected to be performed by a source generation solution. As an example, the [C#/WinRT tool chain][repo_cswinrt] is a source generator solution that is built on top of `ComWrappers` to provide WinRT interop support.
 
-## Implementing a `ComWrappers` subclass.
+## Implementing a `ComWrappers` subclass
 
 Providing a `ComWrappers` subclass means giving enough information to the .NET runtime in order to create and record wrappers for managed objects being projected into COM and COM objects being projected into .NET. Before we look at an outline of the subclass we should define some terms.
 
-**Managed Object Wrapper** &ndash; Managed .NET objects require wrappers to enable consumption from a non-.NET environment. These wrappers are historically called COM Callable Wrapper (**CCW**).
+**Managed Object Wrapper** &ndash; Managed .NET objects require wrappers to enable consumption from a non-.NET environment. These wrappers are historically called COM Callable Wrappers (**CCW**).
 
-**Native Object Wrapper** &ndash; COM objects that are implemented in a non-.NET language require wrappers to enable usage from .NET. These wrappers are historically called Runtime Callable Wrapper (**RCW**).
+**Native Object Wrapper** &ndash; COM objects that are implemented in a non-.NET language require wrappers to enable usage from .NET. These wrappers are historically called Runtime Callable Wrappers (**RCW**).
 
-### Step 1 &ndash; Define methods to implement and understand their intent.
+### Step 1 &ndash; Define methods to implement and understand their intent
 
 The following 3 methods are required to be implemented in order to extend the `ComWrappers` type. Each of these methods represent the user's participation in the creation or deletion of a type of wrapper. The `ComputeVtables()` and `CreateObject()` are for the creation of a Managed Object Wrapper and Native Object Wrapper respectively, and the `ReleaseObjects()` is used by the runtime to make a request for the supplied collection of wrappers be "released" from the underlying native object. The `ReleaseObjects()` can typically be left as throwing `NotImplementedException` since it is only called in an advanced scenario involving the [Reference Tracker framework][api_referencetracker].
 
@@ -85,7 +85,7 @@ class DemoComWrappers : ComWrappers
 }
 ```
 
-Implementing the `ComputeVtables()` method requires deciding what managed types one would like to support. For this tutorial we will supporting the two aforementioned interfaces (that is, `IDemoGetType` and `IDemoStoreType`) and a managed type implementing the two interfaces (that is, `DemoImpl`).
+Implementing the `ComputeVtables()` method requires deciding what managed types one would like to support. For this tutorial we will be supporting the two aforementioned interfaces (that is, `IDemoGetType` and `IDemoStoreType`) and a managed type implementing the two interfaces (that is, `DemoImpl`).
 
 ```csharp
 class DemoImpl : IDemoGetType, IDemoStoreType
@@ -96,11 +96,11 @@ class DemoImpl : IDemoGetType, IDemoStoreType
 }
 ```
 
-For the `CreateObject()` we will also need to determine what we would like to support. In this case though, we only know the COM interfaces we are interested in not the COM classes. The interfaces being consumed from the COM side are the same as the ones we are projecting from the .NET side (that is, `IDemoGetType` and `IDemoStoreType`).
+For the `CreateObject()` method we will also need to determine what we would like to support. In this case though, we only know the COM interfaces we are interested in not the COM classes. The interfaces being consumed from the COM side are the same as the ones we are projecting from the .NET side (that is, `IDemoGetType` and `IDemoStoreType`).
 
-In this demo we will not be implementing `ReleaseObjects()`.
+In this tutorial we will not be implementing `ReleaseObjects()`.
 
-### Step 2 &ndash; Implementing `ComputeVtables()`.
+### Step 2 &ndash; Implementing `ComputeVtables()`
 
 Let's start with the Managed Object Wrapper first &ndash; they are actually much easier. We will be building a [Virtual Method Table][wiki_vtable], or a "vtable", for each interface in order to project them into the COM environment. For this tutorial, we are going to define a vtable as a sequence of pointers, where each pointer represents an implementation of a function on an interface &ndash; order is _very_ important here. In COM, every interface inherits from `IUnknown`. The `IUnknown` type has 3 methods defined in the following order: `QueryInterface()`, `AddRef()`, and `Release()`. After the `IUnknown` methods come the specific interface methods. For example, consider `IDemoGetType` and `IDemoStoreType`. Conceptually, the vtables for the types would look like the following:
 
@@ -113,13 +113,19 @@ Release         | Release
 GetString       | StoreString
 ```
 
-Looking at `DemoImpl`, we already have an implementation for `GetString()` and `StoreString()`, but what about the `IUnknown` functions? How to [implement an `IUnknown` instance][doc_impliunknown] is beyond the scope of this tutorial but it can be done manually in `ComWrappers`, but in this case we are going to let the runtime handle that part. We can get the `IUnknown` implementation using the [`ComWrappers.GetIUnknownImpl()`][api_comwrappers] method.
+Looking at `DemoImpl`, we already have an implementation for `GetString()` and `StoreString()`, but what about the `IUnknown` functions? How to [implement an `IUnknown` instance][doc_impliunknown] is beyond the scope of this tutorial but it can be done manually in `ComWrappers`. However, in this tutorial we are going to let the runtime handle that part. We can get the `IUnknown` implementation using the [`ComWrappers.GetIUnknownImpl()`][api_comwrappers] method.
 
-It would appear we have the implementation for all our methods but unfortunately only the `IUnknown` functions are consumable in a COM vtable. Since COM is outside of the runtime, we will need create native function pointers to our `DemoImpl` implementation. This can be done easily using C# function pointers and the [`UnmanagedCallersOnlyAttribute`][api_unmanagedcallersonly]. We can create a function to insert into the vtable by creating a `static` function that mimics the needed COM function signature. Below is an example of the COM signature for `IDemoGetType.GetString()` &ndash; recall from the COM ABI that the first argument is the instance itself.
+It would appear we have the implementation for all our methods but unfortunately only the `IUnknown` functions are consumable in a COM vtable. Since COM is outside of the runtime, we will need to create native function pointers to our `DemoImpl` implementation. This can be done easily using C# function pointers and the [`UnmanagedCallersOnlyAttribute`][api_unmanagedcallersonly]. We can create a function to insert into the vtable by creating a `static` function that mimics the COM function signature. Below is an example of the COM signature for `IDemoGetType.GetString()` &ndash; recall from the COM ABI that the first argument is the instance itself.
 
 ```csharp
 [UnmanagedCallersOnly]
 public static int GetString(IntPtr _this, IntPtr* str);
+```
+
+The wrapper implementation of `IDemoGetType.GetString()` should consist of marshalling logic and then a dispatch to the managed object being wrapped. All the state for dispatch is contained within the provided `_this` argument. The `_this` argument will actually be of type `ComInterfaceDispatch*`. This type represents a low-level structure with a single member that will be discussed shortly. Further details of this type and its layout are an implementation detail of the runtime and should not be depended upon. In order to retrieve the managed instance from a `ComInterfaceDispatch*` instance, one should use the following:
+
+```csharp
+IDemoGetType inst = ComInterfaceDispatch.GetInstance<IDemoGetType>((ComInterfaceDispatch*)_this);
 ```
 
 Now that we have a C# method that can be inserted into a vtable we can construct it. Note the use of [`RuntimeHelpers.AllocateTypeAssociatedMemory()`][api_allocatetypeassociatedmemory] for allocating memory in a way that works with [unloadable][doc_unloadability] assemblies.
@@ -146,7 +152,7 @@ IDemoGetTypeManagedWrapper.GetString;
 }
 ```
 
-The allocation of vtables is the first part of implementing `ComputeVtables()`. We should also construct comprehensive definitions for types that we are planning to support &ndash; think `DemoImpl`. Using the constructed vtables we can now create a series of `ComInterfaceEntry` instances that represent the complete view of the managed object into COM.
+The allocation of vtables is the first part of implementing `ComputeVtables()`. We should also construct comprehensive definitions for types that we are planning to support &ndash; think `DemoImpl`. Using the constructed vtables we can now create a series of `ComInterfaceEntry` instances that represent the complete view of the managed object in COM.
 
 ```csharp
 {
@@ -164,7 +170,7 @@ The allocation of vtables is the first part of implementing `ComputeVtables()`. 
 }
 ```
 
-The allocation of vtables and entries for the Managed Object Wrapper can and should be done ahead of time as this data is for all instances of the type in question. This work could be done in a `static` constructor or in a module initializer, but it should be ahead so the `ComputeVtables()` implementation is as simple and quick as possible.
+The allocation of vtables and entries for the Managed Object Wrapper can and should be done ahead of time since the data can be used for all instances of the type. The work here could be performed in a `static` constructor or a module initializer, but it should be done ahead of time so the `ComputeVtables()` method is as simple and quick as possible.
 
 ```csharp
 protected override unsafe ComInterfaceEntry* ComputeVtables(object obj, CreateComInterfaceFlags flags, 
@@ -182,7 +188,7 @@ out int count)
 }
 ```
 
-Once the `ComputeVtables()` method is implemented the `ComWrappers` subclass will be able to produce Managed Object Wrappers for instances of `DemoImpl`. Be aware that the returned Managed Object Wrapper is of type `IUnknown*`. If the native API that is being passed the wrapper requires a different support interface a [`Marshal.QueryInterface()`][api_marshalqueryinterface] for that interface must be performed.
+Once the `ComputeVtables()` method is implemented the `ComWrappers` subclass will be able to produce Managed Object Wrappers for instances of `DemoImpl`. Be aware that the returned Managed Object Wrapper is of type `IUnknown*`. If the native API that is being passed the wrapper requires a different interface, a [`Marshal.QueryInterface()`][api_marshalqueryinterface] for that interface must be performed.
 
 ```csharp
 var cw = new DemoComWrappers();
@@ -190,7 +196,7 @@ var demo = new DemoImpl();
 IntPtr ccw = cw.GetOrCreateComInterfaceForObject(demo, CreateComInterfaceFlags.None);
 ```
 
-### Step 3 &ndash; Implementing `CreateObject()`.
+### Step 3 &ndash; Implementing `CreateObject()`
 
 Construction of Native Object Wrappers has more implementation options and a great deal more nuance. The first question to address is permissive the `ComWrappers` subclass will be in support COM types. If the idea is to support all COM types, which is possible, a substantial amount of code will need to be written and/or some very clever uses of `Reflection.Emit` will need to be employed. For this tutorial we will only be support COM instances that implement both `IDemoGetType` and `IDemoStoreType`. Since we know there is a finite set and have restricted that any supplied COM instance must implement both interfaces we could provide a single statically defined wrapper; however, dynamic cases are common enough in COM we will explore both options.
 
@@ -305,7 +311,7 @@ IntPtr iunk = ...; // Get a COM instance from native code.
 object rcw = cw.GetOrCreateObjectForComInstance(ccw, CreateObjectFlags.UniqueInstance);
 ```
 
-### Step 4 &ndash; Native Object Wrapper lifetime details.
+### Step 4 &ndash; Native Object Wrapper lifetime details
 
 The `ComputeVtables()` and `CreateObject()` implementations discussed some wrapper lifetime details but further details need discussion. This can be a short step but can also significantly increase the complexity of the `ComWrappers` design.
 
--- a/docs/standard/native-interop/tutorial-comwrappers.md
+++ b/docs/standard/native-interop/tutorial-comwrappers.md
@@ -2,7 +2,7 @@
 title: "Using the ComWrappers API"
 description: A tutorial on properly implementing the ComWrappers API.
 author: AaronRobinsonMSFT
-ms.date: "02/09/2021"
+ms.date: "09/02/2021"
 helpviewer_keywords:
   - "COM interop, ComWrappers"
   - "RCW"
@@ -10,13 +10,17 @@ helpviewer_keywords:
   - "interoperation with unmanaged code, COM wrappers"
   - "NativeAOT"
 ---
-# Using the `ComWrappers` API
+# Tutorial: Use the `ComWrappers` API
 
-In this tutorial we are going to see how to properly subclass the [`ComWrappers`][api_comwrappers] type to provide an optimized and AOT friendly COM interop solution. It is assumed the reader of this document is familiar COM, its architecture, and existing COM interop solutions.
+In this tutorial, you'll learn how to properly subclass the [`ComWrappers`][api_comwrappers] type to provide an optimized and AOT-friendly COM interop solution. Before starting this tutorial, you should be familiar with COM, its architecture, and existing COM interop solutions.
 
-We will be implementing the following interface definitions. These interfaces and their implementation will allow us to see an example of marshalling and unmarshalling types across the COM/.NET boundary, two distinct approaches to consuming native COM objects in .NET, and a recommended pattern for enabling custom COM interop in .NET 5 and beyond.
+In this tutorial, you'll implement the following interface definitions. These interfaces and their implementations will demonstrate:
 
-All source code used in this tutorial is available in the final section of this document.
+- Marshalling and unmarshalling types across the COM/.NET boundary.
+- Two distinct approaches to consuming native COM objects in .NET.
+- A recommended pattern for enabling custom COM interop in .NET 5 and beyond.
+
+All source code used in this tutorial is available in the [final section of this article](#sample-code).
 
 **C# definitions**
 
@@ -34,7 +38,7 @@ interface IDemoStoreType
 
 **Win32 C++ definitions**
 
-```c++
+```cpp
 MIDL_INTERFACE("92BAA992-DB5A-4ADD-977B-B22838EE91FD")
 IDemoGetType : public IUnknown
 {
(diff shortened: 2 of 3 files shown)


docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+
+Aside from the lifetime, type system, and functional features that are discussed above, a COM compliant implementation of `ComWrappers` requires additional considerations. The following should also be considered for any implementation that will be used on the Windows platform.
+
+[**Apartments**][doc_comapartments] &ndash; COM's organizational structure for threading is called "Apartments" and has strict rules that must be followed for stable operations. This tutorial does not implement apartment aware Native Object Wrappers but for any production ready implementation it must. In order to accomplish this we recommend using the [`RoGetAgileReference`][api_rogetagilereference] API introduced in Windows 8. Prior to Windows 8 needs should consider the [Global Interface Table][doc_globalinterfacetable].
```
gewarren, line 390:
```suggestion
- [**Apartments**][doc_comapartments] &ndash; COM's organizational structure for threading is called "Apartments" and has strict rules that must be followed for stable operations. This tutorial does not implement apartment-aware Native Object Wrappers, but any production ready implementation should be apartment-aware. To accomplish this, we recommend using the [`RoGetAgileReference`][api_rogetagilereference] API introduced in Windows 8. For versions prior to Windows 8, consider the [Global Interface Table][doc_globalinterfacetable].
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+
+[**Security**][doc_comsecurity] &ndash; COM provides a rich security model for class activation and proxied permission.
```
gewarren, line 392:
```suggestion
- [**Security**][doc_comsecurity] &ndash; COM provides a rich security model for class activation and proxied permission.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+[api_marshalqueryinterface]:/dotnet/api/system.runtime.interopservices.marshal.queryinterface
+[api_referencetracker]:/windows/win32/api/windows.ui.xaml.hosting.referencetracker/
+[api_rogetagilereference]:/windows/win32/api/combaseapi/nf-combaseapi-rogetagilereference
+[api_unmanagedcallersonly]:/dotnet/api/system.runtime.interopservices.unmanagedcallersonlyattribute
+
+[doc_comapartments]:/windows/win32/com/processes--threads--and-apartments
+[doc_comsecurity]:/windows/win32/com/security-in-com
+[doc_globalinterfacetable]:/windows/win32/com/when-to-use-the-global-interface-table
+[doc_impliunknown]:/windows/win32/com/using-and-implementing-iunknown
+[doc_nullable]:/dotnet/csharp/nullable-references
```
gewarren, line 409:
```suggestion
[doc_nullable]:../../csharp/nullable-references.md
```

These relative links are verified by the build.

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+[doc_unloadability]:/dotnet/standard/assembly/unloadability
```
gewarren, line 410:
```suggestion
[doc_unloadability]:../assembly/unloadability.md
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+[doc_unsafekeyword]:/dotnet/csharp/language-reference/keywords/unsafe
```
gewarren, line 411:
```suggestion
[doc_unsafekeyword]:../../csharp/language-reference/keywords/unsafe.md
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+
+1) The Native Object Wrapper's Finalizer should be responsible for calling the COM instance's `Release()`. This is the only time when it is safe to call this method since at this point it has been correctly determined by the GC there are no other references to the Native Object Wrapper in the .NET runtime. There can be complexity here if one is properly supporting COM Apartments &ndash; see the "Additional considerations" section.
+
+2) Have the returned Native Object Wrapper implement `IDisposable` and call `Release()` in `Dispose()`. This pattern should only be supported if during the `CreateObject()` call the `CreateObjectFlags.UniqueInstance` flag was passed in. **Note** if this requirement is not followed it is possible for disposed Native Object Wrappers to be reused after being disposed.
+
+## Using the `ComWrappers` subclass
+
+We now have a `ComWrappers` subclass that can be tested. In order to avoid creating a native library that returns a COM instance that implements `IDemoGetType` and `IDemoStoreType` we are going to use the Managed Object Wrapper and treat it as a COM instance &ndash; this must be possible in order to pass it COM anyways.
```
gewarren, line 326:
```suggestion
You now have a `ComWrappers` subclass that can be tested. To avoid creating a native library that returns a COM instance that implements `IDemoGetType` and `IDemoStoreType`, you'll use the Managed Object Wrapper and treat it as a COM instance &ndash; this must be possible in order to pass it COM anyways.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+
+Let's create a Managed Object Wrapper first. We will instantiate a `DemoImpl` instance display its current string state.
```
gewarren, line 328:
```suggestion
Let's create a Managed Object Wrapper first. Instantiate a `DemoImpl` instance and display its current string state.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+
+```csharp
+var demo = new DemoImpl();
+
+string? value = demo.GetString();
+Console.WriteLine($"Initial string: {value ?? "<null>"}");
+```
+
+Now we can create an instance of `DemoComWrappers` and create a Managed Object Wrapper that we could then pass into a COM environment.
```
gewarren, line 337:
```suggestion
Now you can create an instance of `DemoComWrappers` and a Managed Object Wrapper that you can then pass into a COM environment.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+
+```csharp
+var cw = new DemoComWrappers();
+
+IntPtr ccw = cw.GetOrCreateComInterfaceForObject(demo, CreateComInterfaceFlags.None);
+```
+
+Instead of passing the Managed Object Wrapper to a COM environment let's pretend we just received this COM instance and instead create a Native Object Wrapper for it.
```
gewarren, line 345:
```suggestion
Instead of passing the Managed Object Wrapper to a COM environment, pretend you just received this COM instance, so you'll create a Native Object Wrapper for it instead.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+
+msg = msg.ToUpper();
+demo.StoreString(msg.Length, msg.ToUpper());
+Console.WriteLine($"Setting string through managed object: {msg}");
+
+value = getter.GetString();
+Console.WriteLine($"Get string through wrapper: {value}");
+```
+
+Since our `ComWrapper` subclass was designed to support `CreateObjectFlags.UniqueInstance` we can clean up the Native Object Wrapper immediately instead of waiting for a GC to occur.
```
gewarren, line 372:
```suggestion
Since your `ComWrapper` subclass was designed to support `CreateObjectFlags.UniqueInstance`, you can clean up the Native Object Wrapper immediately instead of waiting for a GC to occur.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+IntPtr ccw = cw.GetOrCreateComInterfaceForObject(demo, CreateComInterfaceFlags.None);
+```
+
+Instead of passing the Managed Object Wrapper to a COM environment let's pretend we just received this COM instance and instead create a Native Object Wrapper for it.
+
+```csharp
+var rcw = cw.GetOrCreateObjectForComInstance(ccw, CreateObjectFlags.UniqueInstance);
+```
+
+With the Native Object Wrapper we should be able to cast it to one of the desired interfaces and use it as a normal managed object. We can examine the `DemoImpl` instance and observe the impact of operations on the Native Object Wrapper that is wrapping a Managed Object Wrapper that is in turn wrapping the managed instance is reflected.
```
gewarren, line 351:
```suggestion
With the Native Object Wrapper, you should be able to cast it to one of the desired interfaces and use it as a normal managed object. You can examine the `DemoImpl` instance and observe the impact of operations on the Native Object Wrapper that's wrapping a Managed Object Wrapper that's in turn wrapping the managed instance.
```

I wasn't sure what the last "is reflected" words were trying to convey, so they may need to be added back.

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+Once the `CreateObject()` method is implemented the `ComWrappers` subclass will be able to produce Native Object Wrappers for COM instances that implement both `IDemoGetType` and `IDemoStoreType`.
+
+```csharp
+IntPtr iunk = ...; // Get a COM instance from native code.
+object rcw = cw.GetOrCreateObjectForComInstance(ccw, CreateObjectFlags.UniqueInstance);
+```
+
+### Step 4 &ndash; Native Object Wrapper lifetime details
+
+The `ComputeVtables()` and `CreateObject()` implementations discussed some wrapper lifetime details but further details need discussion. This can be a short step but can also significantly increase the complexity of the `ComWrappers` design.
```
gewarren, line 316:
```suggestion
The `ComputeVtables()` and `CreateObject()` implementations covered some wrapper lifetime details, but there are further considerations. While this can be a short step, it can also significantly increase the complexity of the `ComWrappers` design.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+
+Unlike the Managed Object Wrapper, which is controlled by calls to its `AddRef()`/`Release()`, the lifetime of a Native Object Wrapper is non-deterministically handled by the GC. The question here is, when does the Native Object Wrapper call `Release()` on the `IntPtr` that represents the COM instance? There are two general buckets here.
```
gewarren, line 318:
```suggestion
Unlike the Managed Object Wrapper, which is controlled by calls to its `AddRef()` and `Release()` methods, the lifetime of a Native Object Wrapper is nondeterministically handled by the GC. The question here is, when does the Native Object Wrapper call `Release()` on the `IntPtr` that represents the COM instance? There are two general buckets:
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+
+1) The Native Object Wrapper's Finalizer should be responsible for calling the COM instance's `Release()`. This is the only time when it is safe to call this method since at this point it has been correctly determined by the GC there are no other references to the Native Object Wrapper in the .NET runtime. There can be complexity here if one is properly supporting COM Apartments &ndash; see the "Additional considerations" section.
```
gewarren, line 320:
```suggestion
1) The Native Object Wrapper's Finalizer is responsible for calling the COM instance's `Release()` method. This is the only time when it's safe to call this method. At this point, it's been correctly determined by the GC that there are no other references to the Native Object Wrapper in the .NET runtime. There can be complexity here if you're properly supporting COM Apartments; for more information, see the [Additional considerations](#additional-considerations) section.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+
+2) Have the returned Native Object Wrapper implement `IDisposable` and call `Release()` in `Dispose()`. This pattern should only be supported if during the `CreateObject()` call the `CreateObjectFlags.UniqueInstance` flag was passed in. **Note** if this requirement is not followed it is possible for disposed Native Object Wrappers to be reused after being disposed.
```
gewarren, line 322:
```suggestion
2) The Native Object Wrapper implements `IDisposable` and calls `Release()` in `Dispose()`.

  > [!NOTE]
  > This pattern should only be supported if, during the `CreateObject()` call, the `CreateObjectFlags.UniqueInstance` flag was passed in. If this requirement is not followed, it's possible for disposed Native Object Wrappers to be reused after being disposed.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+return new DemoNativeStaticWrapper()
+{
+    IDemoGetTypeInst = IDemoGetTypeInst,
+    IDemoStoreTypeInst = IDemoStoreTypeInst
+};
+```
+
+#### Dynamic Native Object Wrapper
+
+Dynamic wrappers are more flexible in that they provide a way for types to be query at run time instead of statically. In order to provide this support we will utilize [`IDynamicInterfaceCastable`][api_idynamicinterfacecastable]. Observe that `DemoNativeDynamicWrapper` only implements this interface. The functionality that the interface provides is a chance at run time to determine what type is supported.
```
gewarren, line 245:
```suggestion
Dynamic wrappers are more flexible becaues they provide a way for types to be queried at run time instead of statically. In order to provide this support, you'll utilize [`IDynamicInterfaceCastable`][api_idynamicinterfacecastable]. Observe that `DemoNativeDynamicWrapper` only implements this interface. The functionality that the interface provides is a chance to determine what type is supported at run time.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+{
+    public RuntimeTypeHandle GetInterfaceImplementation(RuntimeTypeHandle interfaceType) =>
+        throw new NotImplementedException();
+
+    public bool IsInterfaceImplemented(RuntimeTypeHandle interfaceType, bool throwIfNotImplemented) =>
+        throw new NotImplementedException();
+}
+```
+
+Let's look at one of the types that `DemoNativeDynamicWrapper` will dynamically support. The `IDataStoreType` is defined below using the default interface methods feature.
```
gewarren, line 259:
```suggestion
Let's look at one of the types that `DemoNativeDynamicWrapper` will dynamically support. The following code defines the `IDataStoreType` type using the *default interface methods* feature.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+
+    void IDemoStoreType.StoreString(int len, string? str)
+    {
+        var inst = ((DemoNativeDynamicWrapper)this).IDemoStoreTypeInst;
+        StoreString(inst, len, str);
+    }
+}
+```
+
+There are two important things to note in this example. The first is `DynamicInterfaceCastableImplementationAttribute`. This attribute is required on any type that is returned from a `IDynamicInterfaceCastable` method. It has the added benefit of making IL Linking easier which means NativeAOT scenarios are more reliable. The second is the cast to `DemoNativeDynamicWrapper`. This is part of the dynamic nature of `IDynamicInterfaceCastable`. The type that is returned from `IDynamicInterfaceCastable.GetInterfaceImplementation()` is used to blanket the type that implements `IDynamicInterfaceCastable`.
```
gewarren, line 275:
```suggestion
There are two important things to note in this example:

1) The `DynamicInterfaceCastableImplementationAttribute` attribute. This attribute is required on any type that is returned from a `IDynamicInterfaceCastable` method. It has the added benefit of making IL Linking easier, which means NativeAOT scenarios are more reliable.
2) The cast to `DemoNativeDynamicWrapper`. This is part of the dynamic nature of `IDynamicInterfaceCastable`. The type that's returned from `IDynamicInterfaceCastable.GetInterfaceImplementation()` is used to blanket the type that implements `IDynamicInterfaceCastable`.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+
+Regardless of which Native Object Wrapper is used, the ability to invoke functions on a COM instance is needed. The implementation of `IDemoStoreTypeNativeWrapper.StoreString()` can serve as an example employing `unmanaged` C# function pointers.
```
gewarren, line 277:
```suggestion
Regardless of which Native Object Wrapper is used, you need the ability to invoke functions on a COM instance. The implementation of `IDemoStoreTypeNativeWrapper.StoreString()` can serve as an example of employing `unmanaged` C# function pointers.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+    int hr = ((delegate* unmanaged<IntPtr, int, IntPtr, int>)(*(*(void***)inst + 3 /* IDemoStoreType.StoreString slot */)))(inst, len, strLocal);
+    if (hr != 0)
+    {
+        Marshal.FreeCoTaskMem(strLocal);
+        Marshal.ThrowExceptionForHR(hr);
+    }
+}
+```
+
+The dereferencing of the COM instance to access its vtable implementation warrants examining. The COM ABI defines that the first pointer of an object is to the type's vtable and from there the desired slot can be accessed. Let's assume the address of the COM object is `0x10000`. The first pointer sized value should be the address of the vtable &ndash; in this example `0x20000`. Once we are at the vtable we look for the 4th slot (3 in zero-based indexing) to access the `StoreString()` implementation.
```
gewarren, line 292:
```suggestion
Let's examine the dereferencing of the COM instance to access its vtable implementation. The COM ABI defines that the first pointer of an object is to the type's vtable and, from there, the desired slot can be accessed. Let's assume the address of the COM object is `0x10000`. The first pointer-sized value should be the address of the vtable &ndash; in this example `0x20000`. Once you're at the vtable, you look for the fourth slot (index 3 in zero-based indexing) to access the `StoreString()` implementation.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+VTable for IDemoStoreType
+0x20000  <Address of QueryInterface>
+0x20008  <Address of AddRef>
+0x20010  <Address of Release>
+0x20018  <Address of StoreString>
+```
+
+Having the function pointer then allows us to dispatch to that member function on that object by passing the object instance as the fist parameter &ndash; that should look familiar based on the function definitions when implementing the Managed Object Wrapper.
+
+Once the `CreateObject()` method is implemented the `ComWrappers` subclass will be able to produce Native Object Wrappers for COM instances that implement both `IDemoGetType` and `IDemoStoreType`.
```
gewarren, line 307:
```suggestion
Once the `CreateObject()` method is implemented, the `ComWrappers` subclass will be able to produce Native Object Wrappers for COM instances that implement both `IDemoGetType` and `IDemoStoreType`.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
...
+0x10000  0x20000
+
+VTable for IDemoStoreType
+0x20000  <Address of QueryInterface>
+0x20008  <Address of AddRef>
+0x20010  <Address of Release>
+0x20018  <Address of StoreString>
+```
+
+Having the function pointer then allows us to dispatch to that member function on that object by passing the object instance as the fist parameter &ndash; that should look familiar based on the function definitions when implementing the Managed Object Wrapper.
```
gewarren, line 305:
```suggestion
Having the function pointer then allows you to dispatch to that member function on that object by passing the object instance as the first parameter. This pattern should look familiar based on the function definitions of the Managed Object Wrapper implementation.
```

gewarren (commented):
This is a long one. Thanks for adding it!

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,936 @@
+---
+title: "Using the ComWrappers API"
+description: A tutorial on properly implementing the ComWrappers API.
+author: AaronRobinsonMSFT
+ms.date: "02/09/2021"
```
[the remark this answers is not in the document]
AaronRobinsonMSFT:
```suggestion
ms.date: "09/02/2021"
```

gewarren:
Here's the [preview link](https://review.docs.microsoft.com/en-us/dotnet/standard/native-interop/tutorial-comwrappers?branch=pr-en-us-25933). Also, consider adding the sample to code to the [samples browser](https://docs.microsoft.com/en-us/samples/browse/) by adding to the https://github.com/dotnet/samples repo. Then you could just link there instead of including it as the last section of the file.

docs/standard/native-interop/tutorial-comwrappers.md
AaronRobinsonMSFT: You captured the idea correctly :)

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,946 @@
...
+
+Since your `ComWrapper` subclass was designed to support `CreateObjectFlags.UniqueInstance`, you can clean up the Native Object Wrapper immediately instead of waiting for a GC to occur.
+
+```csharp
+(rcw as IDisposable)?.Dispose();
+```
+
+## Additional considerations
+
+**Native AOT** &ndash; Ahead of Time (AOT) compilation provides improved startup cost as JIT compilation is avoided. Removing the need for JIT compilation is also often required on some platforms. Supporting AOT was a goal of the `ComWrappers` API, but any wrapper implementation must be careful not to inadvertantly introduce cases where AOT breaks down, such as using reflection. The `Type.GUID` property is an example of where reflection is used, but in a non-obvious way. The `Type.GUID` property uses reflection to inspect the type's attributes and then potentially the type's name and containing assembly in order to generate its value.
```
elinor-fung, line 390:
```suggestion
**Native AOT** &ndash; Ahead-of-time (AOT) compilation provides improved startup cost as JIT compilation is avoided. Removing the need for JIT compilation is also often required on some platforms. Supporting AOT was a goal of the `ComWrappers` API, but any wrapper implementation must be careful not to inadvertantly introduce cases where AOT breaks down, such as using reflection. The `Type.GUID` property is an example of where reflection is used, but in a non-obvious way. The `Type.GUID` property uses reflection to inspect the type's attributes and then potentially the type's name and containing assembly in order to generate its value.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,946 @@
...
+0x20018  <Address of StoreString>
+```
+
+Having the function pointer then allows you to dispatch to that member function on that object by passing the object instance as the first parameter. This pattern should look familiar based on the function definitions of the Managed Object Wrapper implementation.
+
+Once the `CreateObject()` method is implemented, the `ComWrappers` subclass will be able to produce Native Object Wrappers for COM instances that implement both `IDemoGetType` and `IDemoStoreType`.
+
+```csharp
+IntPtr iunk = ...; // Get a COM instance from native code.
+object rcw = cw.GetOrCreateObjectForComInstance(ccw, CreateObjectFlags.UniqueInstance);
```
sbomer, line 318:
```suggestion
object rcw = cw.GetOrCreateObjectForComInstance(iunk, CreateObjectFlags.UniqueInstance);
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,946 @@
...
+{
+    public RuntimeTypeHandle GetInterfaceImplementation(RuntimeTypeHandle interfaceType) =>
+        throw new NotImplementedException();
+
+    public bool IsInterfaceImplemented(RuntimeTypeHandle interfaceType, bool throwIfNotImplemented) =>
+        throw new NotImplementedException();
+}
+```
+
+Let's look at one of the types that `DemoNativeDynamicWrapper` will dynamically support. The following code defines the `IDataStoreType` type using the *default interface methods* feature.
```
sbomer, line 263:
```suggestion
Let's look at one of the interfaces that `DemoNativeDynamicWrapper` will dynamically support. The following code provides the implementation of `IDemoStoreType` using the *default interface methods* feature.
```

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,946 @@
...
+    {
+        var inst = ((DemoNativeDynamicWrapper)this).IDemoStoreTypeInst;
+        StoreString(inst, len, str);
+    }
+}
+```
+
+There are two important things to note in this example:
+
+1) The `DynamicInterfaceCastableImplementationAttribute` attribute. This attribute is required on any type that is returned from a `IDynamicInterfaceCastable` method. It has the added benefit of making IL Linking easier, which means NativeAOT scenarios are more reliable.
```
sbomer, line 281:
```suggestion
1) The `DynamicInterfaceCastableImplementationAttribute` attribute. This attribute is required on any type that is returned from a `IDynamicInterfaceCastable` method. It has the added benefit of making IL trimming easier, which means NativeAOT scenarios are more reliable.
```

docs/fundamentals/toc.yml
```diff
@@ -2474,6 +2474,10 @@ items:
           href: ../standard/native-interop/com-callable-wrapper.md
```
gewarren, line 2474:
```suggestion
          href: ../standard/native-interop/com-callable-wrapper.md
        - name: Tutorial: Use the ComWrappers API
          href: ../standard/native-interop/tutorial-comwrappers.md
```

docs/fundamentals/toc.yml
```diff
@@ -2474,6 +2474,10 @@ items:
...
       - name: Qualifying .NET types for COM interop
         href: ../standard/native-interop/qualify-net-types-for-interoperation.md
+      - name: Tutorials
```
gewarren, line 2477:
```suggestion
```

docs/fundamentals/toc.yml
```diff
@@ -2474,6 +2474,10 @@ items:
...
+        items:
```
gewarren, line 2478:
```suggestion
```

docs/fundamentals/toc.yml
```diff
@@ -2474,6 +2474,10 @@ items:
...
+        - name: Using the ComWrappers API
```
gewarren, line 2479:
```suggestion
```

docs/fundamentals/toc.yml
```diff
@@ -2474,6 +2474,10 @@ items:
...
+          href: ../standard/native-interop/tutorial-comwrappers.md
```
gewarren, line 2480:
```suggestion
```

gewarren (commented):
TOC suggestion

gewarren (approved):
LGTM

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+return new DemoNativeStaticWrapper()
+{
+    IDemoGetTypeInst = IDemoGetTypeInst,
+    IDemoStoreTypeInst = IDemoStoreTypeInst
+};
+```
+
+#### Dynamic Native Object Wrapper
+
+Dynamic wrappers are more flexible because they provide a way for types to be queried at run time instead of statically. In order to provide this support, you'll utilize [`IDynamicInterfaceCastable`][api_idynamicinterfacecastable]. Observe that `DemoNativeDynamicWrapper` only implements this interface. The functionality that the interface provides is a chance to determine what type is supported at run time.
```
vitek-karas, line 249: > types to be queried at run time instead of statically.

I think this is confusing. Even the static one queries the types at runtime. If I understand this correctly the core difference is that with the static one all of the COM interfaces and their managed counterparts must be figured out up front (for each instance it's needed on), the dynamic one is lazy, it only needs to figure out the managed interface when something actually asks for it.

Curiosity: How does the `IDynamicallyInterfaceCastable` play with reflection - for example what happens if I call `obj.GetType().GetInterfaces()` - will it include the dynamic interfaces or not?

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+        var inst = ((DemoNativeDynamicWrapper)this).IDemoStoreTypeInst;
+        StoreString(inst, len, str);
+    }
+}
+```
+
+There are two important things to note in this example:
+
+1) The `DynamicInterfaceCastableImplementationAttribute` attribute. This attribute is required on any type that is returned from a `IDynamicInterfaceCastable` method. It has the added benefit of making IL trimming easier, which means NativeAOT scenarios are more reliable.
+2) The cast to `DemoNativeDynamicWrapper`. This is part of the dynamic nature of `IDynamicInterfaceCastable`. The type that's returned from `IDynamicInterfaceCastable.GetInterfaceImplementation()` is used to blanket the type that implements `IDynamicInterfaceCastable`.
```
vitek-karas, line 282: > The type that's returned from `IDynamicInterfaceCastable.GetInterfaceImplementation()` is used to blanket the type that implements `IDynamicInterfaceCastable`.

It could be my subpar english, but I don't understand what this is trying to say. I know this is about the dynamic castable feature, which is not really com wrappers, so maybe this should link to another doc which explains how this works (I'm curious how this thing works).

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+
+**Global registration** &ndash; Since the `ComWrappers` API was designed as a new phase of COM interop, it needed to have some way to partially integrate with the existing system. There are globally impacting static methods on the `ComWrappers` API that permit registration of a global instance for various support. These methods are designed for `ComWrappers` instances that are expecting to provide comprehensive COM interop support in all cases &ndash; akin to the built-in COM interop system.
+
+[**Reference Tracker support**][api_referencetracker] &ndash; This support is primary used for WinRT scenarios and represents an advanced scenario. For the majority of `ComWrapper` implementations, either a `CreateComInterfaceFlags.TrackerSupport` or `CreateObjectFlags.TrackerObject` flag should throw a <xref:System.NotSupportedException>. If you'd like to enable this support, perhaps on a Windows or even non-Windows platform, it is highly recommended to reference the [C#/WinRT tool chain][repo_cswinrt].
```
vitek-karas, line 398: >  or even non-Windows platform

Does this mean that COM Wrappers work on all platforms? I think that should be mentioned somewhere in the doc. Actually it should mention it either way - whether it's Windows specific or not.

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+```
+IDemoGetType    | IDemoStoreType
+==================================
+QueryInterface  | QueryInterface
+AddRef          | AddRef
+Release         | Release
+GetString       | StoreString
+```
+
+Looking at `DemoImpl`, we already have an implementation for `GetString()` and `StoreString()`, but what about the `IUnknown` functions? How to [implement an `IUnknown` instance][doc_impliunknown] is beyond the scope of this tutorial, but it can be done manually in `ComWrappers`. However, in this tutorial, you'll let the runtime handle that part. You can get the `IUnknown` implementation using the [`ComWrappers.GetIUnknownImpl()`][api_comwrappers] method.
```
vitek-karas, line 120: Is it intentional that the link for `GetIUnknownImpl` points to the more generic `ComWrappers` reference, why not the method page itself?

vitek-karas: Another piece of feedback is on the API reference for this method - it says very little. Would be nice to have some description of what the runtime implementation does (and roughly how) - specifically the QI behavior is interesting I think.

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+
+Now that you have a C# method that can be inserted into a vtable, you can construct the vtable. Note the use of [`RuntimeHelpers.AllocateTypeAssociatedMemory()`][api_allocatetypeassociatedmemory] for allocating memory in a way that works with [unloadable][doc_unloadability] assemblies.
+
+```csharp
+GetIUnknownImpl(
+    out IntPtr fpQueryInterface,
+    out IntPtr fpAddRef,
+    out IntPtr fpRelease);
+
+{
```
vitek-karas, line 143: I would recommend somehow visually separating the above call. At first glance it looks like a method definition and body for `GetIUnknownImpl`  (especially since the inline variable declaration looks exactly like parameter definition). The only syntax difference is the semicolon. (I would say this is C#'s fault, but that doesn't solve anything).

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+    int tableCount = 4;
+    int idx = 0;
+    var vtable = (IntPtr*)RuntimeHelpers.AllocateTypeAssociatedMemory(
+        typeof(DemoComWrappers),
+        IntPtr.Size * tableCount);
+    vtable[idx++] = fpQueryInterface;
+    vtable[idx++] = fpAddRef;
+    vtable[idx++] = fpRelease;
+    vtable[idx++] = (IntPtr)(delegate* unmanaged<IntPtr, IntPtr*, int>)&ABI.
```
vitek-karas, line 152: This is more of a C# rant - why is it necessary to type out the "native delegate" type here? The funny thing is that the compiler knows exactly what it should be, since if I write it wrong it will complain. It would be nice if just `(IntPtr)&Method` worked.

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+        typeof(DemoComWrappers),
+        sizeof(ComInterfaceEntry) * s_DemoImplDefinitionLen);
+    entries[idx].IID = IDemoGetType.IID_IDemoGetType;
+    entries[idx++].Vtable = s_IDemoGetTypeVTable;
+    entries[idx].IID = IDemoStoreType.IID_IDemoStoreType;
+    entries[idx++].Vtable = s_IDemoStoreVTable;
+    Debug.Assert(s_DemoImplDefinitionLen == idx);
+    s_DemoImplDefinition = entries;
+}
+```
```
vitek-karas, line 175: Comment on the full code in the samples repo:
```C#
Debug.Assert(flags.HasFlag(CreateComInterfaceFlags.None));
```

The condition is always true since `None == 0`. It should probably be `Debug.Assert(flags is CreateComInterfaceFlags.None)`.

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+
+    public void StoreString(int len, string? str) =>
+        throw new NotImplementedException();
+}
+```
+
+To construct an instance of this class and provide it as a wrapper, you must define some policy. If this type is used as a wrapper, it would seem that since it implements both interfaces, the underlying COM instance should implement both interfaces too. Given that you're adopting this policy, you'll need to confirm this through calls to [`Marshal.QueryInterface()`][api_marshalqueryinterface] on the COM instance.
+
+```csharp
+int hr = Marshal.QueryInterface(ptr, ref IDemoGetType.IID_IDemoGetType, out IntPtr IDemoGetTypeInst);
```
vitek-karas, line 227: I guess a C# rant - the `IID_IDemoGetType` can't be a readonly field - since the "ref" here is "in/out".

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+```csharp
+var cw = new DemoComWrappers();
+
+IntPtr ccw = cw.GetOrCreateComInterfaceForObject(demo, CreateComInterfaceFlags.None);
+```
+
+Instead of passing the Managed Object Wrapper to a COM environment, pretend you just received this COM instance, so you'll create a Native Object Wrapper for it instead.
+
+```csharp
+var rcw = cw.GetOrCreateObjectForComInstance(ccw, CreateObjectFlags.UniqueInstance);
```
vitek-karas, line 360: I think both `GetOrCreateComInterfaceForObject` and `GetOrCreateObjectForComInstance` should document their behavior with regard to `AddRef`/`Release` on the native pointer. For example in this case, I honestly can't tell if the code should release `ccw` after it's used here, or if the reference is "Taken over" by the RCW creation. (COM conventions seem to dictate that the `ccw` should be released in this case, so maybe this is a leak?)

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+string? value = demo.GetString();
+Console.WriteLine($"Initial string: {value ?? "<null>"}");
+```
+
+Now you can create an instance of `DemoComWrappers` and a Managed Object Wrapper that you can then pass into a COM environment.
+
+```csharp
+var cw = new DemoComWrappers();
+
+IntPtr ccw = cw.GetOrCreateComInterfaceForObject(demo, CreateComInterfaceFlags.None);
```
vitek-karas, line 354: I didn't see this sample mention how to handle cases where the code want to `CoCreateInstance` the COM object? I think it might be a relatively common case. Does it need to PInvoke into the Win32 API, or is there some managed helper to make it easier?

vitek-karas (commented):
Sorry I'm late to the party, but I think the PR interface works better then me filing a large issue or something similar.

This is a great guide/sample - with some effort I was able to make the code work (and if I would not make a simple mistake, it would probably work on the first try). Given how complex the topic is, this is really good.

Almost all my feedback is subjective - I mainly wanted to point out places where I got stuck for a bit, it's very possible that lot of them are unavoidable.

I also included bunch of curious questions - feel free to do what you want with them (including completely ignoring them)

vitek-karas:
If anybody's interested here's my result of following the guide to create a very simple COM wrappers for XmlLite library (just super basic XML parsing is supported).

https://github.com/vitek-karas/public-samples/tree/main/XmlLiteComWrappers

docs/standard/native-interop/tutorial-comwrappers.md
AaronRobinsonMSFT: > description of what the runtime implementation does

That is fair. I will add some notes.

> (and roughly how)

The how is an implementation detail I'd rather not document as it is always subject to change. Also, it shouldn't be relevant in the majority of cases.

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+
+```csharp
+[UnmanagedCallersOnly]
+public static int GetString(IntPtr _this, IntPtr* str);
+```
+
+The wrapper implementation of `IDemoGetType.GetString()` should consist of marshalling logic and then a dispatch to the managed object being wrapped. All the state for dispatch is contained within the provided `_this` argument. The `_this` argument will actually be of type `ComInterfaceDispatch*`. This type represents a low-level structure with a single field, `Vtable`, that will be discussed later. Further details of this type and its layout are an implementation detail of the runtime and should not be depended upon. In order to retrieve the managed instance from a `ComInterfaceDispatch*` instance, use the following code:
+
+```csharp
+IDemoGetType inst = ComInterfaceDispatch.GetInstance<IDemoGetType>((ComInterfaceDispatch*)_this);
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: I prefer the cast and being explicit here since just accepting an `IntPtr`, in my opinion, is a bit too much freedom.

docs/standard/native-interop/tutorial-comwrappers.md
AaronRobinsonMSFT: We should take this offline. There is actually a reason and it relates to our calling convention extension model.

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+    vtable[idx++] = fpAddRef;
+    vtable[idx++] = fpRelease;
+    vtable[idx++] = (IntPtr)(delegate* unmanaged<IntPtr, IntPtr*, int>)&ABI.
+IDemoGetTypeManagedWrapper.GetString;
+    Debug.Assert(tableCount == idx);
+    s_IDemoGetTypeVTable = (IntPtr)vtable;
+}
+```
+
+The allocation of vtables is the first part of implementing `ComputeVtables()`. You should also construct comprehensive definitions for types that you're planning to support &ndash; think `DemoImpl`. Using the constructed vtables, you can now create a series of [`ComInterfaceEntry`][api_cominterfaceentry] instances that represent the complete view of the managed object in COM.
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: "COM View" is a good idea.

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+    entries[idx++].Vtable = s_IDemoGetTypeVTable;
+    entries[idx].IID = IDemoStoreType.IID_IDemoStoreType;
+    entries[idx++].Vtable = s_IDemoStoreVTable;
+    Debug.Assert(s_DemoImplDefinitionLen == idx);
+    s_DemoImplDefinition = entries;
+}
+```
+
+The allocation of vtables and entries for the Managed Object Wrapper can and should be done ahead of time since the data can be used for all instances of the type. The work here could be performed in a `static` constructor or a module initializer, but it should be done ahead of time so the `ComputeVtables()` method is as simple and quick as possible.
+
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: That is fair. In this case those I would argue the types the ComWrappers implementation supports should be left alone because you never know what type will appear from COM.

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+        return s_DemoImplDefinition;
+    }
+
+    // Unknown type
+    count = 0;
+    return null;
+}
+```
+
+Once you've implemented the `ComputeVtables()` method, the `ComWrappers` subclass will be able to produce Managed Object Wrappers for instances of `DemoImpl`. Be aware that the returned Managed Object Wrapper from the call to [`GetOrCreateComInterfaceForObject()`](xref:System.Runtime.InteropServices.ComWrappers.GetOrCreateComInterfaceForObject%2A) is of type `IUnknown*`. If the native API that's being passed to the wrapper requires a different interface, a [`Marshal.QueryInterface()`][api_marshalqueryinterface] for that interface must be performed.
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: > Will the runtime construct such a vtable?

Yes, by default. Users can provide their own though.

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+
+```csharp
+var cw = new DemoComWrappers();
+var demo = new DemoImpl();
+IntPtr ccw = cw.GetOrCreateComInterfaceForObject(demo, CreateComInterfaceFlags.None);
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: I like using the `ccw` and `rcw` for the variables as it is familiar for some.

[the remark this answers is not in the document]
AaronRobinsonMSFT: Sure.

docs/standard/native-interop/tutorial-comwrappers.md
AaronRobinsonMSFT: >I think it might be a relatively common case.

Agreed. I am going to leave that out. It is conceptually simple once you try it. I don't think it is worth adding that concept here though.

docs/standard/native-interop/tutorial-comwrappers.md
AaronRobinsonMSFT: Not at all. It is a hard concept and my phrasing isn't great. I tried to explain it in a single sentence and likely didn't do a great job. I will try again.

I think we should have another tutorial on IDIC in general. /cc @elinor-fung

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+int hr = Marshal.QueryInterface(ptr, ref IDemoGetType.IID_IDemoGetType, out IntPtr IDemoGetTypeInst);
+if (hr != 0)
+{
+    return null;
+}
+
+hr = Marshal.QueryInterface(ptr, ref IDemoStoreType.IID_IDemoStoreType, out IntPtr IDemoStoreTypeInst);
+if (hr != 0)
+{
+    Marshal.Release(ptr);
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: You are absolutely right. This is a failure on my part.

docs/standard/native-interop/tutorial-comwrappers.md
AaronRobinsonMSFT: I know, I know...

docs/standard/native-interop/tutorial-comwrappers.md
elinor-fung: Yeah, I'd agree that `IDynamicInterfaceCastable` is complicated and we should have a dedicated tutorial. (All we have is a [sample](https://github.com/dotnet/samples/tree/main/core/interop/IDynamicInterfaceCastable) and [part of a blog post](https://devblogs.microsoft.com/dotnet/improvements-in-native-code-interop-in-net-5-0/#idynamicinterfacecastable) from 5.0)

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+    entries[idx++].Vtable = s_IDemoStoreVTable;
+    Debug.Assert(s_DemoImplDefinitionLen == idx);
+    s_DemoImplDefinition = entries;
+}
+```
+
+The allocation of vtables and entries for the Managed Object Wrapper can and should be done ahead of time since the data can be used for all instances of the type. The work here could be performed in a `static` constructor or a module initializer, but it should be done ahead of time so the `ComputeVtables()` method is as simple and quick as possible.
+
+```csharp
+protected override unsafe ComInterfaceEntry* ComputeVtables(object obj, CreateComInterfaceFlags flags,
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: > If I implement my own QI - can I just register IUnknown and nothing else here? 

Yep, that is an option. More advanced than even what we have here.

>I assume the runtime provided QI implementation uses the returned table to do its job, right?

Right.

> Is it valid (expected) to return different interface set per-instance (even of the same type)?

That is possible for sure but it would make things difficult to follow and possibly fail in some COM scenarios depending on how the type is exposed.

docs/standard/native-interop/tutorial-comwrappers.md
vitek-karas: > The how is an implementation detail I'd rather not document as it is always subject to change. Also, it shouldn't be relevant in the majority of cases.

I completely agree that it's an implementation detail - I was just curious :-)

docs/standard/native-interop/tutorial-comwrappers.md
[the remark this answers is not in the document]
vitek-karas: > the types the ComWrappers implementation supports should be left alone because you never know what type will appear from COM.

That is absolutely true - but what if the code which uses the ComWrappers is not used... in that case the entire ComWrappers class can be removed. Unless it's rooted somewhere else, like the module constructor.
It's also at odds with "pay for play" - the app would pay the price of initializing it up front, without knowing if it will need it or not.

docs/standard/native-interop/tutorial-comwrappers.md
vitek-karas: Sorry - does that mean that I should PInvoke into `CoCreateInstance` then? (I agree that it's probably pretty simple to do so, I'm just curious if there's better way)

AaronRobinsonMSFT: Yep. Implicit activation has been removed from the runtime but could technically be re-built with `ComWrappers` like C#/WinRT.

docs/standard/native-interop/tutorial-comwrappers.md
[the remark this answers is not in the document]
AaronRobinsonMSFT: >It's also at odds with "pay for play" - the app would pay the price of initializing it up front, without knowing if it will need it or not.

I would like to think more about that UX. The `ComWrappers` system is going to include substantial code gen, so why generate it in the first place if it isn't going to be needed? In fact that seems like a layering issue rather than push something so big into the module initializer vs a `static` type initializer. Perhaps a compromise here would be to add a comment that states module initializers may not be the best approach for trimmer support?

docs/standard/native-interop/tutorial-comwrappers.md
```diff
@@ -0,0 +1,430 @@
...
+- Marshalling and unmarshalling types across the COM/.NET boundary.
+- Two distinct approaches to consuming native COM objects in .NET.
+- A recommended pattern for enabling custom COM interop in .NET 5 and beyond.
+
+All source code used in this tutorial is available in the [dotnet/samples repository](https://github.com/dotnet/samples/tree/main/core/interop/comwrappers/Tutorial).
+
+**C# definitions**
+
+```csharp
+interface IDemoGetType
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: Oh boy. This is meant to be `IDemoGet` type rather than "get type".

docs/standard/native-interop/tutorial-comwrappers.md
AaronRobinsonMSFT: >Actually it should mention it either way - whether it's Windows specific or not.

That is part of the API definition. It is cross platform now. I believe we default to "it is cross-platform unless otherwise stated" &ndash; that seems to be the .NET promise. We updated the API metadata in .NET 6 for cross-platform.

docs/standard/native-interop/tutorial-comwrappers.md
AaronRobinsonMSFT: It is just a linear search in two sets of arrays. The implementation is relatively simple and documented: https://github.com/dotnet/runtime/blob/404a89f843b121dfb3909c645eda80cc3a2aa9cc/src/coreclr/interop/comwrappers.cpp#L370-L456. So much easier than the built-in system...

docs/standard/native-interop/tutorial-comwrappers.md
vitek-karas: It is surprising to me (in a good way). Mainly because I always think about COM as Windows-only. This is definitely COM, so I expected it to be Windows only. That's why I think it's worth pointing it out.

PR #25936, opened 2021-09-02
add reasons for var being nullable

### LINKED ISSUE #22997: Enhance the note about var and nullable reference types
The documentation notes:
"When var is used with nullable reference types enabled, it always implies a nullable reference type even if the expression type isn't nullable.
"

I think it should be worded more strongly, possibly adding something along the lines of ... If you don't want your variable to be nullable, always define its type fully; don't use var.

---
#### Document Details

* ID: 651ba093-2cef-e0f6-be71-ffdf6551c2c4
* Version Independent ID: a4878d20-2a12-74ad-9fc0-9f589a44af35
* Content: [var - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/var)
* Content Source: [docs/csharp/language-reference/keywords/var.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/keywords/var.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixes #22997

Point out that variables declared as `var` are nullable, but that's safe because the static analysis will ensure you don't dereference a null value.

### DIFF
--- a/docs/csharp/language-reference/keywords/var.md
+++ b/docs/csharp/language-reference/keywords/var.md
@@ -1,13 +1,12 @@
 ---
 description: "var - C# reference"
 title: "var - C# reference"
-ms.date: 10/02/2020
+ms.date: 09/02/2021
 f1_keywords: 
   - "var"
   - "var_CSharpKeyword"
 helpviewer_keywords: 
   - "var keyword [C#]"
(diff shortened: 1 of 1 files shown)


IEvangelist (approved):

PR #25940, opened 2021-09-02
Fix lint

gewarren:
Follow up to #25883


BillWagner (approved):
This LGTM.  I'll :shipit: so it gets published tonight.

PR #25958, opened 2021-09-04
Document DynamicDependencyAttribute

sbomer:
Adds documentation for DynamicDependencyAttribute in the instructions for preparing libraries for trimming.


docs/core/deploying/prepare-libraries-for-trimming.md
```diff
@@ -177,6 +177,43 @@ static void InitializeTypeField()
 
 In this case the trim analysis will simply keep public methods of `System.Tuple`, and will not produce further warnings.
 
+### DynamicDependency
+
+This attribute can be used to indicate that a member has a dynamic dependency on other members. This results in the referenced members being kept whenever the member with the attribute is kept, but doesn't silence warnings on its own.
+This can be used together with `UnconditionalSuppressMessageAttribute` to fix some analysis warnings.
```
vitek-karas, line 183: I would add a warning like:
Usage of `DynamicDependencyAttribute` should be limited to cases where no other approach is usable. Trimming tools can't validate that the dependency specified by the attribute will be enough to make the application work, so it's very likely trimming will produce warnings even if the usage of the attribute fully covers the use case. If such warnings are suppressed (via `UnconditionalSuppressMessageAttribute`) trimmer will also not validate that potential future changes in the code are correctly reflected in the usage of `DynamicDependencyAttribute`.

vitek-karas: In short - we don't want people to use this - unless they absolutely have to.

vitek-karas (approved):
Other than the warning which I think should be added, this looks great. Thanks!

docs/core/deploying/prepare-libraries-for-trimming.md
```diff
@@ -177,6 +177,43 @@ static void InitializeTypeField()
...
+
+```csharp
+[DynamicDependency("Helper", "MyType", "MyAssembly")]
+static void RunHelper()
+{
+    var helper = Assembly.Load ("MyAssembly").GetType("MyType").GetMethod("Helper");
```
eerhardt, line 189:
```suggestion
    var helper = Assembly.Load("MyAssembly").GetType("MyType").GetMethod("Helper");
```

docs/core/deploying/prepare-libraries-for-trimming.md
```diff
@@ -177,6 +177,43 @@ static void InitializeTypeField()
...
+    helper.Invoke(null, null);
+}
+```
+
+Without `DynamicDependency`, trimming might remove `Helper` from `MyAssembly` or remove `MyAssembly` completely if it's not referenced elsewhere, producing a warning that indicates a possible failure at runtime. The attribute ensures that `Helper` is kept.
+
+The attribute specifies the members to keep via a `string` or via `DynamicallyAccessedMemberTypes`. The type and assembly are either implicit in the attribute context, or explicitly specified in the attribute (by `Type`, or by `string`s for the type and assembly name).
+
+The type and member strings use the format described at C# documentation comment [format](https://github.com/dotnet/csharplang/blob/master/spec/documentation-comments.md#id-string-format), without the member prefix. The member string should not include the name of the declaring type, and may omit parameters to keep all members of the specified name. Some examples of the format follow:
```
eerhardt, line 198: (nit) Is this format doc'd on docs.microsoft.com at all? I wonder how much linking we should be doing from docs.microsoft.com to github docs. Especially considering the branch `master` might be renamed at some point. Or the doc could be moved.

docs/core/deploying/prepare-libraries-for-trimming.md
```diff
@@ -177,6 +177,43 @@ static void InitializeTypeField()
...
+```csharp
+[DynamicDependency("Method()")]
+[DynamicDependency("Method(System,Boolean,System.String)")]
+[DynamicDependency("MethodOnDifferentType()", typeof(ContainingType))]
+[DynamicDependency("MemberName")]
+[DynamicDependency("MemberOnUnreferencedAssembly", "ContainingType", "UnreferencedAssembly")]
+[DynamicDependency("MemberName", "Namespace.ContainingType.NestedType", "Assembly")]
+// generics
+[DynamicDependency("GenericMethodName`1")]
+[DynamicDependency("GenericMethod``2(``0,``1)")]
```
eerhardt, line 209: Why the double ``` `` ``` here?

eerhardt (approved):
Thanks!

docs/core/deploying/prepare-libraries-for-trimming.md
sbomer: All of the generic method ones were supposed to use double ` `` ` actually - I fixed `GenericMethodName` to do the same. That's just the format for generic method parameters (generic parameters on types use one, generic parameters on methods use two).

docs/core/deploying/prepare-libraries-for-trimming.md
sbomer: I moved this down to the "advanced" section and added some warnings - PTAL.

docs/core/deploying/prepare-libraries-for-trimming.md
sbomer: Turns out it does exist on docs.microsoft.com - I fixed the link. Thanks!

vitek-karas (approved):

PR #25961, opened 2021-09-04
Logically inconsistent `Explicitly Generic Constructs` example snippet

ghost:
## Summary

Was declared like `val function2<'T> : x:obj -> y:obj -> unit` and it was possible to call it like `function2 1. "Hi"`, so I found that it should be changed to logically be identic with the previous declaration of `function1`.

### DIFF
--- a/samples/snippets/fsharp/lang-ref-1/snippet1703.fs
+++ b/samples/snippets/fsharp/lang-ref-1/snippet1703.fs
@@ -1,2 +1,2 @@
-let function2<'T> x y =
+let function2<'T> (x: 'T) (y: 'T) =
     printfn "%A, %A" x y


BillWagner:
ping @dsyme for review.

dulanov:
Some extra context, `function2` was descrabed [here](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/generics/#explicitly-generic-constructs), `function1` has signature `val function1 : x:'a -> y:'a -> unit` and was declared on the same page early like:

```fsharp
let function1 (x: 'a) (y: 'a) =
    printfn "%A %A" x y
```

dsyme (approved):

dsyme:
Thank you! Yes this is a correct fix

PR #25973, opened 2021-09-06
Fix compile error in CA5394 solution code

Youssef1313:
Seems like `GetInt32` is a static method

https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.randomnumbergenerator.getint32?view=net-5.0


gewarren (approved):
Thanks @Youssef1313

PR #25986, opened 2021-09-07
Update table in versioning-sdk-msbuild-vs.md

StephenBonikowsky:
Update Lifecycle table in as well as a couple comments in versioning-sdk-msbuild-vs.md


StephenBonikowsky:
@adegeo could you merge this if everything looks good?

docs/core/porting/versioning-sdk-msbuild-vs.md
```diff
@@ -31,11 +31,15 @@ The support timeframe for the SDK typically matches the VS it's included in.
 | 5.0.4xx          | 16.11              | 16.8                                | Aug '21      | Feb '22*  |
 | 6.0.100-preview4 | 16.10 (No VS)      | 16.10-preview3                      | May '21      | N/A       |
 | 6.0.100-preview5 | 17.0-preview 1     | 16.10                               | June '21     | N/A       |
+| 6.0.100-preview6 | 17.0-preview 2     | 16.10                               | July         | N/A       |
+| 6.0.100-preview7 | 17.0-preview 3     | 16.10                               | August       | N/A       |
+| 6.0.100-rc1      | 17.0-preview 4     | 16.10                               | September    | N/A       |
 | 6.0.100          | 17.0               | 16.10**                             | Nov. '21     |           |
 
+**Note:** The Minimum version is what's required to load the SDK but doesn't cover what's supported. Targeting net6.0 is officially supported in 17.0+ only.
```
adegeo, line 39: We don't use `**Note**` syntax, either a full note or make it standalone paragraph.

```suggestion
> [!NOTE]
> The Minimum version is what's required to load the SDK but doesn't cover what's supported. Targeting net6.0 is officially supported in 17.0+ only.
```

docs/core/porting/versioning-sdk-msbuild-vs.md
```diff
@@ -31,11 +31,15 @@ The support timeframe for the SDK typically matches the VS it's included in.
...
 > \* MSbuild/VS supported for longer
 >
-> \*\* You shouldn't have to upgrade to 17.0 on the day .NET 6 releases. This is the current version and will likely be 16.11 at release.
+> \*\* You shouldn't have to upgrade to 17.0 on the day .NET 6 releases. This is the current version and will likely be 16.11 at release. This version will support downlevel targeting i.e. 5.0 and below but does not allow targeting .NET 6 for the purpose of using .NET 6 or C#10 features.
```
adegeo, line 42: - This should be active voice.
- If it's an unknown right now (shouldn't being used in the sense of "you shouldn't have to, but we don't know yet") then I would remove it.
- I'm unsure what "This is the current version" means, what version is the current version? 16.10? 17.0?
- I think the versioning for C# is `C# 10` not `C#10`
- Use conjunctions `does not` should be `doesn't`

```suggestion
> \*\* You won't have to upgrade to 17.0 on the day .NET 6 releases. This is the current version and will likely be 16.11 at release. This version supports downlevel targeting, such as .NET 5 and below, but doesn't allow targeting .NET 6 for the purpose of using .NET 6 or C# 10 features.
```

Overall though, I find this note confusing and I don't understand "doesn't allow targeting .NET 6 for the purpose of using .NET 6" that doesn't make sense to me.

adegeo (changes requested):
Hey there! A couple of things about this I find confusing. Perhaps you can update the wording?

docs/core/porting/versioning-sdk-msbuild-vs.md
StephenBonikowsky: "This is the current version" refers to 16.10 which is where the "**" for this note are. It will be updated at release.

This is a confusing topic and difficult to express without getting wordy, let me think about it for a bit and see if I can improve it.

docs/core/porting/versioning-sdk-msbuild-vs.md
```diff
@@ -36,7 +36,8 @@ The support timeframe for the SDK typically matches the VS it's included in.
 | 6.0.100-rc1      | 17.0-preview 4     | 16.10                               | September    | N/A       |
 | 6.0.100          | 17.0               | 16.10**                             | Nov. '21     |           |
 
-**Note:** The Minimum version is what's required to load the SDK but doesn't cover what's supported. Targeting net6.0 is officially supported in 17.0+ only.
+> [!NOTE]
+> The Minimum version is what's required to load the SDK but doesn't cover what's supported. Targeting net6.0 is officially supported in 17.0+ only.
 > \* MSbuild/VS supported for longer
```
adegeo, line 41: You'll need an empty line between the `>` to separate the note from the general `*` and `**` markers.

adegeo (approved):
Looks good!

StephenBonikowsky:
Thanks @adegeo I'll work on the wording of that note for the next update to this file.

PR #25990, opened 2021-09-08
Update extension-methods.md

DejanPopovic1:
Compared the use of extension methods to use without extension methods thus enhancing the distinguishing features.


docs/csharp/programming-guide/classes-and-structs/extension-methods.md
```diff
@@ -42,6 +42,12 @@ int i = s.WordCount();
 
 You invoke the extension method in your code with instance method syntax. The intermediate language (IL) generated by the compiler translates your code into a call on the static method. The principle of encapsulation is not really being violated. Extension methods cannot access private variables in the type they are extending.
 
+If extension methods were not used, the `WordCount` method would instead be invoked as:
+```csharp
+string s = "Hello Extension Methods";
+int i = MyExtensions.WordCount(s);
+```
```
IEvangelist, line 49: Technically, `WordCount` is always an extension method regardless of how it is used because its method signature uses the `this` identifier. But I do see what you're trying to communicate and I agree it is valuable. You might want to consider something like this?

> Both the `MyExtensions` class and the `WordCount` method are `static`, and it can be accessed like all other `static` members. The `WordCount` method can be invoked like other `static` methods as follows:
> 
> ```csharp
> string s = "Hello Extension Methods";
> int i = MyExtensions.WordCount(s);
> ```
> 
> The preceding C# code:
> 
> - Declares and assigns a new `string` named `s` with a value of `"Hello Extension Methods"`.
> - Calls `MyExtensions.WordCount` given `s`

IEvangelist (changes requested):

BillWagner:
@DejanPopovic1 Can you address the review comments? Once you do that, this is ready to merge.

docs/csharp/programming-guide/classes-and-structs/extension-methods.md
DejanPopovic1: All good points, really didn't change your suggestion at all, except I added a word in the second bullet point. Reading it end to end makes a lot more sense

IEvangelist (approved):

PR #26031, opened 2021-09-09
Add binder breaking change

### LINKED ISSUE #25779: [Breaking change]: Microsoft.Extensions.Configuration binder will now bind single elements to an array
### Description

Related to https://github.com/dotnet/runtime/pull/57204.

Added an app context switch in 6.0 for configuration binder. As of .NET 6.0 the configuration binder will bind single elements to an array unless setup otherwise via the `Microsoft.Extensions.Configuration.BindSingleElementsToArray` context switch.

#### More context:

This was originally requested as an API addition to BinderOptions, with details explained in issue https://github.com/dotnet/runtime/issues/57325#issuecomment-900508614, but we decided to go with the app context approach instead.

### Version

.NET 6 RC1

### Background and motivation

Previously with JSON Configuration Provider:
```json
{
   "tvshow": {
        "metadata": [
            {
               "name": "PrisonBreak"
            }
         ]
    }
}
```

would bind to below without any problem:
```cs
public class TvShow
{
  public Metadata[] Metadata { get; set; }
}

public class Metadata
{
  public string Name { get; set; }
}
```

However binding to this with an XML configuration provider given input below:
```xml
<TvShow>
  <Metadata>
    <Name>Prison Break</Name>
  </Metadata>
</TvShow>
```

would have no way of indicating that `TvShows` can contain multiple `Metadata` elements and that is bound to an array, so we fail to bind when a single element is provided as the key doesn't contain any `index` in the target array as the binder expects when binding to an array, and we return a `Metadata[]` with a `null` object.

### New behavior

To support binding to array elements for XML configuration provider, we changed the binder to account for a `key` without the `index` suffix when the child elements don't contain any index (a single element).

This is a breaking for other providers where this is not expected. Consider changing the original JSON from:

```json
{
   "tvshow": {
        "metadata": [
            {
               "name": "PrisonBreak"
            }
         ]
    }
}
```

so that `tvshows` is an object with a single `metadata`, this would bind even though json is not specifying an array:

```json
{
   "tvshow": {
        "metadata": {
            "name": "PrisonBreak"
          }
    }
```

### Type of breaking change

### Reason for change

Partner request

### Recommended action

If the user wants to switch back to old behavior they could set the app context switch below:

```C#
AppContext.SetSwitch("Microsoft.Extensions.Configuration.BindSingleElementsToArray", false);
``` 

#### Risks:

This breaking change could impact perf a little bit but only if the flag is set to `true`.

### Feature area

C#

### Affected APIs

`ConfigurationBinder.Bind(..)`

https://apisof.net/catalog/8210dea9-7ecf-ca70-bb5a-9842c62effa4

gewarren:
Fixes #25779

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/extensions/6.0/bind-single-elements-to-array?branch=pr-en-us-26031).

### DIFF
--- a/docs/core/compatibility/extensions/6.0/bind-single-elements-to-array.md
+++ b/docs/core/compatibility/extensions/6.0/bind-single-elements-to-array.md
@@ -51,7 +51,7 @@ The JSON configuration provider would not bind that JSON because it does not spe
 }
 ```
 
-However, when binding the following XML in previous .NET versions, the XML configuration provider has no way of knowing that `TvShows` can contain multiple `Metadata` elements and that they are bound to an array. Thus, the binder fails to bind when a single element is provided. The key doesn't contain any index in the target array, which the binder expects when binding to an array, and the binder returns a `Metadata[]` array with a `null` object.
+In addition, when binding the following XML in previous .NET versions, the XML configuration provider has no way of knowing that `TvShows` can contain multiple `Metadata` elements and that they are bound to an array. Thus, the binder fails to bind when a single element is provided. The key doesn't contain any index in the target array, which the binder expects when binding to an array, and the binder returns a `Metadata[]` array with a `null` object.
 
 ```xml
 <TvShow>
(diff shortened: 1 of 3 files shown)


docs/core/compatibility/extensions/6.0/bind-single-elements-to-array.md
```diff
@@ -0,0 +1,108 @@
...
+  public Metadata[] Metadata { get; set; }
+}
+
+public class Metadata
+{
+  public string Name { get; set; }
+}
+```
+
+The JSON configuration provider would not bind that JSON because it does not specify an array. The JSON would need to appear as follows in order to bind:
```
gewarren, line 40: I'm not sure if I explained this correctly. Please correct if necessary. I'm trying to show what's change for the JSON provider.

BillWagner:

safern (approved):
Thank you!

Issue #25676, opened 2021-08-17
Parameterless struct constructors


BillWagner:
[Feature specification](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/parameterless-struct-constructors.md)

This would allow structs to declare field initializers and no-argument constructors.

See https://github.com/dotnet/docs/pull/24234#discussion_r632409776 for note on language reference.

BillWagner:
This feature likely touches more articles than most, but those edits are likely smaller.

Major changes are needed on:

- [Language reference - struct types](https://docs.microsoft.com/dotnet/csharp/language-reference/builtin-types/struct)
- Programming guide - section on constructors](https://docs.microsoft.com/dotnet/csharp/programming-guide/classes-and-structs/constructors)

- The entire C# docset should be searched for "constructor", "default constructor" and "parameterless constructor" to remove any restriction on struct types and parameterless constructors. In addition, the statements on the compiler generating a parameterless constructor should be checked.
-  The docset should also be searched for "initializer" and "field initialization" to update any rules on field initializers for struct types.
-  Pages for arrays should be checked for how arrays of struct types are initialized. These should be checked for accuracy in light of record structs and parameterless constructors.

pkulikov:
@BillWagner is that fine if I update the language reference article on structs for this item?

BillWagner:
@pkulikov 

> Is that fine if I update the language reference article on structs for this item?

Yes, please.

pkulikov:
@BillWagner I'll continue with that and will check the [article about constructors](https://docs.microsoft.com/dotnet/csharp/programming-guide/classes-and-structs/constructors) as the next step.

pkulikov:
@BillWagner after #26310 is merged, we can say that almost all items from https://github.com/dotnet/docs/issues/25676#issuecomment-900594950 are done. EXCEPT, the two: (1) entire C# docset should be searched for "constructor" and (2) The docset should also be searched for "initializer". I've searched for those entries but capitulated when I saw the number of results. So, it's your call to decide if this item can be closed.

BillWagner:
Thanks for all the details @pkulikov 

I just merged #26310.  I'll do a final search for constructor and initializer. 

My goal in that note was to find any instances where the docs said that a struct couldn't have either initializers or a parameterless constructor. If we don't have any, I'll close this.

BillWagner:
[...]

PR #26037, opened 2021-09-10
C# 10 reference update: Add the parameterless constructor section to the struct reference

pkulikov:
Contributes to #25676 (only updates the **Structure types** reference article)

### DIFF
--- a/docs/csharp/language-reference/builtin-types/struct.md
+++ b/docs/csharp/language-reference/builtin-types/struct.md
@@ -97,7 +97,7 @@ Beginning with C# 10.0, you can declare a parameterless instance constructor in 
 
 :::code language="csharp" source="snippets/shared/StructType.cs" id="ParameterlessConstructor":::
 
-As the preceding example shows, the [default value expression](../operators/default.md) ignores a parameterless constructor and produces the default value of a structure type, which is the value produced by setting all value-type fields to their [default values](default-values.md) and all reference-type fields to `null`. Structure-type array instantiation also ignores a parameterless constructor and produces an array populated with the default values of a structure type.
+As the preceding example shows, the [default value expression](../operators/default.md) ignores a parameterless constructor and produces the default value of a structure type, which is the value produced by setting all value-type fields to their [default values](default-values.md) (the 0-bit pattern) and all reference-type fields to `null`. Structure-type array instantiation also ignores a parameterless constructor and produces an array populated with the default values of a structure type.
 
 Beginning with C# 10.0, you can also initialize an instance field or property at its declaration, as the following example shows:
 
--- a/docs/csharp/language-reference/builtin-types/struct.md
+++ b/docs/csharp/language-reference/builtin-types/struct.md
@@ -103,9 +103,12 @@ Beginning with C# 10.0, you can also initialize an instance field or property at
 
 :::code language="csharp" source="snippets/shared/StructType.cs" id="FieldInitializer":::
 
-As the preceding example shows, if you don't declare a parameterless constructor, the implicit parameterless constructor produces the default value of a structure type, regardless of field initializers. However, if a structure type has no explicit instance constructors and has field initializers, the compiler synthesizes a public parameterless constructor that performs the specified field initializations, as the following example shows:
+If you don't declare a parameterless constructor explicitly, a structure type provides a parameterless constructor whose behavior is as follows:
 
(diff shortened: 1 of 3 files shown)


docs/csharp/language-reference/builtin-types/struct.md
```diff
@@ -77,14 +77,40 @@ When you design a structure type, you have the same capabilities as with a [clas
...
+
+As the preceding example shows, the [default value expression](../operators/default.md) ignores a parameterless constructor and produces the default value of a structure type, which is the value produced by setting all value-type fields to their [default values](default-values.md) and all reference-type fields to `null`. Structure-type array instantiation also ignores a parameterless constructor and produces an array populated with the default values of a structure type.
+
+Beginning with C# 10.0, you can also initialize an instance field or property at its declaration, as the following example shows:
+
+:::code language="csharp" source="snippets/shared/StructType.cs" id="FieldInitializer":::
+
+As the preceding example shows, if you don't declare a parameterless constructor, the implicit parameterless constructor produces the default value of a structure type, regardless of field initializers. However, if a structure type has no explicit instance constructors and has field initializers, the compiler synthesizes a public parameterless constructor that performs the specified field initializations, as the following example shows:
```
BillWagner, line 106: The current wording took me a few times to understand correctly. How about this alternative:

```suggestion
The preceding example demonstrates that the implicit parameterless constructor performs the specified field initializations only if no instance constructors are declared. If you write an explicit parameterless constructor, that constructor must assign all instance fields in its body. 
```

BillWagner (commented):
This is great @pkulikov 

I had a couple comments. Let me know what you think, and then this is ready to :shipit:

docs/csharp/language-reference/builtin-types/struct.md
pkulikov: @BillWagner 

> The preceding example demonstrates that the implicit parameterless constructor performs the specified field initializations only if no instance constructors are declared.

As far as I understand, that's not correct, because there is a difference between the *implicit* parameterless constructor that *always* returns a zeroed instance and the *synthesized* parameterless constructor. This is how I've read this section in spec: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/parameterless-struct-constructors#constructors

Or you're fine to mix the implicit and synthesized parameterless constructors? Not sure, as the synthesized constructor is still explicit, but just written by the compiler.

> If you write an explicit parameterless constructor, that constructor must assign all instance fields in its body.

That's not correct, as the instance fields with field initializers are not required to be assigned in the body of the constructor. In general, I think it's fine to omit that sentence, as the corresponding compiler error message is clear enough.

I'll need to think more about wording here, I see.

docs/csharp/language-reference/builtin-types/struct.md
```diff
@@ -77,14 +77,40 @@ When you design a structure type, you have the same capabilities as with a [clas
...
 
 - You can't declare a [finalizer](../../programming-guide/classes-and-structs/destructors.md) within a structure type.
 
+## Parameterless constructors and field initializers
+
+Beginning with C# 10.0, you can declare a parameterless instance constructor in a structure type, as the following example shows:
+
+:::code language="csharp" source="snippets/shared/StructType.cs" id="ParameterlessConstructor":::
+
+As the preceding example shows, the [default value expression](../operators/default.md) ignores a parameterless constructor and produces the default value of a structure type, which is the value produced by setting all value-type fields to their [default values](default-values.md) and all reference-type fields to `null`. Structure-type array instantiation also ignores a parameterless constructor and produces an array populated with the default values of a structure type.
```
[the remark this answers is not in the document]
pkulikov: Agree, thank you @BillWagner

docs/csharp/language-reference/builtin-types/struct.md
pkulikov: @BillWagner I've rewritten the paragraph. What do you think of the new version?

BillWagner: Looking back at my sharplab.io history, I see what I did wrong. I had one field without an initializer when I was testing. Reviewing again.

BillWagner (approved):
This now looks great @pkulikov 

Thanks for helping me work through those nuances.  I'll :shipit: now.

Issue #25421, opened 2021-07-31
Warnings help discover original design intent – backhanded null states guarantees 


Uzivatel919:
From [Warnings help discover original design intent](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/upgrade-to-nullable-references#warnings-help-discover-original-design-intent)

> There's quite a bit going on in the preceding block of code. This application uses the [AutoMapper](https://automapper.org/) NuGet package to construct a news item from an `ISyndicationItem`. You've discovered that the news story items are constructed and the properties are set in that one statement. That means the design for the `NewsStoryViewModel` indicates that these properties should never have the `null` value. These properties should be **nonnullable reference types**. That best expresses the original design intent. In fact, any `NewsStoryViewModel` is correctly instantiated with non-null values. That makes the following initialization code a valid fix:

This paragraph is particularly weird:

1. Common practice is to validate input parameters all the time. No matter _Automapper_ takes its part in processing inputs from `RssFeedReader`/`ISyndicationItem` no code or textual description expresses guarantees about values stored in `ISyndicationItem`. Even if [RSS 2.0 `<title>` element](https://validator.w3.org/feed/docs/rss2.html) is `required`, it is not obvious why `ISyndicationItem.Title` must have value. Original design is in this case wrong.

2. If something **should** not have `null` value, it does not mean it **could not** have or **does not**. When it **should not**, it is likely `nullable annotated reference type` with some attributes further specifying `null state` circumstances. 

> That means the design for the `NewsStoryViewModel` indicates that these properties should never have the `null` value. These properties should be **nonnullable reference types**.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 313ec178-bb7c-e6be-0760-f7db427b464c
* Version Independent ID: ac631425-4ff7-850c-9fa7-127ff7a9ac06
* Content: [Upgrade to nullable reference types](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/upgrade-to-nullable-references)
* Content Source: [docs/csharp/whats-new/tutorials/upgrade-to-nullable-references.md](https://github.com/dotnet/docs/blob/main/docs/csharp/whats-new/tutorials/upgrade-to-nullable-references.md)
* Product: **dotnet-csharp**
* Technology: **csharp-null-safety**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

[...]

Issue #25420, opened 2021-07-31
Tutorial: Migrate existing code with nullable reference types – terminology


Uzivatel919:
From [Tutorial: Migrate existing code with nullable reference types](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/upgrade-to-nullable-references)

>C# 8 introduces **nullable reference types**, which complement reference types the same way nullable value types complement value types. You declare a variable to be a **nullable reference type** by appending a `?` to the type. For example, `string?` represents a nullable `string`. You can use these new types to more clearly express your design intent: some variables _must always have a value_, others _may be missing a value_. Any existing variables of a reference type would be interpreted as a non-nullable reference type.

Every time I pass through some documentation on `nullable reference types` I hit some bewildering connotation:

1. Term `nullable reference type` appears more as some new feature of reference types than some terse technical name. FMPOV, `nullable reference type` is shorthand for `nullable annotated reference type` but throughout the documentation `nullable reference type` (exactly `nullable annotated reference type`) is presented as some new kind of type. 

> You can use these new types to more clearly express your design intent:

Further in article it is stated 

> These two properties cause `CS8618`, "Non-nullable property is uninitialized". That's clear enough: both `string` properties have the default value of `null` when a `NewsStoryViewModel` is constructed. 

It is obvious that even `non-nullable` (`annotated`) `reference type` can be of `null` value.

I have some doubts if such explanations/expositions can be ever healthy for .NET documentation.

2. Statement
> C# 8 introduces **nullable reference types**, which complement reference types the same way nullable value types complement value types. 

is rather questionable. In fact for `value type` it was impossible to be of `null` value before NVT. For `reference type` nothing changed regard its ability to be of `null` value when NRT were introduced.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 313ec178-bb7c-e6be-0760-f7db427b464c
* Version Independent ID: ac631425-4ff7-850c-9fa7-127ff7a9ac06
* Content: [Upgrade to nullable reference types](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/upgrade-to-nullable-references)
* Content Source: [docs/csharp/whats-new/tutorials/upgrade-to-nullable-references.md](https://github.com/dotnet/docs/blob/main/docs/csharp/whats-new/tutorials/upgrade-to-nullable-references.md)
* Product: **dotnet-csharp**
[...]

Issue #25419, opened 2021-07-31
Attributes extend type annotations – bad definition, bad article name


Uzivatel919:
From [Attributes extend type annotations](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies#attributes-extend-type-annotations)

>  Reference types should never be the null value, and nullable reference types must be checked against null before being dereferenced.

It is not true that `nullable annotated` `reference types` have to be checked before dereference. That is true only for those in `maybe null` state. 

“_…and nullable reference types must be checked against `null` when it is not know whether they are not `null` before being dereferenced._” or similar is more exact.

---

> The attributes that describe the semantics of your API are found in the article on Attributes that impact nullable analysis.

Article under link _Attributes that impact nullable analysis_ has slightly different title – [_Attributes for null-state static analysis_](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis).

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: fc516159-d7e1-46ba-7d31-538a0b811491
* Version Independent ID: 99552131-8839-7c22-fcdc-2baa7722e756
* Content: [Update your codebase to use nullable reference types](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies#generic-definitions-and-nullability)
* Content Source: [docs/csharp/nullable-migration-strategies.md](https://github.com/dotnet/docs/blob/main/docs/csharp/nullable-migration-strategies.md)
* Product: **dotnet-csharp**
* Technology: **csharp-null-safety**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixed in #26046

Issue #25416, opened 2021-07-31
Update libraries to use nullable reference types and communicate nullable rules to callers – *null state* does not exists


Uzivatel919:
From [Update libraries to use nullable reference types and communicate nullable rules to callers](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies)

> If the `TryGetMessage` method returns `true`, the value of `message` isn't null. If the return value is `false`, the value of `message` (and its null state) is null.

Let focus on parenthesized phrase “_(and its null state)_”.

From [Nullable references and static analysis](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types#nullable-references-and-static-analysis)

> In a nullable enabled context, the compiler performs static analysis on variables of any reference type, both nullable and non-nullable. The compiler tracks the null state of each reference variable as either _not null_ or _maybe null_. 

Thus there is no _null state_ . Exist only `maybe null` and `not null`.

For another reference see https://github.com/dotnet/roslyn/issues/55110#issuecomment-886302535

>To fix this, we would probably have to introduce something like "definitely null" flow state in addition to the existing "maybe null" and "not null" flow states. 

I believe that “_(and its null state)_” is there by accident. That since whole paragraph states common expected situation regardless `nullable context` state (on, off, …).

The same wording can be found at [Attributes for null-state static analysis](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis#preconditions-allownull-and-disallownull).

---

#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: fc516159-d7e1-46ba-7d31-538a0b811491
* Version Independent ID: 99552131-8839-7c22-fcdc-2baa7722e756
* Content: [Update your codebase to use nullable reference types](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies#generic-definitions-and-nullability)
* Content Source: [docs/csharp/nullable-migration-strategies.md](https://github.com/dotnet/docs/blob/main/docs/csharp/nullable-migration-strategies.md)
* Product: **dotnet-csharp**
* Technology: **csharp-null-safety**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixed in #26046

PR #26046, opened 2021-09-10
Improve guidance on enabling nullable reference types

### LINKED ISSUE #25417: Choose a strategy for nullable reference types – senseless phrase(?), grammar
From [Choose a strategy for nullable reference types](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies#choose-a-strategy-for-nullable-reference-types).

---

> If any developers on your team forget, that new code is now in the backlog of work to make all code nullable aware.

This sentence likely misses some conclusion. Or am I wrong?

---

> The more mature and stable your project, the better the second strategy.

Does that 1st phrase miss _is_? (_The more mature and stable your project **is**._)

---
#### Document Details

* ID: fc516159-d7e1-46ba-7d31-538a0b811491
* Version Independent ID: 99552131-8839-7c22-fcdc-2baa7722e756
* Content: [Update your codebase to use nullable reference types](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies#generic-definitions-and-nullability)
* Content Source: [docs/csharp/nullable-migration-strategies.md](https://github.com/dotnet/docs/blob/main/docs/csharp/nullable-migration-strategies.md)
* Product: **dotnet-csharp**
* Technology: **csharp-null-safety**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

### LINKED ISSUE #25516: Tutorial: Express your design intent more clearly with nullable and non-nullable reference types – typos
> One lesson from this tutorial is that you made decisions about which variables could or could not be null.

Shouldn’t it be _makes_ or _will make_ instead of _made_.

---

> You've declared that variable to be non-nullable. Its value can't be null.

> The AllParticipants member must take into account that the respondents variable might be null, but the return value can't be null.

Is it really understood this situation as CANNOT? Isn’t it SHOULD NOT, rather. Or “_Its value is considered `non-null`_”. Once again `nullable-annotation` cannot enforce `non-null` value by default. Even if `nullability` _warnings_ are replaced with _errors_ `null-state` static analysis is not bulletproof. E.g. `!` can bring on this.

---

>Enable nullable support.

Already enabled at this place.

---

>  its return signature returns a non-nullable type.

- `Signature` returns nothing. It states.

---
#### Document Details

* ID: 3197c2df-22bd-abab-ac79-e2ab9f43cb92
* Version Independent ID: 2af238b0-3b5e-8add-c356-2230d01d4f9c
* Content: [Design with nullable reference types](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/nullable-reference-types)
* Content Source: [docs/csharp/whats-new/tutorials/nullable-reference-types.md](https://github.com/dotnet/docs/blob/main/docs/csharp/whats-new/tutorials/nullable-reference-types.md)
* Product: **dotnet-csharp**
* Technology: **csharp-null-safety**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Several changes here as well.

- remove the migration tutorial. It didn't have the scope to be useful.

Fixes #25516, #25421, #25420, #26076

- edit and improve the guidance on migrations.

Fixes #25417, #25416, #24980, #25419

- Add article on techniques to address warnings.


docs/csharp/nullable-warnings.md
```diff
@@ -0,0 +1,130 @@
...
+Console.WriteLine(message!.Length);
+```
+
+The null forgiving operator makes the expression *not-null* even if it was *maybe-null* without the `!` applied. In this example, a better solution is to add an attribute to the signature of `IsNotNull`:
+
+:::code language="csharp" source="./snippets/null-warnings/NullTests.cs" id="AnnotatedNullCheck":::
+
+The <xref:System.Diagnostics.CodeAnalysis.NotNullWhenAttribute?displayProperty=nameWithType> informs the compiler that the argument used for the `obj` parameter is *not-null* when the method returns `true`. When the method returns `false`, the argument has the same *null-state* it had before the method was called.
+
+There's a rich set of attributes you can use to describe how your methods and properties affect *null-state*. You can learn about them in the language reference article on [Nullable static analysis attributes](language-reference/attributes/nullable-analysis.md).
```
IEvangelist, line 44:
```suggestion
> [!TIP]
> There's a rich set of attributes you can use to describe how your methods and properties affect *null-state*. You can learn about them in the language reference article on [Nullable static analysis attributes](language-reference/attributes/nullable-analysis.md).
```

I suggest promoting this to a tip as it's really good callout, but it's your call.

docs/csharp/nullable-warnings.md
```diff
@@ -0,0 +1,130 @@
...
+:::code language="csharp" source="./snippets/null-warnings/PersonExamples.cs" id="PersonExample":::
+
+Neither `FirstName` nor `LastName` are guaranteed initialized. If this code is new, consider changing the public interface. The above example could be updated as follows:
+
+:::code language="csharp" source="./snippets/null-warnings/PersonExamples.cs" id="WithConstructor":::
+
+If you require creating a `Person` object before setting the name, you can initialize the properties using a default non-null value:
+
+:::code language="csharp" source="./snippets/null-warnings/PersonExamples.cs" id="Initializer":::
+
```
IEvangelist, line 90: I'm curious, is there ever a valid use case for using the null forgiving operator to say that you're ok with it being `null` to start out, even though you define the type as non-nullable?

```csharp
public class Person
{
    public string FirstName { get; set; } = null!;
    public string LastName { get; set; } = null!;
}
```

docs/csharp/nullable-warnings.md
```diff
@@ -0,0 +1,130 @@
...
+Another alternative may be to change those members to nullable reference types. The `Person` class could be defined as follows if `null` should be allowed for the name:
+
+:::code language="csharp" source="./snippets/null-warnings/PersonExamples.cs" id="NullableMember":::
+
+Existing code may require other changes to inform the compiler about the null semantics for those members. You may have created multiple constructors, and your class may have a private helper method that initializes one or more members. The <xref:System.Diagnostics.CodeAnalysis.MemberNotNullAttribute?displayProperty=nameWithType> and <xref:System.Diagnostics.CodeAnalysis.MemberNotNullWhenAttribute.%23ctor%2A?displayProperty=nameWithType> attributes inform the compiler that a member is *not-null* after the method has been called.
```
IEvangelist, line 95: I do not believe that you intended to have it render this way, but I see the fully qualified type and the type name.type name (which feels a bit redundant). I think the .ctor wasn't rendering correctly as an xref.

image

IEvangelist (approved):
Overall this is awesome, thank you! I have a few bits of feedback.

tdykstra (approved):
The new article is great, a big improvement in understandability over what it replaces -- the structure makes it easy to find what you need if you're looking for something specific, and the explanations and examples are clear and complete. Thank you very much for doing this! If I run into situations not covered here I'll let you know and we can add new H2s or supplement existing ones.

docs/csharp/nullable-warnings.md
BillWagner: I'd rather not recommend it. It's valid, but you're telling the compiler "I trust all callers to do the right thing." I might use it it legacy code, where I don't want to introduce breaking changes. Otherwise, no.

PR #26051, opened 2021-09-11
Source gen overloads breaking change

### LINKED ISSUE #25987: [Breaking change]: New JsonSerializer source-gen overloads can cause source breaking changes when options param is null
### Description

The System.Text.Json source generator feature added new overloads to `JsonSerializer` accepting pre-generated type information via `JsonTypeInfo<T>`/`JsonSerializerContext`, as a perf optimization over pre-existing overloads that take `JsonSerializerOptions` instances and perform runtime reflection. All these types are reference types which could be `null`. See below as an example of the method-signature patterns for both approaches:

**Reflection/`JsonSerializerOptions`-based overloads**
- `public static string JsonSerializer.Serialize<T>(T value, JsonSerializerOptions? options = null)`;
- `public static string JsonSerializer.Serialize(object value, Type type, JsonSerializerOptions? options = null)`;
- `public static T JsonSerializer.Deserialize<T>(string json, JsonSerializerOptions? options = null)`;
- `public static T JsonSerializer.Deserialize(string json, Type type, JsonSerializerOptions? options = null)`;

**Src-gen/`JsonTypeInfo`/`JsonSerializerContext`-based overloads**
- `public static string JsonSerializer.Serialize<T>(T value, JsonTypeInfo<T> jsonTypeInfo)`;
- `public static string JsonSerializer.Serialize(object value, Type type, JsonSerializerContext jsonSerializerContext)`;
- `public static T JsonSerializer.Deserialize<T>(string json, JsonTypeInfo<T> jsonTypeInfo)`;
- `public static object JsonSerializer.Deserialize(string json, Type type, JsonSerializerContext jsonSerializerContext)`;

### Version

.NET 6 Preview 6

### Previous behavior

A customer could have written code that passes `null` as the value for the `JsonSerializerOptions` parameter and it would have been fine.

```cs
entity.Property(e => e.Value).HasConversion(v => JsonSerializer.Serialize(v,null), v => JsonSerializer.Deserialize(v, null));
```

### New behavior

With the new source-gen methods in 6.0 there would be compiler ambiguity such as:

```
"The call is ambiguous between the following methods or properties: 'JsonSerializer.Serialize(TValue, JsonSerializerOptions?)' and 'JsonSerializer.Serialize(TValue, JsonTypeInfo)'"
```

Users now have to update their code in a manner that disambiguates the intended overload, such as performing an explicity cast to the intended target:

```cs
entity.Property(e => e.Value).HasConversion(v => JsonSerializer.Serialize(v, (JsonSerializerOptions)null), v => JsonSerializer.Deserialize(v, (JsonSerializerOptions)null));
```

Other workarounds include omitting the optional parameter `JsonSerializerOptions? options = null` or using named arguments. Note that these options are not possible within lambda expressions and cannot be used as workarounds in those scenarios. 

### Type of breaking change

### Reason for change

New overloads were added to the serializer as a performance optimization. See https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-source-generator/.

### Recommended action

Described in "New behavior" above.

### Feature area

Core .NET libraries

### Affected APIs

All the `JsonSerializer` methods as shipped in .NET 5.0 - https://docs.microsoft.com/dotnet/api/system.text.json.jsonserializer?view=net-5.0.

gewarren:
Fixes #25987 

[Preview link.](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/jsonserializer-source-generator-overloads?branch=pr-en-us-26051)

### DIFF
--- a/docs/core/compatibility/6.0.md
+++ b/docs/core/compatibility/6.0.md
@@ -46,6 +46,7 @@ If you're migrating an app to .NET 6, the breaking changes listed here might aff
 | [FileStream no longer synchronizes file offset with OS](core-libraries/6.0/filestream-doesnt-sync-offset-with-os.md) | Preview 4 |
 | [FileStream.Position updates after ReadAsync or WriteAsync completes](core-libraries/6.0/filestream-position-updates-after-readasync-writeasync-completion.md) | Preview 4 |
 | [New diagnostic IDs for obsoleted APIs](core-libraries/6.0/diagnostic-id-change-for-obsoletions.md) | Preview 5 |
+| [New JsonSerializer source generator overloads](core-libraries/6.0/jsonserializer-source-generator-overloads.md) | Preview 6 |
 | [New System.Linq.Queryable method overloads](core-libraries/6.0/additional-linq-queryable-method-overloads.md) | Preview 3-4 |
 | [Older framework versions dropped from package](core-libraries/6.0/older-framework-versions-dropped.md) | Preview 5 |
 | [Partial and zero-byte reads in DeflateStream, GZipStream, and CryptoStream](core-libraries/6.0/partial-byte-reads-in-streams.md) | Preview 6 |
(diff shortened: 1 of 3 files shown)


tdykstra (approved):
LGTM

gewarren:
Thanks Tom!

PR #26057, opened 2021-09-11
Adding declared variables to the identifier names examples

wilsonneto-swe:
Constants and variables don't seem to be the same thing, then I think would be a good idea to add the "declared constants" in the list of when we use identifiers declarations.

## Summary

Constants and variables don't seem to be the same thing, then I think would be a good idea to add the "declared constants" in the list of when we use identifiers declarations..

### DIFF
--- a/docs/csharp/fundamentals/coding-style/identifier-names.md
+++ b/docs/csharp/fundamentals/coding-style/identifier-names.md
@@ -5,7 +5,7 @@ ms.date: 08/21/2018
 ---
 # Identifier names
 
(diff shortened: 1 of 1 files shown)


BillWagner:
hi 👋  @wilsonneto-dev 

I'm going to close this PR without merging it. My reasoning is that the article is correct as it is written. A *constant* is a *member*. If we call out *constants*, we would need to do the same for all other types of members: fields, methods, properties, and so on. 

Having said that, thank you for your contribution, and I encourage you to submit other PRs where you think we can improve. We appreciate it.

PR #26071, opened 2021-09-12
Add note about indentation of match and DU constructors when using less than 4 spaces

cmeeren:
Indenting pattern match "bodies" (to the right of the arrow) and DU constructor fields one level works fine with 4 spaces of indentation, but with 2 spaces of indentation, due to the initial `| `, the rules make the body/fields line up with the clause/constructor, which makes it more difficult to visually separate the clauses/constructors and the bodies/fields.

This PR fixes that by instructing to indented 4 spaces if using less than 2 spaces of indentation, to ensure sufficient separation.

/cc @cartermp

### DIFF
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -592,6 +592,34 @@ type SynBinding =
         seqPoint: DebugPointAtBinding
 ```
 
+If using less than 4 spaces of indentation, the fields should still be indented 4 spaces to ensure sufficient visual separation between the cases and the fields.
+
+```fsharp
+// Using 2 spaces of indentation
+
+// OK
+type MyUnion =
+  | Case1 of
+      field1: int *
+      field2: string *
+      field3: bool
+  | Case2 of
+      field1: int *
+      field2: string *
+      field3: bool
+      
+// Not OK
+type MyUnion =
+  | Case1 of
+    field1: int *
+    field2: string *
+    field3: bool
+  | Case2 of
+    field1: int *
+    field2: string *
+    field3: bool
+```
+
 You can also use triple-slash `///` comments.
 
 ```fsharp
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -971,7 +971,7 @@ match lam with
 
 ```
 
-If using less than 4 spaces of indentation, the expression should still be indented 4 spaces to ensure sufficient separation between the clauses and the subexpression.
+If using less than 4 spaces of indentation, the expression should still be indented 4 spaces to ensure sufficient visual separation between the clauses and the subexpression.
 
 ```fsharp
 // Using 2 spaces of indentation
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -971,6 +971,28 @@ match lam with
 
 ```
 
+If using less than 4 spaces of indentation, the expression should still be indented 4 spaces to ensure sufficient separation between the clauses and the subexpression.
+
+```fsharp
+// Using 2 spaces of indentation
+
+// OK
+match lam with
+| Var v -> 1
+| Abs(x, body) ->
+    1 + sizeLambda body
+| App(lam1, lam2) ->
+    sizeLambda lam1 + sizeLambda lam2
+
+// Not OK
+match lam with
+| Var v -> 1
+| Abs(x, body) ->
+  1 + sizeLambda body
+| App(lam1, lam2) ->
+  sizeLambda lam1 + sizeLambda lam2
+```
+
 Pattern matching of anonymous functions, starting by `function`, should generally not indent too far. For example, indenting one scope as follows is fine:
 
 ```fsharp


nojaf:
Please stop pushing for this. It is annoying and getting very tiresome.
It is not worth having such an exception. The style guide should not have special rules based on the size of the indentation.
The normal flow is being followed here.

cmeeren:
You are more than welcome to disagree, but I am truly of the opinion that some constructs need to be adjusted for the indentation size to ensure proper visual separation.

The style guide was clearly written with 4 spaces of indentation in mind (in the sense that some problems emerge only with e.g. 2 spaces of indentation and simply do not exist with 4 spaces of indentation), yet the guide explicitly opens up for other indentation levels (in particular, 2 or 3 spaces). I have seen a lot of codebases in the wild with 2 spaces of indentation, which means that ideally, we should make sure that the style guide produces ideal formatting also for this indentation level. I only want what's best for the F# community (which does not exclude the fact that I am scratching my own itch here).

> Please stop pushing for this. It is annoying and getting very tiresome.

I am not sure what you mean by "this"; are you referring only to this particular issue? Based on your wording (in particular, "getting"), I am getting the impression that my annoying you have been going on for some time, though this particular suggestion is not something I have been actively pushing for for a long time. I am therefore unsure what you mean by "this". I am sorry that you find my suggestions tiresome. Still, I stand by my suggestion, for the reasons mentioned above.

cmeeren:
On another note, this is also relevant for DU constructor fields. I updated the PR to reflect this.

BillWagner:
ping @dsyme for his opinion.

cmeeren:
> ping @dsyme for his opinion.

Yes, it might be good to have @dsyme's opinion on the underlying issue, as detailed in my comment above (https://github.com/dotnet/docs/pull/26071#issuecomment-918070600). I see three ways forward:
* Leave the style guide as-is, explicitly opening up for <4 spaces of indentation but with some problems with visual separation that emerge when using 2 spaces of indentation. I'd say this is is not ideal.
* Make 4 spaces the only officially recommended indentation in the style guide. Then there would be no "<4-space problems". Many codebases in the wild use 2 spaces and will then be in violation. I myself use 2 spaces, and this change would probably make me abandon the style guide and any formatters strictly adhering to it, and just format the code myself (as I mostly do already).
* Adapt the guide to <4 spaces of indentation where needed (not many places, AFAIK - some are already adapted, and this PR contains only other change I currently find is needed). Auto-formatting tooling would need to adapt (likely not terribly complicated, as previously mentioned). The style guide would be better overall, IMHO.

dsyme (dismissed):

dsyme:
cc @nojaf 

@nojaf - this also relates to the 
```fsharp
let f ()=
    { abc with
      DEF = 1 }
```
we were discussing.  Consistency with what's proposed here would mean 4-space indentation
```fsharp
let f ()=
    { abc with
        DEF = 1 }
```

nojaf:
I'm a bit surprised here, to be honest.
The guide should not make rules based on a specific number of indentations.

How on earth are you going to explain this to newcomers? If you want to use a certain indent size, it's alright, but at random points, we'll make exceptions for subjective visual preferences.

```fsharp
if a then
  // 2
  b
else
  c

match a with
| true ->
    // 4
    b
| false ->
    // 4
   c
```

People that choose a smaller indent like 2 spaces, generally want just want to go with 2 spaces.
If they feel that visual separation is not good enough with 2 spaces, then why not use 4 spaces in the first place?

I hope you can understand that I find this a very subjective matter and I prefer consistency instead. With such changes, the risk of confusing users is real, and they will open issues with Fantomas, like this one:  https://github.com/fsprojects/fantomas/issues/1234.

dsyme:
@nojaf Yes, on re-reading you are correct. At first glance I thought it was highlighting cases which gave offside warnings or errors, but of course it isn't.  I need more coffee.

@cmeeren Does that make sense?  I think given the feedback we should not merge this

dsyme:
@cmeeren Thank you for your reply :)

I personally don't have any strong opinion about "best" 2-space formatting, so the Fantomas perspective seems to be the right one to make a call - that is @nojaf is asking himself questions like "can we implement this, will we get requests saying we're not implementing it correctly, and can we justify the exceptions with a simple set of consistent rules" 

So I will close this base on that, thanks

dsyme:
> Let me again reiterate that I have never intended to create trouble or extra work just for the sake of it. The style guide opens up for 2 spaces of indentation, and I just wanted that to look as good as possible.

@nojaf I think perhaps it is best if we remove 2-space indentation from the style guide and document it in Fantomas?

nojaf:
All hi, thanks for understanding.
I think the guide should express itself always in units of indentation, regardless of what concrete number of spaces has been chosen.
Believe it or not, some people even use 3 or 5 as indent_size. It might be less confusing to have all the samples with 4 spaces as indentations. We could document potential shortcomings in Fantomas, although these will be limited I guess.

dsyme:
Hmmm..   I think it depends on whether the doc is a "style spec" or "something simple to make F# beginners feel happy".  If the former then "units of indentation", if the latter then "assume four space".  

The revamp I've done organises it more along the "something simple to make F# beginners feel happy" lines.  But I don't want to lose the detailed "spec" quality of it.

PR #26088, opened 2021-09-13
MSBuild breaking change

### LINKED ISSUE #26048: [Breaking change]: MSBuild 17 no longer supports calling GetType()
### Description

MSBuild 17 no longer supports calling the `GetType()` instance method within property functions. This allowed unpredictable code execution during evaluation and could cause Visual Studio hangs.

### Version

.NET SDK 6.0.100-rc1

### Previous behavior

`GetType()` calls in MSBuild property functions would execute sometimes causing unpredictable behavior in VS.

### New behavior

An error during project evaluation `The function "GetType" on type "System.String" is not available for execution as an MSBuild property function.`

### Type of breaking change

### Reason for change

This functionality was not documented and not commonly used. It caused performance and reliability issues with project loading, especially in Visual Studio.

The only known common use of this was in the [CBT](https://commonbuildtoolset.github.io/) system which has been deprecated.

### Recommended action

Replace any calls to `GetType()` with alternate MSBuild logic.

### Feature area

SDK

### Affected APIs

_No response_

gewarren:
Fixes #26048 

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/6.0/calling-gettype-property-functions?branch=pr-en-us-26088)


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

gewarren:
Thanks Bill!

PR #26095, opened 2021-09-14
Change --install-dir documentation to reflect reality

kojoru:
## Summary

The linux `dotnet-install.sh` script says the following:
```
Install Location:
  Location is chosen in following order:
    - --install-dir option
    - Environmental variable DOTNET_INSTALL_DIR
    - $HOME/.dotnet
```

If you review the source code, it matches this documentation. However, the online documentation claims that the default directory is `/usr/share/dotnet`, which is not true. It _is_ the directory where `apt-get` installs it globally though, but I don't think it's relevant in this doc.


tdykstra:
Closing, as this issue was addressed by #26665

PR #26109, opened 2021-09-15
Trim orphaned images

gewarren:
Contributes to #25539


IEvangelist (changes requested):
The docs/media/* images are being as implicit metadata. They should not be deleted, as they are actually refenced as part of the JSON config.

gewarren:
@IEvangelist Could you take another look?

IEvangelist (approved):

PR #26157, opened 2021-09-16
Document aggressive trimming mode

MichalStrehovsky:
## Summary

The default value of `TrimMode` is `link` in .NET 6 so the "aggressive mode" wasn't actually doing anything.


sbomer (approved):
LGTM! I know we were hesitant to document this one, but it's already mentioned at https://docs.microsoft.com/en-us/dotnet/core/deploying/prepare-libraries-for-trimming, and I think it makes sense here since it is talking about aggressive trimming. @vitek-karas might have an opinion on it.

vitek-karas:
I don't have much problem with this - one would get warnings either way. It's just that without full aggressive trimming the app has higher chance of working even with warnings.

PR #26159, opened 2021-09-16
Nit: fix typos in expanded LTTng name

usewits:
Got the correct expanded name here: https://lttng.org/docs/v2.13/#doc-nuts-and-bolts
Linux Trace Toolkit: next generation

The expanded name listed on Wikipedia is close but doesn't match the capitalization:
Linux Trace Toolkit Next Generation
Seems reasonable to use the name listed on lttng.org instead.


gewarren (commented):
Thank you @usewits

gewarren (approved):

PR #26161, opened 2021-09-16
Fixed nonsensical phrase: "The can use the null forgiving operator"

Psycho900:
I'm pretty sure they meant "YOU can use the null forgiving operator" instead of "THE can use the null forgiving operator" (which doesn't make grammatical sense).

## Summary

Fixed a typo in the original document


BillWagner (approved):
Thanks @Psycho900 

I'll :shipit: now.

Psycho900:
That was fast! Thanks a bunch 🙂

PR #26163, opened 2021-09-16
Ux study fixes from test run

DavidCBerry13:
When doing the internal test run, found a number of typos and such, so fixing those before the first actual participant


CamSoper (approved):
LGTM! Good luck with the user study!

IEvangelist (approved):

PR #25937, opened 2021-09-02
First draft of topic for calling http3 from .NET 6

samsp-msft:
## Summary

New topic for .NET 6 RC1.

PR #26170, opened 2021-09-16
Add article to TOC

adegeo:
## Summary

#25937 was merged without being added to TOC. This add it to the TOC.

@BillWagner


BillWagner (approved):
Great catch @adegeo 

Let's :shipit: now.

/cc @samsp-msft

PR #26172, opened 2021-09-16
Update editorconfig example

### LINKED ISSUE #22161: EditorConfig Docs differs from VS2019 templates
Hi,

Not sure if this is the right place to report / ask this question, so perhaps it is better suited as a discussion.
The default .editorconfig different from the one generated by Visual Studio as described here https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2019
Not only the section descriptions and the amount of sections differ, but the settings in them and the amount of settings being set differ. The VS one covers *, cs, csx, vb and vbx files instead of only C# files. 

So I am really wondering 
- which template should I choose
- how do I update the template with new conventions, analysers etc come available
- which files should .NET target?

Personally I like the style from VS2019 better, with the enclosed # and perhaps I would expect links to the docs regarding sections on the docs. Especially since these are extensions of the editorconfig spec. 

Should there be a sync between the generated one in vs2019 and the one here in the docs? Which Is recommended to be used?

The update part might be hard, since every project can customise which rules to use etc as demonstrated by the samples by the linked runtime and roslyn samples.

Some guidance in this regard would be welcome

Thanks,
Max

---
#### Document Details

* ID: e8c8357d-11a6-6f12-b044-5ff44df65d10
* Version Independent ID: 0c103fd7-5cc4-2726-40f8-d759419ae635
* Content: [.NET code style rule options - .NET](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/code-style-rule-options?view=vs-2019)
* Content Source: [docs/fundamentals/code-analysis/code-style-rule-options.md](https://github.com/dotnet/docs/blob/master/docs/fundamentals/code-analysis/code-style-rule-options.md)
* GitHub Login: @gewarren

gewarren:
Copied from the .NET editorconfig template in Visual Studio. Fixes #22161.

### DIFF
--- a/docs/fundamentals/code-analysis/code-style-rule-options.md
+++ b/docs/fundamentals/code-analysis/code-style-rule-options.md
@@ -35,213 +35,138 @@ To help you get started, here is an example *.editorconfig* file with the defaul
 > In Visual Studio 2019 and later versions (on Windows), you can generate this file and save it to a project at **Tools** > **Options** > **Text Editor** > [**C#** or  **Basic**] > **Code Style** > **General**. Then, click the **Generate .editorconfig file from settings** button. For more information, see [Code style preferences](/visualstudio/ide/code-styles-and-code-cleanup).
 
 ```ini
-# Remove the line below if you want to inherit .editorconfig settings from higher directories
+###############################
+# Core EditorConfig Options   #
+###############################
 root = true
-
-# C# files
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @gewarren 

You can :shipit:

PR #26178, opened 2021-09-17
remove coining of "vanity alignment"

smoothdeveloper:
motives: 
* vanity sounds like a character and pejorative judgement
* some languages (haskell AFAIK) use similar alignment commonly, I have doubts it is referred as such at places where it is commonly used
* web search for "vanity alignment" doesn't return anything relevant but this document
 
I'd prefer the pejorative adjective not be associated with alignment styles, the justification for the guidelines stand on its own without the coining of this alignment style as vain.

## Summary

Describe your changes here.

### DIFF
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -56,7 +56,7 @@ let myLongValueName = someExpression
                       |> anotherExpression
 ```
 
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks @smoothdeveloper 

This is a good change.

I'll :shipit: now.

PR #26181, opened 2021-09-17
Note that .NET 6 projects are nullable enabled

### LINKED ISSUE #25483: New templates are nullable enabled
In the latest previews, the new app templates add  `<Nullable>enable</Nullable>` by default. Articles that discuss new projects, and articles on nullable reference types should add this.

BillWagner:
Fixes #25483

I didn't follow the notes in my last comment on #25483 because the "get started" tutorials don't examine the project file, or use constructs that behave different when Nullable is enabled. I believe calling out the different behavior would only distract from the get started experience.


tdykstra (approved):
LGTM

PR #26185, opened 2021-09-18
Add markdown code style to boolean value `false`

limuyuan:
## Summary

I found the boolean value true is applied to markdown code style(`true`), but “false” is not, so I added the code style to `false`.

N/A


BillWagner (approved):
Thanks for making this fix @limuyuan 

PR #26196, opened 2021-09-20
Update simd.md to fix Vector sample code issue

### LINKED ISSUE #26195: Vector sample code has issue which will cause out of range related exception
In https://github.com/dotnet/docs/blob/main/docs/standard/simd.md, Vector<T> sample code, it has logic of handling case about input array length not be integral multiple of Vector.Count (see green block), 
but first half of code cannot handle it and will throw out of range related exception if happens (see red block)

image

---
#### Document Details

* ID: 9e0f3c8a-1548-40cd-1688-d574c0d1db1d
* Version Independent ID: 40f9283e-0200-4398-7dc3-00c2bca59e81
* Content: [SIMD-accelerated types in .NET](https://docs.microsoft.com/en-us/dotnet/standard/simd)
* Content Source: [docs/standard/simd.md](https://github.com/dotnet/docs/blob/main/docs/standard/simd.md)
* GitHub Login: @FIVIL
* Microsoft Alias: **tagoo**

lateapexearlyspeed:
## Summary

In Vector<T> sample code, it has logic of handling case about input array length not be integral multiple of Vector.Count, but first half of code cannot handle it and will throw out of range related exception if happens. This PR is for that.

Fixes #26195

### DIFF
--- a/docs/standard/simd.md
+++ b/docs/standard/simd.md
@@ -80,7 +80,7 @@ double[] SimdVectorProd(double[] left, double[] right)
     var offset = Vector<double>.Count;
     double[] result = new double[left.Length];
     int i = 0;
-    for (i = 0; i < left.Length; i += offset)
+    for (i = 0; i + offset <= left.Length; i += offset)
     {
         var v1 = new Vector<double>(left, i);
         var v2 = new Vector<double>(right, i);
--- a/docs/standard/simd.md
+++ b/docs/standard/simd.md
@@ -77,10 +77,10 @@ The example below demonstrates adding long arrays elements using <xref:System.Nu
 ```csharp
 double[] SimdVectorProd(double[] left, double[] right)
 {
-    var offset = Vector<double>.Count;
+    var vectorSize = Vector<double>.Count;
     double[] result = new double[left.Length];
     int i = 0;
-    for (i = 0; i + offset <= left.Length; i += offset)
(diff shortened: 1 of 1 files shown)


docs/standard/simd.md
```diff
@@ -80,7 +80,7 @@ double[] SimdVectorProd(double[] left, double[] right)
     var offset = Vector<double>.Count;
     double[] result = new double[left.Length];
     int i = 0;
-    for (i = 0; i < left.Length; i += offset)
+    for (i = 0; i + offset <= left.Length; i += offset)
```
gfoidl, line 83:
```suggestion
    for (; i <= left.Length - offset; i += offset)
```
* `i` is already initialized to `0`
* `left.Length - offset` can be kept in a register, and must not be re-evaluated on each iteration (so better performance)

Although I wouldn't use `offset`, rather inline `Vector<double>.Count` or name this variable `vectorSize` or that like. `offset` is quite misleading.
(I see this was there before this PR, but could fix this at the same time.)

lateapexearlyspeed: @gfoidl Fixed, thanks.

gewarren:
@tannergooding Could you review this code change?

tannergooding (approved):

tannergooding:
Sorry for the delay, I had missed the review request here.

PR #26233, opened 2021-09-23
Removed the mention of `DataTestMethod` attribute

### LINKED ISSUE #64: DataTestMethod vs TestMethod
I discovered today that `[DataRow]` can be used with together with `[TestMethod]`, where I had the understanding that `[DataTestMethod]` was required.

Searching both the [source code](https://github.com/Microsoft/testfx/search?l=C%23&q=datatestmethodattribute) and [docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.datatestmethodattribute), I'm a bit confused.

I've find no practical difference between the two attributes in the source code, but the documentation for `[DataTestMethod]` says that it is intended for data driven tests.
Then the [examples](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.datarowattribute#examples) for `[DataRow]` uses `[TestMethod]`.

Haplois:
## Summary

`DataTestMethod` attribute is no longer necessary, updated documentation to use modern way. 

Fixes https://github.com/microsoft/testfx-docs/issues/64

### DIFF
--- a/docs/core/testing/snippets/unit-testing-using-mstest/csharp/PrimeService.Tests/PrimeService_IsPrimeShould.cs
+++ b/docs/core/testing/snippets/unit-testing-using-mstest/csharp/PrimeService.Tests/PrimeService_IsPrimeShould.cs
@@ -22,7 +22,7 @@ namespace Prime.UnitTests.Services
         }
 
         #region Sample_TestCode
-        [DataTestMethod]
+        [TestMethod]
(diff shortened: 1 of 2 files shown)


BillWagner (approved):
Thanks for helping us improve @Haplois 

I've reviewed the changes, and I'll :shipit: now

PR #26240, opened 2021-09-23
Fix repo lint error

adegeo:
## Summary

Fix the repo lint error introduced by #26148

@BillWagner

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -4,9 +4,7 @@
     "MD004": false,
     "MD013": false,
     "MD024": false,
-    "MD025": {
-        "front_matter_title": ""
-    },
+    "MD025": false,
     "MD026": false,
     "MD028": false,
     "MD033": {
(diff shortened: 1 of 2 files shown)


tdykstra (commented):
We would have to do this in every article that uses tabs -- maybe it's preferable to just turn off md024 and md025 globally?

luisquintanilla:
Thanks @adegeo 

I merged this PR because there were no build issue other than the lint error which I didn't know you could make go away by adding the comment you added and as far as I know, the tab formatting is correct because the preview displayed correctly.

adegeo:
@tdykstra I think we could turn it off if the build system catches it, I suspect it might!

adegeo:
@luisquintanilla The lint errors carry over to every other PR. So we can't ignore them 😄

adegeo:
@dotnet/docs Please comment on disabling these markdown linter checks. I think @tdykstra is correct, it's the best way to just ensure tabbed content doesn't mess things up. The build system does catch additional H1

gewarren:
> @dotnet/docs Please comment on disabling these markdown linter checks. I think @tdykstra is correct, it's the best way to just ensure tabbed content doesn't mess things up. The build system does catch additional H1

I'm fine with that!

luisquintanilla (approved):

PR #26254, opened 2021-09-24
Update null-conditional-operators.md

### LINKED ISSUE #21812: First example on this page is misleading
The first example on this page indicates that the following code:

`Dim length As Integer? = customers?.Length`

is the same as this code:

`Dim length As Integer`
`If customers IsNot Nothing Then`
`    length = customers.Length`
`End If`

That implies that length will not be set to a value if customers is Nothing.  That's not true.  "customers?.Length" is actually equal to Nothing if customers is Nothing.  For instance if you have the following code:

`Dim customers as customer = Nothing`
`Dim length as Integer?`
`length = 10`
`length = customers?.length`

length will be equal to Nothing after executing this code.  The example on this page makes it seem that length is not altered if customers is Nothing.  That would mean length would remain equal to 10 if customers is Nothing.  But in reality, length will be set to Nothing after executing the above code.

So the example on the page should be the following instead:

`Dim length As Integer`
`If customers IsNot Nothing Then`
`    length = customers.Length`
`Else`
`    length = Nothing`
`End If`

This represents what actually happens in the code.  If customers is NOT nothing, length gets set to customers.Length.  If customers is nothing, length gets set to Nothing.

---
#### Document Details

* ID: 7427bc30-4d66-b234-4537-200d4419cc24
* Version Independent ID: 786cb51a-cd8d-95a7-a11c-c39245a01e9d
* Content: [Null-conditional Operators - Visual Basic](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/operators/null-conditional-operators)
* Content Source: [docs/visual-basic/language-reference/operators/null-conditional-operators.md](https://github.com/dotnet/docs/blob/master/docs/visual-basic/language-reference/operators/null-conditional-operators.md)

bergerb:
This PR addresses an example to closely match what is happening when using a null conditional operator in VB.  This was a suggestion from a user and I think it is an excellent idea!  Thanks!

## Summary

Clarity Update - update example to match what is occuring with a null conditional operator.

Fixes #21812

### DIFF
--- a/docs/visual-basic/language-reference/operators/null-conditional-operators.md
+++ b/docs/visual-basic/language-reference/operators/null-conditional-operators.md
@@ -31,6 +31,8 @@ For comparison, the alternative code for the first of these expressions without 
 Dim length As Integer
 If customers IsNot Nothing Then
    length = customers.Length
+else
+    length = Nothing
 End If
 ```
 
--- a/docs/visual-basic/language-reference/operators/null-conditional-operators.md
+++ b/docs/visual-basic/language-reference/operators/null-conditional-operators.md
@@ -31,7 +31,7 @@ For comparison, the alternative code for the first of these expressions without 
 Dim length As Integer
 If customers IsNot Nothing Then
    length = customers.Length
-else
+Else
     length = Nothing
 End If
 ```
--- a/docs/visual-basic/language-reference/operators/null-conditional-operators.md
+++ b/docs/visual-basic/language-reference/operators/null-conditional-operators.md
@@ -28,7 +28,7 @@ Dim count As Integer? = customers?(0)?.Orders?.Count()
 For comparison, the alternative code for the first of these expressions without a null-conditional operator is:
 
 ```vb
-Dim length As Integer
+Dim length As Integer?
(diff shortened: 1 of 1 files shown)


docs/visual-basic/language-reference/operators/null-conditional-operators.md
```diff
@@ -31,6 +31,8 @@ For comparison, the alternative code for the first of these expressions without
 Dim length As Integer
```
BillWagner, line 31: Should this be a nullable to match the example on line 19?

```suggestion
Dim length As Integer?
```

BillWagner (commented):
Thanks @bergerb 

I had one quick comment to make the two samples match. After we resolve that question, this is ready to merge.

docs/visual-basic/language-reference/operators/null-conditional-operators.md
bergerb: Thanks!

BillWagner (approved):
Thanks @bergerb 

I'll :shipit: now.

PR #26255, opened 2021-09-24
changed Recommend to Recommended in header

AdrianEdelen:
## Summary

just fixing what I believe is a typo in the header of the doc.


BillWagner (approved):
Thanks @AdrianEdelen 

You're fix is correct, and I'll :shipit: now.

PR #26256, opened 2021-09-24
Specify that rules for explicit generic type arguments also apply to function application

cmeeren:
It makes sense that the same rules should apply to explicit generic parameters in function application, too.

Currently Fantomas formats this in a way that breaks the "avoid name-sensitive alignment" clause:

```f#
Json.serialize<{| child: {| displayName: string; kind: string |}
                  newParent: {| id: string; displayName: string |}
                  requiresApproval: bool |}>
```

Here is a better formatting that is consistent with the rules for function/member/type definitions:

```f#
Json.serialize<
    {| child: {| displayName: string; kind: string |}
       newParent: {| id: string; displayName: string |}
       requiresApproval: bool |}>
```

I am updating the style guide to clarify this so there is no doubt, and so that Fantomas can hopefully be updated accordingly.

/cc @nojaf @dsyme

### DIFF
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -1568,7 +1568,7 @@ val SampleMixedFunction:
 
 ### Formatting explicit generic type arguments and constraints
 
-The guidelines below apply to function definitions, member definitions, and type definitions.
+The guidelines below apply to function definitions, member definitions, type definitions, and function applications.
 
 Keep generic type arguments and constraints on a single line if it’s not too long:
 
@@ -1612,6 +1612,24 @@ let inline f<^T1, ^T2
     and ^T2 : (member Foo3: string -> ^T1 option)>
     =
     // function body
+```
+
+The same rules apply for function applications:
+
+```f#
+// ✔️ OK
+myObj
+|> Json.serialize<
+    {| child: {| displayName: string; kind: string |}
+       newParent: {| id: string; displayName: string |}
+       requiresApproval: bool |}>
+       
+// ✔️ OK
+Json.serialize<
+    {| child: {| displayName: string; kind: string |}
+       newParent: {| id: string; displayName: string |}
+       requiresApproval: bool |}>
+    myObj
 ```
 
 ## Formatting attributes
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -1616,7 +1616,7 @@ let inline f<^T1, ^T2
 
 The same rules apply for function applications:
 
-```f#
+```fsharp
 // ✔️ OK
 myObj
 |> Json.serialize<


BillWagner:
ping @dsyme for review

dsyme:
@nojaf what do you think?  I see the inconsistency but type applications are perhaps a bit different

cmeeren:
> type applications are perhaps a bit different

I don't want to start a big discussion, just wondering, how is it different? With the current behavior of Fantomas, you still get potentially significant code chunks (as in the demo) moved far to the right, in violation of the "avoid name sensitive alignments" clause which is already part of the guide.

cmeeren:
For example, with Fantomas' current behavior, if you rename `serialize` in the example to `serialize2`, the code breaks (won't compile) and Fantomas is no longer able to serialize it. So I'd say it's fairly important that this is changed.

This PR is basically just to make it absolutely clear; I don't see it as a prerequisite for the Fantomas change due to the aforementioned reasons.

dsyme:
> With the current behavior of Fantomas, you still get potentially significant code chunks (as in the demo) moved off to the right and potentially breaking your code on rename (due to offsets changing). This is in violation of the "avoid name sensitive alignments" clause which is already part of the guide.

I see, yes, that's a good argument from principle. I've not often done multi-line generic argument instantiations so my intuitions aren't so good here. The only different I thought of is that the `serialize` and `<` need to be adjacent though that doesn't affect what you've proposed.

But what's the actual rule?  Is it that a multi-line anon record type should always start `{|` on a new line? So
```fsharp
Json.serialize<
    {| child:
        {| displayName: string
           longLongLongLongLongKind: string |}
       newParent:
           {| id: string
              longLongLongLongDisplayName: string |}
       requiresApproval: bool |}>
```
etc.?  It's not really about generic instantiations is it? THis would apply to any multi-line anon record type formatting?

cmeeren:
I don't get it. Sure, the code breaks due to offsets coming from the anonymous record. But the proposed change in this PR is still sound from a principle standpoint, as argued.

The formatting guidelines can of course introduce something about anonymous records specifically, but that is not itself relevant to this PR. The code breakage occur only because of the way Fantomas formats the generic parameters, which I guess it would do whether or not there is an anonymous record there.

dsyme:
@cmeeren I think your suggestion is fine, I just want to discuss with @nojaf what the actual rule is.

The rules immediately preceeding your addition are describing what to do with 

```
thing<lots, of, little, things>
```

Your example is 
```
thing<one-really-big-and-long-thing
           that-covers-multiplelines-via-inner-linebreaks>
```
and saying it should be 
```
thing<
    one-really-big-and-long-thing
    that-covers-multiplelines-via-inner-linebreaks>
```

That's ok - it's just that AFAICS it doesn't really following from the other rules described, and we should consider all the cases where this arises and check we are doing things consistently.

cmeeren:
I see. I'll let you two hash this out. 👍

cmeeren:
@nojaf, just a reminder that @dsyme wants your input here (in case you missed it amidst all the other talk here).

docs/fsharp/style-guide/formatting.md
```diff
@@ -1614,6 +1614,24 @@ let inline f<^T1, ^T2
     // function body
 ```
 
+The same rules apply for function applications:
+
+```f#
```
IEvangelist, line 1619:
```suggestion
```fsharp
```

IEvangelist (approved):

IEvangelist:
> This looks good, thank you. We'll :shipit: when you make the suggested changes.

Hi friends, hey @dsyme - are you good with this, or should it be changed?

nojaf:
Hello, I have missed all of this I'm afraid.

As for `SynExpr.Typed` (`Foo<Bar>`), I don't think there are a lot of rules in place.
Nothing really comes to mind, except for https://github.com/fsprojects/fantomas/issues/1611.

```fsharp
SomeDatabase.CreateCommand<"
                    select name
                    from things
                    where id = :id
   "            >
```
The position of `>` matters in the case of a string.

Fantomas currently does not have anything in place for doing something different when the type is multiline.

On a side note, I'm also wondering about putting the closing `>` on the next line:
```fsharp
Json.serialize<
    {| child: {| displayName: string; kind: string |}
       newParent: {| id: string; displayName: string |}
       requiresApproval: bool |}
>
```

The same for anonymous type definitions, nothing really is in place.
Regardless of length, we put everything on the right-hand side ([example](https://fsprojects.github.io/fantomas-tools/#/fantomas/preview?data=N4KABGBEDGD2AmBTSAuKAXAngB0WAGmALxjAA%2BY0AFgJYA28a5Y8NAztnQIaYByXAW0Ro26AE40AdgHMA3GADWUxmFESZYMgF8AOmJ2SIR40cmIA7gAUuYxJPRMKNFWqlyW7Tj35CR4t5q6%2BoYmobYAjgCuNLZsAILY2GKwAG5cdGgARrCwdHoGoaFckrDoVIhiAGrFNFgAKrQymI6UtAwtrBzcfILCqv4y8kqSLgPSgfkhhdMzs2ZWNnYOpE6j6u6dXj2%2B-esTwbOHR0YR0bEJSanpWTl0gYGQADTgUGyI6OhubKhgANovEFAoUgABIsLgfpApOgngCjKC0nRIsg0P9pgAGZ7TSAASRGSwAyjQAF7ILGFAAscIgAF04VpyYDqVAwTgUVBobDgSDEcifmjCgBGRkmSAAWS4AA8ADJSRDSuzSMpcoUAJnRzLpxgZcKBotZELQkDs8AA8gAzWVmAlYOhk5kI9J81HMiCqkXGSAAURGFqt9ux0DEdHNkE19JFes9BvZkGyuRV0d57IFoQAzB74Xi3mJ0AAxKTpXgWOhyxMmcTI8PayMOmOQ%2BN0csQR1IlOusAUzMtgmIAQ0OB0WCSOLoH1my1l7tgc3pN7Vow64xR%2BH1o2N5ss5P8jsAVmnkAJ2C40EQACFEObYLZrGJeugKpuIJXEAuIEujCuW2uoBvM63nT%2BDsADYDyPE9z0va95VgcwKmgLg3jxFJYAQz5hyfMAXzfMAPyZblwVjP8623F1pgAdjA49TwvK9bAAVUSeDEMQZDUK4dDJEw2c6HnUItUXWsCLZBtbk3AD22mAAOKiINo6CAGFuDYNgFOHNRImgdBr24udX34iNdTrQjRITf8eSdSTCgATlkmioNsMU%2B0yR9px4viTAE98hP1Ez1zE8zSKA6ZBUxDtD2oyC6MQNShy4ty9JwvDSGMkT-LMkjLJ3ELhXC8DTzic0HzENSBAELhMOwgyayM4TDV-ALMrbbK1TsqLoN7ftBwwhLeP0zzDOXVL6rjRruSC1MTEFDM8siwris6gdch6jsqoGmqhrqoixv1CaO0FLtwrxJB7FNSQ6jETAAHVaioXS%2BqSnzoz8hqMvGrKyKFfdZoguJkkiEYABFEFLftisqsQq2qwTat8tKOXscSLOaz7QkFUDwolSUcXNBo7C9PqCSoa90Bu%2BBlWnCkNWh7zYee%2BGoURwKPuCoVKMxqU8XNGhJVNXA720sQvUlJJEBUmgVumXdqfWmHNrhkbOWZlHWbRmSOclAAlRA4DEeAyYpjsqceunVxexmYWVwDJuMQVbI17XdfgXhIgEFyxAtHEHwEb5p0FE35fpkaBEiOhPlLMw82vcqPgqOoRKtqzQnVA9HeveAxVD8O5SjsQY-Bg9qBsE9ioAfXMZxlQDz9htjJWmutjtVVy7Esb%2Bu9MFNMRZVEA27spmXjC83CnrNhn6-elWbaMVV3Q19ueC7nv0Bdt2Kk973ff26v8IV2MQ7D%2Bgc%2BjjjivjiFE5a5OZuxBfO%2B79h0Ezw%2BI8QXP89c8Ki7vLSKnLyu7o7xSltSEE9dos2nm6Q6rcpTVDbGeZQbg%2B6YWNjTEeptvzmzAUmCBTdvowMlHmAGWkJaSAQSMJBACUGDyMMPZKX4WRYKZg3JOJhVQYwIYDUoABxd4wtRbi2HMg6c0sgEMNBObYik9G7TFVOzVuWcj5mDPEOaACgzw-wUO8NgZ0CS9Fiq7eKHZ3L9SHoNGuID0pNkvqjNh6tsTFnMK-C86A4J2HPogYG3NJC1FIXEEYTk15iC3tMExYja6mWsSwq%2BbD7bYgANKIEQNgHGeNJB4j0UIf090PJmI2hYveoDmHSNYcYNMYUCEE37GwKgwijY0NpOY3eQdtpvXAVPDsaYW6FEPG4O0f1pCuyWFdRAmQxQIADIUMJaD6ERKsUjPa0w0xz1vqWaQkgiGSBIcOIk6yOKRFsHUWAx0lgcVITk0xtCmnAMKfMmxqsTBphvj0uIYcKiSDOSkGCMhAnuy8VIXx6kLnhMsa9KJJSYllOgT05%2Bnx-TSkEJkeAXAlKwDYG4Rxr9gUzNHpghmUj2kyMKGmfBPTAbsC4JkO0VT2BUAJJgewUpsWy1poHMeI0CU4I6UsjhPTEnJJOfYPEGjijUGZXkuWBSWmRIWbgpZ8iekqOKAof08R-ojGLKIRAGdFGv34bEdFQLpxrQlayqV7LWngsJaUowaZ7GKpsPJWw5K2BBhoP2D5D54D0R8cOYG0BuAC3Ob1XJVz8nNItTK%2B5kCwBpjiT0m0EgtLjKQOKsNi4Xg0nJFCNgeZ0U-DCVoIAA))

PR #26269, opened 2021-09-26
Async dispose pattern mini fixes

deep-outcome:
## Summary
In commit order:

1. Unacceptable pattern – managed resource must be handled only in managed resources disposal block.
2. Same.
3. More clear definitions on `DisposeAsync` and `DisposeAsyncCore`. Code showing nothing unsaid removal.
4. Clarification on place of disposal, it is likely not known if variables refer to any instances (they were [even] checked for).
5. What happens on producer side (`AnotherAsyncDisposable`) is not conditional from-by consumer pattern. Thus is irrelevant to denote any responsibilities resulting from bad pattern on consumer side. It is not important whether variable is assigned or not. Important is that valid instance referred to by `objOne` `DisposeAsync` is not called while it should and can be.
----
Check also #26270.

### DIFF
--- a/docs/standard/garbage-collection/implementing-disposeasync.md
+++ b/docs/standard/garbage-collection/implementing-disposeasync.md
@@ -29,7 +29,6 @@ Signatures:
 - `public ValueTask DisposeAsync()` – an <xref:System.IAsyncDisposable.DisposeAsync?displayProperty=nameWithType> implementation
 - `protected virtual ValueTask DisposeAsyncCore()`
 
-
 ### The DisposeAsync() method
 
 The `public` parameterless `DisposeAsync()` method is called implicitly in an `await using` statement, and its purpose is to free unmanaged resources, perform general cleanup, and to indicate that the finalizer, if one is present, need not run. Freeing the memory associated with a managed object is always the domain of the [garbage collector](index.md). Because of this, it has a standard implementation:
--- a/docs/standard/garbage-collection/implementing-disposeasync.md
+++ b/docs/standard/garbage-collection/implementing-disposeasync.md
@@ -76,9 +76,9 @@ You may need to implement both the <xref:System.IDisposable> and <xref:System.IA
 
 The <xref:System.IDisposable.Dispose?displayProperty=nameWithType> and <xref:System.IAsyncDisposable.DisposeAsync?displayProperty=nameWithType> implementations are both simple boilerplate code.
 
-In the `Dispose(bool)` overload method, the <xref:System.IDisposable> instance is conditionally disposed of if it is not `null`. The <xref:System.IAsyncDisposable> instance is cast as <xref:System.IDisposable>, and if it is also not `null`, it is disposed of as well. Both instances are then assigned to `null`.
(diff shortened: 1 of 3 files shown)


docs/standard/garbage-collection/implementing-disposeasync.md
```diff
@@ -124,13 +120,10 @@ In the preceding example, each asynchronous clean up operation is implicitly sco
 
 ### Unacceptable pattern
 
-The highlighted lines in the following code show what it means to have "stacked usings". If an exception is thrown from the `AnotherAsyncDisposable` constructor, neither object is properly disposed of. The variable `objTwo` is never assigned because the constructor did not complete successfully. As a result, the constructor for `AnotherAsyncDisposable` is responsible for disposing any resources allocated before it throws an exception.
+The highlighted lines in the following code show what it means to have "stacked usings". If an exception is thrown from the `AnotherAsyncDisposable` constructor, neither object is properly disposed of.
 
 :::code language="csharp" id="dontdothis" source="snippets/dispose-async/ExamplePatterns.cs" highlight="9-10":::
 
-> [!TIP]
-> Avoid this pattern as it could lead to unexpected behavior.
```
IEvangelist, line 132: Why is this being removed? This should stay.

IEvangelist (changes requested):

IEvangelist:
Hi again @Uzivatel919 - in discussing the proposed changes with both @bartonjs and @sharwell, they confirmed that the moving for the `null` assignment is not necessary. For consistency with the tooling / analyzer comments that are provided during the implementation refactoring of both `Dispose` and `DisposeAsync` I'm going to close this PR. Thanks again for the contribution and having an open dialog.

docs/standard/garbage-collection/implementing-disposeasync.md
Uzivatel919: Please, this is eligible for removal since it gives **a tip** on phenomenon that is focused in whole chapter. I my casual day I would not appreciate **a tip** on anything being right 100 % obvious.

Uzivatel919:
> Hi again @Uzivatel919 - in discussing the proposed changes with both @bartonjs and @sharwell, they confirmed that the moving for the `null` assignment is not necessary. For consistency with the tooling / analyzer comments that are provided during the implementation refactoring of both `Dispose` and `DisposeAsync` I'm going to close this PR. Thanks again for the contribution and having an open dialog.

@IEvangelist, @bartonjs, @sharwell, I have to take an objection. Moving `null` assignment into respective _manage resources disposal block_ is not really necessary by virtue of code flow. On the other hand I would say anyone checking with these docs would be expecting that examples of pattern follows the patterns. In this case, separation of logic manipulating un/managed resource to respective code blocks.

@IEvangelist, here are also another changes beside moving `null` assignments. Their externality have not been explained.

docs/standard/garbage-collection/implementing-disposeasync.md
sharwell: ➡️ Users frequently search for code, and then copy paste whichever result shows up without reading all the surrounding text. The callout here is intended to highlight (via color/page style) the fact that copying this result is problematic.

Uzivatel919: Hmm… If I skip that right above is quite big heading **Unacceptable pattern** should not this be a note then?

PR #24234, opened 2021-05-14
Fix inaccurate information for optional parameters

Youssef1313:
`default(SomeType)` doesn't have to be a value type. See [SharpLab](https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEBMBGAsAKHQGYACLEgYRIG8CT6zT0AWEgWQAoqBjEgXhIIApgDMIAVwA2AFy4BKOTToMAvgRVA).

Closes https://github.com/dotnet/docs/issues/24225

PR #26282, opened 2021-09-27
C# docset: Mention explicit struct parameterless constructor where necessary

pkulikov:
Contributes to #25676

I've searched and checked only the "default constructor" and "parameterless constructor" entries in the C# doc set.

### DIFF
--- a/docs/csharp/language-reference/builtin-types/default-values.md
+++ b/docs/csharp/language-reference/builtin-types/default-values.md
@@ -1,7 +1,7 @@
 ---
 title: "Default values of C# types - C# reference"
 description: "Learn the default values of C# types such as bool, char, int, float, double and more."
-ms.date: 12/18/2019
+ms.date: 09/28/2021
 helpviewer_keywords: 
   - "default [C#]"
   - "parameterless constructor [C#]"
@@ -33,7 +33,9 @@ Beginning with C# 7.1, you can use the [`default` literal](../operators/default.
 int a = default;
 ```
 
-For a value type, the implicit parameterless constructor also produces the default value of the type, as the following example shows:
+## Parameterless constructor of a value type
+
+For a value type, the *implicit* parameterless constructor also produces the default value of the type, as the following example shows:
 
 ```csharp-interactive
 var n = new System.Numerics.Complex();
@@ -41,6 +43,9 @@ Console.WriteLine(n);  // output: (0, 0)
 ```
 
 At run time, if the <xref:System.Type?displayProperty=nameWithType> instance represents a value type, you can use the <xref:System.Activator.CreateInstance(System.Type)?displayProperty=nameWithType> method to invoke the parameterless constructor to obtain the default value of the type.
+
+> [!NOTE]
+> In C# 10.0 and later, a [structure type](struct.md) (which is a value type) may have an [explicit parameterless constructor](struct.md#parameterless-constructors-and-field-initializers) that may produce a non-default value of the type. Thus, we recommend using the `default` operator or the `default` literal to produce the default value of a type.
 
 ## C# language specification
 
--- a/docs/csharp/methods.md
+++ b/docs/csharp/methods.md
@@ -136,8 +136,11 @@ A method definition can specify that its parameters are required or that they ar
 The parameter's default value must be assigned by one of the following kinds of expressions:
 
 - A constant, such as a literal string or number.
+- An expression of the form `default(SomeType)`, where `SomeType` can be either a value type or a reference type. If it's a reference type, it's effectively the same as specifying `null`. Beginning with C# 7.1, you can use the `default` literal, as the compiler can infer the type from the parameter's declaration.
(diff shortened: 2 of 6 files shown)


docs/csharp/language-reference/builtin-types/default-values.md
```diff
@@ -42,6 +44,9 @@ Console.WriteLine(n);  // output: (0, 0)
 
 At run time, if the <xref:System.Type?displayProperty=nameWithType> instance represents a value type, you can use the <xref:System.Activator.CreateInstance(System.Type)?displayProperty=nameWithType> method to invoke the parameterless constructor to obtain the default value of the type.
 
+> [!NOTE]
+> In C# 10.0 and later, a [structure type](struct.md) (which is a value type) may have an [explicit parameterless constructor](struct.md#parameterless-constructors-and-field-initializers) that may produce a non-default value of the type. Thus, we recommend using the `default` operator or the `default` literal to produce the default value of a type.
```
pkulikov, line 48: @BillWagner Can I get into trouble even in earlier C# versions (let's say in C# 7) if I use the library written and compiled with C# 10?

docs/csharp/methods.md
```diff
@@ -136,8 +136,11 @@ A method definition can specify that its parameters are required or that they ar
...
 - An expression of the form `new ValType()`, where `ValType` is a value type. Note that this invokes the value type's implicit parameterless constructor, which is not an actual member of the type.
-- An expression of the form `default(SomeType)`, where `SomeType` can be either a value type or a reference type. If it's a reference type, it's effectively the same as specifying `null`.
+
+  > [!NOTE]
+  > In C# 10.0 and later, an expression of the form `new ValType()` is not allowed if it invokes the explicitly defined parameterless constructor of a value type. Use the `default(ValType)` expression or the `default` literal to provide the parameter's default value. For more information about parameterless constructors, see the [Parameterless constructors and field initializers](language-reference/builtin-types/struct.md#parameterless-constructors-and-field-initializers) section of the [Structure types](language-reference/builtin-types/struct.md) article.
```
pkulikov, line 143: This addresses https://github.com/dotnet/docs/pull/24234#discussion_r632409776

docs/csharp/language-reference/builtin-types/default-values.md
BillWagner: @pkulikov I don't think so. I suppose that if you were expecting the 0-bit pattern and didn't get it, that could be unexpected. But, I would think that would be part of the motivation for using a C# 10 library.

More important here, I think, would be to emphasize that the explicit parameterless constructor won't get called for arrays and stackalloc expressions.

docs/csharp/methods.md
BillWagner: I'm not sure what you mean by the first sentence in this paragraph. You can invoke the parameterless constructor. I don't have a suggestion, because I'm not sure what your purpose is.

BillWagner (changes requested):
Thanks @pkulikov 

I had one small suggestion, and a couple larger thoughts to address. Once we resolve those discussions, this is ready.

docs/csharp/methods.md
pkulikov: I mean that the following code generates the error `Error CS1736 Default parameter value for 'entry' must be a compile-time constant`:
```csharp
    public static void Display(Example entry = new Example())
    {    }

    public struct Example
    {
        public Example() { }
    }
```

The error disappears as soon as I remove the explicit parameterless constructor.

In that sense, "is not allowed" equivalent to "results in the compiler error"

pkulikov: @BillWagner I've rephrased the sentence, please check it again.

BillWagner: I like the way you rephrased this. It's more clear when the restriction applies, and why.

docs/csharp/language-reference/builtin-types/default-values.md
pkulikov: > More important here, I think, would be to emphasize that the explicit parameterless constructor won't get called for arrays and stackalloc expressions.

What makes it important in this article? This article is about the default values of C# types, and I'm afraid the content about parameterless constructor is starting to occupy too much space. We may say that the default values are used at array initialization, instead. What do you think?

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/struct#parameterless-constructors-and-field-initializers - already tells that explicit constructors are not called for arrays.

The article about `stackalloc` says that `The content of the newly allocated memory is undefined.` So, it's not even 0-bit. So, saying that explicit parameterless constructor is not called may confuse people into thinking that the default type value is used, which is not true.

BillWagner: Thanks @pkulikov 

I like your analysis here. I'll approve, and :shipit: now

BillWagner (approved):

PR #26283, opened 2021-09-27
--version-suffix example uses incorrect naming

jzabroski:
semver 2.0 should use `rc.1` not `rc1`, as per conversation I had with @joelverhagen where I was using it incorrectly.  I believe the example in this article is likely the root cause as to why I thought `rc1` was correct, so it would be great to update and fix this. See discussion here https://github.com/NuGet/NuGetGallery/issues/8325


tdykstra (approved):
Thanks for correcting this!

PR #26294, opened 2021-09-28
F# style guide: more examples (vanity alignment)

knocte:
## Summary

As a follow-up to #21487
and #21690 we can put more
examples here.

Completes #21487


BillWagner:
@dsyme

dsyme (approved):

PR #26295, opened 2021-09-28
C# 10: Add record struct to articles on records 

### LINKED ISSUE #24681: Update document with 'record' type
[Enter feedback here]

The "Types and Variables" section of this document needs to be updated to include the C# 9.0 'record' type.
---
#### Document Details

* ID: 14618620-83fc-fd68-93a9-03626c76d680
* Version Independent ID: ea7fbbaf-fd41-75cb-28fb-a2f36c8d3967
* Content: [A Tour of C# - C# Guide](https://docs.microsoft.com/en-us/dotnet/csharp/tour-of-csharp/)
* Content Source: [docs/csharp/tour-of-csharp/index.md](https://github.com/dotnet/docs/blob/main/docs/csharp/tour-of-csharp/index.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

### LINKED ISSUE #25484: Add articles on record structs
C# 10 adds `record struct` types. These types follow value semantics (as all value types do). The compiler generates many of the same methods as it does for `record` types. Some aren't applicable because they are only needed to support inheritance for `record` types.

In addition, `record` types can be declared as `record class` types to avoid potential confusion with `record struct` types.

Feature specification: [Record structs](https://github.com/dotnet/csharplang/issues/4334)

### LINKED ISSUE #26289: Equality in inheritance hierarchies does not clearly demonstrate EqualityContract
[Equality in inheritance hierarchies](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record#equality-in-inheritance-hierarchies) example states `and student == student2 returns True although one is a Person variable and one is a Student variable.`, however, in the example, both records are `Student`, so this does not demonstrate the `EqualityContract` clearly.
```
    Person student = new Student("Nancy", "Davolio", 3);
    ...
    Student student2 = new Student("Nancy", "Davolio", 3);
```
---
#### Document Details

* Content: [Records - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record#positional-syntax-for-property-definition)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixes #25484 
Fixes #26289
Fixes #24681

In addition to the notes on #25484, search the C# guide for references to `record` types, and determine if any updates are needed for record structs. There were a few small edits to remove the restriction that all records are reference types.


docs/csharp/fundamentals/types/records.md
```diff
@@ -29,16 +29,18 @@ An immutable type is one that prevents you from changing any property or field v
 
 Immutability isn't appropriate for all data scenarios. [Entity Framework Core](/ef/core/), for example, doesn't support updating with immutable entity types.
 
-## How records differ from classes
+## How records differ from classes and structs
 
-The same syntax that [declares](classes.md#declaring-classes) and [instantiates](classes.md#creating-objects) classes can be used with records. Just substitute the `record` keyword for the `class` keyword. Likewise, the same syntax for expressing inheritance relationships is supported by records. Records differ from classes in the following ways:
+The same syntax that [declares](classes.md#declaring-classes) and [instantiates](classes.md#creating-objects) classes or structs can be used with records. Just substitute the `record` keyword for the `class` keyword, or use `record struct` instead of `struct`. Likewise, the same syntax for expressing inheritance relationships is supported by record classes. Records differ from classes in the following ways:
```
gfoidl, line 34:
```suggestion
The same syntax that [declares](classes.md#declaring-classes) and [instantiates](classes.md#creating-objects) classes or structs can be used with records. Just substitute the `class` keyword with the `record`, or use `record struct` instead of `struct`. Likewise, the same syntax for expressing inheritance relationships is supported by record classes. Records differ from classes in the following ways:
```
For me it's easier to read that way (first what needs to be substituted, then the substitution).

docs/csharp/language-reference/builtin-types/snippets/shared/RecordType.cs
```diff
@@ -22,40 +22,71 @@ public static void Examples()
...
     }
 
     public static class ImmutableRecordType
     {
         // <ImmutableRecord>
         public record Person
         {
-            public string FirstName { get; init; }
-            public string LastName { get; init; }
+            public string FirstName { get; init; } = default!;
```
gfoidl, line 44: Just as question for the style: is `default!` prefered over `null!`?
If so it should be consistent accross the docs.

Some background: I'm moderator in (german language) C# forum, the audience contains quite some entry-level C# users. So I see them asking questions like `default` vs. `null`. 

Maybe
```suggestion
            public string FirstName { get; init; } = default!;    // similarly one could use = null! as it's the same for reference types
```
Is this too much noise?

docs/csharp/tour-of-csharp/index.md
```diff
@@ -80,7 +80,7 @@ The following outline provides an overview of C#'s type system.
   - [Delegate types](../language-reference/builtin-types/reference-types.md#the-delegate-type)
     - User-defined types of the form `delegate int D(...)`
 
-C# programs use *type declarations* to create new types. A type declaration specifies the name and the members of the new type. Six of C#'s categories of types are user-definable: class types, struct types, interface types, enum types, delegate types, and tuple value types.
+C# programs use *type declarations* to create new types. A type declaration specifies the name and the members of the new type. Six of C#'s categories of types are user-definable: class types, struct types, interface types, enum types, delegate types, and tuple value types. You can also declare `record` types, either `record struct`, or `record class`. Record types have additional synthesized members. You should use records primarily for storing values, with minimal associated behavior.
```
gfoidl, line 83: > either `record struct`, or `record class`.

For `record class` the `class` is optional. Should this be pointed out?
Or do we want to add the `class` to the `record` in order to make it more explicit and distinct from `record struct`?

docs/csharp/language-reference/builtin-types/record.md
```diff
@@ -97,27 +118,25 @@ To implement value equality, the compiler synthesizes the following methods:
...
+To implement this feature for `record class` types, the compiler synthesizes a clone method and a copy constructor. The virtual clone method returns a new record initialized by the copy constructor. When you use a `with` expression, the compiler creates code that calls the clone method and then sets the properties that are specified in the `with` expression.
 
-If you need different copying behavior, you can write your own copy constructor. If you do that, the compiler won't synthesize one. Make your constructor `private` if the record is `sealed`, otherwise make it `protected`.
+If you need different copying behavior, you can write your own copy constructor. If you do that, the compiler won't synthesize one. Make your constructor `private` if the record is `sealed`, otherwise make it `protected`. You can write a copy constructor for a `record struct`, but that typically isn't needed.
```
pkulikov, line 137: > You can write a copy constructor for a `record struct`

Is that true? How to do that then?

According to [this example in sharplab](https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEBMBGAsAKAPQGYACLUgdSgBcALAUQA8AHAJwFMBnLqAexgAhCLwDGzCGBYAbDgQDeBUsvJlOovmwSkuNNgFdRNUgAU+sGgAoLpABpxSNgJoBKANwElKrJi/LF+CpBpABuEGykLNikALykMBwA7qbmMFYADA7p7n7BYREsmLGR0Ym0dKTydsUArKQAvh6BKvUE9UA==) `record struct` doesn't have even the default copy constructor.

Since, structs don't allow `protected` members, than the custom copy constructor would be `private`. Something [like that](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK7wCYgNQB8ACATAIwCwAUBfgMwAERtA6lAgBYCiAHgA7ICmAZwFQA9jABCEYQGMuEMNwA2fCgG8KtTfTr9pI5BloCEaaQloAFEbAQAKG7QAacWg4CaASg1b15Lf9peKAA3CAQ+S2sYOysHbiiED1oQWjYoAVsABhdMrz8AzVVNAF9vEooy+kJCSt8CzVDkQOJaAF5aGD4Ad0ibW2IXYg8Abkr/RsDCNubaLpZWWiLHaYBWWmLR/IL8YgBOW25CEcrS8mKgA==):

```csharp
using System;

public class WithExpressionBasicExample
{
    public record struct Point(int X, int Y)
    {
        private Point(Point point) : this(0, 0)
        {  }
    }

    public static void Main()
    {
        var p1 = new Point(1, 1);
        var p2 = p1 with { X = 5 };
        Console.WriteLine(p2);
    }
}
```

But when I run it, I get `Point { X = 5, Y = 1 }`. So, the zeroing copy constructor is not called.

I have an impression, that a user cannot write a custom copy constructor for `record struct`

docs/csharp/language-reference/builtin-types/record.md
```diff
@@ -152,11 +173,11 @@ The following example illustrates inheritance with positional property syntax:
 
 ### Equality in inheritance hierarchies
 
-For two record variables to be equal, the run-time type must be equal. The types of the containing variables might be different. This is illustrated in the following code example:
+This section applies to `record` types, but not `record struct` types. For two record variables to be equal, the run-time type must be equal. The types of the containing variables might be different. This is illustrated in the following code example:
```
pkulikov, line 176:
```suggestion
This section applies to `record class` types, but not `record struct` types. For two record variables to be equal, the run-time type must be equal. The types of the containing variables might be different. This is illustrated in the following code example:
```

docs/csharp/language-reference/builtin-types/snippets/shared/RecordType.cs
BillWagner: Good conversation @gfoidl 

I haven't seen Microsoft adopt a coding standard for either `default` or `null` (except in generic classes where a variable of type `T` could be either a value type or a reference type). I know I've used both, but probably lean a bit toward `default`. 

> Is this too much noise?

I think so.

docs/csharp/tour-of-csharp/index.md
BillWagner: Also a great conversation. I'm going to leave this as is, with `record struct` and `record class`, but not resolve the conversation because I want other reviewers to provide their thoughts as well.

The features for `record struct` and `record class` are similar, and a single description works for both of them. I've used the word "record" to talk about features that apply to both. I've used `record class` and `record struct` where I mean one or the other. In the opening of the article, I did explain that `class` is optional: `record` and `record class` are the same.

docs/csharp/language-reference/builtin-types/record.md
BillWagner: great points @pkulikov 

You can write a copy constructor, but that won't be called for `with` expressions.  See here:

```
using System;

var pt = new Point(1,2);
var pt2 = new Point(pt);
Console.WriteLine(pt2);
var pt3 = pt2 with {X = 5 };
Console.WriteLine(pt3);

public record struct Point(int X, int Y)
{
    public Point(Point src) : this(src.X, src.Y) {}
}
```

Output:

```
Point { X = 1, Y = 2 }
Point { X = 5, Y = 2 }
```

CamSoper (approved):
Looks great! :shipit:

PR #26296, opened 2021-09-28
Update culture-creation-invariant-mode.md

SteveMCarroll:
i had to read this document and the link to explain what globalization invariant mode is actually the second mention which I found confusing when reading it.   thought it might make sense to move the link up for clarity, but hardly critical :)

## Summary

linked the first mention of globalization invariant mode to the explanatory page for clarity

### DIFF
--- a/docs/core/compatibility/globalization/6.0/culture-creation-invariant-mode.md
+++ b/docs/core/compatibility/globalization/6.0/culture-creation-invariant-mode.md
@@ -5,7 +5,7 @@ ms.date: 07/23/2021
 ---
 # Culture creation and case mapping in globalization-invariant mode
 
-This breaking change affects *[globalization-invariant mode](https://tools.ietf.org/rfc/bcp/bcp47.txt)* in two ways:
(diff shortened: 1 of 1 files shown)


docs/core/compatibility/globalization/6.0/culture-creation-invariant-mode.md
```diff
@@ -5,7 +5,7 @@ ms.date: 07/23/2021
 ---
 # Culture creation and case mapping in globalization-invariant mode
 
-This breaking change affects *globalization-invariant mode* in two ways:
+This breaking change affects *[globalization-invariant mode](https://tools.ietf.org/rfc/bcp/bcp47.txt)* in two ways:
```
gewarren, line 8: I don't think this is the correct link.

IEvangelist: The same link is used below, what's incorrect about it?

gewarren: It's a link to an RFC called "Tags for Identifying Languages". I think something like https://github.com/dotnet/runtime/blob/main/docs/design/features/globalization-invariant-mode.md would make a lot more sense.

IEvangelist (approved):

PR #26192, opened 2021-09-19
Implement dispose method fixes

deep-outcome:
### In commit order
 
1. Term “_**Non**Inheritable_” contains typo. Correct is “[_**Not**Inheritable_](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/modifiers/notinheritable)”.

More over this is some mistake. Inheritance is class/object concept. Instead [**NotOverridable**](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/modifiers/notoverridable) belongs there. 

2. Clarification on object resources management.
3. Place of non-deterministic release has no relevance to managed resources conditional block. Further more this can never happen non-deterministically in managed resources conditional block while `bool disposing` parameter is `false`.
4. Introduction of `Dispose()`, `Dispose(bool)` was somehow strange.
5. `Foo` was simplified to spare some lines + `?.` operator made no sense there – after construction `_bar` cannot be `null`.
6. Clarifications on _Implement the dispose pattern_.
7. Clarifications _Implement the dispose pattern for a derived class_ + again `?.` operator removal.
8. Chapters _Implement the dispose pattern with safe handles_ and _Implement the dispose pattern for a derived class with safe handles_ removed since both states nothing unsaid, see. _Implement the dispose pattern|1st example_ and _Implement the dispose pattern for a derived class|1st example_, about _dispose pattern_. Moreover example show excessive information about using safe handles, extern API, … If these should be kept, I would introduce them into some _Additional Examples_ chapter (or to some place more related to `SafeHandle`s).

PR #26302, opened 2021-09-29
Address review feedback and correct deviation from analyzer dispose pattern impls.

IEvangelist:
## Summary

- Corrected issues introduced from #26192, formatting, deviation from analyzer provided implementation of Dispose, code style, etc.
- Add enable implicit usings in new LINQ bits
- Follow up from review feedback that wasn't incorporated: https://github.com/dotnet/docs/pull/26192#pullrequestreview-761065050
- The example for `Foo` needs to remain as such as well, since it aligns with the Framework guidance on implementing the dispose pattern.


gewarren (approved):
LGTM - thanks for following up.

IEvangelist:
> @IEvangelist Should the VB snippets be updated to match the C# snippets?

No, we've deemphasized VB efforts as there are higher priorities.

PR #26309, opened 2021-09-29
Added the section title

pkulikov:
This is to balance the "Parameterless constructor of a value type" section and separate the table with the default values from other article text.


BillWagner (approved):
Thanks @pkulikov 

PR #26310, opened 2021-09-29
Closing updates for the parameterless struct constructors feature

pkulikov:
Contributes to  #25676

- Searched and checked only the "field initializer" and "field initialization" entries in the C# doc set.
- Checked the articles about arrays in the C# guide on the way how they mention array initialization with the default values.


BillWagner (approved):
This LGTM @pkulikov 

I'll :shipit: now.

PR #26358, opened 2021-10-03
Update interfaces.md

kcratss:
Default implementation is not accessible as an inherited member in the class neither its available for override by default.

## Summary

Describe your changes here.


BillWagner (commented):
Thanks @kcratss This is a great change.

I had a few small suggestions. Once we resolve those, this ready to merge.

kcratss:
Thanks @BillWagner for the suggestions, I have updated the PR as per the suggestions.

BillWagner (approved):
Thanks @kcratss 

I've reviewed the changes, and I'll :shipit: now

docs/csharp/fundamentals/types/interfaces.md
```diff
@@ -36,7 +36,7 @@ The following example shows an implementation of the <xref:System.IEquatable%601
...
...d interface inherits the members from its base interfaces. A class that implements a derived interface must implement all members in the derived interface, including all members of the derived interface's base interfaces. That class may be implicitly converted to the derived interface or any of its base interfaces. A class might include an interface multiple times through base classes that it inherits or through interfaces that other interfaces inherit. However, the class can provide an implementation of an interface only one time and only if the class declares the interface as part of the definition of the class (`class ClassName : InterfaceName`). If the interface is inherited because you inherited a base class that implements the interface, the base class provides the implementation of the members of the interface. However, the derived class can reimplement any virtual interface members instead of using the inherited implementation. When interfaces declare a default implementation of a method, any class implementing that interface inherits that implementation (You need to cast the class instance to the interface type to access the default implementation on the Interface member).
```
BillWagner, line 39:
```suggestion
Interfaces can inherit from one or more interfaces. The derived interface inherits the members from its base interfaces. A class that implements a derived interface must implement all members in the derived interface, including all members of the derived interface's base interfaces. That class may be implicitly converted to the derived interface or any of its base interfaces. A class might include an interface multiple times through base classes that it inherits or through interfaces that other interfaces inherit. However, the class can provide an implementation of an interface only one time and only if the class declares the interface as part of the definition of the class (`class ClassName : InterfaceName`). If the interface is inherited because you inherited a base class that implements the interface, the base class provides the implementation of the members of the interface. However, the derived class can reimplement any virtual interface members instead of using the inherited implementation. When interfaces declare a default implementation of a method, any class implementing that interface inherits that implementation (You need to cast the class instance to the interface type to access the default implementation on the Interface member).
```

PR #26379, opened 2021-10-04
Add Documentation for package Validation

Anipik:
## Summary

We added a new package validation feature in the sdk during net6.0 and this change is adding the documentation for it.


docs/compatibility/PackageValidation.md
```diff
@@ -0,0 +1,243 @@
...
+## Enabling Package Validation
+
+Package Validation can be enabled in the .NET project by setting `EnablePackageValidation`. 
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <EnablePackageValidation>true</EnablePackageValidation>
```
gewarren, line 27: We should probably add this property to the msbuild-props.md reference file as well.

docs/compatibility/PackageValidation.md
```diff
@@ -0,0 +1,243 @@
...
+  </PropertyGroup>
+
+</Project>
+```
+
+`EnablePackageValidation` runs a series of checks after the pack task. There are some additional checks which could be run by setting other msbuild properties.
```
gewarren, line 33:
```suggestion
`EnablePackageValidation` runs a series of checks after the pack task. There are some additional checks that can be run by setting other MSBuild properties.
```

gewarren (commented):
I will keep reviewing more later but there are some changes you can make meanwhile:

- File names should be all lower case, and words should be separated with hyphens (e.g. package-validation.md).
- You'll need to add the Markdown files to the appropriate TOC.yml file (the table of contents). Probably this one: https://github.com/dotnet/docs/blob/main/docs/fundamentals/toc.yml

docs/compatibility/PackageValidation.md
```diff
@@ -0,0 +1,243 @@
...
+
+In both of these cases, the consumers can build against .NET Standard 2.0 or NET 5.0 and run on .NET 6.0. In case your binaries are not compatible between these frameworks, consumers could end up with compile and/or runtime errors.
+
+Package Validation will catch these errors at pack time. Here is an example scenario:
+
+Suppose you're writing a game which does a lot of string manipulation. You need to support both .NET Framework and .NET Core consumers. You started with just targeting .NET Standard 2.0 but now you realize you want to take advantage of spans in .NET 6.0 to avoid unnecessary string allocations. In order to do that, you now want to multi-target for .NET Standard 2.0 and .NET 6.0.
+
+You have written the following code:
+
+```c#
```
Youssef1313, line 51:
```suggestion
```csharp
```

docs/compatibility/PackageValidation.md
```diff
@@ -0,0 +1,243 @@
...
+ms.date: 09/29/2021
+---
+
+## Importance of Package Validation
+
+With .NET Core & Xamarin we have made cross-platform a mainstream requirement for library authors. However, we lack validation tooling for cross targeting packages, which can result in packages that don't work well, which in turn hurts our ecosystem. This is especially problematic for emerging platforms where adoption isn't high enough to warrant special attention by library authors.
+
+The tooling we provide as part of the SDK has close to zero validation that multi-targeted packages are well-formed. For example, a package that multi-targets for .NET 6.0 and .NET Standard 2.0 needs to ensure that code compiled against the .NET Standard 2.0 binary can run against the .NET 6.0 binary. We have seen this issue in the wild, even with 1st parties, for example, the Azure AD libraries.
+
+It's easy to think that a change is safe and compatible if source consuming that change continues to compile without changes. However, certain changes may work fine in C# but can cause problems at runtime if the consumer wasn't recompiled, for example, adding a defaulted parameter or changing the value of a constant.
```
safern, line 13:
```suggestion
It's easy to think that a change is safe and compatible if source consuming that change continues to compile without changes. However, certain changes may work fine in C# but can cause problems at runtime if the consumer wasn't recompiled, for example, adding an optional parameter or changing the value of a constant.
```

docs/compatibility/PackageValidation.md
```diff
@@ -0,0 +1,243 @@
...
+    <EnablePackageValidation>true</EnablePackageValidation>
+  </PropertyGroup>
+
+</Project>
+```
+
+`EnablePackageValidation` runs a series of checks after the pack task. There are some additional checks which could be run by setting other msbuild properties.
+
+
+## Validation Compatible Frameworks
```
safern, line 36:
```suggestion
## Validate Compatible Frameworks
```

docs/compatibility/PackageValidation.md
```diff
@@ -0,0 +1,243 @@
...
+        // Do some string operations.
+    }
+#endif
+```
+
+You then try to pack the project (using dotnet pack cmd or using VS) it fails with the following error:
+
+![CompatibleFrameworks](CompatibleFrameworks.png)
+
+You understand that you shouldn't exclude ```DoStringManipulation(string)``` but instead just provide an additional ```DoStringManipulation(ReadOnlySpan<char>)``` method for .NET 6.0 and changes the code accordingly:
```
safern, line 69:
```suggestion
You understand that you shouldn't exclude ```DoStringManipulation(string)``` but instead just provide an additional ```DoStringManipulation(ReadOnlySpan<char>)``` method for .NET 6.0 and change the code accordingly:
```

docs/compatibility/PackageValidation.md
```diff
@@ -0,0 +1,243 @@
...
+![BaselineVersion](BaselineVersion.png)
+
+You realize that while this is not a [source breaking change](https://docs.microsoft.com/dotnet/standard/library-guidance/breaking-changes#source-breaking-change), it's a binary [breaking change](https://docs.microsoft.com/dotnet/standard/library-guidance/breaking-changes#binary-breaking-change). You solve this problem by adding an overload instead:
+
+```c#
+public static HttpClient Connect(string url)
+{
+    return Connect(url, Timeout.InfiniteTimeSpan);
+}
+public static HttpClient Connect(string url, TimeSpan timeout)
```
safern, line 138:
```suggestion

public static HttpClient Connect(string url, TimeSpan timeout)
```

docs/compatibility/PackageValidation.md
```diff
@@ -0,0 +1,243 @@
...
+runtimes/unix/lib/net6.0/A.dll
+```
+
+`lib\net6.0\A.dll` will always be used at compile time regardless of the underlying operating system. `lib\net6.0\A.dll` will also be used at runtime for non-Unix systems, but `runtimes\unix\lib\net6.0\A.dll` will be used at runtime for Unix systems.
+
+When you try to pack this project, you get an error:
+
+![MultipleRuntimes](MultipleRuntimes.png)
+
+you quickly realize your mistake and adds `A.B.Open(string)` to the unix runtime as well.
```
safern, line 181:
```suggestion
you quickly realize your mistake and add `A.B.Open(string)` to the unix runtime as well.
```

docs/compatibility/PackageValidation.md
```diff
@@ -0,0 +1,243 @@
...
+```c#
+#if Unix
+    public static void Open(string path, bool securityDescriptor)
+    {
+        // call unix specific stuff
+    }
+    
+    public static void Open(string path)
+    {
+        // throw not supported exception
```
safern, line 192:
```suggestion
        throw new PlatformNotSupportedException();
```

docs/compatibility/PackageValidation.md
```diff
@@ -0,0 +1,243 @@
...
+
+Left = Left operand of an APICompat comparison.
+
+Right = Right operand of an APICompat comparison.
+
+isBaseline =  true <see langword="true"/> if the suppression is to be applied to a baseline validation. false <see langword="false"/> otherwise.
+
+## Extra features
+
+1) Strict mode:- By default, Package validation runs the api-compat in non-strict mode. Enabling strict mode will change some rules and some other rules will be executed when getting the differences. This is useful when you want both sides we are comparing to be strictly the same on their surface area and identity. This could be enabled for different validators by setting
```
safern, line 237:
```suggestion
1) Strict mode:- By default, Package validation runs api-compat in non-strict mode. Enabling strict mode will change some rules and some other rules will be executed when getting the differences. This is useful when you want both sides we are comparing to be strictly the same on their surface area and identity. This could be enabled for different validators by setting 
```

docs/compatibility/PackageValidation.md
```diff
@@ -0,0 +1,243 @@
...
+
+isBaseline =  true <see langword="true"/> if the suppression is to be applied to a baseline validation. false <see langword="false"/> otherwise.
+
+## Extra features
+
+1) Strict mode:- By default, Package validation runs the api-compat in non-strict mode. Enabling strict mode will change some rules and some other rules will be executed when getting the differences. This is useful when you want both sides we are comparing to be strictly the same on their surface area and identity. This could be enabled for different validators by setting 
+
+`EnableStrictModeForCompatibleTfms` & `EnableStrictModeForCompatibleFrameworksInPackage` respectively.
+
+2) Running using references:- Some of the api compatibility rules require references eg resolving typeforwards, enums. The references will be passed automatically for some all validation among the assemblies in the package.In order to pass the references across versions, user will to manually provide them by using `PackageValidationReferencePath`.
```
safern, line 241: Maybe we should provide an example of the items that would be added to this itemgroup? As it has to be a map of tfm + the search directory?

gewarren (commented):
I also think the files should move to the docs/fundamentals/compatibility folder. And from there, can you put the media files in a "media" subfolder? You'll need to update the image links in the articles too.

gewarren (commented):
Getting really close! I still think the images should be moved into a "media" subfolder too.

gewarren (approved):

safern (approved):
LGTM! Thanks @anipik. Great that we are having documentation soon!

ericstj:
FYI @zkat @mhutch Let us know if you have feedback and we can address in follow up.

PR #26405, opened 2021-10-06
C# reference consolidation: lambdas in queries

pkulikov:
Contributes to #19743


BillWagner (approved):
Thanks @pkulikov 

PR #26412, opened 2021-10-06
Update docs for Windows 11

adegeo:
## Summary

Just did a basic pass on mentioning Windows 11 in places where Windows 10 was explicitly called out.

- Updates a lot of the .NET Framework related install articles to mention Windows 11 where it mentioned Windows 10
- Updated the .NET install docs for Windows to add Windows 11

Ping @jamshedd @gewarren 

@jamshedd Does Windows Server 2022 come with 4.8? I made the assumption it does. I didn't, though, create a page specifically for it, if it's required, I might. The only reason I created a page for Windows 11 and .NET Framework is that it fits logically in the TOC. If people see the TOC and see Windows 8 and 10 called out, they'll ask where is Windows 11.


gewarren (approved):
Thanks for updating all of these!

PR #26413, opened 2021-10-06
Add IL2061-IL2116 trimmer warnings

mateoatr:
Moving the second half of the trimmer warnings in https://github.com/mono/linker/blob/main/docs/error-codes.md. Note that warnings which we have listed on the trimmer docs but that are not currently produced by the trimmer were not added in this list.


gewarren (commented):
I left a bunch of suggestions. Some of them should be applied throughout (if my suggestions are correct).

gewarren:
Also if we are going to move the trimmer docs to a subfolder as #26445 does, then we shouldn't merge this PR until that one is merged.

mateoatr:
> Also if we are going to move the trimmer docs to a subfolder as #26445 does, then we shouldn't merge this PR until that one is merged.

Thanks for suggestions! I'll update the PR applying them in all docs. We can wait for #26445 to be merged before this -- I wasn't aware of that PR.

gewarren:
> Thanks for suggestions! I'll update the PR applying them in all docs. We can wait for #26445 to be merged before this -- I wasn't aware of that PR.

Andy's PR is merged, so you can go ahead and put these new files in the new trimming subfolder.

gewarren (approved):

gewarren:
MSDoc build verifier failure seems unrelated.

PR #26431, opened 2021-10-08
Corrected Linux RID value for "dot net publish" command

MopuriMani:
See, https://docs.microsoft.com/en-us/dotnet/core/rid-catalog#linux-rids for valid values of RID for Linux.
Also removed the shell prompt string to make it easy to copy and paste.


gewarren:
Ping for review @luisquintanilla.

adegeo:
@luisquintanilla  reping

PR #26445, opened 2021-10-09
Organize trimming docs and add incompatibility page

agocke:
Moves trimming docs into a subdirectory.


gewarren (commented):
I left some suggestions. Also, if we move any existing file, we need to add a redirect to .openpublishing.redirection.json. And unfortunately it can't be done at the folder level, so we'll need one entry for each file.

gewarren (approved):

PR #25621, opened 2021-08-12
Long pollling fallback

gewarren:
Contributes to aspnet/announcements#470.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/6.0/blazor-long-polling-fallback?branch=pr-en-us-25621).

PR #26504, opened 2021-10-13
Remove long polling breaking change

gewarren:
Follow up to #25621. The breaking change was reverted in RC 2.


TanayParikh (approved):
Thanks!

PR #26505, opened 2021-10-14
File.Replace exceptions on Unix

### LINKED ISSUE #26323: [Breaking Change] `File.Replace` throws exceptions on more cases on Unix
### Description

In https://github.com/dotnet/runtime/pull/50234 we changed the behavior of `File.Replace` on Unix so that the exceptions it throws now match those that are thrown by the Windows implementation.

### Previous behavior

On Unix, with .NET 5, the `File.Replace` method:

- Throws `IOException` with the message `Is a directory`  when `sourceFileName` is a file and `destinationFileName` is a directory.
- Throws `IOException` with the message `Not a directory`  when `sourceFileName` is a directory and `destinationFileName` is a file.
- Silently succeeds when both `sourceFileName` and `destinationFileName` point to the same file or directory.

### New behavior

On Unix, with .NET 6, the `File.Replace` method now:

- Throws `UnauthorizedAccessException` with the message `The specified path <path> is not a path`, when either `sourceFileName` or `destinationFileName` exists and is not a file, or when both `sourceFileName` and `destinationFileName` point to the same existing directory.
- Throws `IOException` with the message `The source <sourceFileName> and destination <destinationFileName> are the same file` when `sourceFileName` and `destinationFileName` point to the same existing file.

### Type of breaking change

**Question**: I would classify this as a "different run-time behavior", but that case is mentioned in both options below. Which one do I choose for exception difference behaviors?:

### Reason for change

Ensuring we throw the same exceptions for the same reasons across platforms in `File.Replace`.

### Recommended action

If your `File.Replace` invocation on Unix is inside a try catch, make sure to now also catch `UnauthorizedAccessException`, and be aware of the new behaviors that are caught.

### Feature area

Core .NET Libraries

### Affected APIs

`File.Replace` https://docs.microsoft.com/en-us/dotnet/api/system.io.file.replace?view=net-6.0

gewarren:
Fixes #26323 

[Preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/file-replace-exceptions-on-unix?branch=pr-en-us-26505)

### DIFF
--- a/docs/core/compatibility/6.0.md
+++ b/docs/core/compatibility/6.0.md
@@ -45,6 +45,7 @@ If you're migrating an app to .NET 6, the breaking changes listed here might aff
 | [API obsoletions with non-default diagnostic IDs](core-libraries/6.0/obsolete-apis-with-custom-diagnostics.md) | ✔️ | ❌ | Preview 1 |
 | [Changes to nullable reference type annotations](core-libraries/6.0/nullable-ref-type-annotation-changes.md) | ✔️ | ❌ | Preview 1-2 |
 | [Environment.ProcessorCount behavior on Windows](core-libraries/6.0/environment-processorcount-on-windows.md) | ✔️ | ❌ | Preview 2 |
+| [File.Replace on Unix throws exceptions to match Windows](core-libraries/6.0/file-replace-exceptions-on-unix.md) | ✔️ | ❌ | Preview 7 |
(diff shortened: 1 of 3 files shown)


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when you're ready.

PR #26512, opened 2021-10-14
StringBuilder.Append overloads breaking change

### LINKED ISSUE #26428: [Breaking change]: StringBuilder.Append{Line} evaluation order
### Description

C# 10 adds support for better string interpolation, including the ability to target custom "handlers" in addition to strings (see https://devblogs.microsoft.com/dotnet/string-interpolation-in-c-10-and-net-6/ for more details).  `StringBuilder` takes advantage of this with new overloads of `Append` and `AppendLine` that accept a custom interpolated string handler.  Existing calls to these methods may now start binding to the new overloads.  In general, behavior is identical but with improved performance, e.g. rather than a string first being created and then that string appended, the individual components of the interpolated string are appended directly to the builder.  However, in doing so this can change the evaluation order of objects used as format items, which can in turn manifest in a difference in behavior.

### Version

.NET 6 RC 1

### Previous behavior

A call to:
```C#
stringBuilder.Append($"{a} {b}");
```
will compile to the equivalent of:
```C#
stringBuilder.Append(string.Format("{0} {1}", a, b));
```
which means `a` is evaluated, then `b` is evaluated, then a string is created from the results of those evaluations, and then that string is appended to the builder.

### New behavior

A call to:
```C#
stringBuilder.Append($"{a} {b}");
```
will compile to the equivalent of:
```C#
var handler = new StringBuilder.AppendInterpolatedStringHandler(1, 2, stringBuilder);
handler.AppendFormatted(a);
handler.AppendLiteral(" ");
handler.AppendFormatted(b);
stringBuilder.Append(ref handler);
```
which means `a` is evaluated and appended to the builder and then `b` is evaluated and appended to the builder.

If, for example, either `a` or `b` is itself the builder, e.g.
```C#
stringBuilder.Append($"{a} {stringBuilder}");
```
the alternate method of compilation can result in different behavior at run-time.

### Type of breaking change

### Reason for change

It is extremely common for developers to pass interpolated strings to `StringBuilder`, as it's more convenient than manually splitting the string into its constiuent components and calling `StringBuilder.Append` for each.  These new overloads enable the concise syntax and most of the performance of doing the individual calls.

### Recommended action

The vast majority of StringBuilder.Append and StringBuilder.AppendLine use will not notice a functional difference.  If a difference is found and proves to be problematic, simply adding a cast to `(string)` prior to the interpolated string will restore the previous behavior, e.g. `stringBuilder.Append((string)$"{a} {b}")`.  This is not recommended, however, unless it's actually required for compatibility.

### Feature area

Core .NET libraries

### Affected APIs

StringBuilder.Append
StringBuilder.AppendLine

gewarren:
Fixes #26428 

[Preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/stringbuilder-append-evaluation-order?branch=pr-en-us-26512)

### DIFF
--- a/docs/core/compatibility/6.0.md
+++ b/docs/core/compatibility/6.0.md
@@ -54,6 +54,7 @@ If you're migrating an app to .NET 6, the breaking changes listed here might aff
 | [Some parameters in Stream-derived types are renamed](core-libraries/6.0/parameters-renamed-on-stream-derived-types.md) | ✔️ | ❌ | Preview 1 |
 | [Standard numeric format parsing precision](core-libraries/6.0/numeric-format-parsing-handles-higher-precision.md) | ✔️ | ❌ | Preview 2 |
 | [Static abstract members in interfaces](core-libraries/6.0/static-abstract-interface-methods.md) | ❌ | ✔️ | Preview 7 |
+| [StringBuilder.Append overloads and evaluation order](core-libraries/6.0/stringbuilder-append-evaluation-order.md) | ❌ | ✔️ | RC 1 |
 | [System.Drawing.Common only supported on Windows](core-libraries/6.0/system-drawing-common-windows-only.md) | ❌ | ❌ | Preview 7 |
 | [System.Security.SecurityContext is marked obsolete](core-libraries/6.0/securitycontext-obsolete.md) | ✔️ | ❌ | RC 1 |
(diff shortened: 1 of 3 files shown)


stephentoub (approved):
Thanks!

PR #26526, opened 2021-10-15
Updates Sample of cs0675

bschaeublin:
Hi there!

This updated sample fixes #26409.


BillWagner (approved):
Thanks for submitting this fix @bschaeublin  We appreciate it! 

I've reviewed this change, and I'll :shipit: now. You should see the changes on the live site in the next day or two on our regular publishing cycle.

Thanks again!

PR #26539, opened 2021-10-15
FileStream locks on Unix

### LINKED ISSUE #26529: [Breaking change]: On Linux, opening a file without `FileAccess.Write` then locking it, will now lock it with a shared lock
### Description

On Unix, when the user opens a file using `FileStream`, only with a read-only `FileAccess` permission (no `FileAccess.Write`), and then attempts to lock a region of the file using `FileStream.Lock`, the operation should succeed because the file will be locked with a shared or read lock.

```cs
using (FileStream fs = File.OpenRead("testfile")) // Opening with FileAccess.Read only
{
    fs.Lock((long) 3, (long) 1); // Attempting to lock a region of the read-only file
}
```

### Version

.NET 6 RC 1

### Previous behavior

On Unix, when the user opened a file with `FileStream` with read permissions only, and then attempted to lock a region of the file using `FileStream.Lock`, we would attempt to lock the file with a write-lock, which would throw `System.UnauthorizedException: Access to the path is denied.` .
No exception thrown on Windows (code behaved well).

### New behavior

Now, when the user opens a file using `FileStream` with read permissions only on Unix, and then attempts to lock a region of the file using `FileStream.Lock`, we lock the file with a read lock (also known as shared lock). No behavior change on Windows.

### Type of breaking change

### Reason for change

`FileStream.Lock(long position, long length)` is the API we offer to users that want to lock a specific region of a file. We do not offer an API that lets the user choose the underlying locking method, so the user expects this API to correctly determine what is the appropriate locking method depending on the file permissions. The user does not have an easy workaround for this.

### Recommended action

Before .NET 6.0, in order to be able to lock the file, the user would have to either check if the code was being executed in Windows or the `FileStream` was opened with `FileAccess.Write` permission, or wrap the `FileStream.Lock` call with a try catch to capture the `System.UnauthorizedException`.

On .NET 6.0 and above, such workarounds will not be necessary.

### Feature area

Core .NET libraries

### Affected APIs

[System.IO.FileStream.Lock(System.Int64,System.Int64)](https://docs.microsoft.com/en-us/dotnet/api/system.io.filestream.lock?view=net-6.0) (Unix only).

gewarren:
Fixes #26529 

[Preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/filestream-file-locks-unix?branch=pr-en-us-26539)


docs/core/compatibility/core-libraries/6.0/filestream-file-locks-unix.md
```diff
@@ -0,0 +1,50 @@
+---
+title: "Breaking change: FileStream locks files with shared lock on Unix"
+description: Learn about the .NET 6 breaking change where FileStream.Lock now locks files with a read lock when they're opened with read permissions on Unix.
```
carlossanlop, line 3: Should this `FileStream.Lock` mentioned in the description be wrapped by an xref or cref?

docs/core/compatibility/core-libraries/6.0/filestream-file-locks-unix.md
```diff
@@ -0,0 +1,50 @@
...
+{
+    fs.Lock((long) 3, (long) 1); // Attempting to lock a region of the read-only file
+}
+```
+
+There's no change to the behavior on Windows, where the operation always succeeded.
+
+## Previous behavior
+
+On Unix, if you opened a file with <xref:System.IO.FileStream> with read permissions only, and then called <xref:System.IO.FileStream.Lock(System.Int64,System.Int64)?displayProperty=nameWithType> to lock a region of the file, the runtime attempted to lock the file with a write lock. That resulted in an <xref:System.UnauthorizedAccessException> and the message "Access to the path is denied".
```
carlossanlop, line 21: The two consecutive `with` words tripped me a bit. The description in the next section is easier to read:
 
```suggestion
On Unix, if you opened a file using a <xref:System.IO.FileStream> with read permissions only, and then called <xref:System.IO.FileStream.Lock(System.Int64,System.Int64)?displayProperty=nameWithType> to lock a region of the file, the runtime attempted to lock the file with a write lock. That resulted in an <xref:System.UnauthorizedAccessException> and the message "Access to the path is denied".
```

docs/core/compatibility/core-libraries/6.0/filestream-file-locks-unix.md
```diff
@@ -0,0 +1,50 @@
...
+
+## New behavior
+
+Starting in .NET 6, if you open a file using <xref:System.IO.FileStream> with read permissions only on Unix, and then call <xref:System.IO.FileStream.Lock(System.Int64,System.Int64)?displayProperty=nameWithType> to lock a region of the file, the runtime locks the file with a read lock (also known as a *shared lock*).
```
carlossanlop, line 25:
```suggestion
Starting in .NET 6, if you open a file using a <xref:System.IO.FileStream> with read permissions only on Unix, and then call <xref:System.IO.FileStream.Lock(System.Int64,System.Int64)?displayProperty=nameWithType> to lock a region of the file, the runtime locks the file with a read lock (also known as a *shared lock*).
```

carlossanlop (approved):
Some minor suggestions.

docs/core/compatibility/core-libraries/6.0/filestream-file-locks-unix.md
gewarren: No, it's just metadata so I don't think such links would resolve.

PR #26540, opened 2021-10-16
Incomplete info in `linq` equals operator section

### LINKED ISSUE #24528: Incomplete / misleading info in LINQ `equals` operator documentation
Issue
--

At https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/join-clause#the-equals-operator it states (emphasis mine):

>  The `equals` keyword can only be used in a `join` clause and it differs from the `==` operator in ***one important way***. With `equals`, the left key consumes the outer source sequence, and the right key consumes the inner source. The outer source is only in scope on the left side of `equals` and the inner source sequence is only in scope on the right side.

That is, it implies that there is only a single important difference between `==` and `equals`, and identifies that single difference as the scoping behavior.

However, that is *not* the only "important" (imo) difference. In particular, the following two differences are undocumented but very significant:

- `equals` behaves more like `Object.Equals` -- that is, it checks reference types for value equivalence.
- `null equals null` appears to evaluate to *false*, or at least has that effect on the `join` results (note that `Object.Equals(null, null)` is *true*. (This, specifically, was a recent source of confusion for me and is what prompted this feedback.)

I feel that the documentation for `equals` should be expanded to include a more complete description of the specific behavior of that keyword. Also, at the very least, the wording should probably be changed so that it doesn't appear to read as the scope rules being the single important difference.

Example
--

Maybe not helpful, but fwiw:

		object[] dataX = { string.Intern("a"), null, new string("c"), 4 };
		object[] dataY = { string.Intern("a"), null, new string("c"), 4 };

		var query1 = from x in dataX join y in dataY on x equals y select (x, y);
		var query2 = from x in dataX from y in dataY where x == y select (x, y);
		var query3 = from x in dataX from y in dataY where Equals(x, y) select (x, y);

		Console.WriteLine($"query1 returned {string.Join(" / ", query1),-32} [equals]");
		Console.WriteLine($"query2 returned {string.Join(" / ", query2),-32} [==]");
		Console.WriteLine($"query3 returned {string.Join(" / ", query3),-32} [Object.Equals]");

The output demonstrates that `equals` has its own unique currently-undocumented behavior, and that the differences between `equals` and `==` extend beyond scope rules:

```
query1 returned (a, a) / (c, c) / (4, 4)         [equals]
query2 returned (a, a) / (, )                    [==]
query3 returned (a, a) / (, ) / (c, c) / (4, 4)  [Object.Equals]
```

Runnable example [here](https://dotnetfiddle.net/0QIxrb).

---
#### Document Details

* ID: 7f64bda9-3763-e0a0-fbb9-22f6fd50beca
* Version Independent ID: d552325f-7578-efb3-3f6d-927169181391
* Content: [join clause - C# Reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/join-clause)
* Content Source: [docs/csharp/language-reference/keywords/join-clause.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/keywords/join-clause.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

luizhlelis:
## Summary

Fix: incomplete info in `linq` equals operator section

Fixes #24528


BillWagner (approved):
Thanks for making these changes @luizhlelis We appreciate it.

Thanks again.

PR #26543, opened 2021-10-16
In `invalidArg` example is better to use `nameof`

ghost:
Exactly [this example](https://docs.microsoft.com/en-us/dotnet/fsharp/whats-new/fsharp-50#support-for-nameof) was used in release notes for F# 5.0, but unfortunately was not updated in docs.


BillWagner (approved):
Thanks for making this update @dulanov 

I've reviewed the changes, and I'll :shipit: nnow.

PR #26553, opened 2021-10-18
Fix typo in rid-catalog.md

akoeplinger:
The RID version is 11.0 but the text mentioned 11.01


BillWagner (approved):
Thanks @akoeplinger 

PR #26447, opened 2021-10-10
Make `private` article follow ms coding-styles

bschaeublin:
## Summary
Hello everyone 😃 

In #25767 a user states, that the article about private variables does not follow the[ C# Coding Conventions](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#camel-case)

> - Use camel casing ("camelCasing") when naming private or internal fields, and prefix them with _.
> - When working with static fields that are private or internal, use the s_ prefix and for thread static use t_.

Fixes #25767

There are other samples with the same issue, but i did not touch them, as they sometimes do not follow the conventions on purpose.

If i forgot something, let me know.

Issue #25767, opened 2021-08-23
This page doesn't follow the naming convention for c#


PhotoAtomic:
This page doesn't folllow the convention stated here
https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#camel-case

differently from what is reported at the top of the document
>The guidelines in this article are used by Microsoft to develop samples and documentation

due to a later addition to the guidelines occurred here https://github.com/dotnet/docs/commit/631c596d095e9c0815c838a5b8b8ec79c122dd5b

see https://github.com/dotnet/docs/issues/25754 as reference

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 808a5ca6-e3c5-bed2-6bc7-8e0dbf965870
* Version Independent ID: 9a6c2aa5-f5cc-5caf-6592-3048e377440b
* Content: [private keyword - C# Reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/private)
* Content Source: [docs/csharp/language-reference/keywords/private.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/keywords/private.md)
* Product: **dotnet-csharp**
* Technology: **csharp-language-reference**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

IEvangelist:
Hi @PhotoAtomic,
I've marked this issue as `up-for-grabs`, and we'd happily review a pull request to update these samples.
Thank you

PhotoAtomic:
Sorry, after the signaling I've already done my own copy of the coding convention, not interested anymore in the consistency of this doc.

Scarica Outlook per Android<https://aka.ms/AAb9ysg>
________________________________

PR #26610, opened 2021-10-21
Update code style for private fields (#25767)

poke:
Hey, this is a follow-up to PR #26447 for issue #25767 which made it in before I had the chance to open a PR. The merged PR was very defensive in regard to what was changed in order to match the style guide. While going through the examples myself, I did identify a few more examples that I felt should be updated. This PR contains my changes, rebased on top of the already merged PR.

Since a lot of the docs are focused on the language options (instead of the practical guidelines), I do agree that not all examples should blindly follow the guidelines, especially since some rules would make it impossible to show off some syntax features (e.g. public fields).

---

Update example snippets to follow the code style for private and static
fields, and use an underscore or a `s_` prefix.

This commit only changes the styling in those examples where the field
is not the primary focus of the documentation page. This avoids
over-focusing on the styling in favor of keeping the examples succinct,
and also helps not to distract from the actual topic. Otherwise, many
examples, like public fields, couldn’t be explained because the style
guide discourages them.

- Use underscrore prefix for private fields.
- Always specify visibility modifiers.
- Use `s_` prefix for static fields.
- Avoid public field in `override` docs and use a public property
  `Employee.Name` instead.
- Mention abstract properties in `abstract` docs and move explanatory
  comments above lines.

### DIFF
--- a/docs/csharp/language-reference/keywords/snippets/ReadonlyKeywordExamples.cs
+++ b/docs/csharp/language-reference/keywords/snippets/ReadonlyKeywordExamples.cs
@@ -7,14 +7,14 @@ namespace keywords
     // <SnippetReadonlyField>
     class Age
     {
-        readonly int year;
+        private readonly int _year;
         Age(int year)
         {
-            this.year = year;
+            _year = year;
         }
         void ChangeYear()
         {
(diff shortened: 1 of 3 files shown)


IEvangelist (approved):

PR #26612, opened 2021-10-22
Clarify try-catch usage (#26560)

poke:
This addresses #26560 which suggests that the `catch (Exception)` examples on the try-catch docs page shouldn’t be left empty.

I did want to change the usage of the base `Exception` type to use a more specific type but ultimately didn’t because those `Exception` handlers all were logging out the exceptions (which is usually a fine thing to do when done at a central application entry point, e.g. `Main`) and because the examples actually throw example `Exception` types to demonstrate the behavior. At least in the case of those examples, the code was complete enough to show that there were no other unknown exceptions being thrown that could have been incorrectly caught here.

Instead, I have added some clarification on what exception type should be chosen when using a `catch` statement, to prefer specific exceptions over the base type if possible.

### DIFF
--- a/docs/csharp/language-reference/keywords/try-catch.md
+++ b/docs/csharp/language-reference/keywords/try-catch.md
@@ -28,11 +28,12 @@ try
 }
 ```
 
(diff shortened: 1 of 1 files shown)


IEvangelist (approved):

PR #26615, opened 2021-10-22
Update pre-processor documentation to match spec

terrajobst:
The spec is here: https://github.com/dotnet/designs/blob/main/accepted/2020/or-greater-defines/or-greater-defines.md#doc-changes


ericsampson:
Somehow the updated docs should denote that these new symbols are only available in .NET 5 SDK (or greater ;)

ericsampson (commented):
Somehow the updated docs should denote that these new symbols are only available in .NET 5 SDK (or greater ;)

IEvangelist (approved):

PR #26627, opened 2021-10-22
Note that about timing of assembly loads

### LINKED ISSUE #6833: Types derived from MarshalByRefObject need a workaround to work when compiled for x64 runtimes
# The Problem:
A call to a class derived from `MarshalByRefObject`, created with `AppDomain.CreateInstanceFromAndUnwrap()` threw a `FileNotFoundException` because it tried to load the referenced Assemblies in the original, not the newly created `AppDomain`.

# This was my situation:
I have a type derived from `MarshalByRefObject`, that has simple Log methods on it. (references Serilog.dll)
```csharp
internal class Logger : MarshalByRefObject
{
    public void Information(string msg, params object?[] args) { /* implementation that uses Serilog.dll (that already got used in the ctor successfully */ }
}
```
My `<TargetFramework>` is `net48`

I create a different AppDomain with **DIFFERENT** PrivateBinPath and ApplicationBase than the current AppDomain.
This means the current AppDomain cannot resolve the libraries in PrivateBinPath automatically (it could with a AssemblyResolve event handler), but the goal is **NOT** to load into the main AppDomain:
```csharp
var domain = AppDomain.CreateDomain("LoggerAppDomain", null, new AppDomainSetup { PrivateBinPath = "...", ApplicationBase = "..." });
```

The Logger instance gets created (constructor runs fine and loads referenced assemblies correctly)
```csharp
var logger = (Logger) _appDomain.CreateInstanceFromAndUnwrap(GetType().Assembly.Location, typeof(Logger).FullName, false, BindingFlags.Default, null, new object[] { /* some args */ }, null, Array.Empty<object>() );
```

The VisualStudio Debug Window shows Serilog.dll beeing loaded into the `LoggerAppDomain`

Calling `logger.Information(...)` resulted in a `FileNotFoundException` with Serilog.dll not beeing found.

I tried this in a Unit-Test project, and it worked flawlessly.

It did not work in my library project...

# The Solution:
What i found is [this](https://social.msdn.microsoft.com/Forums/sqlserver/de-DE/c6a62351-7bc8-49ea-9bc7-3720de7444da/assemblyloader-filenotfoundexception-trotz-privatebinpath-in-64bit-windows-2008?forum=visualcsharpde): (sorry, link is in german)
It references information from a connect.microsoft.com post and an answer by `David Wrighton from the .NET Framework CLR Team` in english and thankfully also translates it in place, or the information would be lost, because connect.microsoft.com does not exist any more.

The solution is to mark the Method with `virtual` like this:
```csharp
internal class Logger : MarshalByRefObject
{
    public **virtual** void Information(string msg, params object?[] args) { /* implementation that uses Serilog.dll (that already got used in the ctor successfully */ }
}
```
Why? (Translated back from the aforementioned post)

> This problem is caused by the fact that non-virtual remote-enabled methods are jitted differently on x86 and x64 machines. 
>When a method is called, the x86 runtime performs all the necessary remoting transformations and then tries to jitter the method into the target AppDomain. 
>**The behavior of the x64 runtime is subtly different:** 
>**The runtime does not prepare the method for remoting the first time it is called, but rather at an earlier point in time (as soon as it finds this method in an AppDomain).** 
> **If this is a remote-capable method, it will now be jitted twice by the x64 runtime.** 

My code would work fine in a x86 or AnyCPU Project (like my UnitTest project), but the x64 JIT-Compiler `NEEDS` the virtual keyword to properly marshal this call...

Would it please be possible to document this behavior somewhere on the page for MarshalByRefObject or any other Remoting documentation?

jkotas:
## Summary

Add note that timing of static assembly reference loads is unspecified

Fixes https://github.com/dotnet/core/issues/6833

### DIFF
--- a/docs/core/dependency-loading/loading-managed.md
+++ b/docs/core/dependency-loading/loading-managed.md
@@ -12,7 +12,7 @@ All managed assemblies except satellite assemblies and `WinRT` assemblies use th
 
 ## When are managed assemblies loaded?
 
-The most common mechanism to trigger a managed assembly load is a static assembly reference. These references are inserted by the compiler whenever code uses a type defined in another assembly. These assemblies are loaded (`load-by-name`) as needed by the runtime.
+The most common mechanism to trigger a managed assembly load is a static assembly reference. These references are inserted by the compiler whenever code uses a type defined in another assembly. These assemblies are loaded (`load-by-name`) as needed by the runtime. The exact timing of when the static assembly references are loaded is unspecified. It can vary between runtime version and it is influenced by optimizations like inlining.
 
 The direct use of specific APIs will also trigger loads:
 


gewarren (approved):
LGTM

PR #26629, opened 2021-10-22
Document additional environment variables

terrajobst:
This adds documentation for `DOTNET_SKIP_FIRST_TIME_EXPERIENCE` and `SuppressNETCoreSdkPreviewMessage`. Technically, `SuppressNETCoreSdkPreviewMessage` is an MSBuild property, not an environment variable, but it will often be used as an environment variable.


JamieMagee (approved):

tdykstra (approved):
LGTM, just a couple of suggestions to make `true` formatting consistent with the rest of the article.

PR #26635, opened 2021-10-23
Fix two inline code span errors

mondeja:
## Summary

Fixes two inline code span errors found in https://github.com/DavidAnson/markdownlint/pull/447


gewarren (approved):
The first change looks fine in the preview.

PR #26641, opened 2021-10-25
Update re. https://github.com/dotnet/docs/issues/26464

### LINKED ISSUE #26464: Is the Collections guidelines page out of date?
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections?source=docs

... says to prefer to return concrete read-only collections, e.g. `ReadOnlyCollection<T>` instead of `IReadOnlyCollection<T>`:

> ✔️ DO use ReadOnlyCollection<T>, a subclass of ReadOnlyCollection<T>, or in rare cases IEnumerable<T> for properties or return values representing read-only collections.

Following Postel's Law:
> "be conservative in what you send, be liberal in what you accept"

... the concrete type **is** more conservative than the interface, but does this trump 'prefer abstractions over concretes'?

Even if it is deemed by concensus that the this rule still holds, maybe a paragraph is required to show that the alternative has been considered.

SteveDunn:
## Summary

Added a note to the page to say that the guidelines are reprinted from a book dated 2008 and may be out of date/superseded.

Fixes #26464

### DIFF
--- a/docs/standard/design-guidelines/guidelines-for-collections.md
+++ b/docs/standard/design-guidelines/guidelines-for-collections.md
@@ -6,6 +6,9 @@ ms.assetid: 297b8f1d-b11f-4dc6-960a-8e990817304e
 ---
 # Guidelines for Collections
 
+**Note:**
+The guidelines below are reprinted from the 2008 book [Framework Design Guidelines: Conventions, Idioms, and Patterns for Reusable .NET Libraries, 2nd Edition](https://www.informit.com/store/framework-design-guidelines-conventions-idioms-and-9780321545619). Some of these guidelines may now be out of date and some types mentioned may now be superseded, for example, returning the concrete `ReadOnlyCollection` rather than the _interface_ `IReadOnlyCollection` which was introduced 4 years later in .NET 4.5.
+
 Any type designed specifically to manipulate a group of objects having some common characteristic can be considered a collection. It is almost always appropriate for such types to implement <xref:System.Collections.IEnumerable> or <xref:System.Collections.Generic.IEnumerable%601>, so in this section we only consider types implementing one or both of those interfaces to be collections.
 
  ❌ DO NOT use weakly typed collections in public APIs.


BillWagner:
Tagging @tdykstra 

What are your thoughts on this change?

MSDN-WhiteKnight:
If guidelines are outdated, it should be solved in more consistent way. Why should we single out the collections page specifically, if there are pages that are actually more outdated? For example, the [Serialization](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/serialization) talks about BinaryFormatter and DataContractJsonSerializer, which are no longer recommended for serialization. Also, adding such visible note "guidelines are out of date" on top of the page could make reader think that all guidelines are useless and should not be followed, which is not the case. 

Ideally, i'd rather want up-to-date guidelines, instead of outdated guidelines with deprecation notes attached to them. As i understand, the new edition of the book is about to be published: https://www.amazon.com/Framework-Design-Guidelines-Conventions-Addison-Wesley/dp/0135896460/ref=sr_1_2?keywords=Framework+Design+Guidelines&qid=1581295416&sr=8-2 , but will it be possible to publish its contents on Microsoft Docs? It also makes me think, could the Framework Guidelines use the same process as C# standard, when draft version of guidelines is edited in open source repository, and when changes are ready they are published as a book? That way community could suggests changes in that repository, instead of having them constantly shot down in this repository due to copyrighted content.

BillWagner:
@MSDN-WhiteKnight Great comment. 

We are talking with the publisher to acquire the rights to update this content. That would be the best solution. We haven't finished that work.

SteveDunn:
I really liked the original book (I was a community reviewer for it), but it is focused on designing 'Frameworks' (which is fair enough, that's exactly what it says on the tin!), but users are blindly applying some of the guidelines to applications, where they make less sense.  I haven't got a definitive list of those guidelines that are specific to 'Frameworks', but I'll try to get a list together, maybe in a blog post at https://blog.dunnhq.com

SteveDunn:
Perhaps all the CA warnings could be grouped by 'applicability', e.g. applicable to a 'reusable library of framework', or applicable to 'a non reusable application (console/webapi etc.)'

BillWagner:
Hi 👋 @SteveDunn 

We're archiving this content until we work with the authors and publishers to acquire permission to the latest version.  See #27495

Because of that, I'll close this PR.

SteveDunn:
> Hi 👋 @SteveDunn
> 
> We're archiving this content until we work with the authors and publishers to acquire permission to the latest version. See #27495
> 
> Because of that, I'll close this PR.

OK @BillWagner - thanks for the update

SteveDunn:
> If guidelines are outdated, it should be solved in more consistent way. Why should we single out the collections page specifically, if there are pages that are actually more outdated? For example, the [Serialization](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/serialization) talks about BinaryFormatter and DataContractJsonSerializer, which are no longer recommended for serialization. Also, adding such visible note "guidelines are out of date" on top of the page could make reader think that all guidelines are useless and should not be followed, which is not the case.
> 
> Ideally, i'd rather want up-to-date guidelines, instead of outdated guidelines with deprecation notes attached to them. As i understand, the new edition of the book is about to be published: https://www.amazon.com/Framework-Design-Guidelines-Conventions-Addison-Wesley/dp/0135896460/ref=sr_1_2?keywords=Framework+Design+Guidelines&qid=1581295416&sr=8-2 , but will it be possible to publish its contents on Microsoft Docs? It also makes me think, could the Framework Guidelines use the same process as C# standard, when draft version of guidelines is edited in open source repository, and when changes are ready they are published as a book? That way community could suggests changes in that repository, instead of having them constantly shot down in this repository due to copyrighted content.

The publish date for the 3rd Edition says June '2020.  I can still see the collections guidance is still live though (just had someone reference it my team)

SteveDunn:
> Hi 👋 @SteveDunn
> 
> We're archiving this content until we work with the authors and publishers to acquire permission to the latest version. See #27495
> 
> Because of that, I'll close this PR.

Hi @BillWagner - I can still see that it's live. In discussion with team members, this page is still getting quoted as best-practice. https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections?source=docs

PR #26651, opened 2021-10-25
CA5400 is similar to CA5399

kartheekp-ms:
## Summary

Describe your changes here.

`CA5400` analyzer is similar to `CA5399`. The doc says `CA5400` rule is similar to itself which seems incorrect to me.


gewarren (approved):
Great catch @kartheekp-ms

PR #26725, opened 2021-10-29
Update conventions.md

Thecentury:
## Summary

Replaced a namespace `TransactionTests` that contains `let` bindings (which won't compile) with a module with the same name.


BillWagner (approved):
Thanks @Thecentury 

We appreciate it. 

Issue #26752, opened 2021-11-01
Import or Using in the csproj file?


aktxyz:
I tried the following line from this page to add a namespace to the implicit using list ...

`<Import Include="System.IO.Pipes" />`

and that did not seem to work.  Googling around I found another tag name that did work

`<Using Include="System.IO.Pipes" />`

https://dotnetcoretutorials.com/2021/08/31/implicit-using-statements-in-net-6/

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 40c80da1-37e3-efc4-5208-4ef4c4d84cfd
* Version Independent ID: bcf66f2a-bd70-ea8c-e9f0-198c897a8634
* Content: [Breaking change: Implicit `global using` directives in C# projects - .NET](https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/6.0/implicit-namespaces)
* Content Source: [docs/core/compatibility/sdk/6.0/implicit-namespaces.md](https://github.com/dotnet/docs/blob/main/docs/core/compatibility/sdk/6.0/implicit-namespaces.md)
* Product: **dotnet-fundamentals**
* GitHub Login: @gewarren
* Microsoft Alias: **gewarren**

gewarren:
Thanks @aktxyz. It's a little confusing because there were two separate breaking changes - one for Preview 7 and another one for RC 1. `<Import>` worked for Preview 7, and `<Using>` works for RC 1 and later versions. See https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/6.0/implicit-namespaces-rc1 for the RC 1 changes.

I've created PR #26756 to hopefully make this less confusing.

PR #26756, opened 2021-11-01
Remove Preview 7-specific breaking change

gewarren:
The breaking change entry for Preview 7 seems to have been causing confusion because it uses properties that were only valid for Preview 7. I think it's best just to remove it since Preview 7 was no longer supported once RC 1 released.

Related to #26752.


docs/whats-new/dotnet-docs-2021-07-01.md
```diff
@@ -24,7 +24,7 @@ Welcome to what's new in the .NET docs from July 1, 2021 through July 31, 2021.
 ### New articles
 
 - [Culture creation and case mapping in globalization-invariant mode](../core/compatibility/globalization/6.0/culture-creation-invariant-mode.md)
-- [Implicit `global using` directives in C# projects](../core/compatibility/sdk/6.0/implicit-namespaces.md)
+- Implicit `global using` directives in C# projects (Later removed.)
```
tdykstra, line 27: I think it would cause less confusion if we just delete this rather than keep it with "later removed," which raises the question "Why?" That said, a July what's new in docs probably doesn't get much traffic, so it doesn't make a big difference.

docs/core/compatibility/sdk/6.0/implicit-namespaces-rc1.md
```diff
@@ -1,19 +1,21 @@
 ---
 title: "Preview-to-preview breaking change: Implicit `global using` directives disabled for existing projects"
-description: Learn about the preview-to-preview breaking change in .NET 6 where the implicit `global using` directives feature is disabled by default for existing C# projects.
+description: Learn about the preview-to-preview breaking change in .NET 6 where the implicit `global using` directives feature is only enabled for new C# projects.
```
tdykstra, line 3:
```suggestion
description: Learn about the preview-to-preview breaking change in .NET 6 where the implicit `global using` directives feature is enabled by default only for new C# projects.
```

tdykstra (approved):
LGTM

docs/whats-new/dotnet-docs-2021-07-01.md
gewarren, line 27:
```suggestion
```

PR #26789, opened 2021-11-03
Fix operator precedence for switch and with expressions

svick:
According to a comment by @agocke (https://github.com/dotnet/csharplang/discussions/5378#discussioncomment-1579123) and [the source code of the compiler](https://github.com/dotnet/roslyn/blob/30c493a/src/Compilers/CSharp/Portable/Parser/LanguageParser.cs#L10228-L10230), `switch` and `with` expressions have the same precedence, so they should be listed on the same row in the table.


BillWagner (approved):
Thanks for making this change @svick 

I've reviewed this and I'll :shipit: now.

PR #26801, opened 2021-11-03
Add improved definite assignment

### LINKED ISSUE #25671: Improved definite assignment
[Feature spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/improved-definite-assignment.md)

This may not affect a lot of docs. It will be seen as a set of improvements to when a variable is considered to be definitely assigned.

The same improvements will affect not-null / maybe-null analysis.

BillWagner:
Fixes #25671

This is a smaller PR. The main impact of the feature is that existing warning diagnostics are more accurate.

### DIFF
--- a/docs/csharp/whats-new/csharp-10.md
+++ b/docs/csharp/whats-new/csharp-10.md
@@ -115,7 +115,7 @@ int x = 0;
 
 ## Improved definite assignment
 
-There were a number of scenarios were definite assignment, and null-state analysis produced warnings that were false positives. These generally involved comparisons to boolean constants, accessing a variable only in the `true` or `false` statements in an `if` statement, and null coalescing expressions. These examples generate warnings in previous versions of C#, but don't in C# 10:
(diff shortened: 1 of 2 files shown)


docs/csharp/whats-new/csharp-10.md
```diff
@@ -112,6 +113,32 @@ int x = 0;
 > [!NOTE]
 > When using .NET 6.0 preview 5, this feature requires setting the `<LangVersion>` element in your *csproj* file to `preview`.
 
+## Improved definite assignment
+
+There were a number of scenarios were definite assignment, and null-state analysis produced warnings that were false positives. These generally involved comparisons to boolean constants, accessing a variable only in the `true` or `false` statements in an `if` statement, and null coalescing expressions. These examples generate warnings in previous versions of C#, but don't in C# 10:
```
gewarren, line 118:
```suggestion
Prior to C# 10, there were a number of scenarios where definite assignment and null-state analysis produced warnings that were false positives. These generally involved comparisons to boolean constants, accessing a variable only in the `true` or `false` statements in an `if` statement, and null coalescing expressions. These examples generated warnings in previous versions of C#, but don't in C# 10:
```

gewarren (approved):

PR #26823, opened 2021-11-04
Add constructor to breaking change article

### LINKED ISSUE #26763: [WinForms] API validate input parameters and throw ANE
## Selected API validate input parameters and throw `ArgumentNullException`.

Selected API now validate input parameters and throw `ArgumentNullException` where previously a `NullReferenceException` could be thrown, if APIs were invoked with `null` input arguments.
Throwing `ArgumentException` is inline with the behaviours in .NET Runtime, and provides developers with a better debug experience clearly communicating which argument caused the exception.

### Version

.NET 6 RC 1

### Previous behavior

The listed API would throw `NullReferenceException` if invoked with `null` input arguments.

### New behavior

The listed API would throw `ArgumentNullException` if invoked with `null` input arguments.

### Type of breaking change

### Reason for change

Throwing `ArgumentNullException` is inline with the behaviors in .NET Runtime, and provides developers with a better debug experience clearly communicating which argument caused the exception.

### Recommended action

- Review and, if necessary, update the code to prevent passing `null` input arguments in the listed API.
- If the code is handling `NullReferenceException` replace (or supplement) it with handling `ArgumentNullException`.

### Feature area

Windows Forms

### Affected APIs

* [DataGridViewColumnStateChangedEvent](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.datagridviewcolumnstatechangedeventargs)
	- [x] [DataGridViewColumnStateChangedEvent(DataGridViewRow, DataGridViewElementStates) Constructor](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.datagridviewcolumnstatechangedeventargs.-ctor) when `dataGridViewColumn` is `null`

gewarren:
Fixes #26763.

### DIFF
--- a/docs/core/compatibility/windows-forms/6.0/apis-throw-argumentnullexception.md
+++ b/docs/core/compatibility/windows-forms/6.0/apis-throw-argumentnullexception.md
@@ -1,7 +1,7 @@
 ---
 title: "Breaking change: Some APIs throw ArgumentNullException"
 description: Learn about the breaking change in .NET 6 where some APIs validate arguments and now throw an ArgumentNullException.
-ms.date: 01/29/2021
+ms.date: 11/04/2021
 ---
 # Some APIs throw ArgumentNullException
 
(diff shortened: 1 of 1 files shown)


dreddy-work (approved):
LGTM.

PR #26825, opened 2021-11-04
Add nullability change

### LINKED ISSUE #26767: [Non-Breaking change in nullability]: XmlDocument.XmlResolver setter is now nullable
### Description

XmlDocument.XmlResolver setter now accepts nullable reference.
Note that this is only a change in nullability and it's non-breaking.

```diff
public virtual System.Xml.XmlResolver XmlResolver { set { } }
public virtual System.Xml.XmlResolver? XmlResolver { set { } }
```

Changed between 5.0 and 6.0

related: https://github.com/dotnet/runtime/pull/45767

### Version

.NET 6 RC 1

### Previous behavior

XmlDocument.XmlResolver setter was annotated as not accepting null.

### New behavior

XmlDocument.XmlResolver setter now accepts null

### Type of breaking change

### Reason for change

previous annotation was incorrect.

### Recommended action

None.

### Feature area

XML, XSLT

### Affected APIs

System.Xml.XmlDocument.XmlResolver (setter)

gewarren:
Fixes #26767 

[Preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/xmlresolver-nullable?branch=pr-en-us-26825)

### DIFF
--- a/docs/core/compatibility/core-libraries/6.0/xmlresolver-nullable.md
+++ b/docs/core/compatibility/core-libraries/6.0/xmlresolver-nullable.md
@@ -1,0 +1,44 @@
+---
+title: "Breaking change: XmlDocument.XmlResolver nullability change"
+description: Learn about the .NET 6 breaking change where the setter for XmlDocument.XmlResolver was made nullable.
+ms.date: 11/04/2021
+---
+# XmlDocument.XmlResolver nullability change
+
+The nullability for <xref:System.Xml.XmlDocument.XmlResolver?displayProperty=nameWithType> has changed in .NET 6. The property setter now accepts a nullable reference.
+
+## Previous behavior
+
+The setter for <xref:System.Xml.XmlDocument.XmlResolver?displayProperty=nameWithType> was annotated as not accepting `null`:
+
+```csharp
+public virtual System.Xml.XmlResolver XmlResolver { set { } }
+```
+
+## New behavior
+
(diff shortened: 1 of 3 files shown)


docs/core/compatibility/core-libraries/6.0/xmlresolver-nullable.md
```diff
@@ -0,0 +1,44 @@
...
+public virtual System.Xml.XmlResolver? XmlResolver { set { } }
+```
+
+## Version introduced
+
+6.0 RC 1
+
+## Type of breaking change
+
+This change can affect [binary compatibility](../../categories.md#binary-compatibility) if you've overridden this property in a derived type.
```
gewarren, line 32: Not sure about source vs. binary incompatibility for the derived type case.

BillWagner: Adding @MadsTorgersen to be sure, but I think this would be only a source incompatibility change.

IEvangelist (approved):

docs/core/compatibility/core-libraries/6.0/xmlresolver-nullable.md
IEvangelist: I think it's more valuable to have this article out with .NET 6, so we can come back and tweak the wording on this if need be. I'm going to :shipit:

PR #26828, opened 2021-11-05
Fix typos in C# guide

GitHubPang:


BillWagner (approved):
Thanks again @GitHubPang 

These are great, and I'll :shipit: now.

PR #26848, opened 2021-11-06
JSON number handling breaking change

### LINKED ISSUE #26771: [Breaking change]: [JsonNumberHandling] on collection of non-number properties
### Description

Minor breaking change introduced in .NET 6 Preview 1:

Applying `JsonNumberHandlingAttribute` on a collection of non-numbers e.g. List<Poco> was previously (in 5.0) a no-op. Now, `InvalidOperationException` is thrown as it is not a valid configuration.

```cs
public class ClassWith_AttributeOnComplexListProperty
{
    // Attribute not allowed here, previously ignored.
    [JsonNumberHandling(JsonNumberHandling.AllowReadingFromString | JsonNumberHandling.WriteAsString)]
    public List<MyClass> MyList { get; set; }

    // Attribute allowed here and will be honored.
    [JsonNumberHandling(JsonNumberHandling.AllowReadingFromString | JsonNumberHandling.WriteAsString)]
    public List<int> MyOtherList { get; set; }
}
```

### Version

### Previous behavior

Provided in description.

### New behavior

Provided in description.

### Type of breaking change

### Reason for change

It was a side effect of a performance optimization for the number handling feature.

### Recommended action

Remove the attribute on incompatible collection properties.

### Feature area

Core .NET libraries

### Affected APIs

All `JsonSerializer` (de)serialization methods.

gewarren:
Fixes #26771 

[Preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/serialization/6.0/jsonnumberhandlingattribute-behavior?branch=pr-en-us-26848)

### DIFF
--- a/docs/core/compatibility/serialization/6.0/jsonnumberhandlingattribute-behavior.md
+++ b/docs/core/compatibility/serialization/6.0/jsonnumberhandlingattribute-behavior.md
@@ -12,8 +12,6 @@ A minor breaking change was introduced in .NET 6 with regard to the <xref:System
 public List<int> MyList { get; set; }
 ```
 
-<xref:System.Text.Json.Serialization.JsonNumberHandlingAttribute>
-
 ## Previous behavior
 
 Although it was ignored during serialization, <xref:System.Text.Json.Serialization.JsonNumberHandlingAttribute> could be applied to properties that were collections of non-number types. For example:
--- a/docs/core/compatibility/serialization/6.0/jsonnumberhandlingattribute-behavior.md
+++ b/docs/core/compatibility/serialization/6.0/jsonnumberhandlingattribute-behavior.md
@@ -1,0 +1,48 @@
+---
+title: "Breaking change: JsonNumberHandlingAttribute on non-number collection properties"
+description: Learn about the .NET 6 breaking change where JsonNumberHandlingAttribute can now only be applied to properties that are collections of numbers.
+ms.date: 11/05/2021
+---
+# JsonNumberHandlingAttribute on collection properties
+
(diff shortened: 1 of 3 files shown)


docs/core/compatibility/serialization/6.0/jsonnumberhandlingattribute-behavior.md
```diff
@@ -0,0 +1,46 @@
+---
+title: "Breaking change: JsonNumberHandlingAttribute on non-number collection properties"
+description: Learn about the .NET 6 breaking change where JsonNumberHandlingAttribute can now only be applied to properties that are collections of numbers.
+ms.date: 11/05/2021
+---
+# JsonNumberHandlingAttribute on collection properties
+
+A minor breaking change was introduced in .NET 6 with regard to the <xref:System.Text.Json.Serialization.JsonNumberHandlingAttribute> attribute. If you apply the attribute to a property that's a collection of non-number values and attempt to serialize or deserialize the property, an <xref:System.InvalidOperationException> is thrown. The attribute is only valid for properties that are collections of number types, for example:
```
IEvangelist, line 8:
```suggestion
A breaking change was introduced in .NET 6 with regard to the <xref:System.Text.Json.Serialization.JsonNumberHandlingAttribute> attribute. If you apply the attribute to a property that's a collection of non-number values and attempt to serialize or deserialize the property, an <xref:System.InvalidOperationException> is thrown. The attribute is only valid for properties that are collections of number types, for example:
```

I'm not sure it's a good idea to classify this as a "minor" breaking change. Any one who had code in place before that used this attribute in this way, would now experience an exception. It's still a breaking change, even though it may not affect as many consumers.

IEvangelist (approved):
I had one nit, otherwise looks good to me, :shipit: when you're ready.

PR #26849, opened 2021-11-06
AddProvider breaking change

### LINKED ISSUE #26780: [Breaking change]: Adds null check upon logging provider registration
### Description

`Microsoft.Extensions.Logging.LoggerFactory` implements `ILoggerFactory` with a `AddProvider(ILoggerProvider)` method. 

`null` providers are not accepted and the method now throws `ArgumentNullException` if the logging provider parameter is null. 

PR fix: https://github.com/dotnet/runtime/pull/46002
Issue report: https://github.com/dotnet/runtime/issues/45599

### Version

.NET 6 RC 1

### Previous behavior

Prior to new change, `AddProvider` would not perform any argument validation on `provider`, meaning it will consider `null` as a "valid" provider and add it to the collection of providers internally.

Some operations inside the class will unnecessarily throw `NullReferenceException`s due to this behavior, like the `Dispose` method, which will capture these exceptions and do nothing.

### New behavior

`null` providers are not accepted and the method now throws `ArgumentNullException` if the logging provider parameter is null. 

```c#
var factory = new LoggerFactory();
Assert.Throws<ArgumentNullException>(() => ((ILoggerFactory)factory).AddProvider(null));
```

### Type of breaking change

### Reason for change

Some operations inside the class will unnecessarily throw `NullReferenceException`s due to this behavior, like the `Dispose` method, which will capture these exceptions and do nothing.

### Recommended action

N/A

### Feature area

### Affected APIs

https://apisof.net/catalog/7c879577-6459-ae40-6e26-ee46369220ec

gewarren:
Fixes #26780

[Preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/extensions/6.0/addprovider-null-check?branch=pr-en-us-26849)

### DIFF
--- a/docs/core/compatibility/6.0.md
+++ b/docs/core/compatibility/6.0.md
@@ -82,6 +82,7 @@ If you're migrating an app to .NET 6, the breaking changes listed here might aff
 
 | Title | Binary compatible | Source compatible | Introduced |
 | - | - | - | - |
+| [AddProvider checks for non-null provider](extensions/6.0/addprovider-null-check.md) | ✔️ | ❌ | RC 1 |
 | [Microsoft.Extensions.Configuration binder binds single elements to an array](extensions/6.0/bind-single-elements-to-array.md) | ✔️ | ❌ | RC 1 |
 
 ## Globalization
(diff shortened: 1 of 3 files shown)


IEvangelist (approved):

PR #26857, opened 2021-11-07
Lambda improvements

### LINKED ISSUE #25675: Lambda improvements
[Feature specification](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md)

This adds a few small but nice improvements for lambdas.

BillWagner:
Fixes #25675

Update language reference for lambda expression improvements.

### DIFF
--- a/docs/csharp/language-reference/operators/lambda-expressions.md
+++ b/docs/csharp/language-reference/operators/lambda-expressions.md
@@ -1,7 +1,7 @@
 ---
 title: "Lambda expressions - C# reference"
 description: Learn about C# lambda expressions that are used to create anonymous functions.
-ms.date: 09/25/2020
+ms.date: 11/08/2021
 helpviewer_keywords: 
   - "lambda expressions [C#]"
   - "outer variables [C#]"
@@ -40,7 +40,7 @@ You can use lambda expressions in any code that requires instances of delegate t
 
 [!code-csharp-interactive[lambda is argument in LINQ](snippets/lambda-expressions/Introduction.cs#Argument)]
 
-When you use method-based syntax to call the <xref:System.Linq.Enumerable.Select%2A?displayProperty=nameWithType> method in the <xref:System.Linq.Enumerable?displayProperty=nameWithType> class, for example in LINQ to Objects and LINQ to XML, the parameter is a delegate type <xref:System.Func%602?displayProperty=nameWithType>. When you call the <xref:System.Linq.Queryable.Select%2A?displayProperty=nameWithType> method in the <xref:System.Linq.Queryable?displayProperty=nameWithType> class, for example in LINQ to SQL, the parameter type is an expression tree type [`Expression<Func<TSource,TResult>>`](<xref:System.Linq.Expressions.Expression%601>). In both cases you can use the same lambda expression to specify the parameter value. That makes the two `Select` calls to look similar although in fact the type of objects created from the lambdas is different.
+When you use method-based syntax to call the <xref:System.Linq.Enumerable.Select%2A?displayProperty=nameWithType> method in the <xref:System.Linq.Enumerable?displayProperty=nameWithType> class, for example in LINQ to Objects and LINQ to XML, the parameter is a delegate type <xref:System.Func%602?displayProperty=nameWithType>. When you call the <xref:System.Linq.Queryable.Select%2A?displayProperty=nameWithType> method in the <xref:System.Linq.Queryable?displayProperty=nameWithType> class, for example in LINQ to SQL, the parameter type is an expression tree type [`Expression<Func<TSource,TResult>>`](<xref:System.Linq.Expressions.Expression%601>). In both cases, you can use the same lambda expression to specify the parameter value. That makes the two `Select` calls to look similar although in fact the type of objects created from the lambdas is different.
   
 ## Expression lambdas
 
@@ -50,7 +50,7 @@ A lambda expression with an expression on the right side of the `=>` operator is
 (input-parameters) => expression
 ```
 
-The body of an expression lambda can consist of a method call. However, if you are creating [expression trees](../../programming-guide/concepts/expression-trees/index.md) that are evaluated outside the context of the .NET Common Language Runtime (CLR), such as in SQL Server, you should not use method calls in lambda expressions. The methods will have no meaning outside the context of the .NET Common Language Runtime (CLR).
+The body of an expression lambda can consist of a method call. However, if you're creating [expression trees](../../programming-guide/concepts/expression-trees/index.md) that are evaluated outside the context of the .NET Common Language Runtime (CLR), such as in SQL Server, you shouldn't use method calls in lambda expressions. The methods will have no meaning outside the context of the .NET Common Language Runtime (CLR).
 
 ## Statement lambdas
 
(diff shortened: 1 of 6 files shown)


docs/csharp/language-reference/operators/lambda-expressions.md
```diff
@@ -204,12 +204,78 @@ customers.Where(c => c.City == "London");
...
 - The lambda must contain the same number of parameters as the delegate type.
-
 - Each input parameter in the lambda must be implicitly convertible to its corresponding delegate parameter.
-
 - The return value of the lambda (if any) must be implicitly convertible to the delegate's return type.
   
-Note that lambda expressions in themselves don't have a type because the common type system has no intrinsic concept of "lambda expression." However, it's sometimes convenient to speak informally of the "type" of a lambda expression. In these cases the type refers to the delegate type or <xref:System.Linq.Expressions.Expression> type to which the lambda expression is converted.
+## Natural type for lambda expressions
+
+Lambda expressions in themselves don't have a type because the common type system has no intrinsic concept of "lambda expression." However, it's sometimes convenient to speak informally of the "type" of a lambda expression.That informal "type" refers to the delegate type or <xref:System.Linq.Expressions.Expression> type to which the lambda expression is converted.
```
IEvangelist, line 212:
```suggestion
Lambda expressions in themselves don't have a type because the common type system has no intrinsic concept of "lambda expression." However, it's sometimes convenient to speak informally of the "type" of a lambda expression. That informal "type" refers to the delegate type or <xref:System.Linq.Expressions.Expression> type to which the lambda expression is converted.
```

docs/csharp/language-reference/operators/lambda-expressions.md
```diff
@@ -204,12 +204,78 @@ customers.Where(c => c.City == "London");
...
+
+Beginning with C# 10, some lambda expressions have a *natural type*. Instead of forcing you to declare a delegate type, such as `Func<...>` or `Action<...>` for a lambda expression, the compiler may infer the delegate type from the parameters and the type of the expression.  For example, consider the following declaration:
+
+```csharp
+var parse = (string s) => int.Parse(s);
+```
+
+The compiler can infer `parse` to be a `Func<string, int>`. In general, the compiler will use an available `Func` or `Action` delegate, if a suitable one exists. Otherwise, it will synthesize a delegate type. For example, the type must be synthesized if the lambda expression has `ref` parameters. When a lambda expression has a natural type, it can be assigned to a weaker type, such as <xref:System.Object?displayProperty=nameWithType>, or <xref:System.Delegate?displayProperty=nameWithType>:
```
IEvangelist, line 220:
```suggestion
The compiler can infer `parse` to be a `Func<string, int>`. In general, the compiler will use an available `Func` or `Action` delegate, if a suitable one exists. Otherwise, it will synthesize a delegate type. For example, the type must be synthesized if the lambda expression has `ref` parameters. When a lambda expression has a natural type, it can be assigned to a less explicit type, such as <xref:System.Object?displayProperty=nameWithType>, or <xref:System.Delegate?displayProperty=nameWithType>:
```

I feel like "less explicit" seems a bit more accurate, "weaker" feels a bit misleading. Why would it be weaker?

docs/csharp/language-reference/operators/lambda-expressions.md
```diff
@@ -204,12 +204,78 @@ customers.Where(c => c.City == "London");
...
+Beginning with C# 10, you can apply attributes to lambda expressions. The attributes are added before the parameter declaration. The lambda expression's parameter list must be parenthesized when there are attributes:
+
+``` c#
+Func<string, int> parse = [Example(1)] (s) => int.Parse(s);
+var choose = [Example(2)][Example(3)] object (bool b) => b ? 1 : "two";
+```
+
+You can apply any attribute that is valid on <xref:System.AttributeTargets.Method?displayProperty=nameWithType>.
+
+Lambda expressions are invoked differently than methods and local functions. Attributes don't have any effect when the lambda expression is invoked. Attributes on lambda expressions are useful for code analysis, and can be discovered via reflection.
```
IEvangelist, line 278: I'm not sure I entirely follow what you're implying here. If attributes do not have any effect when the lambda expression is invoked, what attributes should be avoided? For example, with Minimal API could I add an attribute to an expression that defines `RequiresScope`? Would that work?

docs/csharp/programming-guide/delegates/delegates-with-named-vs-anonymous-methods.md
```diff
@@ -1,40 +1,41 @@
...
+
+This is called using a named method. Delegates constructed with a named method can encapsulate either a [static](../../language-reference/keywords/static.md) method or an instance method. Named methods are the only way to instantiate a delegate in earlier versions of C#. However, in a situation where creating a new method is unwanted overhead, C# enables you to instantiate a delegate and immediately specify a code block that the delegate will process when it is called. The block can contain either a [lambda expression](../../language-reference/operators/lambda-expressions.md) or an [anonymous method](../../language-reference/operators/delegate-operator.md).
+
+The method that you pass as a delegate parameter must have the same signature as the delegate declaration. A delegate instance may encapsulate either static or instance method.
+
+Although the delegate can use an [out](../../language-reference/keywords/out-parameter-modifier.md) parameter, we do not recommend its use with multicast event delegates because you cannot know which delegate will be called.
```
IEvangelist, line 20:
```suggestion
> [!NOTE]
> Although the delegate can use an [out](../../language-reference/keywords/out-parameter-modifier.md) parameter, we do not recommend its use with multicast event delegates because you cannot know which delegate will be called.
```

I'd promote this to an alert, as is something that a developer skimming this would really need to know.

docs/csharp/whats-new/csharp-10.md
```diff
@@ -73,6 +75,16 @@ valid in C# 8.0 and later.
...
 
+## Lambda expression improvements
+
+C# 10 includes many improvements to how lambda expressions are handled:
+
+- Lambda expressions may have a [natural type](../language-reference/operators/lambda-expressions.md#natural-type-for-lambda-expressions), where the compiler can infer a delegate type from the lambda expression or method group.
+- Lambda expressions may declare a [return type](../language-reference/operators/lambda-expressions.md#declared-return-type) when the compiler can't infer it.
+- [Attributes](../language-reference/operators/lambda-expressions.md#attributes) can be applied to lambda expressions.
+
+These features make lambda expressions more similar to methods and local functions. They make it easier to use lambda expressions without declaring a variable of a delegate type, and they work more seamlessly with the new ASP.NET minimal APIs.
```
IEvangelist, line 86: I'm not sure, but is "minimal" supposed to be title case? I have always referred to it as "ASP.NET Core Minimal APIs", and the word Core is missing.

```suggestion
These features make lambda expressions more similar to methods and local functions. They make it easier to use lambda expressions without declaring a variable of a delegate type, and they work more seamlessly with the new ASP.NET Core Minimal APIs.
```

IEvangelist (approved):
A few minor things, but otherwise looks great! :shipit: when you're ready.

docs/csharp/language-reference/operators/lambda-expressions.md
BillWagner: Added clarification to that. 

I think the answer is no. Unless the ASP.NET team is using reflection to see the attributes on the lambda expression, the library won't have that information.

Issue #21005, opened 2020-10-09
.NET Hub page: New design plus update for .NET 6


IEvangelist:
[.NET Hub page](https://docs.microsoft.com/en-us/dotnet/)

scottaddie:
@IEvangelist Heads up that I'm unassigning myself. Who should this be assigned to?

IEvangelist:
I believe @Rick-Anderson should be, if not you.

Rick-Anderson:
@wadepickett  is the hub expert, assigning him.

wadepickett:
Changing this to reflect just the .NET Hub pages, and creating a new ASP.NET Hub page issue on the aspnetcore.docs repo to track that.

Related to ASP.NET Hub pages issue: AspNetCore.Docs#22599

Reviewed details of what is needed with Bill Wagner.   This update request is aligned with the .NET 6 timeline and there are new HUB designs we will be using that are not ready at the moment. (So I will not be starting on this yet.)

I will be handling both the .NET and ASP.NET Hub pages for .NET 6 changes

For now I reviewed all links, verified the Hub page is working as is.  It is.

IEvangelist:
Closed by #26869

PR #26869, opened 2021-11-08
dotnet 6.0 update to Hub pages

wadepickett:
[Internal Review](https://review.docs.microsoft.com/en-us/dotnet/?branch=pr-en-us-26869)

First round of immediate updates to align with .NET 6.0:
Contributes to #21005

- Updated links for What's New, reference and any other links where a .NET 6.0 specific topic is now available and the version has to be specified to get there.
- Reviewed past data for this hub page to see what might need a change.
- Reduced the number of top highlight links to reflect top most used links, and where top traffic is to be driven to.
- Removed original Maui links, new links will be added as owners provide new ones in a future PR.

### DIFF
--- a/docs/index.yml
+++ b/docs/index.yml
@@ -292,10 +292,6 @@ additionalContent:
         summary: API reference documentation for .NET
         url: ../api/index.md?view=net-6.0
       # Card
-      - title: ".NET Core API reference"
-        summary: API reference documentation for .NET Core
-        url: ../api/index.md?view=netcore-3.1&preserve-view=true
-      # Card
       - title: ".NET Framework API reference"
         summary: API reference documentation for .NET Framework
         url: ../api/index.md?view=netframework-4.8&preserve-view=true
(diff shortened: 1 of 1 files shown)


wadepickett:
@BillWagner & @gewarren,   These are just the immediate changes that need to occur for just the Hub page.  Landing page updates will be a separate PR.

docs/index.yml
```diff
@@ -298,7 +290,7 @@ additionalContent:
       # Card
       - title: ".NET API reference"
         summary: API reference documentation for .NET
-        url: ../api/index.md?view=net-5.0
+        url: ../api/index.md?view=net-6.0
       # Card
       - title: ".NET Core API reference"
```
gewarren, line 295: I don't think we should have separate cards for .NET and .NET Core API ref. So I suggest removing the .NET Core API ref card.

gewarren (approved):
Otherwise, looks great.

wadepickett:
Thanks for the review.  Just waiting for a build at this point, builds are slow today...

gewarren (approved):

PR #26874, opened 2021-11-08
Resolving disposed service provider

### LINKED ISSUE #26783: [Breaking change]: Throw when resolving on a disposed service provider
### Description

Resolving Services after ServiceProvider is Disposed should throw with `ObjectDisposedException`.

PR: https://github.com/dotnet/runtime/pull/45116

### Version

.NET 6 RC 1

### Previous behavior

Prior to the new change, if the application when through the rare case of resolving a service after service provider is disposed, it would lead to deadlock.

### New behavior

Rather than deadlock, it will throw `ObjectDisposedException`.

### Type of breaking change

### Reason for change

To unblock the deadlock scenario.

### Recommended action

N/A

### Feature area

### Affected APIs

`IServiceProvider` `GetService` or `GetRequiredService` with `Dispose` usages in combination.

gewarren:
Fixes #26783 

[Preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/extensions/6.0/service-provider-disposed?branch=pr-en-us-26874)


docs/core/compatibility/extensions/6.0/service-provider-disposed.md
```diff
@@ -0,0 +1,39 @@
...
+
+This change was introduced to fix the deadlock scenario.
+
+## Recommended action
+
+Catch <xref:System.ObjectDisposedException> when calling any of the affected APIs.
+
+## Affected APIs
+
+- <xref:System.IServiceProvider.GetService(System.Type)?displayProperty=fullName>
```
gewarren, line 36: Not sure about the affected APIs - please double check.

IEvangelist (approved):

PR #26888, opened 2021-11-09
Update Program.cs for usability and a tyop

icyfire0573:
detecte is changed to detected; typo 
\t changed to , in above so that the spacing is the same for the headers and the text output
added , to console writeline to delimit the output and changed the "on, de" to not have a comma so the output can be put into a csv seamlessly for examination

## Summary

Describe your changes here.


adegeo:
- Verified the change is required. The article and code uses `,` as a separator but the header code was using a tab.
- Fixed merge conflict.

adegeo (approved):

PR #27010, opened 2021-11-10
Fix text element count not starting at 1

enchant97:
When the code sample `InsertNewlinesEveryTenTextElements` is run, the first newline is 11 characters long when it is meant to be 10. This commit fixes that by making the `textElementCount` start at 1.


IEvangelist:
Close and reopen to build

IEvangelist (approved):

PR #27041, opened 2021-11-11
Added alert immediately following the H1

### LINKED ISSUE #19868: Article is unnecessarily complicated
- "Create a class library" - it is not necessary to have the whole thing in the System.Numbers namespace and then to later include an explanation of why the SUT was put into the System.Numbers namespace. Just put the thing in a "Numbers" namespace to match the name of the project that was created and omit the explanation about not needing `using System`. The side-tangent about how namespaces work in C# is not needed to explain how to get code coverage reports from unit tests - it only adds additional cognitive load and friction.
- "Create test projects" - no explanation is given as to why we're creating two test projects. Furthermore, the naming of the two different test projects are rather confusing, as naming the projects starting with "XUnit" is something I would expect to be doing if we were adding a new project to XUnit itself - but we aren't doing that - we're writing tests for our "Numbers" project, so I would expect our projects to be named something along the lines of "Numbers.Test*"
- Over 50% of the article is just creating a SUT project and unit test project(s). 99% of people who are trying to figure out how to add code coverage to a project already have a SUT project and unit test project. Could the whole article be reworked to minimize/eliminate all of the setup that a person reading the article can reasonably be expected to have already figured out? Just jump straight to the point where coverlet is added to the unit test project and go from there.
- "Tooling" (and all sections thereafter) - these sections are great! Its just that the leadup to it had me so confused, especially regarding the multiple different unit tests projects, that it took a while to even get this far.

In addition to the above comments on the current content of the page, it would be nice to add an additional section that explains how to get code coverage working in Azure pipelines, as currently, there doesn't seem to be any official Microsoft documentation on doing this process end-to-end. I found https://www.meziantou.net/computing-code-coverage-for-a-dotnet-core-project-with-azure-devops-and-coverlet.htm which does a great job of explaining this clearly and concisely.

---
#### Document Details

* ID: 588a6d72-f70f-15f8-7505-5db491c8a8c6
* Version Independent ID: ab6ce421-f5e5-94f1-e707-32763b3a6a24
* Content: [Use code coverage for unit testing - .NET Core](https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-code-coverage?tabs=linux)
* Content Source: [docs/core/testing/unit-testing-code-coverage.md](https://github.com/dotnet/docs/blob/master/docs/core/testing/unit-testing-code-coverage.md)
* Product: **dotnet-core**
* GitHub Login: @IEvangelist
* Microsoft Alias: **dapine**

IEvangelist:
## Summary

Added alert immediately following the H1, letting the reader know that they can safely skip ahead as the first several sections intentionally set up the sample project to test. Also, minor edit pass and added relevant links to Azure content.

Fixes #19868


docs/core/testing/unit-testing-code-coverage.md
```diff
@@ -3,23 +3,26 @@ title: Use code coverage for unit testing
...
-ms.date: 02/10/2021
+ms.date: 11/11/2021
 recommendations: false
 ---
 
 # Use code coverage for unit testing
 
-Unit tests help to ensure functionality, and provide a means of verification for refactoring efforts. Code coverage is a measurement of the amount of code that is run by unit tests - either lines, branches, or methods. As an example, if you have a simple application with only two conditional branches of code (_branch a_, and _branch b_), a unit test that verifies conditional _branch a_ will report branch code coverage of 50%.
+> [!IMPORTANT]
+> This article explains the creation of the example project under test. If you already have a project, you can skip ahead to the [Code coverage tooling](#code-coverage-tooling) section.
```
tdykstra, line 13:
```suggestion
> This article explains the creation of the example project. If you already have a project, you can skip ahead to the [Code coverage tooling](#code-coverage-tooling) section.
```

tdykstra (approved):
LGTM

Issue #27012, opened 2021-11-10
Snippets 5000 should be a required check


BillWagner:
We've been running the Snippets 5000 CI build for some time. We've generally reached the point where we don't merge PRs unless this CI build passes.

We should make it a required check. This would make auto-merge more consistent: As an optional check, it doesn't need to pass. Also, we'd be less tempted to skip it "just this once".

BillWagner:
Upvote this comment 👍 for "Yes, we should make it required."

BillWagner:
Upvote this comment 👍 for "No, keep it optional"

BillWagner:
I updated the settings today. Closing this as finished.

PR #27056, opened 2021-11-12
Fix errors for required statuses

BillWagner:
See #27012 

The way GitHub implements "required checks", if a check doesn't run, it's marked as "pending", and never completes. That meant our "Snippets 5000" would stay in a stuck "pending" state forever on any PR that didn't change source code, or a project file. 

The fix is to remove the "paths" filter, so this check does run on all PRs. See https://github.community/t/expected-waiting-for-status-to-be-reported/18001/6 for details.

The other main functional check is to change the trigger from "pull_request" to "pull_request_target". That runs the actions as configured on the target branch, not as created in the PR. As a result of that change, we won't need to allow checks to run on PRs created on other forks from people who aren't maintainers.

Other changes are cosmetic to improve the UI for the branch protection screen.

Detailed changes:

- Remove the paths filter.
- Update labels so it's easier to find this check in the "branch protection" page.
- Change `pull_request` to `pull_request-target`

### DIFF
--- a/.github/workflows/build-validation.yml
+++ b/.github/workflows/build-validation.yml
@@ -1,28 +1,12 @@
 # This is a basic workflow to help you get started with Actions
-name: Snippets 5000
+name: 'Snippets 5000'
 
 # Controls when the action will run. Triggers the workflow on push or pull request
 # events on the main branch only.
 on:
-  pull_request:
-    paths:
-      - "**.cs"
-      - "**.vb"
-      - "**.fs"
-      - "**.cpp"
-      - "**.h"
-      - "**.xaml"
-      - "**.razor"
-      - "**.cshtml"
-      - "**.vbhtml"
-      - "**.csproj"
-      - "**.vbproj"
-      - "**.fsproj"
-      - "**.vcxproj"
-      - "**.sln"
-      - "**global.json"
-      - "**snippets.5000.json"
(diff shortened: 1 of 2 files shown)


BillWagner:
ping @adegeo

.github/workflows/build-validation.yml
```diff
@@ -1,28 +1,12 @@
...
-      - "**.cshtml"
-      - "**.vbhtml"
-      - "**.csproj"
-      - "**.vbproj"
-      - "**.fsproj"
-      - "**.vcxproj"
-      - "**.sln"
-      - "**global.json"
-      - "**snippets.5000.json"
+  pull_request_target:
```
IEvangelist, line 7: Do we want to run this for .md only file changes? I think we still might want the `paths` filter, if it is applicable.

BillWagner: That's the point: If this action doesn't run, it gets marked as "pending", and the PR is stuck. I did run some tests, and this won't use many resources when there aren't any code changes. The LocateProjects exe will just look at the PR paths and ignore them. One exception might be CI updates in the API repo, where thousands of files change. The link above in the description explains the design from GitHub's support team.

gewarren (approved):
Thanks Bill!

PR #27073, opened 2021-11-14
Clarify information about reference validity in a finalizer (#26710)

shadow-cs:
## Summary

References of other instances in of finalizable instances are not reclaimed before finalizer runs, finalizable objects are sources of GC roots.

Also adds information about possible null references during finalize operation.

Fixes #26710

### DIFF
--- a/docs/standard/garbage-collection/implementing-dispose.md
+++ b/docs/standard/garbage-collection/implementing-dispose.md
@@ -87,7 +87,7 @@ If your class owns a field or property, and its type implements <xref:System.IDi
 :::code language="vb" source="../../../samples/snippets/visualbasic/VS_Snippets_CLR/conceptual.disposable/vb/Foo.vb":::
 
 > [!TIP]
-> There are cases when you may want to perform `null`-checking in a finalizer (which includes the `Dispose(false)` method invoked by a finalizer), one of the primary reasons is if you're unsure whether the instance got fully initialized (ie. exception may be thrown in a constructor).
+> There are cases when you may want to perform `null`-checking in a finalizer (which includes the `Dispose(false)` method invoked by a finalizer), one of the primary reasons is if you're unsure whether the instance got fully initialized (for example, an exception may be thrown in a constructor).
 
 ## Implement the dispose pattern
 
--- a/docs/standard/garbage-collection/implementing-dispose.md
+++ b/docs/standard/garbage-collection/implementing-dispose.md
@@ -77,7 +77,7 @@ The body of the method consists of three blocks of code:
 
   - **Managed objects that consume large amounts of memory or consume scarce resources.** Assign large managed object references to `null` to make them more likely to be unreachable. This releases them faster than if they were reclaimed non-deterministically.
 
-If the method call comes from a finalizer, only the code that frees unmanaged resources should execute. The implementer is responsible for ensuring that the false path doesn't interact with managed objects that may have been reclaimed. This is important because the order in which the garbage collector destroys managed objects during finalization is non-deterministic.
+If the method call comes from a finalizer, only the code that frees unmanaged resources should execute. The implementer is responsible for ensuring that the false path doesn't interact with managed objects that may have been disposed. This is important because the order in which the garbage collector disposes managed objects during finalization is non-deterministic.
 
 ## Cascade dispose calls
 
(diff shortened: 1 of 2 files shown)


shadow-cs:
There is also a tip starting with:

> It is possible for a base class to only reference managed objects, and implement the dispose pattern. In these cases, a finalizer is unnecessary. A finalizer is only required if you directly reference unmanaged resources.

I'd consider adding information that adding an unnecessary finalizer only increases GC pressure by promoting the instance a generation without additional benefit.

Just a thought though, I don't want to complicate things way too much...

docs/standard/garbage-collection/implementing-dispose.md
```diff
@@ -77,7 +77,7 @@ The body of the method consists of three blocks of code:
 
   - **Managed objects that consume large amounts of memory or consume scarce resources.** Assign large managed object references to `null` to make them more likely to be unreachable. This releases them faster than if they were reclaimed non-deterministically.
 
-If the method call comes from a finalizer, only the code that frees unmanaged resources should execute. The implementer is responsible for ensuring that the false path doesn't interact with managed objects that may have been reclaimed. This is important because the order in which the garbage collector destroys managed objects during finalization is non-deterministic.
+If the method call comes from a finalizer, only the code that frees unmanaged resources should execute. The implementer is responsible for ensuring that the false path doesn't interact with managed objects that may have been disposed. This is important because the order in which the garbage collector disposes managed objects during finalization is non-deterministic.
```
shadow-cs, line 80: I'd also consider adding a bit more detail. Namely:

- All instances and their references are live during finalization (subsequent reclamation of the instances are non-deterministic as well, but IMO this is out of scope of this article and not really important since those references are just dead references waiting for GC to run)
- Instance (graph) may be revived (exactly once) during finalization - in most cases, this is bad practice to rely on this
- Exception thrown in a finalizer terminate the runtime (in default configuration), this includes `ObjectDisposedException` which some `Dispose` calls may throw (despite the fact they should be idempotent). This is the real danger of disposing already disposed instances.

BillWagner: Ping @Maoni0 for review on this change. In particular, I wonder if using "dispose" in this paragraph could add confusion, as this is more about finalizers than Dispose methods.

samples/snippets/csharp/VS_Snippets_CLR/conceptual.disposable/cs/Foo.cs
```diff
@@ -9,5 +9,5 @@ public Foo()
         _bar = new Bar();
     }
 
-    public void Dispose() => _bar?.Dispose();
+    public void Dispose() => _bar.Dispose();
```
BillWagner, line 12: @shadow-cs I'm confused about this check. Why would you remove the null-conditional operator without adding a null check?

shadow-cs: This is why I added the `TIP` in the MD, I found it confusing why the `null` check was there it sort of pointed out to a previously stated assumption that the GC may somehow put `null`s there (by reclaiming fields before the instance owning them) which is not the case. The `null`-check is only possible in this case if the `Bar` ctor throws which is not a pattern I'd generally recommend.

docs/standard/garbage-collection/implementing-dispose.md
shadow-cs: I was unsure about how much detail you wanted to put in. I basically wanted to start a discussion on this topic and find the right balance. But I personally expected a bit more detail when I read this article the first time. A link to another article or even a more technical blog post would also help.

But maybe this is just me needing to understand the topic in detail and not really that helpful for someone looking at it in a broader view.

samples/snippets/csharp/VS_Snippets_CLR/conceptual.disposable/cs/Foo.cs
IEvangelist: My thought process is that it doesn't hurt to have it as a null-conditional as it will just no-op. 🤷🏼

shadow-cs: @IEvangelist I agree it doesn't hurt. But look at it from the other side, I began wondering why is the `null` check there, it shouldn't be necessary if `_bar` is initialized properly and then the `reclaimed/destroys` comes in and you begin wondering if it somehow possible `null` will ever be there? Is it possible the runtime does this? In fact, it doesn't which is sort of the point I'm trying to make. But maybe I'm just overthinking it. But chances are, there are other over-thinkers like me out there 😉

docs/standard/garbage-collection/implementing-dispose.md
```diff
@@ -86,6 +86,9 @@ If your class owns a field or property, and its type implements <xref:System.IDi
 :::code language="csharp" source="../../../samples/snippets/csharp/VS_Snippets_CLR/conceptual.disposable/cs/Foo.cs":::
 :::code language="vb" source="../../../samples/snippets/visualbasic/VS_Snippets_CLR/conceptual.disposable/vb/Foo.vb":::
 
+> [!TIP]
+> There are cases when you may want to perform `null`-checking in a finalizer (which includes the `Dispose(false)` method invoked by a finalizer), one of the primary reasons is if you're unsure whether the instance got fully initialized (ie. exception may be thrown in a constructor).
```
IEvangelist, line 90:
```suggestion
> There are cases when you may want to perform `null`-checking in a finalizer (which includes the `Dispose(false)` method invoked by a finalizer), one of the primary reasons is if you're unsure whether the instance got fully initialized (for example, an exception may be thrown in a constructor).
```

IEvangelist (approved):

PR #27080, opened 2021-11-15
Update cs5001.md

machielvisser:
## Summary

Return type void is not allowed when in the Main method when using async.

### DIFF
--- a/docs/csharp/misc/cs5001.md
+++ b/docs/csharp/misc/cs5001.md
@@ -14,7 +14,7 @@ Program does not contain a static 'Main' method suitable for an entry point
 
 This error occurs when no static `Main` method with a correct signature is found in the code that produces an executable file. It also occurs if the entry point function, `Main`, is defined with the wrong case, such as lower-case `main`. For information about the rules that apply to the `Main` method, see [Main() and Command-Line Arguments](../fundamentals/program-structure/main-command-line.md).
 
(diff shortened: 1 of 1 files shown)


docs/csharp/misc/cs5001.md
```diff
@@ -14,7 +14,7 @@ Program does not contain a static 'Main' method suitable for an entry point
 
 This error occurs when no static `Main` method with a correct signature is found in the code that produces an executable file. It also occurs if the entry point function, `Main`, is defined with the wrong case, such as lower-case `main`. For information about the rules that apply to the `Main` method, see [Main() and Command-Line Arguments](../fundamentals/program-structure/main-command-line.md).
 
-If the `Main` method has an `async` modifier, make sure that the [selected C# language version](../language-reference/configure-language-version.md) is 7.1 or higher.
+If the `Main` method has an `async` modifier, make sure that the [selected C# language version](../language-reference/configure-language-version.md) is 7.1 or higher and to use Task as the return type.
```
IEvangelist, line 17:
```suggestion
If the `Main` method has an `async` modifier, make sure that the [selected C# language version](../language-reference/configure-language-version.md) is 7.1 or higher and to use `Task` or `Task<int>` as the return type.
```

IEvangelist (approved):

PR #27102, opened 2021-11-16
Retitle Fact 1 in writing-large-responsive-apps.md

radrow:
Facts cannot be stated in an imperative form. Otherwise, I would call it a "tip".


BillWagner (approved):
Thanks for making this change @radrow 

PR #27104, opened 2021-11-16
Add the HTTP keyword to the expected list of available slug completions

IEvangelist:
## Summary (Auto-merge enabled)

Add the HTTP keyword to the expected list of available slug completions


tdykstra (approved):
LGTM

PR #27096, opened 2021-11-16
Add documentation for CA2255 (new in .NET 6)

jeffhandley:
## Summary

Add documentation for CA2255, which was introduced in .NET 6 via https://github.com/dotnet/roslyn-analyzers/pull/5347.

Closes https://github.com/dotnet/roslyn-analyzers/issues/5723

PR #27114, opened 2021-11-17
Remove backticks from page metadata

jeffhandley:
## Summary

As noted in https://github.com/dotnet/docs/pull/27096#discussion_r750562268, we should not have `` ` `` characters in page metadata (title/description) as it renders poorly where those strings are used. This removes the occurrences found by the regex: `` ^(title|description).*` ``.


IEvangelist (approved):

PR #27122, opened 2021-11-17
Fix the build for unit-testing-best-practices.sln on VS2022 by altering the arguments to ArgumentException

### LINKED ISSUE #27133: unit-testing-best-practices.sln does not compile with error CA2208 related to ArgumentException
### Reproduce

Checkout the main branch and build with VS2020. The compile fails on this line:
```csharp
throw new ArgumentException(nameof(numbers));
```
With this error:
```console
Error	CA2208	Method Add passes parameter name 'numbers' as the message argument to a ArgumentException constructor. Replace this argument with a descriptive message and pass the parameter name in the correct position.	unit-testing-best-practices	.\docs\samples\snippets\core\testing\unit-testing-best-practices\csharp\unit-testing-best-practices\StringCalculator.cs	13	Active
```

### Cause

The root has a Directory.Build.props with this setting
```
    <EnableNETAnalyzers>true</EnableNETAnalyzers>
    <AnalysisLevel>latest</AnalysisLevel>
```
This detects the `CA2208'. The name of the parameter is passed as the first parameter (the message) instead of the second.

The root also has an .editorconfig which has this setting to treat it as error:
```console
# CA2208: Instantiate argument exceptions correctly
dotnet_diagnostic.CA2208.severity = error
```

### Solution

A simple solution suggested by the ['how to fix'](https://docs.microsoft.com/nl-nl/dotnet/fundamentals/code-analysis/quality-rules/ca2208#how-to-fix-violations) is to use null for the message like this:

```csharp
throw new ArgumentException(null, nameof(numbers));
```

Although null is passed in the message of the ArgumentException object will be:
> Value does not fall within the expected range. (Parameter 'numbers')

pauldendulk:
## Summary

The project unit-testing-best-practices.sln does not build with a fresh install of VS2022 because of error CA2208.
The fix page mentioned below suggests two solutions. I implemented the second because that seems to be most in line the original intent to only pass in the name of the violating parameter.

See: https://docs.microsoft.com/nl-nl/dotnet/fundamentals/code-analysis/quality-rules/ca2208#how-to-fix-violations

Fixes #27133

### DIFF
--- a/samples/snippets/core/testing/unit-testing-best-practices/csharp/unit-testing-best-practices/StringCalculator.cs
+++ b/samples/snippets/core/testing/unit-testing-best-practices/csharp/unit-testing-best-practices/StringCalculator.cs
@@ -10,7 +10,7 @@ namespace UnitTestingBestPractices
         {
             if (numbers == null)
             {
-                throw new ArgumentException(nameof(numbers));
+                throw new ArgumentNullException(nameof(numbers));
             }
 
             int result;
@@ -49,7 +49,7 @@ namespace UnitTestingBestPractices
             var result = int.TryParse(number, out var parsedNumber);
             if (result == false)
             {
-                 throw new ArgumentException(nameof(number));
+                throw new ArgumentException($"Unable to parse {nameof(number)} as an Int32.", nameof(number));
             }
 
             return parsedNumber;


pauldendulk:
The snippet build is failing but this seems to be a general problem. Perhaps something wrong with the action. Judging from the error it tries to build main instead of the branch.

gewarren:
@adegeo Can you look at the Snippets 5000 failure? The error is the error that @pauldendulk fixed in the code, so it's odd that it's appearing.

Regarding the commit SHAs in the GitHub Actions, I think this was done for security reasons. Right @IEvangelist?

pauldendulk:
I reverted the change to build-validation.yml. It was an attempt to fix the build. I think that should be a separate issue.

IEvangelist:
> @adegeo Can you look at the Snippets 5000 failure? The error is the error that @pauldendulk fixed in the code, so it's odd that it's appearing.
> 
> Regarding the commit SHAs in the GitHub Actions, I think this was done for security reasons. Right @IEvangelist?

Hi @gewarren - yes, you're correct. I meant to submit my review of this PR yesterday, but I obviously got distracted with other things.

image

samples/snippets/core/testing/unit-testing-best-practices/csharp/unit-testing-best-practices/StringCalculator.cs
```diff
@@ -10,7 +10,7 @@ public int Add(string numbers)
         {
             if (numbers == null)
             {
-                throw new ArgumentException(nameof(numbers));
+                throw new ArgumentException(null, nameof(numbers));
```
IEvangelist, line 13:
```suggestion
                throw new ArgumentNullException(nameof(numbers));
```

What about something like this instead? This should really be an `ArgumentNullException`.

samples/snippets/core/testing/unit-testing-best-practices/csharp/unit-testing-best-practices/StringCalculator.cs
```diff
@@ -49,7 +49,7 @@ private static int HandleSingleNumber(string number)
             var result = int.TryParse(number, out var parsedNumber);
             if (result == false)
             {
-                 throw new ArgumentException(nameof(number));
+                 throw new ArgumentException(null, nameof(number));
```
IEvangelist, line 52:
```suggestion
                 throw new ArgumentException(
                     $"Unable to parse '{numbers}' as an Int32.", nameof(number));
```

And something like this?

IEvangelist:
This is a bit odd, I was curious so I wrote a [quick Sharplab.io](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK7wCYgNQB8ACATAIwCwAUBQG4TIAEAcgKYDuAgsgOYCiyyA9sgBaTAXQC8dABQBKCQD46MVnQ6dUYJjATcAHgGMmABwRR+MWQG5qtRqzW8ByAPLKJ0gCIQETACpRNOgxvJjlxRWUWVS4NLR0DY1NzKRgITX4AMylgnxkZa3IaemZ2LkdBXxZ+dykvH39AnNCFJRU1WO09QxMzCxhUABsBuCU0pkzskLyCinxiAE4pCjoVugASACIAOn0EQVkQAB0YFcONgG8Shz5BEQFZLYBZJgBnF4hOJgBfM+ON/NmCyW5FW622u32cwADAB+OiaN4fULHU4XK5lG4uZS1EINJhbBj8FgyJ6vd6fH4bP4A8hzRbLVabHZ7ZBSaFwhHkpgjdl0Iy0NIMMYyI4nOhnS72DFOSr8HH1AL4wnE0mIim/GD/GbkAD0OrozlQCCMRpAFD1zP2IvN+roGwAahABqgmEF+K8lPwEHQMk6BnQWFAEAALWB0EOupi6IxMXZMDB0ZAQGCfLYbG02y2s3mcpEyG0rDZNdO6nWZiHZ4iw+FkpE8qtw/lJsBCzTW0uFx3O10Yd0vT3e31DANB0MnCN0KMxuMJpMp/HSAAKAs0PnoAHImuuZBsgA==):

```csharp
using System;

var NewArgErrorZero = () => new ArgumentException();
var NewArgErrorOne = (DateTime date) => new ArgumentException(nameof(date));
var NewArgErrorTwo = (DateTime date) => new ArgumentException(null, nameof(date));

Console.WriteLine(
    $".ctor():\n  \"{NewArgErrorZero().Message}\"\n");
Console.WriteLine(
    $".ctor(string? message)\n  \"{NewArgErrorOne(DateTime.Now).Message}\"\n");
Console.WriteLine(
    $".ctor(string? message, string? paramName):\n  \"{NewArgErrorTwo(DateTime.Now).Message}\"\n");

// Output:
//.ctor():
//  "Value does not fall within the expected range."
//
//.ctor(string? message)
//  "date"
//
//.ctor(string? message, string? paramName):
//  "Value does not fall within the expected range. (Parameter 'date')"
```

I would have expected an error message like this to come from the `ArgumentOutOfRangeException` but not the `ArgumentException`. After looking a bit further, this is the correct localized message. I suppose we should probably try to have the `dotnet/runtime` team fix that as to help avoid confusion?

samples/snippets/core/testing/unit-testing-best-practices/csharp/unit-testing-best-practices/StringCalculator.cs
pauldendulk: Ah, that is clearly better. Did not know about the ArgumentNullException.

samples/snippets/core/testing/unit-testing-best-practices/csharp/unit-testing-best-practices/StringCalculator.cs
pauldendulk: My choice was to keep it functionally equivalent to how it was, but the null as first parameter does look weird and would raise some question marks. This is better.

pauldendulk:
> I would have expected an error message like this to come from the ArgumentOutOfRangeException but not the ArgumentException. After looking a bit further, this is the correct localized message. I suppose we should probably try to have the dotnet/runtime team fix that as to help avoid confusion?

Perhaps 'range' here is not intended as meaning a sequence with a beginning and end, but more broadly as 'the collection of things accepted'. It could be confusing though and can be improved.

IEvangelist:
> > I would have expected an error message like this to come from the ArgumentOutOfRangeException but not the ArgumentException. After looking a bit further, this is the correct localized message. I suppose we should probably try to have the dotnet/runtime team fix that as to help avoid confusion?
> 
> Perhaps 'range' here is not intended as meaning a sequence with a beginning and end, but more broadly as 'the collection of things accepted'. It could be confusing though and can be improved.

Yeah, I filled an issue. Perhaps we could improve the default exception messaging. https://github.com/dotnet/runtime/issues/61779

adegeo:
I believe I've figured out the source of the problem. I'm working on a solution. I'll follow up with more information when I'm done.

.github/workflows/build-validation.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-    - uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675 #@v2
```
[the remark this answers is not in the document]
pauldendulk: I reverted this with a new commit. Can you squash the PR?

adegeo:
@pauldendulk You'll need to rebase your fork and get the latest version of this repo to get rid of the old check.

When we changed Snippets 5000 to become a required check we thought we were securing it down by setting it to run on `pull_request_target` but that was actually running it in the context of the **target** branch, our main branch. Thus, Snippets 5000 always checked the code as it exists in our main branch and not in the context of the pull request.

BillWagner:
I messed this up. I badly misread the distinction between `pull_request` and `pull_request_target` in the GitHub action docs. Thanks @adegeo for working through it to address it.

pauldendulk:
@adegeo All checks pass now. I also force pushed a version with one commit.

IEvangelist (approved):

PR #27128, opened 2021-11-17
Restore the Azure for .NET devs tile

scottaddie:
Restore the Azure for .NET devs tile on the hub page. It was removed in https://github.com/dotnet/docs/pull/26869. This tile is key in driving traffic to the relevant Azure content.


IEvangelist (approved):

PR #27136, opened 2021-11-17
Updated to reflect support for .NET 6, 2.X, and 3.X

joshongithub:
Updated to reflect: 
* support for: .NET 6 (changed from just .NET 5)
* support for 2.X instead of 2.0 because 2.1 was also supported
* support for 3.X instead of 3.0 because 3.1 is also supported

## Summary

Describe your changes here.


adegeo:
Hello. Glancing at this PR I'm assuming that the table actually represents the minimum versions, so there is no need to specify `X` or calling out 6 specificaly.

joshongithub:
Now that you point that out, that makes sense. When I read of the table I thought it was outdated. I'll close this PR.

joshongithub:
Now that you point that out, that makes sense. When I read of the table I thought it was outdated. I'll close this PR.

Issue #27139, opened 2021-11-18
Updates for .NET 6 release


gewarren:
This issue tracks various updates needed now that .NET 6 has released.

- [x] Add a what's new article
- [x] Update hub and landing pages with what's new article, etc.
- [x] Remove "preview" references - including preview versions
- [x] Refresh API docs and update moniker to remove "prerelease"
- [x] Update API browser index
- [x] Update .NET docs header
- [x] Update all 'updateeachrelease' articles. Most are done. Remaining articles are:
  - [x] https://github.com/dotnet/docs/blob/main/docs/core/deploying/deploy-with-cli.md
  - [x] https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/keywords/index.md
  - [x] https://github.com/dotnet/docs/blob/main/docs/csharp/whats-new/csharp-version-history.md
- [x] Update ?view= links to latest version, as necessary
- [x] Update support tables - e.g. https://docs.microsoft.com/en-us/dotnet/core/install/windows?tabs=net60#supported-releases
- [x] Update TFMs and preprocessor symbols - https://docs.microsoft.com/en-us/dotnet/standard/frameworks
- [x] Remove preview-to-preview breaking changes
- [x] Update list of APIs that always throw - https://docs.microsoft.com/en-us/dotnet/core/compatibility/unsupported-apis
- [x] Update reference for settings that include a version:
  - [x] [AnalysisLevel](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#analysislevel) and [AnalysisMode](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#analysismode)
- [x] Update code analysis rules that are enabled by default - https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview#enabled-rules
- [x] Update version in breaking change issue template

Other considerations:

- Update samples and [snippets](https://github.com/dotnet/docs/issues/27310) to target new version (if it's LTS)
- Update code analysis level in samples (if necessary)
- Update tutorials to reflect latest version and any template/UI changes
- Update redirects that should always point to the latest version
- API docs: Unfilter any new attributes that should be shown in code signatures in the docs
- API docs: Update exceptions as necessary

Related issue: #26777

PR #27140, opened 2021-11-18
Update monikers for .NET API browser index

gewarren:
Contributes to #27139


IEvangelist (approved):

PR #27153, opened 2021-11-18
Change snippets target

adegeo:


BillWagner (approved):
Thanks @adegeo 

I think I mis-read the docs on these two target types.

Let's :shipit:

PR #27160, opened 2021-11-18
Remove unneeded lines from docfx

gewarren:
It doesn't seem like we need these lines in the docfx.json file?


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

PR #27200, opened 2021-11-20
Rewrite 'Avoid multiple asserts' to 'Avoid multiple acts' #27198

### LINKED ISSUE #27099: 'Avoid multiple asserts' is not good advice (and should be rewritten as Avoid multiple acts)
I am aware that this advice is repeated all over the internet but it is still not a good advice. 😄

The 'Bad' example is indeed bad but it is not because of the multiple asserts but because of the multiple acts

[Fact]
public void Add_EdgeCases_ThrowsArgumentExceptions()
{
    Assert.Throws<ArgumentException>(() => **stringCalculator.Add(null)**); // <-- Act 1 in bold
    Assert.Throws<ArgumentException>(() => **stringCalculator.Add("a")**); // < -- Act 2 in bold
}

Parameterized tests are mentioned as a solution, but parameters are typically only applied in the act of test. This can also be seen in the 'Good' example on the page. The 'input' parameter is only used in the act line.

The page mentions that with multiple asserts it does not 'give you the entire picture as to why your tests are failing', but the opposite is true. With multiple asserts you can test exactly what is wrong with your response. A first asserts can check for null, a second one can assert the item count, a third assert can check the content of the expected items. Such a failing test can tell you exactly what is wrong. The fact that the last assert is never called is not a problem at all, it would not make sense.

The general argument of the section is valuable and would still hold if it was rewritten as 'Avoid multiple acts'.

---
#### Document Details

* ID: 51947dc8-9ae8-21fa-13db-a7ffb3f9bc0e
* Version Independent ID: be661bf5-6f6e-cd18-344b-253edf5751ae
* Content: [Best practices for writing unit tests - .NET](https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#avoid-multiple-asserts)
* Content Source: [docs/core/testing/unit-testing-best-practices.md](https://github.com/dotnet/docs/blob/main/docs/core/testing/unit-testing-best-practices.md)
* GitHub Login: @jpreese
* Microsoft Alias: **wiwagn**

pauldendulk:
## Summary

The problem is not with multiple asserts. Multiple asserts is fine, as noted in the exception to the rule, the problem stems from multiple acts. By rewriting it as 'Avoid multiple acts' the exception to the rule can be removed.

The accompanying code sample is also updated because in the previous one the Assert and Act was conflated. By adding an explicit Act step the explanation becomes easier to understand.

Fixes #27099


docs/core/testing/unit-testing-best-practices.md
```diff
@@ -255,23 +255,20 @@ In unit testing frameworks, `Setup` is called before each and every unit test wi
...
-- Ensures you are not asserting multiple cases in your tests.
+- When the test fails it is not clear which Act is failing.
+- Ensures the test is focussed on just a single case.
 - Gives you the entire picture as to why your tests are failing.
 
-When introducing multiple asserts into a test case, it is not guaranteed that all of the asserts will be executed. In most unit testing frameworks, once an assertion fails in a unit test, the proceeding tests are automatically considered to be failing. This can be confusing as functionality that is actually working, will be shown as failing.
-
-> [!NOTE]
-> A common exception to this rule is when asserting against an object. In this case, it is generally acceptable to have multiple asserts against each property to ensure the object is in the state that you expect it to be in.
+Multiple Acts need to be individually Asserted and it is not guaranteed that all of the Asserts will be executed. In most unit testing frameworks, once an Assert fails in a unit test, the proceeding tests are automatically considered to be failing. This can be confusing as functionality that is actually working, will be shown as failing.
```
pauldendulk, line 271: Now that we talk about Acts instead of Asserts it is possible to remove the exception to the rule that was added as a note.

BillWagner (approved):
Thanks for making these changes @pauldendulk This is more clear guidance.

Thanks again!

PR #27215, opened 2021-11-22
Fix table rendering broken by embedded pipe characters.

DavidAnson:


DavidAnson:
To see the problem, look at the changed lines as rendered by GitHub pre-fix: https://github.com/dotnet/docs/blob/main/docs/core/diagnostics/dotnet-dump.md

DavidAnson:
For comparison, here’s the file post-fix: https://github.com/dotnet/docs/blob/6d7653d4f4601af48e8984619d2c2c5524a22214/docs/core/diagnostics/dotnet-dump.md

gewarren:
The existing syntax doesn't render properly in GitHub, but it does render properly on docs: https://review.docs.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-dump?branch=main#analyze-sos-commands.

And the fix in this PR makes the backslashes visible, so it's not the right fix, if a fix is needed at all: https://review.docs.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-dump?branch=pr-en-us-27215#analyze-sos-commands.

DavidAnson:
The fix is correct per CommonMark specification which is the most formal definition of Markdown and the one GitHub uses. Any ASCII punctuation character can be escaped: https://spec.commonmark.org/0.30/#backslash-escapes

It seems the Docs website uses a different standard or has a bug. Maybe it would be better to use a different character here, one that does not interfere with table layout?

gewarren:
[This page](https://review.docs.microsoft.com/en-us/help/contribute/markdown-reference?branch=master) in our contributor guide say that docs supports CommonMark parsed through Markdig. So I'm not sure where the problem lies. @meganbradley

DavidAnson:
This is interesting...

Tables are not part of the CommonMark spec, so we can't see what they intend, but they do show the escaping backslash in a normal code block: <https://spec.commonmark.org/dingus/?text=%23%20Test%0A%0A%60a%5C%7Cb%60%0A%0A%7C%20Column%20%7C%0A%7C--------%7C%0A%7C%20%60a%5C%7Cb%60%20%7C>

The markdown-it parser is fairly widely used and explicitly allows the behavior I'm describing: <https://markdown-it.github.io/#md3=%7B%22source%22%3A%22%23%20Test%5Cn%5Cn%60a%5C%5C%7Cb%60%5Cn%5Cn%7C%20Column%20%7C%5Cn%7C--------%7C%5Cn%7C%20%60a%5C%5C%7Cb%60%20%7C%22%2C%22defaults%22%3A%7B%22html%22%3Afalse%2C%22xhtmlOut%22%3Afalse%2C%22breaks%22%3Afalse%2C%22langPrefix%22%3A%22language-%22%2C%22linkify%22%3Atrue%2C%22typographer%22%3Atrue%2C%22_highlight%22%3Atrue%2C%22_strict%22%3Afalse%2C%22_view%22%3A%22html%22%7D%7D>

The Markdig team seems to be aware of this scenario and has specifically chosen the current behavior: <https://github.com/xoofx/markdig/issues/236>

GitHub suggests the approach I took, but does not say whether it should apply inside a code span: <https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/organizing-information-with-tables#formatting-content-within-your-table>

GitHub's specification for tables specifically says that pipes can be escaped inside inline spans (which would include code spans) with example 200: <https://github.github.com/gfm/#tables-extension->

But again, there is room for doubt whether the intent is meant to include code spans and it's reasonable to claim that it should not on principle.

This brings me back to my idea above to use a different character than pipe in order to avoid this situation entirely. :)

gewarren:
So something like this?

``| `soshelp <command>` or `help <command>` |`` instead of ``| `soshelp|help <command>` |``

I'll let @tdykstra make the final decision.

DavidAnson:
Yes! I think that's a little clearer to non-nerds as well.

tdykstra:
>So something like this?

> ``| `soshelp <command>` or `help <command>` |`` instead of ``| `soshelp|help <command>` |``

That makes sense to me -- at first glance I didn't even realize that's what they were using the pipe character for here -- it looked like it was part of the command. So I would prefer this approach even if we didn't have an issue with how the pipe characters were displaying.

gewarren:
Closing this in favor of #27622.

PR #27222, opened 2021-11-22
Structure types and with expression: fix incorrect language

### LINKED ISSUE #27177: Nondestructive mutation – not all immutable members can be changes using `with` expression
> Beginning with C# 10, you can use the with expression if you need to mutate immutable properties or fields of a structure-type instance. 

Immutable field cannot be subject to change. That is for sure. Immutable properties only if setter is specified. This narrows down to only option – `init-only` setter.

```csharp
// Immutable
struct WithField { public readonly int i = 3; }
struct WithProperty { public int I { get; } = 7; }
struct WithProperty_3 { public int I { get; init; } = 7; }

// Mutable
struct WithField_2 { public int i = 3; }
struct WithProperty_2 { public int I { get; set; } = 7; }

void Show()
{
  object x;
  // Error CS0191  A readonly field cannot be assigned to(except in a constructor or init - only setter of the type in which the field is defined or a variable initializer)
  x = new WithField() with { i = 5, };
  //  Error CS0200  Property or indexer 'WithProperty.I' cannot be assigned to -- it is read only	
  x = new WithProperty() with { I = 5, };
      
  x = new WithField_2() with { i = 3 }; // Ok
  x = new WithProperty_2() with { I = 3 }; // Ok
  x = new WithProperty_3() with { I = 3 }; // Ok
}
```
---
#### Document Details

* ID: d9220116-e950-04cf-b6d5-70d15a084a22
* Version Independent ID: 0960802e-80c3-cb56-cb8b-ab1f1b20fb4e
* Content: [Structure types - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record#nondestructive-mutation)
* Content Source: [docs/csharp/language-reference/builtin-types/struct.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/builtin-types/struct.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

pkulikov:
Fixes #27177

@BillWagner @Uzivatel919 I agree that the statement "mutate immutable properties or fields" is at least confusing, misleading, so I've removed it and simplified the language. At the same time, I don't think we need to specify that both mutable and init-only members can be updated only: I believe the compiler error messages do that well.

### DIFF
--- a/docs/csharp/language-reference/builtin-types/struct.md
+++ b/docs/csharp/language-reference/builtin-types/struct.md
@@ -73,7 +73,7 @@ The compiler may make use of the `readonly` modifier for performance optimizatio
 
 ## Nondestructive mutation
 
-Beginning with C# 10, you can use the [`with` expression](../operators/with-expression.md) if you need to mutate immutable properties or fields of a structure-type instance. A `with` expression makes a copy of its operand with specified properties and fields modified. You use [object initializer](../../programming-guide/classes-and-structs/object-and-collection-initializers.md) syntax to specify what members to modify and their new values, as the following example shows:
+Beginning with C# 10, you can use the [`with` expression](../operators/with-expression.md) to produce a copy of a structure-type instance with the specified properties and fields modified. You use [object initializer](../../programming-guide/classes-and-structs/object-and-collection-initializers.md) syntax to specify what members to modify and their new values, as the following example shows:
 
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
I like the wording here @pkulikov 

I'll :shipit: now.

PR #27223, opened 2021-11-22
Fix formatting

pkulikov:
An empty line after the list is essential. Current rendering:
image


BillWagner (approved):
Thanks @pkulikov 

This is an improvement.  I'll :shipit: now.

PR #27239, opened 2021-11-23
Reword confusing sentence

Youssef1313:
This part of sentence looks confusing to. The space is saved for any `namespace` declaration that's using this new syntax. What does "common" really mean here?

Will update the PR with another wording instead of removing this part if someone knows the original intent.


BillWagner (approved):
Thanks @Youssef1313 

I like this alternative. I think I wrote this sentence originally. If so, I was going for some wiggle room around those files that have multiple namespace declarations. I really don't think that's needed here. Your version is better.

I'll :shipit: now.

PR #27269, opened 2021-11-24
Update interpolated-string-handler.md

SteveDunn:
## Summary

Remove text assumed to be from a tutorial template


BillWagner (approved):
Thanks for spotting this and fixing it @SteveDunn 

I can't believe we missed it in our internal reviews. 😞 

PR #27281, opened 2021-11-25
Add using directive to HelloWorld.cs

rodrimc:
## Summary

The documentation has the following statement: "The "Hello, World" program starts with a `using` directive that references the System namespace". However the HelloWorld.cs is missing the `using` directive.

### DIFF
--- a/docs/csharp/tour-of-csharp/snippets/shared/HelloWorld.cs
+++ b/docs/csharp/tour-of-csharp/snippets/shared/HelloWorld.cs
@@ -1,4 +1,6 @@
-﻿class Hello
+using System;
+
+class Hello
 {
     static void Main()
     {


MSDN-WhiteKnight:
I'm pretty sure it's because of implicit usings in .NET 6: https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/6.0/implicit-namespaces-rc1 We need to update the tour text instead.

BillWagner:
This is a good conversation @rodrimc @MSDN-WhiteKnight 

I wasn't sure which direction to go in updating this for C# 10. I've been leaning toward the recommendation to update this to match the new template for C# 10 (global implicit usings, and top-level programs). I think that lets us introduce each concept one at a time. That would be a larger change, and it's in my plan for the month or two. In the meantime, this change is an improvement in correctness.

Thoughts?

rodrimc:
Considering this HelloWorld.cs is the very first C# code sample discussed in the "Tour of C# language" docs. I think it worth editing it to explicitly add the `using` directive for the sake of presenting the language to newcomers. It might also be worth adding a few words about implicit usings, maybe linking it to the page @MSDN-WhiteKnight pointed (https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/6.0/implicit-namespaces-rc1). 

If we explain the use of `usings` in the introduction and then briefly discuss about the implicit `using`s feature, then I think that's okay to skip adding `using`s in following code samples.

BTW, if you hit the `Run` button, the interactive window shows an error saying `The name 'Console' does not exist in the current context`. One can get rid of that error by explicitly adding `using System;`.

MSDN-WhiteKnight:
>BTW, if you hit the Run button, the interactive window shows an error saying The name 'Console' does not exist in the current context. One can get rid of that error by explicitly adding using System;.

If this does not actually work, i agree, it's better to fix it for now, and update for implicit usings later. Bad that interactive runner environment does not match snippets CI environment.

BillWagner (approved):
Thanks for the discussion.

I'll approve and merge this now. It does improve the page as it's currently presented. 

When I make the next set of major updates, I'll run tests on which order of topics works best for newcomers to the language. Then, I'll make the updates.

PR #27370, opened 2021-12-02
Update protobuf-data-types.md

VladMusidze:
Since DecimalValue is a partial class with its other part being defined in DecimalValue.proto properties `Units` and `Nanos` are already defined. Existing code will not compile

### DIFF
--- a/docs/architecture/grpc-for-wcf-developers/protobuf-data-types.md
+++ b/docs/architecture/grpc-for-wcf-developers/protobuf-data-types.md
@@ -166,9 +166,6 @@ namespace CustomTypes
             Nanos = nanos;
         }
 
-        public long Units { get; }
-        public int Nanos { get; }
-
         public static implicit operator decimal(CustomTypes.DecimalValue grpcDecimal)
         {
             return grpcDecimal.Units + grpcDecimal.Nanos / NanoFactor;


IEvangelist:
> Since DecimalValue is a partial class with its other part being defined in DecimalValue.proto properties `Units` and `Nanos` are already defined. Existing code will not compile

Hi @VladSabre, I'm not so sure. I do not see any other properties with those names, the `message DecimalValue` has two fields named `units` and `nanos`:

```protobuf
package CustomTypes;

// Example: 12345.6789 -> { units = 12345, nanos = 678900000 }
message DecimalValue {

    // Whole units part of the amount
    int64 units = 1;

    // Nano units of the amount (10^-9)
    // Must be same sign as units
    sfixed32 nanos = 2;
}
```
The .ctor for the `partial class DecimalValue` accepts these two parameters named `units` and `nanos`, and assigns them to the `get` only props - to me this is accurate and should compile:

```csharp
namespace CustomTypes
{
    public partial class DecimalValue
    {
        private const decimal NanoFactor = 1_000_000_000;
        public DecimalValue(long units, int nanos)
        {
            Units = units;
            Nanos = nanos;
        }

        public static implicit operator decimal(CustomTypes.DecimalValue grpcDecimal)
        {
            return grpcDecimal.Units + grpcDecimal.Nanos / NanoFactor;
        }

        public static implicit operator CustomTypes.DecimalValue(decimal value)
        {
            var units = decimal.ToInt64(value);
            var nanos = decimal.ToInt32((value - units) * NanoFactor);
            return new CustomTypes.DecimalValue(units, nanos);
        }
    }
}
```

IEvangelist (changes requested):
I'm not so sure that this proposed change is accurate, in fact; I believe it breaks to example.

VladSabre:
Hi @IEvangelist 
I believe I haven't made myself perfectly clear :)
The code you specified is indeed correct, and it also contains the changes I proposed

IEvangelist:
> Hi @IEvangelist I believe I haven't made myself perfectly clear :) The code you specified is indeed correct, and it also contains the changes I proposed

Oops, I copied the code incorrectly. I meant to include the properties that you removed. I'm not actually familiar enough with what's generated, I'll have to try this to ensure that you're correct.

IEvangelist:
I heard from @JamesNK - who confirmed this, thank you.

IEvangelist (approved):

PR #27376, opened 2021-12-02
Clean various build warnings and errors

IEvangelist:
## Summary

Clean various build warnings and errors

Related to #26984 and #26999

### DIFF
--- a/docs/fundamentals/code-analysis/quality-rules/snippets/csharp/all-rules/ca2208.cs
+++ b/docs/fundamentals/code-analysis/quality-rules/snippets/csharp/all-rules/ca2208.cs
@@ -2,7 +2,7 @@
 
 namespace ca2208
 {
-    #pragma warning disable CA2208 // Instantiate argument exceptions correctly
+#pragma warning disable CA2208 // Instantiate argument exceptions correctly
     //<snippet1>
     public class Book
     {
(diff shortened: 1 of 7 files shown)


docs/fundamentals/code-analysis/quality-rules/snippets/csharp/all-rules/ca2208.cs
```diff
@@ -2,6 +2,7 @@
 
 namespace ca2208
 {
+#pragma warning disable CA2208 // Instantiate argument exceptions correctly
```
gewarren, line 5: Oh interesting, for some reason, this is one rule we've explicitly turned on in the editorconfig: https://github.com/dotnet/docs/blob/266c099450b3b5433e04dbec73515ef1c9a330de/.editorconfig#L81. Actually there are new settings in .NET 6 where we can bulk enable security rules - I'll look into that.

gewarren (approved):

docs/fundamentals/code-analysis/quality-rules/snippets/csharp/all-rules/ca2208.cs
IEvangelist: It is an error, I think that's what we wanted. This PR adds an explicit disable/restore to remove the error from the example that shows the issue.

PR #27383, opened 2021-12-02
Add note about order of JSON objects

### LINKED ISSUE #18201: Polymorphic deserialization example fails on valid JSON
The Polymorphic deserialization example on the following url fails on valid JSON

> Title: How to write custom converters for JSON serialization - .NET
> URL: https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?view=netcore-3.0#feedback

In JSON, An object is an unordered set of name/value pairs (ref https://www.json.org/json-en.html). This means that the JSON from the example:
`[
  {
    "TypeDiscriminator": 1,
    "CreditLimit": 10000,
    "Name": "John"
  },
  {
    "TypeDiscriminator": 2,
    "OfficeNumber": "555-1234",
    "Name": "Nancy"
  }
]`
Is equivalent to the following:
`[
  {
    "CreditLimit": 10000,
    "TypeDiscriminator": 1,
    "Name": "John"
  },
  {
    "OfficeNumber": "555-1234",
    "TypeDiscriminator": 2,
    "Name": "Nancy"
  }
]`
This is because the order of the properties in the object does not matter in JSON. Javascript implementations are fully entitled to take in the first example and return the second example and expect it to be valid.

However the example code will throw an exception if the first property name is not "TypeDiscriminator". 

Please add a warning that the example code requires JSON object name/value pairs to stay in order even though this is not required in JSON, or amend the example.

---
#### Document Details

* Content: [How to write custom converters for JSON serialization - .NET](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?view=netcore-3.0#feedback)
* Product: **dotnet**
* GitHub Login: @tdykstra

tdykstra:
Fixes #18201

### DIFF
--- a/docs/standard/serialization/system-text-json-converters-how-to.md
+++ b/docs/standard/serialization/system-text-json-converters-how-to.md
@@ -322,7 +322,10 @@ Built-in features provide a limited range of [polymorphic serialization](system-
 
 Suppose, for example, you have a `Person` abstract base class, with `Employee` and `Customer` derived classes. Polymorphic deserialization means that at design time you can specify `Person` as the deserialization target, and `Customer` and `Employee` objects in the JSON are correctly deserialized at run time. During deserialization, you have to find clues that identify the required type in the JSON. The kinds of clues available vary with each scenario. For example, a discriminator property might be available or you might have to rely on the presence or absence of a particular property. The current release of `System.Text.Json` doesn't provide attributes to specify how to handle polymorphic deserialization scenarios, so custom converters are required.
 
-The following code shows a base class, two derived classes, and a custom converter for them. The converter uses a discriminator property to do polymorphic deserialization. The type discriminator isn't in the class definitions but is created during serialization and is read during deserialization. The example code requires JSON object name/value pairs to stay in order, which is not a standard requirement of JSON.
+The following code shows a base class, two derived classes, and a custom converter for them. The converter uses a discriminator property to do polymorphic deserialization. The type discriminator isn't in the class definitions but is created during serialization and is read during deserialization.
+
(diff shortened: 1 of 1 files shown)


docs/standard/serialization/system-text-json-converters-how-to.md
```diff
@@ -322,7 +322,7 @@ Built-in features provide a limited range of [polymorphic serialization](system-
...
 
-The following code shows a base class, two derived classes, and a custom converter for them. The converter uses a discriminator property to do polymorphic deserialization. The type discriminator isn't in the class definitions but is created during serialization and is read during deserialization.
+The following code shows a base class, two derived classes, and a custom converter for them. The converter uses a discriminator property to do polymorphic deserialization. The type discriminator isn't in the class definitions but is created during serialization and is read during deserialization. The example code requires JSON object name/value pairs to stay in order, which is not a standard requirement of JSON.
```
IEvangelist, line 325:
```suggestion
The following code shows a base class, two derived classes, and a custom converter for them. The converter uses a discriminator property to do polymorphic deserialization. The type discriminator isn't in the class definitions but is created during serialization and is read during deserialization.

> [!IMPORTANT]
> The example code requires JSON object name/value pairs to stay in order, which is not a standard requirement of JSON.
```

I'd promote this to an alert as it is very important to consumers. This will likely be a gotcha that readers will struggle with, so we should call more attention to it.

IEvangelist (approved):
Just a suggestion, otherwise :shipit:

tdykstra:
@IEvangelist Thanks for the reviews!

PR #27387, opened 2021-12-03
Test multi redirect files with docs-verifier

gewarren:


IEvangelist:
Hi @gewarren - how did the test go? Is everything working as expected? I wasn't sure if the tool that you updated recently correctly discovered all redirection files, that might be something we'd need to update if this doesn't work as expected. Let me know, and I can help with that too if needed.

gewarren:
> Hi @gewarren - how did the test go? Is everything working as expected? I wasn't sure if the tool that you updated recently correctly discovered all redirection files, that might be something we'd need to update if this doesn't work as expected. Let me know, and I can help with that too if needed.

This PR is a test for updating that tool (docs-verifier) - not for testing that multiple redirection files work in OPS. So I'm currently working on updating the tool. Getting there.

gewarren:
This is working as expected now.

PR #27398, opened 2021-12-03
Add ConfigureAwait(false) to DisposeAsync example

eerhardt:
This example calls `ConfigureAwait(false)` in `DisposeAsyncCore()`, but doesn't in `DisposeAsync`. The example should consistently use `ConfigureAwait`.


gewarren (approved):

IEvangelist (approved):

PR #27409, opened 2021-12-05
Reader positioning in STJ converters

### LINKED ISSUE #18763: Add an example when actual object you want to deserialize is wrapped in non essential layers
I had hard time deserializing the list of objects wrapped in several non-essential layers. I couldn't understand why I was getting `JsonException` with "read too much, or not enough" message (more details [here](https://stackoverflow.com/questions/62147178/system-text-json-deserialization-fails-with-jsonexception-read-to-much-or-not-e)).

I finally got an explanation on [Stackoverflow](https://stackoverflow.com/a/62155881/761755), but I believe the requirement "When reading an object, JsonConverter<T>.Read() must leave the Utf8JsonReader positioned on the EndObject token of the object where it was originally positioned." should be added to this page to help other developers, when they face similar issue.

---
#### Document Details

* Content: [How to write custom converters for JSON serialization - .NET](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to)
* Product: **dotnet**
* GitHub Login: @tdykstra

### LINKED ISSUE #21147: Not clear that when creating a custom converter Utf8JsonReader should exit Read method positioned on EndArray/EndObject token and with correct depth
It is worth mentioning that upon exiting Read method Utf8JsonReader's TokenType & CurrentDepth should follow some rules depending on TokenType/Depth of the reader upon entering the method & underlying json structure (object/array vs simple property)

E.g. given this json
```
{
   "property1":"value1",
   "property2":{
      "nestedLevel1":{
         "nestedLevel2":"nestedLevelValue2"
      }
   }
}
```
and the model to deserialize to

```
public class Model
{
   public string Property1 {get; set;}

   [JsonPropertyName("property2")
   [JsonConverter(typeof(NestedLevel2Converter))]
   public string NestedLevel2Property {get; set;}
}
```
And the converter itself

```
public sealed class NestedLevel2Converter: JsonConverter<string>
    {
        public override string Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
        {
            throw new NotImplementedException();
            // upon existing we should have reader.TokenType == JsonTokenType.EndObject
            // && reader.CurrentDepth == depth of the reader upon entering Read method
            // https://github.com/dotnet/runtime/blob/98fc7edd7fbfba0e0120dbd4533b1ad7c73cbb7d/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonConverterOfT.cs#L439
        }

        public override void Write(Utf8JsonWriter writer, string value, JsonSerializerOptions options)
        {
            throw new NotImplementedException();
        }
    }
```

As mentioned above, as per https://github.com/dotnet/runtime/blob/98fc7edd7fbfba0e0120dbd4533b1ad7c73cbb7d/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonConverterOfT.cs#L439 when Read method exits Utf8JsonReader has to have specific state.
Documentation doesn't mention this, and it's not clear that this rule should be followed to avoid JsonException _read too much or not enough_ exception for this case.

---
#### Document Details

* Content: [How to write custom converters for JSON serialization - .NET](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to#other-custom-converter-samples)
* Product: **dotnet**
* GitHub Login: @tdykstra

tdykstra:
Fixes #18763
Fixes #21147

### DIFF
--- a/docs/standard/serialization/system-text-json-converters-how-to.md
+++ b/docs/standard/serialization/system-text-json-converters-how-to.md
@@ -1,7 +1,7 @@
 ---
 title: "How to write custom converters for JSON serialization - .NET"
 description: "Learn how to create custom converters for the JSON serialization classes that are provided in the System.Text.Json namespace."
-ms.date: 08/19/2021
+ms.date: 12/04/2021
 no-loc: [System.Text.Json, Newtonsoft.Json]
 zone_pivot_groups: dotnet-version
 helpviewer_keywords: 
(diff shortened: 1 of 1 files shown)


gewarren (approved):
The `Stack<T>` link text won't be formatted as code, just FYI.

tdykstra:
Thanks for the review! I'm using code style there to prevent Stack from being localized and to avoid the open angle bracket from being interpreted as the start of an HTML tag.

PR #27410, opened 2021-12-05
C# reference: Consolidated jump statements

### LINKED ISSUE #10746: Abridged Reference: Statements / selection and jump statements

pkulikov:
Fixes #10746

✔️ [Internal preview: Jump statements (C# reference)](https://review.docs.microsoft.com/en-us/dotnet/csharp/language-reference/statements/jump-statements?branch=pr-en-us-27410)


docs/csharp/language-reference/statements/jump-statements.md
```diff
@@ -0,0 +1,114 @@
...
+  - "goto keyword [C#]"
+---
+# Jump statements (C# reference)
+
+The following statements unconditionally transfer control:
+
+- The [`break` statement](#the-break-statement): terminates the closest enclosing [iteration statement](iteration-statements.md) or [`switch` statement](selection-statements.md#the-switch-statement).
+- The [`continue` statement](#the-continue-statement): starts a new iteration of the closest enclosing [iteration statement](iteration-statements.md).
+- The [`return` statement](#the-return-statement): terminates execution of the function in which it appears and returns control to the caller.
+- The [`goto` statement](#the-goto-statement): transfers control to a statement that is marked by a label.
```
pkulikov, line 27: The [spec about jump statements](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/statements#jump-statements) also includes the `throw` statement in the list of jump statements. Should we cover `throw` in this article (and then to have the separate article only for `try` and its variants)?

docs/csharp/language-reference/statements/jump-statements.md
```diff
@@ -0,0 +1,114 @@
...
+  :::code language="csharp" interactive="try-dotnet-method" source="snippets/jump-statements/ReturnStatement.cs" id="WithoutExpression":::
+
+  You can use the `return` statement without expression in a function that doesn't compute a value, that is:
+
+  - a [method](../../programming-guide/classes-and-structs/methods.md), [local function](../../programming-guide/classes-and-structs/local-functions.md), or [lambda expression](../operators/lambda-expressions.md) with the result type [`void`](../builtin-types/void.md)
+  - an [async](../keywords/async.md) method, local function, or lambda expression with the result type <xref:System.Threading.Tasks.Task> or <xref:System.Threading.Tasks.ValueTask>
+  - the `set` accessor of a [property](../../programming-guide/classes-and-structs/properties.md) or an [indexer](../../programming-guide/indexers/index.md)
+  - a [constructor](../../programming-guide/classes-and-structs/constructors.md)
+  - a [finalizer](../../programming-guide/classes-and-structs/finalizers.md)
+  - the `add` and `remove` accessors of an [event](../../programming-guide/events/index.md)
```
pkulikov, line 64: This list is very spec-like. Alternatives:
(1) Remove it at all, leave only the introduction sentence
(2) Shorten it, make it not exhaustive, that is, mention only method and properties as examples
(3) Do nothing, leave as it is
(4) Something else

@BillWagner what do you think?

pkulikov (commented):
@BillWagner I still have several questions on how to continue. Other than those, it's ready for review.

docs/csharp/language-reference/statements/jump-statements.md
IEvangelist: I would expect to see `throw` here, as it in my mind is a jump statement. But I defer to @BillWagner

docs/csharp/language-reference/statements/jump-statements.md
```diff
@@ -0,0 +1,114 @@
...
+  You use the `return` statement with expression in a function that computes a value, that is:
+
+  - a [method](../../programming-guide/classes-and-structs/methods.md), [local function](../../programming-guide/classes-and-structs/local-functions.md), or [lambda expression](../operators/lambda-expressions.md) with a non-void result type
+  - an [async](../keywords/async.md) method, local function, or lambda expression with the result type <xref:System.Threading.Tasks.Task%601> or <xref:System.Threading.Tasks.ValueTask%601>
+  - the `get` accessor of a [property](../../programming-guide/classes-and-structs/properties.md) or an [indexer](../../programming-guide/indexers/index.md)
+  - an [overloaded operator](../operators/operator-overloading.md)
+
+  By default, the `return` statement returns the value of an expression. Beginning with C# 7.0, you can return a reference to a variable. To do that, use the `return` statement with the [`ref` keyword](../keywords/ref.md), as the following example shows:
+
+  :::code language="csharp" source="snippets/jump-statements/ReturnStatement.cs" id="RefReturn":::
```
IEvangelist, line 81:
```suggestion
  :::code language="csharp" interactive="try-dotnet-method" source="snippets/jump-statements/ReturnStatement.cs" id="RefReturn":::
```

I verified that this one does work in the interactive experience.

docs/csharp/language-reference/statements/jump-statements.md
```diff
@@ -0,0 +1,114 @@
...
+- With expression, as in the following example:
+
+  :::code language="csharp" interactive="try-dotnet-method" source="snippets/jump-statements/ReturnStatement.cs" id="WithExpression":::
+
+  You use the `return` statement with expression in a function that computes a value, that is:
+
+  - a [method](../../programming-guide/classes-and-structs/methods.md), [local function](../../programming-guide/classes-and-structs/local-functions.md), or [lambda expression](../operators/lambda-expressions.md) with a non-void result type
+  - an [async](../keywords/async.md) method, local function, or lambda expression with the result type <xref:System.Threading.Tasks.Task%601> or <xref:System.Threading.Tasks.ValueTask%601>
+  - the `get` accessor of a [property](../../programming-guide/classes-and-structs/properties.md) or an [indexer](../../programming-guide/indexers/index.md)
+  - an [overloaded operator](../operators/operator-overloading.md)
```
IEvangelist, line 77:
```suggestion
  - A [method](../../programming-guide/classes-and-structs/methods.md), [local function](../../programming-guide/classes-and-structs/local-functions.md), or [lambda expression](../operators/lambda-expressions.md) with a non-void result type.
  - An [async](../keywords/async.md) method, local function, or lambda expression with the result type <xref:System.Threading.Tasks.Task%601> or <xref:System.Threading.Tasks.ValueTask%601>.
  - The `get` accessor of a [property](../../programming-guide/classes-and-structs/properties.md) or an [indexer](../../programming-guide/indexers/index.md).
  - An [overloaded operator](../operators/operator-overloading.md).
```

I think we'd want sentences for the list here.

docs/csharp/language-reference/statements/jump-statements.md
```diff
@@ -0,0 +1,114 @@
...
+## The `goto` statement
+
+The `goto` statement transfers control to a statement that is marked by a label, as the following example shows:
+
+:::code language="csharp" interactive="try-dotnet-method" source="snippets/jump-statements/GotoStatement.cs" id="NestedLoops":::
+
+As the preceding example shows, you can use the `goto` statement to get out of a nested loop.
+
+> [!TIP]
+> When you work with nested loops, consider refactoring separate loops into separate methods. That may lead to a simpler, more readable code without the `goto` statement.
```
IEvangelist, line 94: Indeed, a `goto` is often a code smell.

IEvangelist (commented):
This looks great, I'll let @BillWagner approve and merge. I had a few suggestions, but overall, this is great.

docs/csharp/language-reference/statements/jump-statements.md
```diff
@@ -0,0 +1,114 @@
...
+  - "goto statement [C#]"
+  - "goto keyword [C#]"
+---
+# Jump statements (C# reference)
+
+The following statements unconditionally transfer control:
+
+- The [`break` statement](#the-break-statement): terminates the closest enclosing [iteration statement](iteration-statements.md) or [`switch` statement](selection-statements.md#the-switch-statement).
+- The [`continue` statement](#the-continue-statement): starts a new iteration of the closest enclosing [iteration statement](iteration-statements.md).
+- The [`return` statement](#the-return-statement): terminates execution of the function in which it appears and returns control to the caller.
```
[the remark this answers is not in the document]
pkulikov: Is there any actual feedback from readers about this possible confusion? The current reference article doesn't have any such feedback. Spec, which is usually very specific, doesn't mention that topic. I think exiting method clearly implies exiting a loop. I suggest to keep it simple.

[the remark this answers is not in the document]
pkulikov: Also, the ref return example has the `return` statement inside a loop. I guess that may be enough?

docs/csharp/language-reference/statements/jump-statements.md
BillWagner: I thought about this for a while. There are good arguments on both sides.  In the end, I think it will be more clear to link to an article about `throw`, `try`, `catch` and `finally` (and probably exception filters) from this article than to cover `throw` here.

My reasoning is that `throw` is strongly related to `try`, `catch`, and `finally`. 

The `throw` statement is a jump statement in the spec. But what about the `throw` expression? It's not even a *statement*. If I were looking for `throw`, I'd expect to find information on both the *statement* and the *expression*. (Although, I think they will stay separate in the spec because of the 1st level headings)

docs/csharp/language-reference/statements/jump-statements.md
```diff
@@ -0,0 +1,114 @@
...
+- The [`goto` statement](#the-goto-statement): transfers control to a statement that is marked by a label.
+
+## The `break` statement
```
BillWagner, line 29: There are a few other cases here that could be explored. This may be another case where not qualifying a statement can be clarifying. I'm not sure yet.

- What about nested `switch` statements? A `break` only exits the inner-most switch. 
- What about a `break` inside a loop in the statements for a *switch-condition*. The `break` will exit the loop, but not the `switch` statement.
- Any level of nesting of those conditions, among others.

I don't have alternative, and it may not be necessary, but it's worth thinking through. From customer questions, the nested loops example *is* important.

docs/csharp/language-reference/statements/jump-statements.md
BillWagner: I think it will be easier to read using the introductory sentence structure. Consider something like my suggestion on line 51.

docs/csharp/language-reference/statements/jump-statements.md
```diff
@@ -0,0 +1,114 @@
...
+
+## The `continue` statement
+
+The `continue` statement starts a new iteration of the closest enclosing [iteration statement](iteration-statements.md) (that is, `for`, `foreach`, `while`, or `do` loop), as the following example shows:
+
+:::code language="csharp" interactive="try-dotnet-method" source="snippets/jump-statements/ContinueStatement.cs" id="BasicExample":::
+
+## The `return` statement
+
+The `return` statement terminates execution of the function in which it appears and returns control and the function's result, if any, to the caller. The `return` statement can be any of the following two forms:
```
BillWagner, line 51: Regarding the next question, I might consider something like this:

```suggestion
The `return` statement terminates execution of the function in which it appears and returns control and the function's result, if any, to the caller. The `return` statement has an expression following the `return` keyword when the member produce a value. The `return` statement must not have an expression when the member doesn't prodcue a value. When the `return` statement has an expression, that expression must be convertible to the type returned by the member. The only exception to this rule is `async` functions, where the expression must be convertible to the type argument for <xref:System.Threading.Tasks.Task%601?displayProperty=nameWithType>. Return statements in `async` methods that return <xref:System.Threading.Tasks.Task?displayProperty=nameWithType> must not have an expression.
```

docs/csharp/language-reference/statements/jump-statements.md
```diff
@@ -0,0 +1,114 @@
...
+  :::code language="csharp" interactive="try-dotnet-method" source="snippets/jump-statements/ReturnStatement.cs" id="WithExpression":::
+
+  You use the `return` statement with expression in a function that computes a value, that is:
+
+  - a [method](../../programming-guide/classes-and-structs/methods.md), [local function](../../programming-guide/classes-and-structs/local-functions.md), or [lambda expression](../operators/lambda-expressions.md) with a non-void result type
+  - an [async](../keywords/async.md) method, local function, or lambda expression with the result type <xref:System.Threading.Tasks.Task%601> or <xref:System.Threading.Tasks.ValueTask%601>
+  - the `get` accessor of a [property](../../programming-guide/classes-and-structs/properties.md) or an [indexer](../../programming-guide/indexers/index.md)
+  - an [overloaded operator](../operators/operator-overloading.md)
+
+  By default, the `return` statement returns the value of an expression. Beginning with C# 7.0, you can return a reference to a variable. To do that, use the `return` statement with the [`ref` keyword](../keywords/ref.md), as the following example shows:
```
BillWagner, line 79: Is it worth mentioning that `ref return` is not valid in an `async` method?

docs/csharp/language-reference/statements/jump-statements.md
```diff
@@ -0,0 +1,114 @@
...
+
+  :::code language="csharp" interactive="try-dotnet-method" source="snippets/jump-statements/ReturnStatement.cs" id="RefReturn":::
+
+  For more information about ref returns, see [Ref returns and ref locals](../../programming-guide/classes-and-structs/ref-returns.md).
+
+## The `goto` statement
```
BillWagner, line 85: I think it's worth mentioning that there are restrictions on the target label of a `goto` statement.

The [standard](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/statements.md#13104-the-goto-statement) has some very elegant language to define the restriction:

> If a label with the given name does not exist in the current function member, or if the `goto` statement is not within the scope of the label, a compile-time error occurs.

This rule means you can't use `goto` to transfer control *into* a `try` block, `catch` block, `finally` block, or other nested scope.

It might be sufficient to say you can't use `goto` to transfer control out of the current member, or into any nested scope, such as a `try` block.

BillWagner (commented):
Overall, this looks great @pkulikov 

I think I answered all your questions, and raised a few of my own. Consider all my questions as "items for discussion". I'm not sure how best to resolve each point, but I wanted to make sure we considered them.

docs/csharp/language-reference/statements/jump-statements.md
pkulikov: @BillWagner thank you; so, I've only added the link to the article about `throw` for now. (Later, I'll consolidate all exception-related statements into one article.)

docs/csharp/language-reference/statements/jump-statements.md
pkulikov: From the current feedback to the `break` article, I've concluded that nested loops and `switch` inside a loop are worth mentioning. As for other variants, I assume (1) they are less frequent (and maybe should be refactored), (2) may be reasoned about by analogy from the existing examples. So, I don't think the article should explore other cases. What do you think @BillWagner?

docs/csharp/language-reference/statements/jump-statements.md
pkulikov: @BillWagner thank you, I've rewritten this section based on your suggestion. Please review it again.

docs/csharp/language-reference/statements/jump-statements.md
pkulikov: I don't think so, as I've added the link to the article about ref returns just after the example.

docs/csharp/language-reference/statements/jump-statements.md
pkulikov: @BillWagner thanks, I've updated the section accordingly.

pkulikov:
@BillWagner thank you. I think I've addressed all of your feedback, please review the new commits.
@IEvangelist thank you for your review too.

BillWagner (approved):
Thanks again @pkulikov 

I really like the final version.

I'll :shipit: now.

PR #27466, opened 2021-12-09
Remove reference to absent code

vishallama:
## Summary

There is a reference to `Seq.toList` in the current doc, which no longer seems necessary. The said reference seems part of code in a previous version of the doc. Removing it is safe, and it will make the doc more up-to-date.


BillWagner (approved):
Thanks for making this update @vishallama 

Thanks again!

PR #27475, opened 2021-12-09
Fix snippet interactivity

pkulikov:
When published, this interactive snippet requires `using System;` directive. So, I've just moved the code to the separate file and added that directive.

This is about the last snippet in the following section (now it also misses the last closing brace):
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/statements/jump-statements#the-goto-statement


BillWagner (commented):
Thanks again @pkulikov 

This is a great change, and I'll :shipit: now.

BillWagner (approved):

PR #27476, opened 2021-12-09
Disable snippet interactivity

pkulikov:
Another issue with an interactive snippet:
image

This is the first snippet in the following section:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/statements/jump-statements#the-goto-statement

If you prefer interactivity over the KeyValuePair deconstruction, please let me know: I'll refactor the loop then.

In general, is there any work done/planned on improving the interactive experience in the docs, like this or supporting C# 8 and later?

/cc: @IEvangelist

### DIFF
--- a/docs/csharp/language-reference/statements/jump-statements.md
+++ b/docs/csharp/language-reference/statements/jump-statements.md
@@ -74,7 +74,7 @@ For more information about ref returns, see [Ref returns and ref locals](../../p
 
 The `goto` statement transfers control to a statement that is marked by a label, as the following example shows:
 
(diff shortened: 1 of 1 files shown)


IEvangelist:
Hi @pkulikov, I too prefer the interactivity. I'm honestly not sure about the plans for that functionality, I'm certain @BillWagner would know.

BillWagner (approved):
Thanks @pkulikov 

I agree that disabling interactivity for now is the right choice.

The team is working on the plan to update C# interactive to match fidelity with the latest C# features. I don't have a deadline for that, or which features we can expect first.

PR #27505, opened 2021-12-11
Add macOS 12 Monterey to Documentation

say25:
## Summary

Adds macOS 12 to the Documentation as it is a supported runtime https://github.com/dotnet/runtime/blob/4fb7f773f3df89411216ff658cfc85dc397f5f5a/src/libraries/Microsoft.NETCore.Platforms/src/runtime.json#L2136-L2152


IEvangelist (approved):

PR #27514, opened 2021-12-12
Update ms.date and VS screenshots

ardalis:


IEvangelist (approved):

PR #27516, opened 2021-12-12
Shift operators: Make shift count info more visible

### LINKED ISSUE #27486: Need to specify that left shift is modulo number of bits in the left operand
Currently, this documentation doesn't specify what happens when you left shift a number more than the number of bits in the number, e.g., left shift a 64 bit long 66 times. I was surprised to learn that this doesn't just keep shifting in 0s from the right 66 times but rather has the same effect as shifting the number 2 times, i.e., if x is a 64 bit long then x << 66 == x<<2. The documentation should call that out.

BTW: This is also left out in the Microsoft C++ documentation but the C++ language standard actually requires that the shift is modulo number of bits in the left operand.

Regards,
Martin

---
#### Document Details

* ID: 02b8f7c2-379e-7876-0d36-25b06a294baa
* Version Independent ID: 2410c828-0c66-9efe-ae87-9b07708aa731
* Content: [Bitwise and shift operators - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators#feedback)
* Content Source: [docs/csharp/language-reference/operators/bitwise-and-shift-operators.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/operators/bitwise-and-shift-operators.md)
* Product: **dotnet-csharp**
* GitHub Login: @pkulikov
* Microsoft Alias: **wiwagn**

pkulikov:
Fixes #27486

### DIFF
--- a/docs/csharp/language-reference/operators/bitwise-and-shift-operators.md
+++ b/docs/csharp/language-reference/operators/bitwise-and-shift-operators.md
@@ -56,7 +56,7 @@ You can also use the `~` symbol to declare finalizers. For more information, see
 
 ## Left-shift operator \<\<
 
-The `<<` operator shifts its left-hand operand left by the [number of bits defined by its right-hand operand](#shift-count-of-the-shift-operators).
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks @pkulikov 

This is great. I'll :shipit: now.

PR #27517, opened 2021-12-12
Updated the workdir for UNIX

belyaevsa:
## Summary

The DOTNET_ROOT dir after using the dotnet_install.sh script for Unix has changed to `$HOME/.dotnet` (with a dot in the beginning). Changed the TIP area


IEvangelist (approved):
This looks good to me, I'm going to let @adegeo confirm this though before merging.

adegeo (approved):

adegeo:
This is indeed correct. Thank you for spotting and fixing this!

PR #27525, opened 2021-12-13
Add important alert about user secrets usage

### LINKED ISSUE #27488: Add a note that user secrets will not be loaded when using the worker service sdk
Due to a bug or missing feature user secrets aren't loaded unless your startup project uses the web sdk or directly references the `Microsoft.Extensions.Configuration.UserSecrets` package. [This is a known issue](https://github.com/dotnet/extensions/issues/2743#issuecomment-562264614) since at least .net core 3.1 and still exists in .net 6.

The worker service template sets up user secrets but doesn't reference this package. At the very least a note should be added to this page, and probably the [user secrets page](https://docs.microsoft.com/en-us/aspnet/core/security/app-secrets), mentioning this so users don't waste their time trying to figure out why it doesn't work. It'd also be nice if the template could be updated with this workaround since the issue has been around for 3 versions now with no fix in sight.

---
#### Document Details

* ID: ae847f16-ff74-ab68-7163-f713b447430c
* Version Independent ID: ff379d26-0d85-6d74-67da-4a3cd260b04d
* Content: [Worker Services in .NET](https://docs.microsoft.com/en-us/dotnet/core/extensions/workers)
* Content Source: [docs/core/extensions/workers.md](https://github.com/dotnet/docs/blob/main/docs/core/extensions/workers.md)
* GitHub Login: @IEvangelist
* Microsoft Alias: **dapine**

IEvangelist:
## Summary

Add important alert about user secrets usage, update image to .NET 6, and fixes #27488

### DIFF
--- a/docs/core/extensions/workers.md
+++ b/docs/core/extensions/workers.md
@@ -3,7 +3,7 @@ title: Worker Services in .NET
 description: Learn how to implement a custom IHostedService and use existing implementations with .NET.
 author: IEvangelist
 ms.author: dapine
-ms.date: 11/15/2021
+ms.date: 12/13/2021
 ms.topic: overview
 ---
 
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @IEvangelist 

You can :shipit: when ready.

PR #27526, opened 2021-12-13
Language version consistency (9 -> 9.0)

Youssef1313:
This is consistent with the rest of the article. I noticed this from:

image


BillWagner:
Thanks @Youssef1313 

I've got a question for @MadsTorgersen to make sure we should standardize on "9" vs. "9.0". As soon as I get the answer, I'll either merge this or ask for changes.

BillWagner:
@Youssef1313 

I'm closing this one. After chatting with @MadsTorgersen and @BethMassi the correct fix would be to standardize on "9" instead of "9.0".

Thanks.

Youssef1313:
@BillWagner fwiw, the compiler actually uses C# 9.0, for example in:

https://github.com/dotnet/roslyn/blob/8c77f529a5981819f15a68deb616b6e04ad8d1aa/src/Compilers/CSharp/Portable/CSharpResources.resx#L6452 should it be updated as well?

Youssef1313:
Also in 

https://github.com/dotnet/roslyn/blob/8c77f529a5981819f15a68deb616b6e04ad8d1aa/src/Compilers/CSharp/Portable/LanguageVersion.cs#L346-L347

BillWagner:
@Youssef1313 I would ask @jaredpar if those changes are above the bar.

PR #27604, opened 2021-12-16
Lambda expressions: More attribute examples and style edits

pkulikov:
Added example of attributes applied to the input parameters and a return value.
Also, the last paragraph of the **Attributes** section looks to be important enough for a highlight.
Furthermore, removed future tense and reduced amount of passive voice.

### DIFF
--- a/docs/csharp/language-reference/operators/lambda-expressions.md
+++ b/docs/csharp/language-reference/operators/lambda-expressions.md
@@ -207,24 +207,24 @@ The general rules for type inference for lambdas are as follows:
 - Each input parameter in the lambda must be implicitly convertible to its corresponding delegate parameter.
 - The return value of the lambda (if any) must be implicitly convertible to the delegate's return type.
   
-## Natural type for lambda expressions
+## Natural type of a lambda expression
 
-Lambda expressions in themselves don't have a type because the common type system has no intrinsic concept of "lambda expression." However, it's sometimes convenient to speak informally of the "type" of a lambda expression. That informal "type" refers to the delegate type or <xref:System.Linq.Expressions.Expression> type to which the lambda expression is converted.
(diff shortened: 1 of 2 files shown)


IEvangelist (approved):
Looks good to me, we'll :shipit: when @BillWagner sees this and is ready. Thank you

IEvangelist:
Hey, @pkulikov - I think you may have broken a bookmark as part of this PR:

**Line 82: [Warning]** Cannot find bookmark `#natural-type-for-lambda-expressions` in:
_docs/csharp/language-reference/operators/lambda-expressions.md_ 
> did you mean `#natural-type-of-a-lambda-expression`?

**Line 83: [Warning]** Cannot find bookmark `#declared-return-type` in:
_docs/csharp/language-reference/operators/lambda-expressions.md_
> did you mean `#explicit-return-type`?

docs/csharp/language-reference/operators/lambda-expressions.md
```diff
@@ -250,32 +250,40 @@ The compiler can't infer a parameter type for `s`. When the compiler can't infer
...
+
+```csharp
+var sum = ([Example(1)] int a, [Example(2), Example(3)] int b) => a + b;
+var inc = [return: Example(1)] (int s) => s++;
+```
+
+As the preceding examples show, you must parenthesize the input parameters when you add attributes to a lambda expression or its parameters.
 
-Lambda expressions are invoked through the underlying delegate type. That is different than methods and local functions. The delegate's `Invoke` method won't check attributes on the lambda expression. Attributes don't have any effect when the lambda expression is invoked. Attributes on lambda expressions are useful for code analysis, and can be discovered via reflection. One consequence of this decision is the <xref:System.Diagnostics.ConditionalAttribute?displayProperty=nameWithType> cannot be applied to a lambda expression.
+> [!NOTE]
```
BillWagner, line 285: I agree this should be called out. Maybe change from Note to Important?

```suggestion
> [!IMPORTANT]
```

BillWagner (commented):
I'll agree this is ready to go, once the warnings are resolved. (I also think it's due to the changed anchor tags).

I had one optional suggestion.

Thanks again @pkulikov

pkulikov:
@IEvangelist thank you!

PR #27607, opened 2021-12-16
Edit pass record reference

### LINKED ISSUE #27149: Records (C# reference)|Positional syntax for property definition – compiler action taken list is wrong
> When you use the positional syntax for property definition, the compiler creates:
> - A public init-only auto-implemented property for each positional parameter provided in the record declaration.
>     - For record types and readonly record struct types: An init-only property can only be set in the constructor or by using a property initializer.
>    - For record struct types: A read-write property that can be set in a constructor, property initializer, or assignment after construction.
> - A primary constructor whose parameters match the positional parameters on the record declaration.
> - For record struct types, a parameterless constructor that sets each field to its default value.
> - A Deconstruct method with an out parameter for each positional parameter provided in the record declaration. This method is provided only if there are two or more positional parameters. The method deconstructs properties defined by using positional syntax; it ignores properties that are defined by using standard property syntax.

That _init-only_ is there extra. See point 2 in original text.

`Init-only` property can be set in constructor or by property initializer in all cases not just on record. This should be reworded somehow. Likely just noting `init-only` choice for autogenerated properties.

Again this is normal behavior of (R/W) property. Basically the same as # 2.

In contrast to # 5 why there are only noted parameters?

A lie. Such ctor is not generated even for `non-record` `struct`.

A lie. All `class`, `struct` and `readonly struct` `records` get generated `Deconstruct` method even for just 1 positional parameter.

---
#### Document Details

* Content: [Records - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

### LINKED ISSUE #27154: Positional syntax for property definition – generated ctor does not use explicitly declared property automatically
> If the generated auto-implemented property definition isn't what you want, you can define your own property of the same name. If you do that, the generated constructor and deconstructor will use your property definition.

No, it will not be used by generated .ctor. 

```csharp
> record Rec(int Int, string String, double Double)
. {
.     public int Int { get; } = int.Parse(String);
.     public string String { get; } = Int.ToString();
.     public double Double { get; }
. }
> new Rec(1, "3", 4.4).ToString()
"Rec { Int = 3, String = 1, Double = 0 }"
```

It also gets some kind of formula few words later.

>  You must add the explicit assignment of the declared property to its corresponding positional parameter.

This sentence is also of course wrong since source-destination dependency is exactly opposite.

This whole paragraph deservers rewording. For instance

> For instance, the following example declares the FirstName and LastName properties of a positional record public, but restricts the Id positional parameter to internal. 

becomes little esoteric.

---

Example bellow is also little beside point when showing some output and `Main` not relevant to explicit property declarations.

Part with deconstructor is correct.
---
#### Document Details

* Content: [Records - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record#positional-syntax-for-property-definition)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

### LINKED ISSUE #27157: Nondestructive mutation – space for improvements on `with` with `record struct`
> The compiler doesn't synthesize a copy constructor for record struct types. You can write one, but the compiler won't generate calls to it for with expressions. Instead, the compiler uses assignment.

I guess nobody can nick what exactly is _assginment_ in sentence “Instead, the compiler uses assignment.”. I am for something like _Compiler instead employs value type characteristics and assigns instance to new variable effectively making copy. Then values are assigned._

---
#### Document Details

* Content: [Records - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record#nondestructive-mutation)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

### LINKED ISSUE #27158: Built-in formatting for display – reference types same as value types *print* result of `ToString` method call
> For reference types, the type name of the object that the property refers to is displayed instead of the property value. In the following example, the array is a reference type, so System.String[] is displayed instead of the actual array element values:

`Person { FirstName = Nancy, LastName = Davolio, ChildNames = System.String[] }`

---

Above is complete nonsense. See C# interactive sample.

```csharp
> class ClassWithOveriddenToString
{
  public override string ToString() => "Hi, the type name is not available right now. But I’m an object definitely!";
}
> struct Struct { }
> record struct RecStru(ClassWithOveriddenToString obj, Struct val);
> { Write(new RecStru(new ClassWithOveriddenToString(), new Struct()).ToString()); }
"RecStru { obj = Hi, the type name is not available right now. But I’m an object definitely!, val = Submission#28+Struct }"
```
So statement from documentation is fairly far away from truth.

---
#### Document Details

* Content: [Records - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record#built-in-formatting-for-display)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

### LINKED ISSUE #27171: Equality in inheritance hierarchies – `EqualityContract` property is not `sealed` for `sealed` `record`
> If the record type is sealed, this property is sealed.

When `record` is `sealed`, `EqualityContract` is not `sealed`. Whole `type` is `sealed` thus all members are free of this burden.

```csharp
sealed record SealedRecord { }

// Decompilation

internal sealed record SealedRecord
{
  private Type EqualityContract { [CompilerGenerated] get { return typeof(SealedRecord); }}
  public override string ToString() { … }
  private bool PrintMembers(StringBuilder builder) { return false; }
  public override int GetHashCode() { return EqualityComparer<Type>.Default.GetHashCode(EqualityContract); }
  public bool Equals(SealedRecord? other) { return (object)this == other || ((object)other != null && EqualityContract == other!.EqualityContract); }
  private SealedRecord(SealedRecord original) { }
  public SealedRecord() { }
}
```
---
#### Document Details

* Content: [Records - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record#equality-in-inheritance-hierarchies)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

### LINKED ISSUE #27173: PrintMembers formatting in derived records – sealing method in sealed types has no use
> For a sealed record that derives from another record: protected sealed override bool PrintMembers(StringBuilder builder);

When type (record) is `sealed` cannot be inherited thus sealing its members has no use. So correct signature is `protected override bool PrintMembers(StringBuilder builder)`;

---
#### Document Details

* Content: [Records - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record#printmembers-formatting-in-derived-records)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

### LINKED ISSUE #27174: PrintMembers formatting in derived records – senseless note
>  Note
> In C# 10 and later, the compiler will synthesize PrintMembers when a base record has sealed the ToString method. You can also create your own implementation of PrintMembers.

Before C# 10 it is not possible to seal `ToString`. `PrintMembers` is created all the time. It is once told just fine some line above.

> In C# 10 and later, your implementation of ToString may include the sealed modifier, which prevents the compiler from synthesizing a ToString implementation for any derived records. Effectively, that means the ToString output won't include the runtime type information. 

Maybe author was about to note this new `ToString` sealing feature. This note is just intermix of some information.
 
---
#### Document Details

* Content: [Records - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record#printmembers-formatting-in-derived-records)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixes #27149 

- In addition to the notes on that issues, emphasize the difference in the behavior of the synthesized parameterless constructor based on the presence of field initializers and other instance constructors.

Fixes #27154

- Fix the description of positional parameters where you declare the property declaration in your source.

Fixes #27157 

- Reword sentence on `with` expressions and assignment. I kept this somewhat vague on how fields or properties are assigned, because the compiler may perform optimizations such as blitting the memory for the struct.

Fixes #27158 

- Correctly that the string for each property is obtained by calling `ToString` on that property type. That may often print the type name, but that depends on the type.

Fixes #27171 

- Clarify that the property is `sealed` because the enclosing type is `sealed`. (The property declaration need not include the `sealed` modifier).

Fixes #27173 

- The method signature doesn't need the `sealed` modifier because the type is sealed.

Fixes #27174 

- Add motivation for sealing ToString in records.


BillWagner:
Each fix is in a separate commit for easier review.

Ping @adegeo

docs/csharp/language-reference/builtin-types/record.md
```diff
@@ -53,12 +53,12 @@ You can use positional parameters to declare properties of a record and to initi
...
+* A public auto-implemented property for each positional parameter provided in the record declaration.
+  - For `record` types and `readonly record struct` types: An [init-only](../keywords/init.md) property.
+  - For `record struct` types: A read-write property.
 * A primary constructor whose parameters match the positional parameters on the record declaration.
 * For record struct types, a parameterless constructor that sets each field to its default value.
-* A `Deconstruct` method with an `out` parameter for each positional parameter provided in the record declaration. This method is provided only if there are two or more positional parameters. The method deconstructs properties defined by using positional syntax; it ignores properties that are defined by using standard property syntax.
+* A `Deconstruct` method with an `out` parameter for each positional parameter provided in the record declaration. The method deconstructs properties defined by using positional syntax; it ignores properties that are defined by using standard property syntax.
```
IEvangelist, line 61: I'm not sure it's a good idea to remove the messaging around the fact that positional records will only generate a `Deconstruct` when there are two or more positional props. Is that called out anywhere else?

```suggestion
* A `Deconstruct` method with an `out` parameter for each positional parameter provided in the record declaration. The method deconstructs properties defined by using positional syntax; it ignores properties that are defined by using standard property syntax. This method is provided only if there are two or more positional parameters.
```

IEvangelist (approved):
One clarifying question/suggestion.

docs/csharp/language-reference/builtin-types/record.md
BillWagner: I removed it because it's incorrect. The compiler *does* generate a deconstruct for a record with only one member.

PR #27648, opened 2021-12-21
Update mutually recursive record definition example

cartermp:
It's incomplete without this change :)


BillWagner (approved):
Thanks @cartermp 

This LGTM. I'll :shipit: now.

PR #27658, opened 2021-12-23
Fix typo in dotnet build --os option description

veloek:


includes/cli-os.md
```diff
@@ -6,4 +6,4 @@ ms.topic: include
 ---
 - **`--os <OS>`**
 
-  Specifies the target operating system (OS). This is a shorthand syntax for setting the [Runtime Identifier (RID)](../docs/core/rid-catalog.md), where the provided value is combined with the default RID. For example, on a `win-x64` machine, specifying `--os os` sets the RID to `os-x64`. If you use this option, don't use the `-r|--runtime` option. Available since .NET 6 Preview 7.
+  Specifies the target operating system (OS). This is a shorthand syntax for setting the [Runtime Identifier (RID)](../docs/core/rid-catalog.md), where the provided value is combined with the default RID. For example, on a `win-x64` machine, specifying `--os os` sets the RID to `win-x64`. If you use this option, don't use the `-r|--runtime` option. Available since .NET 6 Preview 7.
```
tdykstra, line 9:
```suggestion
  Specifies the target operating system (OS). This is a shorthand syntax for setting the [Runtime Identifier (RID)](../docs/core/rid-catalog.md), where the provided value is combined with the default RID. For example, on a `win-x64` machine, specifying `--os linux` sets the RID to `linux-x64`. If you use this option, don't use the `-r|--runtime` option. Available since .NET 6.
```

tdykstra (changes requested):
Thanks for catching that error, the fix just needs to be tweaked a bit.

tdykstra (approved):

PR #27671, opened 2021-12-26
Point constructor name omitted

Userbit:


pkulikov:
@Userbit that omission may be by design to demonstrate the target-typed new expression

docs/csharp/tour-of-csharp/snippets/shared/Types.cs
```diff
@@ -118,7 +118,7 @@ public static void Examples()
             //</CreatePairObject>
 
             //<ImplicitCastToBase>
-            Point a = new(10, 20);
+            Point a = new Poin(10, 20);
```
gfoidl, line 121:
```suggestion
            Point a = new Point(10, 20);
```
But I strongly believe that https://github.com/dotnet/docs/pull/27671#issuecomment-1001193747 applies here.

PR #27711, opened 2022-01-03
Fix bug 🐛 in code example, and update `ms.date`

### LINKED ISSUE #27504: Article: `Options pattern in .NET`; Section: `IValidateOptions for complex validation`; Bug in code example
In the section [IValidateOptions for complex validation](https://docs.microsoft.com/en-us/dotnet/core/extensions/options#ivalidateoptions-for-complex-validation) there is the following code sample:
```csharp
class ValidateSettingsOptions : IValidateOptions<SettingsOptions>
{
    public SettingsOptions _settings { get; private set; }

    public ValidateSettingsOptions(IConfiguration config) =>
        _settings = config.GetSection(SettingsOptions.ConfigurationSectionName)
                          .Get<SettingsOptions>();

    public ValidateOptionsResult Validate(string name, SettingsOptions options)
    {
        string result = "";
        var rx = new Regex(@"^[a-zA-Z''-'\s]{1,40}$");
        Match match = rx.Match(options.SiteTitle);
        if (string.IsNullOrEmpty(match.Value))
        {
            result += $"{options.SiteTitle} doesn't match RegEx\n";
        }

        if (options.Scale < 0 || options.Scale > 1000)
        {
            result += $"{options.Scale} isn't within Range 0 - 1000\n";
        }

        if (_settings.Scale is 0 && _settings.VerbosityLevel <= _settings.Scale)
        {
            result += "VerbosityLevel must be > than Scale.";
        }

        return result != null
            ? ValidateOptionsResult.Fail(result)
            : ValidateOptionsResult.Success;
    }
}
```

The returned result depends on whether the `result` variable is `null`, but it can never be null in this code example, so the return value will be always `ValidateOptionsResult.Fail(result)` which makes it a bad code example.

There are two options to fix this:

1. Assign `result` variable `null` on initialization.
2. On the return statement, check if the `result` is an empty string.

---
#### Document Details

* Content: [Options pattern in .NET](https://docs.microsoft.com/en-us/dotnet/core/extensions/options)
* Content Source: [docs/core/extensions/options.md](https://github.com/dotnet/docs/blob/main/docs/core/extensions/options.md)
* GitHub Login: @IEvangelist
* Microsoft Alias: **dapine**

IEvangelist:
## Summary

Fix bug 🐛 in code example, and update `ms.date`

Fixes #27504

### DIFF
--- a/docs/core/extensions/snippets/configuration/console-json/ValidateSettingsOptions.cs
+++ b/docs/core/extensions/snippets/configuration/console-json/ValidateSettingsOptions.cs
@@ -33,7 +33,7 @@ class ValidateSettingsOptions : IValidateOptions<SettingsOptions>
             failure.AppendLine("VerbosityLevel must be > than Scale.");
         }
 
-        return failure is { Length: > 0 }
+        return failure.Length > 0
             ? ValidateOptionsResult.Fail(failure.ToString())
             : ValidateOptionsResult.Success;
     }
--- a/docs/core/extensions/snippets/configuration/console-json/ValidateSettingsOptions.cs
+++ b/docs/core/extensions/snippets/configuration/console-json/ValidateSettingsOptions.cs
@@ -33,7 +33,7 @@ class ValidateSettingsOptions : IValidateOptions<SettingsOptions>
             failure.AppendLine("VerbosityLevel must be > than Scale.");
         }
 
-        return failure is {  Length: > 0 }
+        return failure is { Length: > 0 }
             ? ValidateOptionsResult.Fail(failure.ToString())
             : ValidateOptionsResult.Success;
     }
--- a/docs/core/extensions/snippets/configuration/console-json/ValidateSettingsOptions.cs
+++ b/docs/core/extensions/snippets/configuration/console-json/ValidateSettingsOptions.cs
@@ -1,4 +1,5 @@
-﻿using System.Text.RegularExpressions;
+﻿using System.Text;
+using System.Text.RegularExpressions;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Options;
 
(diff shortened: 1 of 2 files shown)


BillWagner (approved):
This LGTM @IEvangelist 

You can :shipit: when ready.

docs/core/extensions/snippets/configuration/console-json/ValidateSettingsOptions.cs
```diff
@@ -14,26 +15,26 @@ class ValidateSettingsOptions : IValidateOptions<SettingsOptions>
...
 
         if (_settings.Scale is 0 && _settings.VerbosityLevel <= _settings.Scale)
         {
-            result += "VerbosityLevel must be > than Scale.";
+            failure.AppendLine("VerbosityLevel must be > than Scale.");
         }
 
-        return result != null
-            ? ValidateOptionsResult.Fail(result)
+        return failure is { Length: > 0 }
```
Youssef1313, line 36:
```suggestion
        return failure.Length > 0
```

docs/core/extensions/snippets/configuration/console-json/ValidateSettingsOptions.cs
```diff
@@ -14,26 +15,26 @@ class ValidateSettingsOptions : IValidateOptions<SettingsOptions>
 
     public ValidateOptionsResult Validate(string name, SettingsOptions options)
     {
-        string result = "";
+        StringBuilder failure = new();
         var rx = new Regex(@"^[a-zA-Z''-'\s]{1,40}$");
```
Youssef1313, line 19: Unrelated to the changes, but I think it makes more sense to make this a private static field so that allocation is done only once. The private static field could be directly initialized or could be made lazy. I don't have a strong opinion which is better.

Youssef1313 (commented):
ASP.NET docs have a similar snippet:

https://github.com/dotnet/AspNetCore.Docs/blob/0799d9bda43805ef4cea766d8567fb963fd8744d/aspnetcore/fundamentals/configuration/options/samples/6.x/OptionsValidationSample/Configuration/MyConfigValidation.cs

https://github.com/dotnet/AspNetCore.Docs/blob/0799d9bda43805ef4cea766d8567fb963fd8744d/aspnetcore/fundamentals/configuration/options/samples/3.x/OptionsValidationSample/Configuration/MyConfigValidation.cs

@Rick-Anderson to consider similar changes for ASP.NET

docs/core/extensions/snippets/configuration/console-json/ValidateSettingsOptions.cs
IEvangelist: I do not like the idea of having a class-scoped variable as we do not control the number of times that `Validate` is called, nor do we know from what number of threads this is called concurrently. Not saying it's an issue, but it leaves the door open for the contention and stale reads/writes.

PR #27725, opened 2022-01-04
Restrict all actions that rely on the `GITHUB_TOKEN`

IEvangelist:
## Summary

In this PR, we're explicitly declaring the individual actions' `permissions`. This overrides the default permissions for the `GITHUB_TOKEN`, restricting it to the bare minimal access in order for the given action to run. For more information, see:

- [Workflow syntax for GitHub Actions: `permissions`](https://docs.github.com/actions/learn-github-actions/workflow-syntax-for-github-actions#permissions)
- [Permissions for the `GITHUB_TOKEN`](https://docs.github.com/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -163,7 +163,7 @@ csharp_space_between_parentheses = false
 csharp_space_between_square_brackets = false
 
 # Namespace preference
-csharp_style_namespace_declarations = file_scoped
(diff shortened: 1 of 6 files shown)


adegeo (approved):
I think this should be fine.

PR #27745, opened 2022-01-06
Fix code format

lindexi:
## Summary

Fix code format and fix the API usage.

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR/Asynchronous_File_IO_async/cs/AsynchronousFileIO.csproj
+++ b/samples/snippets/csharp/VS_Snippets_CLR/Asynchronous_File_IO_async/cs/AsynchronousFileIO.csproj
@@ -1,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0-windows</TargetFramework>
+    <Nullable>enable</Nullable>
+    <UseWPF>true</UseWPF>
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+  </PropertyGroup>
+
+</Project>
--- a/samples/snippets/csharp/VS_Snippets_CLR/Asynchronous_File_IO_async/cs/example.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR/Asynchronous_File_IO_async/cs/example.cs
@@ -15,16 +15,16 @@ namespace WpfApplication
 
         private async void Button_Click(object sender, RoutedEventArgs e)
         {
-            string StartDirectory = @"c:\Users\exampleuser\start";
-            string EndDirectory = @"c:\Users\exampleuser\end";
+            string startDirectory = @"c:\Users\exampleuser\start";
+            string endDirectory = @"c:\Users\exampleuser\end";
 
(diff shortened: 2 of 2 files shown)


IEvangelist (approved):

IEvangelist:
Hi @lindexi - thank you for fixing this. There is a build error, could you please also add a new .csproj for this?

>  Check failure on line 1 in samples/snippets/csharp/VS_Snippets_CLR/Asynchronous_File_IO_async/cs/example.cs
> 
> GitHub Actions
> / snippets-build
> 
> samples/snippets/csharp/VS_Snippets_CLR/Asynchronous_File_IO_async/cs/example.cs#L1
> 
> 😵 Project missing. A project (and optionally a solution file) must be in this directory or one of the parent directories to validate and build this code

lindexi:
@IEvangelist  Sorry, I can not compile the project after I added the csproj file. Because we define the same MainWindow class in `example.cs` and `example3.cs` and `example4.cs`

Should I rename the MainWindow to MainWindow3 or MainWindow4 in  `example3.cs` and `example4.cs` ?

samples/snippets/csharp/VS_Snippets_CLR/Asynchronous_File_IO_async/cs/AsynchronousFileIO.csproj
```diff
@@ -0,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0-windows</TargetFramework>
+    <Nullable>enable</Nullable>
+    <UseWPF>true</UseWPF>
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
```
Youssef1313, line 7: Why is this needed?

lindexi: @Youssef1313  Because we define the same MainWindow class in example.cs and example3.cs and example4.cs

Youssef1313: The problem with `EnableDefaultCompileItems` being false is that it beats the purpose of the csproj, which is to make sure the code compiles properly. If the examples are completely independent, I'd make a folder for each of them, each with its own csproj. @IEvangelist Do you agree?

Youssef1313: Actually `example3` and `example4` doesn't seem to be referenced. So they can just be deleted?

lindexi: @IEvangelist Can I delete the example3.cs and example4.cs file and create the MainWindow2.xaml file?

IEvangelist: You'd have to verify that those two files are in fact not being referenced, if you can prove that then yes.

PR #27752, opened 2022-01-06
Update whatsnew action

BillWagner:
Make sure none of the queries request organization information. The default GitHub access token prevents those queries.


IEvangelist (approved):

PR #27802, opened 2022-01-11
Update ca1829.md

ddobric:
SImple correction. Lenght is more efficient. Length or Count is not.

## Summary

Describe your changes here.

### DIFF
--- a/docs/fundamentals/code-analysis/quality-rules/ca1829.md
+++ b/docs/fundamentals/code-analysis/quality-rules/ca1829.md
@@ -26,7 +26,7 @@ The <xref:System.Linq.Enumerable.Count%2A> LINQ method was used on a type that s
 
 ## Rule description
 
-This rule flags the <xref:System.Linq.Enumerable.Count%2A> LINQ method calls on collections of types that have equivalent, but more efficient `Length` or `Count` property to fetch the same data. `Length` or `Count` property does not enumerate the collection, hence is more efficient.
+This rule flags the <xref:System.Linq.Enumerable.Count%2A> LINQ method calls on collections of types that have equivalent, but more efficient `Length` or `Count` property to fetch the same data. `Length` property does not enumerate the collection, hence is more efficient.
 
 This rule flags <xref:System.Linq.Enumerable.Count%2A> calls on the following collection types with `Length` property:
 


gewarren:
@ddobric The article states that the Count property is more efficient on the following collections:

"This rule flags Count calls on the following collection types with the Count property:

System.Collections.ICollection
System.Collections.Generic.ICollection<T>
System.Collections.Generic.IReadOnlyCollection<T>"

Can you elaborate more and why you think we should remove the Count property from the sentence you changed?

Youssef1313 (changes requested):
The original sentence is correct. I see the confusion comes from the fact that there is a `Count` **method** (inefficient), and a `Count` **property** (efficient - and in the common case comes from `ICollection<T>`).

For example, see how `List<T>` implements this property:

https://github.com/dotnet/runtime/blob/4e2102b73b4ab673242fcb480937130b12d04d1b/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L126-L127

*Side note:* even the LINQ Count method have a fast-path for certain types and in many cases it doesn't enumerate the collection as well. This is a thing that I think might not be accurate in the documentation.

The LINQ Count implementation for reference:

https://github.com/dotnet/runtime/blob/cf5b231fcbea483df3b081939b422adfb6fd486a/src/libraries/System.Linq/src/System/Linq/Count.cs#L11

Youssef1313:
@mavasani @Evangelink Might have more context on the side note I mentioned. There *might* be some historical reasons where in old TFMs the LINQ Count method used to always enumerate.

Evangelink:
> The original sentence is correct. I see the confusion comes from the fact that there is a `Count` **method** (inefficient), and a `Count` **property** (efficient - and in the common case comes from `ICollection<T>`).

Yes.

> @mavasani Manish Vasani FTE @Evangelink Amaury Levé FTE Might have more context on the side note I mentioned. There _might_ be some historical reasons where in old TFMs the LINQ Count method used to always enumerate.

I am pretty sure this optimization wasn't in place when the rule was designed.

Now we could argue whether the rule still make sense for the `.Count()` (method) vs `.Count`  (property) case. Accessing the property is still faster than having a method call with casts but the difference might be neglectable now. This being said, if the implementation changed, that means that depending on the runtime we might have a slower experience so I would be in favor of keeping the rule as-is and closing this PR.

gewarren:
Okay, closing this PR. Thanks for the input!

PR #27809, opened 2022-01-12
Version deserialization breaking change

### LINKED ISSUE #26292: [Breaking change]: JsonSerializer throws FormatException when deserializing Version types with leading or trailing whitespace
### Description

JsonSerializer goes from permitting leading and trailing whitespace when deserializing Version types to throwing `FormatException`. Breaking change introduced in .NET 7 Preview 1.

### Version

### Previous behavior

Provided in description.

### New behavior

Provided in description.

### Type of breaking change

### Reason for change

We optimized the implementation of the underlying `Version` converter. This resulted in the implementation being made to align with the behavior for other primitive types supported by `System.Text.Json` e.g. `DateTime` and `Guid` which also disallow leading and trailing spaces.

### Recommended action

To get the old behavior back, add a custom converter for the type which permits whitespace:

```cs
internal sealed class VersionConverter : JsonConverter<Version>
{
    public override Version Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
    {
        string? versionString = reader.GetString();
        if (Version.TryParse(versionString, out Version? result))
        {
            return result;
        }

        ThrowHelper.ThrowJsonException();
        return null;
    }

    public override void Write(Utf8JsonWriter writer, Version value, JsonSerializerOptions options)
    {
        writer.WriteStringValue(value.ToString());
    }
}
```

### Feature area

Core .NET libraries

### Affected APIs

All `System.Text.Json.JsonSerializer.Deserialize` method overloads.

gewarren:
Fixes #26292

[Preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/serialization/7.0/deserialize-version-with-whitespace?branch=pr-en-us-27809)

### DIFF
--- a/docs/core/compatibility/7.0.md
+++ b/docs/core/compatibility/7.0.md
@@ -17,3 +17,9 @@ If you're migrating an app to .NET 7, the breaking changes listed here might aff
 | Title | Binary compatible | Source compatible | Introduced |
 | - | - | - | - |
 | [Validate CompressionLevel for BrotliStream](core-libraries/7.0/compressionlevel-validation.md) | ❌ | ✔️ | Preview 1 |
+
+## Serialization
+
+| Title | Binary compatible | Source compatible | Introduced |
+| - | - | - | - |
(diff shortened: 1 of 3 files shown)


tdykstra (approved):
LGTM

PR #27820, opened 2022-01-12
Add conceptual documentation for "disabled runtime marshalling"

jkoritzinsky:
## Summary

Add some conceptual documentation for the upcoming `System.Runtime.InteropServices.DisableRuntimeMarshallingAttribute` in our new interop documentation.

This PR is expected to fail until after dotnet/runtime#63320 is merged and dotnet/dotnet-api-docs is updated with API docs including the new attribute.

cc: @dotnet/interop-contrib

### DIFF
--- a/docs/standard/native-interop/best-practices.md
+++ b/docs/standard/native-interop/best-practices.md
@@ -84,6 +84,8 @@ GUIDs are usable directly in signatures. Many Windows APIs take `GUID&` type ali
 
 Blittable types are types that have the same bit-level representation in managed and native code. As such they do not need to be converted to another format to be marshaled to and from native code, and as this improves performance they should be preferred. Some types are not blittable but are known to contain blittable contents. These types have similar optimizations as blittable types when they are not contained in another type, but are not considered blittable when in fields of structs or for the purposes of [`UnmanagedCallersOnlyAttribute`](xref:System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute).
 
+### Blittable types when runtime marshalling is enabled
+
 **Blittable types:**
 
 - `byte`, `sbyte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `single`, `double`
@@ -125,6 +127,10 @@ public struct UnicodeCharStruct
 `string` contains blittable contents if it isn't contained in another type and it's being passed as an argument that is marked with `[MarshalAs(UnmanagedType.LPWStr)]` or the `[DllImport]` has `CharSet = CharSet.Unicode` set.
 
 You can see if a type is blittable or contains blittable contents by attempting to create a pinned `GCHandle`. If the type isn't a string or considered blittable, `GCHandle.Alloc` will throw an `ArgumentException`.
+
+### Blittable types when runtime marshalling is disabled
+
+When [runtime marshalling is disabled](disabled-marshalling.md), the rules for which types are blittable are significantly simpler. All types that are [C# `unmanaged`](../../csharp/language-reference/builtin-types/unmanaged-types.md) types are blittable. All types that are not C# `unmanaged` types are not blittable. The concept of "types with blittable contents" does not apply when runtime marshalling is disabled.
 
 ✔️ DO make your structures blittable when possible.
 
--- a/docs/standard/native-interop/best-practices.md
+++ b/docs/standard/native-interop/best-practices.md
@@ -54,7 +54,7 @@ If you *do* use `StringBuilder`, one last gotcha is that the capacity does **not
 
 ✔️ CONSIDER using `char[]`s from an `ArrayPool`.
 
-For more information on string marshalling, see [Default Marshalling for Strings](../../framework/interop/default-marshalling-for-strings.md) and [Customizing string marshalling](customize-parameter-marshalling.md#customizing-string-parameters).
+For more information on string marshalling, see [Default Marshalling for Strings](../../framework/interop/default-marshaling-for-strings.md) and [Customizing string marshalling](customize-parameter-marshalling.md#customizing-string-parameters).
 
 > __Windows Specific__
 > For `[Out]` strings the CLR will use `CoTaskMemFree` by default to free strings or `SysStringFree` for strings that are marked
--- a/docs/standard/native-interop/best-practices.md
+++ b/docs/standard/native-interop/best-practices.md
@@ -25,7 +25,7 @@ The guidance in this section applies to all interop scenarios.
 |---------|---------|----------------|---------|
 | <xref:System.Runtime.InteropServices.DllImportAttribute.PreserveSig>   | `true` |  keep default  | When this is explicitly set to false, failed HRESULT return values will be turned into exceptions (and the return value in the definition becomes null as a result).|
 | <xref:System.Runtime.InteropServices.DllImportAttribute.SetLastError> | `false`  | depends on the API  | Set this to true if the API uses GetLastError and use Marshal.GetLastWin32Error to get the value. If the API sets a condition that says it has an error, get the error before making other calls to avoid inadvertently having it overwritten.|
-| <xref:System.Runtime.InteropServices.DllImportAttribute.CharSet> | Compiler-defined (specified in the [charset documentation](./charset.md))  | Explicitly  use `CharSet.Unicode` or `CharSet.Ansi` when strings or characters are present in the definition | This specifies marshaling behavior of strings and what `ExactSpelling` does when `false`. Note that `CharSet.Ansi` is actually UTF8 on Unix. _Most_ of the time Windows uses Unicode while Unix uses UTF8. See more information on the [documentation on charsets](./charset.md). |
+| <xref:System.Runtime.InteropServices.DllImportAttribute.CharSet> | Compiler-defined (specified in the [charset documentation](./charset.md))  | Explicitly  use `CharSet.Unicode` or `CharSet.Ansi` when strings or characters are present in the definition | This specifies marshalling behavior of strings and what `ExactSpelling` does when `false`. Note that `CharSet.Ansi` is actually UTF8 on Unix. _Most_ of the time Windows uses Unicode while Unix uses UTF8. See more information on the [documentation on charsets](./charset.md). |
 | <xref:System.Runtime.InteropServices.DllImportAttribute.ExactSpelling> | `false` | `true`             | Set this to true and gain a slight perf benefit as the runtime will not look for alternate function names with either an "A" or "W" suffix depending on the value of the `CharSet` setting ("A" for `CharSet.Ansi` and "W" for `CharSet.Unicode`). |
 
 ## String parameters
@@ -36,7 +36,7 @@ Remember to mark the `[DllImport]` as `Charset.Unicode` unless you explicitly wa
 
 ❌ DO NOT use `[Out] string` parameters. String parameters passed by value with the `[Out]` attribute can destabilize the runtime if the string is an interned string. See more information about string interning in the documentation for <xref:System.String.Intern%2A?displayProperty=nameWithType>.
 
-❌ AVOID `StringBuilder` parameters. `StringBuilder` marshaling *always* creates a native buffer copy. As such, it can be extremely inefficient. Take the typical scenario of calling a Windows API that takes a string:
+❌ AVOID `StringBuilder` parameters. `StringBuilder` marshalling *always* creates a native buffer copy. As such, it can be extremely inefficient. Take the typical scenario of calling a Windows API that takes a string:
 
 1. Create a `StringBuilder` of the desired capacity (allocates managed capacity) **{1}**.
 2. Invoke:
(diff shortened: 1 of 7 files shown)


docs/standard/native-interop/best-practices.md
```diff
@@ -126,6 +128,10 @@ public struct UnicodeCharStruct
 
 You can see if a type is blittable or contains blittable contents by attempting to create a pinned `GCHandle`. If the type isn't a string or considered blittable, `GCHandle.Alloc` will throw an `ArgumentException`.
 
+### Blittable types when runtime marshalling is disabled
+
+When [runtime marshalling is disabled](disabled-marshalling.md), the rules for which types are blittable are significantly simpler. All types that are [C# `unmanaged`](../../csharp/language-reference/builtin-types/unmanaged-types.md) types are blittable. All types that are not C# `unmanaged` types are not blittable. The concept of "types with blittable contents" does not apply when runtime marshalling is disabled.
```
tannergooding, line 133: Would it be worth clarifying that this largely impacts `bool` and `char` on RyuJIT (.NET 6+). Other implementations may of course have their own differences here, but calling out the "common" case seems helpful at least.

docs/standard/native-interop/best-practices.md
```diff
@@ -84,6 +84,8 @@ GUIDs are usable directly in signatures. Many Windows APIs take `GUID&` type ali
 
 Blittable types are types that have the same bit-level representation in managed and native code. As such they do not need to be converted to another format to be marshaled to and from native code, and as this improves performance they should be preferred. Some types are not blittable but are known to contain blittable contents. These types have similar optimizations as blittable types when they are not contained in another type, but are not considered blittable when in fields of structs or for the purposes of [`UnmanagedCallersOnlyAttribute`](xref:System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute).
 
+### Blittable types when runtime marshalling is enabled
+
 **Blittable types:**
 
 - `byte`, `sbyte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `single`, `double`
```
tannergooding, line 91: This covers `string` below and I would assume is no longer supported when `runtime marshalling is disabled`. Is it worth explicitly calling that out?

tannergooding: I missed the sentence that does cover this when reading the first time so looks like no action is really needed, other than maybe explicitly calling our `string`/`arrays` if desired.
> The concept of "types with blittable contents" does not apply when runtime marshalling is disabled.

👍

docs/standard/native-interop/disabled-marshaling.md
```diff
@@ -0,0 +1,45 @@
...
+ms.date: 01/12/2022
+---
+
+# Disabled runtime marshalling
+
+When the [`System.Runtime.CompilerServices.DisableRuntimeMarshallingAttribute`](xref:System.Runtime.CompilerServices.DisableRuntimeMarshallingAttribute) attribute is applied to an assembly, the runtime disables most built-in support for data marshalling between managed and native representations. This document describes the features that are disabled and how .NET types map to native types when marshalling is disabled.
+
+## Scenarios where marshalling is disabled
+
+When the `DisableRuntimeMarshallingAttribute` is applied to an assembly, it affects P/Invokes and Delegate types in the assembly, as well as any calls to unmanaged function pointers in the assembly. It does not affect any P/Invoke or interop delegate types defined in other assemblies. It also does not disable marshalling for the runtime's built-in COM interop support. The built-in COM interop support can be enabled or disabled through a [feature switch](https://github.com/dotnet/runtime/blob/main/docs/workflow/trimming/feature-switches.md).
```
tannergooding, line 13: Could you give some clarifying examples here?

It's not exactly clear which of the below work vs not and what the behavior is given
```csharp
[UnmanagedFunctionPointer(...)]
public delegate void Callback();

[DllImport(...)]
public static extern void Import(Callback callback);
```

Where you have the varying combinations of `Import` defined in `A` and `Callback` defined in `B` where `A` and/or `B` have `DisableRuntimeMarshalling` defined.

tannergooding: -- It's also "somewhat" unfortunate that `COM interop` is controlled by a separate switch I think. It would be nice to have a "one stop shop" to disable all built-in marshalling and enforce blittable conventions.

It's not a big deal though, it just means one extra thing to tweak in my own libraries.

docs/standard/native-interop/disabled-marshaling.md
```diff
@@ -0,0 +1,45 @@
...
+| `byte`    | `uint8_t`          |
+| `sbyte`   | `int8_t`           |
+| `short`   | `int16_t`          |
+| `ushort`  | `uint16_t`         |
+| `int`     | `int32_t`          |
+| `uint`    | `uint32_t`         |
+| `long`    | `int64_t`          |
+| `ulong`   | `uint64_t`         |
+| `char`    | `char16_t` (`CharSet` on the P/Invoke has no effect) |
+| `System.IntPtr` | `intptr_t`   |
```
tannergooding, line 41: Would it be worth expanding this to be:
```
| C# Keyword | .NET Type | Native Type|
```

That would allow covering `nint` and `nuint`, rather than only displaying them as `IntPtr`/`UIntPtr`.

tannergooding: It would likewise be nice to then add
```
| N/A | `System.Runtime.InteropServices.CLong` | `long` |
| N/A | `System.Runtime.InteropServices.CULong` | `unsigned long` |
| N/A | `System.Runtime.InteropServices.NFloat` | `something representing its 32-bits on 32-bit and 64-bits on 64-bits, intended for iOS` |
```

There is also technically `Vector64/128/256<T>` which are the `__m64/128/256` ABI types; but which we don't currently support (but intend to support eventually).

tannergooding (approved):
Overall LGTM. Left a few comments/suggestions that might help improve clarity in some areas.

docs/standard/native-interop/disabled-marshaling.md
jkoritzinsky: Since this feature doesn't totally disable P/Invoke and delegate-based interop, it didn't make sense to fully disable COM interop as part of this attribute. Since COM would still have to generate IL stubs and use marshalling for the `this` parameter, it doesn't make sense (at least to me) to disable the marshalling but not the built-in COM system as a whole.

AaronRobinsonMSFT: > It would be nice to have a "one stop shop" to disable all built-in marshalling and enforce blittable conventions.

I'm not against that, but it seems hard in my mind. Does it mean the PIA itself &ndash; the containing assembly of the COM type &ndash; applies to that definition of the interface? Does it apply the GUID for the type? If the assembly using COM interop uses an interface does the attribute apply to the call? (I assume not). I guess we could also apply the attribute to the assembly that activates the COM object, but that also gets weird.

If you can come up with a crisp set of rules that makes intuitive sense I'd be very interested.

docs/standard/native-interop/disabled-marshaling.md
```diff
@@ -0,0 +1,97 @@
...
+When the `DisableRuntimeMarshallingAttribute` is applied to an assembly, the following attributes will have no effect or throw an exception:
+
+- [`LCIDConversionAttribute`](xref:System.Runtime.InteropServices.LCIDConversionAttribute) on a P/Invoke or a delegate
+- `SetLastError=true` on a P/Invoke
+- `ThrowOnUnmappableChar=true` on a P/Invoke
+- `BestFitMapping=true` on a P/Invoke
+- .NET variadic argument method signatures (varargs)
+- `in`, `ref`, `out` parameters
+
+## Default rules for marshaling common types
```
AaronRobinsonMSFT, line 26:
```suggestion
## Default rules for marshalling common types
```

AaronRobinsonMSFT (commented):
It would appear the names of the files need to be updated too "marshalling". Sad panda.

docs/standard/native-interop/disabled-marshaling.md
jkoritzinsky: Looking in the redirection file, we renamed to single-l "marshaling" a while back in #12005. See https://github.com/dotnet/docs/issues/11986.

AaronRobinsonMSFT: Bah. So, what are we doing here? I thought we agreed on two ll, but could be misremembering.

jkoritzinsky: I think we as the interop team agreed on two Ls, but we never updated the style guide that the docs team uses.

@gewarren, can we update the style guide and rename things back to "marshalling"?

jkoritzinsky: I've submitted a request to update the style guide. Let's see where this goes.

tannergooding: It would be good to also update and ensure its covered in https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalizing-compound-words-and-common-terms

CC. @bartonjs as he may know if we have concrete guidance on `marshaling` vs `marshalling`

AaronRobinsonMSFT (approved):
All good to me. I'd like resolution on the l vs ll debate though.

docs/standard/native-interop/disabled-marshaling.md
gewarren: Based on email from Celia Parker today (she spent 7 hours researching this), it sounds like we are sticking to one `l` for now.

"[...] That should be reason enough to support the argument of not changing "marshal" to "marshall" in the IESG.
Unless you have much more compelling reasoning with more powerful evidence to support it, I believe we should leave the IESG as it is."

That said, is this good to merge?

tannergooding: I think so, unless @bartonjs has any pushback based on the FDG

tannergooding: -- (We can also always fix it up later if there is pushback there)

bartonjs: It's not in the standard spellings table.

Based on existing public types (which @tannergooding already pointed out), I believe our de facto in .NET is `-ller` and `-lling`.  Here are all the hits of Marshal in our refs.cs files where it's not the last word in the type/method name or followed by a capital letter:

```
System.Reflection.Metadata.cs:        public System.Reflection.Metadata.BlobHandle GetMarshallingDescriptor() { throw null; }
System.Reflection.Metadata.cs:        public System.Reflection.Metadata.BlobHandle GetMarshallingDescriptor() { throw null; }
System.Reflection.Metadata.cs:        public void AddMarshallingDescriptor(System.Reflection.Metadata.EntityHandle parent, System.Reflection.Metadata.BlobHandle descriptor) { }
System.Runtime.InteropServices.cs:    public partial interface ICustomMarshaler
System.Runtime.InteropServices.cs:        [System.ObsoleteAttribute("Marshalling as Currency may be unavailable in future releases.")]
System.Runtime.InteropServices.cs:        [System.ObsoleteAttribute("Marshalling as VBByRefString may be unavailable in future releases.")]
System.Runtime.InteropServices.cs:        [System.ObsoleteAttribute("Marshalling as AnsiBStr may be unavailable in future releases.")]
System.Runtime.InteropServices.cs:        [System.ObsoleteAttribute("Marshalling as TBstr may be unavailable in future releases.")]
System.Runtime.InteropServices.cs:        [System.ObsoleteAttribute("Marshalling arbitrary types may be unavailable in future releases. Please specify the type you wish to marshal as.")]
System.Runtime.InteropServices.cs:        public static void RegisterForMarshalling(ComWrappers instance) { }
```

Counting the obsoletion messages, it's 9:1, ignoring them, it's 4:1.  (And that's not counting things like the `MarshallerHelpers` from the new interop generator)

Wikipedia's current style for the word marshalling is the two-l form (https://en.wikipedia.org/wiki/Marshalling_(computer_science)).  While I generally agree with "US spelling" when there's a country-based difference, in this case, I believe one-l is wrong.

gewarren:
> All good to me. I'd like resolution on the l vs ll debate though.

We've sent a request to the Cloud & AI Style Guide to update the guidance to use two l's. I'm assuming they'll approve this request, so I've updated the wording in this PR to use two l's.

docs/standard/native-interop/customize-struct-marshaling.md
```diff
@@ -259,7 +259,7 @@ struct BString
 };
 ```
 
-When using a WinRT-based API, you may need to marshal a string as an `HSTRING`. Using the <xref:System.Runtime.InteropServices.UnmanagedType.HString?displayProperty=nameWithType> value, you can marshal a string as a `HSTRING`.
+When using a WinRT-based API, you may need to marshal a string as an `HSTRING`. Using the <xref:System.Runtime.InteropServices.UnmanagedType.HString?displayProperty=nameWithType> value, you can marshal a string as a `HSTRING`. `HSTRING` marshalling is only suppported on runtimes with built-in WinRT support. WinRT support was [removed in .NET 5](/core/compatibility/interop/5.0/built-in-support-for-winrt-removed.md), so `HSTRING` marshalling is not supported in .NET 5.0 or newer.
```
gewarren, line 262:
```suggestion
When using a WinRT-based API, you may need to marshal a string as an `HSTRING`. Using the <xref:System.Runtime.InteropServices.UnmanagedType.HString?displayProperty=nameWithType> value, you can marshal a string as a `HSTRING`. `HSTRING` marshalling is only suppported on runtimes with built-in WinRT support. WinRT support was [removed in .NET 5](../../core/compatibility/interop/5.0/built-in-support-for-winrt-removed.md), so `HSTRING` marshalling is not supported in .NET 5.0 or newer.
```

PR #27841, opened 2022-01-13
Updated the wrong function calls with the correct one

vcrobe:
## Summary

Describe your changes here.


BillWagner (approved):
Thanks for spotting this and making the fix @vcrobe 

PR #27851, opened 2022-01-14
Update nullable-references.md

IEvangelist:
**Enabled auto-merge**

Correct nullable context identifiers. `enabled` and `disabled` do not exist, instead it is `enable` and `disable`


BillWagner (approved):
Thanks for catching this @IEvangelist 

Let's :shipit:

Issue #26720, opened 2021-10-28
[Breaking change]: API validate input parameters and throw ANE 


RussKie:
### Description

Selected API now validate input parameters and throw `ArgumentNullException` where previously a `NullReferenceException` could be thrown, if APIs were invoked with `null` input arguments.
Throwing `ArgumentNullException` is inline with the behaviours in .NET Runtime, and provides developers with a better debug experience clearly communicating which argument caused the exception.

### Previous behavior

The listed API would throw `NullReferenceException` if invoked with `null` input arguments.

### New behavior

The listed API would throw `ArgumentNullException` if invoked with `null` input arguments.

### Type of breaking change

- [X] **Binary incompatible**: Existing binaries may encounter a breaking change in behavior, such as failure to load/execute or different run-time behavior.
- [ ] **Source incompatible**: Source code may encounter a breaking change in behavior when targeting the new runtime/component/SDK, such as compile errors or different run-time behavior.

### Reason for change

Throwing `ArgumentNullException` is inline with the behaviours in .NET Runtime, and provides developers with a better debug experience clearly communicating which argument caused the exception.

### Recommended action

- Review and, if necessary, update the code to prevent passing `null` input arguments in the listed API.
- If the code is handling `NullReferenceException` replace (or supplement) it with handling `ArgumentNullException`.

### Feature area

Windows Forms

#### Affected APIs

* [ToolStripArrowRenderEventArgs](https://docs.microsoft.com/dotnet/api/system.windows.forms.toolstriparrowrendereventargs)
	- [x] [ToolStripArrowRenderEventArgs(Graphics, ToolStripItem, Rectangle, Color, ArrowDirection) Constructor](https://docs.microsoft.com/dotnet/api/system.windows.forms.toolstriparrowrendereventargs.-ctor) when `Graphics` is `null`
* [ToolStripItemRenderEventArgs](https://docs.microsoft.com/dotnet/api/system.windows.forms.toolstripitemrendereventargs)
	- [x] [ToolStripItemRenderEventArgs(Graphics, ToolStripItem) Constructor](https://docs.microsoft.com/dotnet/api/system.windows.forms.toolstripitemrendereventargs.-ctor) when `Graphics` is `null`
	- [x] [ToolStripItemRenderEventArgs(Graphics, ToolStripItem) Constructor](https://docs.microsoft.com/dotnet/api/system.windows.forms.toolstripitemrendereventargs.-ctor) when `ToolStripItem` is `null`
[...]

PR #27854, opened 2022-01-14
Add ANE breaking change for Preview 1

gewarren:
Contributes to #26720 (also need to update API ref).

[Preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/windows-forms/7.0/apis-throw-argumentnullexception?branch=pr-en-us-27854).

### DIFF
--- a/docs/core/compatibility/7.0.md
+++ b/docs/core/compatibility/7.0.md
@@ -24,3 +24,9 @@ If you're migrating an app to .NET 7, the breaking changes listed here might aff
 | Title | Binary compatible | Source compatible | Introduced |
 | - | - | - | - |
 | [Deserialize Version type with leading or trailing whitespace](serialization/7.0/deserialize-version-with-whitespace.md) | ❌ | ✔️ | Preview 1 |
+
+## Windows Forms
+
+| Title | Binary compatible | Source compatible | Introduced |
+| - | - | - | - |
+| [Some APIs throw ArgumentNullException](windows-forms/7.0/apis-throw-argumentnullexception.md) | ❌ | ✔️ | Preview 1 |
--- a/docs/core/compatibility/7.0.md
+++ b/docs/core/compatibility/7.0.md
@@ -14,19 +14,19 @@ If you're migrating an app to .NET 7, the breaking changes listed here might aff
 
 ## Core .NET libraries
 
-| Title | Binary compatible | Source compatible | Introduced |
-| - | - | - | - |
+| <div style="width:290px">Title</div> | Binary compatible | Source compatible | Introduced |
+| - | :-: | :-: | - |
 | [API obsoletions with non-default diagnostic IDs](core-libraries/7.0/obsolete-apis-with-custom-diagnostics.md) | ✔️ | ❌ | Preview 1 |
 | [Validate CompressionLevel for BrotliStream](core-libraries/7.0/compressionlevel-validation.md) | ❌ | ✔️ | Preview 1 |
 
 ## Serialization
 
-| Title | Binary compatible | Source compatible | Introduced |
-| - | - | - | - |
+| <div style="width:290px">Title</div> | Binary compatible | Source compatible | Introduced |
+| - | :-: | :-: | - |
 | [Deserialize Version type with leading or trailing whitespace](serialization/7.0/deserialize-version-with-whitespace.md) | ❌ | ✔️ | Preview 1 |
 
 ## Windows Forms
 
-| Title | Binary compatible | Source compatible | Introduced |
-| - | - | - | - |
+| <div style="width:290px">Title</div> | Binary compatible | Source compatible | Introduced |
+| - | :-: | :-: | - |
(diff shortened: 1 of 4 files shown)


RussKie:
Is there a reason why the menu entry reads differently?
image

docs/core/compatibility/7.0.md
```diff
@@ -24,3 +24,9 @@ If you're migrating an app to .NET 7, the breaking changes listed here might aff
 | Title | Binary compatible | Source compatible | Introduced |
 | - | - | - | - |
 | [Deserialize Version type with leading or trailing whitespace](serialization/7.0/deserialize-version-with-whitespace.md) | ❌ | ✔️ | Preview 1 |
+
+## Windows Forms
+
+| Title | Binary compatible | Source compatible | Introduced |
+| - | - | - | - |
+| [Some APIs throw ArgumentNullException](windows-forms/7.0/apis-throw-argumentnullexception.md) | ❌ | ✔️ | Preview 1 |
```
RussKie, line 32: Not really related to this PR, but this gives me an OCD...
image

If possible I recommend [center aligning](https://markdown.land/markdown-table#aligning-columns) the images (2nd and 3rd columns).
It'd also be great to "fix" the width of the last three columns too. Unfortunately MD syntax does not provide a way to do this, but looks like there are few possible clever ways of achieving this (e.g. [this](https://stackoverflow.com/questions/36121672/set-table-column-width-via-markdown)).

RussKie (approved):

docs/core/compatibility/windows-forms/7.0/apis-throw-argumentnullexception.md
```diff
@@ -0,0 +1,45 @@
...
+- If your code handles <xref:System.NullReferenceException>, replace or add an additional handler for <xref:System.ArgumentNullException>.
+
+## Affected APIs
+
+The following table lists the affected APIs and specific parameters:
+
+| Method/property | Parameter name |
+|-|-|
+| <xref:System.Windows.Forms.ToolStripArrowRenderEventArgs.%23ctor(System.Drawing.Graphics,System.Windows.Forms.ToolStripItem,System.Drawing.Rectangle,System.Drawing.Color,System.Windows.Forms.ArrowDirection)> | `g` |
+| <xref:System.Windows.Forms.ToolStripItemRenderEventArgs.%23ctor(System.Drawing.Graphics,System.Windows.Forms.ToolStripItem)> | `g` |
```
RussKie, line 42: We had a follow up in https://github.com/dotnet/winforms/pull/6507 to make the other arg non-null as well
Not sure the best way to document this:
```suggestion
| <xref:System.Windows.Forms.ToolStripItemRenderEventArgs.%23ctor(System.Drawing.Graphics,System.Windows.Forms.ToolStripItem)> | `g` or `item` |
```
or
```suggestion
| <xref:System.Windows.Forms.ToolStripItemRenderEventArgs.%23ctor(System.Drawing.Graphics,System.Windows.Forms.ToolStripItem)> | `g` |
| <xref:System.Windows.Forms.ToolStripItemRenderEventArgs.%23ctor(System.Drawing.Graphics,System.Windows.Forms.ToolStripItem)> | `item` |
```

I updated https://github.com/dotnet/docs/issues/26720 as well

docs/core/compatibility/7.0.md
gewarren: Good suggestion on the centering. I tried the div tag, but got a linter error so took it out. I think the many &nbsp; is just too ugly in the raw Markdown.

RussKie: > Good suggestion on the centering. I tried the div tag, but got a linter error so took it out. 

Do you know who owns the linter? May be the rules can be tweaked or exclusions added?

> I think the many   is just too ugly in the raw Markdown.

I think we should prioritise the end user experience, but I agree that adding "&nbsp;" may be a non scalable maintenance burden.

docs/core/compatibility/windows-forms/7.0/apis-throw-argumentnullexception.md
```diff
@@ -0,0 +1,45 @@
...
+- If your code handles <xref:System.NullReferenceException>, replace or add an additional handler for <xref:System.ArgumentNullException>.
+
+## Affected APIs
+
+The following table lists the affected APIs and specific parameters:
+
+| Method/property | Parameter name |
+|-|-|
+| <xref:System.Windows.Forms.ToolStripArrowRenderEventArgs.%23ctor(System.Drawing.Graphics,System.Windows.Forms.ToolStripItem,System.Drawing.Rectangle,System.Drawing.Color,System.Windows.Forms.ArrowDirection)> | `g` |
+| <xref:System.Windows.Forms.ToolStripItemRenderEventArgs.%23ctor(System.Drawing.Graphics,System.Windows.Forms.ToolStripItem)> | `g` and `item` |
```
RussKie, line 42: From a programmers perspective this reads as "the exception is thrown if both args are null", where as in reality the exception is thrown if any of the args is null.

PR #27874, opened 2022-01-18
fix HeapAffinitizeRanges name

dickens-code:
## Summary
fix HeapAffinitizeRanges name in runtimeconfig.json, as noted here: https://github.com/dotnet/runtime/blob/158b26e68681c3b28c6abeb3c9828506222f18fd/src/coreclr/gc/gcconfig.h#L95


gewarren (approved):
Thank you @dickens-code

PR #27881, opened 2022-01-19
"and" to "or"

gewarren:
Follow up to #27854.


RussKie (approved):

gewarren:
@RussKie I did try adding an exclusion for `div` to the markdown linter and that worked. But our build system also looks for HTML tags separately and logs a suggestion, which we are "graded" against for content health.

RussKie:
Bugger...

PR #27900, opened 2022-01-20
Add section for telemetry gathered by the Templating Engine

baronfel:
## Summary

After doing some planning about telemetry for the template engine, I realized that we did in fact collect telemetry from users instantiating templates, and we should declare that in the docs alongside the rest of the declarations.

### DIFF
--- a/docs/core/tools/telemetry.md
+++ b/docs/core/tools/telemetry.md
@@ -117,7 +117,7 @@ Except for `--verbosity` and `--sdk-package-version`, all the other values are h
 
 ### Template engine telemetry
 
-In addition to the options listed above, the `dotnet new` template instantiation command collects additional data for Microsoft-authored templates (after 2.1.100):
+The `dotnet new` template instantiation command collects additional data for Microsoft-authored templates, starting with .NET Core 2.1.100 SDK:
 
 * `--framework`
 * `--auth`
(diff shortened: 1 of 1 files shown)


docs/core/tools/telemetry.md
```diff
@@ -115,6 +115,13 @@ A subset of commands sends selected options if they're used, along with their va
 
 Except for `--verbosity` and `--sdk-package-version`, all the other values are hashed starting with .NET Core 2.1.100 SDK.
 
+### Template Engine telemetry
```
tdykstra, line 118:
```suggestion
### Template engine telemetry
```

docs/core/tools/telemetry.md
```diff
@@ -115,6 +115,13 @@ A subset of commands sends selected options if they're used, along with their va
...
+
+In addition to the options listed above, the `dotnet new` template instantiation command collects additional data for Microsoft-authored templates (after 2.1.100):
```
tdykstra, line 120:
```suggestion
The `dotnet new` template instantiation command collects additional data for Microsoft-authored templates, starting with .NET Core 2.1.100 SDK:
```

tdykstra (approved):
Looks good, I left a couple of suggestions.

baronfel:
Thanks for the suggestions, @tdykstra, I took them in their entirety.

PR #27910, opened 2022-01-20
The correct values for System.GC.CpuGroup in runtimeconfig.json should be true/false

cshung:
## Summary

The correct values for `System.GC.CpuGroup` in runtimeconfig.json should be true/false instead of 1/0.

As indicated [here](https://github.com/dotnet/runtime/blob/1d751ba8563fc099c6a75c41687d2f282a05f916/src/coreclr/gc/gcconfig.h#L78), `System.GC.CpuGroup` is a `BOOL_CONFIG`, and therefore it should be written as `true` or `false` in `runtimeconfig.json`. We did the same with `System.GC.Server`.

@dickens-code

### DIFF
--- a/docs/core/run-time-config/garbage-collector.md
+++ b/docs/core/run-time-config/garbage-collector.md
@@ -220,7 +220,7 @@ Example:
 
 | | Setting name | Values | Version introduced |
 | - | - | - | - |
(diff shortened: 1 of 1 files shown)


gewarren (approved):
Thank you @cshung

PR #27914, opened 2022-01-21
Use language keywords instead of upper case type names

333fred:
We were inconsistent about `String` (in extension parameters) vs `string` (in the array types). This was confusing to at least one person trying to learn extensions from this tutorial, and using `String` isn't great practice anyway, so I corrected the sample to always use `string` instead.


BillWagner:
Thanks @333fred 

The changes look great.

Can you add a .csproj file in this folder so it passes our CI build? I notice this file has several "Main" methods, so you'll need to pick one as the entry point as well. Otherwise, the default project file should work.

333fred:
@BillWagner done

BillWagner (approved):
Thanks again @333fred 

PR #27932, opened 2022-01-22
Fix link warnings to the C# standard

BillWagner:
The ECMA committee removed clause 4 from the standard. Its content was redundant.

As such, the anchor links moved.

Each commit updates links to one of the clauses, starting with (new) clause 6.


tdykstra (approved):
LGTM

PR #27957, opened 2022-01-26
Update simd.md

OrganizationUsername:
Fixed System.IndexOutOfRangeException

## Summary
Changed `for` loop comparison to remain within bounds.

### DIFF
--- a/docs/standard/simd.md
+++ b/docs/standard/simd.md
@@ -80,7 +80,7 @@ double[] SimdVectorProd(double[] left, double[] right)
     var offset = Vector<double>.Count;
     double[] result = new double[left.Length];
     int i = 0;
-    for (i = 0; i < left.Length; i += offset)
+    for (i = 0; i + offset < left.Length; i += offset)
     {
         var v1 = new Vector<double>(left, i);
         var v2 = new Vector<double>(right, i);
--- a/docs/standard/simd.md
+++ b/docs/standard/simd.md
@@ -72,25 +72,43 @@ var mResult = Matrix4x4.Multiply(m1, m2);
 
 The <xref:System.Numerics.Vector%601> gives the ability to use longer vectors. The count of a <xref:System.Numerics.Vector%601> instance is fixed, but its value <xref:System.Numerics.Vector%601.Count%2A?displayProperty=nameWithType> depends on the CPU of the machine running the code.
 
-The example below demonstrates how to calculate the element-wise product of two arrays using <xref:System.Numerics.Vector%601>.
+The following example demonstrates how to calculate the element-wise sum of two arrays using <xref:System.Numerics.Vector%601>.
 
 ```csharp
-double[] SimdVectorProd(double[] left, double[] right)
+double[] Sum(double[] left, double[] right)
 {
-    var offset = Vector<double>.Count;
-    double[] result = new double[left.Length];
-    int i = 0;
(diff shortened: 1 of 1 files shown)


gewarren:
@tannergooding Could you check this small code fix to prevent an IndexOutOfRangeException?

docs/standard/simd.md
```diff
@@ -80,7 +80,7 @@ double[] SimdVectorProd(double[] left, double[] right)
     var offset = Vector<double>.Count;
     double[] result = new double[left.Length];
     int i = 0;
-    for (i = 0; i < left.Length; i += offset)
+    for (i = 0; i + offset < left.Length; i += offset)
```
tannergooding, line 83: The change here is correct, but its also not the "best" way to do all this and misses another case where `right.Length < left.Length`.

If we were to update the sample to avoid bugs, etc; I'd probably write this as the below. This switches to using `Sum` which I believe is simpler to understand/validate and which is always accelerated (multiplication is generally accelerated but there are edge cases where it isn't). It validates the inputs and tracks the length in a fashion that is more extensible for the code to handle other interesting scenarios if that's ever desired.

---------------------------

> The example below demonstrates how to calculate the element-wise sum of two arrays using <xref:System.Numerics.Vector%601>.
```csharp
double[] Sum(double[] left, double[] right)
{
    if (left is null)
    {
        throw new ArgumentNullException(nameof(left));
    }

    if (right is null)
    {
        throw new ArgumentNullException(nameof(right));
    }

    if (left.Length != right.Length)
    {
        throw new ArgumentException($"{nameof(left)} and {nameof(right)} are not the same length");
    }

    int length = left.Length;
    double[] result = new double[length];

    // Get the number of elements that can't be processed in the vector
    // NOTE: Vector<T>.Count is a JIT time constant and will get optimized accordingly
    int remaining = length % Vector<double>.Count;

    for (int i = 0; i < length - remaining; i += Vector<double>.Count)
    {
        var v1 = new Vector<double>(left, i);
        var v2 = new Vector<double>(right, i);
        (v1 * v2).CopyTo(result, i);
    }

    for (int i = length - remaining; i < length; i++)
    {
        result[i] = left[i] * right[i];
    }

    return result;
}
```

Treit: @tannergooding isn't your example still doing multiplication instead of computing the sum?

tannergooding: Yes, fixed.

gewarren:
@OrganizationUsername Can you update the code sample to the one that Tanner suggested?

gewarren:
This PR is similar to #26196.

BillWagner:
ping @gewarren @tannergooding 

Should this be merged or closed?

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/docs?pullRequest=27957) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/docs?pullRequest=27957) before we can accept your contribution.<br/><br/>:x: OrganizationUsername [sign now](https://cla.dotnetfoundation.org/dotnet/docs?pullRequest=27957)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/docs?pullRequest=27957) it.</sub>

gewarren (approved):

PR #27967, opened 2022-01-26
Remove null coalescing in LINQ

### LINKED ISSUE #27831: Example not accepted by Visual Studio 2019
When I try updating my code in line with this example. Visual Studio highlights the subpet?. bit with the message "CS8072: An expression tree lambda may not contain a null propagating operator.". If LINQ does not allow this, why are you showing it in your example?

---
#### Document Details

* ID: d433265d-15f4-bd2e-a3d1-8148a3968783
* Version Independent ID: 75d2c893-1d07-f836-62f1-e767edfcebac
* Content: [Perform left outer joins (LINQ in C#)](https://docs.microsoft.com/en-us/dotnet/csharp/linq/perform-left-outer-joins)
* Content Source: [docs/csharp/linq/perform-left-outer-joins.md](https://github.com/dotnet/docs/blob/main/docs/csharp/linq/perform-left-outer-joins.md)
* Product: **dotnet-csharp**
* Technology: **csharp-linq**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

IEvangelist:
## Summary

Remove null coalescing in LINQ

Fixes #27831

✔️ [Preview: Perform left outer joins](https://review.docs.microsoft.com/en-us/dotnet/csharp/linq/perform-left-outer-joins?branch=pr-en-us-27967)

Hey @BillWagner, I'm curious about how you feel about not including a .csproj in this case? If we do, the code will break as it's actually not valid C# and won't compile as is. It's not a valid top-level statement app. [Here is the error](https://github.com/dotnet/docs/runs/4953280225?check_suite_focus=true).


BillWagner:
@IEvangelist 

My first impression is to modify the organization of the code such that it can compile (even if that means using snippet tags to extract only some of it for the article).

If that's not a major time investment, we should keep slowing improving.

BillWagner (approved):
This LGTM @IEvangelist 

Let's :shipit:

PR #27970, opened 2022-01-26
Update title and content to reference .NET 6

jongalloway:
## Summary

Update title and content to reference .NET 6

### DIFF
--- a/docs/core/porting/index.md
+++ b/docs/core/porting/index.md
@@ -1,6 +1,6 @@
 ---
-title: Port from .NET Framework to .NET 5
-description: Understand the porting process and discover tools you may find helpful when porting a .NET Framework project to .NET 5 (and .NET Core 3.1).
+title: Port from .NET Framework to .NET 6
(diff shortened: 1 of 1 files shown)


docs/core/porting/index.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: Port from .NET Framework to .NET 5
-description: Understand the porting process and discover tools you may find helpful when porting a .NET Framework project to .NET 5 (and .NET Core 3.1).
+title: Port from .NET Framework to .NET 6
```
Youssef1313, line 2: Does it make sense to mention only ".NET", or ".NET 5+"?

jongalloway:
Closing as dupe of https://github.com/dotnet/docs/pull/27972

docs/core/porting/index.md
adegeo: @Youssef1313 We'll eventually revisit these titles when the manager in charge of this content gets back to restructuring/editing/adding stuff. For now I think we'll just call out 6 directly.

PR #27975, opened 2022-01-27
Add note that IDE0050 is now a refactoring

Youssef1313:
https://github.com/dotnet/roslyn/commit/584878b3903b8222f29e332a4a3c66b5fc798730

### DIFF
--- a/docs/fundamentals/code-analysis/style-rules/ide0050.md
+++ b/docs/fundamentals/code-analysis/style-rules/ide0050.md
@@ -15,6 +15,9 @@ dev_langs:
 ---
 # Convert anonymous type to tuple (IDE0050)
 
+> [!IMPORTANT]
+> This analyzer was changed to a refactoring in Visual Studio 2022 17.0
+
 |Property|Value|
 |-|-|
 | **Rule ID** | IDE0050 |
--- a/docs/fundamentals/code-analysis/style-rules/ide0050.md
+++ b/docs/fundamentals/code-analysis/style-rules/ide0050.md
@@ -16,7 +16,7 @@ dev_langs:
 # Convert anonymous type to tuple (IDE0050)
 
 > [!IMPORTANT]
-> This analyzer was changed to a refactoring in Visual Studio 2022 17.0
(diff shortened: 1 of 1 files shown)


Youssef1313:
FYI @CyrusNajmabadi

gewarren:
@Youssef1313 How is the new refactoring different to this one that's been around for a few years? https://docs.microsoft.com/en-us/visualstudio/ide/reference/convert-anonymous-type-to-tuple?view=vs-2022

Youssef1313:
The screenshot in that article actually shows IDE0050, so that is the analyzer/codefix that was documented here.

With it now being refactoring, you cannot enforce it on build or change a severity.

gewarren:
> The screenshot in that article actually shows IDE0050, so that is the analyzer/codefix that was documented here.
> 
> With it now being refactoring, you cannot enforce it on build or change a severity.

Okay, so I guess that documentation doesn't need to be updated then.

Youssef1313:
There is no IDE0050 anymore.

I even see @CyrusNajmabadi suggested deleting the documentation in his PR, which we may do if there is a good redirection.

Youssef1313:
@gewarren Actually the link you provided looks like a good redirection. Makes sense to retire the article completely?

docs/fundamentals/code-analysis/style-rules/ide0050.md
```diff
@@ -15,6 +15,9 @@ dev_langs:
 ---
 # Convert anonymous type to tuple (IDE0050)
 
+> [!IMPORTANT]
+> This analyzer was changed to a refactoring in Visual Studio 2022 17.0
```
gewarren, line 19:
```suggestion
> This style rule was removed and converted to a Visual Studio refactoring in Visual Studio 2022. For information about the refactoring, see [Convert anonymous type to tuple](/visualstudio/ide/reference/convert-anonymous-type-to-tuple).
```

gewarren:
> @gewarren Actually the link you provided looks like a good redirection. Makes sense to retire the article completely?

I don't think we should delete the article, because people may be using a NuGet analyzer package or the .NET 5 set of analyzers.

gewarren (approved):
Thanks for catching this!

PR #27982, opened 2022-01-27
Update nullable-references.md

IEvangelist:
I missed these two, they're incorrect as well and should be `enable` and `disable`, not `enabled` and `disabled`.


BillWagner (approved):
Thanks @IEvangelist 

Approving now.

PR #27985, opened 2022-01-28
Fix a few typos

GitHubPang:


BillWagner (approved):
Thanks again @GitHubPang 

PR #27993, opened 2022-01-28
Fix protocolMapping ex. to refer to correct config

DreadLordMikey:
Fix bindingConfiguration reference in ProtocolMapping example so that it correctly references previous code example.

## Summary

The protocolMapping example refers to a bindingConfiguration mentioned earlier in this document. The current bindingConfiguration value is "MyBindingConfiguration", which is incorrect; the previous example specifies this value as "MyBindingConfig". This fix makes them match, so that the example is clear and correct.


tdykstra (approved):
Thanks, @DreadLordMikey

PR #28009, opened 2022-01-31
Disambiguate TFMs from Preprocessor Symbols

fowl2:
## Summary

This confused me for a while trying to work on as it's included right under the project file section of https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-a-target-framework.

Ps. I *think* the link URLs are correct as per [Use links in documentation](https://docs.microsoft.com/en-us/contribute/how-to-write-links), but it seems like GitHub doesn't like cross-repo links.


tdykstra (approved):
This looks good, thanks for the contribution!

PR #28015, opened 2022-01-31
fix order of rules by rule number

heikow10:
## Summary

The entries are now sorted by the rule number.


gewarren (approved):
Thanks @heikow10

PR #28022, opened 2022-01-31
Fix unrecognized tag [!INFORMATION]

BLM16:
## Summary

[!INFORMATION] doesn't render a box like [!NOTE] or [!TIP]. It has been changed to [!NOTE].


BLM16:
I wasn't sure if I should be updating the `ms.date`. Should this be changed for all edits?

Issue #27310, opened 2021-11-30
Update code snippets to use .NET 6 features where appropriate


tdykstra:
.NET 6 features to implement in code snippets where appropriate:
* Implicit usings
* Nullable reference types (NRT)
* File-scoped namespace declarations
* Top-level statements

The following project files target .NET 6 but omit implicit usings and/or NRT:

* [x] https://github.com/dotnet/docs/blob/main/docs/csharp/fundamentals/program-structure/snippets/top-level-statements-1/top-level-statements.csproj
* [x] https://github.com/dotnet/docs/blob/main/docs/csharp/fundamentals/program-structure/snippets/top-level-statements-2/top-level-statements.csproj
* [x] https://github.com/dotnet/docs/blob/main/docs/csharp/fundamentals/program-structure/snippets/top-level-statements-3/top-level-statements.csproj
* [x] https://github.com/dotnet/docs/blob/main/docs/csharp/fundamentals/program-structure/snippets/top-level-statements-4/top-level-statements.csproj
* [x] https://github.com/dotnet/docs/blob/main/docs/csharp/fundamentals/program-structure/snippets/top-level-statements-5/top-level-statements.csproj
* [x] https://github.com/dotnet/docs/blob/main/docs/csharp/fundamentals/types/snippets/anonymous-types/anonymous-types.csproj
* [x] https://github.com/dotnet/docs/blob/main/docs/csharp/fundamentals/types/snippets/namespaces/namespaces.csproj
* [x] https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/keywords/snippets/keywords.csproj
* [x] https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/operators/snippets/patterns/patterns.csproj
* [x] https://github.com/dotnet/docs/blob/main/docs/csharp/snippets/iterators/iterators.csproj
* [x] https://github.com/dotnet/docs/blob/main/docs/csharp/whats-new/tutorials/snippets/record-types/record-types.csproj
* [x] https://github.com/dotnet/docs/blob/main/docs/standard/serialization/snippets/system-text-json-supported-collection-types/csharp/SystemTextJsonSamples.csproj
* [x] https://github.com/dotnet/docs/blob/main/docs/standard/serialization/snippets/system-text-json-use-dom-utf8jsonreader-utf8jsonwriter/csharp/SystemTextJsonSamples.csproj

The following project files target .NET Core 3.1:

[...]

PR #28030, opened 2022-02-01
Implement NRT and implicit usings

### LINKED ISSUE #27693: Non-nullable property 'Summary' must contain a non-null value when exiting constructor. CS8618
Hi,

When creating a new C# 6 project the `<Nullable>enable</Nullable>` property causes warnings, e.g.
_"Non-nullable property 'Summary' must contain a non-null value when exiting constructor. Consider declaring the property as nullable. "_
1. Could you mention something about it in the docs?
2. Can you also mention and/or show an example of using record structs? This is the preferred way in .NET 6 & C# 9 and up?

Thanks

---
#### Document Details

* Content Source: [docs/standard/serialization/system-text-json-how-to.md](https://github.com/dotnet/docs/blob/main/docs/standard/serialization/system-text-json-how-to.md)
* GitHub Login: @tdykstra

tdykstra:
Fixes #27693
Contributes to #27310

### DIFF
--- a/docs/standard/serialization/snippets/system-text-json-how-to/csharp/DeserializeRequiredProperty.cs
+++ b/docs/standard/serialization/snippets/system-text-json-how-to/csharp/DeserializeRequiredProperty.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Text.Json;
+﻿using System.Text.Json;
 
 namespace SystemTextJsonSamples
 {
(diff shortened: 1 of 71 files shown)


docs/standard/serialization/snippets/system-text-json-how-to/csharp/DeserializeRequiredProperty.cs
```diff
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Text.Json;
+﻿using System.Text.Json;
```
BillWagner, line 1: Not necessary, but since all of these samples use `System.Text.Json`, you could consider adding it to the global usings.

BillWagner (approved):
This LGTM @tdykstra 

You can :shipit: when ready.

docs/standard/serialization/snippets/system-text-json-how-to/csharp/DeserializeRequiredProperty.cs
tdykstra: That's a good idea. I'll do that for the next batches of snippet files but will keep the STJ `using` in the files that have the whole file including `using`s displayed in the docs.

PR #28054, opened 2022-02-02
Add SG dependencies section

Youssef1313:
@sharwell @jmarolf can someone review? Thanks!

### DIFF
--- a/docs/csharp/roslyn-sdk/source-generators-overview.md
+++ b/docs/csharp/roslyn-sdk/source-generators-overview.md
@@ -139,6 +139,9 @@ In this guide, you'll explore the creation of a source generator using the <xref
 
 ## Dependencies in source generators
 
+> [!IMPORTANT]
+> Source generator package dependencies must be `netstandard2.0`.
+
 If you want to add package dependencies to your source generator, you will need extra work in the *csproj* file:
 
 - Each `PackageReference` should have `GeneratePathProperty="true"` and `PrivateAssets="all"`.
--- a/docs/csharp/roslyn-sdk/source-generators-overview.md
+++ b/docs/csharp/roslyn-sdk/source-generators-overview.md
@@ -137,6 +137,32 @@ In this guide, you'll explore the creation of a source generator using the <xref
 
     :::image type="content" source="media/source-generators/source-generated-program.png" lightbox="media/source-generators/source-generated-program.png" alt-text="Visual Studio: Auto-generated Program.g.cs file.":::
 
+## Dependencies in source generators
+
+If you want to add package dependencies to your source generator, you will need extra work in the *csproj* file:
+
+- Each `PackageReference` should have `GeneratePathProperty="true"` and `PrivateAssets="all"`.
+- You need to add the following target:
+
+    ```xml
+    <Target Name="GetDependencyTargetPaths">
+      <ItemGroup>
+        <!-- Do this for every package you reference. The "PKG" prefix remains the same, while any periods in package name is replaced with underscores -->
+        <TargetPathWithTargetPlatformMoniker Include="$(PKGThePackageYouReference)\lib\netstandard2.0\ThePackageYouReference.dll" IncludeRuntimeDependency="false" />
+      </ItemGroup>
+    </Target>
+    ```
+
+- The `GetTargetPath` target should depend on the new target you added in the previous step, you do that as follows:
+
+    ```xml
+    <PropertyGroup>
+      <GetTargetPathDependsOn>$(GetTargetPathDependsOn);GetDependencyTargetPaths</GetTargetPathDependsOn>
+    </PropertyGroup>
+    ```
+
+For a complete *csproj* example, see [CSharpSourceGeneratorSamples.csproj](https://github.com/dotnet/roslyn-sdk/blob/5496633deaa473b7fdcf686ac4da3a1c9a58a6f8/samples/CSharp/SourceGenerators/SourceGeneratorSamples/CSharpSourceGeneratorSamples.csproj#L13-L30).
+
 ## Next steps
 
 The [Source Generators Cookbook](https://github.com/dotnet/roslyn/blob/main/docs/features/source-generators.cookbook.md) goes over some of these examples with some recommended approaches to solving them. Additionally, we have a [set of samples available on GitHub](https://github.com/dotnet/roslyn-sdk/tree/main/samples/CSharp/SourceGenerators) that you can try on your own.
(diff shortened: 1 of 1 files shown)


docs/csharp/roslyn-sdk/source-generators-overview.md
```diff
@@ -137,6 +137,32 @@ In this guide, you'll explore the creation of a source generator using the <xref
...
+
+If you want to add package dependencies to your source generator, you will need extra work in the *csproj* file:
+
+- Each `PackageReference` should have `GeneratePathProperty="true"` and `PrivateAssets="all"`.
+- You need to add the following target:
+
+    ```xml
+    <Target Name="GetDependencyTargetPaths">
+      <ItemGroup>
+        <!-- Do this for every package you reference. The "PKG" prefix remains the same, while any periods in package name is replaced with underscores -->
```
Youssef1313, line 150:
```suggestion
        <!-- Do this for every package you reference. The "PKG" prefix remains the same, while any periods in package name are replaced with underscores -->
```

docs/csharp/roslyn-sdk/source-generators-overview.md
```diff
@@ -137,6 +137,32 @@ In this guide, you'll explore the creation of a source generator using the <xref
...
+If you want to add package dependencies to your source generator, you will need extra work in the *csproj* file:
+
+- Each `PackageReference` should have `GeneratePathProperty="true"` and `PrivateAssets="all"`.
+- You need to add the following target:
+
+    ```xml
+    <Target Name="GetDependencyTargetPaths">
+      <ItemGroup>
+        <!-- Do this for every package you reference. The "PKG" prefix remains the same, while any periods in package name are replaced with underscores -->
+        <TargetPathWithTargetPlatformMoniker Include="$(PKGThePackageYouReference)\lib\netstandard2.0\ThePackageYouReference.dll" IncludeRuntimeDependency="false" />
```
sharwell, line 151: ❗ This won't work due to the way Roslyn's assembly loader works. All assemblies must be located in the _same folder on disk_ in order for them to be usable. Naturally, the easiest location ends up being `$(OutputPath)`, like this:

https://github.com/KathleenDollard/JackFruit/blob/3855504030438acc9ce79e507be6b941d4d36289/JackFruitAppModel.CSharp/JackfruitAppModel.CSharp.fsproj#L37-L46

sharwell (changes requested):

docs/csharp/roslyn-sdk/source-generators-overview.md
Youssef1313: @sharwell Does this mean the source generator sample I quoted the code from is incorrect?

Youssef1313: @sharwell btw, the sdk itself uses the directory as the source generator sample:

https://github.com/dotnet/sdk/blob/d694a5e325c88b44e8054874cacde0ad391e8634/src/RazorSdk/SourceGenerators/Microsoft.NET.Sdk.Razor.SourceGenerators.csproj#L30

sharwell: > Does this mean the source generator sample I quoted the code from is incorrect?

Likely yes. @genlu where is the issue filed where people noticed the behavior change when we restricted assembly load locations?

sharwell: > the sdk itself uses the directory as the source generator sample

I believe this source generator is special and doesn't run in the same way as others. It might not be impacted.

Youssef1313: @sharwell Are you referring to https://github.com/dotnet/roslyn/issues/56442, which was fixed by https://github.com/dotnet/roslyn/pull/56662?

jmarolf: I do not think our current samples are correct here. I recommend something like this as a means of deploying dependencies: https://github.com/jmarolf/generator-start/blob/d0db05af009b29406caed7bde007d472fffc8c2f/src/WebAssembly.Generators/WebAssembly.Generators.csproj#L35. Manually placing all the dlls in the right place has been less error-prone in my experience. There is also the complexity that today the dependency _must_ be netstandard2.0 and _cannot_ be a reference assembly. It must be loadable and runnable on both .NET Framework and .NET Core. This is hard to know about a dependency and one of the reasons I have shied away from official documentation on this topic as troubleshooting quickly devolves into needing to understand how to make your dependencies loadable and runnable in a variety of contexts. Still applaud the effort though and its clear we _must_ have some guidance for people.

My trepidation is implying that this is something that will "just work". That couldn't be further from the case. The only way to make it work is to know, in intimate detail, how your source generator/ analyzer is being loaded, how its runtime dependencies are being identified, and how to correctly package all runtime dependencies so that they are discoverable. Nuget and MSBuild do a respectable job of abstracting all of this away in .NET so most developers will not be familiar with how to do this "manually" with their own build targets. Part of the hesitation here is also that it is a much simpler message to say "third party dependencies are not supported, wait for us to add first-class support in nuget and msbuild" but that may be too long in coming.

Youssef1313: > I do not think our current samples are correct here

@jmarolf The line you linked is almost the same as the sample.

I'll add a note about the `netstandard2.0` requirement.

Youssef1313: @sharwell I tested now. The `OutputPath` approach didn't work while the other worked.

Youssef1313: @sharwell https://github.com/Youssef1313/generator-dependencies

Youssef1313: Ping @sharwell @jmarolf

Bosch-Eli-Black: > > Does this mean the source generator sample I quoted the code from is incorrect?
> 
> Likely yes. @genlu where is the issue filed where people noticed the behavior change when we restricted assembly load locations?

@sharwell In case you'd still like a link to the issue, I believe it's https://github.com/dotnet/roslyn/issues/56442 🙂

BillWagner:
@Youssef1313 

I'm closing this as very stale. It's been more than 2 years since the last comment, and there are now conflicts. If you want to reopen the PR, address the conflicts and the comments, we'll take another look.

PR #28072, opened 2022-02-03
System.CommandLine documentation

### LINKED ISSUE #25030: Document System.CommandLine library
Document the System.CommandLine library and its use in the .NET CLI.

https://devblogs.microsoft.com/dotnet/announcing-net-6-preview-1/#net-cli
https://github.com/dotnet/command-line-api/blob/main/docs/Features-overview.md

Related to #27339

### LINKED ISSUE #27339: Links to command line argument parsing
I know that .Net Core added a really cool command line argument parser, but when I search the documentation, all I can find is examples how how to use the args array and parse it manually.  Google doesn't help because it prioritizes old command line parser nuget packages. there should be a link on this page to the articles on using the command line parser.  It's such a shame, -- I'm having such a hard time finding the docs, I'll probably fall back to the 3rd party package that I used to use in the pre-core days.

---
#### Document Details

* ID: cde79d6e-21ab-c158-8f41-1821ad920628
* Version Independent ID: 69a297cb-677e-2b05-0187-0d0ec0890b00
* Content: [How to display command-line arguments](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/tutorials/how-to-display-command-line-arguments)
* Content Source: [docs/csharp/fundamentals/tutorials/how-to-display-command-line-arguments.md](https://github.com/dotnet/docs/blob/main/docs/csharp/fundamentals/tutorials/how-to-display-command-line-arguments.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

### LINKED ISSUE #28071: System.CommandLine documentation
Publish conceptual documentation for the System.CommandLine library,

tdykstra:
Fixes #25030
Fixes #27339

-- | -- | --
Content | docs/standard/commandline/index.md | https://review.docs.microsoft.com/en-us/dotnet/standard/commandline/index?branch=pr-en-us-28072
Content | docs/standard/commandline/get-started-tutorial.md | https://review.docs.microsoft.com/en-us/dotnet/standard/commandline/get-started-tutorial?branch=pr-en-us-28072
Content | docs/standard/commandline/syntax.md | https://review.docs.microsoft.com/en-us/dotnet/standard/commandline/syntax?branch=pr-en-us-28072
Content | docs/standard/commandline/define-commands.md | https://review.docs.microsoft.com/en-us/dotnet/standard/commandline/define-commands?branch=pr-en-us-28072
Content | docs/standard/commandline/model-binding.md | https://review.docs.microsoft.com/en-us/dotnet/standard/commandline/model-binding?branch=pr-en-us-28072
Content | docs/azure/sdk/dependency-injection.md | https://review.docs.microsoft.com/en-us/dotnet/azure/sdk/dependency-injection?branch=pr-en-us-28072
Content | docs/standard/commandline/customize-help.md | https://review.docs.microsoft.com/en-us/dotnet/standard/commandline/customize-help?branch=pr-en-us-28072
Content | docs/standard/commandline/dependency-injection.md | https://review.docs.microsoft.com/en-us/dotnet/standard/commandline/dependency-injection?branch=pr-en-us-28072
Content | docs/standard/commandline/handle-termination.md | https://review.docs.microsoft.com/en-us/dotnet/standard/commandline/handle-termination?branch=pr-en-us-28072
Content | docs/standard/commandline/tab-completion.md | https://review.docs.microsoft.com/en-us/dotnet/standard/commandline/tab-completion?branch=pr-en-us-28072
Content | docs/standard/commandline/use-middleware.md | https://review.docs.microsoft.com/en-us/dotnet/standard/commandline/use-middleware?branch=pr-en-us-28072


docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,572 @@
...
+As this example illustrates, the value of the option may be explicit (`quiet` for `--verbosity`) or implicit (nothing follows `--global`). Options that have no value specified are typically Boolean parameters that default to `true` if the option is specified on the command line.
+
+For some Windows command-line apps, you identify an option by using a leading slash (`/`) with the option name. For example:
+
+```console
+msbuild /version
+        ^------^
+```
+
+The POSIX style (two hyphens before long-form option names) is the default for `System.CommandLine`, but the library can be configured to use other prefixes.
```
jonsequitur, line 80: Is there a code example for this? The way it's configured is just to specify the literal string, e.g. `new Option<int>("/version")`

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,572 @@
...
+Arguments can have expected types, and the app errors if the argument can't be parsed into the expected type. For example, the following command errors because "silent" isn't one of the valid values for `--verbosity`:
+
+```dotnetcli
+dotnet build --verbosity silent
+```
+
+```output
+System.InvalidOperationException: Cannot parse
+argument 'silent' for option '-v' as expected type
+Microsoft.DotNet.Cli.VerbosityOptions.
```
baronfel, line 117:
```
Cannot parse argument 'silent' for option '-v' as expected type 'Microsoft.DotNet.Cli.VerbosityOptions'. Did you mean one of the following?
Detailed
Diagnostic
Minimal
Normal
Quiet
```

baronfel: This is the new wording that System.CommandLine emits for enum suggestions in the most recent public beta.

docs/standard/commandline/get-started-tutorial.md
```diff
@@ -0,0 +1,432 @@
...
+
+## Install the System.CommandLine package
+
+* Run the following command:
+
+  ```dotnetcli
+  dotnet add package System.CommandLine --prerelease
+  ```
+
+  The `-prerelease` option is necessary because the library is still in beta.
```
jonsequitur, line 57:
```suggestion
  The `--prerelease` option is necessary because the library is still in beta.
```

docs/standard/commandline/tab-completion.md
```diff
@@ -0,0 +1,82 @@
...
+
+  * For PowerShell, add the contents of [*dotnet-suggest-shim.ps1*](https://github.com/dotnet/command-line-api/blob/master/src/System.CommandLine.Suggest/dotnet-suggest-shim.ps1) to your PowerShell profile. You can find the expected path to your PowerShell profile by running the following command in your console:
+
+    ```console
+    echo $profile
+    ```
+
+Once the user's shell is set up, completions will work for all apps that are built by using `System.CommandLine`.
+
+For *cmd.exe* on Windows there is no pluggable tab completion mechanism, so no shim script is available. For other shells, [look for a GitHub issue that is labeled `Area-Completions`](https://github.com/dotnet/command-line-api/issues?q=is%3Aissue+is%3Aopen+label%3A%22Area-Completions%22). If you don't find an issue, you can [open a new one](https://github.com/dotnet/command-line-api/issues).
```
baronfel, line 41: This is a very helpful section - I would suggest also calling out the shim script for `zsh`, because it's the default shell for MacOS these days.

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,572 @@
...
+| `--verbosity`    | Option for install command        |
+| `quiet`          | Argument for `--verbosity` option |
+
+A token can contains spaces if it's enclosed in quotation marks (`"`). Here's an example:
+
+```console
+dotnet tool search "ef migrations add"
+```
+
+## Commands
```
jonsequitur, line 45: Commands are often commonly referred to as "verbs". Mentioning that might help some people understand the concept.

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,572 @@
...
+
+## Case sensitivity
+
+Command and option names and aliases are case-sensitive by default according to POSIX convention, and `System.CommandLine` follows this convention. If you want your CLI to be case insensitive, define aliases for the various casing alternatives. For example, `--additional-probing-path` could have aliases `--Additional-Probing-Path` and `--ADDITIONAL-PROBING-PATH`.
+
+In some command-line tools, a difference in casing specifies a difference in function. For example, [`git clean -X`](https://git-scm.com/docs/git-clean#Documentation/git-clean.txt--X) behaves differently than [`git clean -x`](https://git-scm.com/docs/git-clean#Documentation/git-clean.txt--x). The .NET CLI is all lowercase.
+
+## The `--` token
+
+POSIX convention interprets the double-dash (`--`) token as an escape mechanism. Everything to the right of the double-dash token is interpreted as arguments for the command. This functionality can be used to submit arguments that look like options, since it prevents them from being interpreted as options.
```
jonsequitur, line 175: The term "following" might be more appropriate for people using right-to-left languages in their terminal.

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,572 @@
...
+
+Suppose *myapp* takes a `message` argument, and you want the value of `message` to be `--interactive`. The following command line won't work:
+
+```console
+myapp --interactive
+```
+
+If `myapp` has an `--interactive` option, this input will be interpreted as referring to that option. If the app doesn't have an `--interactive` option, the result will be an unmatched option error.
```
jonsequitur, line 183: This second sentence isn't accurate. If there's no `--interactive` option defined, then this will be parsed as an argument. The `--` escape mechanism is only necessary when there's a matching identifier in the command line app.

docs/standard/commandline/syntax.md
tdykstra: Changing the text to

```
`System.CommandLine` supports both POSIX and Windows prefix conventions. When you [configure an option](define-commands.md#define-options), you specify the option name including the prefix.
```
The link takes you to a code example.

docs/standard/commandline/syntax.md
tdykstra: I mention that under the subcommand section -- should the root command also be called a verb?

jonsequitur: No, I think you have it right to just associate that term with subcommands.

docs/standard/commandline/syntax.md
baronfel: It might be worth calling out that if you want to support short forms of enums, it's incumbent on you to provide that variations. The 'full' verbosity enum for the .NET CLI includes the following enum cases:


```csharp
    public enum VerbosityOptions
    {
        quiet,
        q,
        minimal,
        m,
        normal,
        n,
        detailed,
        d,
        diagnostic,
        diag
    }
```

but that has its own complications - comparisons and verbosity-level checks in the codebase are more complex as a result. We have to check conditions like `verbosity != VerbosityOptions.diagnostic && verbosity != VerbosityOptions.diag`, for example.

docs/standard/commandline/customize-help.md
```diff
@@ -0,0 +1,115 @@
...
+---
+# How to customize help in apps that are built with the System.Commandline library
+
+You can customize help for a specific command, option, or argument, and you can add or replace whole help sections.
+
+The examples in this article work with the following command-line application:
+
+:::code language="csharp" source="snippets/customize-help/csharp/Program.cs" id="original" :::
+
+When help is requested, the following output is produced:
```
KathleenDollard, line 20: Maybe:

```suggestion
Without customization, when help is requested, the following output is produced:
```

docs/standard/commandline/customize-help.md
```diff
@@ -0,0 +1,115 @@
...
+
+Usage:
+  scl [options]
+
+Options:
+  --file <file>                                               The file to print out. [default: scl.runtimeconfig.json]
+  --light-mode                                                Determines whether the background color will be black or
+                                                              white
+  --color                                                     Specifies the foreground color of console output
+  <Black|Blue|Cyan|DarkBlue|DarkCyan|DarkGray|DarkGreen|Dark  [default: White]
```
KathleenDollard, line 34: Do we have an issue for finding a solution here that is less ugly (a System.CommandLine, not docs question)

docs/standard/commandline/snippets/customize-help/csharp/Program.cs
```diff
@@ -0,0 +1,183 @@
...
+                Console.WriteLine($"File contents:\n{file?.OpenText().ReadToEnd()}");
+            },
+            fileOption,
+            lightModeOption,
+            foregroundColorOption);
+
+        await rootCommand.InvokeAsync(args);
+        // </original>
+        Console.WriteLine("Default help");
+        await rootCommand.InvokeAsync("--light-mode false --color Yellow");
```
KathleenDollard, line 55: It may be easier for the user if we include -h here, even if the specific CLI design does not require it.

KathleenDollard: It's a Friday afternoon, but doesn't this print in the correct order due to oddities in async?

docs/standard/commandline/snippets/customize-help/csharp/Program.cs
```diff
@@ -0,0 +1,183 @@
...
+            lightModeOption,
+            foregroundColorOption);
+
+        // <first2columns>
+        var parser = new CommandLineBuilder(rootCommand)
+                .UseDefaults()
+                .UseHelp(ctx =>
+                {
+                    ctx.HelpBuilder.CustomizeSymbol(fileOption,
+                        firstColumnText: "--file <FILEPATH>");
```
KathleenDollard, line 99: I really like the second example, but I believe this could be done with the options ArgumentDisplayName (not sure I am remembering that name precisely.)

docs/standard/commandline/snippets/customize-help/csharp/Program.cs
```diff
@@ -0,0 +1,183 @@
...
+                        secondColumnText: "Specifies the foreground color. " +
+                            "Choose a color that provides enough contrast " +
+                            "with the background color. " +
+                            "For example, a yellow foreground can't be read " +
+                            "against a light mode background.");
+                    ctx.HelpBuilder.CustomizeLayout(
+                        _ =>
+                            HelpBuilder.Default
+                                .GetLayout()
+                                .Skip(1) // Skip the default description section.
```
KathleenDollard, line 167: If this is the default _command_ description section, it may help to add that word.

docs/standard/commandline/customize-help.md
```diff
@@ -0,0 +1,115 @@
...
+
+```output
+Description:
+  Read a file
+
+Usage:
+  scl [options]
+
+Options:
+  --file <FILEPATH>                                           The file to print out. [default: scl.runtimeconfig.json]
```
KathleenDollard, line 65: See note on this example in the snippet (and other comments)

docs/standard/commandline/customize-help.md
```diff
@@ -0,0 +1,115 @@
...
+
+Change the layout by adding a call to `HelpBuilder.CustomizeLayout` in the lambda passed to the `CommandLineBuilder.UseHelp` method:
+
+:::code language="csharp" source="snippets/customize-help/csharp/Program.cs" id="description" highlight="13-21" :::
+
+The `HelpBuilder.Default` class lets you reuse pieces of existing help formatting functionality and compose them into your custom help.
+
+The help output now looks like this:
+
+```output
```
KathleenDollard, line 90: I love this example!

docs/standard/commandline/snippets/define-commands/csharp/Program.cs
```diff
@@ -0,0 +1,18 @@
...
+
+class Program
+{
+    static async Task Main(string[] args)
+    {
+        var rootCommand = new RootCommand("Sample command-line app");
+
+        rootCommand.SetHandler(() =>
+        {
+            Console.WriteLine(rootCommand.Description);
```
KathleenDollard, line 12: Would the intent be more clear if this was "Hello, World". I do not think users would ever output the description.

docs/standard/commandline/snippets/define-commands/csharp/Program2.cs
```diff
@@ -0,0 +1,191 @@
...
+        {
+            Console.WriteLine($"<delay> argument = {delayArgumentValue}");
+            Console.WriteLine($"<message> argument = {messageArgumentValue}");
+        },
+        delayArgument, messageArgument);
+
+        await rootCommand.InvokeAsync(args);
+        // </definearguments>
+        Console.WriteLine("Sample command line.");
+        await rootCommand.InvokeAsync("21 \"Hello world!\"");
```
KathleenDollard, line 40: I was surprised this was not a lambda

docs/standard/commandline/snippets/define-commands/csharp/Program2.cs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    static async Task FromAmong(string[] args)
+    {
+        // <staticlist>
+        var languageOption = new Option<string>(
+            "--language",
+            "An option that that must be one of the values of a static list.")
+                .FromAmong(
+                    "csharp",
```
KathleenDollard, line 170: Can we include VB on principle, unless the context is one we know VB doesn't support (I'm jumping around a bit in my review)/

docs/standard/commandline/define-commands.md
```diff
@@ -0,0 +1,208 @@
...
+```console
+myapp -f net6.0
+myapp --framework net6.0
+```
+
+Command aliases work the same way.
+
+```csharp
+var command = new Command("serialize");
+command.AddAlias("serialise");
```
KathleenDollard, line 102: Lovely example

docs/standard/commandline/define-commands.md
```diff
@@ -0,0 +1,208 @@
...
+  -?, -h, --help          Show help and usage information
+```
+
+If the command line for this example app doesn't include `--endpoint`, an error message is displayed and the command handler isn't called:
+
+```output
+Option '--endpoint' is required.
+```
+
+## Hidden commands, options, and arguments
```
KathleenDollard, line 135: Somewhere lets add why you would do this:

The most common reason to do this is to continue to support a feature you would like to deprecate. It can also be used for admin or preview features.

docs/standard/commandline/define-commands.md
```diff
@@ -0,0 +1,208 @@
...
+
+```output
+Options:
+  --version               Show version information
+  -?, -h, --help          Show help and usage information
+```
+
+## Multiple arguments
+
+By default, you can repeat the option name to specify multiple arguments for an option that has maximum [arity](syntax.md#argument-arity) greater than one.
```
KathleenDollard, line 151: Can we start this with something about calling the CLI

```suggestion
By default, when you call a command you can repeat an option name to specify multiple arguments for an option that has maximum [arity](syntax.md#argument-arity) greater than one.
```

docs/standard/commandline/define-commands.md
```diff
@@ -0,0 +1,208 @@
...
+myapp --items one --items two --items three
+```
+
+To allow multiple arguments without repeating the option name, set `Option.AllowMultipleArgumentsPerToken` to `true`. This setting lets you enter the following command line.
+
+```console
+myapp --items one two three
+```
+
+The same setting has a different effect if maximum argument arity is 1. It allows you to repeat an option but takes only the last value on the line. In the following example, the value `three` would be passed to the app.
```
KathleenDollard, line 163: @jonsequitur Is this in the Posix standard, or is otherwise a good idea? (SCL, not docs question)

docs/standard/commandline/define-commands.md
```diff
@@ -0,0 +1,208 @@
...
+Here's an example of command-line input and the resulting output for the preceding example code:
+
+```console
+myapp --language not-a-language
+```
+
+```output
+Argument 'not-a-language' not recognized. Must be one of:
+        'csharp'
+        'fsharp'
```
KathleenDollard, line 184: If you take my advice to add VB, add it here too

docs/standard/commandline/define-commands.md
```diff
@@ -0,0 +1,208 @@
...
+myapp --items one two three
+```
+
+The same setting has a different effect if maximum argument arity is 1. It allows you to repeat an option but takes only the last value on the line. In the following example, the value `three` would be passed to the app.
+
+```console
+myapp --item one --item two --item three
+```
+
+## List valid argument values
```
KathleenDollard, line 169: Shall we add something along the lines of: 

If your list of values is contained in an enum, you can just use the enum as the type of the option to get similar support.

docs/standard/commandline/dependency-injection.md
```diff
@@ -0,0 +1,37 @@
...
+---
+# How to do dependency injection in System.CommandLine
+
+[!INCLUDE [scl-preview](../../../includes/scl-preview.md)]
+
+Use a [custom binder](model-binding.md#model-binding-more-than-16-options-and-arguments) to inject custom types into a command handler.
+
+We recommend handler-specific dependency injection (DI) for the following reasons:
+
+* Command-line apps are often short-lived processes, in which startup cost can have a noticeable impact on performance. Optimizing performance is particularly important when tab completions have to be calculated. The startup cost of configuring a DI container up front is acceptable in Web and GUI apps because they tend to be relatively long-lived processes.
```
KathleenDollard, line 20: Consider: 

```suggestion
* Command-line apps are often short-lived processes, in which startup cost can have a noticeable impact on performance. Optimizing performance is particularly important when tab completions have to be calculated. Unlike Web and GUI apps that tend to be relatively long-lived processes, CLI apps are often short lived where unnecessary startup time is not appropriate.
```

docs/standard/commandline/get-started-tutorial.md
```diff
@@ -0,0 +1,434 @@
...
+> * Assign an option recursively to all subcommands under a command.
+> * Work with multiple levels of nested subcommands.
+> * Create aliases for commands and options.
+> * Work with `string`, `string[]`, `int`, `bool`, `FileInfo` and enum option types.
+> * Bind option values to command handler code.
+> * Use custom code for parsing and validating options.
+
+## Prerequisites
+
+* A code editor, such as [Visual Studio Code](https://code.visualstudio.com/) with the [C# extension](https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.csharp).
```
KathleenDollard, line 34: Should we include VS here? If the user has VS, they do not need VSCode or the extension or the .NET SDK

docs/standard/commandline/get-started-tutorial.md
```diff
@@ -0,0 +1,434 @@
...
+
+## Install the System.CommandLine package
+
+* Run the following command:
+
+  ```dotnetcli
+  dotnet add package System.CommandLine --prerelease
+  ```
+
+  The `--prerelease` option is necessary because the library is still in beta.
```
KathleenDollard, line 59: How do we note to update this when we go to GA

docs/standard/commandline/get-started-tutorial.md
```diff
@@ -0,0 +1,434 @@
...
+
+* Creates an [option](syntax.md#options) named `--file` of type `FileInfo` and assigns it to the [root command](syntax.md#commands):
+
+  :::code language="csharp" source="snippets/get-started-tutorial/csharp/Stage1/Program.cs" id="option" :::
+
+* Specifies that `ReadFile` is the method that will be called when the root command is invoked:
+
+  :::code language="csharp" source="snippets/get-started-tutorial/csharp/Stage1/Program.cs" id="sethandler" :::
+
+* Displays the contents of the specified file when the root command is invoked:
```
KathleenDollard, line 75: Consider:

```suggestion
* The ReadFile method displays the contents of the specified file when the root command is invoked:
```

docs/standard/commandline/get-started-tutorial.md
```diff
@@ -0,0 +1,434 @@
...
+
+   ```xml
+   <ItemGroup>
+     <None Update="sampleQuotes.txt">
+       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+     </None>
+   </ItemGroup>
+   ```
+
+   This markup causes the text file to be copied to the *bin/debug/net6.0* folder when you build the app. So when you run the executable in that folder, you can access the file by name without specifying a folder path.
```
KathleenDollard, line 182: Consider: 

```suggestion
   TAdding this markup causes the text file to be copied to the *bin/debug/net6.0* folder when you build the app. So when you run the executable in that folder, you can access the file by name without specifying a folder path.
```

KathleenDollard (commented):
Preserving this much of my review

docs/standard/commandline/get-started-tutorial.md
```diff
@@ -0,0 +1,434 @@
...
+
+This section creates the final version of the app. When finished, the app will have the following commands and options:
+
+* root command with a global\* option named `--file`
+  * `quotes` command
+    * `read` command with options named `--delay`, `--fgcolor`, and `--light-mode`
+    * `add` command with arguments named `quote` and `byline`
+    * `delete` command with option named `--search-terms`
+
+\* A global option is available to the command it's assigned to and recursively to all its subcommands.
```
KathleenDollard, line 314: Is the slash a typo?

docs/standard/commandline/get-started-tutorial.md
```diff
@@ -0,0 +1,434 @@
...
+
+   The result is the following command hierarchy:
+
+   * Root command
+     * `quotes`
+       * `read`
+       * `add`
+       * `delete`
+
+   The app now implements the recommended pattern where the parent command (`quotes`)specifies an area or group, and its children commands (`read`, `add`, `delete`) are actions.
```
KathleenDollard, line 362:
```suggestion
   The app now implements the recommended pattern where the parent command (`quotes`) specifies an area or group, and its children commands (`read`, `add`, `delete`) are actions.
```

docs/standard/commandline/get-started-tutorial.md
```diff
@@ -0,0 +1,434 @@
...
+```
+
+Run subcommand `delete` with search strings from the beginning of the file, and then look at the beginning of the text file to see where text was removed:
+
+```console
+scl quotes delete --search-terms David "You can do" Antoine "Perfection is achieved"
+```
+
+> [!NOTE]
+> If you're running in the *bin/debug/net6.0* folder, that folder is where you'll find the file with changes from the `add` and `delete` commands. The copy of the file in the project folder remains unchanged.
```
KathleenDollard, line 430: do you want to add that this is because of the setting in the .csproj that says to copy it.

docs/standard/commandline/index.md
```diff
@@ -0,0 +1,53 @@
...
+ms.date: 02/03/2022
+no-loc: [System.CommandLine]
+helpviewer_keywords:
+  - "command line interface"
+  - "command line"
+  - "System.CommandLine"
+ms.topic: overview
+---
+
+# System.CommandLine overview
```
KathleenDollard, line 13: This section is great. Since it is super important I wanted you to know I reviewed it.

docs/standard/commandline/model-binding.md
```diff
@@ -0,0 +1,225 @@
...
+---
+# How to bind arguments to handlers in System.CommandLine
+
+[!INCLUDE [scl-preview](../../../includes/scl-preview.md)]
+
+The process of parsing arguments and providing them to command handler code is called *model binding*. `System.CommandLine` has the ability to bind many argument types built in. For example, integers, enums, and file system objects such as `FileInfo` and `DirectoryInfo` can be bound. `FileInfo` and `DirectoryInfo` are examples of a general convention whereby any type that has a constructor that takes a single `string` parameter can be bound without your having to write any custom code. Several `System.CommandLine` types can also be bound.
+
+## Built-in argument validation
+
+Arguments have expected types and [arity](syntax.md#argument-arity). `System.CommandLine` rejects arguments that don't match these expectations.
```
KathleenDollard, line 20: Do we need to define arity?

docs/standard/commandline/model-binding.md
```diff
@@ -0,0 +1,225 @@
...
+
+:::code language="csharp" source="snippets/model-binding/csharp/ReturnExitCode.cs" id="returnexitcode" :::
+
+However, if the lambda itself needs to be async, you can't return a `Task<int>`. In that case, use `InvocationContext.ExitCode`. You can get the `InvocationContext` instance injected into your lambda just by including it as one of the parameters, as in the following example:
+
+:::code language="csharp" source="snippets/model-binding/csharp/ContextExitCode.cs" id="contextexitcode" :::
+
+If you don't have asynchronous work to do, you can use the `Action` overloads. In that case, set the exit code by using `InvocationContext.ExitCode` the same way you would with an async lambda.
+
+If you don't explicitly set the exit code, and your handler exits normally, the exit code is automatically set to 0. If an exception is thrown, the exit code is set to 1.
```
KathleenDollard, line 101:
```suggestion
If you don't explicitly set the exit code, and your handler exits normally, the exit code is automatically set to 0. If an exception is thrown and caught by System.CommandLine code, the exit code is set to 1.
```

docs/standard/commandline/model-binding.md
```diff
@@ -0,0 +1,225 @@
...
+
+Here's sample command-line input and resulting output from the preceding example:
+
+```console
+myapp --enum red
+myapp --enum RED
+```
+
+```output
+Read
```
KathleenDollard, line 121: Is this a typo

docs/standard/commandline/model-binding.md
```diff
@@ -0,0 +1,225 @@
...
+
+* Parsing of custom types, such as the `Person` class in the following example:
+
+  :::code language="csharp" source="snippets/model-binding/csharp/ParseArgument.cs" id="persontype" :::
+
+  :::code language="csharp" source="snippets/model-binding/csharp/ParseArgument.cs" id="personoption" :::
+
+* Parsing of other kinds of input strings (for example, parse "1,2,3" into `int[]`).
+
+* Dynamic arity. For example, you have two arguments that are defined as string arrays, and you have to handle a sequence of strings in the command line input. The `OnlyTake` method enables you to dynamically divide up the input strings between the arguments.
```
KathleenDollard, line 221: Consider: 

```suggestion
* Dynamic arity. For example, you have two arguments that are defined as string arrays, and you have to handle a sequence of strings in the command line input. A method like `OnlyTake` enables you to dynamically divide up the input strings between the arguments.
```

docs/standard/commandline/snippets/get-started-tutorial/csharp/Stage3/Program.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                if (result.Tokens.Count == 0)
+                {
+                    return new FileInfo("sampleQuotes.txt");
+
+                }
+                var filePath = result.Tokens.Single().Value;
+                if (!File.Exists(filePath))
+                {
+                    result.ErrorMessage = "File does not exist";
```
KathleenDollard, line 25: Consider:

// This allows customization of the error behavior, like a custom message. Without this, missing files will still be reported.

docs/standard/commandline/snippets/handle-termination/csharp/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+            CancellationToken token) =>
+        {
+            await DoRootCommand(urlOptionValue, token);
+        },
+        urlOption);
+
+        await rootCommand.InvokeAsync(args);
+    }
+
+    public static async Task<int> DoRootCommand(
```
KathleenDollard, line 24: Will Ctl-C cancel this example?

docs/standard/commandline/snippets/model-binding/csharp/Program.cs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    static async Task OnlyTake(string[] args)
+    {
+        // <onlytake>
+        var rootCommand = new RootCommand
+        {
+            new Argument<string[]>(name: "arg1", parse: result =>
+            {
+                result.OnlyTake(2);
```
KathleenDollard, line 144: Not obvious where this method is (OnlyTake(int))

docs/standard/commandline/snippets/tab-completion/csharp/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+    private Option<DateTime> dateOption = 
+        new ("--date", "The day of week to schedule. Should be within two weeks.");
+    
+    public DateCommand() : base("schedule", "Makes an appointment for sometime in the next week.")
+    {
+        this.AddArgument(subjectArgument);
+        this.AddOption(dateOption);
+
+        // <dateoption>
+        dateOption.AddCompletions((ctx) => {
```
KathleenDollard, line 21: @jonsequitur  @baronfel  When would you do this rather than FromAmong?

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+
+## Tokens
+
+`System.CommandLine` parses command-line input into *tokens*, which are strings delimited by spaces. For example, consider the following command line:
+
+```dotnetcli
+dotnet tool install dotnet-suggest --global --verbosity quiet
+```
+
+This input is parsed into tokens `tool`, `install`, `dotnet-suggest`, `--global`, `--verbosity`, and `quiet`.
```
KathleenDollard, line 27: I think this is strictly correct, but a bit confusing as to whether dotnet is a token. I think it might be easier to use the text "parsed tokens" and then not include it in the table below.

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+
+Tokens are interpreted as commands, options, or arguments. The command-line app that is being invoked determines how the tokens after the first one are interpreted. The first token identifies the app. The following table shows how `System.CommandLine` interprets the preceding example:
+
+| Token            | Parsed as                         |
+|------------------|-----------------------------------|
+| `dotnet`         | Root command - runs *dotnet.exe*  |
```
KathleenDollard, line 33: dotnet appearing here seems a bit confusing to someone with a tenuous grasp on what a token is.

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+dotnet tool update dotnet-suggest --global
+                                  ^------^
+
+dotnet tool update dotnet-suggest --global true
+                                  ^-----------^
+```
+
+Some options have required arguments. For example in the .NET CLI, `--output` requires a folder name argument. If the argument is not provided, the command fails.
+
+Arguments can have expected types, and the app errors if the argument can't be parsed into the expected type. For example, the following command errors because "silent" isn't one of the valid values for `--verbosity`:
```
KathleenDollard, line 110: Is "app errors" a phrase we often use in docs?

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+Minimal
+Normal
+Quiet
+```
+
+Arguments also have expectations about how many values can be provided. Examples are provided in the [section on argument arity](#argument-arity).
+
+## Order of options and arguments
+
+You can provide options before arguments or arguments before options on the command line. The following commands are equivalent:
```
KathleenDollard, line 129: Do we ever have problems with ambiguity? if --prerelease had an optional string argument, for example

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+```
+
+Options can be specified in any order. The following commands are equivalent:
+
+```dotnetcli
+dotnet add package System.CommandLine --prerelease --no-restore --source https://api.nuget.org/v3/index.json
+dotnet add package System.CommandLine --source https://api.nuget.org/v3/index.json --no-restore --prerelease 
+```
+
+When there are multiple arguments, the order does matter. The following commands are not equivalent:
```
KathleenDollard, line 143: Well, they might be. Maybe 

These pass a list with the same values to the handler, but they differ in the order of the values.

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+dotnet publish --out ./publish
+dotnet publish --ou ./publish
+dotnet publish --o ./publish
+```
+
+`System.CommandLine` doesn't support automatic aliases.
+
+## Case sensitivity
+
+Command and option names and aliases are case-sensitive by default according to POSIX convention, and `System.CommandLine` follows this convention. If you want your CLI to be case insensitive, define aliases for the various casing alternatives. For example, `--additional-probing-path` could have aliases `--Additional-Probing-Path` and `--ADDITIONAL-PROBING-PATH`.
```
KathleenDollard, line 174: @jonsequitur  I hate this solution and think we should have an option to be case insensitive (at some future point after GA).

docs/standard/commandline/customize-help.md
jonsequitur: Help customization allows people to change this output.

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+In this example, the `--project` option is passed to the `dotnet run` command, and the `--message` option with its argument is passed as a command-line option to *myapp* when it runs.
+
+The `--` token is not always required for passing options to an app that you run by using `dotnet run`. Without the double-dash, the `dotnet run` command automatically passes on to the app being run any options that aren't recognized as applying to `dotnet run` itself or to MSBuild. So the following command lines are equivalent because `dotnet run` doesn't recognize the arguments and options:
+
+```dotnetcli
+dotnet run -- quotes read --delay 0 --fg-color red
+dotnet run quotes read --delay 0 --fg-color red
+```
+
+This behavior doesn't follow POSIX standards, and `System.CommandLine` doesn't have built-in support for it.
```
KathleenDollard, line 215: If we are saying that the .NET CLI is weird and does not use standard SCL defaults, I think we should say that more clearly @baronfel

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+```dotnetcli
+dotnet build -v quiet
+dotnet build -v=quiet
+dotnet build -v:quiet
+```
+
+A POSIX convention lets you omit the delimiter when you are specifying a single-character option alias. For example, the following commands are equivalent:
+
+```console
+myapp -vquiet
```
KathleenDollard, line 230: Oh, that is truly awful @jonsequitur But I can live iwth it if it is Posix. Can we have a feature request that authors can turn off this horribleness

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+myapp -v quiet
+```
+
+System.CommandLine supports this syntax by default, but many of the .NET CLI commands and options don't support it.
+
+## Argument arity
```
KathleenDollard, line 236: Can we add that arity is generally inferred based on the type. And the example was for OneOrMany or similar, so I think we will need a bit more here.

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+
+`System.CommandLine` can be configured to accept multiple arguments for one option without repeating the option name.
+
+In the following example, the list passed to the `myapp` command would contain "a", "b", "c", and "d":
+
+```console
+myapp --list a b c --list d
+```
+
+This syntax pattern isn't enabled for .NET CLI commands.
```
KathleenDollard, line 279: Do we need so many comments on the >NET CLI being werd? @baronfel

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+
+## Option bundling
+
+POSIX recommends that you support *bundling* of single-character options, also known as *stacking*. Bundled options are single-character option aliases specified together after a single hyphen prefix. The options before the last one in a bundle can't have required arguments. For example, the following command lines are equivalent:
```
KathleenDollard, line 283: Consider:

```suggestion
POSIX recommends that you support *bundling* of single-character options, also known as *stacking*. Bundled options are single-character option aliases specified together after a single hyphen prefix. Only the last option can specify an argument. For example, the following command lines are equivalent:
```

docs/standard/commandline/define-commands.md
jonsequitur: Yes. For maximum arity 1, it's a POSIX convention to allow a repeated option to override the value. 

`myapp --items one two three` for applying `one`, `two`, and `three` to a single option would be surprising for many people used to POSIX conventions, which is why it is no longer the default.

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+```
+
+```output
+False
+True
+False
+True
+```
+
+In the .NET CLI, this pattern doesn't always hold true. Some Boolean options result in the same behavior when you pass `false` as when you pass `true`. This behavior results when .NET CLI code that implements the option only checks for the presence or absence of the option, ignoring the value. An example is `--no-restore` for the `dotnet build` command. Pass `no-restore false` and the restore operation will be skipped the same as when you specify `no-restore true` or `no-restore`.
```
KathleenDollard, line 317: My opinion is solidfiying. Somewhere in our world we should have "How the .NET CLI is weird". It isn't that important when people are learning SCL, and it might be to someone that isn't ever going to use SCL

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+
+```dotnetcli
+dotnet --help
+dotnet -h
+dotnet /h
+dotnet -?
+dotnet /?
+```
+
+Specific commands, options, or arguments may be *hidden*, which means they don't show up in help output but they can be specified on the command line.
```
KathleenDollard, line 355: Why is this under the help option?

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+  -?, -h, --help  Show command line help.
+
+Commands:
+  package    List all package references of the project or solution.
+  reference  List all project-to-project references of the project.
+```
+
+App users might be accustomed to different ways to request help on different platforms, so apps built on `System.CommandLine` respond to many ways of requesting help. The following commands are all equivalent:
+
+```dotnetcli
```
KathleenDollard, line 347: Do we have an easy way for nothing following the command to give help. Personally I think it is a bad idea, but it is not uncommon.

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+```
+
+## Response files
+
+A *response file* is a file that contains a set of [tokens](syntax.md#tokens) for a command-line app. Response files are a feature of `System.CommandLine` that is useful in two scenarios:
+
+* To invoke a command-line app by specifying input that is longer than the character limit of the terminal.
+* To invoke the same command repeatedly without retyping the whole line.
+
+To use a response file, enter the file name prefixed by an `@` sign wherever in the line you want to insert commands, options, and arguments. The following lines are equivalent:
```
KathleenDollard, line 376: Is this a Posix standard?

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+### The `[parse]` directive
+
+Both users and developers may find it useful to see how an app will interpret a given input. One of the default features of a `System.CommandLine` app is the `[parse]` directive, which lets you preview the result of parsing command input. For example:
+
+```console
+myapp [parse] --delay not-an-int --interactive --file file.txt extra
+```
+
+```output
+![ myapp [ --delay !<not-an-int> ] [ --interactive <True> ] [ --file <filename.txt> ] *[ --fgcolor <White> ] ]   ???--> extra
```
KathleenDollard, line 440: Apparent typo:

```suggestion
![ myapp [ --delay !<not-an-int> ] [ --interactive <True> ] [ --file <file.txt> ] *[ --fgcolor <White> ] ]   ???--> extra
```

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+```
+
+In the preceding example:
+
+* The command (`myapp`), its child options, and the arguments to those options are grouped using square brackets.
+* For the option result `![ --delay <not-an-int> ]`, the `!` indicates a parsing error. The value `not-an-int` for an `int` option can't be parsed to the expected type. The error is also flagged by `!` in front of the command that contains the errored option: `![ myapp...`.
+* For the option result `*[ --fgcolor <White> ]`, the option wasn't specified on the command line, so the configured default was used. `White` is the effective value for this option.
```
KathleenDollard, line 447: Consider:

```suggestion
* For the option result `*[ --fgcolor <White> ]`, the option wasn't specified on the command line, so the configured default was used. `White` is the effective value for this option. The asterisk indicates that the value is the default. 
```

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+dotnet [suggest] buil
+```
+
+```output
+build
+build-server
+msbuild
+```
+
+## Design guidance
```
KathleenDollard, line 464: Note to @KathleenDollard Stopped here on review

KathleenDollard (commented):
Partial review

KathleenDollard (commented):
More partial review

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+
+```dotnetcli
+dotnet --version
+```
+
+```output
+6.0.100
+```
+
+## Response files
```
KathleenDollard, line 369: @jonsequitur  Please review this section based on the recent/planned changes.

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+
+### Short-form aliases
+
+In general, we recommend that you minimize the number of short-form option aliases that you define.
+
+In particular, avoid using any of the following aliases differently than their common usage in the .NET CLI and other .NET command-line apps:
+
+* `-i` for `--interactive`.
+
+  This option signals to the user that they may be prompted for inputs to questions that the command needs answered. For example, prompting for a username.
```
KathleenDollard, line 486:
```suggestion
  This option signals to the user that they may be prompted for inputs to questions that the command needs answered. For example, prompting for a username. Your CLI may be used in scripts, so use caution in prompting uses that have not specified this switch. 
```

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+
+* `-o` for `--output`.
+
+  Some commands produce files as the result of their execution. This option should be used to help determine where those files should be located. In cases where a single file is created, this option should be a file path. In cases where many files are created, this option should be a directory path.
```
KathleenDollard, line 490: @jonsequitur If there is a single option, is it reasonable to accept a directory? Do we want to explicitly suggest that the corresponding types be used, maybe via an example

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+
+  This option often refers to a named Build Configuration, like `Debug` or `Release`. You can use any name you want for a configuration, but most tools are expecting one of those. This setting is often used to configure other properties in a way that makes sense for that configuration&mdash;for example, doing less code optimization when building the `Debug` configuration.
+
+* `-f` for `--framework`
+
+  This option is used to select a single [Target Framework Moniker (TFM)](../frameworks.md) to execute for, so if your CLI application has differing behavior based on which TFM is chosen, you should support this flag.
+
+* `-p` for `--property`
+
+  If your application eventually invokes MSBuild, the user is going to need to customize that call in some way. This option allows for MSBuild properties to be provided on the command line and passed on to the underlying MSBuild call. If your app doesn't use MSBuild but needs a set of key-value pairs, consider using this same option name to take advantage of users' expectations.
```
KathleenDollard, line 508: @baronfel Can you check this language. 

```suggestion
  If your application eventually invokes MSBuild, the user will often need to customize that call in some way. This option allows for MSBuild properties to be provided on the command line and passed on to the underlying MSBuild call. If your app doesn't use MSBuild but needs a set of key-value pairs, consider using this same option name to take advantage of users' expectations.
```

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+
+* `-v` for `--verbosity`.
+
+  Commands often provide output to the user on the console; this option is used to specify the amount of output the user requests. For more information, see [The `--verbosity` option](#the---verbosity-option) later in this article.
+
+There are also some aliases with common usage limited to the .NET CLI. You can use these aliases for other options in your apps, but be aware of the possibility of confusion.
+
+* `-c` for `--configuration`
+
+  This option often refers to a named Build Configuration, like `Debug` or `Release`. You can use any name you want for a configuration, but most tools are expecting one of those. This setting is often used to configure other properties in a way that makes sense for that configuration&mdash;for example, doing less code optimization when building the `Debug` configuration.
```
KathleenDollard, line 500: I really like that the other options here explained where to use it. Consider: 

```suggestion
  This option often refers to a named Build Configuration, like `Debug` or `Release`. You can use any name you want for a configuration, but most tools are expecting one of those. This setting is often used to configure other properties in a way that makes sense for that configuration&mdash;for example, doing less code optimization when building the `Debug` configuration. Consider this option if your command has different modes of operation.
```

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+
+Make command names as short and easy to spell as possible. For example, if `class` is clear enough don't make the command `classification`.
+
+### Lowercase names
+
+Define names in lowercase only, except you can make uppercase aliases to make commands or options case insensitive.
+
+### Kebab case names
+
+Use [kebab case](https://en.wikipedia.org/wiki/Letter_case#Kebab_case) to distinguish words. For example, the .NET CLI option that is defined as [`--additionalprobingpath`](../../core/tools/dotnet.md#runtime-options) should have been defined as `--additional-probing-path`.
```
KathleenDollard, line 524: I think this example makes this very clear, and I'm OK with this pointing out a shortcoming with the .NET CLI

@baronfel

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+
+Case sensitivity does not apply to options that are based on enums. Enum names are matched regardless of casing.
```
KathleenDollard, line 526: I think this section is about command and option names, and this seems to be on how enum values are treated. If so. it doesn't quite seem to fit.

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+
+  If your application eventually invokes MSBuild, the user is going to need to customize that call in some way. This option allows for MSBuild properties to be provided on the command line and passed on to the underlying MSBuild call. If your app doesn't use MSBuild but needs a set of key-value pairs, consider using this same option name to take advantage of users' expectations.
+
+* `-r` for `--runtime`
+
+  If your application can run on different runtimes, or has runtime-specific logic, consider supporting this option as a way of specifying a [Runtime Identifier](../../core/rid-catalog.md). If your app supports --runtime, consider supporting `--os` and `--arch` also. These options let you specify just the OS or the architecture parts of the RID, leaving the part not specified to be determined from the current platform. For more information, see d[dotnet publish](../../core/tools/dotnet-publish.md).
+
+### Short names
+
+Make command names as short and easy to spell as possible. For example, if `class` is clear enough don't make the command `classification`.
```
KathleenDollard, line 516: @jonsequitur For this section, if a name is in Pascal instead of kebab, will we change it?

```suggestion
Make names for commands, options, and arguments as short and easy to spell as possible. For example, if `class` is clear enough don't make the command `classification`.
```

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+
+These are the standard names, but existing apps sometimes use `Silent` in place of `Quiet`, and `Trace`, `Debug`, or `Verbose` in place of `Diagnostic`.
+
+Each app defines its own criteria that determine what gets displayed at each level. Typically an app only needs three levels:
+
+* Quiet
+* Normal
+* Diagnostic
+
+If an app doesn't need five different levels, the option should still define the same five settings. In that case, `Minimal` and `Normal` will produce the same output, and `Detailed` and `Diagnostic` will likewise be the same.
```
KathleenDollard, line 561:
```suggestion
If an app doesn't need five different levels, the option should still define the same five settings. In that case, `Minimal` and `Normal` will produce the same output, and `Detailed` and `Diagnostic` will likewise be the same. This allows your users to just type what they are familiar with, and the best fit will be used.

File: docs/standard/commandline/syntax.md [hunk 68]
```diff
@@ -0,0 +1,577 @@
...
+
+The expectation for `Quiet` is that no output is displayed on the console. However, if an app offers an interactive mode, the app should do one of the following alternatives:
+
+* Display prompts for input when `--interactive` is specified, even if `--verbosity` is `Quiet`.
+* Disallow the use of `--verbosity Quiet` and `--interactive` together.
+
+Otherwise the app will wait for input without telling the user what it's waiting for.
```
KathleenDollard, line 568: Maybe add something like: 

```suggestion
Otherwise the app will wait for input without telling the user what it's waiting for. It will appear that your application froze and the user will have no idea the application is waiting for input. 
```

docs/standard/commandline/tab-completion.md
```diff
@@ -0,0 +1,86 @@
...
+  - "command line"
+  - "System.CommandLine"
+ms.topic: how-to
+---
+
+# Tab completion for System.CommandLine
+
+[!INCLUDE [scl-preview](../../../includes/scl-preview.md)]
+
+Apps that use `System.CommandLine` have built-in support for tab completion in certain shells. To enable it, the end user has to take a few steps once per shell. Tab completion is automatic for static values in your app, such as enum values or values you define by calling [FromAmong](define-commands.md#list-valid-argument-values). You can also customize the tab completion by getting values dynamically at runtime.
```
KathleenDollard, line 17: @jonsequitur does tab-completion work dynamically for files? Should we include that here?

```suggestion
Apps that use `System.CommandLine` have built-in support for tab completion in certain shells. To enable it, the end user has to take a few steps once per shell. Once the user does this, tab completion is automatic for static values in your app, such as enum values or values you define by calling [FromAmong](define-commands.md#list-valid-argument-values). You can also customize the tab completion by getting values dynamically at runtime.
```

docs/standard/commandline/tab-completion.md
```diff
@@ -0,0 +1,86 @@
...
+
+  * For PowerShell, add the contents of [*dotnet-suggest-shim.ps1*](https://github.com/dotnet/command-line-api/blob/master/src/System.CommandLine.Suggest/dotnet-suggest-shim.ps1) to your PowerShell profile. You can find the expected path to your PowerShell profile by running the following command in your console:
+
+    ```console
+    echo $profile
+    ```
+
+Once the user's shell is set up, completions will work for all apps that are built by using `System.CommandLine`.
+
+For *cmd.exe* on Windows there is no pluggable tab completion mechanism, so no shim script is available. For other shells, [look for a GitHub issue that is labeled `Area-Completions`](https://github.com/dotnet/command-line-api/issues?q=is%3Aissue+is%3Aopen+label%3A%22Area-Completions%22). If you don't find an issue, you can [open a new one](https://github.com/dotnet/command-line-api/issues).
```
KathleenDollard, line 45:
```suggestion
For *cmd.exe* on Windows (the Windows Command Prompt) there is no pluggable tab completion mechanism, so no shim script is available. For other shells, [look for a GitHub issue that is labeled `Area-Completions`](https://github.com/dotnet/command-line-api/issues?q=is%3Aissue+is%3Aopen+label%3A%22Area-Completions%22). If you don't find an issue, you can [open a new one](https://github.com/dotnet/command-line-api/issues).
```

docs/standard/commandline/tab-completion.md
```diff
@@ -0,0 +1,86 @@
...
+* `Label` is the completion value to be shown.
+* `SortText` ensures that the values in the list are presented in the right order. It's set by converting `i` to a two-digit string, so that sorting is based on 01, 02, 03, and so on, through 14. If you don't set this parameter, sorting is based on the `Label`, which in this example is in short date format and won't sort correctly.
+
+There are other `CompletionItem` properties, such as `Documentation` and `Detail`, but they aren't used yet in `System.CommandLine`.
+
+The dynamic tab completion list created by this code also appears in help output:
+
+```output
+Description:
+  Makes an appointment for sometime in the next week.
```
KathleenDollard, line 68: This says week, instead of two weeks. I think week might be better as the help would be prettier.

docs/standard/commandline/use-middleware.md
```diff
@@ -0,0 +1,53 @@
...
+  - "command line interface"
+  - "command line"
+  - "System.CommandLine"
+ms.topic: how-to
+---
+# How to use middleware in System.CommandLine
+
+[!INCLUDE [scl-preview](../../../includes/scl-preview.md)]
+
+This article explains how to work with middleware in command-line apps that are built with the `System.CommandLine` library.
```
KathleenDollard, line 16: Consider: 

```suggestion
This article explains how to work with middleware in command-line apps that are built with the `System.CommandLine` library. This is an advanced topic that most System.CommandLine users will not need to consider.
```

docs/standard/commandline/use-middleware.md
```diff
@@ -0,0 +1,53 @@
...
+
+Each call in the pipeline can take action based on the `ParseResult` and return early, or choose to call the next item in the pipeline. The `ParseResult` can even be replaced during this phase. The last call in the chain is the handler for the specified command.
+
+## Add to the middleware pipeline
+
+You can add a call to this pipeline by calling `CommandLineBuilder.AddMiddleware.` Here's an example of code that enables a custom [directive](syntax.md#directives). After creating a root command named `rootCommand`, the code as usual adds options, arguments, and handlers. Then the middleware is added:
+
+:::code language="csharp" source="snippets/use-middleware/csharp/Program.cs" id="middleware" :::
+
+In the preceding code, the middleware writes out "Hi!" if the directive `[just-say-hi]` is found in the parse result. When this happens, because the provided `next` delegate isn't called, the command's normal handler isn't invoked.
```
KathleenDollard, line 30: Found this a bit hard to read, consider:

```suggestion
In the preceding code, the middleware writes out "Hi!" if the directive `[just-say-hi]` is found in the parse result. When this happens the command's normal handler isn't invoked, because the middleware does not call the `next` delegate. 
```

includes/cli-aliases.md
```diff
@@ -0,0 +1,12 @@
+---
```
KathleenDollard, line 1: This is in two places. Should we consolidate a bit? Also, the discussion in the CLI design guidelines has more information.

KathleenDollard (commented):
Wow, this a great, and large, piece of work! Awesome. 

I had a number of comments, but they are all small. Let's use tomorrow's meeting to resolve any of these that are not easy to resolve, and I will approve once we've settled this.

gewarren:
Should we replace occurrences of "How to do dependency injection" with "How to configure dependency injection"?

tdykstra:
@gewarren Yes that sounds better. 😄

baronfel:
Top-level question - I noticed that a getting started sample's binary output had been committed as part of this changeset - was that intended? I'm not used to seeing raw compiled outputs committed to source control.

docs/standard/commandline/snippets/get-started-tutorial/csharp/Stage1/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "scl": {
```
baronfel, line 3: it's worth noting that as of 7.0.100, the SDK will be able to use these `commandLineArgs` as well with `dotnet run` if you specify the `--launch-profile` argument. So `dotnet run --launch-profile scl` would allow for a consistent experience in VS/VSMac/CLI

docs/standard/commandline/snippets/get-started-tutorial/csharp/Stage2/scl.csproj
```diff
@@ -0,0 +1,20 @@
...
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.CommandLine" Version="2.0.0-beta2.21617.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Update="sampleQuotes.txt">
```
baronfel, line 15: It might be simpler to represent this as a `<Content Include="sampleQuotes.txt">`? Content items are included in the output directory by default.

docs/standard/commandline/syntax.md
```diff
@@ -0,0 +1,577 @@
...
+```
+
+## Design guidance
+
+The following sections present guidance that we recommend you follow when designing a CLI. Think of what your app expects on the command line as similar to what a REST API server expects in the URL. Consistent rules for REST APIs are what make them readily usable to client app developers. In the same way, users of your command-line apps will have a better experience if the CLI design follows common patterns.
+
+### Commands and subcommands
+
+If a command has subcommands, the command should function as an area, or a grouping identifier for the subcommands, rather than specify an action. When you invoke the app, you specify the grouping command and one of its subcommands. For example, try to run `dotnet tool`, and you get an error message because the `tool` command only identifies a group of tool-related subcommands, such as `install` and `list`. You can run `dotnet tool install`, but `dotnet tool` by itself would be incomplete.
+
```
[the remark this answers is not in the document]
baronfel: docker is a good example of this: they have both `docker build` (the legacy version from the 1.0 version of the CLI) as well as `docker image build` (under the `image` area that they've been evolving into.

docs/standard/commandline/syntax.md
tdykstra: >Do we have an easy way for nothing following the command to give help. Personally I think it is a bad idea, but it is not uncommon.

@jonsequitur ?

docs/standard/commandline/syntax.md
tdykstra: Because it explains why you don't always see all commands or options in help output.

docs/standard/commandline/model-binding.md
tdykstra: I think the following example explains it but I can add more text if you think it needs it.

docs/standard/commandline/snippets/get-started-tutorial/csharp/Stage1/Properties/launchSettings.json
tdykstra: Added to the how-to-test instructions in the tutorial.

docs/standard/commandline/snippets/get-started-tutorial/csharp/Stage2/scl.csproj
tdykstra: When I change to content include and build the app, sampleQuotes.txt doesn't show up in bin/debug/net6.0.

tdykstra:
>Top-level question - I noticed that a getting started sample's binary output had been committed as part of this changeset - was that intended? I'm not used to seeing raw compiled outputs committed to source control.

Good question, I would have expected the gitignore to keep those out of the PR. I don't know why they're there.

docs/standard/commandline/model-binding.md
KathleenDollard: It seems like this should be Red, not Read, or am I missing something

KathleenDollard (approved):
A couple of dangling things. 

* I had some possible alternate text, all of which you may accept or reject based on how you think it should be. I did not resolve these. 
* _There is an apparent typo of Read vs Red - marked in an unresolved comment._
* _I believe there is typo of a slash, marked with an unresolved comment. If that is not a typo, just resolve the comment._
* Chet had a suggestion for an alternate way to copy to the build output directory. I think we can use your approach, and follow up later as to why the alternative did not work.

Since it is just down to these small items. I am going to approve, but please check on those two potential typos.

docs/standard/commandline/get-started-tutorial.md
tdykstra: The bullet points are continuations of the intro text: 

The preceding code... 

* Creates an option ...
* Specifies that ReadFile is the handler ... 
* Displays the file...

docs/standard/commandline/snippets/get-started-tutorial/csharp/Stage2/scl.csproj
tdykstra: It works with content include if I keep the CopyAlways line.

PR #28073, opened 2022-02-04
Add Win10 21H2 in installation requirements

ychikazawa:
## Summary

.NET Framework can be used in Windows 10 21H2, but this document doesn't discribe that.
So, I added new row in Installation requirements.


adegeo (approved):
Thank you for the contribution!

PR #28125, opened 2022-02-08
Clarify SDK/MSBuild roles in CA1416

rainersigwald:
This is a minor nit, but MSBuild isn't directly creating these attributes, the .NET SDK targets are.


gewarren (approved):
Thanks @rainersigwald

PR #28138, opened 2022-02-09
Breaking change for FullPath

### LINKED ISSUE #27916: [Breaking change]: FileSystemEventArgs.FullPath returns a truly fully qualified path
### Description

Initially the `FullPath` property of `FileSystemEventArgs` and `RenamedEventArgs` returned whatever directory path it was passed in from `FileSystemWatcher`. That could mean even a relative path, that does not match with the true intent and name of the **FullPath** and **OldFullPath** properties on the event args classes.

### Version

.NET 7 Preview 1
Reverted in .NET 7 Preview 5

### Previous behavior

`FullPath` and `OldFullPath` simply mirror what is was passed into FileSystemWatcher initially, which can include a relative path.

### New behavior

The `FullPath` and `OldFullPath` will now return a fully qualified path even when FileSystemWatcher is initialized with a relative path.

### Type of breaking change

### Reason for change

To align the actual result with the intent and name of the `FullPath`/`OldFullPath` properties from the event args.

### Recommended action

The users will have to check their usage of FullPath and OldFullPath when dealing with FileSystemEventArgs and RenamedEventArgs objects and make sure they account for the path change.

### Feature area

Core .NET libraries

### Affected APIs

Only the FullPath and OldFullPath properties are affected.

gewarren:
Fixes #27916.

[Preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/7.0/filesystemeventargs-fullpath?branch=pr-en-us-28138).

cc @slask

### DIFF
--- a/docs/core/compatibility/7.0.md
+++ b/docs/core/compatibility/7.0.md
@@ -23,6 +23,7 @@ If you're migrating an app to .NET 7, the breaking changes listed here might aff
 | Title | Binary compatible | Source compatible | Introduced |
 | - | :-: | :-: | - |
 | [API obsoletions with non-default diagnostic IDs](core-libraries/7.0/obsolete-apis-with-custom-diagnostics.md) | ✔️ | ❌ | Preview 1 |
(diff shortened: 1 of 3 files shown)


BillWagner (approved):
Thanks @gewarren 

This LGTM. You can :shipit: when ready

PR #28145, opened 2022-02-09
Unity 2021.2 supports .NET Standard 2.1

teo-tsirpanis:
See also #25496.


BillWagner (approved):
Thanks for making this clarification @teo-tsirpanis 

I've reviewed the change, and I'll :shipit: now

PR #28224, opened 2022-02-12
Add clarification regarding default values

### LINKED ISSUE #28200: Are there cases where removing a parameter default value would be acceptable?
RE:
> ❌ DISALLOWED: Changing the default value of a property, field, or parameter

Consider the following scenario:
```
public void Foo(int a = 1, int b = 2) { ... }
```

Let's say we want to add a new overload of `Foo` that also accepts `int c`. The following change should (in my understanding) be both binary-compatible and source-compatible:

```
public void Foo(int a, int b) => Foo(a, b, 3);
public void Foo(int a = 1, int b = 2, int c = 3) { ... }
```

(1) Would this be allowed in Microsoft's .NET libraries/runtime?
(2) If it is true that this change does not break either binary or source compatibility, could we add a note about that (I'm happy to file a PR) to this rule for the benefit of those who reference this document to understand breaking changes for .NET generally?

---
#### Document Details

* GitHub Login: @gewarren

madelson:
Clarify the compatibility implications of changing or removing parameter default values. 

Fix #28200

## Summary

Describe your changes here.

### DIFF
--- a/docs/core/compatibility/index.md
+++ b/docs/core/compatibility/index.md
@@ -220,6 +220,11 @@ Changes in this category modify the public surface area of a type. Most of the c
 
 - ❌ **DISALLOWED: Changing the default value of a property, field, or parameter**
 
+  Note that changing or removing a *parameter* default value is not a binary break. Removing a parameter default value is a source break and changing a parameter default value could result in a behavioral break after recompilation.
+  
+  For this reason, removing parameter default values is considered acceptible from a compatibility perspective in the specific case of "moving" those default values to a new method overload in order to eliminate ambiguity. For example, if we have a method `Foo(int a = 1)` and we introduce an overload of `Foo` with 2 optional parameters `a` and `b`, we can preserve compatibility my moving `a`'s default value to the new overload leaving us with `Foo(int a)` and `Foo(int a = 1, int b = 2)`. This will allow `Foo()` to compile.
+ 
+
 - ❌ **DISALLOWED: Changing the precision of a numeric return value**
 
 - ❓ **REQUIRES JUDGMENT: A change in the parsing of input and throwing new exceptions (even if parsing behavior is not specified in the documentation**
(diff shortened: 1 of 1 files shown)


docs/core/compatibility/index.md
```diff
@@ -220,6 +220,10 @@ Changes in this category modify the public surface area of a type. Most of the c
 
 - ❌ **DISALLOWED: Changing the default value of a property, field, or parameter**
 
+  Note that changing or removing a *parameter* default value is not a binary break. Removing a parameter default value is a source break and changing a parameter default value could result in a behavioral break after recompilation.
```
gewarren, line 223:
```suggestion
  Changing or removing a *parameter* default value is not a binary break. Removing a parameter default value is a source break, and changing a parameter default value could result in a behavioral break after recompilation.
```

docs/core/compatibility/index.md
```diff
@@ -220,6 +220,10 @@ Changes in this category modify the public surface area of a type. Most of the c
...
+
+  For this reason, removing parameter default values is considered acceptible from a compatibility perspective in the specific case of "moving" those default values to a new method overload in order to eliminate ambiguity. For example, if we have a method `Foo(int a = 1)` and we introduce an overload of `Foo` with 2 optional parameters `a` and `b`, we can preserve compatibility my moving `a`'s default value to the new overload leaving us with `Foo(int a)` and `Foo(int a = 1, int b = 2)`. This will allow `Foo()` to compile.
```
gewarren, line 225:
```suggestion
  For this reason, removing parameter default values is acceptable in the specific case of "moving" those default values to a new method overload to eliminate ambiguity. For example, consider an existing method `MyMethod(int a = 1)`. If you introduce an overload of `MyMethod` with two optional parameters `a` and `b`, you can preserve compatibility by moving the default value of `a` to the new overload. Now the two overloads are `MyMethod(int a)` and `MyMethod(int a = 1, int b = 2)`. This pattern allows `MyMethod()` to compile.
```

gewarren: We're discouraged from using "Foo" as a name.

bartonjs (approved):
With @gewarren 's suggestions, LGTM

gewarren (approved):

PR #28245, opened 2022-02-15
Fix exception to TreatWarningsAsErrors

eatdrinksleepcode:
**TreatWarningsAsErrors** was described as an exception to itself, which is clearly not correct.


BillWagner (approved):
Thanks for spotting this and fixing it @eatdrinksleepcode 

PR #28246, opened 2022-02-15
Remove ref to JSON DOM as read-only

tdykstra:
In .NET 6 a mutable DOM is available in System.Text.Json.


gewarren (approved):
Should we note it's available in .NET 6 and later? Or is that in the link?

tdykstra:
That's explained in the linked article.

PR #28266, opened 2022-02-16
Fix GitHub Actions typo: `pull_quest` → `pull_request`

swharden:
This PR replaces `pull_quest` with `pull_request`, affecting 3 pages:

* https://docs.microsoft.com/en-us/dotnet/devops/dotnet-build-github-action
* https://docs.microsoft.com/en-us/dotnet/devops/dotnet-secure-github-action
* https://docs.microsoft.com/en-us/dotnet/devops/dotnet-test-github-action

To be sure I double-checked and confirmed there is no such thing as a `pull_quest` 
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows


gewarren (approved):
Thank you @swharden

PR #28295, opened 2022-02-17
Fix api surface and excluded symbols supported rules

Evangelink:
## Summary

Fix top level table for rules supporting `api_surface` and `excluded_symbol_names`.

cc @mavasani


gewarren (approved):
Thank you so much @Evangelink!

Evangelink:
@gewarren I have fixed the conflict.

PR #28322, opened 2022-02-21
Corrected an improper "or" to "nor"

WhitWaldo:
## Summary
Because neither of the states can be, the proper word here is "nor", not "or".


BillWagner (approved):
Good catch @WhitWaldo 

PR #28339, opened 2022-02-22
Update system-text-json-migrate-from-newtonsoft-how-to.md

Rick-Anderson:
Folks reporting having trouble finding Newtonsoft.Json vs System.Text.Json so make this H2 explicit. See https://github.com/dotnet/AspNetCore.Docs/issues/25100

Once this is merged, I can link to it in several places.

### DIFF
--- a/docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
+++ b/docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
@@ -14,7 +14,7 @@ helpviewer_keywords:
 ms.topic: how-to
 ---
 
-# How to migrate from Newtonsoft.Json to System.Text.Json
+# Compare and migrate from Newtonsoft.Json to System.Text.Json
 
 This article shows how to migrate from [Newtonsoft.Json](https://www.newtonsoft.com/json) to <xref:System.Text.Json>.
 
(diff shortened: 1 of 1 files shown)


tdykstra (commented):
Only the first paragraph after the new h2 is related to a comparison, and it doesn't really compare the two.  A better link target for people looking for a comparison is the Table of differences H2. Perhaps we could change that h2 to your proposed title "System.Text.Json and Newtonsoft.Json comparison"

Rick-Anderson:
The title is really deceptive. Maybe we should change the title to "Compare and migrate from NJ to STJ"

This is the critical part I'd like folks to find that want to decide what to use:

System.Text.Json focuses primarily on performance, security, and standards compliance. It has some key differences in default behavior and doesn't aim to have feature parity with Newtonsoft.Json. For some scenarios, System.Text.Json currently has no built-in functionality, but there are recommended workarounds. For other scenarios, workarounds are impractical.

We're investing in adding the features that have most often been requested. 

> doesn't really compare the two.

It makes a major comparison, performance, security, and standards compliance

Sounds like the easiest approach is to change the title.

docs/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to.md
```diff
@@ -14,7 +14,7 @@ helpviewer_keywords:
 ms.topic: how-to
 ---
 
-# How to migrate from Newtonsoft.Json to System.Text.Json
+# Compare and migrate Newtonsoft.Json to System.Text.Json
```
tdykstra, line 17: "migrate" without "from" sounds odd, but "compare ... from" also sounds odd. 

How about 

```suggestion
# Compare Newtonsoft.Json to System.Text.Json, and migrate to System.Text.Json
```

tdykstra (approved):

PR #28345, opened 2022-02-23
Correct sample to avoid overflowing on x64

RussKie:
## Summary

The sample in https://docs.microsoft.com/dotnet/framework/ui-automation/expose-a-server-side-ui-automation-provider is incorrect, see https://source.dot.net/#System.Private.CoreLib/IntPtr.cs,102
image

Noted by the customer in https://github.com/dotnet/winforms/discussions/6736#discussioncomment-2226931.

### DIFF
--- a/samples/snippets/visualbasic/VS_Snippets_Wpf/UIAFragmentProvider_snip/VisualBasic/ListFragment.vb
+++ b/samples/snippets/visualbasic/VS_Snippets_Wpf/UIAFragmentProvider_snip/VisualBasic/ListFragment.vb
@@ -209,7 +209,7 @@ Class ParentList
     Protected Overrides Sub WndProc(ByRef m As Message)
         Const WM_GETOBJECT As Integer = &H3D
 
-        If m.Msg = WM_GETOBJECT AndAlso CType(CType(m.LParam, Long), Int32) = AutomationInteropProvider.RootObjectId Then
+        If m.Msg = WM_GETOBJECT AndAlso CInt(CLng(m.LParam)) = AutomationInteropProvider.RootObjectId Then
             m.Result = AutomationInteropProvider.ReturnRawElementProvider(Me.Handle, m.WParam, m.LParam, DirectCast(Me, IRawElementProviderSimple))
             Return
         End If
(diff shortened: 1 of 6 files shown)


samples/snippets/visualbasic/VS_Snippets_Wpf/UIAFragmentProvider_snip/VisualBasic/ListFragment.vb
```diff
@@ -209,7 +209,7 @@ Class ParentList
     Protected Overrides Sub WndProc(ByRef m As Message)
         Const WM_GETOBJECT As Integer = &H3D
 
-        If m.Msg = WM_GETOBJECT AndAlso m.LParam.ToInt32() = AutomationInteropProvider.RootObjectId Then
+        If m.Msg = WM_GETOBJECT AndAlso CType(CType(m.LParam, Long), Int32) = AutomationInteropProvider.RootObjectId Then
```
RussKie, line 212: @KlausLoeffelmann may I get you to confirm this?

gewarren:
Looks like we'll need to make the corresponding change in the dotnet-api-docs repo too. The C# file is here: https://github.com/dotnet/dotnet-api-docs/blob/6131a03016f1506793c62c1f6dafde8819da4dad/snippets/csharp/System.Windows.Automation/AutomationEventArgs/.ctor/ListFragment.cs#L227

samples/snippets/visualbasic/VS_Snippets_Wpf/UIAFragmentProvider_snip/VisualBasic/ListFragment.vb
gewarren: Ping @KlausLoeffelmann

KlausLoeffelmann: Sorry for the late reply.
Why does `ToInt32` not work here? Wrong conversion of the sign-bit?
Anyway, if there is a need to first convert into `Long` and then `Int`, I'd do it with `CInt` and `CLng`, so, from Memory it would be:

```suggestion
        If m.Msg = WM_GETOBJECT AndAlso CInt(CLng(m.LParam)) = AutomationInteropProvider.RootObjectId Then
```

Would that work?

RussKie: > Why does `ToInt32` not work here? Wrong conversion of the sign-bit?

Please refer to https://source.dot.net/#System.Private.CoreLib/IntPtr.cs,102.

RussKie:
@gewarren may I get a hand with the build failures?

gewarren (approved):

RussKie:
Thank you @gewarren

PR #28368, opened 2022-02-23
Update numbers-in-csharp-local.md

jaigak:
## Summary
Fixes misleading statement. 0.3 repeating forever is exactly equal to 1/3 however floating-point types don't support representing repeating decimals and rounds to a finite number of digits instead.


BillWagner (approved):
Thanks for making this fix @Jaiganeshkumaran It's a good clarification.  I've reviewed the PR, and I'll :shipit: now.

PR #28379, opened 2022-02-23
Update snippet projects

tdykstra:
Contributes to #27310 

Some of these were updated to target .NET 5 because they are used in the .NET 5 version of tutorials.


BillWagner (approved):
This is a great set of changes @tdykstra 

You can :shipit: when ready.

Issue #28383, opened 2022-02-24
OKR: Discovery 


mcleblanc:
GOAL:
Drive discovery and awareness through SEO improvements. 

OBJECTIVES:
Perform selective SEO audits on the following candidate topics:
|Article|PR|
|---|---|
|[.NET introduction and overview](https://docs.microsoft.com/en-us/dotnet/core/introduction)| #28395|
|[Architectural principles](https://docs.microsoft.com/en-us/dotnet/architecture/modern-web-apps-azure/architectural-principles)|#28638|
|[.NET SDK telemetry - .NET CLI](https://docs.microsoft.com/en-us/dotnet/core/tools/telemetry)| #28395|
|[.NET SDK overview](https://docs.microsoft.com/en-us/dotnet/core/sdk)| #28395|
|[Formatting types in .NET](https://docs.microsoft.com/en-us/dotnet/standard/base-types/formatting-types)|#28395|
|[Async in depth](https://docs.microsoft.com/en-us/dotnet/standard/async-in-depth)|#28439|
|[Task-based Asynchronous Pattern (TAP)](https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/task-based-asynchronous-pattern-tap)|#28439|
|[Parallel Programming in .NET](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/)|#28439|
|[BinaryFormatter security guide](https://docs.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide)|#28439|
|[Cross Platform P/Invoke](https://docs.microsoft.com/en-us/dotnet/standard/native-interop/cross-platform)|#28638|
|[Install .NET on Linux Distributions](https://docs.microsoft.com/en-us/dotnet/core/install/linux)||
|[Manually install .NET on Linux - .NET](https://docs.microsoft.com/en-us/dotnet/core/install/linux-scripted-manual)||
|[Install .NET on Linux with Snap - .NET](https://docs.microsoft.com/en-us/dotnet/core/install/linux-snap)||
|[Install .NET on Debian - .NET](https://docs.microsoft.com/en-us/dotnet/core/install/linux-debian)||
|[Dapr for .NET Developers](https://docs.microsoft.com/en-us/dotnet/architecture/dapr-for-net-developers/)|#28638|
[...]

PR #28395, opened 2022-02-24
SEO OKR

tdykstra:
Contributes to #28383 

* Intro to .NET

  Organic search 21.61%. It turns out that many people still include ".net core" in their search strings, so I added that to the title and description. Also changed h1 to "What is ...".

* .NET telemetry

  Organic search 5.10%. The reason for low organic search volume is that almost all traffic comes from an aka.ms link that is displayed by the CLI the first time a CLI command is used after installation (aka.ms/dotnet-cli-telemetry). It probably won't be possible to get the organic search rate up much higher for this article, as most traffic will always be from that aka.ms link. But since the telemetry is more closely related to the CLI in particular than to the SDK in general, I added "CLI" in a couple of places.

* SDK overview

  Organic search 16.19%. Since people are still searching for .NET Core SDK I added "formerly known as the .NET Core SDK" to the meta description. Also changed title to "What is ...".

* Formatting types

  Organic search 17.95%. Some of this article is high-level intro to topics covered in detail in separate articles (standard and custom format strings, composite formatting), other topics are covered in adequate detail in this article. I changed metadata and h1 to emphasize the overview/intro character of the article, to explain better what is meant by "formatting types", and to call out in the description the topics covered in detail in this article. Also removed some superfluous material that interrupted the article introduction.

### DIFF
--- a/docs/core/introduction.md
+++ b/docs/core/introduction.md
@@ -1,6 +1,6 @@
 ---
-title: .NET and .NET Core - introduction and overview
-description: Learn about .NET and .NET Core. .NET is a free, open-source development platform for building many kinds of apps.
+title: .NET (and .NET Core) - introduction and overview
+description: Learn about .NET (and .NET Core). .NET is a free, open-source development platform for building many kinds of apps.
 author: tdykstra
 ms.date: 02/24/2022
 ms.custom: "updateeachrelease"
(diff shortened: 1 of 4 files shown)


BillWagner (approved):
This really looks great @tdykstra 

You can :shipit: when ready

PR #28405, opened 2022-02-25
Add clarifications on records

### LINKED ISSUE #27664: Clarify C# 9 / C# 10 differences
The article starts out with a call to C# 9, but I found the switch to C# 10 features to be very easy to miss.  I made an original assumption that in C# 9, record / record class / record struct, were all introduced at the same time.  I also don't see clarification that a `record struct` is a "value type".  Since `record` is called out in the beginning as a "reference type", I began to assume that a `record struct` was as well.  In between the C# 9 and C# 10 section (mostly at the top, by the way), it's called out that a `record` uses "value equality" (while also mentioning "reference type" again), but it's never mentioned that a `record struct` ALSO uses value equality.  Maybe some of that assumes underlying knowledge of a `struct`, since its the same, but I think it would be good to call it out explicitly since `record class` called it out.

---
#### Document Details

* Content: [Records - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixes #27664

Rework the introduction to clearly state records are a C# 9 feature, and record structs are C# 10.

The later sections in the article combine the discussion to emphasize the similarities between records and record structs.

### DIFF
--- a/docs/csharp/language-reference/builtin-types/record.md
+++ b/docs/csharp/language-reference/builtin-types/record.md
@@ -10,7 +10,7 @@ helpviewer_keywords:
 ---
 # Records (C# reference)
 
-Beginning with C# 9, you use the `record` keyword to define a [reference type](reference-types.md) that provides built-in functionality for encapsulating data. C# 10 allow `record class` as a synonym to clarify a reference type, and `record struct` to define a [value type](value-types.md) with similar functionality. You can create record types with immutable properties by using positional parameters or standard property syntax.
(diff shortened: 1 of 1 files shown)


IEvangelist (approved):
Looks good to me, just a few nits. Take 'em or leave 'em.

PR #28408, opened 2022-02-25
Remove confusing text in code sample

Treit:
"Debug mode" doesn't make any sense and adds unnecessary confusion in this code sample.


BillWagner (approved):
Thanks for making this change @Treit We appreciate it!

PR #28410, opened 2022-02-26
Update local-functions.md

chrisxfire:
Fixes a possible error.


BillWagner:
Hi :wave: @chrisxfire 

I've reviewed this change, and the original text is correct. A `Task` transitions to the `Faulted` state when an uncaught exception is thrown while it is running. The resulting exception is thrown when that faulted `Task` is awaited.

I'll close this, but I can see where the original text may lead to confusion. I'd look at a PR that used language similar to what I wrote above.

chrisxfire:
Ah, my mistake, Bill.  Thanks for reviewing!

Chris

________________________________

PR #28458, opened 2022-03-02
Bulk update incorrect or inconsistent slugs.

IEvangelist:
Bulk update to fix incorrect or inconsistent language identifier (slugs) for triple backtick code blocks.

- `C#` or `c#` should be `csharp`.
- `XML` should be `xml`.
- `dotnet` should be `dotnetcli`.


tdykstra (approved):
LGTM

PR #28467, opened 2022-03-03
SEO OKR

tdykstra:
Contributes to #28383 

* Managed Threading

  Organic search 3%. This is a relatively small link farm of very low value, such that a very low organic search rate is understandable. I'm retiring the article and redirecting its URL to [Managed threading basics](https://docs.microsoft.com/en-us/dotnet/standard/threading/managed-threading-basics).

* Cryptographic Services

  Organic search 15%. "Cryptographic Services" is not very specific about the contents of the article. I changed the title, description, and H1 to specify encryption, digital signatures, hashes, and CNG classes. Rearranged the first paragraph and added an H2 with a reference to "cryptography".

* Properties

  Organic search 21%. There is another article on C# Properties, in the Programming guide. The Programming guide article gets 79% of traffic from organic search and is probably the appropriate page to land on for someone searching for C# properties. However, the title and H1 of this C# concepts section article don't specify C# so I'm adding "in C#" to both.

* Records

  Organic search 14%.  There are several articles about C# records.  The one with highest organic search rate at 88% is the [tutorial](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/records). Next is the [C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record) at 48%.  This introduction to records article may be losing out in search results because it doesn't have C# anywhere in title, H1, or first paragraph text, so I'm adding C# in those places.

If adding "C#" in title and H1 makes a difference in organic search traffic for the Properties and Records articles, there are quite a few other articles that don't now mention C# in those places and we should consider updating those as well.


gewarren (approved):
LGTM

Issue #28372, opened 2022-02-23
[Breaking change]: Change default of SslServerAuthenticationOptions.AllowRenegotiation to false


rzikm:
### Description

PR https://github.com/dotnet/runtime/pull/65777 changed the default value of the `SslServerAuthenticationOptions.AllowRenegotiation` property to `false`.

Original issue: https://github.com/dotnet/runtime/issues/65547

### Version

.NET 7 Preview 3

### Previous behavior

Client-side renegotiation was allowed by the server by default.

### New behavior

Client-side renegotiation must be explicitly enabled on the server side.

### Type of breaking change

- [X] **Binary incompatible**: Existing binaries may encounter a breaking change in behavior, such as failure to load/execute or different run-time behavior.
- [X] **Source incompatible**: Source code may encounter a breaking change in behavior when targeting the new runtime/component/SDK, such as compile errors or different run-time behavior.

### Reason for change

Client-side Renegotiation is viewed as insecure by the industry (e.g. it has been removed from TLS 1.3 entirely). We should disable it by default.

### Recommended action

If client-side renegotiation is required, set the `SslServerAuthenticationOptions.AllowRenegotiation` to `true` when initializing server-side of the `SslStream`.

### Feature area

Networking

### Affected APIs

Besides `SslServerAuthenticationOptions` class itself, following overloads of `SslStream.AuthenticateAsServer` are affected, as they construct `SslServerAuthenticationOptions` internally
```cs
public virtual void AuthenticateAsServer(X509Certificate serverCertificate);
public virtual void AuthenticateAsServer(X509Certificate serverCertificate, bool clientCertificateRequired, bool checkCertificateRevocation);
public virtual void AuthenticateAsServer(X509Certificate serverCertificate, bool clientCertificateRequired, SslProtocols enabledSslProtocols, bool checkCertificateRevocation);
```

PR #28473, opened 2022-03-03
ASP.NET Core 7 breaking change for SignalR Hub DI

serpent5:
Documents https://github.com/aspnet/Announcements/issues/479.

[Internal Preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/7.0?branch=pr-en-us-28473).


docs/core/compatibility/aspnet-core/7.0/signalr-hub-method-parameters-di.md
```diff
@@ -0,0 +1,82 @@
...
+}
+```
+
+## Type of breaking change
+
+This change affects [source compatibility](../../categories.md#source-compatibility).
+
+## Reason for change
+
+We believe the likelihood of breaking apps to be very low as it isn't common to have a type in DI and as an argument in your Hub method at the same time.
```
gewarren, line 51: This isn't really a "reason for change"; nevertheless, I left a small suggestion.

```suggestion
The likelihood of breaking apps is low as it isn't common to have a type in DI and as an argument in your Hub method at the same time.
```

gewarren (approved):

docs/core/compatibility/aspnet-core/7.0/signalr-hub-method-parameters-di.md
BrennanConroy: Isn't the "reason for change" in the context of making a breaking change? So this seems perfectly reasonable as the reason we were fine making a breaking change.

gewarren: That's not how I've seen this field used in the .NET breaking changes. I think of it more like "why it improves the experience". For example, [this issue](https://github.com/dotnet/docs/issues/28372).

serpent5: What should I do with this one? A similar announcement for DI in MVC's API controllers has this:

> This same behavior is already implemented in minimal APIs.
>
> The likelihood of breaking apps is low as it isn't common to have a type in DI and as an argument in your API controller action at the same time.

I see that as covering both why the change was made and why the change was acceptable as being breaking. Would you like to add a line similar to the first one in the example I gave above?

PR #28492, opened 2022-03-04
Update docs-verifier-tryfix.yml

Youssef1313:


.github/workflows/docs-verifier-tryfix.yml
```diff
@@ -28,6 +28,6 @@ jobs:
     - name: Push changes
       run: |
           git config --global user.name github-actions
-          git config --global user.email github-actions@github.com
+          git config --global user.email 41898282+github-actions[bot]@users.noreply.github.com
           git commit -am "Automated: Fix links"
-          git push
+          hub push origin
```
Youssef1313, line 35: @IEvangelist `git push` produces:

```
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
```

I'm not sure why, but will see if using `hub` along with `GITHUB_TOKEN` will resolve this. Per https://hub.github.com/hub.1.html#github-oauth-authentication, the environment variable will be read by `hub`.

IEvangelist: I think that you'd also need to set `permissions` too, right? If you're doing a `push`, see https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token. I'm not sure though.

Youssef1313: Thanks @IEvangelist. I added it, can you review?

I hope it works with these changes, otherwise I'm out of ideas :(

IEvangelist (approved):

PR #28514, opened 2022-03-07
If/then/else is not so different

pbachmann:
Maybe it's just me, but when I first started with F#, I was scared to write code that said if FileIsOk then WriteFile because I was intimidated by documentation that made a clear distinction between F#'s cool new if-then-else expression and the discredited if-then-else statements of lesser languages. Scott Wlaschin even went as far as to say use of if-then-else is strongly discouraged.
The changes I made to this page are designed to help programmers coming from imperative languages relax a bit and feel good about themselves again.

## Summary

Describe your changes here.

### DIFF
--- a/docs/fsharp/language-reference/conditional-expressions-if-then-else.md
+++ b/docs/fsharp/language-reference/conditional-expressions-if-then-else.md
@@ -19,7 +19,7 @@ In the previous syntax, *expression1* runs when the Boolean expression evaluates
 
 Like other languages, the `if...then...else` construct can be used to conditionally execute code. In F#, `if...then...else` is an expression and produces a value by the branch that executes. The types of the expressions in each branch must match.
 
-If there is no explicit `else` branch, the overall type is `unit`, and the type of the `then` branch must also be `unit`. 
+If there is no explicit `else` branch, the overall type is `unit`, and the type of the `then` branch must also be `unit`.
 
 When chaining `if...then...else` expressions together, you can use the keyword `elif` instead of `else if`; they are equivalent.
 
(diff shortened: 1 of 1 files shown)


BillWagner:
Thanks @pbachmann 

I've tagged @dsyme for review. From my reading, both versions are accurate. The connotation is different, and I'd like him to review.

KathleenDollard:
I think we can wait for @dsyme, but I like this change!

pbachmann:
@BillWagner  - Yes, there is no question that the original text and no doubt almost all of the F# documentation is accurate. Programmers are famous for being accurate!
My suggestion is that the effect of the words on the target audience might also be an important consideration.

docs/fsharp/language-reference/conditional-expressions-if-then-else.md
```diff
@@ -17,7 +17,7 @@ if boolean-expression then expression1 [ else expression2 ]
...
+Like other languages, the `if...then...else` construct can be used to conditionally execute code. In F#, however, `if...then...else` is an expression and must also produce a value, being the value of the last expression in the branch that executes. The types of the values produced in each branch must match. If there is no explicit `else` branch, its type is `unit`. Therefore, if the type of the `then` branch is any type other than `unit`, there must be an `else` branch with the same return type. When chaining `if...then...else` expressions together, you can use the keyword `elif` instead of `else if`; they are equivalent.
```
dsyme, line 20:
```suggestion
Like other languages, the `if...then...else` construct can be used to conditionally execute code. In F#, `if...then...else` is an expression and produces a value by the branch that executes. The types of the expressions in each branch must match.

If there is no explicit `else` branch, the overall type is `unit`, and the type of the `then` branch must also be `unit`. 

When chaining `if...then...else` expressions together, you can use the keyword `elif` instead of `else if`; they are equivalent.
```

dsyme (approved):

dsyme:
I made some mods

dsyme:
I think the overall text could still be improved futher - and quite a few of the doc pages suffer from this overly technical language.

BillWagner (approved):
Thanks everyone, I'll :shipit: now.

PR #28536, opened 2022-03-08
Fix using declaration terminology

BillWagner:
The new form is referred to as the `using` declaration in the specification. We should use the same in docs.

### DIFF
--- a/docs/csharp/language-reference/keywords/using-statement.md
+++ b/docs/csharp/language-reference/keywords/using-statement.md
@@ -18,7 +18,7 @@ The following example shows how to use the `using` statement.
 
 :::code language="csharp" source="snippets/usings.cs" id="SnippetFirstExample":::
 
-Beginning with C# 8.0, the *using declaration* doesn't require braces:
+The *using declaration*, which was introduced in C# 8.0, doesn't require braces:
 
 :::code language="csharp" source="snippets/usings.cs" id="SnippetModernUsing":::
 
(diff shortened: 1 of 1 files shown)


docs/csharp/language-reference/keywords/using-statement.md
```diff
@@ -18,17 +18,17 @@ The following example shows how to use the `using` statement.
...
+Both `using` forms ensure that <xref:System.IDisposable.Dispose%2A> (or <xref:System.IAsyncDisposable.DisposeAsync%2A>) is called even if an exception occurs within the `using` block. You can achieve the same result by putting the object inside a `try` block and then calling <xref:System.IDisposable.Dispose%2A> (or <xref:System.IAsyncDisposable.DisposeAsync%2A>) in a `finally` block; in fact, this is how the `using` statement and the `using` declaration is translated by the compiler. The code example earlier expands to the following code at compile time (note the extra curly braces to create the limited scope for the object):
```
gewarren, line 31:
```suggestion
Both `using` forms ensure that <xref:System.IDisposable.Dispose%2A> (or <xref:System.IAsyncDisposable.DisposeAsync%2A>) is called even if an exception occurs within the `using` block. You can achieve the same result by putting the object inside a `try` block and then calling <xref:System.IDisposable.Dispose%2A> (or <xref:System.IAsyncDisposable.DisposeAsync%2A>) in a `finally` block; in fact, this is how the `using` statement and the `using` declaration are translated by the compiler. The code example earlier expands to the following code at compile time (note the extra curly braces to create the limited scope for the object):
```

docs/csharp/language-reference/keywords/using-statement.md
```diff
@@ -18,17 +18,17 @@ The following example shows how to use the `using` statement.
 
 :::code language="csharp" source="snippets/usings.cs" id="SnippetFirstExample":::
 
-Beginning with C# 8.0, you can use the following alternative syntax for the `using` statement that doesn't require braces:
+Beginning with C# 8.0, the *using declaration* doesn't require braces:
```
gewarren, line 21:
```suggestion
The *using declaration*, which was introduced in C# 8.0, doesn't require braces:
```

gewarren (approved):
Thanks Bill. I left a couple suggestions.

PR #28543, opened 2022-03-09
Update trimming libraries docs

jtschuster:
## Summary

Updates the docs about preparing libraries for trimming. Moves `<EnableTrimAnalyzer>` higher in the list and rewords the sample application documentation

### DIFF
--- a/docs/core/deploying/trimming/prepare-libraries-for-trimming.md
+++ b/docs/core/deploying/trimming/prepare-libraries-for-trimming.md
@@ -27,19 +27,27 @@ These instructions show how to enable and resolve static analysis warnings to pr
 > [!TIP]
 > Ensure you're using the .NET 6 SDK or later for these steps. They will not work correctly in previous versions.
 
-### Set IsTrimmable
+### Enable the Roslyn analyzer
 
-Set `<IsTrimmable>true</IsTrimmable>` (in .NET 6+) in your library project. This will mark your assembly as "trimmable". Being trimmable means when your library is used in a trimmed application, the assembly can have its unused members trimmed in the final output.
+The quickest and easiest way to see trim warnings is to enable the Roslyn analyzer for trim compatibility. The Roslyn analyzer is useful for quick feedback in your IDE and catches many issues, but it's currently incomplete. It doesn't cover all trim analysis warnings, but the set of patterns it understands will improve over time to give more complete coverage. The Roslyn analyzer also isn't able to analyze the implementations of reference assemblies that your library depends on. Because the analysis is incomplete, it's important to follow the steps outlined in the [Show all warnings](#show-all-warnings-with-sample-application) to ensure that your library is fully compatible with trimming.
 
-Setting `<IsTrimmable>true</IsTrimmable>` enables a Roslyn analyzer for trim compatibility. The Roslyn analyzer is useful for quick feedback in your IDE, but it's currently incomplete. It doesn't cover all trim analysis warnings, but the set of patterns it understands will improve over time to give more complete coverage. The Roslyn analyzer also isn't able to analyze the implementations of reference assemblies that you depend on. It's important to follow the steps outlined in the rest of this article to ensure that your library is fully compatible with trimming.
+#### Set `EnableTrimAnalyzer`
 
-Alternatively, you can just set `<EnableTrimAnalyzer>true</EnableTrimAnalyzer>` (in .NET 6+) in your library project. This will not have any effect on the output, but it will enable trim analysis during build via the Roslyn analyzer.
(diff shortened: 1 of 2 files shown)


docs/core/deploying/trimming/prepare-libraries-for-trimming.md
```diff
@@ -27,27 +27,35 @@ These instructions show how to enable and resolve static analysis warnings to pr
 > [!TIP]
 > Ensure you're using the .NET 6 SDK or later for these steps. They will not work correctly in previous versions.
 
-### Set IsTrimmable
+### Enable the Roslyn analyzer
 
-Set `<IsTrimmable>true</IsTrimmable>` (in .NET 6+) in your library project. This will mark your assembly as "trimmable". Being trimmable means when your library is used in a trimmed application, the assembly can have its unused members trimmed in the final output.
+The quickest and easiest way to see trim warnings is to enable the Roslyn analyzer for trim compatibility. The Roslyn analyzer is useful for quick feedback in your IDE and catches many issues, but it's currently incomplete. It doesn't cover all trim analysis warnings, but the set of patterns it understands will improve over time to give more complete coverage. The Roslyn analyzer also isn't able to analyze the implementations of reference assemblies that your library depends on. Because the analysis is incomplete, it's important to follow the steps outlined in the [Show all warnings](#show-all-warnings-with-sample-application) to ensure that your library is fully compatible with trimming.
```
sbomer, line 32: I think we want to encourage people to set `IsTrimmable` over `EnableTrimAnalyzer`, because it will both enable the analyzer and allow the library to be trimmed if used in an application - so I think we should lead with `IsTrimmable` as the very first recommendation. Maybe we could make it more clear in the very first sentence of this section that it enables the analyzer?

docs/core/deploying/trimming/prepare-libraries-for-trimming.md
```diff
@@ -27,27 +27,35 @@ These instructions show how to enable and resolve static analysis warnings to pr
...
+
+To create your sample app, first create a separate application project with `dotnet new` and modify the project file to look like the following. You'll need to include a reference to your library, as well as add `<TrimmerRootAssembly Include="YourLibraryName" />` in an `<ItemGroup>` tag, and `<TrimmerDefaultAction>link</TrimmerDefaultAction>` in a `<PropertyGroup>` tag. No changes to the source code are necessary. Once your project file is updated, run `dotnet publish` with the [runtime identifier (RID)](../../rid-catalog.md) you want to target, and `net6.0` as the target framework.
```
sbomer, line 46: Maybe mention `PublishTrimmed` here too. I think this section would be more readable as a bullet list. I'd like this section to start with a pretty concise explanation+example of how to enable this for people who just want to reference the instructions. Then I'd put more detailed explanation after the example. How about this?:

To create your sample app, first ... add the following to your project file:
- Set `<PublishTrimmed>true</PublishTrimmed>`
- Include a reference to your library
- Add `<TrimmerRootAssembly`... to ensure that every part of the library is analyzed.
- Add `<TrimmerDefaultAction` ... to prevent analysis of unused parts of assemblies which your library depends on.

... (project file and publish command) ...

Then later include more details about what `TrimmerRootAssembly` and `TrimmerDefaultAction` do, and remove the bullets there.

docs/core/deploying/trimming/prepare-libraries-for-trimming.md
```diff
@@ -64,13 +72,9 @@ The extra step of creating an app project just to get complete warnings for a li
 ```
 
 ```dotnetcli
-dotnet publish -c Release
+dotnet publish -c Release -r <RID> -f net6.0
```
sbomer, line 75: For the example, I would set the RID either in the project file, or on the command-line, not both. Maybe we could point to the self-contained publish docs https://docs.microsoft.com/en-us/dotnet/core/deploying/deploy-with-cli#self-contained-deployment if you think it's worth clarifying.

I'd suggest setting `RID` on the command-line only, and the TFM in the project file only, to match those docs.

sbomer (commented):
Thanks for updating this!

docs/core/deploying/trimming/prepare-libraries-for-trimming.md
jtschuster: I agree, that's much cleaner.

docs/core/deploying/trimming/prepare-libraries-for-trimming.md
```diff
@@ -27,28 +27,37 @@ These instructions show how to enable and resolve static analysis warnings to pr
 > [!TIP]
 > Ensure you're using the .NET 6 SDK or later for these steps. They will not work correctly in previous versions.
```
agocke, line 28: We may also want to note that the best results here will come from installing the .NET 7 preview SDK.

agocke (approved):
LGTM, thanks!

PR #28544, opened 2022-03-09
Update interfaces.md

absundr:
## Summary

Fixes a typo in the documentation

### DIFF
--- a/docs/csharp/fundamentals/types/interfaces.md
+++ b/docs/csharp/fundamentals/types/interfaces.md
@@ -28,7 +28,7 @@ Interfaces can contain instance methods, properties, events, indexers, or any co
 
 To implement an interface member, the corresponding member of the implementing class must be public, non-static, and have the same name and signature as the interface member.
 
-When a class or struct implements an interface, the class or struct must provide an implementation for all of the members that the interface declares but doesn't provide a default implementation for the interface. However, if a base class implements an interface, any class that's derived from the base class inherits that implementation.
+A class or struct that implements an interface must provide an implementation for all declared members without a default implementation provide by the interface. However, if a base class implements an interface, any class that's derived from the base class inherits that implementation.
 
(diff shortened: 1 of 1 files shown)


docs/csharp/fundamentals/types/interfaces.md
```diff
@@ -28,7 +28,7 @@ Interfaces can contain instance methods, properties, events, indexers, or any co
 
 To implement an interface member, the corresponding member of the implementing class must be public, non-static, and have the same name and signature as the interface member.
 
-When a class or struct implements an interface, the class or struct must provide an implementation for all of the members that the interface declares but doesn't provide a default implementation for. However, if a base class implements an interface, any class that's derived from the base class inherits that implementation.
+When a class or struct implements an interface, the class or struct must provide an implementation for all of the members that the interface declares but doesn't provide a default implementation for the interface. However, if a base class implements an interface, any class that's derived from the base class inherits that implementation.
```
BillWagner, line 31: I agree that the original language is awkward. I'm not sure this helps. How about the following wording:

```suggestion
A class or struct that implements an interface must provide an implementation for all declared members without a default implementation provide by the interface. However, if a base class implements an interface, any class that's derived from the base class inherits that implementation.
```

BillWagner (commented):
Thanks for making this suggestion @abhijithsundr 

I proposed an alternative wording for the change. Let me know what you think and then I'll merge this.

docs/csharp/fundamentals/types/interfaces.md
abhijithsundr: Great, this works.

abhijithsundr: Thank you

BillWagner (approved):
Thanks again @abhijithsundr 

I'll :shipit: now.

PR #28587, opened 2022-03-09
Incorrect output in example

NikRimington:
When running this code the output comment does not match the actual output. 
It actually outputs `pst` where as the comment said it output `PDT`

## Summary

I updated the comment to show the correct expected output.


tdykstra (approved):
Thanks @NikRimington

Issue #28584, opened 2022-03-09
Orleans xref sweep


IEvangelist:
Update all existing content that references files in the `dotnet/orleans` repository to instead use the xref syntax that will rely on the Orleans API reference docs. Sweep each of the following folders:

- [x] #28598 _deployment_ folder
- [x] #28700 _grains_ folder
- [x] #28703 _host_ folder
- [x] #28722 _implementation_ folder
- [x] #28749 _migration_ folder
- [x] #28752 _resources_ folder
- [ ] #28754 _streaming_ folder
- [ ] #28754 _tutorial-and-samples_ folder
- [ ] #28754 _whats-new_ folder

PR #28598, opened 2022-03-09
Replace links with xrefs of type in the deployment folder

IEvangelist:
## Summary

Replace links with xrefs of type in the _deployment_ folder

Part of #28584


adegeo (approved):
LGTM

PR #28603, opened 2022-03-10
Fix documention on specifying connection pooling settings

JasonLoKiSmith:
## Summary

Describe your changes here.
The documentation specifies that a developer must specify minimum connections in a pool as "MinPoolSize", this is incorrect. There needs to be some spaces in between for it to work correctly


tdykstra (approved):
Thanks @JasonLoKiSmith

PR #28614, opened 2022-03-10
Remove example of checking an instance being of nullable type

### LINKED ISSUE #28597: Duplicate example at section "How to identify a nullable value type" of article "Nullable value types"
The first and last example of the section "How to identify a nullable value type" of [this](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types#how-to-identify-a-nullable-value-type) article "Nullable value types" are too similar, consider removing the last example. 

---
#### Document Details

* ID: b3e65a97-5e33-ab78-cd40-b20c3219ca2f
* Version Independent ID: 1ae0aed9-e565-44ae-aebe-c55a0a02c2c5
* Content: [Nullable value types - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types)
* Content Source: [docs/csharp/language-reference/builtin-types/nullable-value-types.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/builtin-types/nullable-value-types.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BartoszKlonowski:
This pull request fixes #28597 

It removes the last example of checking whether an instance is of nullable type and replaces it with the brief explanation how to achieve that.
This explanation points (without links between sections) to the first example from the same section and mention the `typeof` operator as an additional tool needed if there's an instance to be checked instead of a type.

**NOTE:** The last example has been reported as duplicated, but in fact it is quite different case. The first applies to the type being checked, while the last show the same goal but for instance.
However they are *very* similar so it may introduce some confusions and I decided to go with the short explanation mentioning the way to do such check.

### DIFF
--- a/docs/csharp/language-reference/builtin-types/nullable-value-types.md
+++ b/docs/csharp/language-reference/builtin-types/nullable-value-types.md
@@ -109,10 +109,6 @@ Also, don't use the [is](../operators/type-testing-and-cast.md#is-operator) oper
 
 [!code-csharp-interactive[is operator example](snippets/shared/NullableValueTypes.cs#IsOperator)]
 
-You can use the code presented in the following example to determine whether an instance is of a nullable value type:
-
-[!code-csharp-interactive[whether an instance is of a nullable type](snippets/shared/NullableValueTypes.cs#IsInstanceNullable)]
-
 > [!NOTE]
 > The methods described in this section are not applicable in the case of [nullable reference types](nullable-reference-types.md).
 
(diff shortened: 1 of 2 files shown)


BillWagner (approved):
Thanks @BartoszKlonowski 

This is great. 

I had submitted a different PR that fixes the build error. Once I verify that, I'll merge this.

BillWagner:

PR #28615, opened 2022-03-10
Update snippets for NRT

tdykstra:
Contributes to #27310

Hide whitespace when reviewing - changing to file-scoped namespaces changes indentation, which makes it look like the whole file was changed if you don't hide whitespace.

### DIFF
--- a/docs/csharp/fundamentals/object-oriented/snippets/objects/Equality.cs
+++ b/docs/csharp/fundamentals/object-oriented/snippets/objects/Equality.cs
@@ -1,7 +1,4 @@
-
-using System;
-
-public class equality
+﻿public class equality
 {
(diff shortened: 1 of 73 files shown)


docs/csharp/fundamentals/object-oriented/snippets/objects/Equality.cs
```diff
@@ -1,7 +1,4 @@
-
-using System;
-
-public class equality
+﻿public class equality
```
gewarren, line 1:
```suggestion
﻿public class Equality
```

gewarren (approved):
LGTM

tdykstra:
Thanks for the review @gewarren !

PR #28619, opened 2022-03-10
add documentation for new CLI telemetry in .NET 6.0.300

baronfel:
## Summary

With https://github.com/dotnet/sdk/pull/24166 the .NET CLI added the ability to detect the presence of CI systems during invocation. We need to document the data checked accordingly.


tdykstra (approved):
LGTM

baronfel:
@tdykstra would you suggest to hold changes like this until the matching version actually publically is released? That's great if so, I just wanted to be ahead of the curve here.

tdykstra:
If we're only a few days ahead I think it's fine to merge this now, otherwise we can put a do not merge label on it for now.

baronfel:
I removed the DO NOT MERGE label since 6.0.300 has released. This is ready to go!

PR #28624, opened 2022-03-10
generic attributes are allowed in C# 11

BillWagner:
The note in the generics and attributes article said that generic classes can't derive from `System.Attribute`. That changes in C# 11.

While making that change, move snippets into this folder.


IEvangelist (approved):
I really could have used generic attributes just recently! Looks good to me :shipit:

PR #28629, opened 2022-03-11
Remove mention implicit await using call

bitbonk:
Maybe the idea of this sentence was something else but as it stands now, I think the `DisposeAsync()` method is not generally called implicitly in an `await using` statement.  So I suggest to remove this to prevent confusion.

## Summary

Describe your changes here.


bitbonk:
My bad, I misread that as `await using` is implicitly called.

PR #28637, opened 2022-03-11
HostBuilder is now the preferred way to build a silo.

richorama:
## Summary

in the Orleans documentation the code snippets used for creating a silo host are out of date. The preferred technique is to use a HostBuilder rather than a SiloHostBuilder.


IEvangelist (approved):
This looks good to me, but I want to verify with @ReubenBond and @benjaminpetit first.

benjaminpetit:
@richorama is correct, we even removed `SiloHostBuilder` in 4.0

PR #28638, opened 2022-03-11
SEO OKR

tdykstra:
Contributes to #28383 

* Cross Platform P/Invoke

  Organic search January: 17%, February: 63%. The February rate is the norm for this article.  It normally gets around 1,000 PV/month, but on one day, January 13, it got 2,585 PV from Microsoft referrals, skewing the organic search percentage down to 17%. Number of visitors stayed low at 34 for the day, so all of the extra traffic came from one or a few visitors. So 63% is the accurate organic search rate and there is not necessarily an SEO problem, but I added an intro paragraph with the keywords that were provided in the meta description.

* Dapr for .NET Developers

  Organic search 16%. This is the cover page for an e-book. Most of the traffic comes from hub and landing pages, and other pages from the same e-book. A change that might improve SEO would be to spell out the name Distributed Application Runtime somewhere besides just the meta description, but the page can't be modified without getting out of sync with the e-book.

* Architectural principles

  This is an e-book chapter, so any changes to the article would make it out of sync with the e-book it's a copy of.

* C# template changes in .NET 6

  Organic search 18%. Direct traffic rate is 75% due to an aka.ms link included in a code comment in the Program.cs file created by the .NET 6 C# console application project template (aka.ms/new-console-template). I changed title, h1, and first paragraph to emphasize the top keywords -- .NET 6, console app, and program.cs.

* Define types and their members - A tour of C#

  Organic search 15%. This is part of the "tour of C#" series, and most of the traffic comes from "next" and "previous" buttons and menu selections -- 70% of traffic is Microsoft. All of the top keywords include C#, but that was missing from the h1, so I added it. There are similar C# pages competing for the same keywords as this article, for example, "define C# classes" leads to the [C# fundamentals article on classes](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/types/classes), which is actually a better result than this article would be for that search.

* General Structure of a C# Program

  Organic search 16%. This is part of a series in the C# fundamentals section, and Microsoft traffic is 83%, mostly from other fundamentals pages, tour of C# pages, and the C# landing page. A search for "C# program structure" produces this article as the first result and highlights the first paragraph as the quick question answer. No changes needed.

* The building blocks of C# programs

  Organic search 7%. This is a Tour of C# article and most of its traffic comes from those pages and other C# pages. Few searches look for "building blocks", so I revised the first paragraph to specify some of the names most likely to be searched for.

The Tour of C# articles were inconsistent in how they included the series title in the article title -- one had "A tour of C#" in front of the article title, two had it after the article title, and one didn't have it at all.  I made all use the prefix arrangement.

### DIFF
--- a/docs/core/tutorials/top-level-templates.md
+++ b/docs/core/tutorials/top-level-templates.md
@@ -1,6 +1,7 @@
 ---
 title: "C# console app template changes in .NET 6"
 description: The .NET 6 project template for C# console apps uses top-level statements. Understand what's changed and how to use existing learning materials with the new syntax.
+ms.custom: updateeachrelease
 ms.date: 03/14/2022
 ---
 # .NET 6 C# console app template generates top-level statements
@@ -29,7 +30,7 @@ namespace MyApp // Note: actual namespace depends on the project name.
 }
 ```
 
-These two forms represent the same program. Both are valid with C# 10.0. When you use the newer version, you only need to write the body of the `Main` method. You don't need to include the other program elements. You have two options to work with tutorials that haven't been updated yet for .NET 6:
+These two forms represent the same program. Both are valid with C# 10.0. When you use the newer version, you only need to write the body of the `Main` method. You don't need to include the other program elements. You have two options to work with tutorials that haven't been updated to use .NET 6+ templates:
 
 - Use the new program style, adding new top-level statements as you add features.
 - Convert the new program style to the older style, with a `Program` class and a `Main` method.
--- a/docs/core/tutorials/top-level-templates.md
+++ b/docs/core/tutorials/top-level-templates.md
@@ -1,18 +1,18 @@
 ---
-title: "C# template changes in .NET 6"
-description: The C# console app template now uses top-level statements. Understand what's changed and how to use existing learning materials with the new syntax.
-ms.date: 03/11/2022
+title: "C# console app template changes in .NET 6"
+description: The .NET 6 project template for C# console apps uses top-level statements. Understand what's changed and how to use existing learning materials with the new syntax.
+ms.date: 03/14/2022
 ---
-# New C# templates generate top-level statements
+# .NET 6 C# console app template generates top-level statements
 
-Starting with .NET 6, new C# projects using the `console` template generate different code than previous versions:
+Starting with .NET 6, the project template for new C# console apps generates the following code in the *Program.cs* file:
 
 ```csharp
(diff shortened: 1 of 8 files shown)


jkoritzinsky (approved):
P/Invoke docs changes look good to me

docs/csharp/tour-of-csharp/program-building-blocks.md
```diff
@@ -47,13 +51,13 @@ In the following example, each instance of the `Color` class has a separate copy
 
 :::code language="csharp" source="./snippets/shared/ClassesObjects.cs" ID="ColorClassDefinition":::
 
-As shown in the previous example, *read-only fields* may be declared with a `readonly` modifier. Assignment to a read-only field can only occur as part of the field’s declaration or in a constructor in the same class.
+As shown in the previous example, *read-only fields* may be declared with a `readonly` modifier. Assignment to a read-only field can only occur as part of the field's declaration or in a constructor in the same class.
 
 ## Methods
```
gewarren, line 56: Should this be nested under "Members"?

gewarren (approved):
Great work!

docs/core/tutorials/top-level-templates.md
```diff
@@ -1,18 +1,18 @@
 ---
-title: "C# template changes in .NET 6"
-description: The C# console app template now uses top-level statements. Understand what's changed and how to use existing learning materials with the new syntax.
-ms.date: 02/28/2022
+title: "C# console app template changes in .NET 6"
+description: The .NET 6 project template for C# console apps uses top-level statements. Understand what's changed and how to use existing learning materials with the new syntax.
+ms.date: 03/14/2022
 ---
-# New C# templates generate top-level statements
+# .NET 6 C# console app template generates top-level statements
```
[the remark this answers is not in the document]
tdykstra: Good point, but given that plans are being laid to handle this differently in .NET 7 I'll add ms.custom: updateeachrelease instead as a reminder to update this doc when a different approach is implemented.

docs/csharp/tour-of-csharp/program-building-blocks.md
tdykstra: Logically it should, but as a practical matter it would make a very big h2 section with h3s into an h3 with a bunch of h4s, and then **Fields** and **Other function members** would have to be demoted too, so I think keeping it as an h2 makes sense.

PR #28666, opened 2022-03-15
Update Snippet for the event pattern documentation

yrsahl:
Moved \<SnippetDeclareEventHandler\> to include the instantiation of the FileSearcher class as 'fileLister'. 
As it stands the documentation shows 'fileLister' being referred to with no explanation where it originated, making it ambiguous.


BillWagner (approved):
Thanks for spotting this and making the fix @YannikSahl 

Thanks again.

PR #28707, opened 2022-03-16
Add notes about arity and file system types

### LINKED ISSUE #28706: Add example for FileSystemInfo
* Add FileSystemInfo example
* Explain how arity is determined by default
* Explain that required option doesn't have to be provided on command line if it has a default value

---
#### Document Details

* ID: 75416731-e081-9d04-e683-e0fba0d6b388
* Version Independent ID: 1eec00b7-55ba-34ce-225e-3549b9de4b25
* Content: [How to bind arguments to handlers in System.CommandLine](https://docs.microsoft.com/en-us/dotnet/standard/commandline/model-binding)
* Content Source: [docs/standard/commandline/model-binding.md](https://github.com/dotnet/docs/blob/main/docs/standard/commandline/model-binding.md)
* GitHub Login: @gewarren

tdykstra:
Fixes #28706


docs/standard/commandline/model-binding.md
```diff
@@ -149,13 +149,17 @@ Because `AllowMultipleArgumentsPerToken` is set to `true`, the following input r
 
 ### File system types
 
-Since command line applications often have to work with the file system, `FileInfo` and `DirectoryInfo` are supported. The following example shows the use of `FileInfo`, but it would work also with `DirectoryInfo`.
+Command line applications that work with the file system can use the <xref:System.IO.FileSystemInfo>, <xref:System.IO.FileInfo>, and <xref:System.IO.DirectoryInfo> types. The following example shows the use of `FileSystemInfo`:
```
gewarren, line 152:
```suggestion
Command-line applications that work with the file system can use the <xref:System.IO.FileSystemInfo>, <xref:System.IO.FileInfo>, and <xref:System.IO.DirectoryInfo> types. The following example shows the use of `FileSystemInfo`:
```

docs/standard/commandline/define-commands.md
```diff
@@ -146,6 +148,15 @@ Options:
   -?, -h, --help          Show help and usage information
 ```
 
+
+## Set argument arity
+
+You can explicitly set argument [arity](syntax.md#argument-arity) by using the `Arity` property, but in most cases that is not necessary. `System.CommandLine` automatically determines the argument arity based on the argument type:
+
+* `Boolean` arguments - `ZeroOrOne`
```
gewarren, line 156: Are these enum fields or ? If so, maybe put this info in a table with a column header.

gewarren (approved):

docs/standard/commandline/define-commands.md
```diff
@@ -132,6 +132,8 @@ If the command line for this example app doesn't include `--endpoint`, an error
 Option '--endpoint' is required.
 ```
 
+If a required option has a default value, the option doesn't have to be specified on the command line. In that case, the default value provides the required option value.
+
 ## Hidden commands, options, and arguments
 
 You might want to support a a command, option, or argument, but avoid making it easy to discover. For example, it might be a deprecated or administrative or preview feature. Use the `IsHidden` property to prevent users from discovering such features by using tab completion or help, as shown in the following example:
```
jonsequitur, line 139: There's a duplicate `a` here: `You might want to support a a command`.

docs/standard/commandline/define-commands.md
jonsequitur: They are static members of the `ArgumentArity` type.

docs/standard/commandline/model-binding.md
```diff
@@ -149,13 +149,17 @@ Because `AllowMultipleArgumentsPerToken` is set to `true`, the following input r
...
+:::code language="csharp" source="snippets/model-binding/csharp/Program.cs" id="filesysteminfo" :::
+
+With `FileInfo` and `DirectoryInfo` the pattern matching code is not required:
+
+:::code language="csharp" source="snippets/model-binding/csharp/Program.cs" id="filesysteminfo" :::
 
 ### Anything with a string constructor
 
-`FileInfo` and `DirectoryInfo` aren't special cases. Any type having a constructor that takes a single string parameter can be bound in this way. For example, the same code works with a `Uri` instead.
+The file system types aren't special cases. Any type having a constructor that takes a single string parameter can be bound in this way. For example, code that would work with `FileInfo` works with a <xref:System.Uri> instead.
```
jonsequitur, line 162: This is no longer true after the work to support NativeAOT compilation, unfortunately. These are special cases. (`System.Uri` is another, and I expect the list might grow over time).

docs/standard/commandline/define-commands.md
tdykstra: Good idea. It's not an enum but it works like one.

docs/standard/commandline/model-binding.md
tdykstra: Are these now the only special cases or are there others that we should mention?

jonsequitur: Technically, pretty much all types are now special cases. The two main conventions we used before doing the trimming work were the single-string constructor convention and the presence of a `TypeConverter`. Removing the latter also meant types like `DateTime` and `int` needed to be accounted for with special cases. You can see the implementations here: https://github.com/dotnet/command-line-api/blob/021ec68a4cb510c2cc125c6ebb78b9cfd4e3847a/src/System.CommandLine/Binding/ArgumentConverter.StringConverters.cs#L11

docs/standard/commandline/model-binding.md
```diff
@@ -157,12 +157,30 @@ With `FileInfo` and `DirectoryInfo` the pattern matching code is not required:
...
 
-### Anything with a string constructor
+### Other supported types
 
-The file system types aren't special cases. Any type having a constructor that takes a single string parameter can be bound in this way. For example, code that would work with `FileInfo` works with a <xref:System.Uri> instead.
+Many types that have a constructor that takes a single string parameter can be bound in this way. For example, code that would work with `FileInfo` works with a <xref:System.Uri> instead.
 
 :::code language="csharp" source="snippets/model-binding/csharp/Program.cs" id="uri" :::
 
+Besides the file system types and `Uri`, the following types are supported:
```
jonsequitur, line 166: I expect this list will continue to grow until we can find a generalized approach that's also supportable via NativeAOT compilation.

baronfel: I suspect we'll end up at the static interfaces feature that the C# team is working on as a long-term, reflection-free solution. But that's just my guess.

PR #28747, opened 2022-03-21
Add NoDefaultExcludes to the templatepack.csproj

Gimly:
## Summary

The `NoDefaultExcludes` removes the default behaviors of `dotnet pack` that doesn't pack all the files and folders that starts with a `.`, such as `.gitignore`.

In the case of templates, we would probably want those files to be available if they have been defined in the template. This default behavior therefore doesn't make sense for the templates.


tdykstra:
ping @adegeo

adegeo (approved):
This looks like a good idea.

PR #28781, opened 2022-03-23
Change invalid ReadyToRun CLI argument to PublishReadyToRun

huan086:
Ready to run is controlled via the property `PublishReadyToRun`. The `ReadyToRun` property does not work.

## Summary

As per [ready to run documentation](https://docs.microsoft.com/en-us/dotnet/core/deploying/ready-to-run), the property to use is `PublishReadyToRun`


gewarren (approved):
Thank you @huan086

PR #28788, opened 2022-03-24
Ping breaking change

### LINKED ISSUE #28720: [Breaking change]: PlatformNotSupportedException on non-privileged Linux processes when custom payload is specified to Ping
### Description

On Linux, non-privileged processes cannot send raw IP packets and Ping functionality is implemented by interfacing with the `ping` utility. This utility, however, does not support specifying custom payload for the ICMP ping packets so the payload provided by user code was ignored.

PR https://github.com/dotnet/runtime/pull/64625 introduced a check which throws an exception in such cases.

Original issue: https://github.com/dotnet/runtime/issues/62458

### Version

.NET 7 Preview 2

### Previous behavior

Ping packet payload is silently ignored (i.e. not sent) on non-privileged Linux processes.

### New behavior

Attempt to send custom payload when running in non-privileged Linux process leads to `PlatformNotSupportedException` being thrown.

### Type of breaking change

### Reason for change

It is better to signal to the user that the operation cannot be performed instead of silently dropping the payload.

### Recommended action

If ping payload is necessary, run application as `root` or grant `cap_net_raw` capability using `setcap` utility.

Otherwise, use an overload of `Ping.SendPingAsync` which does not accept the custom payload, or pass in an empty array.

### Feature area

Networking

### Affected APIs

All overloads of the following methods which accept `byte[] buffer` parameter
Ping.Send
Ping.SendAsync
Ping.SendPingAsync

gewarren:
Fixes #28720 

[Internal preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/networking/7.0/ping-custom-payload-linux?branch=pr-en-us-28788).

### DIFF
--- a/docs/core/compatibility/7.0.md
+++ b/docs/core/compatibility/7.0.md
@@ -37,6 +37,7 @@ If you're migrating an app to .NET 7, the breaking changes listed here might aff
 | Title | Binary compatible | Source compatible | Introduced |
 | - | :-: | :-: | - |
 | [AllowRenegotiation default is false](networking/7.0/allowrenegotiation-default.md) | ❌ | ❌ | Preview 3 |
+| [Custom ping payloads on Linux](networking/7.0/ping-custom-payload-linux.md) | ❌ | ✔️ | Preview 2 |
 
 ## Serialization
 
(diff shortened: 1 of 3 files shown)


BillWagner (approved):
This LGTM. You can :shipit: when you're ready.

rzikm (approved):
LGTM

PR #28826, opened 2022-03-28
Add EventSource docs

noahfalk:
This is an initial set of docs for using EventSource. The goal was to
capture most of the beginner + intermediate material from
https://github.com/Microsoft/dotnet-samples/blob/master/Microsoft.Diagnostics.Tracing/EventSource/docs/EventSource.md
For brevity I dropped some info about lesser used functionality
(event localization and ETW channel support) while adding some
more info on usage outside the PerfView+ETW ecosystem. I also
added information on ActivityIDs which I haven't found documented
anywhere.

After this PR goes in I'd like to trim the EventSource class api-docs
and add a link to this content as the main place to show overall usage. I
also want to add a link to this content from the dotnet-samples github
page and mark that one as being no longer maintained.

This PR is intended to address
dotnet/diagnostics#2908 - EventSource start/stop
dotnet/diagnostics#2904 - "fast" eventsource
dotnet/diagnostics#2902 - custom EventSource with dotnet-trace+vs
dotnet/diagnostics#1347 - activity ids

It partially covers dotnet/diagnostics#2445 - special case LogAlways behavior
but I think we should do more in the API docs before marking that one complete

## Internal previews

- [EventSource](https://review.docs.microsoft.com/en-us/dotnet/core/diagnostics/eventsource?branch=pr-en-us-28826)
- [Getting Started with EventSource](https://review.docs.microsoft.com/en-us/dotnet/core/diagnostics/eventsource-getting-started?branch=pr-en-us-28826)
- [Instrument Code to Create EventSource Events](https://review.docs.microsoft.com/en-us/dotnet/core/diagnostics/eventsource-instrumentation?branch=pr-en-us-28826)
- [Collect and View EventSource Traces](https://review.docs.microsoft.com/en-us/dotnet/core/diagnostics/eventsource-collect-and-view-traces?branch=pr-en-us-28826)
- [EventSource Activity IDs](https://review.docs.microsoft.com/en-us/dotnet/core/diagnostics/eventsource-activity-ids?branch=pr-en-us-28826)


noahfalk:
There were some numbered lists in the collect-and-view page that keep repeating "1." I tried replacing that with numbers that count upwards 1,2,3... but then the linter complains that they aren't expected to increment. I'm not sure how to both make the page look visually correct and make the linter happy?

noahfalk:
@dotnet/dotnet-diag

IEvangelist (changes requested):
Hi @noahfalk,

I started reviewing this, and then I realized that I was suggesting a lot of changes. Could you please revise this a bit, before I jump back into a full review? Here are the priorities:

1. The documentation writing style should be active instead of passive. For example:
  > We can run the code below to see Activity tracking in action.

  _Becomes_:

  > You can run the code below to see Activity tracking in action.
2. Do not manually insert new lines, instead let your text editor wrap for you.

noahfalk:
Thanks @IEvangelist, I'll make a pass.

> Do not manually insert new lines, instead let your text editor wrap for you.

This feels awkward if you mean what I think you mean. If there was a 10 line long paragraph at ~100 characters per line are you suggesting the markdown file viewed in notepad would have a 1000 character line with a single newline at the end and a large horizontal scroll bar?
My editor for this was notepad which isn't going to wrap automatically for me. However even if I change to use a different editor that does auto-wrap I've seen issues in the past with this approach:
- GitHub suggestions are per file line making the comments very coarse and the suggestion text large
- other developers reading the text in their editors may not have auto-wrapping
- build errors are more challenging to locate within such long lines
In the past I got requests from other reviewers that they would appreciate if I did not rely on auto-wrapping because it made it awkward for them to work with the text. If there is a general consensus among the doc team that auto-wrapping is the preferred style I'll follow it, but if it is a matter of personal preference then I'd rather leave it as-is.

noahfalk:
And thanks @gewarren, I see you already made a bunch of improvements : D

noahfalk:
I notice that in GitHub's side by side view it doesn't show many characters per line which caused it to wrap more aggresively. Switching to the unified view shows far more characters per line, but if it is an issue I could probably find something that reformats the text to wrap with fewer characters per line.

docs/core/diagnostics/eventsource-activity-ids.md
```diff
@@ -0,0 +1,287 @@
...
+    {
+        DemoEventSource.Log.DebugMessage("HelperB");
+        await Task.Delay(100); // pretend to do some work
+    }
+}
+
+[EventSource(Name ="Demo")]
+class DemoEventSource : EventSource
+{
+    public static DemoEventSource Log = new DemoEventSource();
```
gfoidl, line 90: Either
```suggestion
    public static readonly DemoEventSource Log = new DemoEventSource();
```
or
```suggestion
    public static DemoEventSource Log { get; } = new DemoEventSource();
```
to have it as static property.

The same in other code examples showing an `EventSource`.

noahfalk:
I made the editing pass, but didn't change word wrapping at this point. If there is a doc team consensus that newlines should only be inserted at paragraph breaks (or some other heuristic?) then let me know and I will.

tdykstra:
> If there is a general consensus among the doc team that auto-wrapping is the preferred style I'll follow it, but if it is a matter of personal preference then I'd rather leave it as-is.

It's a matter of personal preference, but most of our .NET docs use line breaks only to separate paragraphs.

docs/core/diagnostics/eventsource-activity-ids.md
noahfalk: Updated all the examples to use a getter

IEvangelist (approved):

noahfalk:
Thanks all!

PR #28833, opened 2022-03-28
Fixing references to Attribute Type in Code

shhafiz:
Since the attribute type was named as AuthorAttribute, the successive use of this attribute should match this name.

## Summary

Describe your changes here.


shhafiz:
Incorrect understanding, the published code is fine as is.

PR #28848, opened 2022-03-29
Use net7 preview

BillWagner:
Switch our CI build to use .NET 7 preview.

### DIFF
--- a/.github/workflows/build-validation.yml
+++ b/.github/workflows/build-validation.yml
@@ -9,7 +9,7 @@ on:
     types: [opened, synchronize, reopened]
 
 env:
-  DOTNET_INSTALLER_CHANNEL: '6.0'
+  DOTNET_INSTALLER_CHANNEL: '7.0'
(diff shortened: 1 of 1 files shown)


tdykstra (approved):

.github/workflows/build-validation.yml
```diff
@@ -7,9 +7,15 @@ on:
   pull_request:
     branches: [ main ]
     types: [opened, synchronize, reopened]
+  workflow_dispatch:
+    inputs:
+      reason:
+        description: 'The reason for running the workflow'
+        required: true
+        default: 'Manual run'
```
tdykstra, line 15: Was the workflow-dispatch addition intentional?

BillWagner: Yes. Once this gets merged, you can run tests on your fork before opening a PR. 

(Notice that I had to make one additional change, because I missed one flag that needs to change)

PR #28878, opened 2022-03-31
Fix fsharp azure tables

### LINKED ISSUE #28828: F# Azure Table Storage sample is broken
**Issue description**

The [F# Azure Table Storage sample](https://docs.microsoft.com/en-us/dotnet/fsharp/using-fsharp-on-azure/table-storage) is broken. The main culprit is this:

```fsharp
type Customer(firstName, lastName, email: string, phone: string) =
    interface ITableEntity with
        member this.ETag
            with get (): ETag = 
                raise (System.NotImplementedException())
            and set (v: ETag): unit = 
                raise (System.NotImplementedException())
        member this.PartitionKey
            with get (): string = 
                raise (System.NotImplementedException())
            and set (v: string): unit = 
                raise (System.NotImplementedException())
        member this.RowKey
            with get (): string = 
                raise (System.NotImplementedException())
            and set (v: string): unit = 
                raise (System.NotImplementedException())
        member this.Timestamp
            with get (): Nullable<DateTimeOffset> = 
                raise (System.NotImplementedException())
            and set (v: Nullable<DateTimeOffset>): unit = 
                raise (System.NotImplementedException())
    new() = Customer(null, null, null, null)
    member val Email = email with get, set
    member val PhoneNumber = phone with get, set
    member val PartitionKey = lastName with get, set
    member val RowKey = firstName with get, set
```

The member properties all return exceptions. I'm not sure if this code was accidentally added instead of the valid code or never tested, but if you copy this code and try it out, it'll pop as soon as you try to access e.g. PKey or RKey.

There's no direct equivalent C# sample but I looked for some on docs and found a few e.g. [here](https://docs.microsoft.com/en-us/azure/cosmos-db/table/create-table-dotnet?tabs=azure-portal%2Cvisual-studio) which uses a very different approach of using the `TableEntity` type instead of `ITableEntity` and mapping between that and a normal POCO (something that is common to do in F# too, with a record).

I've tried fixing the sample:

```fsharp
type Customer(firstName, lastName, email: string, phone: string) =

    interface ITableEntity with
        member val PartitionKey = lastName with get, set
        member val RowKey = firstName with get, set
        member val Timestamp = Nullable() with get, set
        member val ETag = ETag null with get, set

    new() = Customer(null, null, null, null)
    member val Email = email with get, set
    member val PhoneNumber = phone with get, set
```

but this code gives an error (at least, from Azurite):

```
Azure.RequestFailedException: An error occurred while processing this request.
RequestId:d0f242b0-c511-4bd8-84bf-d072debc64e1
Time:2022-03-28T11:21:44.798Z
Status: 400 (Bad Request)
ErrorCode: InvalidInput

Content:
{"odata.error":{"code":"InvalidInput","message":{"lang":"en-US","value":"An error occurred while processing this request.\nRequestId:d0f242b0-c511-4bd8-84bf-d072debc64e1\nTime:2022-03-28T11:21:44.798Z"}}}

Headers:
Server: Azurite-Table/3.16.0
x-ms-error-code: REDACTED
x-ms-request-id: d0f242b0-c511-4bd8-84bf-d072debc64e1
x-ms-version: REDACTED
Date: Mon, 28 Mar 2022 11:21:44 GMT
Connection: keep-alive
Keep-Alive: REDACTED
Transfer-Encoding: chunked
Content-Type: application/json; odata=minimalmetadata; streaming=true; charset=utf-8

   at Azure.Data.Tables.TableRestClient.InsertEntity(String table, Nullable`1 timeout, Nullable`1 responsePreference, IDictionary`2 tableEntityProperties, QueryOptions queryOptions, CancellationToken cancellationToken)
   at Azure.Data.Tables.TableClient.AddEntity[T](T entity, CancellationToken cancellationToken)
   at <StartupCode$FSI_0095>.$FSI_0095.main@()
Stopped due to error
```

Unfortunately there's no really clue here as to what the offending field is, and I'm not sure if this is an "I've not populated the ITableEntity correctly" vs "Azurite issue" vs something else.

If you use a normal `TableEntity` this works fine e.g.

```fsharp
let t = TableEntity (dict [ "PartitionKey", "Harp"; "RowKey", "Walter"; "Email", "Walter@contoso.com"; "PhoneNumber", "425-555-0101" ])
table.AddEntity t
```

Happy to submit a PR with the correct code fix, but without e.g. seeing a valid C# sample, it's hard to do that.

isaacabraham:
## Summary

Fixes #28828. @jsquire would you care to have a look through this code sample and maybe try it out to see what you think?

* `printfn`, not `Console.WriteLine`
* `expr |> ignore`, not `ignore(expr)`
* F# lists `[ ]`, not the `Generic.List`
* Removal of unnecessary syntax elements (parentheses etc.)
* Removal of unnecessary / misleading use of `Async` methods which were immediately blocking with `.Result` - just switched to the sync method instead.

There's still some elements that I personally feel could be simplified - the printfns could be removed if you're working in a REPL e.g. VSCode / Visual Studio (or even a notebook) since the outputs from Azure would be printed to the interactive console anyway e.g. on line 70 we assign a query output to `results`, and then print the row and partition key out.

```fsharp
let results = table.Query<Customer> "PartitionKey eq 'Smith'"

for customer in results do
    printfn $"customer: {customer.RowKey} {customer.PartitionKey}"
```

This *could* be simplified to just:

```fsharp
table.Query<Customer> "PartitionKey eq 'Smith'"
```

Because F# automatically outputs the following to the interactive console:

```
val it: Pageable<Customer> =
  seq
    [FSI_0021+Customer {Email = "Ben@contoso.com";
                        PartitionKey = "Smith";
                        PhoneNumber = "425-555-0103";
                        RowKey = "Ben";};
     FSI_0021+Customer {Email = "Jeff@contoso.com";
                        PartitionKey = "Smith";
                        PhoneNumber = "425-555-0102";
                        RowKey = "Jeff";}]
```

Anyway, that's a secondary thing. The main point is that this script now works end-to-end and uses some more idiomatic F# elements.

@jsquire If you're cool with this, I can update the MD to reflect the correct line numbers afterwards.


isaacabraham:
@jsquire I tried to get it working with Records. I couldn't find a nice way of doing it unfortunately - there are too many compromises that need to be made e.g. making all fields mutable, needing to explicitly implement the `ITableEntity` interface, (apparently) not being allowed to have a property of type ETag on the record etc. - it just wasn't worth it.

My gut feel is that people in the F# world will either use this kind of class-based approach or more likely use standard F# records and then have a mapping layer (either hand-written or using e.g. reflection / code-gen) to map between the record and the built-in `TableEntity` class which is very simple to work with at the boundary of the app.

samples/snippets/fsharp/azure/table-storage.fsx
```diff
@@ -1,111 +1,90 @@
...
-            with get (): string = 
-                raise (System.NotImplementedException())
-            and set (v: string): unit = 
-                raise (System.NotImplementedException())
-        member this.Timestamp
-            with get (): Nullable<DateTimeOffset> = 
-                raise (System.NotImplementedException())
-            and set (v: Nullable<DateTimeOffset>): unit = 
-                raise (System.NotImplementedException())
+        member val ETag = ETag "" with get, set
```
jsquire, line 36: Would there be any value in having these delegate to the class-level implementation or vice-versa?

samples/snippets/fsharp/azure/table-storage.fsx
```diff
@@ -1,111 +1,90 @@
...
 // Create the table if it doesn't exist.
-ignore(table.CreateIfNotExists())
+table.CreateIfNotExists() |> ignore
 
 //
 // Add an entity to a table. The last name is used as a partition key.
 //
 
 type Customer(firstName, lastName, email: string, phone: string) =
     interface ITableEntity with
```
jsquire, line 35: Would it be worth considering a comment here to indicate that the shadowing is needed because serialization doesn't work with explicit implementations of `ITableEntity`?

jsquire (commented):
Thank you, @isaacabraham!   This looks great and definitely reads more like F# to my untrained eye.

jsquire:
> This _could_ be simplified to just:
> 
> ```fsharp
> table.Query<Customer> "PartitionKey eq 'Smith'"
> ```
> 
> Because F# automatically outputs the following to the interactive console:

In your opinion, would it be more idiomatic to just assume the FSI behavior?   There doesn't seem to be any trick or complexity being demonstrated to extract the results, so I don't think we'd lose anything important.

samples/snippets/fsharp/azure/table-storage.fsx
isaacabraham: Good question. I've never even done something like this before :-) It looks very much like if you have an interface and class member with the same name and type, F# seems to treat them as one and the same. I've not looked at the IL or decompiled it into C# to see. @cartermp can you confirm what happens there?

samples/snippets/fsharp/azure/table-storage.fsx
isaacabraham: Yes, I'll add that in as a comment. As we both agree, this is really a sub-optimal solution - obviously it would be ideal if we could remove that limitation on the Storage SDK.

isaacabraham:
> > This _could_ be simplified to just:
> > ```fsharp
> > table.Query<Customer> "PartitionKey eq 'Smith'"
> > ```
> > Because F# automatically outputs the following to the interactive console:
> 
> In your opinion, would it be more idiomatic to just assume the FSI behavior? There doesn't seem to be any trick or complexity being demonstrated to extract the results, so I don't think we'd lose anything important.

Agreed, I'll do that too.

IEvangelist:
Hi @isaacabraham - this looks great. When I look at the preview, I'm noticing that the corresponding article must have used range identifiers to pull in parts of the code from the file that you're proposing changes for. With that said, we'll have to also fix the article to adjust that.

For example:

image

image

image

One other concern that I have is the removal of async for an I/O bound code sample. While I agree that they example was incorrect before, as you mentioned they used `.Result` - it would be really nice if we could fix the async parts to make this truly asynchronous. Is that something that you're able to update?

IEvangelist:
Yes, there are live previews internal to MS - I'll review them myself if you update the MD. That would be perfect, and these are awesome changes. Thank you so much @isaacabraham.

jsquire:
> I'm happy to do that now if you're both happy with the code sample itself

Big thumbs-up from me.

samples/snippets/fsharp/azure/table-storage.fsx
cartermp: In genera it's idiomatic F# to do it this way. I wouldn't try to implement it at the class level and then have the interface just call that.

IEvangelist:
Hi @isaacabraham - just ping me when you have a chance to amend the MD line numbers. I'll take a look at the internal preview to ensure it aligns correctly. Thanks

jsquire:
> Nearly two months have passed since following up on this open PR @isaacabraham, any updates or plans? Or do you think we should close this?

@IEvangelist:  If Isaac isn't able to update the line numbers, I'd rather one of us took that on instead of closing it.   We've already got Isaac's expertise on the language captured, it would be a shame to waste it.

PR #28879, opened 2022-03-31
Set process return code

### LINKED ISSUE #28868: Sample shows incorrect means of setting exit code
This page shows a `Task<int>` method returning an exit code that is then dropped instead of actually being used to set the exit code for the process.

---
#### Document Details

* ID: 28554e6c-1983-ca7e-e9e0-200578729a28
* Version Independent ID: d8604fab-bf6f-e820-7db8-b22e3d724335
* Content: [How to handle termination in System.CommandLine](https://docs.microsoft.com/en-us/dotnet/standard/commandline/handle-termination)
* Content Source: [docs/standard/commandline/handle-termination.md](https://github.com/dotnet/docs/blob/main/docs/standard/commandline/handle-termination.md)
* GitHub Login: @tdykstra

tdykstra:
Fixes #28868

### DIFF
--- a/docs/standard/commandline/handle-termination.md
+++ b/docs/standard/commandline/handle-termination.md
@@ -19,6 +19,8 @@ To handle termination, inject a [CancellationToken](xref:System.Threading.Cancel
 
 Cancellation actions can also be added directly using the <xref:System.Threading.CancellationToken.Register%2A?displayProperty=nameWithType> method.
 
+For information about an alternative way to set the process exit code, see [Set exit codes](model-binding.md#set-exit-codes).
+
 ## See also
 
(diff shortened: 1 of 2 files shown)


gewarren (approved):

AArnott (commented):
That's one way to do it. But it's not the System.CommandLine-style way to do it, which is to set `InvocationContext.ExitCode` and then `return await rootCommand.InvokeAsync(args);` (I think).

tdykstra:
The options for setting exit code are discussed in https://docs.microsoft.com/en-us/dotnet/standard/commandline/model-binding#set-exit-codes

PR #28890, opened 2022-04-01
Add simplified null parameter checks

### LINKED ISSUE #28443: [17.1] Parameter null checking
This adds the `!!` syntax for the compiler to add a null check on that parameter. See dotnet/csharplang#2145 and the [feature spec](https://github.com/dotnet/csharplang/blob/main/proposals/param-nullchecking.md)

BillWagner:
Fixes #28443 

Add what's new information for the `!!` operator, available in VS 17.1, and .NET SDK 6.0.200 as preview features.


docs/csharp/language-reference/operators/null-parameter-check.md
```diff
@@ -0,0 +1,81 @@
...
+    }
+    // ...
+}
+```
+
+It's intended for library authors to provide runtime checks even when APIs have been annotated for nullable reference types. Some libraries support projects that aren't nullable enabled. Some projects support multiple versions including some that predate nullable reference types. These checks can simplify the necessary validation. You should only add the `!!` operator on those parameters that require a null-check for safety. In the preceding example, the `message` parameter doesn't have a runtime null check.
+
+There are a few rules that govern where you can add the `!!` operator on a parameter declaration:
+
+- The `!!` operator directs the compiler to add runtime behavior. It can't be applied to a declaration without an implementation. You can't add `!!` to the parameter of an abstract member, an interface method without an implementation, or a parameter on a delegate type declaration.
```
IEvangelist, line 63: > an interface method without an implementation

Does that include default interface implementations?

docs/csharp/language-reference/operators/null-parameter-check.md
```diff
@@ -0,0 +1,81 @@
...
+- The parameter type must be a type that can be compared to `null`. For example, the parameter can't be a non-nullable value type. If the parameter is a type parameter, it can't be constrained to be a non-nullable type (for example the `struct`, `notnull`, and `unmanaged` constraints are disallowed.)
+
+In almost all cases, the code the compiler generates for `!!` on an argument is consistent with the preceding example. However, the null check is inserted before any code you write in the method. The compiler generated null-check in constructors appears before the compiler generated calls to field initializers and before the call to any other constructor or base constructor. You can't catch the <xref:System.ArgumentNullException?displayProperty=nameWithType> thrown by the null check added by the compiler.
```
IEvangelist, line 66:
```suggestion
In almost all cases, the code the compiler generates for `!!` on an argument is consistent with the preceding example. However, the null check is inserted before any code you write in the method. The compiler generated null-check in constructors appears before the compiler generated calls to field initializers and before the call to any other constructor or base constructor.

> [!IMPORTANT]
> You can't catch the <xref:System.ArgumentNullException?displayProperty=nameWithType> thrown by the null check added by the compiler.
```

This needs to stand out.

IEvangelist (dismissed):
Looks awesome! Just have a few suggestions, otherwise :shipit:

docs/csharp/language-reference/operators/null-parameter-check.md
BillWagner: It doesn't. An interface method with a default implementation has an implementation. I may try to reword this. The point is if there's implementation to add the runtime check to, the compiler can't add the runtime check.

docs/csharp/language-reference/operators/null-parameter-check.md
BillWagner: Hmmm. I was almost tempted to delete this sentence. I think it would be writing nonsensical code.  Consider what the compiler generates:

```csharp
void M(string s)
{
    if (s is null)
    {
        throw new NullArgumentException(nameof(s));
    }
    // code....
}
```

Why would anyone write the following:

```csharp
void M(string s)
{
    try {
    if (s is null)
    {
        throw new NullArgumentException(nameof(s));
    }
    // code....
    } catch (NullArgumentException e)
    {
        // Why would you throw and catch in the same method?
    } 
}
```

docs/csharp/language-reference/operators/null-parameter-check.md
```diff
@@ -0,0 +1,81 @@
+---
+title: "!! (null validation) operator - C# reference"
+description: "Learn about simplified null-parameter checks. This operator instructs the compiler to add runtime checks that the argument used for a parameter isn't null."
+ms.date: 03/30/2022
+helpviewer_keywords:
```
Youssef1313, line 5: Can you add f1 keywords?

BillWagner: Yes! Where do I find the keys for those?

Youssef1313: @BillWagner They are not added to roslyn yet. So Roslyn can use any thing you choose. Though `!!_CSharpKeyword` will be consistent with the rest.

docs/csharp/language-reference/operators/null-parameter-check.md
```diff
@@ -0,0 +1,81 @@
...
+With null state analysis, any parameter annotated with `!!` has an initial null state of *not null*. The compiler has already added a null check. The compiler generates a warning when you write a method where a parameter is nullable and you've added a null check. Consider the following example:
+
+```csharp
+void Method(string? name!!)
+{
+    // ...
+}
+```
+
+The compiler issues the warning *"Nullable type 'string?' is null-checked and will throw if null."* The signature indicates that the method should accept `null` as the argument for `name`, and yet throws when that argument is supplied.
```
gfoidl, line 79: That's exactly my point from above. Here (`string?`) no null-check needed at runtime, as null is allowed.
With `string` as null-check should be provided, regardless nullable annotations enabled or not.

docs/csharp/language-reference/operators/null-parameter-check.md
```diff
@@ -0,0 +1,81 @@
...
+
+This feature is intended for library authors whose libraries will be used by code that may not enable nullable reference types. When all components are compiled in a nullable enabled context, the need for extra null parameter checks is reduced, and may not be needed at all. However, when consumers may not have enabled nullable reference types, that code may call methods with null arguments. You can add the runtime checks to improve the resilience of your code. The `!!` syntax makes those extra checks more concise.
```
gfoidl, line 81: Please drop this paragraph, due the aforementioned reasons.

Especially for libraries one never knows what inputs will arrive at runtime. So no matter if nullable annotations are enabled or not, argument validation has to be done for arguments that disallow null.

docs/csharp/whats-new/csharp-11.md
```diff
@@ -81,14 +82,40 @@ These types aren't directly represented in metadata. They include annotations th
...
 ## Newlines in string interpolations
 
-The text inside the `{` and `}` characters for a string interpolation can now span multiple lines. The text between the `{` and `}` markers is parsed as C#. Any legal C#, including newlines, is allowed. This can make it easier to read string interpolations that use longer C# expressions, like pattern matching `switch` expressions, or LINQ queries.
+The text inside the `{` and `}` characters for a string interpolation can now span multiple lines. The text between the `{` and `}` markers is parsed as C#. Any legal C#, including newlines, is allowed. This feature makes it easier to read string interpolations that use longer C# expressions, like pattern matching `switch` expressions, or LINQ queries.
 
 You can learn more about the newlines feature in the [string interpolations](../language-reference/tokens/interpolated.md) article in the language reference.
+
+## Simplified parameter null checks
+
+The `!!` operator provides null validation parameter syntax. Adding `!!` to a parameter declaration instructs the compiler to add a runtime check for that parameter. For example:
```
gfoidl, line 99: "parameter declaration": should it be "parameter name" instead?

For me the declaration is the whole `string name`, whilst the name is only `name` where the `!!` is actually applied.

I know this is pedantic, but it's the docs 😉

docs/csharp/whats-new/csharp-11.md
```diff
@@ -81,14 +82,40 @@ These types aren't directly represented in metadata. They include annotations th
...
+{
+    if (name is null)
+    {
+        throw new ArgumentNullException(nameof(name));
+    }
+    // ...
+}
+```
+
+This feature provides a concise syntax for runtime null parameter checking. It's intended for library authors to provide runtime checks even when APIs have been annotated for nullable reference types. Some libraries support projects that aren't nullable enabled. Some projects support multiple versions including some that predate nullable reference types. These checks can simplify the necessary validation. You can learn more in the language reference article on [null parameter checks](../language-reference/operators/null-parameter-check.md).
```
gfoidl, line 121:
```suggestion
This feature provides a concise syntax for runtime null parameter checking. It's intended for library authors to provide runtime checks even when APIs have been annotated for nullable reference types. These checks can simplify the necessary validation. You can learn more in the language reference article on [null parameter checks](../language-reference/operators/null-parameter-check.md).
```
Then this paragraph is 👍🏻

docs/csharp/language-reference/operators/null-parameter-check.md
```diff
@@ -0,0 +1,81 @@
+---
+title: "!! (null validation) operator - C# reference"
+description: "Learn about simplified null-parameter checks. This operator instructs the compiler to add runtime checks that the argument used for a parameter isn't null."
+ms.date: 03/30/2022
+helpviewer_keywords:
+  - "null-forgiving operator [C#]"
```
Youssef1313, line 6:
```suggestion
  - "null validation operator [C#]"
```

docs/csharp/language-reference/operators/null-parameter-check.md
gfoidl: ad 1)

I think it's a difference whether `NullReferenceException` (NRE) is thrown from anywhere down the stack, or if an `ArgumentNullException` (ANE) is thrown. The latter is part of the "API-contract" of that method. (Therefore ideally it should be documented (xml doc-comments) too).

NRE is kind of unexpected failure, that shouldn't happen. That's IMO why nullability annotations came to C#, so that the compiler can aid in detecting NREs by static analysis.

ANE on the other hand as part of the API-contract can happen. And they are not XOR to NRT, the work together very well.
Example:
```c#
public int Hash_A(string value!!)
{
    return 42;
}

public int Hash-B(string? value)
{
    if (value is null) return 0;
    return 42;
}
```
In `Hash_A` the type is `string` that doesn't allow null. But this "doesn't allow" is only for the static analyis in Roslyn, _not_ for the run-time / execution-time. So argument validation needs to be done -- in the example very handy with the new `!!`-operator.

For `Hash_B` the type is `string?` so for Roslyn's static analysis `null` is allowed, and as part of the API-contract it's also state that `null` can be passed in an no exception should occur in this case. The implementation can handle `null`.

---

ad 2)

> Does NRT do anything useful if I still need to check everything against null?

One only needs to check types that are annotated to not-allow-null (in public APIs).
NRT are still very useful (and I like them very much), as these annotation are part of the signature / API-contract.
So without looking into the docs or xml-doc-comments for a method I can see in the signature if the type allows null or not.
When the type disallows null, and I still pass a null in -- then it's my fault and an ANE should be thrown.
This is my bug, and very easy to track down to fix it.

If there wouldn't be a ANE thrown, rather anywhere and anytime later down the stack a NRE gets throws it's a nightmere to reproduce that bug and cover this by test, in order for a proper fix.
"Fail early" is very helpful.

> If a class library is only accessed from code known to have NRT enabled, is a null check still needed? 

Yes.
NRT is mostyl for static analysis, and being part of the API-contract.
For execution-time the null-check is needed.

I don't think the rules changed / respectively need to be changed.
With NRT we got another useful tool to express the intention: does a type allow to be null / disallows to be null
But this doesn't effect execution-time, there everything remains the same.

> If the only difference is where an exception is thrown, and what type of exception ...

These are big differences. In semantics of the exception, and in the handling of an exception.
The more precise (in terms of type and location) an exception is, the easier it's to deal with it.

Side note: the worst exception is GDI+ generic error
Nobody knows where it comes, and where it goes 😉 

---

So to sum up:
* Runtime exceptions are useful to handle that failure
* NRT are another useful tool to express intent as part of the API-contract
* at execution-time null-checks are still needed (by manual checking or by `!!`)
* NRT and null-checks play together -- they are friends, not opponents

docs/csharp/language-reference/operators/null-parameter-check.md
```diff
@@ -0,0 +1,101 @@
+---
+title: "!! (null validation) operator - C# reference"
+description: "Learn about simplified null-parameter checks. This operator instructs the compiler to add runtime checks that the argument used for a parameter isn't null."
+ms.date: 03/30/2022
+helpviewer_keywords:
```
Youssef1313, line 5: f1 not yet added.

docs/csharp/language-reference/operators/snippets/shared/u2ngxcbv.ld0~
```diff
@@ -0,0 +1,141 @@
+﻿using System;
```
Youssef1313, line 1: Is this file unintentionally checked in?

BillWagner: Good catch. No. It shouldn't be.

docs/csharp/language-reference/operators/null-parameter-check.md
BillWagner: I just handed push that update yet. It is now.

docs/csharp/language-reference/operators/null-parameter-check.md
```diff
@@ -0,0 +1,103 @@
+---
+title: "!! (null validation) operator - C# reference"
+description: "Learn about simplified null-parameter checks. This operator instructs the compiler to add runtime checks that the argument used for a parameter isn't null."
+ms.date: 03/30/2022
+f1_keywords: 
+  - "!!_CSharpKeyword"
```
Youssef1313, line 6: @RikkiGibson Now Roslyn can support `!!` in F1 :)

RikkiGibson: Oh, great. Do we need to wire anything up to make that happen?

Youssef1313: @RikkiGibson Likely yes we'll need to update [CSharpHelpContextService](https://github.com/dotnet/roslyn/blob/main/src/VisualStudio/CSharp/Impl/LanguageService/CSharpHelpContextService.cs). [Tests](https://github.com/dotnet/roslyn/blob/main/src/VisualStudio/CSharp/Test/F1Help/F1HelpTests.cs)

IEvangelist (approved):

gfoidl:
@BillWagner thanks for the updates! Now it reads very well and correct.

PR #28893, opened 2022-04-01
Update struct.md

chessboards:
Readonly static members are possible. I may have not exhausted the list of members excluded from being applicable to both readonly and static (those that can modify the structure instance), so further edits may be necessary.

## Summary

Fixes unclear and misleading information regarding the usage of static readonly members.

### DIFF
--- a/docs/csharp/language-reference/builtin-types/struct.md
+++ b/docs/csharp/language-reference/builtin-types/struct.md
@@ -67,7 +67,7 @@ Typically, you apply the `readonly` modifier to the following kinds of instance 
 
   :::code language="csharp" source="snippets/shared/StructType.cs" id="ReadonlyWithInit":::
 
-You can apply the `readonly` modifier to static members of a structure type excluding static members able to modify the structure instance.
+You can apply the `readonly` modifier to static fields of a structure type, but not any other static members, such as properties or methods.
 
 The compiler may make use of the `readonly` modifier for performance optimizations. For more information, see [Write safe and efficient C# code](../../write-safe-efficient-code.md).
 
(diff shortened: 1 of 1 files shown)


BillWagner:
Hi 👋 @chessboards 

That statement on disallowing `readonly` should stay. It's meant to cover this example:

```csharp
struct S
{
    // CS 8657: Can't apply "readonly" to a static method:
    public static readonly void M()
    {

    }
}
```

How did you interpret that statement? Would a different wording make that more clear?

chessboards:
I think the wording could be changed to make the distinction clearer between static members that should not interact with the structure instance or state (methods, events, props with setters), and those that have no state changing logic (therefore can be made readonly). The wording of `members` led me to include static fields and get-only properties as also being excluded, which they are not. https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/members

docs/csharp/language-reference/builtin-types/struct.md
```diff
@@ -67,7 +67,7 @@ Typically, you apply the `readonly` modifier to the following kinds of instance
 
   :::code language="csharp" source="snippets/shared/StructType.cs" id="ReadonlyWithInit":::
 
-You can't apply the `readonly` modifier to static members of a structure type.
+You can apply the `readonly` modifier to static members of a structure type excluding static members able to modify the structure instance.
```
BillWagner, line 70: Thanks for adding your comments @chessboards. That helps. I'd suggest updating this PR as follows:

```suggestion
You can apply the `readonly` modifier to static fields of a structure type, but not any other static members, such as properties or methods.
```

BillWagner (changes requested):
Thanks @chessboards 

I made a suggestion to this PR. Once you accept that, I'll :shipit:

BillWagner (approved):

BillWagner:
thanks again @chessboards 

I've set this to auto merge as soon as the builds finish.  You should see it live tomorrow, on our regular publishing cycle.

Thanks again!

PR #28943, opened 2022-04-07
Recommend using the 0x prefix in App.config

cshung:
## Summary

Update the GC Configuration page to recommend using 0x prefix in `App.config` to avoid a known unfortunate bug.

@dotnet/gc

### DIFF
--- a/docs/core/runtime-config/garbage-collector.md
+++ b/docs/core/runtime-config/garbage-collector.md
@@ -1,39 +1,35 @@
 ---
 title: Garbage collector config settings
 description: Learn about run-time settings for configuring how the garbage collector manages memory for .NET Core apps.
-ms.date: 07/10/2020
+ms.date: 04/20/2022
 ms.topic: reference
 ---
 # Runtime configuration options for garbage collection
 
-This page contains information about garbage collector (GC) settings that can be changed at run time. If you're trying to achieve peak performance of a running app, consider using these settings. However, the defaults provide optimum performance for most applications in typical situations.
+This page contains information about settings for the .NET runtime garbage collector (GC). If you're trying to achieve peak performance of a running app, consider using these settings. However, the defaults provide optimum performance for most applications in typical situations.
 
 Settings are arranged into groups on this page. The settings within each group are commonly used in conjunction with each other to achieve a specific result.
 
 > [!NOTE]
 >
-> - These settings can also be changed dynamically by the app as it's running, so any run-time settings you set may be overridden.
+> - These settings can also be changed dynamically by the app as it's running, so any configuration options you set may be overridden.
 > - Some settings, such as [latency level](../../standard/garbage-collection/latency.md), are typically set only through the API at design time. Such settings are omitted from this page.
 > - For number values, use decimal notation for settings in the *runtimeconfig.json* file and hexadecimal notation for environment variable settings. For hexadecimal values, you can specify them with or without the "0x" prefix.
 > - If you're using the environment variables, .NET 6 standardizes on the prefix `DOTNET_` instead of `COMPlus_`. However, the `COMPlus_` prefix will continue to work. If you're using a previous version of the .NET runtime, you should still use the `COMPlus_` prefix, for example, `COMPlus_gcServer`.
 
 ## Ways to specify the configuration
 
-For different versions of the the .NET runtime, there are different ways to specify the configuration values, here is a summary:
+For different versions of the the .NET runtime, there are different ways to specify the configuration values. The following table shows a summary.
 
-| config location      | .net versions this location applies to | formats  | how it's interpreted                                         |
+| Config location      | .NET versions this location applies to | Formats  | How it's interpreted                                         |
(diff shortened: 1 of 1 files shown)


docs/core/runtime-config/garbage-collector.md
```diff
@@ -15,6 +15,7 @@ Settings are arranged into groups on this page. The settings within each group a
 > - These settings can also be changed dynamically by the app as it's running, so any run-time settings you set may be overridden.
 > - Some settings, such as [latency level](../../standard/garbage-collection/latency.md), are typically set only through the API at design time. Such settings are omitted from this page.
 > - For number values, use decimal notation for settings in the *runtimeconfig.json* file and hexadecimal notation for environment variable settings. For hexadecimal values, you can specify them with or without the "0x" prefix.
```
gewarren, line 17:
```suggestion
> - For number values, use decimal notation for settings in the *runtimeconfig.json* file and hexadecimal notation for environment variable and *app.confg* settings. For hexadecimal values, you can specify them with or without the "0x" prefix. However, for *app.config* settings, we recommend always adding the "0x" prefix. The prefix ensures that the number is interpreted as hexadecimal and not decimal in all versions of .NET Framework.
```

docs/core/runtime-config/garbage-collector.md
```diff
@@ -15,6 +15,7 @@ Settings are arranged into groups on this page. The settings within each group a
...
+> - Due to an unfortunate bug, while numbers in the App.config without the 0x prefix is interpreted as decimal in .NET framework 4.7.2, numbers for GC config only are interpreted as hex in .NET framework 4.8 or above. To avoid issues, we recommend always specifying numbers as hex with the 0x prefix in App.config, that way it will work on any .NET framework versions.
```
gewarren, line 18:
```suggestion
```

gewarren (commented):
What do you think about combining it with the previous bullet point? It makes sense to me.

cshung:
Thanks for the rewrite, but I am afraid the rewrite is less clear though. 

If I were a customer trying to set the `GCHeapCount` the natural base to use is decimal and I would write 12 there. If I read your message, it isn't clear why I would want to change it to hex and add the 0x prefix. It is more obscure that way and doesn't make anything better. The one with decimal notation will work just fine (on 4.7.2.) and mostly fine on 4.8 (but got misinterpreted as 18, which I wouldn't notice until I really look into it in the debugger).

That's a trap that we want customers to avoid. The reason I added this note is to tell customers about this trap so they can avoid it.

I don't have a strong preference for emphasizing it is a bug, but the message needs to redirect the user to use the 0x hex notation even when it is unnatural to do so.

docs/core/runtime-config/garbage-collector.md
gewarren:
```suggestion
> - For number values, use decimal notation for settings in the *runtimeconfig.json* file and hexadecimal notation for environment variable and *app.config* settings. For hexadecimal values, you can specify them with or without the "0x" prefix. However, for *app.config* settings, we recommend always adding the "0x" prefix. In certain versions of .NET Framework, if you omit the prefix on a number value, it's interpreted as a decimal value.
```

gewarren:
@cshung I made a different suggestion. Is that more clear?

Maoni0:
so, I already heard feedback from customers that the description of how to specify configs sounded confusing. with this additional note it's even more info to take just to specify a config. I think if I were a customer I'd appreciate something less to read. how about changing the description to something like a table? please let me know if this seems clearer to both of you.



| config location    | .net versions this location applies to | formats  | how it's interpreted                                         |
| ------------------ | -------------------------------------- | -------- | ------------------------------------------------------------ |
| app.config         | .NET Framework                         | 0xn   | n is interpret as a hex value  (also see note below)                              |
| env var            | .NET Framework, .NET Core              | 0xn or n | n is interpret as a hex n is interpret as a hex value number in either format |
| runtimeconfig.json | .NET Core                              | n        | n is interpreted as a decimal value.                         |

Note that you could specify a value with the 0x prefix for app.config but we do not recommend that. Due to an unfortunate bug, on .NET Framework 4.8+ it's interpreted as hex but on previous versions of .NET Framework it's interpreted as decimal. To avoid having to change your config please always use the 0x prefix when specifying a value in your app.config.

An example, if I want to specify 12 heaps for GCHeapCount, this is how I would specify it for an executable named a.exe [give the example how to specify that in all 3 ways]

gewarren:
I think it's a great idea @Maoni0!

Maoni0:
looks good. my text about the example sounds too casual for official doc - @gewarren could you please help us make it more suitable for official documentation?

I'd be totally fine if we left this in the GC doc page just because some of this is specific to GC (eg the 4.8 bug). and it's nice to have just one page to go to.

gewarren:
How does this look? [Preview](https://review.docs.microsoft.com/en-us/dotnet/core/runtime-config/garbage-collector?branch=pr-en-us-28943#ways-to-specify-the-configuration)

Maoni0 (approved):

PR #28952, opened 2022-04-08
Document new method group conversion

### LINKED ISSUE #28942: [17.2] Allow cached delegate for method groups
This has a small scope, and the spec is already updated. C# compilers are permitted to cache the delegate instance for a static method.  See dotnet/roslyn#5835. This has minimal semantic impact on developers.

BillWagner:
Fixes #28942

Beginning in C# 11, the compiler is allowed to cache the delegate created from a method group conversion. The compiler now does this.


CamSoper (approved):
LGTM

PR #28974, opened 2022-04-11
Commit to Add download alert - Blazor for ASP NET Web Forms Developers

erjain:
Add alerts to the Blazor-for-ASP-NET-Web-Forms-Developers eBook, explaining that each article is but an excerpt from the whole book. And provides an actionable link and thumbnail of the book.


nishanil:
@erjain Can you revert the author changes? We don't need to modify that.

IEvangelist (approved):

PR #28986, opened 2022-04-12
Fix fish tab completion command

MatisseHack:
## Summary

The current fish shell completions command doesn't work for me. It calls `dotnet complete` with no arguments, which does make suggestions, but the suggestions are not based on what I've already typed. Adding `commandline -cp` passes the currently typed command to improve suggestion relevancy.


tdykstra (approved):
Thanks @MatisseHack !

PR #28991, opened 2022-04-12
move comment to correct block

omarjuul:
## Summary

If I understand the explanation correctly, the third (and fourth) code block uses the feature, not the second.
This tiny PR fixes this error.


BillWagner (approved):
Thanks for spotting this and submitting a fix @omarjuul We appreciate it!

PR #28999, opened 2022-04-12
Fix table cell entry for anonymous records

TrickyCat:
At the moment table cell entry with a code snippet for anonymous records is being rendered in HTML on the Microsoft Docs site as:

`{\|\|}` and not like `{||}`

That's because the current Markdown engine considers pipes even inside the backticked blocks as column delimiters while rendering a table :pensive:.

## Summary

Use HTML `<code>...</code>` block instead.


tdykstra (approved):
Thanks for the correction!

PR #29006, opened 2022-04-12
Added version info for libmsquic

samsp-msft:
Incompatibility needed to be noted.


IEvangelist (approved):

PR #29055, opened 2022-04-13
Update xml-documentation.md

matthewcrews:
## Summary

Fixes a small grammar error

Fixes: N/A


gewarren (commented):
I think the original wording was more correct. Compare it to line 64, for example.

PR #29081, opened 2022-04-15
Note alternative organization option

### LINKED ISSUE #26395: Reason for separating the business logic and tests
In this example you separate the business logic (NewTypes.csproj) from the test cases.

What is the reason for that separation?

The only thing that I could come up with is that it might be easier to identify all test projects in the pipeline when you want to run the tests. But running `dotnet test` on the solution will automatically run all test projects in the solution, ignoring the business logic projects.

There are some downsides to this separation, especially if you use the Visual Studio IDE:

- If you want to use these projects in a Visual Studio Solution, you'd have to put the solution file in the root folder, even though I feel that the solution is part of the source.
- Adding a project in your solution in Visual Studio (even if you right-click the solution folder called `test` or `src`) will add the project to the root folder, not the `test` or `src` folder. You need to manually move it on disk, including editing the `.sln` file, or removing and re-adding the project from/to the solution.
- I'm a big fan of keeping code closely together if you edit that code at the same time: if you change a project, you're very likely to also change the test code for that same project.

---
#### Document Details

* ID: f9b54562-9072-a864-be3d-44ed7965c798
* Version Independent ID: a69ec985-ed05-90dd-a686-dc956b0cf5df
* Content: [Organizing and testing projects with the .NET CLI - .NET](https://docs.microsoft.com/en-us/dotnet/core/tutorials/testing-with-cli#feedback)
* Content Source: [docs/core/tutorials/testing-with-cli.md](https://github.com/dotnet/docs/blob/main/docs/core/tutorials/testing-with-cli.md)
* GitHub Login: @cartermp

tdykstra:
Fixes #26395

### DIFF
--- a/docs/core/tutorials/testing-with-cli.md
+++ b/docs/core/tutorials/testing-with-cli.md
@@ -1,16 +1,19 @@
 ---
 title: Organizing and testing projects with the .NET CLI
 description: This tutorial explains how to organize and test .NET projects from the command line.
-ms.date: 09/10/2018
+ms.date: 04/15/2022
 ---
 
 # Organizing and testing projects with the .NET CLI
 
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @tdykstra 

You can :shipit: when ready.

PR #29091, opened 2022-04-16
Allow demo code to compile

pbachmann:
Function to should return Error not Exception.

## Summary

Fix demo code that has not compiled since it was published five and a half years ago.


IEvangelist (approved):

PR #29112, opened 2022-04-19
Update High CPU Tutorial

mikelle-rogers:
Initial changes to clarify tutorial

## Summary

Update tutorial for clarity

Fixes [dotnet/diagnostics/2901](https://github.com/dotnet/diagnostics/issues/2901)


docs/core/diagnostics/debug-highcpu.md
```diff
@@ -83,10 +83,10 @@ With the web app running, immediately after startup, the CPU isn't being consume
 
 `https://localhost:5001/api/diagscenario/highcpu/60000`
 
-Now, rerun the [dotnet-counters](dotnet-counters.md) command. To monitor just the `cpu-usage`, specify `System.Runtime[cpu-usage]` as part of the command.
+Now, rerun the [dotnet-counters](dotnet-counters.md) command. To monitor just the `cpu-usage`, add --counters `System.Runtime[cpu-usage]` as part of the command.
```
gfoidl, line 86:
```suggestion
Now, rerun the [dotnet-counters](dotnet-counters.md) command. To monitor just the `cpu-usage`, add `--counters System.Runtime[cpu-usage]` as part of the command.
```

docs/core/diagnostics/debug-highcpu.md
```diff
@@ -83,10 +83,10 @@ With the web app running, immediately after startup, the CPU isn't being consume
...
 
 ```dotnetcli
-dotnet-counters monitor --counters System.Runtime[cpu-usage] -p 22884 --refresh-interval 1
+dotnet-counters monitor -p 22884 --refresh-interval 1
```
gfoidl, line 89:
```suggestion
dotnet-counters monitor -p 22884 --refresh-interval 1 --counters System.Runtime[cpu-usage]
```
Otherwise it's the same command as before. And as the text states "add" put it at the end -- not in the middle as the previous code was (left side of diff).

docs/core/diagnostics/debug-highcpu.md
```diff
@@ -96,17 +96,35 @@ Press p to pause, r to resume, q to quit.
     Status: Running
 
 [System.Runtime]
+    % Time in GC since last GC (%)                         0
```
gfoidl, line 99: This does not match the text. Only cpu usage should be shown here.

docs/core/diagnostics/debug-highcpu.md
```diff
@@ -96,17 +96,35 @@ Press p to pause, r to resume, q to quit.
...
+    Working Set (MB)                                      63
 ```
 
 Throughout the duration of the request, the CPU usage will hover around 25% . Depending on the host machine, expect varying CPU usage.
 
 > [!TIP]
 > To visualize an even higher CPU usage, you can exercise this endpoint in multiple browser tabs simultaneously.
 
-At this point, you can safely say the CPU is running higher than you expect.
+At this point, you can safely say the CPU is running higher than you expect, meaning the app has a slow request.
```
gfoidl, line 125: > meaning the app has a slow request.

There can be a myriad of reasons for higher cpu. Maybe write it as something like
```suggestion
At this point, you can safely say the CPU is running higher than you expect. With the knowledge of the app-code, and the requests made this means the app has a slow request. 
```
?

docs/core/diagnostics/debug-highcpu.md
```diff
@@ -83,32 +83,50 @@ With the web app running, immediately after startup, the CPU isn't being consume
...
-At this point, you can safely say the CPU is running higher than you expect.
+At this point, you can safely say the CPU is running higher than you expect. Identifying the effects of a problem is key to finding the cause. We will use the effect of high CPU consumption in addition to diagnostic tools to find the cause of the problem.
 
-## Trace generation
+## Analyze High CPU with Profiler
 
-When analyzing a slow request, you need a diagnostics tool that can provide insights into what the code is doing. The usual choice is a profiler, and there are different profiler options to choose from.
+When analyzing an app with high CPU usage, you need a diagnostics tool that can provide insights into what the code is doing. The usual choice is a profiler, and there are different profiler options to choose from.
 
 ### [Linux](#tab/linux)
```
tommcdon, line 132: Should we explain why we suggest using the Linux perf tool and not dotnet-trace?  I'm guessing the reason is something like: Although Dotnet-Trace is cross-platform but since Event Pipe "CPU" traces only measure elapsed time and not actual CPU time.

tommcdon (approved):
Left a comment for your consideration

docs/core/diagnostics/debug-highcpu.md
mikelle-rogers: I had the same question. @josalem Why are we using the Linux perf tool and not dotnet-trace?

josalem: `dotnet-trace` is great for CPU investigations to a certain point. The limitations of safe-point bias and managed-only callstacks means that it can't give _exact_ information the same way a kernel-aware profiler like `perf` can.

If your performance investigation stops at managed code, generally `dotnet-trace` will be sufficient. If you are analyzing exact timing of your application and it involves native code, you may need something like `perf`. 

I think of it like two differently powered magnifying glasses. You can start with `dotnet-trace` and then look at a higher resolution using `perf` (or `ETW` on Windows).

PR #29114, opened 2022-04-19
Fix build suggestions

gewarren:
Contributes to #28815


BillWagner (approved):
This looks great @gewarren 

You can :shipit: when ready.

PR #29135, opened 2022-04-21
Fix a typo

sixtymage:
## Summary

Fix a grammar typo in the explanation of Type patterns in the **Pattern matching enhancements** section of the _What's new in C# 9.0_ article.

### DIFF
--- a/docs/csharp/whats-new/csharp-9.md
+++ b/docs/csharp/whats-new/csharp-9.md
@@ -197,7 +197,7 @@ For more information, see [Top-level statements](../fundamentals/program-structu
 
 C# 9 includes new pattern matching improvements:
 
-- ***Type patterns*** match a variable is a type
+- ***Type patterns*** match a variable that is a type
 - ***Parenthesized patterns*** enforce or emphasize the precedence of pattern combinations
 - ***Conjunctive `and` patterns*** require both patterns to match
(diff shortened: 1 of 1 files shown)


docs/csharp/whats-new/csharp-9.md
```diff
@@ -197,7 +197,7 @@ For more information, see [Top-level statements](../fundamentals/program-structu
 
 C# 9 includes new pattern matching improvements:
 
-- ***Type patterns*** match a variable is a type
+- ***Type patterns*** match a variable that is a particular type
```
BillWagner, line 200: I think this could be misleading. The variable already has a type. The ***Type pattern*** checks that the object a variable refers to matches a type. How about this:

```suggestion
- ***Type patterns*** match an object matches a particular type
```

BillWagner (commented):
Hi :wave: @sixtymage 

I agree that this should change. I did suggest a different wording. What do you think?

BillWagner (approved):

sixtymage:
> Hi 👋 @sixtymage
> 
> I agree that this should change. I did suggest a different wording. What do you think?

Hey Bill. Sorry, been away for a few days so I couldn't reply. I really like your latest version. Nice one!

PR #29141, opened 2022-04-21
Acrolinx 4/21

gewarren:


docs/azure/sdk/resource-management.md
```diff
@@ -170,7 +170,7 @@ Console.WriteLine($"Vnet: {id.Parent.Name}");
 Console.WriteLine($"Subnet: {id.Name}");
 ```
 
-However, keep in mind that some of those properties could be null. You can usually tell by the id string itself which type a resource ID is, but if you are unsure, check if the properties are null or use the Try methods to retrieve the values as it's shown below:
+However, keep in mind that some of those properties could be null. You can usually tell by the ID string itself which type a resource ID is. But if you're unsure, check if the properties are null, or use the `Try` methods to retrieve the values as is shown in the following example.
```
tdykstra, line 173:
```suggestion
However, keep in mind that some of those properties could be null. You can usually tell by the ID string itself which type a resource ID is. But if you're unsure, check if the properties are null, or use the `Try` methods to retrieve the values, as shown in the following example.
```

docs/azure/sdk/resource-management.md
```diff
@@ -218,7 +218,7 @@ ResourceGroup resourceGroup = await subscription.GetResourceGroups().GetAsync(id
 AvailabilitySet availabilitySet = await resourceGroup.GetAvailabilitySets().GetAsync(id.Name);
 ```
 
-This approach required a lot of code and 3 API calls to Azure. The same can be done with less code and without any API calls by using extension methods that we have provided on the client itself. These extension methods allow you to pass in a resource identifier and retrieve a scoped resource client. The object returned is a *[Resource]* mentioned above, since it has not reached out to Azure to retrieve the data yet the Data property will be null.
+This approach required a lot of code and three API calls to Azure. The same can be done with less code and without any API calls by using extension methods that we've provided on the client itself. These extension methods allow you to pass in a resource identifier and retrieve a scoped resource client. The object returned is a *[Resource]* mentioned previously. Since it hasn't reached out to Azure to retrieve the data yet, the `Data` property will be null.
```
tdykstra, line 221: ? a *[Resource]* mentioned previously

It's unclear which previous reference to [Resource] this refers to, as there are many of them.

tdykstra (approved):
LGTM, a couple of comments.

PR #29158, opened 2022-04-23
Remove repeating word in dotnet-install script doc

N0D4N:
## Summary

Remove repeating "since" word in dotnet-install script docs


tdykstra (approved):
Thanks for the correction @N0D4N

PR #29170, opened 2022-04-25
Fixed Compiler Error CS1009 in sample

stesee:
## Summary

The sample code was not valid. A single '\\' needs to be escaped. I changed it to a '/' thats for every os valid.


gewarren (approved):
Thank you @stesee!

PR #29227, opened 2022-04-29
Clarify the minimum VS/MSBuild version for .NET SDK 6.0.300

baronfel:
We got some user feedback over email that it wasn't apparent if 6.0.300 was an aberrant version in some way, or if this change was a new minimum version bound. This change should reiterate that it's a new minimum VS Version Bound to use the 6.0.300+ SDK.

### DIFF
--- a/docs/core/compatibility/sdk/6.0/vs-msbuild-version.md
+++ b/docs/core/compatibility/sdk/6.0/vs-msbuild-version.md
@@ -5,7 +5,7 @@ ms.date: 03/25/2022
 ---
 # Version requirements for .NET 6 SDK
 
-Starting with the .NET 6.0.300, the .NET SDK no longer loads in version 16.11 or earlier of Visual Studio or MSBuild.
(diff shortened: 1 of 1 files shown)


docs/core/compatibility/sdk/6.0/vs-msbuild-version.md
```diff
@@ -5,19 +5,19 @@ ms.date: 03/25/2022
 ---
 # Version requirements for .NET 6 SDK
 
-The NET 6.0.300 SDK no longer loads in version 16.11 or earlier of Visual Studio or MSBuild.
+Starting with the .NET 6.0.300, the .NET SDK no longer loads in version 16.11 or earlier of Visual Studio or MSBuild.
```
baronfel, line 8:
```suggestion
Starting with the .NET SDK 6.0.300, the .NET SDK no longer loads in version 16.11 or earlier of Visual Studio or MSBuild.
```

gewarren (approved):
Thank you!

PR #29240, opened 2022-05-02
Fix typo in record.md

ericmutta:
The following sentence:

> The `record` type were introduced in C# 9;

Should most likely read:

> The `record` type **was** introduced in C# 9;

It could also read as follows:

> The `record` **types** were introduced in C# 9;

...but as far as I can tell from the [C# 9 docs](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#record-types) there was only one `record` type back then (what are now called `record class` types) so we can't use the plural "types" in the sentence I edited.


BillWagner (approved):
Thanks for submitting this fix @ericmutta 

ericmutta:
Many thanks @BillWagner for the prompt follow-up 👍

PR #29244, opened 2022-05-02
Clarify where events can be raised from

### LINKED ISSUE #29236: Fix inconsistent description of where events can be invoked from
I noticed that in this article it states that "Events are a special kind of multicast delegate that can only be invoked from within the class or struct where they are declared (the publisher class)." But in the comments of the previous code snippet, it states "Wrap the event in a protected virtual method to enable derived classes to raise the event."

I think it would be clearer if the first sentence stated "from within the class (or derived classes) or struct", because when I first read it I thought that events could only be invoked from the publisher class, and never in their derived classes.

---
#### Document Details

* ID: f782a9c5-2dc8-d8d1-b17d-246ab8524863
* Version Independent ID: f6f825c5-c64b-d13b-1c7e-a01bfb93093d
* Content: [event - C# Reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/event?source=docs)
* Content Source: [docs/csharp/language-reference/keywords/event.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/keywords/event.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

Julian-C10:
## Summary

Making it clear that events can be raised from the publisher class _and_ classes derived from the publisher class.

Fixes #29236

### DIFF
--- a/docs/csharp/language-reference/keywords/event.md
+++ b/docs/csharp/language-reference/keywords/event.md
@@ -21,7 +21,7 @@ The following example shows how to declare and raise an event that uses <xref:Sy
 
 [!code-csharp[csrefKeywordsModifiers#7](~/samples/snippets/csharp/VS_Snippets_VBCSharp/csrefKeywordsModifiers/CS/csrefKeywordsModifiers.cs#7)]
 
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks for submitting this fix @Julian-C10 

I'll :shipit: now.

Thanks again!

Julian-C10:
Great! Thank you @BillWagner

PR #29284, opened 2022-05-05
Fix grammar typo

GitHubPang:
Use "an" instead of "a" for `InvalidEnumArgumentException`.


gewarren (approved):
Thank you @GitHubPang

PR #29289, opened 2022-05-05
Emphasize that `SCH_USE_STRONG_CRYPTO` is used only for clients

### LINKED ISSUE #61739: .NET 4.8 - TLS / SCHANNEL / UseStrongCrypto
### Description

According to the documentation, to use "strong" TLS crypto ciphers on .NET 4.7 or newer you need do nothing (i.e. the default settings are sensible), see: https://docs.microsoft.com/en-us/dotnet/framework/network-programming/tls

My application targets .NET 4.8 and runs on Windows Server 2016.

If I understand correctly, as long as I pass SslProtocols.None to AuthenticateAsClientAsync(String, X509CertificateCollection, SslProtocols, Boolean) then the RC4 should be filtered out and disabled, see: https://docs.microsoft.com/en-us/windows/win32/secauthn/tls-cipher-suites-in-windows-10-v1607

This is because, by default "DontEnableSchUseStrongCrypto" defaults to "false" (as "don't enable strong crypto" is false, .NET does enable), and hence SCH_USE_STRONG_CRYPTO is true.

The problem I have is when I act as a server (and use AuthenticateAsServerAsync).

After digging in to the .NET 4.8 Reference Source, it seems that the flag "UseStrongCrypto" is only ever passed to "SecureCredential" when acting as a client (in AcquireClientCredentials), and the server equivalent code path (in AcquireServerCredentials) doesn't set the flag.

Is this correct? i.e. "strong" TLS crypto ciphers are only enforced automatically when acting as a client and not when acting as a server? If so, the documentation doesn't make this clear.

Or have I missed something? My application is running on a fully patch Windows Server 2016, targeting .NET 4.8 Framework, but when I scan the port it lists RC4 as being allowed. Everything I've read from the documentation implies this should not be the case.

Any pointers/help is much appreciated.

### Reproduction Steps

Act as a TLS server using .NET 4.8 on Windows Server 2016

### Expected behavior

"Strong" crypto is enabled and hence RC4 ciphers are disabled

### Actual behavior

RC4 ciphers are still enabled

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

Windows Server 2016
.NET Framework 4.8

### Other information

_No response_

rzikm:
Closes https://github.com/dotnet/runtime/issues/61739

### DIFF
--- a/docs/framework/network-programming/tls.md
+++ b/docs/framework/network-programming/tls.md
@@ -146,7 +146,7 @@ The switches have the same effect whether you're doing HTTP networking (<xref:Sy
 
 ### Switch.System.Net.DontEnableSchUseStrongCrypto
 
-A value of `false` for `Switch.System.Net.DontEnableSchUseStrongCrypto` causes your app to use strong cryptography. A value of `false` for `DontEnableSchUseStrongCrypto` uses more secure network protocols (TLS 1.2, TLS 1.1, and TLS 1.0) and blocks protocols that are not secure. For more info, see [The SCH_USE_STRONG_CRYPTO flag](#the-sch_use_strong_crypto-flag). A value of `true` disables strong cryptography for your app.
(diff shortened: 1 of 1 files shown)


wfurt (approved):
LGTM> All the pages talk about .NET Framework, right? (e.g. no need to mention .NET 7)

rzikm:
Yes, the page is only for .NET Framework

PR #29302, opened 2022-05-05
CLI reference article for dotnet dev-certs

### LINKED ISSUE #26147: dotnet dev-certs Documentation
**Issue description**
`dotnet dev-certs` has very minimal documentation. Apart from the `--help` text that comes up in the CLI tool itself, and a bit in [Generate self-signed certificates with the .NET CLI](https://docs.microsoft.com/en-us/dotnet/core/additional-tools/self-signed-certificates-guide), there is actually very little in the way of documentation of the actual behaviors of all the various flags.

My team and I identified some behaviors that we cannot find documented _anywhere_, which I will list below. Apologies in advance for any formatting errors, I pulled this out of a Slack thread. I'll go back through and edit after I post if I missed anything.

* It does not generate a physical certificate file by default on Windows, but does store the certificate in the user's certificate store.
* If you pass `-ep /path/to/file.cert` on Windows, it will output a physical file as well as adding the certificate to the user's store.
* On Windows, not passing `--trust` still creates the certificate in the user's store, but does not, indeed, add the certificate to a trusted list.
* On Linux, .NET stores the cert in `~/.dotnet/corefx/cryptography/x509stores/my/`, which accurately mimics the path of the certificate store on Windows and OSX (basically, it mocks the `X509CertStore2.StoreName.My` and `X509CertStore2.StoreLocation.CurrentUser` pattern you see in .NET and Windows).
* When you pass an export path on Linux, it will both export the file to the path you give it (including your filename), and it will also put a PFX file (with the filename helpfully set to the fingerprint of the cert) in the default directory. This behavior is functionally identical to Windows and OSX from the perspective of .NET application code.
* Linux .NET doesn't like Windows-generated PFX files, but can load Windows-generated PEM files just fine.
* .NET creates a PEM-encoded cert if you pass `--format Pem`. However, this will be a "password protected" PEM by default, with no key file. It's unclear what the password is; potentially it's an empty string. More testing needs to be done. *EDIT*: I have done some testing here, and I actually can't get this to work in code at all. The only way to get it to work is to provide a password via `-p`, and use the overload of `CreateFromEncryptedPemFile` which includes the cert file path, the password, and the key file path. Using `--format Pem` does not produce a PEM that can be loaded by .NET's `CreateFromEncryptedPemFile`. It seems like the overload for the file path and password (no key file path) is for a type of PEM that `dev-certs` cannot create (that is, one with the key encrypted in the same file as the certificate PEM).
    * This specific behavior should be noted, since there shouldn't be an output of `dotnet dev-certs` that cannot be loaded by a `dotnet` program. This is actually an issue with the dev-certs tool itself, not necessarily with documentation.
* If you want to generate a PEM with separate cert and key files, you have to pass `-np` (no password).
* You always have to pass an export path (`-ep`) if you want to get a PEM file.
* The result of the previous two points is that you will get a file named `<yourcertname>.<yourcertextension>` and also a file named `<yourcertname>.key` file in the directory you pass as part of the export path. For example, the following will generate a file named `localhost.pem` and a file named `localhost.key` in the `/home/user` directory: `dotnet dev-certs --format Pem -ep /home/user/localhost.pem -np`
* `--clean` does _not_ get rid of any physical files which have been exported. It only clears the certificate store of the generated certificate. This is even the case if you have exported your certificate to the (Linux) certificate store directory. 

_________

Importantly, some of this behavior is an active blocker when working with .NET and Docker. We've found that [this guide](https://docs.microsoft.com/en-us/aspnet/core/security/docker-https?view=aspnetcore-5.0) does not work at all in our environment. We're still working through some of the pain points here, and might well be holding the tool wrong, so to speak.

However, given that all of the above points had to be sussed out by actively running the command repeatedly under many different circumstances has made troubleshooting the base problem (using dev certs in Docker) difficult, to say the least.

Normally, I would try to open a PR with this info included, but there are many different behaviors here that aren't necessarily related. As written, these discovered behaviors would make a good blog post, but I don't know where they should go in official .NET documentation. Hence, this issue! I'd be happy to submit some PRs after getting some guidance from the core team as to where this sort of docs should live.
_________

**Target framework**

- [x] .NET Core
- [ ] .NET Framework
- [ ] .NET Standard

<details>
<summary><strong>dotnet --info output</strong> or <strong>About VS info</strong></summary>

```console
$ dotnet --info
.NET SDK (reflecting any global.json):
 Version:   5.0.206
 Commit:    0cf287db1b

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19043
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\5.0.206\

Host (useful for support):
  Version: 5.0.9
  Commit:  208e377a53

.NET SDKs installed:
  5.0.206 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 5.0.9 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.18 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.9 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.1.18 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.9 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```
</details>

---
#### Document Details

* ID: b7e47565-cc59-b1a6-24be-0143c550c0b2
* Version Independent ID: c6859ca6-e2b1-3645-872c-60390ffd429a
* Content: [.NET CLI](https://docs.microsoft.com/en-us/dotnet/core/tools/)
* Content Source: [docs/core/tools/index.md](https://github.com/dotnet/docs/blob/main/docs/core/tools/index.md)
* Technology: **dotnet-cli**
* GitHub Login: @tdykstra

### LINKED ISSUE #27236: Page should explain how it works
Hello, I'm trying to understand how to setup a new Macbook with a self-signed certificate, where my website is running in a container.

While on this journey I'm finding some things confusing. After coding in some form for 35 years now, when I find the going tough or confusing, I think it's okay to blame my tools and documentation.

These commands here are fishy. The first one specifies a file but the second one does not. How does the second one know which certificate to trust? Does the first one have side effects that aren't explicit or explained?

```
dotnet dev-certs https -ep $env:USERPROFILE\.aspnet\https\aspnetapp.pfx -p crypticpassword
dotnet dev-certs https --trust
```

So I turn to the build-in help documentation for that command. Alas, it says nothing that might explain the above, leaving me lost. There's an information gap here. It says "Trust the certificate ..." but which certificate?

I don't want to use commands that may have side-effects I am unaware of because I won't discover exactly what made something work, and I won't be able to automate the steps to setup the correct state in a different environment. 

```
Usage: dotnet dev-certs https [options]

Options:
  -ep|--export-path  Full path to the exported certificate
  -p|--password      Password to use when exporting the certificate with the private key into a pfx file or to encrypt the Pem exported key
  -np|--no-password  Explicitly request that you don't use a password for the key when exporting a certificate to a PEM format
  -c|--check         Check for the existence of the certificate but do not perform any action
  --clean            Cleans all HTTPS development certificates from the machine.
  -i|--import        Imports the provided HTTPS development certificate into the machine. All other HTTPS developer certificates will be cleared out
  --format           Export the certificate in the given format. Valid values are Pfx and Pem. Pfx is the default.
  -t|--trust         Trust the certificate on the current platform. When combined with the --check option, validates that the certificate is trusted.
  -v|--verbose       Display more debug information.
  -q|--quiet         Display warnings and errors only.
  -h|--help          Show help information
```

There's not even a description for what `dev-certs` does. But wait, there's another fishy command later in the page.

```
dotnet user-secrets remove "Kestrel:Certificates:Development:Password" -p aspnetapp\aspnetapp.csproj
dotnet dev-certs https --clean
```

Clean? What's it cleaning? We just created a file. Does it delete that file? Or was there a side effect when that original command was executed that now needs cleaning? Nothing is explained.

Next is the part where a user secret is setup. It seems to be writing a setting for Kestrel to read, but how does Kestrel know it needs to open that PFX file we just created?

Since when did documentation become a murder mystery.  I don't understand why Microsoft's documentation no longer explains how the tricks are done. Why aren't there links out to the documentation about the mechanisms being employed here?

I have similar feedback here: https://github.com/MicrosoftDocs/visualstudio-docs/issues/7337

Thanks, Luke

---
#### Document Details

* ID: 129ac9c9-bab2-7ae0-75f8-cf3b7ad16a56
* Version Independent ID: 19debb34-1672-c5bc-395b-2ec523c25b8f
* Content: [Generate Self-Signed Certificates Overview - .NET](https://docs.microsoft.com/en-us/dotnet/core/additional-tools/self-signed-certificates-guide?source=docs)
* Content Source: [docs/core/additional-tools/self-signed-certificates-guide.md](https://github.com/dotnet/docs/blob/main/docs/core/additional-tools/self-signed-certificates-guide.md)
* GitHub Login: @angee

tdykstra:
Fixes #26147
Fixes #27236


docs/core/tools/dotnet-dev-certs.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Options
+
+- **`-c|--check`**
+
+  Checks for the existence of the development certificate but doesn't perform any action. Use this option with the `--trust` option to check if the certificate is not only valid but also trusted.
+
+- **`--clean`**
+
+  Removes all HTTPS development certificates from the machine. On Windows, it doesn't get rid of any physical files that have been exported. It only clears the certificate store of the generated certificate.
```
javiercn, line 63: Removes all HTTPS development certificates from the certificate store.

> On Windows, it doesn't get rid of any physical files that have been exported

On Windows, it doesn't get rid of any physical files that have been exported. It only clears the certificate store of the generated certificate.

This is confusing, I would omit any reference to exported files because these two behaviors are orthogonal.

javiercn: Exporting the cert is meant to allow it to be consumed by other tools.

docs/core/tools/dotnet-dev-certs.md
```diff
@@ -0,0 +1,186 @@
...
+  Cleaning HTTPS development certificates
+  from the machine.
+  A prompt might get displayed to confirm
+  the removal of some of the certificates.
+
+  HTTPS development certificates
+  successfully removed from the machine.
+  ```
+
+- **`-ep|--export-path`**
```
javiercn, line 77: Export path, password, no-password and format work in tandem

docs/core/tools/dotnet-dev-certs.md
```diff
@@ -0,0 +1,186 @@
...
+  [--format] [-i|--import] [-np|--no-password]
+  [-p|--password] [-q|--quiet] [-t|--trust]
+  [-v|--verbose] [--version]
+
+dotnet dev-certs https -h|--help
+```
+
+## Description
+
+The `dotnet dev-certs` command generates a self-signed certificate to enable HTTPS use in local web app development. The command recognizes options that perform related functions, such as checking if a certificate is already present, and importing and exporting certificates from and to certificate files.
```
javiercn, line 28: The `dotnet dev-certs` command manages a self-signed certificate to enable HTTPS use in local web app development. Its main functions are:
* Generating a certificate
* Trusting a certificate
* Exporting a certificate in various formats so that it can be consumed by other tools.
* Removing the generated certificate from the machine.
* Importing an existing certificate generated by the tool into the machine.

blowdart (approved):

javiercn (approved):

PR #29303, opened 2022-05-05
Fix required TFM for Source Generator runtime

ahmad2smile:
## Summary

For Source Generator to work in Visual Studio TFM has to be targeted to `.netstandard 2.0` but for `dotnet` CLI it can be targeted to any TFM.

Changes:

Corrected misleading TIP regarding Source Generator not working with higher TFM than `.netstandard 2.0`

### DIFF
--- a/docs/csharp/roslyn-sdk/source-generators-overview.md
+++ b/docs/csharp/roslyn-sdk/source-generators-overview.md
@@ -71,7 +71,7 @@ In this guide, you'll explore the creation of a source generator using the <xref
     :::code language="xml" source="snippets/source-generators/SourceGenerator/SourceGenerator.csproj":::
 
     > [!TIP]
-    > The source generator project needs to target the `netstandard2.0` TFM, otherwise it will not work.
+    > The source generator project needs to target the `netstandard2.0` TFM for Visual Studio, for `dotnet` CLI with relative SDK it can work for any TFM.
 
 1. Create a new C# file named _HelloSourceGenerator.cs_ that specifies your own Source Generator like so:
 


BillWagner (approved):
Thanks for updating this @ahmad2smile 

sharwell (changes requested):
Last time I checked, **netstandard2.0** is the only scenario supported by the compiler. Situations where other targets work is merely coincidental, and should not be encouraged without explicit approval from @dotnet/roslyn-compiler.

ahmad2smile:
@sharwell So my basis for these changes were my own experiments with running Source Generator with `netstandard 2.1`, `.net 5` and `.net 6`.

But my project would still not work in Visual Studio which I eventually got to fix with help from this comment: https://github.com/dotnet/roslyn/discussions/56532#discussioncomment-1356504

Which actually explains why Source Generators targetting higher `.netstandard`/`.net` TFM doesn't work in Visual Studio but work with other tooling like `dotnet` CLI.

And I think basis for aforementioned comment is actually your own comment here: https://github.com/dotnet/roslyn/discussions/47701#discussioncomment-79566

But I do see some caution that can be pointed out as mentioned here: https://github.com/dotnet/roslyn/discussions/47701#discussioncomment-1356495

jaredpar:
I understand that this change is technically correct but it is still against our guidance. Customers generally don't differentiate between these scenarios. The expectation is that if their code builds with `dotnet build` then it will build with `msbuild` (and vice versa). Having a generator that targets anything other than `netstandard2.0` breaks this and creates a lot of customer confusion. 

Developers can safely target .NET Core in their generators but it's a much more complex setup that involves effectively multi-targeting the NuPkg and toggling the setup based on whether MSBuild is running on .NET Framework or .NET Core. It's certaintly doable but more of an expert level scenario. 

The message from the compiler team remains that source generators and analyzers need to target `netstandard2.0`. The compiler has even started enforcing this by issuing warnings in cases where we detect generators / analyzers not doing this because of the customer confusion that it creates.

BillWagner:
Closing based on https://github.com/dotnet/docs/pull/29303#issuecomment-1119785397

ahmad2smile:
I understand having documentation reflect official position on the support, but I think it should also point out kind of gotcha or unintended consequences, like using `dotnet` CLI with a Source Generator that has TFM higher than `netstandard 2.0` would work.

TIPs are percise place for this kinda information, wouldn't you say? Finding the same info that I needed to understand in Github Issues and Dicussion is just not nice, which I wanted others to avoid with the help of this PR.

I imagine alot of ppl like me, thought that `dotnet` CLI and `msbuild` in VS would work exactly the same. Cuasing more confusion in face of lack of comprehensive documents on a new feature, which made me question myself on using a new feature and just giving up.

A reference for these kinda frustrations is this article: https://turnerj.com/blog/the-pain-points-of-csharp-source-generators

PR #29304, opened 2022-05-05
Correct incorrect info in regex Unicode docs

GrabYourPitchforks:
The "Unicode general categories" table in the regex docs contains some misleading wording. This cleans it up a bit.

/cc @tarekgh to check the work here. I didn't want to go too overboard by changing all the entries in the table, but I think these changes are good enough to avoid misleading people.


tarekgh (approved):

GrabYourPitchforks:
I'll wait for signoff from @gewarren or @adegeo before committing. That way the docs crew can check things.

IEvangelist (approved):

PR #29305, opened 2022-05-06
Fix typo - "sybte"

GitHubPang:
The data type `sybte` appears to be misspelt here and it should be [`sbyte`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/built-in-types), I suppose.


gewarren (approved):
Thank you @GitHubPang

PR #29309, opened 2022-05-06
Added xrefs to background service exception behavior

IEvangelist:
## Summary

Added xrefs to background service exception behavior


gewarren (approved):
LGTM, thanks.

PR #29320, opened 2022-05-06
Document a new config switch added to the portable thread pool

kouvel:
The new config switch was added in https://github.com/dotnet/runtime/pull/67399


docs/core/runtime-config/threading.md
```diff
@@ -104,6 +104,7 @@ In some cases, the thread pool detects work items that block its threads. To com
...
 | `System.Threading.ThreadPool.Blocking.DelayStepMs` | After the thread count based on `ThreadsToAddWithoutDelay` is reached, this value specifies how much additional delay to add per `ThreadsPerDelayStep` threads, which would be applied before each new thread is created. | .NET 6 |
 | `System.Threading.ThreadPool.Blocking.MaxDelayMs` | After the thread count based on `ThreadsToAddWithoutDelay` is reached, this value specifies the max delay to use before each new thread is created. | .NET 6 |
+| `System.Threading.ThreadPool.Blocking.IgnoreMemoryUsage` | By default, the rate of thread injection in response to blocking is limited by heuristics that determine whether there is sufficient physical memory available. In some situations, it may be preferable to inject threads more quickly even in low-memory situations. The memory usage heuristics may be disabled by turning off this switch. | .NET 7 |
```
gewarren, line 107:
```suggestion
| `System.Threading.ThreadPool.Blocking.IgnoreMemoryUsage` | By default, the rate of thread injection in response to blocking is limited by heuristics that determine whether there is sufficient physical memory available. In some situations, it may be preferable to inject threads more quickly even in low-memory situations. You can disable the memory usage heuristics by turning off this switch. | .NET 7 |
```

gewarren (approved):
Just one small suggestion to use active voice over passive.

PR #29328, opened 2022-05-09
Update cs0433.md

increddibelly:
## Summary

Remove a line of text that duplicates the text directly above it.
perhaps line 19 is too specific and should be changed to: "This error can also occur if:" - I proposed this in https://github.com/dotnet/docs/pull/29329

PR #29329, opened 2022-05-09
Update cs0433.md

increddibelly:
## Summary

Removed some text that was too specific, so the description becomes useful to more users.
see also https://github.com/dotnet/docs/pull/29328


BillWagner (approved):
This also LGTM @increddibelly 

I'll :shipit: now.

PR #29355, opened 2022-05-10
Collectible assembly breaking change

### LINKED ISSUE #29200: [Breaking change]: Resolving collectible assembly into a non-collectible AssemblyLoadContext now throws
### Description

We were incorrectly allowing resolving collectible assemblies into a non-collectible `AssemblyLoadContext`. That may seemingly work ok in some cases and lead to runtime crashes or unexpected `NullReferenceException` in others.
The change in https://github.com/dotnet/runtime/pull/68550 prevents that incorrect behavior by throwing an exception when the `AssemblyLoadContext.Load` or the `AssemblyLoadContext.Resolving` event returns a collectible `Assembly` and the `AssemblyLoadContext` is non-collectible. 

### Version

.NET 7 Preview 5

### Previous behavior

Returning a collectible `Assembly` in the `AssemblyLoadContext.Load` override or the `AssemblyLoadContext.Resolving` event of a non-collectible `AssemblyLoadContext` doesn't cause any exceptions to be thrown.

### New behavior

Returning a collectible `Assembly` in the `AssemblyLoadContext.Load` override or the `AssemblyLoadContext.Resolving` event of a non-collectible `AssemblyLoadContext` results in throwing a `System.IO.FileLoadException` with `System.NotSupportedException` inner exception.

### Type of breaking change

### Reason for change

This change fixes a bug. The collectible `Assembly` would be collected while the `AssemblyLoadContext` that has a reference to it is alive for the rest of the process lifetime. So if the code running in that context references anything from that `Assembly` after it is collected, it would crash the runtime or result in unexpected `NullReferenceException`, possibly even `AccessViolationException` and possibly other kinds of bad behavior.

### Recommended action

Don't return collectible assemblies in `AssemblyLoadContext.Load` overload or `AssemblyLoadContext.Resolving` event of a non-collectible `AssemblyLoadContext`. Possible workaround is to change the `AssemblyLoadContext` to collectible by passing `true` in the  `isCollectible` argument to its constructor and then keep a reference to that `AssemblyLoadContext` forever to make sure it is never collected.

### Feature area

Core .NET libraries

### Affected APIs

`AssemblyLoadContext.Load` override
`AssemblyLoadContext.Resolving` event

gewarren:
Fixes #29200 

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/7.0/collectible-assemblies?branch=pr-en-us-29355).

### DIFF
--- a/docs/core/compatibility/7.0.md
+++ b/docs/core/compatibility/7.0.md
@@ -29,6 +29,7 @@ If you're migrating an app to .NET 7, the breaking changes listed here might aff
 | - | :-: | :-: | - |
 | [API obsoletions with non-default diagnostic IDs](core-libraries/7.0/obsolete-apis-with-custom-diagnostics.md) | ✔️ | ❌ | Preview 1 |
 | [C++/CLI projects in Visual Studio](core-libraries/7.0/cpluspluscli-compiler-version.md)
+| [Collectible Assembly in non-collectible AssemblyLoadContext](core-libraries/7.0/collectible-assemblies.md) | ❌ | ✔️ | Preview 5 |
 | [FullPath and OldFullPath return fully qualified path](core-libraries/7.0/filesystemeventargs-fullpath.md) | ✔️ | ❌ | Preview 1 |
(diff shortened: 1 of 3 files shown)


IEvangelist (approved):

PR #29364, opened 2022-05-11
Fix lambda example in delegate operator page

Eideren:
## Summary
Fixed invalid syntax

### DIFF
--- a/docs/csharp/language-reference/operators/delegate-operator.md
+++ b/docs/csharp/language-reference/operators/delegate-operator.md
@@ -54,7 +54,7 @@ Action a = StaticFunction;
 Before C# 11, you'd need to use a lambda expression to reuse a single delegate object:
 
 ```csharp
-Action a = () = StaticFunction;
+Action a = () => StaticFunction();
 ```
 
 ## C# language specification


docs/csharp/language-reference/operators/delegate-operator.md
```diff
@@ -54,7 +54,7 @@ Action a = StaticFunction;
 Before C# 11, you'd need to use a lambda expression to reuse a single delegate object:
 
 ```csharp
-Action a = () = StaticFunction;
+Action a = () => StaticFunction();
```
IEvangelist, line 57:
```suggestion
Action a = () => StaticFunction;
```

Close, I think it just needed the _goes to_ operator, but not the parentheses on `StaticFunction`. Let's double check with @BillWagner

BillWagner: Either will work in this instance. That's because there is only one overload of `StaticFunction`. I'd prefer the version with the parentheses because it would work when there are more overloads.

BillWagner (approved):
This LGTM @Eideren 

Thanks for the extra 👀 on this @IEvangelist 

I'll :shipit: now.

PR #29412, opened 2022-05-11
Update dumps.md

tarurar:
## Summary

Just fixing a grammatical error

### DIFF
--- a/docs/core/diagnostics/dumps.md
+++ b/docs/core/diagnostics/dumps.md
@@ -10,7 +10,7 @@ A dump is a file that contains a snapshot of the process at the time it was crea
 
 ## Collect dumps
 
-Dumps can be collected in a variety of ways depending on which platform you are running your app on.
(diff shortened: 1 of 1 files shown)


IEvangelist (approved):

docs/core/diagnostics/dumps.md
```diff
@@ -10,7 +10,7 @@ A dump is a file that contains a snapshot of the process at the time it was crea
 
 ## Collect dumps
 
-Dumps can be collected in a variety of ways depending on which platform you are running your app on.
+Dumps can be collected in a variety of ways depending which platform you are running your app on.
```
gewarren, line 13: Hmm, I actually think the original version is correct. While the "on" is duplicated, it has different meanings in each place. "depending on [which platform]" and "running your app on [a platform]". @tdykstra do you agree?

tdykstra: Agree

PR #29422, opened 2022-05-11
Clarify new() constraint condition

### LINKED ISSUE #29362: CA1812 and new() constraint
One way to silence `CA1812` is to have a method with a generic constraint like shown below. Currently, the documentation actually contains a _counter-example_ of this `new()` constraint, which is a bit confusing.

Perhaps add this example to the documentation? 

```c#
SilenceCA1812<MyClass>();

private static void SilenceCA1812<T>() where T : new()
{
}
```

---
#### Document Details

* ID: 2d34eee1-2619-6701-31e2-7d56cc690ebb
* Version Independent ID: 8a9b0c36-3000-7b71-c0f2-a5b01e54174f
* Content: [CA1812: Avoid uninstantiated internal classes (code analysis) - .NET](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1812)
* Content Source: [docs/fundamentals/code-analysis/quality-rules/ca1812.md](https://github.com/dotnet/docs/blob/main/docs/fundamentals/code-analysis/quality-rules/ca1812.md)
* GitHub Login: @gewarren

gewarren:
Fixes #29362.

### DIFF
--- a/docs/fundamentals/code-analysis/quality-rules/ca1812.md
+++ b/docs/fundamentals/code-analysis/quality-rules/ca1812.md
@@ -62,7 +62,7 @@ It is safe to suppress a warning from this rule. We recommend that you suppress 
 
 - The class is created automatically by the runtime or ASP.NET. Some examples of automatically created classes are those that implement <xref:System.Configuration.IConfigurationSectionHandler?displayProperty=fullName> or <xref:System.Web.IHttpHandler?displayProperty=fullName>.
 
-- The class is passed as a type parameter that has a [`new` constraint](../../../csharp/language-reference/keywords/new-constraint.md). The following example will be flagged by rule CA1812:
+- The class is used as a type parameter in a class definition and has a [`new` constraint](../../../csharp/language-reference/keywords/new-constraint.md). The following example will be flagged by rule CA1812:
 
     ```csharp
     internal class MyClass


IEvangelist:
@gewarren - I'm not so sure. I feel like the original is just as accurate as the changes proposed here. I guess my question is, "the CA1812 warning is issued because the compiler is unaware of the `internal class` being implicitly instantiated by the generic `MyGeneric.Create` method, how does this clarify that?"

docs/fundamentals/code-analysis/quality-rules/ca1812.md
```diff
@@ -62,7 +62,7 @@ It is safe to suppress a warning from this rule. We recommend that you suppress
 
 - The class is created automatically by the runtime or ASP.NET. Some examples of automatically created classes are those that implement <xref:System.Configuration.IConfigurationSectionHandler?displayProperty=fullName> or <xref:System.Web.IHttpHandler?displayProperty=fullName>.
 
-- The class is passed as a type parameter that has a [`new` constraint](../../../csharp/language-reference/keywords/new-constraint.md). The following example will be flagged by rule CA1812:
+- The class is used as a type parameter in a class definition and has a [`new` constraint](../../../csharp/language-reference/keywords/new-constraint.md). The following example will be flagged by rule CA1812:
```
BillWagner, line 65: Reading the other comments, maybe this clarifies why this example generates the warning:

```suggestion
- The class is used as a type parameter in a class definition and has a [`new` constraint](../../../csharp/language-reference/keywords/new-constraint.md). The generic class instantiates its type parameter, but the isn't instantiated elsewhere. The following example will be flagged by rule CA1812:
```

BillWagner (approved):
This LGTM @gewarren 

I made one possible suggestion for how to clarify why a generic class can show this warning. You can decide if that helps, or not, or if there's better language here.

gewarren:
@IEvangelist The original issue is not about why the existing example raises CA1812. Rather, it's [specifically about the wording](https://github.com/dotnet/docs/issues/29362#issuecomment-1124286140) "[It's safe to suppress the warning when] the class is passed as a type parameter that has a `new()` constraint". It's a true statement when the type parameter is a part of a generic _class definition_. But when the type parameter is on a _method declaration_, CA1812 isn't raised (as the OP proves with their example), so we need to be more specific about what will cause CA1812 to be raised.

docs/fundamentals/code-analysis/quality-rules/ca1812.md
gewarren: The original issue wasn't about why this case shows the warning, so I'm going to resolve this comment.

PR #29449, opened 2022-05-13
Fix System.CommandLine model-binding ComplexType.cs example

isobel-cullen:
## Summary

Description for the `lastNameOption` referred to the first name. I assume this is incorrect.


IEvangelist (approved):

PR #29454, opened 2022-05-13
Change ref from .NET Framework to .NET

adegeo:
## Summary

I was reading through this thought I would clarify that it's not tied to the .NET Framework product, but to .NET as a whole.

@BillWagner @gewarren


gewarren (approved):
LGTM

PR #29457, opened 2022-05-13
Update "z" custom format specifier

### LINKED ISSUE #28225: Undocumented breaking change since .NET Core 3.x: Formatting DateTime values with "zzz"
**Issue description**

Sometime around .NET Core 3.x, the behaviour of the `zzz` custom Date-and-Time formatting token changed but this was not/is not documented.

Previously - and in .NET Framework 4.x, using the `zzz` specifier to format a `DateTime` (not `DateTimeOffset`) would always result in the Local Computer clock's UTC offset, regardless of `DateTime.Kind`. This is the current documented behavior:

https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings#zzzSpecifier

> With [DateTime](https://docs.microsoft.com/en-us/dotnet/api/system.datetime) values, the "zzz" custom format specifier represents the signed offset of the local operating system's time zone from UTC, measured in hours and minutes. It doesn't reflect the value of an instance's [DateTime.Kind](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.kind) property. For this reason, the "zzz" format specifier is not recommended for use with [DateTime](https://docs.microsoft.com/en-us/dotnet/api/system.datetime) values.

However, in .NET Core 3.1 and later (including .NET 5, .NET 6, etc), the behaviour was changed to now start respecting `DateTime.Kind`.

Here's a screenshot showing identical programs demonstrating  the change between .NET Framework 4.8 (in Linqpad 5) and .NET 6 (in Linqpad 7):

```
DateTime dtUtc         = new DateTime( ticks: 637802866463298955, kind: DateTimeKind.Utc         );
DateTime dtUnspecified = new DateTime( ticks: 637802866463298955, kind: DateTimeKind.Unspecified );
DateTime dtLocal       = new DateTime( ticks: 637802866463298955, kind: DateTimeKind.Local       );

DateTime[] values = new[] { dtUtc, dtUnspecified, dtLocal };

foreach( DateTime v in values )
{
	v.ToString( "yyyy-MM-dd'T'HH':'mm':'sszzz" ).Dump( v.Kind.ToString() );
	v.ToString( "yyyy-MM-dd'T'HH':'mm':'ssK" ).Dump( v.Kind.ToString() );
}
```

image

To my surprise, only a few people have noticed it so far: I've only seen this StackOverflow QA from 2020 talking about it: https://stackoverflow.com/questions/65004352/c-sharp-datetime-tostring-with-zzz-breaks-in-dotnet-framework-but-not-in-dotnet

The StackOverflow post also mentions that the text of the Visual Studio MDA for date+time formats is wrong too:

> 'A UTC DateTime is being converted to text in a format that is only correct for local times. This can happen when calling DateTime.ToString using the 'z' format specifier, which will include a local time zone offset in the output. In that case, either use the 'Z' format specifier, which designates a UTC time, or use the 'o' format string, which is the recommended way to persist a DateTime in text. 

...but I'm unsure where to file a bug for MDA help text.

Of note, the reknown [Jon Skeet chimed in](https://stackoverflow.com/a/65004441/159145):

> Note that .NET Core (and .NET 5.0) apparently doesn't behave as documented. While you could argue it's "fixed" in .NET Core, I'd suggest that **behaving in an undocumented way is a bug in itself**, and one that could make code migration harder than expected.

------------------

I was going to submit a quick PR, but as this involves documenting a breaking change I assume it means it's more complicated than adding a sentence of two buried in [that paragraph](https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings#zzzSpecifier)

**Target framework**

- [x] .NET Core
- [x] .NET Framework
- [ ] .NET Standard

<details>
<summary><strong>dotnet --info output</strong> or <strong>About VS info</strong></summary>

```console
<replace>
```
</details>

gewarren:
Fixes #28225.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings?branch=pr-en-us-29457#offset-z-format-specifier).

### DIFF
--- a/docs/standard/base-types/custom-date-and-time-format-strings.md
+++ b/docs/standard/base-types/custom-date-and-time-format-strings.md
@@ -1,7 +1,7 @@
 ---
 title: "Custom date and time format strings"
 description: Learn to use custom date and time format strings to convert DateTime or DateTimeOffset values into text representations, or to parse strings for dates & times.
-ms.date: "03/30/2017"
+ms.date: 05/12/2022
 ms.topic: reference
 dev_langs: 
 - "csharp"
@@ -356,7 +356,7 @@ The following example includes the "HH" custom format specifier in a custom form
 
 The "K" custom format specifier represents the time zone information of a date and time value. When this format specifier is used with <xref:System.DateTime> values, the result string is defined by the value of the <xref:System.DateTime.Kind%2A?displayProperty=nameWithType> property:
 
-- For the local time zone (a <xref:System.DateTime.Kind%2A?displayProperty=nameWithType> property value of <xref:System.DateTimeKind.Local?displayProperty=nameWithType>), this specifier is equivalent to the "zzz" specifier and produces a result string containing the local offset from Coordinated Universal Time (UTC); for example, "-07:00".
+- For the local time zone (a <xref:System.DateTime.Kind%2A?displayProperty=nameWithType> property value of <xref:System.DateTimeKind.Local?displayProperty=nameWithType>), this specifier produces a result string containing the local offset from Coordinated Universal Time (UTC); for example, "-07:00".
 
 - For a UTC time (a <xref:System.DateTime.Kind%2A?displayProperty=nameWithType> property value of <xref:System.DateTimeKind.Utc?displayProperty=nameWithType>), the result string includes a "Z" character to represent a UTC date.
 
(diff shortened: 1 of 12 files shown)


tarekgh (approved):
Thanks @gewarren for fixing the docs.

PR #29464, opened 2022-05-15
Update using-delegates.md

wuyuansushen:
## Summary

The instantiated delegate is an object. It should be argument rather than parameter when use.


docs/csharp/programming-guide/delegates/using-delegates.md
```diff
@@ -18,7 +18,7 @@ A delegate object is normally constructed by providing the name of the method th
...
+Delegate types are derived from the <xref:System.Delegate> class in .NET. Delegate types are [sealed](../../language-reference/keywords/sealed.md)—they cannot be derived from— and it is not possible to derive custom classes from <xref:System.Delegate>. Because the instantiated delegate is an object, it can be passed as a argument, or assigned to a property. This allows a method to accept a delegate as a parameter, and call the delegate at some later time. This is known as an asynchronous callback, and is a common method of notifying a caller when a long process has completed. When a delegate is used in this fashion, the code using the delegate does not need any knowledge of the implementation of the method being used. The functionality is similar to the encapsulation interfaces provide.
```
BillWagner, line 21: one small grammar nit:

```suggestion
Delegate types are derived from the <xref:System.Delegate> class in .NET. Delegate types are [sealed](../../language-reference/keywords/sealed.md)—they cannot be derived from— and it is not possible to derive custom classes from <xref:System.Delegate>. Because the instantiated delegate is an object, it can be passed as an argument, or assigned to a property. This allows a method to accept a delegate as a parameter, and call the delegate at some later time. This is known as an asynchronous callback, and is a common method of notifying a caller when a long process has completed. When a delegate is used in this fashion, the code using the delegate does not need any knowledge of the implementation of the method being used. The functionality is similar to the encapsulation interfaces provide.
```

BillWagner (approved):
Thanks @wuyuansushen 

With one small grammar fix, this is great.
I'll :shipit: now.

PR #29465, opened 2022-05-15
Missing name Console.

LukaHietala:
Fixes: Program.cs(36,9): error CS0103: The name 'Console' does not exist in the current context.

### DIFF
--- a/docs/csharp/fundamentals/types/snippets/classes/Program.cs
+++ b/docs/csharp/fundamentals/types/snippets/classes/Program.cs
@@ -1,4 +1,6 @@
-﻿public class Person
+using System;
+
+public class Person
 {
     // Constructor that takes no arguments:
     public Person()


docs/csharp/fundamentals/types/snippets/classes/Program.cs
```diff
@@ -1,4 +1,6 @@
-﻿public class Person
+using System;
```
Youssef1313, line 1: The using is unnecessary due to the use of `ImplicitUsings`:

https://github.com/dotnet/docs/blob/09d90f573fcf78373a7c18b93938b74f97e6ddf3/docs/csharp/fundamentals/types/snippets/classes/classes.csproj#L7

IEvangelist:
Hi @LukaH0, this PR isn't needed as @Youssef1313 called attention to already, this project is set up to use implicit using. I will close this PR, but let me know if you have any additional questions. Thank you

Issue #27520, opened 2021-12-13
F# arrays - element access syntax


jcmrva:
https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/arrays#access-elements

For this section, the text should be updated to match the code, maybe something like:

> You can access array elements by using brackets (`[` and `]`). The original dot syntax (`.[index]`) is still supported but no longer recommended as of F# 6.0.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6cda87e0-998a-086f-f5ed-2290d7a97966
* Version Independent ID: 45152c64-0643-3c5a-93cb-a2f689170887
* Content: [Arrays in F#](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/arrays)
* Content Source: [docs/fsharp/language-reference/arrays.md](https://github.com/dotnet/docs/blob/main/docs/fsharp/language-reference/arrays.md)
* Product: **dotnet-fsharp**
* GitHub Login: @cartermp
* Microsoft Alias: **dotnetcontent**

BillWagner:
Thanks for noting this @jcmrva 

I've added this to our backlog, and added the "up for grabs" label for any community members interested in submitting a PR.

eriawan:
@BillWagner 
This issue should be closed by PR #27520 because that PR is the one that fixed the F# array element access syntax.

BillWagner:
Fixed in #29468

PR #29468, opened 2022-05-15
Update F# element access syntax #27520

jcmrva:
## Summary

Remove suggestion of dot operator for array element access.

Fixes #Issue_Number #27520


BillWagner (approved):
Thanks for making this change @jcmrva 

PR #29540, opened 2022-05-19
update dependencies for roslyn samles

BillWagner:
The .NET version was out of date.


IEvangelist:
Did you verify that these work? I thought there was a limitation with what versions were supported for some of these, but can't recall?

IEvangelist (approved):

BillWagner:
> Did you verify that these work? I thought there was a limitation with what versions were supported for some of these, but can't recall?

Yes. The add must use be compiled against netstandard 2.0. and full framework for loading into VS.

PR #29555, opened 2022-05-20
Fix note callout in work-with-resx-files-programmatically.md

hmmwhatsthisdo:
## Summary
This PR corrects a `[!NOTE]` callout block in work-with-resx-files-programmatically.md, which was previously rendering incorrectly due to the callout block header being mistyped:

image


IEvangelist (approved):

PR #29562, opened 2022-05-20
Fix typo

DAXaholic:
`maxTries` should be rather be `maxRetries` as the former one is not a valid property on the [retry options](https://docs.microsoft.com/en-us/dotnet/api/azure.core.retryoptions?view=azure-dotnet) object.


CamSoper (approved):
LGTM! Thank you for the fix!

PR #29577, opened 2022-05-21
Update components.md

oneolddev:
The original sample code will generate *warning CS8625: Cannot convert null literal to non-nullable reference type* when *net6.0* is targeted with *Nullable enabled* which are the defaults when creating a Blazor application in VS2022.  This change resolves the warning.

### References

https://docs.microsoft.com/en-us/dotnet/csharp/nullable-warnings
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving

## Summary

Initializes referenced component with a null forgiving default.


danroth27 (approved):
Looks good.

IEvangelist (approved):

PR #29580, opened 2022-05-22
System.CommandLine beta 4

tdykstra:


docs/standard/commandline/model-binding.md
```diff
@@ -183,30 +191,30 @@ Besides the file system types and `Uri`, the following types are supported:
...
-`System.CommandLine` allows you to use some types in handlers by adding parameters for them to the handler signature. The available types include:
+`System.CommandLine` allows you to use some types in handlers by providing a `SetHandler` overload that gives you access to the <xref:System.CommandLine.Invocation.InvocationContext>. The available types include:
 
+* <xref:System.CommandLine.Invocation.InvocationContext>
 * <xref:System.Threading.CancellationToken>
 * <xref:System.CommandLine.IConsole>
-* <xref:System.CommandLine.Invocation.InvocationContext>
 * <xref:System.CommandLine.Parsing.ParseResult>
 
 Other types can be injected by using custom binders. For more information, see [Dependency injection](dependency-injection.md).
```
baronfel, line 201: I wouldn't call this injection anymore, since (IIRC correct me if I'm wrong here @jonsequitur), there's no DI of the System.CommandLine types anymore. If you want the cancellationToken, you need to get the InvocationContext and access the property on it. If you want an IConsole, you need to get the InvocationContext and access the property on it. And so on.

gewarren (approved):
LGTM

docs/standard/commandline/model-binding.md
jonsequitur: I didn't actually remove those yet but it's probably a good idea. `SetHandler` won't inject them anymore in any case.

docs/standard/commandline/dependency-injection.md
```diff
@@ -13,7 +13,7 @@ ms.topic: how-to
 
 [!INCLUDE [scl-preview](../../../includes/scl-preview.md)]
 
-Use a [custom binder](model-binding.md#model-binding-more-than-16-options-and-arguments) to inject custom types into a command handler.
+Use a [custom binder](model-binding.md#model-binding-more-than-8-options-and-arguments) to inject custom types into a command handler.
```
jonsequitur, line 16: I wonder if "model binding" is the correct term here, since we don't do any kind of binding of rich models. We really just bind parameters to the handler. 

@KathleenDollard @baronfel

jonsequitur (approved):

docs/standard/commandline/dependency-injection.md
KathleenDollard: I like parameter binding.

jonsequitur (approved):

PR #29585, opened 2022-05-23
Change "finally" to "catch"

### LINKED ISSUE #29584: CA2000 "try/finally" should likely be "try/catch"
The page reads "To return a disposable object, the object must be constructed in a `try/finally` block outside of a using block."
I am assuming it should read `try/catch` instead of `try/finally`, as the disposable is returned from the method and as such should not be disposed yet?

---
#### Document Details

* ID: 80e2f8df-7039-c826-764e-88abefb6744b
* Version Independent ID: f75bace0-4c3a-5936-8e19-9e2334fd9848
* Content: [CA2000: Dispose objects before losing scope (code analysis) - .NET](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2000)
* Content Source: [docs/fundamentals/code-analysis/quality-rules/ca2000.md](https://github.com/dotnet/docs/blob/main/docs/fundamentals/code-analysis/quality-rules/ca2000.md)
* GitHub Login: @gewarren

davidwalschots:
## Summary

A disposable is returned from the method and as such shouldn't be disposed yet. A `Dispose` call within a `finally` block would always be executed, which is not what one would want in this situation. A `catch` block makes more sense.

Fixes #29584

### DIFF
--- a/docs/fundamentals/code-analysis/quality-rules/ca2000.md
+++ b/docs/fundamentals/code-analysis/quality-rules/ca2000.md
@@ -52,7 +52,7 @@ To fix a violation of this rule, call <xref:System.IDisposable.Dispose%2A> on th
 
 You can use the [`using` statement](../../../csharp/language-reference/keywords/using-statement.md) ([`Using`](../../../visual-basic/language-reference/statements/using-statement.md) in Visual Basic) to wrap objects that implement <xref:System.IDisposable>. Objects that are wrapped in this manner are automatically disposed at the end of the `using` block. However, the following situations should not or cannot be handled with a `using` statement:
 
-- To return a disposable object, the object must be constructed in a `try/finally` block outside of a `using` block.
+- To return a disposable object, the object must be constructed in a `try/catch` block outside of a `using` block.
 
 - Do not initialize members of a disposable object in the constructor of a `using` statement.
 


IEvangelist:
Hi @davidwalschots - did read through [Example 1](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2000#example-1)? It explicitly states the reason why to use `try/finally` and not `catch`. I think that this is accurate, and changing the text would make this incorrect.

davidwalschots:
@IEvangelist I see. I didn't see the connection between the two, but now I understand it. Maybe adding a reference from the example to the bullet point would make this clearer. 

Reading the example, I am not sure why I'd go for `finally` over `catch`. To explain why, I'll quickly guide you through why I expected `catch`.

I ran into the following (simplified) example in a code base which triggered CA2000 (in .NET 4.8/C# 7.3):
```csharp
public HttpResponseMessage Foo(Bar[] someArray)
{
    var response = new HttpResponseMessage(HttpStatusCode.OK);
    response.Content = new ByteArrayContent(someArray);

    return response;
}
```
After making the following change it no longer triggers CA2000:
```csharp
public HttpResponseMessage Foo(Bar[] someArray)
{
    var response = new HttpResponseMessage(HttpStatusCode.OK);
    try
    {
        response.Content = new ByteArrayContent(someArray);

        return response;
    }
    catch
    {
        response.Dispose();
        throw;
    }
}
```
Applying the same pattern to the example, I would expect this to be equivalent:
```csharp
public SerialPort OpenPort2(string portName)
{
   SerialPort port = new SerialPort(portName);
   try
   {
      port.Open();
      SomeMethod();
      // Add any other methods above this line

      return port;
   }
   catch
   {
      port.Close();
      //   ^^^^^ Interestingly enough Close is called instead of Dispose. 
      //         While the internals of SerialPort might itself also just
      //         call Close, this does feel like a leaky abstraction and
      //         might not be the best example.
      throw;
   }
}
```
I'm not really seeing why I'd pick the `finally` version over the `catch` version. There might be valid reasons for it, in which case I'd be very interested in understanding the reasoning.

I'm not sure, but the following text seems to only explain the difference between `OpenPort1` and `OpenPort2`, not an actual reason for using `finally` vs `catch`:
> If you're implementing a method that returns a disposable object, use a try/finally block without a catch block to make sure that the object is disposed. By using a try/finally block, you allow exceptions to be raised at the fault point and make sure that object is disposed.

Lastly, note that the following text seems incorrect, due to the method throwing an exception. In other words: while `port` is indeed `null`, it is no longer returned due to the exception throw within `try` bubbling up.
> The returned port object will contain the opened SerialPort object if the operations of the method succeeded, or it will be null if an operation failed.

GeeLaw:
`catch`ing and re`throw`ing is different from just doing a `finally`:

- Due to the two-pass exception handling model of CLR, the universal `catch` will prevent any exception filters lower down from examining the state when the exception is first `throw`n (they could get a chance to examine the state after this universal `catch` is done re`throw`ing).
- `catch`ing and re`throw`ing is presumably less efficient than a simple `finally`.
- Semantically, `catch`ing means there's a high chance for the exception to be handled, which is not the case here. The `finally` block spells whatever clean-up that must be done.

In the case of clean-up, a simple `finally` is an idiom and preferred over `catch`ing and re`throw`ing.

The first point can be demonstrated with the following program:

```csharp
using System;

namespace Example
{
    internal class Program
    {
        static int value = 0;

        static void UseTryFinally()
        {
            value = 0;
            try
            {
                try
                {
                    value = 5;
                    throw new Exception();
                }
                finally
                {
                    Console.WriteLine("UseTryFinally: inner finally, value = {0}.", value);
                    value = 0;
                }
            }
            catch when (value == 5)
            {
                Console.WriteLine("UseTryFinally: outer catch when (value == 5), value = {0}.", value);
            }
            catch when (value != 5)
            {
                Console.WriteLine("UseTryFinally: outer catch when (value != 5), value = {0}.", value);
            }
        }

        static void UseTryCatchRethrow()
        {
            value = 0;
            try
            {
                try
                {
                    value = 5;
                    throw new Exception();
                }
                catch
                {
                    Console.WriteLine("UseTryCatchRethrow: inner catch-rethrow, value = {0}.", value);
                    value = 0;
                    throw;
                }
            }
            catch when (value == 5)
            {
                Console.WriteLine("UseTryCatchRethrow: outer catch when (value == 5), value = {0}.", value);
            }
            catch when (value != 5)
            {
                Console.WriteLine("UseTryCatchRethrow: outer catch when (value != 5), value = {0}.", value);
            }
        }

        static void Main()
        {
            UseTryFinally();
            UseTryCatchRethrow();
        }
    }
}
```

The output is

```text
UseTryFinally: inner finally, value = 5.
UseTryFinally: outer catch when (value == 5), value = 0.
UseTryCatchRethrow: inner catch-rethrow, value = 5.
UseTryCatchRethrow: outer catch when (value != 5), value = 0.
```

Note how the outer handler is different in the two cases.

davidwalschots:
@GeeLaw thank you for your detailed explanation. I wasn't yet aware of the two-pass exception handling model of the CLR. After reading more about it and looking through your example I understand the reasoning. In itself using `finally` does make a lot of sense as "finally" reads better semantically. 

It's quite likely I'm not the only one who would just use `catch` here to reduce the amount of code, as I'm sure many devs are unaware of the two-pass exception handling semantics. It might be a good idea to link from the CA2000 page to a page explaining the two-pass model and semantics, such that not only guidance is given, but devs can also gain deeper understanding of the underlying reasoning.

I'll close this PR now.

PR #29597, opened 2022-05-24
Breaking change doc - clarify when adding fields to structs is allowed

GrabYourPitchforks:
## Summary

Based on an internal chat, clarifying when it's acceptable to add instance fields to structs. Removes the word "state" and refers instead to public and nonpublic fields, which are more generally understood terms.

tl;dr: If a struct goes out the door containing only public instance fields (or no fields at all), we can never add instance fields to it in any future release. It's both a source breaking change and - due to the recent introduction of `[SkipLocalsInit]` - risks leaking uninitialized memory.


docs/core/compatibility/index.md
```diff
@@ -183,9 +183,13 @@ Changes in this category modify the public surface area of a type. Most of the c
 
    The return value of a method or the type of a property or field cannot be modified. For example, the signature of a method that returns an <xref:System.Object> cannot be changed to return a <xref:System.String>, or vice versa.
 
-- ❌ **DISALLOWED: Adding a field to a struct that previously had no state**
+- ❌ **DISALLOWED: Adding an instance field to a struct that has no nonpublic fields**
 
-  Definite assignment rules allow the use of uninitialized variables so long as the variable type is a stateless struct. If the struct is made stateful, code could end up with uninitialized data. This is both potentially a source breaking and a binary breaking change.
+  If a struct has only public fields or has no fields at all, callers can declare locals of that struct type without calling the struct's ctor or first initializing the local to _default(T)_, so long as all public fields are set on the struct before first use. Adding any new fields - public or nonpublic - to such a struct is a source breaking change for these callers, as the compiler will now require the additional fields to be initialized.
```
gewarren, line 188:
```suggestion
  If a struct has only public fields or has no fields at all, callers can declare locals of that struct type without calling the struct's constructor or first initializing the local to `default(T)`, so long as all public fields are set on the struct before first use. Adding any new fields - public or nonpublic - to such a struct is a source breaking change for these callers, as the compiler will now require the additional fields to be initialized.
```

docs/core/compatibility/index.md
```diff
@@ -183,9 +183,13 @@ Changes in this category modify the public surface area of a type. Most of the c
...
+
+  Additionally, adding any new fields - public or nonpublic - to such a struct is a binary breaking change to callers who have applied `[SkipLocalsInit]` to their code. Since the compiler wasn't aware of these fields at compile time, it could emit IL which does not fully initialize the struct, leading to the struct being created from uninitialized stack data.
```
gewarren, line 190:
```suggestion
  Additionally, adding any new fields - public or nonpublic - to a struct with no fields or only public fields is a binary breaking change to callers who've applied `[SkipLocalsInit]` to their code. Since the compiler wasn't aware of these fields at compile time, it could emit IL that doesn't fully initialize the struct, leading to the struct being created from uninitialized stack data.
```

gewarren (approved):

gewarren:
@GrabYourPitchforks Putting this back on your radar.

docs/core/compatibility/index.md
terrajobst, line 188: Why does it matter whether or not there were fields before? If there were fields, adding a new field is a breaking change, no?

```C#
struct Foo {
   int X;
   int Y;
   int Z; // new
}

static Foo CreateFoo()
{
   return new Foo { X = 1, Y = 2 }; // error: Z is not initialized
}
```

gewarren: Your example is covered by the description because all the existing fields are public (has no nonpublic fields). So it's disallowed. If any of the existing fields were private, then the struct would have previously had to be initialized to `default(T)`, and that would apply to the new field too. I'll make a suggestion to change the wording from "Adding an instance field to a struct that has no nonpublic fields" to "Adding an instance field to a struct that has only public fields or no fields".

docs/core/compatibility/index.md
```diff
@@ -183,9 +183,13 @@ Changes in this category modify the public surface area of a type. Most of the c
 
    The return value of a method or the type of a property or field cannot be modified. For example, the signature of a method that returns an <xref:System.Object> cannot be changed to return a <xref:System.String>, or vice versa.
 
-- ❌ **DISALLOWED: Adding a field to a struct that previously had no state**
+- ❌ **DISALLOWED: Adding an instance field to a struct that has no nonpublic fields**
```
gewarren, line 186:
```suggestion
- ❌ **DISALLOWED: Adding an instance field to a struct that has only public fields or no fields**
```

gewarren: The double negative is a little confusing I think.

PR #29599, opened 2022-05-24
.NET Standard 2 does not support net461

SimonCropp:
it is misleading to imply that is does when there are so many incompatibilities between the two
https://www.google.com/search?q=netStandard2+net461+error

Having it as subtext is pretty poor way to communicate

> While NuGet considers .NET Framework 4.6.1 as supporting .NET Standard 1.5 through 2.0, there are several issues with consuming .NET Standard libraries that were built for those versions from .NET Framework 4.6.1 projects. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher.

### DIFF
--- a/includes/net-standard-2.0.md
+++ b/includes/net-standard-2.0.md
@@ -3,7 +3,7 @@
 | .NET implementation         | Version support                                   |
 |-----------------------------|---------------------------------------------------|
 | .NET and .NET Core          | 2.0, 2.1, 2.2, 3.0, 3.1, 5.0, 6.0                 |
-| .NET Framework <sup>1</sup> | 4.6.1 <sup>2</sup>, 4.6.2, 4.7, 4.7.1, 4.7.2, 4.8 |
+| .NET Framework <sup>1</sup> | <sup>2</sup> 4.6.2, 4.7, 4.7.1, 4.7.2, 4.8 |
 | Mono                        | 5.4, 6.4                                          |
 | Xamarin.iOS                 | 10.14, 12.16                                      |
 | Xamarin.Mac                 | 3.8, 5.16                                         |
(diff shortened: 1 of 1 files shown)


IEvangelist:
I defer to @terrajobst on this one. From the standpoint of the docs, we were relying on Immo's interactive table as a source of truth. Immo, are you good with this change?

terrajobst (approved):
The suggested wording seems clearer. Thanks @SimonCropp!

includes/net-standard-2.0.md
```diff
@@ -3,7 +3,7 @@
 | .NET implementation         | Version support                                   |
 |-----------------------------|---------------------------------------------------|
 | .NET and .NET Core          | 2.0, 2.1, 2.2, 3.0, 3.1, 5.0, 6.0                 |
-| .NET Framework <sup>1</sup> | 4.6.1 <sup>2</sup>, 4.6.2, 4.7, 4.7.1, 4.7.2, 4.8 |
+| .NET Framework <sup>1</sup> | <sup>2</sup> 4.6.2, 4.7, 4.7.1, 4.7.2, 4.8        |
```
gewarren, line 6:
```suggestion
| .NET Framework <sup>1</sup> | 4.6.2, 4.7, 4.7.1, 4.7.2, 4.8 <sup>2</sup>        |
```

includes/net-standard-2.0.md
```diff
@@ -13,7 +13,7 @@
 
 <sup>1</sup> The versions listed for .NET Framework apply to .NET Core 2.0 SDK and later versions of the tooling. Older versions used a different mapping for .NET Standard 1.5 and higher. You can [download tooling for .NET Core tools for Visual Studio 2015](https://github.com/dotnet/core/blob/main/release-notes/download-archives) if you cannot upgrade to Visual Studio 2017 or a later version.
 
-<sup>2</sup> The versions listed here represent the rules that NuGet uses to determine whether a given .NET Standard library is applicable. While NuGet considers .NET Framework 4.6.1 as supporting .NET Standard 1.5 through 2.0, there are several issues with consuming .NET Standard libraries that were built for those versions from .NET Framework 4.6.1 projects. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher.
+<sup>2</sup> The versions listed here represent the rules that NuGet uses to determine whether a given .NET Standard library is applicable. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher.
```
gewarren, line 16:
```suggestion
<sup>2</sup> The versions listed here represent the rules that NuGet uses to determine whether a given .NET Standard library is applicable. For .NET Framework projects that need to use .NET Standard 2.0, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher.
```

gewarren (commented):
Left a couple suggestions. I think it would look funny to have the superscript 2 at the start of the list.

terrajobst (changes requested):
Oh, I didn't realize that @SimonCropp was suggesting removing 4.6.1 from the table. I think that would be a mistake. People need to know the compatibility rules baked into NuGet because it controls which assets are selected. I think we should leave 4.6.1 there or it can result in surprises.

includes/net-standard-2.0.md
terrajobst: I think we should revert this part.

includes/net-standard-2.0.md
terrajobst: I like that!

SimonCropp:
>  I think we should leave 4.6.1 there or it can result in surprises.

the point of this PR is that there are so many issues with mixing  4.6.1 with netstandard2, that having in that table is leading people down a path of pain,

SimonCropp:
> Sadly, this is the kind of bug we can't fix without breaking even more stuff

i get that. and i am ok with this. but given the caveats and problems people have, i think having net461 in that table does more harm than good

includes/net-standard-2.0.md
IEvangelist, line 6:
```suggestion
| .NET Framework <sup>1</sup> | 4.6.1 <sup>2</sup>, 4.6.2, 4.7, 4.7.1, 4.7.2, 4.8 |
```

Revert this per @terrajobst

IEvangelist (approved):

PR #29626, opened 2022-05-25
Decrypt EnvelopedCms breaking change

### LINKED ISSUE #29263: [Breaking change]: Decrypting EnvelopedCms no longer unwraps double-wrapped content
### Description

In .NET Core 2.0 on macOS and Linux, the EnvelopedCms implementation wrapped content in an extra ASN.1 OCTET STRING value.  In order to maintain compatibility with processing content created with this error, the EnvelopedCms class still looked at the decrypted content and tried removing the extra data on Windows when using an external private key as well as on macOS and Linux.

Unfortunately, this opportunistic compatibility code cannot distinguish between documents that were created incorrectly and documents that were created correctly but have the same data shape.

### Version

.NET 7 Preview 6

### Previous behavior

When executing any overload of `EnvelopedCms.Decrypt` on macOS or Linux, or the `EnvelopedCms.Decrypt(RecipientInfo,AsymmetricAlgorithm)` on Windows, if the decrypted content started with the byte value `0x04` and a legally-encoded ASN.1 BER length value that was less than or equal to the number of bytes remaining in the content, the data provided in `envelopedCms.ContentInfo.Content` property only received the data associated with the content octets portion of the value when treated as an ASN.1 OCTET STRING.

For example, if the initially decrypted content was the byte series `{ 0x04, 0x03, 0x01, 0x02, 0x03 }` or `{ 0x04, 0x03, 0x01, 0x02, 0x03, [continued content] }` the value of `envelopedCms.ContentInfo.Content` was the byte series `{ 0x01, 0x02, 0x03 }`.

Values that did not start with 0x04, or started with 0x04 but were not followed by an acceptly-encoded length value, were fully reported.

### New behavior

The EnvelopedCms class will no longer attempt to work around the previous issue, and will always report the decrypted content faithfully.

Users who are processing documents that were created by the .NET Core 2.0 version of the EnvelopedCms class on macOS or Linux will see extra data at the beginning of their content.

### Type of breaking change

### Reason for change

The compatibility code could not differentiate between incorrectly created documents and documents that were legitimately transporting data that looked like a BER-encoded ASN.1 OCTET STRING.

Due to the nature of the BER encoding, callers that were negatively impacted by this compatibility code could not easily recover their missing data.

### Recommended action

Callers that are reading documents created with EnvelopedCms on .NET Core 2.0 for macOS or Linux can remove the extra data with the `AsnDecoder` class from the `System.Formats.Asn1` NuGet package, which is already a dependency of the `EnvelopedCms` class:

```C#
envelopedCms.Decrypt(...);

byte[] content = envelopedCms.ContentInfo.Content;

if (envelopedCms.ContentInfo.Oid.Value == "1.2.840.113549.1.7.1")
{
    if (content?.Length > 0 && content[0] == 0x04)
    {
        try
        {
            content = AsnDecoder.ReadOctetString(content, AsnEncodingRules.BER, out _);
        }
        catch (AsnContentException)
        {
        }
    }
}
```

### Feature area

Cryptography

### Affected APIs

This affects the data provided in `P:System.Security.Cryptography.Pkcs.ContentInfo.Content` from `P:System.Security.Cryptography.Pkcs.EnvelopedCms.ContentInfo` after a call to

All Operating Systems:
* `M:System.Security.Cryptography.Pkcs.EnvelopedCms.Decrypt(System.Security.Cryptography.Pkcs.RecipientInfo, System.Security.Cryptography.AsymmetricAlgorithm)`

Operating Systems, other than Windows:
* `O:System.Security.Cryptography.Pkcs.EnvelopedCms.Decrypt`

gewarren:
Fixes #29263

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/cryptography/7.0/decrypt-envelopedcms?branch=pr-en-us-29626)

### DIFF
--- a/docs/core/compatibility/cryptography/7.0/decrypt-envelopedcms.md
+++ b/docs/core/compatibility/cryptography/7.0/decrypt-envelopedcms.md
@@ -11,7 +11,7 @@ Unfortunately, this opportunistic compatibility code cannot distinguish between 
 
 ## Previous behavior
 
-Previously, if the decrypted content started with the byte value 0x04 and a legally encoded ASN.1 BER length value that was less than or equal to the number of bytes remaining in the content, the data provided in the `envelopedCms.ContentInfo.Content` property only received the data associated with the content octets portion of the value when treated as an ASN.1 OCTET STRING.
+Previously, if the decrypted content started with the byte value `0x04` and a legally encoded ASN.1 BER length value that was less than or equal to the number of bytes remaining in the content, the data provided in the `envelopedCms.ContentInfo.Content` property only received the data associated with the content octets portion of the value when treated as an ASN.1 OCTET STRING.
 
 For example, if the initially decrypted content was the byte series `{ 0x04, 0x03, 0x01, 0x02, 0x03 }` or `{ 0x04, 0x03, 0x01, 0x02, 0x03, [continued content] }`, the value of `envelopedCms.ContentInfo.Content` was the byte series `{ 0x01, 0x02, 0x03 }`.
 
--- a/docs/core/compatibility/cryptography/7.0/decrypt-envelopedcms.md
+++ b/docs/core/compatibility/cryptography/7.0/decrypt-envelopedcms.md
@@ -1,0 +1,71 @@
+---
+title: "Breaking change: Decrypting EnvelopedCms doesn't double unwrap"
+description: Learn about the .NET 7 breaking change in cryptography where decrypting EnvelopedCms no longer removes extra data that was introduced by a bug in .NET Core 2.0.
+ms.date: 05/24/2022
+---
+# Decrypting EnvelopedCms doesn't double unwrap
+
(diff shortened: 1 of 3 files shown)


bartonjs (approved):

BillWagner (approved):
This LGTM @gewarren 

I found one small nit, and then this is ready to :shipit:

docs/core/compatibility/cryptography/7.0/decrypt-envelopedcms.md
```diff
@@ -0,0 +1,71 @@
...
+
+Unfortunately, this opportunistic compatibility code cannot distinguish between documents that were created incorrectly and documents that were created correctly but have the same data shape.
+
+## Previous behavior
+
+Previously, if the decrypted content started with the byte value `0x04` and a legally encoded ASN.1 BER length value that was less than or equal to the number of bytes remaining in the content, the data provided in the `envelopedCms.ContentInfo.Content` property only received the data associated with the content octets portion of the value when treated as an ASN.1 OCTET STRING.
+
+For example, if the initially decrypted content was the byte series `{ 0x04, 0x03, 0x01, 0x02, 0x03 }` or `{ 0x04, 0x03, 0x01, 0x02, 0x03, [continued content] }`, the value of `envelopedCms.ContentInfo.Content` was the byte series `{ 0x01, 0x02, 0x03 }`.
+
+Values that did not start with 0x04, or started with 0x04 but were not followed by an acceptably-encoded length value, were fully reported.
```
gewarren, line 18:
```suggestion
Values that did not start with `0x04`, or started with `0x04` but were not followed by an acceptably encoded length value, were fully reported.
```

PR #29627, opened 2022-05-25
Clarify version requirements

### LINKED ISSUE #29604: Technically correct but misleading claim about MultiSubnetFailover in Framework ADONet docs
**Issue description**

The page [SqlClient Support for High Availability, Disaster Recovery](https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql/sqlclient-support-for-high-availability-disaster-recovery) ([Source](https://github.com/dotnet/docs/blob/main/docs/framework/data/adonet/sql/sqlclient-support-for-high-availability-disaster-recovery.md)) contains the following statement:

> Setting MultiSubnetFailover=True isn't required with .NET Framework 4.6.1 or later versions.

While technically correct , it is potentially misleading, in that it only applies to .NET Framework, and not .NET Core or .NET 5+. The feature that makes this true on .NET Framework 4.6.1 (TransparentNetworkIPResolution) is not present in Core or .NET 5.

While this is technically a docs area for .NET Framework only, it is largely still applicable to .NET Core and .NET 5+, albeit ideally with users using Microsoft.Data.SqlClient, rather than System.Data.SqlClient. This is unsuprising since it remains an evolotuion of the same cvodebase, albeit with differences in the lower levels, since the C++ portions of the driver were rewritten to C# for .NET Core/.NET 5+.

It also is one of the top search results for MultiSubnetFailover.

See also dotnet/SqlClient#1097 (a bug report where the confusing nature of this wording was pointed out) and MicrosoftDocs/azure-docs#93400 (A bug about another copy of this same wording in a different doc collection).

**Target framework**

- [X] .NET Core
- [X] .NET Framework
- [ ] .NET Standard

gewarren:
Fixes #29604.

### DIFF
--- a/docs/framework/data/adonet/sql/sqlclient-support-for-high-availability-disaster-recovery.md
+++ b/docs/framework/data/adonet/sql/sqlclient-support-for-high-availability-disaster-recovery.md
@@ -6,11 +6,11 @@ ms.assetid: 61e0b396-09d7-4e13-9711-7dcbcbd103a0
 ---
 # SqlClient Support for High Availability, Disaster Recovery
 
-This topic discusses SqlClient support (added in .NET Framework 4.5) for high-availability, disaster recovery with the Always On features -- Always On availability groups (AGs) and Always On failover cluster instances (FCIs) with SQL Server 2012 or later.  For more information about either Always On feature, see SQL Server Books Online.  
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

PR #29642, opened 2022-05-26
clarify how type can declare new static members

### LINKED ISSUE #29471: Static members of an `interface` can also be re-implemented
in this [section](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/types/interfaces):

> "To implement an interface member, the corresponding member of the implementing class <ins>**must**</ins> be public, <ins>**non-static**</ins>, and have the same name and signature as the interface member."

but I can implement an static member in an interface and re-implement it in a class.
**`public`**  is also **not mandatory.**

**Example:**
```csharp
   
    interface IZ
    {
        static private string _name = "";
        public static string Name(string name)
        {
            _name = name;
            return _name;
        }
    }

    class StaticReImp : IZ
    {
        private static string _name = "";
        public static string Name(string name)
        {
            _name = name + " <<<static member can be reimplement>>>";
            return _name;
        }
    }
```
I ran it successfully on C#10 - .NET 7.

thanks in advance.

---
#### Document Details

* ID: d9fd340a-0f74-8602-28a1-9569296a5951
* Version Independent ID: 1f0a9a2b-6495-370d-620f-a6dabec7df3b
* Content: [Interfaces - define behavior for multiple types](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/types/interfaces)
* Content Source: [docs/csharp/fundamentals/types/interfaces.md](https://github.com/dotnet/docs/blob/main/docs/csharp/fundamentals/types/interfaces.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
Fixes #29471

When a type declares a static member with the same signature as an interface it implements, it's not an override. Nor does it hide the method. Instead, it's a unique method declared in a different type.

### DIFF
--- a/docs/csharp/fundamentals/types/interfaces.md
+++ b/docs/csharp/fundamentals/types/interfaces.md
@@ -26,7 +26,10 @@ For more information about abstract classes, see [Abstract and Sealed Classes an
 
 Interfaces can contain instance methods, properties, events, indexers, or any combination of those four member types. Interfaces may contain static constructors, fields, constants, or operators. Beginning with C# 11, interface members that aren't fields may be `static abstract`. An interface can't contain instance fields, instance constructors, or finalizers. Interface members are public by default, and you can explicitly specify accessibility modifiers, such as `public`, `protected`, `internal`, `private`, `protected internal`, or `private protected`. A `private` member must have a default implementation.
 
-To implement an interface member, the corresponding member of the implementing class must be public, non-static, and have the same name and signature as the interface member. When an interface declares static members, a type implementing that interface may also declare static members with the same signature. Those are distinct and uniquely identified by the type declaring the member. The static member declared in a type *doesn't* override the static member declared in the interface.
+To implement an interface member, the corresponding member of the implementing class must be public, non-static, and have the same name and signature as the interface member.
+
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


docs/csharp/fundamentals/types/interfaces.md
```diff
@@ -26,7 +26,7 @@ For more information about abstract classes, see [Abstract and Sealed Classes an
...
 
-To implement an interface member, the corresponding member of the implementing class must be public, non-static, and have the same name and signature as the interface member.
+To implement an interface member, the corresponding member of the implementing class must be public, non-static, and have the same name and signature as the interface member. When an interface declares static members, a type implementing that interface may also declare static members with the same signature. Those are distinct and uniquely identified by the type declaring the member. The static member declared in a type *doesn't* override the static member declared in the interface.
```
gewarren, line 29: Would it be better to put it on a note block or separate paragraph? If you're just learning about interfaces, this additional info is a little separate from the concept of implementing interface members.

gewarren (approved):
Just one question/suggestion.

docs/csharp/fundamentals/types/interfaces.md
BillWagner: That's a great suggestion!

PR #29698, opened 2022-06-02
Questionable word: "alignment"

OlegAlexander:
## Summary

I found the word "alignment" questionable here. I think the author meant to say ".NET-style specifier" instead because we're talking about format specifiers in this section. The next section, _Aligning expressions in interpolated strings_, talks about alignment. Of course, I could be wrong. Thank you.


KathleenDollard (approved):
Thanks so much for improving our docs by helping us get these important things right!

PR #29704, opened 2022-06-02
Add no response action

adegeo:
## Summary

I looked into the no response bot and it seems it's been deprecated for a year. They suggested using this action.

@gewarren 

Do we need to add auth for this action in the github settings?

How do we remove the no-response bot app?

### DIFF
--- a/.github/workflows/no-response.yml
+++ b/.github/workflows/no-response.yml
@@ -6,7 +6,7 @@ on:
   issue_comment:
     types: [created]
   schedule:
-    # Schedule for five minutes after the hour, every hour
+    # Schedule every day at 00:05
     - cron: '5 0 * * *'
 
 jobs:
--- a/.github/workflows/no-response.yml
+++ b/.github/workflows/no-response.yml
@@ -7,7 +7,7 @@ on:
     types: [created]
   schedule:
     # Schedule for five minutes after the hour, every hour
-    - cron: '5 * * * *'
+    - cron: '5 0 * * *'
 
 jobs:
   noResponse:
--- a/.github/workflows/no-response.yml
+++ b/.github/workflows/no-response.yml
@@ -1,0 +1,23 @@
+name: No Response
+
+# Both `issue_comment` and `scheduled` event types are required for this Action
+# to work properly.
+on:
+  issue_comment:
+    types: [created]
+  schedule:
+    # Schedule for five minutes after the hour, every hour
+    - cron: '5 * * * *'
+
+jobs:
+  noResponse:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/no-response.yml
```diff
@@ -0,0 +1,23 @@
+name: No Response
+
+# Both `issue_comment` and `scheduled` event types are required for this Action
+# to work properly.
+on:
+  issue_comment:
+    types: [created]
+  schedule:
+    # Schedule for five minutes after the hour, every hour
+    - cron: '5 * * * *'
```
BillWagner, line 10: Wouldn't it be sufficient to run once a day?

```suggestion
    - cron: '5 0 * * *'
```

BillWagner (approved):
This LGTM @adegeo 

My only suggestion is to run it once a day, instead of every hour.

gewarren:
It's too bad that the action in lives in a private organization. Is there not something in the dotnet org we can use?

adegeo:
@gewarren We can fork it and reference it?

.github/workflows/no-response.yml
```diff
@@ -0,0 +1,23 @@
+name: No Response
+
+# Both `issue_comment` and `scheduled` event types are required for this Action
+# to work properly.
+on:
+  issue_comment:
+    types: [created]
+  schedule:
+    # Schedule for five minutes after the hour, every hour
```
Youssef1313, line 9:
```suggestion
    # Schedule every day at 00:05
```

IEvangelist:
I'm curious, how exactly does this work? If someone posts an issue, and we don't touch it for 14 days - does it get closed? Ideally, that wouldn't be the case. How does it know when we engage, is that the `issue_comment` part of the trigger? This seems a bit incorrect as is, as it would have no way of knowing if a comment was from a bot and not one of the team members.

.github/workflows/no-response.yml
```diff
@@ -0,0 +1,23 @@
...
+    - cron: '5 0 * * *'
+
+jobs:
+  noResponse:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: lee-dohm/no-response@9bb0a4b5e6a45046f00353d5de7d90fb8bd773bb
+        with:
+          token: ${{ github.token }}
+          responseRequiredLabel: needs-more-info
```
Youssef1313, line 19: @IEvangelist I think it works on this label.

adegeo: Yeah, it basically works just like the existing bot does. The bot app that was doing this for us points to this action as its replacement.

adegeo:
@IEvangelist The issue comment trigger is only used for detecting when the issue author actually responds after the "needs more info" label was added. The label is removed and if the issue was closed, it's reopened.

PR #29708, opened 2022-06-03
Update right-to-left regex search section

gewarren:
Follow-up to https://github.com/dotnet/dotnet-api-docs/pull/8116.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options?branch=pr-en-us-29708#right-to-left-mode).

### DIFF
--- a/docs/standard/base-types/regular-expression-options.md
+++ b/docs/standard/base-types/regular-expression-options.md
@@ -15,7 +15,7 @@ helpviewer_keywords:
 
 # Regular expression options
 
-By default, the comparison of an input string with any literal characters in a regular expression pattern is case sensitive, white space in a regular expression pattern is interpreted as literal white-space characters, and capturing groups in a regular expression are named implicitly as well as explicitly. You can modify these and several other aspects of default regular expression behavior by specifying regular expression options. Most of these options, which are listed in the following table, can be included inline as part of the regular expression pattern, or they can be supplied to a <xref:System.Text.RegularExpressions.Regex?displayProperty=nameWithType> class constructor or static pattern matching method as a <xref:System.Text.RegularExpressions.RegexOptions?displayProperty=nameWithType> enumeration value.
+By default, the comparison of an input string with any literal characters in a regular expression pattern is case sensitive, white space in a regular expression pattern is interpreted as literal white-space characters, and capturing groups in a regular expression are named implicitly as well as explicitly. You can modify these and several other aspects of default regular expression behavior by specifying regular expression options. Some of these options, which are listed in the following table, can be included inline as part of the regular expression pattern, or they can be supplied to a <xref:System.Text.RegularExpressions.Regex?displayProperty=nameWithType> class constructor or static pattern matching method as a <xref:System.Text.RegularExpressions.RegexOptions?displayProperty=nameWithType> enumeration value.
 
 |RegexOptions member|Inline character|Effect|
 |-------------------------|----------------------|------------|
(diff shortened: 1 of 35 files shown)


BillWagner (approved):
This LGTM @gewarren 

Once you add a project file to address the build issues, you can :shipit:

docs/standard/base-types/regular-expression-options.md
```diff
@@ -11,12 +11,11 @@ helpviewer_keywords:
...
+By default, the comparison of an input string with any literal characters in a regular expression pattern is case sensitive, white space in a regular expression pattern is interpreted as literal white-space characters, and capturing groups in a regular expression are named implicitly as well as explicitly. You can modify these and several other aspects of default regular expression behavior by specifying regular expression options. Most of these options, which are listed in the following table, can be included inline as part of the regular expression pattern, or they can be supplied to a <xref:System.Text.RegularExpressions.Regex?displayProperty=nameWithType> class constructor or static pattern matching method as a <xref:System.Text.RegularExpressions.RegexOptions?displayProperty=nameWithType> enumeration value.
```
stephentoub, line 18:
```suggestion
By default, the comparison of an input string with any literal characters in a regular expression pattern is case sensitive, white space in a regular expression pattern is interpreted as literal white-space characters, and capturing groups in a regular expression are named implicitly as well as explicitly. You can modify these and several other aspects of default regular expression behavior by specifying regular expression options. Some of these options, which are listed in the following table, can be included inline as part of the regular expression pattern, or they can be supplied to a <xref:System.Text.RegularExpressions.Regex?displayProperty=nameWithType> class constructor or static pattern matching method as a <xref:System.Text.RegularExpressions.RegexOptions?displayProperty=nameWithType> enumeration value.
```

docs/standard/base-types/regular-expression-options.md
```diff
@@ -11,12 +11,11 @@ helpviewer_keywords:
   - ".NET regular expressions, options"
   - "inline option constructs"
   - "options parameter"
-ms.assetid: c82dc689-7e82-4767-a18d-cd24ce5f05e9
 ---
 
-# Regular Expression Options
+# Regular expression options
```
stephentoub, line 16: The below table is also missing the new RegexOptions.NonBacktracking being added in .NET 7.  Presumably we should add that as part of a separate PR.

docs/standard/base-types/regular-expression-options.md
```diff
@@ -283,19 +282,27 @@ The following example uses the inline option `(?x)` to ignore pattern white spac
...
 
-The <xref:System.Text.RegularExpressions.RegexOptions.RightToLeft?displayProperty=nameWithType> option changes the search direction only; it does not interpret the regular expression pattern from right to left. For example, the regular expression `\bb\w+\s` matches words that begin with the letter "b" and are followed by a white-space character. In the following example, the input string consists of three words that include one or more "b" characters. The first word begins with "b", the second ends with "b", and the third includes two "b" characters in the middle of the word. As the output from the example shows, only the first word matches the regular expression pattern.
+### Example
+
+The regular expression `\bb\w+\s` matches words that begin with the letter "b" and are followed by a white-space character. In the following example, the input string consists of three words that include one or more "b" characters. The first and second words begin with "b" and the third word end with "b". As the output from the right-to-left search example shows, only the first and second words match the regular expression pattern, with the second word being matched first.
```
stephentoub, line 294: > matches words that begin with the letter "b" and are followed by a white-space character

As written, the pattern also requires at least two characters in the word.  If the intention is that it can include a word consisting only of "b", the "\w+" should instead be "\w*".  If the intention is that words do need to be at least two characters, then it's fine.

stephentoub, line 294:
```suggestion
The regular expression `\bb\w+\s` matches words that begin with the letter "b" and are followed by a white-space character. In the following example, the input string consists of three words that include one or more "b" characters. The first and second words begin with "b" and the third word ends with "b". As the output from the right-to-left search example shows, only the first and second words match the regular expression pattern, with the second word being matched first.
```

stephentoub (commented):
Thanks.

gewarren:
> Thanks.

Thank you for the review @stephentoub!

docs/standard/base-types/regular-expression-options.md
gewarren, line 294:
```suggestion
The regular expression `\bb\w+\s` matches words with two or more characters that begin with the letter "b" and are followed by a white-space character. In the following example, the input string consists of three words that include one or more "b" characters. The first and second words begin with "b" and the third word ends with "b". As the output from the right-to-left search example shows, only the first and second words match the regular expression pattern, with the second word being matched first.
```

docs/standard/base-types/regular-expression-options.md
```diff
@@ -283,19 +282,27 @@ The following example uses the inline option `(?x)` to ignore pattern white spac
...
+
+### Lookaheads
+
+The lookahead assertion (the `(?=`*subexpression*`)` language element) and the lookbehind assertion (the `(?<=`*subexpression*`)` language element) do not change direction. The lookahead assertions look to the right; the lookbehind assertions look to the left. For example, the regular expression `(?<=\d{1,2}\s)\w+,?\s\d{4}` uses the lookbehind assertion to test for a date that precedes a month name. The regular expression then matches the month and the year. For information on lookahead and lookbehind assertions, see [Grouping Constructs](grouping-constructs-in-regular-expressions.md).
```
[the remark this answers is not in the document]
gewarren: This lookahead/lookbehind section is in the context of RTL searches, so when it says "do not change direction", I think it's just referring to the fact that a lookahead is still to the right and a lookbehind is still to the left, even in a RTL search.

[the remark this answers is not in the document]
stephentoub: Ah, ok.

PR #29711, opened 2022-06-03
Fix typo

siriak:


BillWagner (approved):
Thanks for spotting this and submitting the fix @siriak

You should see the change on our live site in the next day or two, on our regular publishing cycle. 

Thanks again for helping us improve! 🎆

PR #29716, opened 2022-06-03
Update 'scl read' example to use quotes command

HolisticDeveloper:
## Summary

After the last code changes in the tutorial, the reader is instructed to run `scl read --file nofile` to see the new custom error message `File does not exist`. However, running this as is also shows the errors `Required command was not provided.` and `Unrecognized command or argument 'read'.` By changing the command to `scl quotes read --file nofile` the extra errors are eliminated.


tdykstra (approved):
Thanks @HolisticDeveloper !

PR #29777, opened 2022-06-09
Update statement for .NET 7

Youssef1313:


tdykstra (commented):
Although there is a .NET 7 Preview button, clicking it on this tutorial just gives you the same tutorial as the .NET 6 version; there are no .NET 7 changes in the tutorial. So I'd rather not add language here that  implies there is a new .NET 7 version.

PR #29785, opened 2022-06-09
Add example of concatenated interpolated strings

### LINKED ISSUE #29440: Breaking-up interpolated strings into pieces
I failed to find on the page (and [this one](https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/string-interpolation) too) the answer to the following question.

If the interpolated string is very long and I want to break it up into multiple pieces (each having an interpolation expression), then does each part have to be prefixed with `$`?
Example: Original interpolated string `$"{a}{b}"`.
Which break-up is right, 
this `$"{a}" + "{b}"` or
this `$"{a}" + $"{b}"`?

---
#### Document Details

* ID: d0665571-35b7-0841-5719-b69456b35cfb
* Version Independent ID: 0410420b-3b12-6d96-40cf-56f22ac19a1d
* Content: [$ - string interpolation - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated)
* Content Source: [docs/csharp/language-reference/tokens/interpolated.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/tokens/interpolated.md)
* Product: **dotnet-csharp**
* GitHub Login: @pkulikov
* Microsoft Alias: **wiwagn**

SeanKilleen:
## Summary

Adds a description of how to break up interpolated strings, based on feedback in #29440.

Fixes #29440.

### DIFF
--- a/docs/csharp/language-reference/tokens/interpolated.md
+++ b/docs/csharp/language-reference/tokens/interpolated.md
@@ -94,6 +94,20 @@ The following example uses implicit conversion to <xref:System.FormattableString
 
 :::code language="csharp" source="./snippets/string-interpolation.cs" id="Snippet4":::
 
+## Combining multiple interpolated strings
+
+To combine multiple interpolated strings, use the interpolation expression for each:
+
+```csharp
+var firstWord = "Hello";
+var secondWord = "World";
+
+var combinedInterpolatedString = $"{firstWord}, " + $"{secondWord}!";
+
+Console.WriteLine(combinedInterpolatedString)
+// output: "Hello, World!"
+```
+
 ## Other resources
 
 If you're new to string interpolation, see the [String interpolation in C#](../../tutorials/exploration/interpolated-strings.yml) interactive tutorial. You can also check another [String interpolation in C#](../../tutorials/string-interpolation.md) tutorial. That tutorial demonstrates how to use interpolated strings to produce formatted strings.


docs/csharp/language-reference/tokens/interpolated.md
```diff
@@ -94,6 +94,20 @@ The following example uses implicit conversion to <xref:System.FormattableString
 
 :::code language="csharp" source="./snippets/string-interpolation.cs" id="Snippet4":::
 
+## Combining multiple interpolated strings
+
+To combine multiple interpolated strings, use the interpolation expression for each:
+
+```csharp
```
SeanKilleen, line 101: ℹ️ I'm happy to move this to a sample code file if desired; just seemed like overkill in this instance.

IEvangelist (approved):

docs/csharp/language-reference/tokens/interpolated.md
IEvangelist: This is fine as is, I agree it's rather simple.

docs/csharp/language-reference/tokens/interpolated.md
```diff
@@ -94,6 +94,20 @@ The following example uses implicit conversion to <xref:System.FormattableString
...
+var firstWord = "Hello";
+var secondWord = "World";
+
+var combinedInterpolatedString = $"{firstWord}, " + $"{secondWord}!";
```
gewarren, line 105: Why wouldn't this just be:

```
var combinedInterpolatedString = $"{firstWord}, {secondWord}!";
```

@BillWagner

BillWagner: That would work, but wouldn't show the example as concatenating two interpolated strings. Admittedly, in this case, a single interpolated string would be clear. This would normally be done when all the source expressions were more complex.

SeanKilleen: @gewarren it could be, for sure. I was using an example provided in the initial issue, which discussed concatenating multiple strings together specifically. So it may be that in the context of the example, the usefulness is lost.

Perhaps I could add something like "bear in mind, you could also accomplish this via" and use the syntax you've provided above?

docs/csharp/language-reference/tokens/interpolated.md
```diff
@@ -94,6 +94,20 @@ The following example uses implicit conversion to <xref:System.FormattableString
 
 :::code language="csharp" source="./snippets/string-interpolation.cs" id="Snippet4":::
 
+## Combining multiple interpolated strings
+
+To combine multiple interpolated strings, use the interpolation expression for each:
```
gewarren, line 99: I'd change this to:

```
To combine multiple interpolated strings, add the `$` special character to each string literal.
```

PR #29789, opened 2022-06-09
Add example for range indices from arrays

SeanKilleen:
## Summary

Adds an explanation and example of range indices from arrays, and notes that taking a range from an array creates a new array with a copy of the array's data.

Fixes #28329


IEvangelist (approved):
Again, I think this addition is valuable. Let's get @BillWagner's thoughts.

BillWagner (approved):
Thanks for making this fix @SeanKilleen 

I'll :shipit: now.

PR #29792, opened 2022-06-09
Documenting Native AOT deployment

### LINKED ISSUE #29780: Create a Native AOT deployment document content
**Help us make content visible**

NativeAOT, Native AOT, AOT

**Describe the new article**

.NET 7 has a preview release of [Native AOT form factor](https://github.com/dotnet/designs/blob/main/accepted/2020/form-factors.md#native-aot-form-factors) and need documentation for users. It should be under [dotnet->core->deployment](https://docs.microsoft.com/en-us/dotnet/core/deploying). The content at minimum should describe the form factor at high level, ways to generate a native aot application, and scope. The article should follow the pattern of other form factors on that section (`Trimming`, `SingleFile`), and should summarize from somewhat technical content already available in our repos ([developer overview](https://github.com/dotnet/runtime/blob/main/docs/workflow/building/coreclr/nativeaot.md), [source location readme links](https://github.com/dotnet/runtime/tree/main/src/coreclr/nativeaot))

LakshanF:
## Summary

Created a starting content for NativeAOT deployment model to get feedback

Fixes #29780

[Rendered view](https://github.com/LakshanF/docs/blob/NativeAotDocs/docs/core/deploying/native-aot.md)

### DIFF
--- a/docs/core/deploying/native-aot.md
+++ b/docs/core/deploying/native-aot.md
@@ -62,7 +62,6 @@ The app will be available in the publish directory and will contain all the code
 
 ## Impact of using the Native AOT deployment
 
-Native AOT applications are [self-contained](index.md#publish-self-contained) and don't use any shared libraries, as is the case with the default [framework-dependent deployment model](index.md#publish-framework-dependent). Consider the native AOT deployment model when the benefits to deployment and the ability to optimize a single app outweigh the benefits of sharing for certain types of workloads. The gain is most significant for workloads with a high number of deployed instances: cloud infrastructure, hyper-scale services, popular apps, or games.
 
 Native AOT applications comes with a few fundamental limitations and compatibility issues. The key limitations include:
 
--- a/docs/core/deploying/native-aot.md
+++ b/docs/core/deploying/native-aot.md
@@ -73,7 +73,7 @@ Native AOT applications comes with a few fundamental limitations and compatibili
 
 The first release of native AOT in .NET 7 has additional limitations. These include:
 
-- Should be targeted for console type applications.
+- Should be targeted for console type applications (not ASP.NET Core).
 - Not all the runtime libraries are fully annotated to be native AOT compatible (that is, some warnings in the runtime libraries are not actionable by end developers).
 - Limited diagnostic support (debugging and profiling).
 
--- a/docs/core/deploying/native-aot.md
+++ b/docs/core/deploying/native-aot.md
@@ -34,7 +34,7 @@ Publish a native AOT application using the [dotnet publish](../tools/dotnet-publ
 
 02. Ensure pre-requisites that are needed for publishing a native AOT application are in the machine.
 
-    For Windows pre-requisites, install [Visual Studio 2022](https://visualstudio.microsoft.com/vs/), including Desktop development with C++ workload.
+    On Windows, install [Visual Studio 2022](https://visualstudio.microsoft.com/vs/), including Desktop development with C++ workload.
 
     On Linux, install clang and developer packages for libraries that .NET runtime depends on.
 
--- a/docs/core/deploying/native-aot.md
+++ b/docs/core/deploying/native-aot.md
@@ -48,7 +48,7 @@ Publish a native AOT application using the [dotnet publish](../tools/dotnet-publ
     sudo apk add clang gcc lld musl-dev build-base zlib-dev
     ```
 
-03. Publish the app for a specific runtime identifier using `dotnet publish -r <RID>`. 
+03. Publish the app for a specific runtime identifier using `dotnet publish -r <RID>`.
 
     The following example publishes the app for Windows as a native AOT application on a machine with the required pre-requisites installed.
 
--- a/docs/core/deploying/native-aot.md
+++ b/docs/core/deploying/native-aot.md
@@ -7,7 +7,7 @@ ms.date: 06/09/2022
 ---
 # Native AOT Deployment
 
-Publishing your app as *native AOT* produces an app that is [*self-contained*](index.md#publish-self-contained) and that has been ahead-of-time (AOT) compiled to native code. Native AOT apps startup very quickly and use less memory. Users of the application can run it on a machine that doesn't have the .NET runtime installed.
+Publishing your app as *native AOT* produces an app that is [*self-contained*](index.md#publish-self-contained) and that has been ahead-of-time (AOT) compiled to native code. Native AOT apps start up very quickly and use less memory. Users of the application can run it on a machine that doesn't have the .NET runtime installed.
 
 The native AOT deployment model uses an IL to native compiler. Native AOT apps don't use a Just-In-Time (JIT) compiler when the application runs. Native AOT apps can run in restricted environments where a JIT is not allowed. Native AOT applications target a specific runtime environment, such as Linux x64 or Windows x64, just like publishing a [self-contained app](index.md#publish-self-contained).
 
--- a/docs/core/deploying/native-aot.md
+++ b/docs/core/deploying/native-aot.md
@@ -30,27 +30,31 @@ Publish a native AOT application using the [dotnet publish](../tools/dotnet-publ
     </PropertyGroup>
     ```
 
-02. Publish the app for for a specific runtime identifier using `dotnet publish -r <RID>`. There are pre-requisites that are needed to be installed for publishing a native AOT application. Some examples are given below for each platform.
+02. Ensure pre-requisites that are needed for publishing a native AOT application are in the machine.
 
-    For Windows pre-requisites, install [Visual Studio 2022](https://visualstudio.microsoft.com/vs/community/), including Desktop development with C++ workload. The following example publishes the app for Windows as a native AOT application on a machine with the required pre-requisites installed.
-
-    `dotnet publish -r win-x64 -c Release`
+    For Windows pre-requisites, install [Visual Studio 2022](https://visualstudio.microsoft.com/vs/), including Desktop development with C++ workload.
 
     For Linux pre-requisites, install clang and developer packages for libraries that .NET runtime depends on.
 
-    *Ubuntu (16.04+)*
+    *Ubuntu (18.04+)*
 
     ```sh
     sudo apt-get install clang zlib1g-dev
     ```
 
-    *Alpine (3.14+)*
+    *Alpine (3.15+)*
 
     ```sh
     sudo apk add clang gcc lld musl-dev build-base zlib-dev
     ```
 
-    The following example publishes the app for Linux as a native AOT application. A native AOT binary produced on Linux machine is only going to work on same or newer Linux version than the build machine Linux version. For example, native AOT binary produced on Ubuntu 20.04 is going to run on Ubuntu 22.04, but it is not going to run on Ubuntu 16.04.
+03. Publish the app for a specific runtime identifier using `dotnet publish -r <RID>`. 
+
+    The following example publishes the app for Windows as a native AOT application on a machine with the required pre-requisites installed.
+
+    `dotnet publish -r win-x64 -c Release`
+
+    The following example publishes the app for Linux as a native AOT application. A native AOT binary produced on Linux machine is only going to work on same or newer Linux version. For example, native AOT binary produced on Ubuntu 20.04 is going to run on Ubuntu 20.04 and later, but it is not going to run on Ubuntu 18.04.
 
     `dotnet publish -r linux-arm64 -c Release`
 
@@ -65,7 +69,7 @@ Native AOT applications comes with a few fundamental limitations and compatibili
 - No dynamic loading (for example, `Assembly.LoadFile`)
 - No runtime code generation (for example, `System.Reflection.Emit`)
 - No C++/CLI
-- No built-in COM
+- No built-in COM (only applies to Windows)
 - Requires trimming, which has [limitations](trimming/incompatibilities.md)
 
 The first release of native AOT in .NET 7 has additional limitations. These include:
@@ -80,5 +84,5 @@ The following table shows supported compilation targets when targeting .NET 7.
 
 | Platform | Supported architecture |
 | ------------ | --------------------------- |
-| Windows  | X64, ARM64 |
-| Linux    | X64, ARM64 |
+| Windows  | x64, Arm64 |
+| Linux    | x64, Arm64 |
--- a/docs/core/deploying/native-aot.md
+++ b/docs/core/deploying/native-aot.md
@@ -16,7 +16,7 @@ There are some limitations in the .NET NativeAOT deployment model, with the main
 
 ## Publish NativeAOT - CLI
 
-Publish a NativeAOT application using the [dotnet publish](../../tools/dotnet-publish.md) command.
+Publish a NativeAOT application using the [dotnet publish](../tools/dotnet-publish.md) command.
 
 01. Add `<PublishAot>true</PublishAot>` to your project file.
 
@@ -40,7 +40,7 @@ Publish a NativeAOT application using the [dotnet publish](../../tools/dotnet-pu
 
 The app will be available in the publish directory and will contain all the code needed to run in it, including the coreclr runtime.
 
-Native libraries that can be consumed by other programming languages can be published using the `NativeLib=Static` property for static libraries that can be linked at compile time. Shared libraries that are required at runtime can be created at publish time by using the `NativeLib=Shared` property.
+NativeAOT libraries that can be consumed by other programming languages can be published using the `NativeLib=Static` property for static libraries that can be linked at compile time. Shared libraries that are required at runtime can be created at publish time by using the `NativeLib=Shared` property.
 
 ## Impact of using the NativeAOT deployment
 
@@ -51,7 +51,7 @@ NativeAOT applications comes with a few fundamental limitations and compatibilit
 - No dynamic loading (for example, `Assembly.LoadFile`)
 - No runtime code generation (for example, `System.Reflection.Emit`)
 - No C++/CLI
-- no built-in COM
+- No built-in COM
 - Requires trimming, which has [limitations](trimming/incompatibilities.md)
 
 The first release of NativeAOT in .NET 7 is experimental and has additional limitations. These include:
--- a/docs/core/deploying/native-aot.md
+++ b/docs/core/deploying/native-aot.md
@@ -36,7 +36,7 @@ Publish a native AOT application using the [dotnet publish](../tools/dotnet-publ
 
     For Windows pre-requisites, install [Visual Studio 2022](https://visualstudio.microsoft.com/vs/), including Desktop development with C++ workload.
 
-    For Linux pre-requisites, install clang and developer packages for libraries that .NET runtime depends on.
+    On Linux, install clang and developer packages for libraries that .NET runtime depends on.
 
     *Ubuntu (18.04+)*
 
--- a/docs/core/deploying/native-aot.md
+++ b/docs/core/deploying/native-aot.md
@@ -9,7 +9,7 @@ ms.date: 06/09/2022
 
 .NET NativeAOT deployment model uses an optimized ahead-of-time (AOT) compiler that generates an application that contains native executable code. NativeAOT applications do not use the just-in-time (JIT) compiler when the application runs. There are performance benefits (startup time, working set improvements etc.) and the ability to run .NET applications in restricted environment where a JIT is not allowed with using NativeAOT applications. NativeAOT applications have to be [self-contained applications](index.md#publish-self-contained) that targets a specific runtime environment (RID) such as Linux x64 or Windows x64.
 
-There are some limitations in .NET NativeAOT deployment model, the main one being, that no runtime code generation is possible. See [Impact of using the NativeAOT deployment](#impact-of-using-the-nativeaot-deployment) for additional details. The support in .NET 7 release is targeted towards console type applications. 
+There are some limitations in .NET NativeAOT deployment model, the main one being, that no runtime code generation is possible. See [Impact of using the NativeAOT deployment](#impact-of-using-the-nativeaot-deployment) for additional details. The support in .NET 7 release is targeted towards console type applications.
 
 > [!WARNING]
 > NativeAOT deployment is an experimental feature in .NET 7.
@@ -45,15 +45,17 @@ The app should be available in the publish directory and will contain all the co
 NativeAOT applications are [self-contained application](index.md#publish-self-contained) and do not use any shared libraries as is the case with default [framework-dependent deployment model](index.md#publish-framework-dependent). NativeAot deployment model should be considered when the benefits to deployment and the ability to optimize a single app outweigh the benefits of sharing for certain types of workloads. The gain is most significant for workloads with a high number of deployed instances: cloud infrastructure, hyper-scale services, popular apps or games.
 
 NativeAOT applications comes with a few fundamental limitations and compatibility issues. The key limitations include:
- - No dynamic loading (e.g. Assembly.LoadFile)
- - No runtime code generation (e.g. System.Reflection.Emit)
- - No C++/CLI, no built-in COM and WinRT interop support
- - No unconstrained reflection
+
+- No dynamic loading (e.g. Assembly.LoadFile)
+- No runtime code generation (e.g. System.Reflection.Emit)
+- No C++/CLI, no built-in COM and WinRT interop support
+- No unconstrained reflection
 
 The first release of NativeAOT in .NET 7 is experimental and has additional limitations. These include:
- - Should be targeted for console type applications
- - Not all the runtime libraries are fully annotated to be NativeAOT compatible (i.e., some warnings in the runtime libraries are not actionable by end developers)
- - No full diagnostic support (debugging and profiling)
+
+- Should be targeted for console type applications
+- Not all the runtime libraries are fully annotated to be NativeAOT compatible (i.e., some warnings in the runtime libraries are not actionable by end developers)
+- No full diagnostic support (debugging and profiling)
 
 ## Platform/architecture restrictions
 
--- a/docs/core/deploying/native-aot.md
+++ b/docs/core/deploying/native-aot.md
@@ -1,26 +1,28 @@
 ---
-title: NativeAOT deployment overview
-description: Learn what NativeAOT deployments are and why you should consider using it as part of the publishing your app with .NET 7 and later.
+title: Native AOT deployment overview
+description: Learn what native AOT deployments are and why you should consider using it as part of the publishing your app with .NET 7 and later.
 author: lakshanf
 ms.author: lakshanf
 ms.date: 06/09/2022
 ---
-# NativeAOT Deployment
+# Native AOT Deployment
 
-The .NET NativeAOT deployment model uses an optimized ahead-of-time (AOT) compiler that generates an application that contains native executable code. NativeAOT apps don't use the just-in-time (JIT) compiler when the application runs. Native AOT provides performance benefits (for example, startup time and working set improvements) and the ability to run .NET applications in restricted environments where a JIT is not allowed. NativeAOT applications are [self-contained applications](index.md#publish-self-contained) that target a specific runtime environment (RID), such as Linux x64 or Windows x64.
+Publishing your app as *native AOT* produces an app that is [*self-contained*](index.md#publish-self-contained) and that has been ahead-of-time (AOT) compiled to native code. Native AOT apps startup very quickly and use less memory. Users of the application can run it on a machine that doesn't have the .NET runtime installed.
 
-There are some limitations in the .NET NativeAOT deployment model, with the main one being that no run-time code generation is possible. For more information, see [Impact of using the NativeAOT deployment](#impact-of-using-the-nativeaot-deployment). The support in the .NET 7 release is targeted towards console-type applications.
+The native AOT deployment model uses an IL to native compiler. Native AOT apps don't use a Just-In-Time (JIT) compiler when the application runs. Native AOT apps can run in restricted environments where a JIT is not allowed. Native AOT applications target a specific runtime environment, such as Linux x64 or Windows x64, just like publishing a [self-contained app](index.md#publish-self-contained).
+
+There are some limitations in the .NET native AOT deployment model, with the main one being that run-time code generation is not possible. For more information, see [Impact of using the Native AOT deployment](#impact-of-using-the-native-aot-deployment). The support in the .NET 7 release is targeted towards console-type applications.
 
 > [!WARNING]
-> NativeAOT deployment is an experimental feature in .NET 7.
+> Native AOT is supported in .NET 7 but only a limited number of libraries are fully compatible with native AOT in .NET 7.
 
-## Publish NativeAOT - CLI
+## Publish Native AOT - CLI
 
-Publish a NativeAOT application using the [dotnet publish](../tools/dotnet-publish.md) command.
+Publish a native AOT application using the [dotnet publish](../tools/dotnet-publish.md) command.
 
 01. Add `<PublishAot>true</PublishAot>` to your project file.
 
-    This will produce a NativeAOT app and shows PublishAot compatibility warnings during build.
+    This will produce a native AOT app and shows PublishAot compatibility warnings during build.
 
     ```xml
     <PropertyGroup>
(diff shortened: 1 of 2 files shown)


docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,67 @@
+---
+title: NativeAOT deployment overview
+description: Learn what NativeAOT deployments are and why you should consider using it as part of the publishing your app with .NET 7 and later.
+author: lakshanf
+ms.author: lakshanf
+ms.date: 06/09/2022
+---
+# NativeAOT Deployment
```
MichalStrehovsky, line 8: @samsp-msft Are we going to officially use "Native AOT" or "NativeAOT"?

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,67 @@
...
+
+.NET NativeAOT deployment model uses an optimized ahead-of-time (AOT) compiler that generates an application that contains native executable code. NativeAOT applications do not use the just-in-time (JIT) compiler when the application runs. There are performance benefits (startup time, working set improvements etc.) and the ability to run .NET applications in restricted environment where a JIT is not allowed with using NativeAOT applications. NativeAOT applications have to be [self-contained applications](index.md#publish-self-contained) that targets a specific runtime environment (RID) such as Linux x64 or Windows x64.
+
+There are some limitations in .NET NativeAOT deployment model, the main one being, that no runtime code generation is possible. See [Impact of using the NativeAOT deployment](#impact-of-using-the-nativeaot-deployment) for additional details. The support in .NET 7 release is targeted towards console type applications.
+
+> [!WARNING]
+> NativeAOT deployment is an experimental feature in .NET 7.
```
MichalStrehovsky, line 15: Do we still call it experimental or is it a preview?

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,67 @@
...
+
+There are two ways to publish your app as NativeAOT:
+
+01. Specify the PublishAot flag directly to the dotnet publish command. See [dotnet publish](../tools/dotnet-publish.md) for details.
+
+    ```dotnetcli
+    dotnet publish -c Release -r win-x64 -p:PublishAot=true
+    ```
```
MichalStrehovsky, line 23: Let's not talk about this one.

PublishAot sets PublishTrimmed to true. We decided we don't want to advertise passing it on the command line because it bypasses analyzers and other things that emulate trimmed behavior in `dotnet run`.

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Impact of using the NativeAOT deployment
+
+NativeAOT applications are [self-contained application](index.md#publish-self-contained) and do not use any shared libraries as is the case with default [framework-dependent deployment model](index.md#publish-framework-dependent). NativeAot deployment model should be considered when the benefits to deployment and the ability to optimize a single app outweigh the benefits of sharing for certain types of workloads. The gain is most significant for workloads with a high number of deployed instances: cloud infrastructure, hyper-scale services, popular apps or games.
+
+NativeAOT applications comes with a few fundamental limitations and compatibility issues. The key limitations include:
+
+- No dynamic loading (e.g. Assembly.LoadFile)
+- No runtime code generation (e.g. System.Reflection.Emit)
+- No C++/CLI, no built-in COM and WinRT interop support
```
MichalStrehovsky, line 51:
```suggestion
- No C++/CLI
- no built-in COM
```

WinRT interop support was removed from CoreCLR in 6.0 and we don't have to worry about not having it in NativeAOT.

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,67 @@
...
+- No unconstrained reflection
```
MichalStrehovsky, line 52:
```suggestion
```

Maybe let's mention it requires trimming and comes with all the issues that trimming has. It's not a limitation of nativeAOT per-se.

docs/core/deploying/native-aot.md
LakshanF: I'm ok with either and we have NativeAOT in our source (e.g. https://github.com/dotnet/runtime/pull/68950)

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,66 @@
...
+
+    The following example publishes the app for Linux as a NativeAOT application.
+
+    `dotnet publish -r linux-arm64`
+
+The app should be available in the publish directory and will contain all the code needed to run in it including the coreclr runtime.
+
+## Impact of using the NativeAOT deployment
+
+NativeAOT applications are [self-contained application](index.md#publish-self-contained) and do not use any shared libraries as is the case with default [framework-dependent deployment model](index.md#publish-framework-dependent). NativeAot deployment model should be considered when the benefits to deployment and the ability to optimize a single app outweigh the benefits of sharing for certain types of workloads. The gain is most significant for workloads with a high number of deployed instances: cloud infrastructure, hyper-scale services, popular apps or games.
```
samsp-msft, line 43:
```suggestion
NativeAOT applications are [self-contained](index.md#publish-self-contained) and do not use any shared libraries as is the case with the default [framework-dependent deployment model](index.md#publish-framework-dependent). NativeAOT deployment model should be considered when the benefits to deployment and the ability to optimize a single app outweigh the benefits of sharing for certain types of workloads. The gain is most significant for workloads with a high number of deployed instances: cloud infrastructure, hyper-scale services, popular apps or games.
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,66 @@
...
+- No runtime code generation (e.g. System.Reflection.Emit)
+- No C++/CLI
+- no built-in COM
+- Requires trimming and should be aware of trimming [limitations](trimming/incompatibilities.md)
+
+The first release of NativeAOT in .NET 7 is experimental and has additional limitations. These include:
+
+- Should be targeted for console type applications
+- Not all the runtime libraries are fully annotated to be NativeAOT compatible (i.e., some warnings in the runtime libraries are not actionable by end developers)
+- No full diagnostic support (debugging and profiling)
```
samsp-msft, line 57:
```suggestion
- Limited diagnostic support (debugging and profiling)
```

samsp-msft (approved):
A couple of minor linguistic tweaks

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,66 @@
+---
+title: NativeAOT deployment overview
+description: Learn what NativeAOT deployments are and why you should consider using it as part of the publishing your app with .NET 7 and later.
+author: lakshanf
+ms.author: lakshanf
+ms.date: 06/09/2022
+---
+# NativeAOT Deployment
+
+.NET NativeAOT deployment model uses an optimized ahead-of-time (AOT) compiler that generates an application that contains native executable code. NativeAOT applications do not use the just-in-time (JIT) compiler when the application runs. There are performance benefits (startup time, working set improvements etc.) and the ability to run .NET applications in restricted environment where a JIT is not allowed with using NativeAOT applications. NativeAOT applications have to be [self-contained applications](index.md#publish-self-contained) that targets a specific runtime environment (RID) such as Linux x64 or Windows x64.
```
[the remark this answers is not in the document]
gewarren:
```suggestion
The .NET NativeAOT deployment model uses an optimized ahead-of-time (AOT) compiler that generates an application that contains native executable code. NativeAOT apps don't use the just-in-time (JIT) compiler when the application runs. Native AOT provides performance benefits (for example, startup time and working set improvements) and the ability to run .NET applications in restricted environments where a JIT is not allowed. NativeAOT applications are [self-contained applications](index.md#publish-self-contained) that target a specific runtime environment (RID), such as Linux x64 or Windows x64.
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,66 @@
...
+
+There are some limitations in .NET NativeAOT deployment model, the main one being, that no runtime code generation is possible. See [Impact of using the NativeAOT deployment](#impact-of-using-the-nativeaot-deployment) for additional details. The support in .NET 7 release is targeted towards console type applications.
```
gewarren, line 12:
```suggestion
There are some limitations in the .NET NativeAOT deployment model, with the main one being that no run-time code generation is possible. For more information, see [Impact of using the NativeAOT deployment](#impact-of-using-the-nativeaot-deployment). The support in the .NET 7 release is targeted towards console-type applications.
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,66 @@
...
+
+    The following example publishes the app for Windows as a NativeAOT application.
+
+    `dotnet publish -r win-x64`
+
+    The following example publishes the app for Linux as a NativeAOT application.
+
+    `dotnet publish -r linux-arm64`
+
+The app should be available in the publish directory and will contain all the code needed to run in it including the coreclr runtime.
```
gewarren, line 39:
```suggestion
The app will be available in the publish directory and will contain all the code needed to run in it, including the coreclr runtime.
```

docs/core/deploying/native-aot.md
gewarren:
```suggestion
NativeAOT applications are [self-contained](index.md#publish-self-contained) and don't use any shared libraries, as is the case with the default [framework-dependent deployment model](index.md#publish-framework-dependent). Consider the NativeAOT deployment model when the benefits to deployment and the ability to optimize a single app outweigh the benefits of sharing for certain types of workloads. The gain is most significant for workloads with a high number of deployed instances: cloud infrastructure, hyper-scale services, popular apps, or games.
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Impact of using the NativeAOT deployment
+
+NativeAOT applications are [self-contained application](index.md#publish-self-contained) and do not use any shared libraries as is the case with default [framework-dependent deployment model](index.md#publish-framework-dependent). NativeAot deployment model should be considered when the benefits to deployment and the ability to optimize a single app outweigh the benefits of sharing for certain types of workloads. The gain is most significant for workloads with a high number of deployed instances: cloud infrastructure, hyper-scale services, popular apps or games.
+
+NativeAOT applications comes with a few fundamental limitations and compatibility issues. The key limitations include:
+
+- No dynamic loading (e.g. Assembly.LoadFile)
```
gewarren, line 47:
```suggestion
- No dynamic loading (for example, `Assembly.LoadFile`)
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,66 @@
...
+- No runtime code generation (e.g. System.Reflection.Emit)
```
gewarren, line 48:
```suggestion
- No run-time code generation (for example, `System.Reflection.Emit`)
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,66 @@
...
+- No C++/CLI
+- no built-in COM
```
gewarren, line 50:
```suggestion
- No built-in COM
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,66 @@
...
+- Requires trimming and should be aware of trimming [limitations](trimming/incompatibilities.md)
```
gewarren, line 51:
```suggestion
- Requires trimming, which has [limitations](trimming/incompatibilities.md)
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,66 @@
...
+
+The first release of NativeAOT in .NET 7 is experimental and has additional limitations. These include:
+
+- Should be targeted for console type applications
```
gewarren, line 55:
```suggestion
- Should be targeted for console type applications.
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,66 @@
...
+- Not all the runtime libraries are fully annotated to be NativeAOT compatible (i.e., some warnings in the runtime libraries are not actionable by end developers)
```
gewarren, line 56:
```suggestion
- Not all the runtime libraries are fully annotated to be NativeAOT compatible (that is, some warnings in the runtime libraries are not actionable by end developers).
```

docs/core/deploying/native-aot.md
gewarren:
```suggestion
- Limited diagnostic support (debugging and profiling).
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,66 @@
...
+- No full diagnostic support (debugging and profiling)
+
+## Platform/architecture restrictions
+
+Supported compilation targets are described in the table below when targeting .NET 7.
```
gewarren, line 61:
```suggestion
The following table shows supported compilation targets when targeting .NET 7.
```

gewarren (commented):
I left a few additional suggestions, incorporating Sam's into mine.

docs/core/deploying/native-aot.md
LakshanF: Thanks!

docs/core/deploying/native-aot.md
LakshanF: Thanks!

docs/core/deploying/native-aot.md
kant2002: NativeAOT IMO better for googling things. I would immidiately get only relevant results.

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,70 @@
...
+
+    `dotnet publish -r win-x64 -c Release`
+
+    The following example publishes the app for Linux as a NativeAOT application.
+
+    `dotnet publish -r linux-arm64 -c Release`
+
+The app will be available in the publish directory and will contain all the code needed to run in it, including the coreclr runtime.
+
+NativeAOT libraries that can be consumed by other programming languages can be published using the `NativeLib=Static` property for static libraries that can be linked at compile time. Shared libraries that are required at runtime can be created at publish time by using the `NativeLib=Shared` property.
```
kant2002, line 43: Based on my experience Shared library mode is break implicit expectations of dev community. A lot of developers would not realize that their native code would have CLR running inside their app and second DLL compiled with NativeAOT in same process would be problematic. That should be highlighted since a lot of people want develop in C# and use artifacts in unmanaged world.

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,70 @@
...
+- No runtime code generation (for example, `System.Reflection.Emit`)
+- No C++/CLI
+- No built-in COM
+- Requires trimming, which has [limitations](trimming/incompatibilities.md)
+
+The first release of NativeAOT in .NET 7 is experimental and has additional limitations. These include:
+
+- Should be targeted for console type applications.
+- Not all the runtime libraries are fully annotated to be NativeAOT compatible (that is, some warnings in the runtime libraries are not actionable by end developers).
+- Limited diagnostic support (debugging and profiling).
```
kant2002, line 61: I would somehow emphasize that Reflection DO WORKING with NativeAOT. Common theme among developers is that NativeAOT working without reflection. Maybe worth saying that NativeAOT has limited support for reflection. You can inspect types, invoke members but do not load assemblies at runtime.

docs/core/deploying/native-aot.md
MichalStrehovsky: > and second DLL compiled with NativeAOT in same process would be problematic

What kind of problems did you run into with that setup?

It's not expected to be more problematic than any other "there are two automatic memory managers in the process that don't know about each other" situation. The other library doesn't have to be NativeAOT-compiled.

NativeAOT doesn't have the problem as CoreCLR that there can only be one of those in the process.

docs/core/deploying/native-aot.md
MichalStrehovsky: @richlander @jkotas Do we care about how Native AOT/NativeAOT is spelled in docs? I recall there was some conversation about it but I can't find it now. The Preview 3 blog talks about "Native AOT".

docs/core/deploying/native-aot.md
jkotas: NativeAOT is not experimental feature for .NET 7. We should say that it is supported, but only limited number of libraries are fully compatible with NativeAOT in .NET 7.

docs/core/deploying/native-aot.md
jkotas, line 43: The static libraries are feature with a rough-edges. I would omit it from the official docs. For .NET 7, we are only going to support NativeAOT for apps.

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,70 @@
...
+
+    ```xml
+    <PropertyGroup>
+        <PublishAot>true</PublishAot>
+    </PropertyGroup>
+    ```
+
+01. Publish the app as for a specific runtime identifier using `dotnet publish -r <RID>`
+
+    The following example publishes the app for Windows as a NativeAOT application.
```
jkotas, line 33: This should talk about pre-requisites you need to install for the compilation.

(It should be simplified version of https://github.com/dotnet/runtime/blob/main/src/coreclr/nativeaot/docs/prerequisites.md.)

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,70 @@
...
+
+    `dotnet publish -r win-x64 -c Release`
+
+    The following example publishes the app for Linux as a NativeAOT application.
+
+    `dotnet publish -r linux-arm64 -c Release`
```
jkotas, line 39: We may want to mention that the NativeAOT binary produced on Linux machine is only going to work on same or newer Linux version than the build machine Linux version. For example, NativeAOT binary produced on Ubuntu 20.04 is going to run on Ubuntu 22.04, but it is not going to run on Ubuntu 16.04.

docs/core/deploying/native-aot.md
kant2002: I have only limited exposure to this setup and in very ìnteresting` circumstances. I try to create custom GC and it chokes because some resources was shared between NativeAOT GC initialization phase and CoreCLR Gc initialization phase. If you can have 2 runtimes and 2 GC within same process then I my remark would not be applicable. I may misinterpret things.

MichalStrehovsky: This is getting really off topic but if you're referring to this setup: https://twitter.com/KooKiz/status/1533133858312445956, the problem there is about trying to do `[UnmanagedCallersOnly(Entrypoint="GC_Initialize")]`. It won't link because Native AOT uses that symbol name. One can run into the same problem with `[UnmanagedCallersOnly(Entrypoint="printf")]`. The solution is: "don't do that". The symbolic name is gone after the linker runs and doesn't pose problems for a "finished" library.

kant2002: Yeah. You are right. I did try something like what's in that tweet about year ago, but did not manage to finish experiment  properly.

docs/core/deploying/native-aot.md
richlander: I think we should avoid this "ProjectN" style name. We should call it "Native AOT". We want to avoid the idea that we're creating a new brand or something similar. R2R is bad enough (on this topic). The more we can of this capability as a new build output, the better. I also don't think it always has to be capitalized, like "native AOT" should be fine within a sentence. It's "... native AOT differs from R2R since R2R is managed-first and purely an optimization on the normal functioning of the CoreCLR managed runtime."

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,70 @@
...
+description: Learn what NativeAOT deployments are and why you should consider using it as part of the publishing your app with .NET 7 and later.
+author: lakshanf
+ms.author: lakshanf
+ms.date: 06/09/2022
+---
+# NativeAOT Deployment
+
+The .NET NativeAOT deployment model uses an optimized ahead-of-time (AOT) compiler that generates an application that contains native executable code. NativeAOT apps don't use the just-in-time (JIT) compiler when the application runs. Native AOT provides performance benefits (for example, startup time and working set improvements) and the ability to run .NET applications in restricted environments where a JIT is not allowed. NativeAOT applications are [self-contained applications](index.md#publish-self-contained) that target a specific runtime environment (RID), such as Linux x64 or Windows x64.
+
+There are some limitations in the .NET NativeAOT deployment model, with the main one being that no run-time code generation is possible. For more information, see [Impact of using the NativeAOT deployment](#impact-of-using-the-nativeaot-deployment). The support in the .NET 7 release is targeted towards console-type applications.
```
richlander, line 12:
```suggestion
There are some limitations in the .NET NativeAOT deployment model, with the main one being that run-time code generation is not possible. For more information, see [Impact of using the NativeAOT deployment](#impact-of-using-the-nativeaot-deployment). The support in the .NET 7 release is targeted towards console-type applications.
```

docs/core/deploying/native-aot.md
richlander, line 43: This content shouldn't show up here. We should walk users through the basic flow and not mention any advanced scenarios as part of that.

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,70 @@
...
+
+    This will produce a native AOT app and shows PublishAot compatibility warnings during build.
+
+    ```xml
+    <PropertyGroup>
+        <PublishAot>true</PublishAot>
+    </PropertyGroup>
+    ```
+
+02. Publish the app for for a specific runtime identifier using `dotnet publish -r <RID>`. There are pre-requisites that are needed to be installed for publishing a native AOT application. For Windows, install [Visual Studio 2022](https://visualstudio.microsoft.com/vs/community/), including Desktop development with C++ workload. For Linux, install clang and developer packages for libraries that .NET runtime depends on.
```
jkotas, line 33: > For Linux, install clang and developer packages for libraries that .NET runtime depends on.

I think this needs more details. We should have the command line to install these pre-requisites on Debian and Alpine at least. (Debian and Alpine is what we are publishing dotnet containers for.)

docs/core/deploying/native-aot.md
LakshanF: Removed references to advanced scenarios

docs/core/deploying/native-aot.md
LakshanF: I couldn't find the Debian pre-requisites and included Ubuntu and Alpine from the link above

docs/core/deploying/native-aot.md
LakshanF: I'm not sure if affirming reflection working is a good idea, given that there are some restrictions. The doc mentions requiring trimming, which implies reflection support. @MichalStrehovsky, do you have any views on this?

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,84 @@
...
+
+    This will produce a native AOT app and shows PublishAot compatibility warnings during build.
+
+    ```xml
+    <PropertyGroup>
+        <PublishAot>true</PublishAot>
+    </PropertyGroup>
+    ```
+
+02. Publish the app for for a specific runtime identifier using `dotnet publish -r <RID>`. There are pre-requisites that are needed to be installed for publishing a native AOT application. Some examples are given below for each platform.
```
richlander, line 33:
```suggestion
02. Publish the app for a specific runtime identifier using `dotnet publish -r <RID>`. There are pre-requisites that are needed to be installed for publishing a native AOT application. Some examples are given below for each platform.
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,84 @@
...
+
+    For Windows pre-requisites, install [Visual Studio 2022](https://visualstudio.microsoft.com/vs/community/), including Desktop development with C++ workload. The following example publishes the app for Windows as a native AOT application on a machine with the required pre-requisites installed.
```
richlander, line 35: We should just link to the overall VS page, not the community specific one. We don't know which VS version the reader uses.

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,84 @@
...
+
+    `dotnet publish -r win-x64 -c Release`
+
+    For Linux pre-requisites, install clang and developer packages for libraries that .NET runtime depends on.
+
+    *Ubuntu (16.04+)*
```
richlander, line 41:
```suggestion
    *Ubuntu (18.04+)*
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,84 @@
...
+
+    ```sh
+    sudo apt-get install clang zlib1g-dev
+    ```
+
+    *Alpine (3.14+)*
```
richlander, line 47:
```suggestion
    *Alpine (3.15+)*
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,84 @@
...
+
+    ```sh
+    sudo apk add clang gcc lld musl-dev build-base zlib-dev
+    ```
+
+    The following example publishes the app for Linux as a native AOT application. A native AOT binary produced on Linux machine is only going to work on same or newer Linux version than the build machine Linux version. For example, native AOT binary produced on Ubuntu 20.04 is going to run on Ubuntu 22.04, but it is not going to run on Ubuntu 16.04.
```
richlander, line 53:
```suggestion
    The following example publishes the app for Linux as a native AOT application. A native AOT binary produced on Linux machine is only going to work on same or newer Linux version. For example, native AOT binary produced on Ubuntu 20.04 is going to run on Ubuntu 20.04 and later, but it is not going to run on Ubuntu 18.04.
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,84 @@
...
+
+    `dotnet publish -r linux-arm64 -c Release`
+
+The app will be available in the publish directory and will contain all the code needed to run in it, including the coreclr runtime.
```
richlander, line 57: > including the coreclr runtime

I was under the impression that native AOT apps did not include the CoreCLR runtime.

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Impact of using the Native AOT deployment
+
+Native AOT applications are [self-contained](index.md#publish-self-contained) and don't use any shared libraries, as is the case with the default [framework-dependent deployment model](index.md#publish-framework-dependent). Consider the native AOT deployment model when the benefits to deployment and the ability to optimize a single app outweigh the benefits of sharing for certain types of workloads. The gain is most significant for workloads with a high number of deployed instances: cloud infrastructure, hyper-scale services, popular apps, or games.
```
richlander, line 61: This paragraph doesn't communicate anything obvious to me. Perhaps just delete it?

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,84 @@
...
+
+Native AOT applications comes with a few fundamental limitations and compatibility issues. The key limitations include:
+
+- No dynamic loading (for example, `Assembly.LoadFile`)
+- No runtime code generation (for example, `System.Reflection.Emit`)
+- No C++/CLI
+- No built-in COM
```
richlander, line 68:
```suggestion
- No built-in COM (only applies to Windows)
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,84 @@
...
+- Requires trimming, which has [limitations](trimming/incompatibilities.md)
+
+The first release of native AOT in .NET 7 has additional limitations. These include:
+
+- Should be targeted for console type applications.
```
richlander, line 73: > console type applications

We should be more specific. ASP.NET Core apps are console apps.

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,84 @@
...
+- Limited diagnostic support (debugging and profiling).
+
+## Platform/architecture restrictions
+
+The following table shows supported compilation targets when targeting .NET 7.
+
+| Platform | Supported architecture |
+| ------------ | --------------------------- |
+| Windows  | X64, ARM64 |
+| Linux    | X64, ARM64 |
```
richlander, line 84:
```suggestion
| Windows  | x64, Arm64 |
| Linux    | x64, Arm64 |
```

docs/core/deploying/native-aot.md
LakshanF: Thanks!

docs/core/deploying/native-aot.md
LakshanF: Thanks!

docs/core/deploying/native-aot.md
LakshanF: It contains a stripped down version, https://github.com/dotnet/runtime/blob/main/docs/workflow/building/coreclr/nativeaot.md

docs/core/deploying/native-aot.md
LakshanF: Wanted to convey the type of applications that are good for native AOT application but can take this out if it's not helpful

docs/core/deploying/native-aot.md
LakshanF: Yes, it is important to exclude the perception that ASP.NET Core apps are possible.

Is there a suitable phrasing to convey this? I was going with VS project templates that seems to restrict to non-ASP.NET apps.

richlander:
It looks like you are not committing the suggestions. Is that intentional or am I misreading the commit history?

docs/core/deploying/native-aot.md
richlander: Right, but it doesn't say what you think it does.

richlander:
I only see commits from you. That was my point: https://github.com/dotnet/docs/pull/29792/commits.

jkotas:
@LakshanF The usual github workflow when people make a bunch of suggestions is to commit the suggestions if you generally agree with them, and apply any additional edits on top of them if necessary.

docs/core/deploying/native-aot.md
LakshanF: Can you suggest the change? That will give me a chance to fix at least 1 change for all the commit snafus I've been making.

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,88 @@
+---
+title: Native AOT deployment overview
+description: Learn what native AOT deployments are and why you should consider using it as part of the publishing your app with .NET 7 and later.
+author: lakshanf
+ms.author: lakshanf
+ms.date: 06/09/2022
+---
+# Native AOT Deployment
+
+Publishing your app as *native AOT* produces an app that is [*self-contained*](index.md#publish-self-contained) and that has been ahead-of-time (AOT) compiled to native code. Native AOT apps startup very quickly and use less memory. Users of the application can run it on a machine that doesn't have the .NET runtime installed.
```
jtschuster, line 10:
```suggestion
Publishing your app as *native AOT* produces an app that is [*self-contained*](index.md#publish-self-contained) and that has been ahead-of-time (AOT) compiled to native code. Native AOT apps start up very quickly and use less memory. Users of the application can run it on a machine that doesn't have the .NET runtime installed.
```
From what I found, "start up" should be two words when used a verb.

docs/core/deploying/native-aot.md
richlander:
```suggestion
The app will be available in the publish directory and will contain all the code needed to run in it, including a stripped-down version of the coreclr runtime.
```

richlander: @LakshanF

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,88 @@
...
+
+    `dotnet publish -r win-x64 -c Release`
+
+    The following example publishes the app for Linux as a native AOT application. A native AOT binary produced on Linux machine is only going to work on same or newer Linux version. For example, native AOT binary produced on Ubuntu 20.04 is going to run on Ubuntu 20.04 and later, but it is not going to run on Ubuntu 18.04.
+
+    `dotnet publish -r linux-arm64 -c Release`
+
+The app will be available in the publish directory and will contain all the code needed to run in it, including the coreclr runtime.
+
+## Impact of using the Native AOT deployment
```
richlander, line 63:
```suggestion
## Limitations of Native AOT deployment
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,88 @@
...
+description: Learn what native AOT deployments are and why you should consider using it as part of the publishing your app with .NET 7 and later.
+author: lakshanf
+ms.author: lakshanf
+ms.date: 06/09/2022
+---
+# Native AOT Deployment
+
+Publishing your app as *native AOT* produces an app that is [*self-contained*](index.md#publish-self-contained) and that has been ahead-of-time (AOT) compiled to native code. Native AOT apps start up very quickly and use less memory. Users of the application can run it on a machine that doesn't have the .NET runtime installed.
+
+The native AOT deployment model uses an IL to native compiler. Native AOT apps don't use a Just-In-Time (JIT) compiler when the application runs. Native AOT apps can run in restricted environments where a JIT is not allowed. Native AOT applications target a specific runtime environment, such as Linux x64 or Windows x64, just like publishing a [self-contained app](index.md#publish-self-contained).
```
richlander, line 12:
```suggestion
The benefit of native AOT is most significant for workloads with a high number of deployed instances, such as cloud infrastructure and hyper-scale services. It is currently not supported with ASP.NET Core, but only console apps.

The native AOT deployment model uses an IL to native compiler. Native AOT apps don't use a Just-In-Time (JIT) compiler when the application runs. Native AOT apps can run in restricted environments where a JIT is not allowed. Native AOT applications target a specific runtime environment, such as Linux x64 or Windows x64, just like publishing a [self-contained app](index.md#publish-self-contained).
```

docs/core/deploying/native-aot.md
richlander:
```suggestion
```

docs/core/deploying/native-aot.md
richlander:
```suggestion
- Should be targeted for console type applications (not ASP.NET Core).
```

docs/core/deploying/native-aot.md
```diff
@@ -0,0 +1,88 @@
...
+
+The app will be available in the publish directory and will contain all the code needed to run in it, including the coreclr runtime.
+
+## Impact of using the Native AOT deployment
+
+Native AOT applications are [self-contained](index.md#publish-self-contained) and don't use any shared libraries, as is the case with the default [framework-dependent deployment model](index.md#publish-framework-dependent). Consider the native AOT deployment model when the benefits to deployment and the ability to optimize a single app outweigh the benefits of sharing for certain types of workloads. The gain is most significant for workloads with a high number of deployed instances: cloud infrastructure, hyper-scale services, popular apps, or games.
+
+Native AOT applications comes with a few fundamental limitations and compatibility issues. The key limitations include:
+
+- No dynamic loading (for example, `Assembly.LoadFile`)
```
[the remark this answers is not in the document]
LakshanF: I'm a little concerned with this message as is - native AOT will trim unused code and will only include the code that the app requires.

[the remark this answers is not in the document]
richlander: They will still be bigger than FDD apps, right?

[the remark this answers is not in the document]
LakshanF: I think SCD-FDD terminology might confuse with Native AOT with regards to talking about the size as well as the implication that the coreclr runtime is a separate entity. Native AOT is really a stand alone native binary (the closest would be to say its a trimmed SCD single file application with stripped down version of the runtime embedded in it but that would be confusing as well).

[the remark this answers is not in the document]
richlander: I don't think that's the right frame. I grasp that people thinking working on Native AOT think of it as a separate universe. In order for us to think about it as supported and for it to be integrated into this doc-set, we need to think of it as another point on the spectrum of deployment choices.

The spectrum looks like this:

portable FDD -> RID-specific FDD -> self-contained -> single file -> native AOT

> Native AOT is really a stand alone native binary (the closest would be to say its a trimmed SCD single file application with stripped down version of the runtime embedded in it but that would be confusing as well).

That's thinking of it all as a series of tech choices. In reality, native aot sits naturally at the point I put it at in that spectrum.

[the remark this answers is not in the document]
LakshanF: I like your framing native aot in terms of deployment choices like this but worry that native aot apps including the runtime libraries as a size limitation might not be clear to developers. The choices are really between shared vs non-shared library code but given that the doc briefly mentions the benefits of native aot apps at the start, I'm ok with FDD-SCD phrasing.

If you can think of a way to word this better, please do so. Otherwise I'll commit the suggestion above.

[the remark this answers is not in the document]
richlander: > , as compared to framework-dependent apps

I added that. Does that help?

We could also add it last in the list. That might help.

Issue #26214, opened 2021-09-21
Add logging configuration example for ASP.NET Core


lmolkova:
I'd love to see here how to configure logging with ASP.NET Core and Microsoft.Extensions.Azure including code and `appsettings.json` content - I can't find it being documented anywhere

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 337e8ed0-9a24-93ce-c7cf-478c0e24dc06
* Version Independent ID: 2151de27-a308-a249-277d-7addb659775e
* Content: [Logging with the Azure SDK for .NET](https://docs.microsoft.com/en-us/dotnet/azure/sdk/logging)
* Content Source: [docs/azure/sdk/logging.md](https://github.com/dotnet/docs/blob/main/docs/azure/sdk/logging.md)
* Product: **dotnet-azure**
* GitHub Login: @CamSoper
* Microsoft Alias: **casoper**

lmolkova:
resolved in https://github.com/dotnet/docs/pull/26698, thank you!

PR #29798, opened 2022-06-10
Document logging with AddAzureClientsCore

scottaddie:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/29205
Related to https://github.com/dotnet/docs/issues/26214

Officially document the approach taken here to enable log forwarding: https://github.com/bradygaster/FileNewBlazorServerToAzureContainerApps/pull/1

[Internal review page](https://review.docs.microsoft.com/dotnet/azure/sdk/logging?branch=pr-en-us-29798)

### DIFF
--- a/docs/azure/sdk/logging.md
+++ b/docs/azure/sdk/logging.md
@@ -9,7 +9,7 @@ author: camsoper
 
 # Logging with the Azure SDK for .NET
 
-The [Azure SDK](https://azure.microsoft.com/downloads/) for .NET client libraries include the ability to log client library operations. This logging allows you to monitor I/O requests and responses that client libraries are making to Azure services. Typically, the logs are used to debug or diagnose communication issues. This article describes the following approaches to enable logging with the Azure SDK for .NET:
+The Azure SDK for .NET's client libraries include the ability to log client library operations. This logging allows you to monitor I/O requests and responses that client libraries are making to Azure services. Typically, the logs are used to debug or diagnose communication issues. This article describes the following approaches to enable logging with the Azure SDK for .NET:
 
 - [Enable logging with built-in methods](#enable-logging-with-built-in-methods)
 - [Configure custom logging](#configure-custom-logging)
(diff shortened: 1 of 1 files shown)


JoshLove-msft:
> Officially document the approach taken here to enable log forwarding: [bradygaster/FileNewBlazorServerToAzureContainerApps#1](https://github.com/bradygaster/FileNewBlazorServerToAzureContainerApps/pull/1)

I'm curious why we don't register the clients using AddAzureClients in this app. We are instead constructing them by hand in Program.cs instead of allowing the DI methods to do their thing.

IEvangelist (approved):

docs/azure/sdk/logging.md
```diff
@@ -92,7 +98,7 @@ using AzureEventSourceListener listener = new AzureEventSourceListener((e, messa
 
 ## Map to ASP.NET Core logging
 
-When the <xref:Microsoft.Extensions.Azure.AzureClientServiceCollectionExtensions.AddAzureClients%2A> extension method is called, the <xref:Microsoft.Extensions.Azure.AzureEventSourceLogForwarder> service is registered. The `AzureEventSourceLogForwarder` service enables you to use the standard ASP.NET Core logging configuration for logging.
+When the <xref:Microsoft.Extensions.Azure.AzureClientServiceCollectionExtensions.AddAzureClients%2A> or the <xref:Microsoft.Extensions.Azure.AzureClientServiceCollectionExtensions.AddAzureClientsCore%2A> extension method is called, the <xref:Microsoft.Extensions.Azure.AzureEventSourceLogForwarder> service is registered. The `AzureEventSourceLogForwarder` service enables you to use the standard ASP.NET Core logging configuration for logging.
```
jsquire, line 101: I'm on the fence here with whether to mention `AddAzureClientsCore` explicitly or just call out that `AzureEventSourceLogForwarder` can be registered as a singleton if, for some odd reason, you prefer not to register your SDK clients via `AddAzureClients` - which is the recommended approach.

Thoughts?

jsquire:
> I'm curious why we don't register the clients using AddAzureClients in this app. We are instead constructing them by hand in Program.cs instead of allowing the DI methods to do their thing.

There aren't any explicit clients created/used.   The clients for that app are implicit to the Data Protection extensions for Blobs/KeyVault.  For some reason, these do not implicitly enable log forwarding.

docs/azure/sdk/logging.md
JoshLove-msft: I don't have a strong opinion on AddAzureClientsCore vs registering AzureEventSourceLogForwarder, but I think it is a good idea to try to call out that it is unusual to use either of these two approaches as opposed to just using AddAzureClients.

jsquire: .... unless you're not actually using explicit clients.   😄    _(which is an uncommon corner case, admittedly)_

JoshLove-msft: Yeah, I guess we can provide an example where someone might do this, e.g. your app doesn't actually use any Azure SDK clients but it uses some other package that uses the Azure SDK under the hood, such as the data protection extensions.

bradygaster:
Being transparent - I think the *real* solution here is that we either add methods to the data protection API to dial up to the Azure SDK, or we add code to the Azure SDK clients "stuff" that enables data protection (and we make it obvious and do obvious things when developers call it - it should be a turnkey method, in a sense). The way it is now, the methods in the doc actually call to the data plane SDKs when they're executed. That would mean someone would have to build the builder before the explicit Build is called. That's - meh.

docs/azure/sdk/logging.md
scottaddie: For as long as `AddAzureClientsCore` remains on the public API surface, we need to document its purpose. I like Josh's recommendation of providing an example. I'll add this.

jsquire (approved):

JoshLove-msft (approved):

PR #29812, opened 2022-06-11
Typos and clarifications

OlegAlexander:
## Summary

Regarding lightweight syntax, I believe this sentence is outdated because both the lightweight and the verbose syntax are always on by default. And I believe it is necessary to use indentation for a function definition in both the lightweight and the verbose syntax.

Regarding "The result is again 202." This result isn't mentioned yet. It's mentioned later in the Function Composition section.

Thank you.


BillWagner (approved):
Thanks again for helping us improve @OlegAlexander 

Issue #29336, opened 2022-05-09
When a type loaded by System.Reflection.MetadataLoadContext "is" another type


Forgind:
**Help us make content visible**

I was looking at https://github.com/dotnet/docs/blob/main/docs/standard/assembly/inspect-contents-using-metadataloadcontext.md

The issue is that I'm unclear on how to check if a type "is" another type. For instance, let's say Foo extends Bar. If I have Foo f, loaded from an assembly inspected using MLC, and I check `if (f is Bar)`, it returned false. If I loaded the type Bar from the same MLC that Foo was loaded from, it still returned false.

My more specific issue was when I had a class from the MLC-loaded assembly with properties, and I tried to inspect the type of those properties. The workaround I found was walking from that type to the base type (eventually object) and checking at each stage if its name matched the one I was looking for, but that's messier.

[Edit to add document details]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6d4a1a18-d470-6a2e-30d3-8f89a82aaaf8
* Version Independent ID: f1e8d3e4-e92e-f8e4-85fd-666101347b2f
* Content: [How to: Inspect assembly contents using MetadataLoadContext](https://docs.microsoft.com/en-us/dotnet/standard/assembly/inspect-contents-using-metadataloadcontext)
* Content Source: [docs/standard/assembly/inspect-contents-using-metadataloadcontext.md](https://github.com/dotnet/docs/blob/main/docs/standard/assembly/inspect-contents-using-metadataloadcontext.md)
* Product: **dotnet-fundamentals**
* GitHub Login: @MSDN-WhiteKnight
* Microsoft Alias: **dapine**

svick:
> If I have Foo f, loaded from an assembly inspected using MLC

This should be impossible. If you load an assembly using the MLC, then you can't have an instance of a type from that assembly.

> The workaround I found was walking from that type to the base type (eventually object) and checking at each stage if its name matched the one I was looking for, but that's messier.

Why do you need to check the whole inheritance hierarchy? Checking just the name of the type in question should be enough.

MSDN-WhiteKnight:
`is` operator could never work with MetadataLoadContext, it's meaning is baked into language so it could only work with a type defined at compile time. `IsAssignableFrom` works for MLC, but it is not compatible with runtime reflection. If you test using a type loaded in MLC too, it should work:

```
[...]

PR #29831, opened 2022-06-14
Clarify MetadataLoadContext limitations

MSDN-WhiteKnight:
Clarify MetadataLoadContext limitations when testing types for equality or assignability with other types

Related to: https://github.com/dotnet/docs/issues/29336

### DIFF
--- a/docs/standard/assembly/inspect-contents-using-metadataloadcontext.md
+++ b/docs/standard/assembly/inspect-contents-using-metadataloadcontext.md
@@ -33,6 +33,10 @@ The following code sample creates <xref:System.Reflection.MetadataLoadContext>, 
 
 [!code-csharp[](snippets/inspect-contents-using-metadataloadcontext/MetadataLoadContextSnippets.cs#CreateContext)]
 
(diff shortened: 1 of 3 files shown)


docs/standard/assembly/inspect-contents-using-metadataloadcontext.md
```diff
@@ -33,6 +33,10 @@ The following code sample creates <xref:System.Reflection.MetadataLoadContext>,
 
 [!code-csharp[](snippets/inspect-contents-using-metadataloadcontext/MetadataLoadContextSnippets.cs#CreateContext)]
 
+If you need to test types in <xref:System.Reflection.MetadataLoadContext> for equality or assignability, only use type objects loaded into that context. Mixing <xref:System.Reflection.MetadataLoadContext> types with runtime types is not supported. For example, suppose we have a type `testedType` in <xref:System.Reflection.MetadataLoadContext>. If you need to test whether another type is assignable from it, do not use code like `typeof(MyType).IsAssignableFrom(testedType)`. Use code like this instead:
```
gewarren, line 36:
```suggestion
If you need to test types in <xref:System.Reflection.MetadataLoadContext> for equality or assignability, only use type objects loaded into that context. Mixing <xref:System.Reflection.MetadataLoadContext> types with runtime types is not supported. For example, consider a type `testedType` in <xref:System.Reflection.MetadataLoadContext>. If you need to test whether another type is assignable from it, don't use code like `typeof(MyType).IsAssignableFrom(testedType)`. Use code like this instead:
```

IEvangelist (approved):

PR #29903, opened 2022-06-17
Update ide0054-ide0074.md

HugoRoss:
## Summary

Value in the example changed from `1` to `5` as in C# it looks unesthetic to write `x += 1` - one would write `++x` or `x++` instead.

Fixes n/a


gewarren (approved):
Thanks @HugoRoss

PR #29906, opened 2022-06-17
Fixes #27367

### LINKED ISSUE #27367: Provide a reason why implicit typing should be avoided in loop variable in foreach loops
According to C# Coding Conventions, "Don't use implicit typing to determine the type of the loop variable in foreach loops."
But no clear reason is written here, so I don't know why we should use `foreach (char ch in laugh)` instead of `foreach (var ch in laugh)`.
You can find usually easily which type the `var` is, as Intellisense suggests like other `var`s.
```
Dictionary<Foo, Bar> X = new();
// Add some elements to X...
foreach(KeyValuePair<Foo, Bar> Y in X)
{
// Some Process here
}
```
Above might be difficult to guess, but that is what the C# implementation of `Dictionary` is, so not hard to remember.

Is this rule relevant to the note below it("Be careful not to accidentally change a type of an element of the iterable collection...")?

---
#### Document Details

* ID: 40f42506-f4a5-859a-0266-ded002056d83
* Version Independent ID: f940f2ad-e13d-4d37-6439-f8afcbc4e4c7
* Content: [C# Coding Conventions](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions)
* Content Source: [docs/csharp/fundamentals/coding-style/coding-conventions.md](https://github.com/dotnet/docs/blob/main/docs/csharp/fundamentals/coding-style/coding-conventions.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

rafcon-dev:
Added a clarification on why implicit typing on foreach loops is not desirable.

### DIFF
--- a/docs/csharp/fundamentals/coding-style/coding-conventions.md
+++ b/docs/csharp/fundamentals/coding-style/coding-conventions.md
@@ -202,7 +202,7 @@ The following sections describe practices that the C# team follows to prepare co
 
     :::code language="csharp" source="./snippets/coding-conventions/program.cs" id="Snippet7":::
 
-- Don't use implicit typing to determine the type of the loop variable in [`foreach`](../../language-reference/statements/iteration-statements.md#the-foreach-statement) loops.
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks for submitting this fix @rafcon-dev 

rafcon-dev:
That's great @BillWagner , glad I could contribute a bit

PR #29911, opened 2022-06-17
Add links to code style rules

gewarren:
Add links from language reference to related code style rules. If you don't like this, I'm open to feedback!

(Hide white space in diff.)


BillWagner (approved):
This is really great @gewarren 

I really like the way this will drive traffic to the style rules, and ideally drive behavior to adopt a better coding style.

Of the two styles used in this PR, I like the `[!TIP]` style. I think it stands out more. 

Two thumbs up. Way up!  👍 👍 

Let's :shipit:

PR #29921, opened 2022-06-18
Fixes #22697

### LINKED ISSUE #22697: How does this help me overload an operator?
[Enter feedback here]

I'm looking to so some operator overloading, and there is nothing here that's of any help. Even the links were useless. When someone is looking to overload an operator they need two things. The list of which operators can be overloaded (that was provided). Examples on how to override the operators (those are missing).

edit by @billwagner to add blank line (formatting only)

---
#### Document Details

* ID: f91d85ae-8b37-e837-a031-144571362fbb
* Version Independent ID: 5ec5c397-c7c6-700c-bc36-a8dfb2396dad
* Content: [Operator overloading - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/operator-overloading)
* Content Source: [docs/csharp/language-reference/operators/operator-overloading.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/operators/operator-overloading.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

rafcon-dev:
Changed name of section, which seemed to imply that all the following operators were overloadable, which isn't the case. Added note to point the user to check more details on each operator's page, and a reminder to check the example code. I still believe the table text could be less verbose, and maybe move some of it to numbered notes below, as is the case for the binary operators.

Fixes #22697

- edit by @billwagner Add the "fixes" tag for automation and linking.


BillWagner:
Hi 👋  @rafcon-dev 

I don't think this change addresses the issue very well.

I'd recommend a different set of changes:

- Add text for the rules for all overloaded operator definitions. For example, that one of the operands must be the type in which the overloaded operator is defined).
- Split the table into two tables. 
- The first lists only the operators that can be overloaded. The right column can explain any restrictions (like overloading `==` and `!=` as a pair). 
- The second lists only the operators that can't be overloaded. There, the right column explains alternatives to overloading the operator.

Finally, by changing the header, all anchors need to be updated. There are build issues with the current changes.

What do you think?

rafcon-dev:
@BillWagner Yeah, I believe that's a good way of tidying up this page. I'll get to it when I get some time, hopefully in the next couple days!
When you say all anchors need to be updated, you are referring to other pages that could reference this one table right? So I should grep the rest of the docs to get the references to the table and update those, is that it?

BillWagner:
@rafcon-dev 

> When you say all anchors need to be updated, you are referring to other pages that could reference this one table right? So I should grep the rest of the docs to get the references to the table and update those, is that it?

Yes, that's correct. The build report shows only these two anchor:

- https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/operators/operator-overloading.md?plain=1#L14
- https://github.com/dotnet/docs/blob/main/docs/csharp/misc/cs1020.md?plain=1#L15

BillWagner:

pkulikov:
@rafcon-dev I would be glad to update this article as well according to the suggestion of @BillWagner 
please let me know if you're overwhelmed and fine with me continuing this work

rafcon-dev:
@pkulikov . Hey. Actually I have the changes locally, haven't got the time to upload them indeed, but I'll get to it on the next days as I'll have some free time to do so. Cheers, thank you for the heads up.

rafcon-dev:
@BillWagner Let me know if this is more in line with what you had in mind. Ended up overhauling the text making it less verbose. Also, the tables are smaller so I believe the example above should be more visible, I don't believe there's a need to again point the user to it tbh.

BillWagner (approved):
Thanks again @rafcon-dev 

This now LGTM (once I commit the two small suggestions on the title casing)

I have some small concerns about the lack of punctuation or spacing between the different expressions in the tables, but that could be my eyes. Let's :shipit: and watch for any updated feedback.

PR #29957, opened 2022-06-22
Remove repeated content

GitHubPang:
The text is identical to the 3 lines above, so I removed the duplicate.


gewarren (approved):
Thanks @GitHubPang

PR #29964, opened 2022-06-22
Fix up language rules index

gewarren:
Link to the article/rule itself instead of the option names.


BillWagner (approved):
This LGTM @gewarren 

Let's :shipit:

PR #29966, opened 2022-06-22
Update Null-Conditional Operator for Member Access

### LINKED ISSUE #29953: Clarification on Null-Conditional Operator for Member Access
The document for the null-conditional operator for member access should clearly state whether the use of the operator always guarantees that there will never be a `NullReferenceException` w.r.t. to the left-hand operand. The following statement exists in the document but does so only in the context of the `Invoke` method called on delegates:

> The `?.` operator evaluates its left-hand operand no more than once, guaranteeing that it cannot be changed to `null` after being verified as non-null.

If the above statement is true regardless of the type of the left-hand operand then it should be moved higher up in the doc and outside the section that pertains to its use with delegates.

---
#### Document Details

* ID: 0a25abe4-2b4e-119e-85e9-af19ed06e8c0
* Version Independent ID: 682b8c86-e49a-1d6d-8562-c9f919082bdc
* Content: [Member access operators and expressions - C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-)
* Content Source: [docs/csharp/language-reference/operators/member-access-operators.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/operators/member-access-operators.md)
* Product: **dotnet-csharp**
* GitHub Login: @pkulikov
* Microsoft Alias: **wiwagn**

prasadu-microsoft:
Clarification that the null-conditional operator used in member access evaluates the left-hand operand just once regardless of whether it is a delegate type or not.

Related to issue https://github.com/dotnet/docs/issues/29953

## Summary

Reorganizes the documentation to make it clear that the left-hand operand is evaluated once regardless of the type of the operand.

Fixes #29953

### DIFF
--- a/docs/csharp/language-reference/operators/member-access-operators.md
+++ b/docs/csharp/language-reference/operators/member-access-operators.md
@@ -139,6 +139,9 @@ If `a.x` or `a[x]` is of a non-nullable value type `T`, `a?.x` or `a?[x]` is of 
 
 In the preceding example, if you don't use the `??` operator, `numbers?.Length < 2` evaluates to `false` when `numbers` is `null`.
 
+> [!NOTE]
+> The `?.` operator evaluates its left-hand operand no more than once, guaranteeing that it cannot be changed to `null` after being verified as non-null.
+
 The null-conditional member access operator `?.` is also known as the Elvis operator.
 
 ### Thread-safe delegate invocation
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks @prasadu-microsoft 

This LGTM. I'll :shipit: now.

BillWagner:
> @BillWagner - I wonder if this can actually be organized even more clearly in a section that specifically discusses the thread-safety aspect of the null-conditional operator? Do you have any suggestions (feel free to make edits)?

It's not guaranteed to be thread-safe in the general case. It's not an atomic operation. If there are multiple cores, it's possible to change state after the expression is evaluated. The null-case works because evaluating the expression and caching it in a compiler generated temporary makes it reachable. But, it's not thread safe in that state could change between caching the value and invoking the right argument.

prasadu-microsoft:
> > @BillWagner - I wonder if this can actually be organized even more clearly in a section that specifically discusses the thread-safety aspect of the null-conditional operator? Do you have any suggestions (feel free to make edits)?
> 
> It's not guaranteed to be thread-safe in the general case. It's not an atomic operation. If there are multiple cores, it's possible to change state after the expression is evaluated. The null-case works because evaluating the expression and caching it in a compiler generated temporary makes it reachable. But, it's not thread safe in that state could change between caching the value and invoking the right argument.

I mean thread-safe with respect to "non-nullness". 

```csharp
leftHandOperand?.MyMethod();
```

Is the operator usage above equivalent in all cases to the following code regardless of the type of the operand? Otherwise, my doc change may not be a good one:

```csharp
var tempVar = leftHandOperand;
if (tempVar != null)
{
    tempVar.MyMethod();
}
```

BillWagner:
> Is the operator usage above equivalent in all cases to the following code regardless of the type of the operand?

Yes, semantically it is. (The compiler can make some optimizations, but there cannot be any observable differences).

> I mean thread-safe with respect to "non-nullness".

I'm still objecting to the use of "thread-safe". It's about side-effects, not threads. Multiple threads aren't required. A single threaded GUI app using preemptive scheduling could still see the side effects. It's actually a stronger guarantee.

PR #29968, opened 2022-06-22
x86 host path on x64/arm64

### LINKED ISSUE #29740: [Breaking change]: 32-bit .NET host will not be added to PATH on 64-bit Windows platforms
### Description

We are modifying the code in x86 version of .NET Windows Installers, that adds x86 host location to the PATH - Program Files (x86)\dotnet. Going forward, x86 host path will not be added on 64-bit Windows systems.

See pending PR for the actual change in .NET 7: https://github.com/dotnet/runtime/pull/69902

This will be backported, and will affect, .NET 6 and .NET Core 3.1, as well.

### Version

.NET 7 Preview 6

### Previous behavior

X86 host location will be added to PATH, even on x64/arm64 systems. Depending on which .NET architecture installer is run first, user's machine could have either native (x64/arm64) or x86 host as first in the PATH list.

### New behavior

Going forward, x86 host location will only be added to the PATH on x86 systems.

Customers that need x86 host in the PATH on x64/arm64 systems need to add host location to the PATH manually.

This will affect .NET Core 3.1, .NET 6, .NET 7, and future versions.

### Type of breaking change

### Reason for change

Currently, x86 host will be added to PATH, even on x64/arm64 systems. Depending on which .NET architecture installer is run first, user's machine could have either native (x64/arm64) or x86 host as first in the PATH list.

This presents problems with initial .NET installation, and gets further complicated during .NET servicing events. Any of these installation scenarios could modify the order of .NET hosts in PATH.

The experience is non-deterministic, and has a high chance of regressing user's .NET runtime experience.

This change streamlines .NET host experience on Windows 64-bit systems. Only 64-bit hosts will be available in system PATH - x64 host on x64 system and arm64 host on arm64 systems.

As a result, we remove the ambiguity of order of .NET hosts in system PATH - only one host will be in the PATH.

### Recommended action

Customers that need x86 host in the PATH on x64/arm64 systems need to add the host location to the PATH manually.

### Feature area

Deployment

### Affected APIs

_No response_

gewarren:
Fixes #29740

I added this breaking change to multiple versions since it will be backported.

[Internal preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/deployment/7.0/x86-host-path?branch=pr-en-us-29968).


docs/core/compatibility/deployment/7.0/x86-host-path.md
```diff
@@ -0,0 +1,36 @@
...
+---
+# x86 host path on 64-bit Windows
+
+The x86 versions of .NET installers for Windows have been modified to no longer add the x86 host location (*Program Files (x86)\dotnet*) to the `PATH` environment variable on 64-bit Windows systems.
+
+This change affects .NET Core 3.1, .NET 6, .NET 7, and future versions.
+
+## Previous behavior
+
+The x86 host location was added to `PATH`, even on x64/arm64 systems. Depending on which .NET architecture installer was run first, a user's machine could have either the native (x64/arm64) or x86 host listed first in `PATH`.
```
richlander, line 14: Correct casing is "Arm64"

docs/core/compatibility/deployment/7.0/x86-host-path.md
```diff
@@ -0,0 +1,36 @@
...
+
+## New behavior
+
+Going forward, the x86 host location is only added to the `PATH` environment variable on x86 systems.
```
richlander, line 18: It would be good to clarify that this change only affects the `dotnet` host. It doesn't not affect 32-bit/x86 application hosts, like `myapp.exe`. They will continue to find the x86 runtime (assuming it is installed) correctly.

IEvangelist (approved):

PR #29971, opened 2022-06-22
ContentRootPath breaking change

### LINKED ISSUE #29945: Microsoft.Extensions.Hosting ContentRootPath no longer uses CWD when C:\Windows\system32
### Description

[IHostEnvironment.ContentRootPath](https://docs.microsoft.com/dotnet/api/microsoft.extensions.hosting.ihostenvironment.contentrootpath) is the default directory where `appsettings.json` and other content files are loaded in a hosted application, including ASP.NET apps. This value defaults to `Environment.CurrentDirectory`, the current working directory of the application. This allows the same app to be executed under different working directories and use the content from each directory.

When a Windows process (either application or service) is launched without specifying a working directory, the working directory of the process that created it is used. See [CreateProcess](https://docs.microsoft.com/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa). When the process is launched by Windows Shell or `services.exe`, the current directory is set to `%windir%\system32` or the "System" special folder. This is because that is the current directory of the Windows Shell and `services.exe` processes.

This behavior is confusing and causes hosted applications to fail when the ContentRootPath is `%windir%\system32` because the application tries loading files from that directory, which do not exist. For example, the `appsettings.json` file is not found at runtime and the settings aren't applied.

Starting with .NET 7 Preview 6, when a hosted application is launched with the current directory set to the "System" special folder, it defaults the ContentRootPath to [AppContext.BaseDirectory](https://docs.microsoft.com/dotnet/api/system.appcontext.basedirectory).

### Version

.NET 7 Preview 6

### Previous behavior

[Host.CreateDefaultBuilder](https://docs.microsoft.com/dotnet/api/microsoft.extensions.hosting.host.createdefaultbuilder) defaulted the ContentRootPath to `Environment.CurrentDirectory` regardless of the value of the current directory.

### New behavior

[Host.CreateDefaultBuilder](https://docs.microsoft.com/dotnet/api/microsoft.extensions.hosting.host.createdefaultbuilder) will no longer default the ContentRootPath property to the current directory if it is the System special folder on Windows. Instead the base directory of the application is used.

### Type of breaking change

### Reason for change

App developers were not expecting the ContentRootPath to be `C:\Windows\system32` when their application was launched by Windows in certain cases (for example when the app was packaged as an MSIX or started by the Start Menu). In these cases it was better to default the ContentRootPath to be the base directory of the application.

### Recommended action

If you want to use the previous behavior, you can set the ContentRootPath explicitly when creating the `IHostBuilder`:

```C#
Host
    .CreateDefaultBuilder()
    .UseContentRoot(Environment.CurrentDirectory)
    ....
```

### Feature area

Core .NET libraries, Extensions

### Affected APIs

https://docs.microsoft.com/dotnet/api/microsoft.extensions.hosting.host.createdefaultbuilder - all overloads

gewarren:
Fixes #29945.

[Internal preview](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/extensions/7.0/contentrootpath-hosted-app?branch=pr-en-us-29971).

### DIFF
--- a/docs/core/compatibility/7.0.md
+++ b/docs/core/compatibility/7.0.md
@@ -51,6 +51,12 @@ If you're migrating an app to .NET 7, the breaking changes listed here might aff
 | - | :-: | :-: | - |
 | [Multi-level lookup is disabled](deployment/7.0/multilevel-lookup.md) | ❌ | ✔️ | Preview 4 |
 
+## Extensions
+
+| Title | Binary compatible | Source compatible | Introduced |
+| - | :-: | :-: | - |
+| [ContentRootPath for apps launched by Windows Shell](extensions/7.0/contentrootpath-hosted-app.md) | ❌ | ✔️ | Preview 6 |
+
 ## Networking
 
 | Title | Binary compatible | Source compatible | Introduced |
--- a/docs/core/compatibility/extensions/7.0/contentrootpath-hosted-app.md
+++ b/docs/core/compatibility/extensions/7.0/contentrootpath-hosted-app.md
@@ -5,11 +5,11 @@ ms.date: 06/22/2022
 ---
 # ContentRootPath for apps launched by Windows Shell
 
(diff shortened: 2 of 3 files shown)


tdykstra (approved):
LGTM

gewarren:
Thanks @tdykstra!

PR #29990, opened 2022-06-24
Bug fix in code example

OlegAlexander:
## Summary

This code example seems to have multiple issues:

- It's referring to a variable called `array` instead of `arr`.
- It's referring to an undeclared variable called `x`.
- It's referring to a counter `i` when the actual counter is `j`.
- It's not incrementing `sum`.

The fixed code example actually runs and produces the same `sum` result as the example above it. 

I also changed the loop counters from `j` to `i` because `i` is standard.

Thank you.


BillWagner (approved):
Thanks again @OlegAlexander 

This LGTM. I'll :shipit: now.

PR #29996, opened 2022-06-24
Document new PublishRelease and PackRelease properties

nagilson:
## Summary

Add documentation for the new properties we are adding to the SDK. 
* The pack property was included here because it is specific to the SDK and not part of MSBuild. 
Reflects the changes we're adding here https://github.com/dotnet/sdk/pull/25991

### DIFF
--- a/docs/core/project-sdk/msbuild-props.md
+++ b/docs/core/project-sdk/msbuild-props.md
@@ -139,6 +139,15 @@ You can specify properties such as `PackageId`, `PackageVersion`, `PackageIcon`,
 </PropertyGroup>
 ```
 
+### PackRelease 
+
+The `PackRelease` property is similar to the [PublishRelease](#publishrelease) property, except that it changes the default behavior of `dotnet pack`.
+```xml
+<PropertyGroup>
+	<PackRelease>true</PackRelease>
+</PropertyGroup>
+```
+
 ## Publish-related properties
 
 The following MSBuild properties are documented in this section:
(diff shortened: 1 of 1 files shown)


nagilson:
Will fix the markdown format after some meetings.

docs/core/project-sdk/msbuild-props.md
```diff
@@ -304,6 +315,19 @@ In .NET 5 and earlier versions, reference assemblies are always written to the `
 
 For more information, see [Write reference assemblies to intermediate output](../compatibility/sdk/6.0/write-reference-assemblies-to-obj.md).
 
+### PublishRelease
+
+The `PublishRelease` property informs `dotnet publish` to leverage `Release` configuration instead of `Debug` by default. Add it to a `Directory.Build.props` file.
```
gewarren, line 320: @nagilson Is there a reason to add it to a Directory.Build.props file instead of a project file?

nagilson: If the property is added to the `.csproj` file, it won't be evaluated early enough for the configuration change to propagate properly

gewarren, line 320:
```suggestion
The `PublishRelease` property informs `dotnet publish` to leverage the `Release` configuration instead of the `Debug` configuration. We recommend adding this property to a `Directory.Build.props` file instead of a project file so that it's evaluated early enough for the configuration change to propagate.
```

gewarren (approved):
Thanks for adding these!

PR #30006, opened 2022-06-27
Updated console output to match code

dutts:
Line 51 asks the user to modify the program.cs output to the string "Hello World with C#" but line 67 stated you should get "Hello World with C#!" as output.

## Summary

Modified the docs so that the expected output matches the proposed code change.


tdykstra (approved):
Thanks for fixing this, @dutts !

PR #30010, opened 2022-06-27
Changed processor to process

rafcon-dev:
To be in line with what's on the nint and nuint pages https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types#native-sized-integers.


BillWagner (approved):
Thanks @rafcon-dev 

I'll :shipit: now.

PR #30015, opened 2022-06-27
Add the PublishProtocol property to the list of telemetry data disclosures

baronfel:
## Summary

With https://github.com/dotnet/sdk/pull/26287 we started logging the PublishProtocol MSBuild property when a Publish is performed. We need to document this in the telemetry data disclosures.


tdykstra (approved):
LGTM

PR #30019, opened 2022-06-27
Fix example define symbols for .NET framework

mgurlitz:
## Summary

These example symbols do not line up with the ones listed in the table directly above. This may confuse or mislead readers, who expect the .NET Framework symbols to follow the same underscored format as the .NET 5+ ones.

I tested with the .NET 6 SDK targeting .NET Framework 4.7.2, `NET20_OR_GREATER` was defined while `NET_2_0_OR_GREATER` was not.

The `NET11_OR_GREATER` and `NET10_OR_GREATER` symbols **were not defined**, however they aren't listed in the table either.


gewarren (approved):
Thanks @mgurlitz

PR #30020, opened 2022-06-27
Fix casing of Arm64 etc.

gewarren:
Inspired by https://github.com/dotnet/docs/pull/29968#discussion_r907781348.

### DIFF
--- a/docs/core/deploying/single-file/overview.md
+++ b/docs/core/deploying/single-file/overview.md
@@ -164,7 +164,7 @@ In .NET Core 3.x, publishing as a single file produced one file, consisting of t
 
 Starting with .NET 5, only managed DLLs are bundled with the app into a single executable. When the app starts, the managed DLLs are extracted and loaded in memory, avoiding the extraction to a folder. On Windows, this approach means that the managed binaries are embedded in the single file bundle, but the native binaries of the core runtime itself are separate files.
 
-To embed those files for extraction and get one output file, like in .NET Core 3.x, set the property `IncludeNativeLibrariesForSelfExtract` to `true`. For more information about extraction, see [Including native libraries](#including-native-libraries).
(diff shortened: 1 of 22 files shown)


docs/core/deploying/single-file/overview.md
```diff
@@ -152,7 +152,40 @@ For example, add the following property to the project file of an assembly to em
 </PropertyGroup>
 ```
 
-## Other considerations
+## Compress assemblies in single file app
+
+Starting with .NET 6, single file apps can be created with compression enabled on the embedded assemblies. Set `EnableCompressionInSingleFile` property to `true` to achieve this. The produced single-file will have all of the embedded assemblies compressed which can significantly reduce the size of the executable. Compression comes with a performance cost. On application start, the assemblies must be decompressed into memory, which takes some time. It's recommended to measure both the size impact and startup cost impact of enabling compression before using it as the impact varies a lot between different applications.
```
tdykstra, line 157:
```suggestion
Starting with .NET 6, single-file apps can be created with compression enabled on the embedded assemblies. Set the `EnableCompressionInSingleFile` property to `true` to achieve this. The single file that is produced will have all of the embedded assemblies compressed, which can significantly reduce the size of the executable. Compression comes with a performance cost. On application start, the assemblies must be decompressed into memory, which takes some time. We recommend that you measure both the size impact and startup cost impact of enabling compression before using it. The impact varies a lot between different applications.
```

docs/core/deploying/single-file/overview.md
```diff
@@ -152,7 +152,40 @@ For example, add the following property to the project file of an assembly to em
 </PropertyGroup>
 ```
 
-## Other considerations
+## Compress assemblies in single file app
```
tdykstra, line 155:
```suggestion
## Compress assemblies in single-file apps
```

tdykstra (approved):
LGTM, a few suggestions.

PR #29925, opened 2022-06-18
Fixes #28385

rafcon-dev:
Adds clarification code on how  init prevents assignment after constructor. Added clarification on the usefulness of init. Changed the property of the example (seconds) to a one that more likely would benefit from init (YearOfBirth in a Person class, for example). Lowered verbosity on the expression-bodied examples, as they really didn't add anything useful in the specific context of the init article.

Fixes #28385

- edit by @billwagner  Add Fixes tag in the description for automation and linking.

PR #30033, opened 2022-06-29
Fix incorrect info in keywords/init.md

GitHubPang:
## Summary

- Fix wrong property and field name (should be `YearOfBirth` instead of `Seconds`), due to #29925.
- Remove apostrophe in a comment ("it's value" > "its value").


BillWagner (approved):
Thanks again @GitHubPang 

I'll :shipit: now.

PR #30057, opened 2022-06-30
Clarify changes to string format for float/double

### LINKED ISSUE #29952: General format specifier (G) is incorrect for .NET 6
The table on this page says:

> If the precision specifier is omitted or zero, the type of the number determines the default precision, as indicated in the following table.
...
> Single | 7 digits
> Double | 15 digits

But when I run the following code on .NET 6, I'm seeing 9 digits for Single and 17 digits for Double:

```C#
float f = 123456789f;
f.ToString("G")
```
prints
```
123456790
```

and

```C#
double d = 12345678912345678;
d.ToString("G")
```
prints
```
12345678912345678
```

cc @tannergooding 

---
#### Document Details

* ID: 0ae38fce-fc98-9420-6b0b-ceee4e885f2d
* Version Independent ID: f23035e0-f271-b48c-f9a9-8582be80b4ad
* Content: [Standard numeric format strings](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings#general-format-specifier-g)
* Content Source: [docs/standard/base-types/standard-numeric-format-strings.md](https://github.com/dotnet/docs/blob/main/docs/standard/base-types/standard-numeric-format-strings.md)
* GitHub Login: @adegeo
* Microsoft Alias: **adegeo**

adegeo:
## Summary

@eerhardt @tannergooding 

Fixes #29952

### DIFF
--- a/docs/standard/base-types/standard-numeric-format-strings.md
+++ b/docs/standard/base-types/standard-numeric-format-strings.md
@@ -200,10 +200,10 @@ The general ("G") format specifier converts a number to the more compact of eith
 |<xref:System.Int64>                        |19 digits|
 |<xref:System.UInt64>                       |20 digits|
 |<xref:System.Numerics.BigInteger>          |Unlimited (same as ["R"](#RFormatString))|
-|<xref:System.Half>                         |5 digits|
-|<xref:System.Single>                       |9 digits (7 digits in .NET Framework)|
-|<xref:System.Double>                       |17 digits (15 digits in .NET Framework)|
-|<xref:System.Decimal>                      |29 digits|
+|<xref:System.Half>                         |Smallest round-trippable number of digits to represent the number|
+|<xref:System.Single>                       |Smallest round-trippable number of digits to represent the number (in .NET Framework, G7 is the default)|
+|<xref:System.Double>                       |Smallest round-trippable number of digits to represent the number (in .NET Framework, G15 is the default)|
+|<xref:System.Decimal>                      |Smallest round-trippable number of digits to represent the number|
 
 Fixed-point notation is used if the exponent that would result from expressing the number in scientific notation is greater than -5 and less than the precision specifier; otherwise, scientific notation is used. The result contains a decimal point if required, and trailing zeros after the decimal point are omitted. If the precision specifier is present and the number of significant digits in the result exceeds the specified precision, the excess trailing digits are removed by rounding.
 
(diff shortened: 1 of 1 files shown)


docs/standard/base-types/standard-numeric-format-strings.md
```diff
@@ -192,18 +192,18 @@ The following example formats a <xref:System.Double> and an <xref:System.Int32>
...
-|<xref:System.Half>|3 digits|
-|<xref:System.Single>|7 digits|
-|<xref:System.Double>|15 digits|
-|<xref:System.Decimal>|29 digits|
+|<xref:System.Int64>                        |19 digits|
+|<xref:System.UInt64>                       |20 digits|
+|<xref:System.Numerics.BigInteger>          |Unlimited (same as ["R"](#RFormatString))|
+|<xref:System.Half>                         |3 digits|
+|<xref:System.Single>                       |9 digits (7 digits in .NET Framework)|
+|<xref:System.Double>                       |17 digits (15 digits in .NET Framework)|
```
tannergooding, line 205: `Half` would be 5, not `3`. But `5`, `9`, and `17` aren't really "accurate" here at all.

For the floating-point types, we print the "shortest roundtrippable string" (equivalent to `R`) and so even if more digits of precision are available, we may not surface them if there is a shorter string that also equates to the same value.

adegeo: @tannergooding  Hmm. So are you saying there is no default precision value for the types when omitted? That's what the table is explaining, the default precision.

tannergooding: For Half, Single, and Double on .NET Core 3.1 and later, that's roughly correct.

They don't have a precision and `G5`/`G9`/`G17` will all turn a distinct result from what `G` does. The same is true for `decimal` going back to .NET Framework.

adegeo: @tannergooding 

I think that is what the paragraphs after the table explain:

> When used with a <xref:System.Double> value, the "G17" format specifier ensures that the original <xref:System.Double> value successfully round-trips. This is because <xref:System.Double> is an IEEE 754-2008-compliant double-precision (`binary64`) floating-point number that gives up to 17 significant digits of precision. On .NET Framework, we recommend its use instead of the ["R" format specifier](#RFormatString), since in some cases "R" fails to successfully round-trip double-precision floating point values.

Should we update this with more information that states single/decimal/float/double all return a value that round-trips if available?

tannergooding: The issue with that paragraph is the "up to 17 significant digits" part.

The most digits you need to guarantee a roundtrippable value for `double` is 17. However, the actual value represented often has many more digits (up to 767 of them).

The same applies to `float` where the most you need is `9` but where the actual represented may have up to `112` significant digits and similarly for `Half` with `5` and up to `19`.

tannergooding (approved):

adegeo:
I got some more feedback from tanner that I'll implement

gewarren:
> I got some more feedback from tanner that I'll implement

Can you put it into draft mode then so it doesn't get merged accidentally?

tannergooding (approved):

PR #30078, opened 2022-07-01
Reordering words in "Named arguments" section

petarmari266:
## Summary

Changed the order of two words which in my opinion improves subject coherence. _"We can supply named arguments for each method parameter."_

Now that I think of it, maybe the first sentence should also be changed to: ```Named arguments free you from matching the order of arguments to the order of parameters in the parameter lists of called methods.```.

Sorry for nitpicking. I understand the usage as it is currently written but I guess it can cause misunderstandings for beginners.


BillWagner:
Hi @petarmari266 

I like this wording much better:

> Named arguments free you from matching the order of arguments to the order of parameters in the parameter lists of called methods.

If you make that change in this PR, I'll :shipit:

petarmari266:
Also, would you mind if I were to do more of these little commits which might improve readability or understanding of the existing text?

BillWagner:
> Also, would you mind if I were to do more of these little commits which might improve readability or understanding of the existing text?

That would be very helpful. Thank you.

PR #30096, opened 2022-07-05
Update to correct typo with ReferenceEquals usage

rsoulard-cadenza:
- Edited usage of ReferenceEquals() in the EqualOperator() function to use the reference of the right value object rather than null. I believe this was a typo since checking against a null reference with the left value object a second time makes little sense as the function would have already returned. It makes more sense to check if the left and right value object reference the same object at that point in the function.

## Summary

Changed call to `ReferenceEquals(left, null)` to `ReferenceEquals(left, right)`.


IEvangelist (approved):

GitHubPang:
https://github.com/dotnet-architecture/eShopOnContainers/blob/dev/src/Services/Ordering/Ordering.Domain/SeedWork/ValueObject.cs#L11
🤔

PR #30120, opened 2022-07-07
Update unmanaged.md

WAftring:
Adding emphasis to clean up of unmanaged object clean up.

## Summary

Describe your changes here.

### DIFF
--- a/docs/standard/garbage-collection/unmanaged.md
+++ b/docs/standard/garbage-collection/unmanaged.md
@@ -16,7 +16,7 @@ ms.assetid: a17b0066-71c2-4ba4-9822-8e19332fc213
 
 # Cleaning up unmanaged resources
 
-For a majority of the objects that your app creates, you can rely on the [.NET garbage collector](index.md) to handle memory management. However, when you create objects that include unmanaged resources, you must explicitly release those resources when you finish using them. The most common types of unmanaged resources are objects that wrap operating system resources, such as files, windows, network connections, or database connections. Although the garbage collector is able to track the lifetime of an object that encapsulates an unmanaged resource, it doesn't know how to release and clean up the unmanaged resource.
+For a majority of the objects that your app creates, you can rely on the [.NET garbage collector](index.md) to handle memory management. However, when you create objects that include unmanaged resources, you **must** explicitly release those resources when you finish using them. The most common types of unmanaged resources are objects that wrap operating system resources, such as files, windows, network connections, or database connections. Although the garbage collector is able to track the lifetime of an object that encapsulates an unmanaged resource, it doesn't know how to release and clean up the unmanaged resource.
 
 If your types use unmanaged resources, you should do the following:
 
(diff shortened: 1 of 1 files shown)


docs/standard/garbage-collection/unmanaged.md
```diff
@@ -31,15 +31,15 @@ If your types use unmanaged resources, you should do the following:
   - Define a [finalizer](../../csharp/programming-guide/classes-and-structs/finalizers.md). Finalization enables the non-deterministic release of unmanaged resources when the consumer of a type fails to call <xref:System.IDisposable.Dispose%2A?displayProperty=nameWithType> to dispose of them deterministically.
 
     > [!WARNING]
-    > However, because object finalization can be a complex and an error-prone operation, we recommend that you use a safe handle instead of providing your own finalizer.
+    > Object finalization can be a complex and error-prone operation, we recommend that you use a safe handle instead of providing your own finalizer.
```
gewarren, line 34:
```suggestion
    > Object finalization can be a complex and error-prone operation. We recommend that you use a safe handle instead of providing your own finalizer.
```

docs/standard/garbage-collection/unmanaged.md
```diff
@@ -31,15 +31,15 @@ If your types use unmanaged resources, you should do the following:
...
 
 ## In this section
 
 [Implementing a Dispose method](implementing-dispose.md) describes how to implement the dispose pattern for releasing unmanaged resources.
 
-[Using objects that implement `IDisposable`](using-objects.md) describes how consumers of a type ensure that its <xref:System.IDisposable.Dispose%2A> implementation is called. We recommend using the C# `using` (or the Visual Basic `Using`) statement to do this.
+[Using objects that implement `IDisposable`](using-objects.md) describes how consumers of a type ensure that its <xref:System.IDisposable.Dispose%2A> implementation is called. We strongly recommend using the C# `using` (or the Visual Basic `Using`) statement to do this.
```
gewarren, line 42: @BillWagner Do you think this new adverb is justified?

gewarren (approved):

docs/standard/garbage-collection/unmanaged.md
BillWagner: I'd probably leave it off. The general convention is to prefer `using` when you wouldn't have a `try` - `catch` for cases where recovery is possible.  If the code would include `try` - `catch` - `finally`, writing the `finally` clause is likely cleaner.

IEvangelist: I agree with Bill, there are times when we recommend not having implementations in a `using` too.

gewarren, line 42:
```suggestion
[Using objects that implement `IDisposable`](using-objects.md) describes how consumers of a type ensure that its <xref:System.IDisposable.Dispose%2A> implementation is called. We recommend using the C# `using` (or the Visual Basic `Using`) statement to do this.
```

gewarren: Per previous conversation.

PR #30126, opened 2022-07-07
Improving the readability of the paragraph

petarmari266:
## Summary

Using the word "either" in the sentence which provides the first example, and beginning the second sentence with "Alternatively", which provides the second example, seems wrong to me, or worthy of improvement.


BillWagner (approved):
Thanks @petarmari266 

PR #30135, opened 2022-07-08
Promote use of GenerateDocumentationFile over DocumentationFile

drewnoakes:
For a documentation file to be correctly picked up by the IDE, it must have a specific name and be in a specific location. It is easier and more maintainable to opt into documentation file generation with a boolean value than with a file path. For this reason, we should promote use of `GenerateDocumentationFile`, with only a passing reference to the older `DocumentationFile` property for rare case that the default behaviour should be overridden.


docs/csharp/language-reference/compiler-options/output.md
```diff
@@ -17,26 +17,30 @@ The following options control compiler output generation.
...
+The **GenerateDocumentationFile** option generates an XML file containing the documentation comments from the source code files of the compilation. This file allows consumers of your binaries to read API-level documentation directly in their IDE. To learn more about documenting your code, see [Recommended Tags for Documentation Comments](../xmldoc/recommended-tags.md).
+
+The easiest way to enable this for your project is via the `GenerateDocumentationFile` property:
 
 ```xml
-<DocumentationFile>path/to/file.xml</DocumentationFile>
+<GenerateDocumentationFile>true</GenerateDocumentationFile>
 ```
 
+By default, the documentation file shared the name of the assembly it is documenting, with the `.dll` extension substituted for `.xml`. In the rare case that you need to customise the documentation file's path, you can use `<DocumentationFile>` to override the default path.
```
drewnoakes, line 36:
```suggestion
By default, the documentation file shares the name of the assembly it is documenting, with the `.dll` extension substituted for `.xml`. In the rare case that you need to customise the documentation file's path, you can use the `<DocumentationFile>` property to override the default path.
```

drewnoakes:
I'm closing this as each of the inbound links requires investigation and I don't have time to complete that work right now.

PR #30144, opened 2022-07-08
Update article based on #22246

### LINKED ISSUE #22246: Shouldnt the custom console logger's IsEnabled function use >= instead of ==?
Because the text right below it says 

> Loggers should also be enabled for all higher log levels.

Also, what does "higher" mean in this case? I've only ever seen developers only make a LogLevel enum thats like this...

``` csharp
public enum LogLevel
{
   Error = 0,
   Warn = 5,
   Info = 10, 
   Verbose = 15  //int values are sometimes ^2 flags, and sometimes the enum carries no explicit int
}
```

---
#### Document Details

* ID: b609eb80-71e8-7a3c-4c4c-60cf9a315cfb
* Version Independent ID: e111baf5-04b7-c497-ce86-eb687e121751
* Content: [Implement a custom logging provider in .NET](https://docs.microsoft.com/en-us/dotnet/core/extensions/custom-logging-provider)
* Content Source: [docs/core/extensions/custom-logging-provider.md](https://github.com/dotnet/docs/blob/master/docs/core/extensions/custom-logging-provider.md)
* GitHub Login: @IEvangelist
* Microsoft Alias: **dapine**

IEvangelist:
## Summary

Add details about how this implementation requires that each log level requires an explicit configuration entry in order to log.

Fixes #22246

### DIFF
--- a/docs/core/extensions/custom-logging-provider.md
+++ b/docs/core/extensions/custom-logging-provider.md
@@ -3,7 +3,7 @@ title: Implement a custom logging provider in .NET
 description: Learn how to implement a custom logging provider in your .NET applications.
 author: IEvangelist
 ms.author: dapine
-ms.date: 01/04/2022
+ms.date: 07/08/2022
 ms.topic: how-to
 ---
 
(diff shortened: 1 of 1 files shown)


BillWagner (approved):

PR #30149, opened 2022-07-09
Fix `gc-committed` counter name

nxtn:
There is no such counter.

See dotnet/runtime#50604.

https://github.com/dotnet/runtime/blob/v6.0.0/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L94
https://github.com/dotnet/runtime/blob/v7.0.0-preview.5.22301.12/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L94


gewarren (approved):
Thanks @nxtn

PR #30160, opened 2022-07-11
Address feedback from #30101

### LINKED ISSUE #30101: Github Actions: Add Note/Warning - for Full Path to the project DLL in docker
**Page in question**
[Github Action Tutorial](https://docs.microsoft.com/en-us/dotnet/devops/create-dotnet-github-action#the-dockerfile)

**Problematic Content**
```dockerfile
FROM mcr.microsoft.com/dotnet/sdk:6.0
COPY --from=build-env /app/out .
ENTRYPOINT [ "dotnet", "/DotNet.GitHubAction.dll" ]
```

**Issue I ran into**
I built a service which uses the default docker template (the 'add docker support' feature). Seeing that I use the `runtime`, I didn't pay attention to the `entrypoint`!!
The template uses a relative path to the DLL using a `workdir` of `/app`. Inside the runner it sets the `workdir` to `/github/workspace`, something different. Thus, when the entrypoint executes it fails!

**Proposed Solution**
Add a highlight to
```dockerfile
ENTRYPOINT [ "dotnet", "/DotNet.GitHubAction.dll" ] # this line
```
Add a note, similar to the one explaining size differences between SDK and runtime but indicating that providing a **full path** to the DLL is required in order to work in the runner.

Additionally, I'd propose that the powers that be update the default docker template to use a full path. 

```dockerfile
FROM base AS final
WORKDIR /app
COPY --from=publish /app/publish .
ENTRYPOINT ["dotnet", "/app/myapplication.dll"] # a 5 character change here
```

Or to use the same approach as this example where the workdir isn't needed. 

---
#### Document Details

* ID: c32d49db-bd6c-c5db-96c9-fe426191d6ca
* Version Independent ID: 491a0389-f24d-a398-4496-873461b6180d
* Content: [Tutorial: Create a GitHub Action with .NET](https://docs.microsoft.com/en-us/dotnet/devops/create-dotnet-github-action#the-dockerfile)
* Content Source: [docs/devops/create-dotnet-github-action.md](https://github.com/dotnet/docs/blob/main/docs/devops/create-dotnet-github-action.md)
* Product: **dotnet**
* GitHub Login: @IEvangelist
* Microsoft Alias: **dapine**

IEvangelist:
## Summary

Add a warning about the slight difference in the _Dockerfile_.

Fixes #30101

### DIFF
--- a/docs/devops/create-dotnet-github-action.md
+++ b/docs/devops/create-dotnet-github-action.md
@@ -3,7 +3,7 @@ title: "Tutorial: Create a GitHub Action with .NET"
 description: Learn how to create a GitHub Action with a containerized .NET app.
 author: IEvangelist
 ms.author: dapine
-ms.date: 05/18/2022
+ms.date: 07/11/2022
 ms.topic: tutorial
 recommendations: false
 ---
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @IEvangelist 

You can :shipit: when ready.

PR #30186, opened 2022-07-13
Operator precedence: mention unsigned right shift operator

pkulikov:


BillWagner (approved):
Thanks for spotting this and catching it @pkulikov I missed it when I added the `>>>` operator.

I'll :shipit: now.

PR #30188, opened 2022-07-13
C# bitwise and shift operators: fix interactive snippets syntax

pkulikov:
Interactive C# code UX is broken for the [bitwise and shift operators article](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators):
image

This PR should fix that. Please check that before merging.


BillWagner (approved):
Thanks @pkulikov for fixing this. I'll :shipit: now.

PR #30215, opened 2022-07-13
Add attribute target specifier keywords

CollinAlpert:
## Summary

This PR adds the attribute target specifier keywords to the "Contextual keywords" section. It does not add the keywords `event` and `return`, as these are already listed in the table above. I wasn't sure about the link for the keywords, but am happy about suggestions of better targets.


BillWagner:
Hi @CollinAlpert 

Thanks for doing this. I'm going to close this because of how [attribute targets](https://docs.microsoft.com/dotnet/csharp/language-reference/language-specification/attributes#213-attribute-specification) are defined in the language spec. They aren't "contextual keywords", but rather "attribute target specifiers".

PR #30219, opened 2022-07-13
Change can to can't

jtschuster:
I'm guessing this is a typo or missing a 'but this will error at compile time'


tdykstra:
"can't" wouldn't make sense in context -- the "For example..." sentence illustrates the statement "Reflection mode supports the use of non-public accessors of public properties."

PR #30177, opened 2022-07-12
update resource management docs keep align with latest SDK

xboxeer:
## Summary
Since we've GAed some major RP's SDK Packages, our current docs for .net mgmt sdk is out of date. Update the docs to catch up the latest SDK.
Also added some sections based on our customer interview findings.
Update resource management docs

Fixes https://github.com/Azure/azure-sdk-for-net/issues/29797

Internal preview:

- [Resource management using the Azure SDK for .NET](https://review.docs.microsoft.com/en-us/dotnet/azure/sdk/resource-management?branch=pr-en-us-30177&tabs=dotnetcli)

PR #30221, opened 2022-07-14
Update the version of Azure ResourceManager libs

ArthurMa1978:
## Summary

Have released stable version for `Azure.ResourceManager.Compute` and `Azure.ResourceManager.Network`.


IEvangelist:
We're in the process of reviewing #30177, which already contains these changes. I'm going to close this one in favor of #30177.

PR #30239, opened 2022-07-15
Add note on suppressing trimming warnings

MichalStrehovsky:
Inspired by the suppression justification from https://github.com/dotnet/linker/pull/2891.

Cc @vitek-karas @sbomer @tlakollo @eerhardt


vitek-karas (approved):
Thanks a lot!

I'm still terrible at thinking about trimming the right way - I think what the linker does today, not what it "Should do" in theory.

MichalStrehovsky:
> I'm still terrible at thinking about trimming the right way - I think what the linker does today, not what it "Should do" in theory.

The more aggressive constant propagation in .NET 7 already puts the IL linker into these territories today. It doesn't kick in often, but when it does, statically used properties and methods get removed.

eerhardt (approved):

sbomer (approved):
Thank you!

MichalStrehovsky:
@adegeo is this okay to merge?

IEvangelist (approved):

PR #30244, opened 2022-07-16
Document StripSymbols property for native AOT

jkotas:
The property was introduced by https://github.com/dotnet/runtime/issues/69847 and it is one of the frequently asked questions for native AOT publishing.

### DIFF
--- a/docs/core/deploying/native-aot.md
+++ b/docs/core/deploying/native-aot.md
@@ -70,11 +70,11 @@ The native AOT publishing follows platform conventions for native toolchains. Th
 
 Set `StripSymbols` property to true to produce the debug information in a separate .dbg file and exclude it from the native binary on Linux. This property has no effect on Windows.
 
-    ```xml
-    <PropertyGroup>
-        <StripSymbols>true</StripSymbols>
-    </PropertyGroup>
-    ```
+```xml
+<PropertyGroup>
+    <StripSymbols>true</StripSymbols>
+</PropertyGroup>
+```
 
 ## Limitations of Native AOT deployment
 
--- a/docs/core/deploying/native-aot.md
+++ b/docs/core/deploying/native-aot.md
@@ -64,6 +64,18 @@ The app will be available in the publish directory and will contain all the code
 
 Check out the [native AOT samples](https://github.com/dotnet/samples/tree/main/core/nativeaot) available in the dotnet/samples repository on GitHub. The samples includes [Linux](https://github.com/dotnet/samples/blob/main/core/nativeaot/HelloWorld/Dockerfile) and [Windows](https://github.com/dotnet/samples/blob/main/core/nativeaot/HelloWorld/Dockerfile.windowsservercore-x64) Dockerfiles that demonstrate how to automate installation of prerequisites and publishing .NET projects with native AOT using containers.
 
+### Native Debug Information
+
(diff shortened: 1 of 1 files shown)


jkotas:
cc  @am11 @dotnet/ilc-contrib

am11:
Thank you!

There is also a similar property `NativeDebugSymbols`, which works cross-platform. The differences are:

* `NativeDebugSymbols=false` only strips the debug symbols. There are boat load of additional symbols which `StripSymbols=true` shaves off (including the debug symbols).
* `NativeDebugSymbols=false` does not generate a (large) .dbg file, `StripSymbols=true` does. Consumers can copy that file to their production environment on-demand in rare cases when they need to debug something in production.

Publishing MVC app (`mvc1`) with `PublishAot=true` shows these stats:
* Without stripping: 135M
* `NativeDebugSymbols=false`:101M
* `StipSymbols=true`: 44M (and size of mvc1.dbg is 91 MB) 

The symbols which are exported by `UnmanagedCallersOnlyAttribute` are preserved by all three modes, and things continue to work with `extern <signature>` in C, or `extern "C" <signature>` in C++.

Perhaps we can include some of this information in the docs to clarify that it is safe to use this option.

jkotas:
This is the entrypoint doc for publishing as native AOT. We need to balance the level of details in the doc to what is needed for common cases.

We get questions about the stripping symbols on Linux a lot. It is what made me to submit this PR.  I do not think we ever got a question about `NativeSymbols`. We may eventually want to start a second more advanced doc topic that has the options like `NativeSymbols` and the level of detail that most people publishing as native AOT do not needed to know.

am11:
I agree, it can be captured in some other document which focuses on `NativeLibrary=Shared`. I was thinking about clarifying `UnmanagedCallersOnlyAttribute` usage with `/p:NativeLibrary=shared` being safe when `StripSymbols=true`. The NativeAOT entry-level docs are more focused on `OutputType=Exe`.

LGTM. Thanks!

MichalStrehovsky (approved):
Thank you!

docs/core/deploying/native-aot.md
```diff
@@ -64,6 +64,18 @@ The app will be available in the publish directory and will contain all the code
 
 Check out the [native AOT samples](https://github.com/dotnet/samples/tree/main/core/nativeaot) available in the dotnet/samples repository on GitHub. The samples includes [Linux](https://github.com/dotnet/samples/blob/main/core/nativeaot/HelloWorld/Dockerfile) and [Windows](https://github.com/dotnet/samples/blob/main/core/nativeaot/HelloWorld/Dockerfile.windowsservercore-x64) Dockerfiles that demonstrate how to automate installation of prerequisites and publishing .NET projects with native AOT using containers.
 
+### Native Debug Information
+
+The native AOT publishing follows platform conventions for native toolchains. The default behavior of native toolchains on Windows is to produce debug information in separate .pdb file. The default behavior of native toolchains on Linux is to include the debug information in the native binary that makes the native binary significantly larger.
```
IEvangelist, line 69:
```suggestion
The native AOT publishing follows platform conventions for native toolchains. The default behavior of native toolchains on Windows is to produce debug information in a separate _.pdb_ file. The default behavior of native toolchains on Linux is to include the debug information in the native binary which makes the native binary significantly larger.
```

IEvangelist (approved):
A few nits

PR #30247, opened 2022-07-16
Fix placement of summary paragraph

eatdrinksleepcode:
When the section "List patterns" was added in https://github.com/dotnet/docs/commit/b181812600f3a1efe8a57fefd7dee80a469d9f93, it was placed after the paragraph that was intended to be at the end of the article. This commit puts the summary paragraph back at the end where it was intended to be.


BillWagner (approved):
Thanks for updating this @eatdrinksleepcode 

I don't know how I missed that, but I'm glad you caught it.

I'll :shipit: now.

PR #30252, opened 2022-07-17
Fix the snippet for a property pattern

### LINKED ISSUE #30248: Confusing information on deconstruction in pattern matching
The "Multiple Inputs" section gives this example

```csharp
public record Order(int Items, decimal Cost);

public decimal CalculateDiscount(Order order) =>
    order switch
    {
        (Items: > 10, Cost: > 1000.00m) => 0.10m,
        (Items: > 5, Cost: > 500.00m) => 0.05m,
        Order { Cost: > 250.00m } => 0.02m,
        null => throw new ArgumentNullException(nameof(order), "Can't calculate discount on null order"),
        var someObject => 0m,
    };
```

It then follows with another example:
> ...If the Order type defines a suitable [Deconstruct](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/deconstruct) method, you can omit the property names from the pattern and use deconstruction to examine properties:

```csharp
public decimal CalculateDiscount(Order order) =>
    order switch
    {
        ( > 10,  > 1000.00m) => 0.10m,
        ( > 5, > 50.00m) => 0.05m,
        Order { Cost: > 250.00m } => 0.02m,
        null => throw new ArgumentNullException(nameof(order), "Can't calculate discount on null order"),
        var someObject => 0m,
    };
```

The implication is that the first example works without a `Deconstruct` method, while the second example needs a `Deconstruct` method. However, that is not correct. The first example also uses deconstruction in the first two patterns, and needs a suitable `Deconstruct` method (like from a record) or it will not compile.

The third pattern in each example is the same and is _not_ using deconstruction. Together the examples paint a confusing picture of what is deconstruction and what isn't, and when you can use it versus not. This is particularly true for someone who, even if they are familiar with C#, is not experienced with deconstruction 👋 .

---
#### Document Details

* ID: 46225236-316d-bf2c-add4-6cea9b579ddc
* Version Independent ID: 528c02d9-3551-c7df-817e-ea50b2760f51
* Content: [Pattern matching overview - C# guide](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching#feedback)
* Content Source: [docs/csharp/fundamentals/functional/pattern-matching.md](https://github.com/dotnet/docs/blob/main/docs/csharp/fundamentals/functional/pattern-matching.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

pkulikov:
Fixes #30248

Also removes unnecessary type-check as the parameter is already of the `Order` type.

### DIFF
--- a/docs/csharp/fundamentals/functional/snippets/patterns/OrderProcessor.cs
+++ b/docs/csharp/fundamentals/functional/snippets/patterns/OrderProcessor.cs
@@ -10,9 +10,9 @@ class OrderProcessor
     public decimal CalculateDiscount(Order order) =>
         order switch
         {
-            (Items: > 10, Cost: > 1000.00m) => 0.10m,
-            (Items: > 5, Cost: > 500.00m) => 0.05m,
-            Order { Cost: > 250.00m } => 0.02m,
+            { Items: > 10, Cost: > 1000.00m } => 0.10m,
+            { Items: > 5, Cost: > 500.00m } => 0.05m,
+            { Cost: > 250.00m } => 0.02m,
             null => throw new ArgumentNullException(nameof(order), "Can't calculate discount on null order"),
             var someObject => 0m,
         };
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks again @pkulikov 

This looks great.

I'll :shipit: now.

PR #30292, opened 2022-07-20
Fixed csharp middleware example to use args

peteraritchie:
The existing example ignores the args and _always_ assumes the command line is `[just-say-hi] --delay 42 --message "Hello world!"`, which doesn't show a realistic implementation of middleware.  This change makes the example use `args` so it will show middleware being used in the right circumstances.

## Summary

Describe your changes here.


tdykstra (approved):
Thanks, @peteraritchie !

PR #30300, opened 2022-07-21
Default auth scheme handling

gewarren:
Related to aspnet/announcements#490.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/7.0/default-authentication-scheme?branch=pr-en-us-30300).

### DIFF
--- a/docs/core/compatibility/aspnet-core/7.0/default-authentication-scheme.md
+++ b/docs/core/compatibility/aspnet-core/7.0/default-authentication-scheme.md
@@ -13,11 +13,10 @@ Previously, users were required to set the default authentication scheme, which 
 builder.Services.AddAuthentication("MyDefaultScheme");
 ```
 
-Moving forward, when a *single* authentication scheme is registered, that scheme will be treated as the default scheme. For example, "myscheme" will be treated as the default scheme in the following code.
+Moving forward, when a *single* authentication scheme is registered, that scheme will be treated as the default scheme. For example, "MyDefaultScheme" will be treated as the default scheme in the following code.
 
 ```csharp
-builder.Services.AddAuthentication().AddOAuth("myscheme");
-```
+builder.Services.AddAuthentication().AddOAuth("MyDefaultScheme");
 
 This change might expose unintended behavior changes in applications, such as authentication options being validated earlier than expected.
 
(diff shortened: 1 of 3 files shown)


docs/core/compatibility/aspnet-core/7.0/default-authentication-scheme.md
```diff
@@ -0,0 +1,68 @@
...
+
+```csharp
+builder.Services.AddAuthentication("MyDefaultScheme");
+```
+
+Moving forward, when a *single* authentication scheme is registered, that scheme will be treated as the default scheme. For example, "myscheme" will be treated as the default scheme in the following code.
+
+```csharp
+builder.Services.AddAuthentication().AddOAuth("myscheme");
+```
```
IEvangelist, line 20:
```suggestion
Moving forward, when a *single* authentication scheme is registered, that scheme will be treated as the default scheme. For example, "MyDefaultScheme" will be treated as the default scheme in the following code.

```csharp
builder.Services.AddAuthentication().AddOAuth("MyDefaultScheme");
```
```

File: docs/core/compatibility/aspnet-core/7.0/default-authentication-scheme.md [hunk 2]
```diff
@@ -0,0 +1,68 @@
...
+
+```csharp
+builder.Services.AddAuthentication().AddCookie();
+```
+
+This impacted the behavior of authentication handlers in the application layer.
+
+## New behavior
+
+Starting in ASP.NET Core 7.0, if (and only if) a single scheme is registered in an application, that scheme is treated as the default. In the following code, the `CookieDefaults.AuthenticationScheme` is treated as the default scheme.
```
IEvangelist, line 40:
```suggestion
Starting in ASP.NET Core 7.0, when a single scheme is registered in an application, that scheme is treated as the default. In the following code, the `CookieDefaults.AuthenticationScheme` is treated as the default scheme.
```

IEvangelist (approved):
Looks good, a few nits.

docs/core/compatibility/aspnet-core/7.0/default-authentication-scheme.md
gewarren: I would like to keep this as-is to emphasize that it's in that specific case only.

captainsafia: Agreed. I included this language specifically because there was confusion in the announcement's discussion about whether this policy applied if more than one scheme was registered (it doesn't).

captainsafia (approved):

PR #30379, opened 2022-07-28
Generic math docs

### LINKED ISSUE #30142: Generic math conceptual docs
Resources:

Blog posts:

- https://devblogs.microsoft.com/dotnet/dotnet-7-generic-math/
- https://devblogs.microsoft.com/dotnet/preview-features-in-net-6-generic-math/

[Design doc](https://github.com/dotnet/designs/tree/main/accepted/2021/statics-in-interfaces)

[.NET Community standup video](https://www.youtube.com/watch?v=Odvs1wIu_cc&list=PLdo4fOcmZ0oX-DBuRG4u58ZTAJgBAeQ-t&index=8)

gewarren:
Fixes #30142.

Content was mostly taken from Tanner's blog posts - thanks Tanner!

[Internal preview](https://review.docs.microsoft.com/en-us/dotnet/standard/generics/math?branch=pr-en-us-30379).


docs/standard/generics/math.md
```diff
@@ -0,0 +1,199 @@
...
+| <xref:System.Numerics.IRootFunctions%601>          | Exposes root functions supporting `cbrt(x)` and `sqrt(x)`.                                                               |
+| <xref:System.Numerics.ITrigonometricFunctions%601> | Exposes trigonometric functions supporting `acos(x)`, `asin(x)`, `atan(x)`, `cos(x)`, `sin(x)`, and `tan(x)`.            |
+
+### Parsing interfaces
+
+Parsing is a common programming concept. It's typically used to support converting user input to a specified type, or deserializing data from a file. These interfaces are in the <xref:System> namespace.
+
+| Interface name                     | Description                                                                                      |
+|------------------------------------|--------------------------------------------------------------------------------------------------|
+| <xref:System.IParsable%601>        | Exposes support for `T.Parse(string, IFormatProvider)`.                                          |
```
gewarren, line 159: I left out the two formattable interfaces since I didn't see them anywhere in dotnet/runtime.

docs/csharp/whats-new/tutorials/snippets/staticinterfaces/RepeatSequence.cs
```diff
@@ -3,6 +3,10 @@
     private const char Ch = 'A';
     public string Text = new string(Ch, 1);
 
+    public RepeatSequence()
```
gewarren, line 6: This was to account for https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/breaking-changes/compiler%20breaking%20changes%20-%20dotnet%207#6.

docs/standard/generics/math.md
```diff
@@ -0,0 +1,199 @@
...
+| <xref:System.Numerics.IUnsignedNumber%601>             | Exposes APIs common to all unsigned number types.                                                      |
+| <xref:System.Numerics.IAdditiveIdentity%602>           | Exposes the concept of `(x + T.AdditiveIdentity) == x`.                                                |
+| <xref:System.Numerics.IMinMaxValue%601>                | Exposes the concept of `T.MinValue` and `T.MaxValue`.                                                  |
+| <xref:System.Numerics.IMultiplicativeIdentity%602>     | Exposes the concept of `(x * T.MultiplicativeIdentity) == x`.                                          |
+
+<sup>1</sup>The binary floating-point types are <xref:System.Double> (`double`), <xref:System.Half>, and <xref:System.Single> (`float`).
+
+<sup>2</sup>The binary integer types are <xref:System.Byte> (`byte`), <xref:System.Int16> (`short`), <xref:System.Int32> (`int`), <xref:System.Int64> (`long`), <xref:System.IntPtr> (`nint`), <xref:System.SByte> (`sbyte`), <xref:System.UInt16> (`ushort`), <xref:System.UInt32> (`uint`), <xref:System.UInt64> (`ulong`), and <xref:System.UIntPtr> (`nuint`).
```
BillWagner, line 55: optional.  If you want to link to the C# keywords for the numeric types, they are:
- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types
- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types

docs/standard/generics/math.md
```diff
@@ -0,0 +1,199 @@
...
+
+The interface you're most likely to use directly is <xref:System.Numerics.INumber%601>, which roughly corresponds to a "real" number. If a type implements this interface, it means that a value has a sign and well-defined order, and you can compare it to other values of the same type. <xref:System.Numerics.INumberBase%601> confers more advanced concepts, such as "complex" and "imaginary" numbers, for example, the square root of a negative number. Other interfaces, such as <xref:System.Numerics.IFloatingPointIeee754%601>, were created because not all operations make sense for all number&mdash;for example, calculating the floor of a number only makes sense for floating-point types. In the .NET base class library, the floating-point type <xref:System.Double> implements <xref:System.Numerics.IFloatingPointIeee754%601> but <xref:System.Int32> does not.
```
BillWagner, line 57: nit: This reads awkward to me (number -> number type):

```suggestion
The interface you're most likely to use directly is <xref:System.Numerics.INumber%601>, which roughly corresponds to a "real" number. If a type implements this interface, it means that a value has a sign and well-defined order, and you can compare it to other values of the same type. <xref:System.Numerics.INumberBase%601> confers more advanced concepts, such as "complex" and "imaginary" numbers, for example, the square root of a negative number. Other interfaces, such as <xref:System.Numerics.IFloatingPointIeee754%601>, were created because not all operations make sense for all number types&mdash;for example, calculating the floor of a number only makes sense for floating-point types. In the .NET base class library, the floating-point type <xref:System.Double> implements <xref:System.Numerics.IFloatingPointIeee754%601> but <xref:System.Int32> does not.
```

docs/standard/generics/math.md
BillWagner, line 159: I'd call out `TryParse` here as well.

BillWagner (approved):
This is really great work @gewarren 

I had a few comments, and then this is ready to go.

One that we may need to resolve: I changed `abstract` to `virtual` in most cases. C# 11 now allows `static virtual` in addition to `static abstract` members in interfaces. Stating both at each occurrence feels cumbersome. Maybe state both at the first instance in an article, and then simply "virtual"? What do you think? (The spec now refers to `static virtual` in that an `abstract` member is also `virtual`)

docs/standard/generics/interfaces.md
```diff
@@ -36,13 +36,19 @@ Generic interfaces provide type-safe counterparts to nongeneric interfaces for o
...
+- The **<xref:System.Collections.Generic.IEnumerable%601>** generic interface provides a generic enumerator structure. The <xref:System.Collections.Generic.IEnumerator%601> generic interface implemented by generic enumerators inherits the nongeneric <xref:System.Collections.IEnumerator> interface; the <xref:System.Collections.IEnumerator.MoveNext%2A> and <xref:System.Collections.IEnumerator.Reset%2A> members, which do not depend on the type parameter `T`, appear only on the nongeneric interface. This means that any consumer of the nongeneric interface can also consume the generic interface.
+
+## Mathematical functionality
+
+.NET 7 introduces generic interfaces in the <xref:System.Numerics?displayProperty=fullName> namespace that describe number-like types and the functionality available to them. The 20 numeric types that the .NET base class library provides, for example, <xref:System.Int32> and <xref:System.Double>, have been updated to implement these interfaces. The most prominent of these interfaces is <xref:System.Numerics.INumber%601>, which roughly corresponds to a "real" number.
```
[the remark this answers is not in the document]
gewarren: Since they're covered in the main article, I think the "see also" link is enough.

docs/standard/generics/math.md
gewarren, line 159:
```suggestion
| <xref:System.IParsable%601> | Exposes support for `T.Parse(string, IFormatProvider)` and `T.TryParse(string, IFormatProvider, out TSelf)`.  |
```

docs/standard/generics/math.md
```diff
@@ -0,0 +1,199 @@
...
+| <xref:System.Numerics.ITrigonometricFunctions%601> | Exposes trigonometric functions supporting `acos(x)`, `asin(x)`, `atan(x)`, `cos(x)`, `sin(x)`, and `tan(x)`.            |
+
+### Parsing interfaces
+
+Parsing is a common programming concept. It's typically used to support converting user input to a specified type, or deserializing data from a file. These interfaces are in the <xref:System> namespace.
+
+| Interface name                     | Description                                                                                      |
+|------------------------------------|--------------------------------------------------------------------------------------------------|
+| <xref:System.IParsable%601>        | Exposes support for `T.Parse(string, IFormatProvider)`.                                          |
+| <xref:System.ISpanParsable%601>    | Exposes support for `T.Parse(ReadOnlySpan<char>, IFormatProvider)`.                              |
```
gewarren, line 160:
```suggestion
| <xref:System.ISpanParsable%601> | Exposes support for `T.Parse(ReadOnlySpan<char>, IFormatProvider)` and `T.TryParse(ReadOnlySpan<char>, IFormatProvider, out TSelf)`. |
```

docs/csharp/whats-new/csharp-11.md
```diff
@@ -91,12 +91,12 @@ These types aren't directly represented in metadata. They include annotations th
...
+You can add *static abstract members* in interfaces to define interfaces that include overloadable operators, other static members, and static properties. The primary scenario for this feature is to use mathematical operators in generic types. For example, you can implement the `System.IAdditionOperators<TSelf, TOther, TResult>` interface in a type that implements `operator +`. Other interfaces define other mathematical operations or well-defined values. You can learn about the new syntax in the article on [interfaces](../language-reference/keywords/interface.md#static-abstract-and-virtual-members). Interfaces that include `static abstract` methods are typically [generic interfaces](../programming-guide/generics/generic-interfaces.md). Furthermore, most will declare a constraint that the type parameter [implements the declared interface](../programming-guide/generics/constraints-on-type-parameters.md#type-arguments-implement-declared-interface).
```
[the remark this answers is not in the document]
tannergooding: Should the first line be `static abstract or virtual members`?

docs/csharp/whats-new/csharp-11.md
```diff
@@ -91,12 +91,12 @@ These types aren't directly represented in metadata. They include annotations th
 
 There are several language features that enable generic math support:
 
-- static virtual members in interfaces
+- static abstract members in interfaces
 - checked user defined operators
 - relaxed right-shift requirements
```
tannergooding, line 96: Maybe `relaxed shift operators` is a better term? This features relaxes a requirement around the right hand side of a `left shift`, `signed (arithmetic) right shift`, or `unsigned (logical) right shift`. Previously it was required that the signature was `T operator <<(T left, int right)`, that is the right hand side must be `int`. Now it can be any type.

docs/csharp/whats-new/tutorials/static-abstract-interface-methods.md
```diff
@@ -1,19 +1,19 @@
 ---
-title: Explore static abstract members in interfaces - preview
+title: Explore static abstract members in interfaces
```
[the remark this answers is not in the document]
tannergooding: Potentially the file name as well?

docs/csharp/whats-new/tutorials/static-abstract-interface-methods.md
```diff
@@ -29,17 +29,17 @@ public static double MidPoint(double left, double right) =>
...
-Any type that implements the `INumber` interface must include a definition for `operator +`, and `operator /`. The denominator is defined by `(T.One + T.One)` to create the value `2` for any numeric type. The readonly properties `Zero` and `One` are also implemented by every type that implements `INumber`. Using that property rather than `2` forces the denominator to be the same type as the two parameters. This implementation has the potential for overflow if `left` and `right` are both large enough values. There are alternative algorithms that can avoid this potential issue.
+Any type that implements the <xref:System.Numerics.INumber%601> interface must include a definition for `operator +`, and for `operator /`. The denominator is defined by `(T.One + T.One)` to create the value `2` for any numeric type. The readonly properties `Zero` and `One` are also implemented by every type that implements `INumber`. Using that property rather than `2` forces the denominator to be the same type as the two parameters. (This implementation has the potential for overflow if `left` and `right` are both large enough values. There are alternative algorithms that can avoid this potential issue.)
```
tannergooding, line 36: `T.CreateChecked(2)`, `T.CreateSaturating(2)`, or `T.CreateTruncating(2)` are also ways you can achieve the `(T.One + T.One)` logic.

For more complex values, it would likely be the preferred mechanism.

docs/csharp/whats-new/tutorials/static-abstract-interface-methods.md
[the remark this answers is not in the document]
gewarren: @BillWagner Is it worth updating the file name? We'd have to add a redirect.

docs/csharp/whats-new/tutorials/static-abstract-interface-methods.md
tannergooding: `CreateChecked` - This one will throw an overflow exception if the input cannot fit in the destination
`CreateSaturating` - This one will "clamp" to `T.MinValue`/`T.MaxValue` if the input cannot fit in the destination
`CreateTruncating` - This one will "wrap" around if the input cannot fit in the destination.

For an example, given a too large value:
* `byte.CreateChecked(384)` will throw `OverflowException`
* `byte.CreateSaturating(384)` will result in `255` because it is greater than `byte.MaxValue` (which is `255`)
* `byte.CreateTruncating(384)` will result in `128` because it takes the lowest 8-bits (`384` has a hex representation of `0x0180` and the lowest 8-bits is `0x80` which is `128`)

For example, Given a too small value:
* `byte.CreateChecked(-384)` will throw `OverflowException`
* `byte.CreateSaturating(-384)` will result in `0` because it is greater than `byte.MinValue` (which is `0`)
* `byte.CreateTruncating(-384)` will result in `128` because it takes the lowest 8-bits (`384` has a hex representation of `0xFE80` and the lowest 8-bits is `0x80` which is `128`)

There are some special considerations for IEEE 754 floating-point types, like `float` and `double`, as they have special values `PositiveInfinity`, `NegativeInfinity`, and `NaN`. All three APIs behave as `Saturate` instead. Likewise, while `MinValue` and `MaxValue` represent the largest negative/positive "normal" number, the actual min/max values are `NegativeInfinity` and `PositiveInfinity`, so they clamp to these values instead.

docs/standard/generics/math.md
```diff
@@ -0,0 +1,199 @@
...
+
+```csharp
+static T Add<T>(T left, T right)
+    where T : INumber<T>
+{
+    return left + right;
+}
+```
+
+In this method, the type parameter `T` is constrained to be a type that implements the new <xref:System.Numerics.INumber%601> interface. <xref:System.Numerics.INumber%601> implements the <xref:System.Numerics.IAdditionOperators%603> interface, which contains the [+ operator](xref:System.Numerics.IAdditionOperators%603.op_Addition(%600,%601)). That allows the method to generically add the two numbers. The method can be used with any of .NET's numeric types, because they've all been updated to implement <xref:System.Numerics.INumber%601> in .NET 7.
```
tannergooding, line 21:
```suggestion
In this method, the type parameter `T` is constrained to be a type that implements the new <xref:System.Numerics.INumber%601> interface. <xref:System.Numerics.INumber%601> implements the <xref:System.Numerics.IAdditionOperators%603> interface, which contains the [+ operator](xref:System.Numerics.IAdditionOperators%603.op_Addition(%600,%601)). That allows the method to generically add the two numbers. The method can be used with any of .NET's built-in numeric types, because they've all been updated to implement <xref:System.Numerics.INumber%601> in .NET 7.
```

I think clarifying "built-in" is worthwhile as users may expect some form of "shape" matching against any .NET type instead.

docs/standard/generics/math.md
tannergooding, line 55: It's worth adding `Int128` and `UInt128` here:
```suggestion
<sup>2</sup>The binary integer types are <xref:System.Byte> (`byte`), <xref:System.Int16> (`short`), <xref:System.Int32> (`int`), <xref:System.Int64> (`long`), <xref:System.Int128>, <xref:System.IntPtr> (`nint`), <xref:System.SByte> (`sbyte`), <xref:System.UInt16> (`ushort`), <xref:System.UInt32> (`uint`), <xref:System.UInt64> (`ulong`), <xref:System.UInt128>, and <xref:System.UIntPtr> (`nuint`).
```

`System.Numerics.BigInteger` is also part of this group, but may not be as important to users.

docs/standard/generics/math.md
tannergooding: > it means that a value has a sign

This may be confusing to some users as users may consider `unsigned` types to have no sign (even though they technically have one, its just always positive).

docs/standard/generics/math.md
tannergooding: `IFormattable` and `ISpanFormattable` are also implemented by all number (and many of the other) types. They represent the inverse operation from `IParsable` and allow generating a formatted string using the `ToString()` APIs.

gewarren: @tannergooding But they aren't new, correct? And they aren't generic?

docs/csharp/whats-new/tutorials/static-abstract-interface-methods.md
gewarren: Thanks, I'll add this info. BTW, do you know why the Create* methods never come up in IntelliSense (code completion)?

PR #30394, opened 2022-07-28
Address issues from #30393

### LINKED ISSUE #30393: Address keyboard input accessibility issues
Address accessibility issues, where references to keyboard input are not using the appropriate HTML element — which causes screen readers to inconsistently represent the content's intent.

IEvangelist:
## Summary

Address accessibility issues, where references to keyboard input are not using the appropriate HTML element — which causes screen readers to inconsistently represent the content's intent.

Fixes #30393


BillWagner (approved):
Thanks again @IEvangelist 

This LGTM. (I admit I skimmed the changes.)

PR #30408, opened 2022-07-29
Add trimming breaking change

### LINKED ISSUE #30401: Breaking change for trimming in .NET 7 Preview 7
# Breaking change notice

In .NET 7, trimming will trim all assemblies in the app by default. This breaking change only affects apps which are published with `PublishTrimmed=true`, and it only affects *apps with trim warnings*.

## Description

This is a behavior change for console apps (i.e., plain .NET core apps that don't use WindowsDesktop, Android, iOS, WASM, or ASP.NET SDKs). The change is that trimming will now trim all the assemblies in the app by default, as opposed to the pre-.NET 7 behavior, which only trimmed assemblies that had opted-in with `<IsTrimmable>true</IsTrimmable>` in the library project file.

## Effects

Apps which may have previously worked with `PublishTrimmed` may not in .NET 7. Only apps *with trim warnings* may be affected. If your app has no trim warnings, the change in behavior should not cause any adverse affects, and will likely decrease the app size.

## Example

An app which uses `Newtonsoft.Json` or `System.Text.Json` without source generation to serialize and deserialize a type in the user project may function before the change because types in the user project were fully preserved. One or more trim warnings (warning codes `ILxxxx`) would have been present.

After the change, types in the user project will be trimmed and serialization may fail or produce unexpected results.

## Fixes

The best fix is to resolve all trim warnings in your application. For your own libraries, that means resolving the warnings yourself (see https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/fixing-warnings for more information). For other libraries, you may contact the author to request that they resolve the warnings, or you may choose a different library which already supports trimming.

For instance, instead of `Newtonsoft.Json` you could choose `System.Text.Json` *[with source generation](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-source-generation?pivots=dotnet-6-0)*, which  supports trimming.

## Workaround

To go back to the previous behavior, use the `TrimMode` property. The pre-.NET 7 behavior is

```
<TrimMode>partial</TrimMode>
```

The .NET 7+ behavior is

```
<TrimMode>full</TrimMode>
```

gewarren:
Fixes #30401. Related to #30279.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/core/compatibility/deployment/7.0/trim-all-assemblies?branch=pr-en-us-30408).

### DIFF
--- a/docs/core/compatibility/deployment/7.0/trim-all-assemblies.md
+++ b/docs/core/compatibility/deployment/7.0/trim-all-assemblies.md
@@ -27,7 +27,7 @@ This change can affect [source compatibility](../../categories.md#source-compati
 
 ## Reason for change
 
-This change helps to decrease app size by default instead of users having to opt in.
+This change helps to decrease app size without users having to explicitly opt in.
 
 ## Recommended action
 
--- a/docs/core/compatibility/deployment/7.0/trim-all-assemblies.md
+++ b/docs/core/compatibility/deployment/7.0/trim-all-assemblies.md
@@ -1,0 +1,54 @@
+---
+title: "Breaking change: All assemblies trimmed by default"
+description: Learn about the .NET 7 breaking change in deployment where all assemblies in a console app are trimmed by default.
+ms.date: 07/28/2022
+---
+# All assemblies trimmed by default
+
+Trimming now trims all assemblies in console apps, by default. This change only affects apps that are published with `PublishTrimmed=true`, and it only breaks apps that had existing trim warnings. It also only affects plain .NET apps that don't use the Windows Desktop, Android, iOS, WASM, or ASP.NET SDK.
+
+## Previous behavior
+
+Previously, only assemblies that were opted-in with `<IsTrimmable>true</IsTrimmable>` in the library project file were trimmed.
+
+## New behavior
+
+Starting in .NET 7, trimming trims all the assemblies in the app by default. Apps that may have previously worked with `PublishTrimmed` may not work in .NET 7. However, only apps with trim warnings will be affected. If your app has no trim warnings, the change in behavior should not cause any adverse affects, and will likely decrease the app size.
+
+For example, an app that uses `Newtonsoft.Json` or `System.Text.Json` without source generation to serialize and deserialize a type in the user project may have functioned before the change, because types in the user project were fully preserved. However, one or more trim warnings (warning codes `ILxxxx`) would have been present. Now, types in the user project are trimmed, and serialization may fail or produce unexpected results.
+
+## Version introduced
+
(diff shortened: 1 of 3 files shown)


docs/core/compatibility/deployment/7.0/trim-all-assemblies.md
```diff
@@ -0,0 +1,54 @@
...
+
+.NET 7 Preview 7
+
+## Type of breaking change
+
+This change can affect [source compatibility](../../categories.md#source-compatibility).
+
+## Reason for change
+
+This change helps to decrease app size without users having to explicitly opt in.
```
gewarren, line 30: @agocke Please check this reason as I just deduced it.

docs/core/compatibility/deployment/7.0/trim-all-assemblies.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Previous behavior
+
+Previously, only assemblies that were opted-in with `<IsTrimmable>true</IsTrimmable>` in the library project file were trimmed.
+
+## New behavior
+
+Starting in .NET 7, trimming trims all the assemblies in the app by default. Apps that may have previously worked with `PublishTrimmed` may not work in .NET 7. However, only apps with trim warnings will be affected. If your app has no trim warnings, the change in behavior should not cause any adverse affects, and will likely decrease the app size.
+
+For example, an app that uses `Newtonsoft.Json` or `System.Text.Json` without source generation to serialize and deserialize a type in the user project may have functioned before the change, because types in the user project were fully preserved. However, one or more trim warnings (warning codes `ILxxxx`) would have been present. Now, types in the user project are trimmed, and serialization may fail or produce unexpected results.
```
agocke, line 18:
```suggestion
If your app did have trim warnings you may see changes in behavior or exceptions. For example, an app that uses `Newtonsoft.Json` or `System.Text.Json` without source generation to serialize and deserialize a type in the user project may have functioned before the change, because types in the user project were fully preserved. However, one or more trim warnings (warning codes `ILxxxx`) would have been present. Now, types in the user project are trimmed, and serialization may fail or produce unexpected results.
```

docs/core/compatibility/deployment/7.0/trim-all-assemblies.md
agocke, line 30:
```suggestion
This change helps to decrease app size without users having to explicitly opt in and aligns with user expectations that the entire app is trimmed unless noted otherwise.
```

agocke (approved):
LGTM except for small wording suggestions. Thanks!

PR #30429, opened 2022-07-30
Update `HighEntropyVA` option's default value

xiang17:
The default value for `HighEntropyVA` was changed to true for all .NET Core and .NET Standard since 2018: https://github.com/dotnet/msbuild/pull/3207

For .NET Framework, it was also enabled by default if it's not neither CLR 2.0 nor .NET Framework 4.0, which is all the versions starting with .NET Framework 4.5, according to [this list of .NET Framework versions](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/versions-and-dependencies) (if we ignore .NET Framework 1.0 and 1.1).

I didn't leave out the "starting with .NET Framework 4.5" part because `.NET Framework 3.5` is still an officially supported version of [.NET Framework](https://dotnet.microsoft.com/download/dotnet-framework).

## Summary

Update doc to match with the actual default value of the `HighEntropyVA` option.

### DIFF
--- a/docs/csharp/language-reference/compiler-options/security.md
+++ b/docs/csharp/language-reference/compiler-options/security.md
@@ -71,6 +71,6 @@ The **HighEntropyVA** compiler option tells the Windows kernel whether a particu
 <HighEntropyVA>true</HighEntropyVA>
 ```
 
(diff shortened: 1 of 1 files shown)


xiang17:
CC'ing @rainersigwald, who is the original author of [the code change for this default value](https://github.com/dotnet/msbuild/pull/3207).

rainersigwald (approved):

BillWagner (approved):
Thanks for making this fix @xiang17 

I'll :shipit: now.

PR #30439, opened 2022-08-01
Unit Testing Chapter for MAUI eBook

michaelstonis:
## Summary

Unit testing chapter for maui eBook


docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+This pattern ensures that unit tests are readable, self-describing, and consistent.
+
+## Dependency injection and unit testing
+
+One of the motivations for adopting a loosely-coupled architecture is that it facilitates unit testing. One of the types registered with the dependency injection service is the `IAppEnvironmentService` interface. The following code example shows an outline of this class:
+
+```csharp
+public class OrderDetailViewModel : ViewModelBase
+{
+    private IOrderService _ordersService;
```
IEvangelist, line 37: I see that you do not mention this field in the text below. But as someone who is familiar with DI and unit testing, I'm a bit befuddled by the lack of assignment to the `_ordersService` field or at least mention of its intent.

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+
+This approach allows the `IAppEnvironmentService` object to be passed into the `OrderDetailViewModel` class at runtime, and in the interests of testability, it allows a mock class to be passed into the `OrderDetailViewModel` class at test time. The main advantage of this approach is that it enables unit tests to be executed without requiring unwieldy resources such as runtime platform features, web services, or databases.
+
+## Testing MVVM applications
+
+Testing models and view models from MVVM applications is identical to testing any other classes, and uses the same tools and techniques; this includes features such as unit testing and mocking. However, some patterns that are typical to model and view model classes can benefit from specific unit testing techniques.
```
IEvangelist, line 56:
```suggestion
Testing models and view models from MVVM applications is identical to testing any other class, and uses the same tools and techniques; this includes features such as unit testing and mocking. However, some patterns that are typical to model and view model classes can benefit from specific unit testing techniques.
```

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+        IDialogService dialogService, INavigationService navigationService, ISettingsService settingsService)
+        : base(dialogService, navigationService, settingsService)
+    {
+        _appEnvironmentService = appEnvironmentService;
+        _settingsService = settingsService;
+    }
+}
+```
+
+The `OrderDetailViewModel` class has a dependency on the `IAppEnvironmentService` type, which the dependency injection container resolves when it instantiates an `OrderDetailViewModel` object. However, rather than create an `IAppEnvironmentService` object which utilizes real servers, devices and configurations to unit test the `OrderDetailViewModel` class, instead, replace the `IAppEnvironmentService` object with a mock for the purpose of the tests.
```
IEvangelist, line 50:
```suggestion
The `OrderDetailViewModel` class has a dependency on the `IAppEnvironmentService` type, which the dependency injection container resolves when it instantiates an `OrderDetailViewModel` object. However, rather than create an `IAppEnvironmentService` object which utilizes real servers, devices and configurations to unit test the `OrderDetailViewModel` class, instead, replace the `IAppEnvironmentService` object with a mock for the purpose of the tests.
```

We shouldn't assume that everyone reading this eBook is familiar with the term "mock". Please define it as part of it's first occurrence.

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+
+This approach allows the `IAppEnvironmentService` object to be passed into the `OrderDetailViewModel` class at runtime, and in the interests of testability, it allows a mock class to be passed into the `OrderDetailViewModel` class at test time. The main advantage of this approach is that it enables unit tests to be executed without requiring unwieldy resources such as runtime platform features, web services, or databases.
+
+## Testing MVVM applications
+
+Testing models and view models from MVVM applications is identical to testing any other classes, and uses the same tools and techniques; this includes features such as unit testing and mocking. However, some patterns that are typical to model and view model classes can benefit from specific unit testing techniques.
+
+> [!TIP]
+> Test one thing with each unit test.
```
IEvangelist, line 59: Why—I mean, I know why...but your reader will be asking themselves, why. Add the reason why as part of this statement.

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+
+Don't be tempted to make a unit test exercise more than one aspect of the unit's behavior. Doing so leads to tests that are difficult to read and update. It can also lead to confusion when interpreting a failure.
+
+The eShopOnContainers mobile app uses [xUnit](https://xunit.github.io/) to perform unit testing, which supports two different types of unit tests:
+
+|Testing Type | Description |
+|---------|---------|
+| Facts | Tests that are always true, which test invariant conditions |
+| Theories | Tests that are only true for a particular set of data. |
```
IEvangelist, line 68:
```suggestion
| Testing Type | Attribute | Description                                                  |
|--------------|-----------|--------------------------------------------------------------|
| Facts        | `Fact`    | Tests that are always true, which test invariant conditions. |
| Theories     | `Theory`  | Tests that are only true for a particular set of data.       |
```

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+
+The unit tests included with the eShopOnContainers mobile app are fact tests, so each unit test method is decorated with the `[Fact]` attribute.
```
IEvangelist, line 70:
```suggestion
The unit tests included with the eShopOnContainers mobile app are fact tests, so each unit test method is decorated with the `Fact` attribute.
```

No need to annotate the attribute as it is declared in code, instead just call it `Fact`.

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+[Fact]
+public async Task OrderPropertyIsNotNullAfterViewModelInitializationTest()
+{
+    var orderService = new OrderMockService();
+    var orderViewModel = new OrderDetailViewModel(orderService);
+
+    var order = await orderService.GetOrderAsync(1, GlobalSetting.Instance.AuthToken);
+    await orderViewModel.InitializeAsync(order);
+
+    Assert.NotNull(orderViewModel.Order);
```
IEvangelist, line 86:
```suggestion
    // Arrange
    var orderService = new OrderMockService();
    var orderViewModel = new OrderDetailViewModel(orderService);

    // Act
    var order = await orderService.GetOrderAsync(1, GlobalSetting.Instance.AuthToken);
    await orderViewModel.InitializeAsync(order);

    // Assert
    Assert.NotNull(orderViewModel.Order);
```

Maybe have these comments here for the first one, with a note that establishes (and reaffirms this pattern from early mentions in this chapter), but states that they're not needed when the pattern is simple like this.

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+}
+```
+
+This unit test checks that the `Order` property of the `OrderDetailViewModel` instance will have a value after the `InitializeAsync` method has been invoked. The InitializeAsync`` method is invoked when the view model's corresponding view is navigated to. For more information about navigation, see [Navigation](navigation.md).
```
IEvangelist, line 90:
```suggestion
This unit test checks that the `Order` property of the `OrderDetailViewModel` instance will have a value after the `InitializeAsync` method has been invoked. The `InitializeAsync` method is invoked when the view model's corresponding view is navigated to. For more information about navigation, see [Navigation](navigation.md).
```

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+
+Implementing the `INotifyPropertyChanged` interface allows views to react to changes that originate from view models and models. These changes are not limited to data shown in controls -- they are also used to control the view, such as view model states that cause animations to be started or controls to be disabled.
+
+Properties that can be updated directly by the unit test can be tested by attaching an event handler to the `PropertyChanged` event and checking whether the event is raised after setting a new value for the property. The following code example shows such a test:
+
+```csharp
+[Fact]
+public async Task SettingOrderPropertyShouldRaisePropertyChanged()
+{
+    bool invoked = false;
```
IEvangelist, line 104:
```suggestion
    var invoked = false;
```

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+
+## Testing message-based communication
+
+View models that use the `MessagingCenter` class to communicate between loosely-coupled classes can be unit tested by subscribing to the message being sent by the code under test, as demonstrated in the following code example:
+
+```csharp
+[Fact]
+public void AddCatalogItemCommandSendsAddProductMessageTest()
+{
+    bool messageReceived = false;
```
IEvangelist, line 130:
```suggestion
    var messageReceived = false;
```

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+    var catalogService = new CatalogMockService();
+    var catalogViewModel = new CatalogViewModel(catalogService);
+
+    .NET MAUI.MessagingCenter.Subscribe<CatalogViewModel, CatalogItem>(
```
IEvangelist, line 134:
```suggestion
    MessagingCenter.Subscribe<CatalogViewModel, CatalogItem>(
```

Not sure what happened here, please verify that my suggestion is correct, but this will not compile.

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+    Assert.Throws<ArgumentException>(() => listView.Behaviors.Add(behavior));
+}
+```
+
+This unit test will throw an exception because the `ListView` control does not have an event named `OnItemTapped`. The `Assert.Throws<T>` method is a generic method where `T` is the type of the expected exception. The argument passed to the `Assert.Throws<T>` method is a lambda expression that will throw the exception. Therefore, the unit test will pass provided that the lambda expression throws an `ArgumentException`.
+
+> [!TIP]
+> Avoid writing unit tests that examine exception message strings.
+
+Exception message strings might change over time, and so unit tests that rely on their presence are regarded as brittle.
```
IEvangelist, line 170:
```suggestion
> Avoid writing unit tests that examine exception message strings. Exception message strings might change over time, and so unit tests that rely on their presence are regarded as brittle.
```

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+
+```csharp
+[Fact]
+public void CheckValidationPassesWhenBothPropertiesHaveDataTest()
+{
+    var mockViewModel = new MockViewModel();
+    mockViewModel.Forename.Value = "John";
+    mockViewModel.Surname.Value = "Smith";
+
+    bool isValid = mockViewModel.Validate();
```
IEvangelist, line 186:
```suggestion
    var isValid = mockViewModel.Validate();
```

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+As well as checking that validation succeeds, validation unit tests should also check the values of the `Value`, `IsValid`, and `Errors` property of each `ValidatableObject<T>` instance, to verify that the class performs as expected. The following code example demonstrates a unit test that does this:
+
+```csharp
+[Fact]
+public void CheckValidationFailsWhenOnlyForenameHasDataTest()
+{
+    var mockViewModel = new MockViewModel();
+    mockViewModel.Forename.Value = "John";
+
+    bool isValid = mockViewModel.Validate();
```
IEvangelist, line 203:
```suggestion
    var isValid = mockViewModel.Validate();
```

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
+
+    Assert.True(isValid);
+}
+```
+
+This unit test checks that validation succeeds when the two `ValidatableObject<T>` properties in the `MockViewModel` instance both have data.
+
+As well as checking that validation succeeds, validation unit tests should also check the values of the `Value`, `IsValid`, and `Errors` property of each `ValidatableObject<T>` instance, to verify that the class performs as expected. The following code example demonstrates a unit test that does this:
```
IEvangelist, line 194: Why? This goes against what you stated earlier in this chapter. You said this earlier, which kind of reads as a contradiction:

> "Don't be tempted to make a unit test exercise more than one aspect of the unit's behavior."

IEvangelist (changes requested):
Looks great, just needs a bit more work to get this one ready for merge.

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
 ---
 title: Unit Testing
-description: Unit Testing
+description: Providing application testing and improving code quality
```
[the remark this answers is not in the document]
michaelstonis: Totally understand. Sometimes it is just hard to come up with something snappy in a single sentence. This is great though!

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,223 @@
...
 author: michaelstonis
 no-loc: [MAUI]
-ms.date: 07/12/2022
+ms.date: 07/11/2022
 ---
 
-# Unit Testing
+# Unit testing
+
+Mobile apps have unique problems that desktop and web-based applications don't typically encounter. Mobile users will differ by their devices, network connectivity, availability of services, and various other factors. Therefore, mobile apps should be tested as they would be used in the real world to improve their quality, reliability, and performance. Many types of testing should be performed on an app, including unit testing, integration testing, and user interface testing. Unit testing is the most common form of testing and is essential to building high-quality applications.
```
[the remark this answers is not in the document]
michaelstonis: Sounds good me to.

docs/architecture/maui/unit-testing.md
michaelstonis: Good point. Updating the language for mocks to help better flesh it out. I looked for an docs article to link for this for a deeper dive, but I didn't find anything direct. Feels like that could be a good opportunity for a callout though.

docs/architecture/maui/unit-testing.md
michaelstonis: Expanded and added a link for additional best practices.

docs/architecture/maui/unit-testing.md
michaelstonis: I like the way that this reads far better.

docs/architecture/maui/unit-testing.md
michaelstonis: This was an accidental replacement. Updated to just `MessagignCenter...`

docs/architecture/maui/unit-testing.md
```diff
@@ -1,9 +1,227 @@
...
+    Assert.Throws<ArgumentException>(() => listView.Behaviors.Add(behavior));
+}
+```
+
+This unit test will throw an exception because the `ListView` control does not have an event named `OnItemTapped`. The `Assert.Throws<T>` method is a generic method where `T` is the type of the expected exception. The argument passed to the `Assert.Throws<T>` method is a lambda expression that will throw the exception. Therefore, the unit test will pass provided that the lambda expression throws an `ArgumentException`.
+
+> [!TIP]
+> Avoid writing unit tests that examine exception message strings. Exception message strings might change over time, and so unit tests that rely on their presence are regarded as brittle.
+
+Exception message strings might change over time, and so unit tests that rely on their presence are regarded as brittle.
```
IEvangelist, line 174:
```suggestion
```

docs/architecture/maui/unit-testing.md
IEvangelist: I think adding the additional explanation would be perfect here.

IEvangelist (approved):

docs/architecture/maui/unit-testing.md
michaelstonis: This is entirely just a code update issue. I had updated the DI process along the way here and just missed updating the variables here. Updating references now.

michaelstonis: yes, I just pushed the change up.

PR #30467, opened 2022-08-02
remove snap as option for Alpine

wfurt:
## Summary

There does not seems to be snap package for Alpine.
That leaves binaries and dotnet-isntall IMHO.


adegeo (approved):
Thanks!

PR #30505, opened 2022-08-05
Program.cs(351,17): error CS0116: A namespace cannot directly contain…

DebasisPaul:
… members such as field s, methods or statements ; Whats The Possible Solution Here.

using System;

 namespace ConsoleApp6
 {
    {
        int sum = 0;
        for (int number = 1; number < 21; number++)
        {
        if (number % 3 == 0)
        {
            sum = sum + number;
        }
        }
        Console.WriteLine($"The sum is {sum}");
    }
 }

// In the video tutorial there is no problem in this code. But when i do the same code the error come out.

## Summary

Describe your changes here.

### DIFF
--- a/docs/csharp/tour-of-csharp/tutorials/branches-and-loops.yml
+++ b/docs/csharp/tour-of-csharp/tutorials/branches-and-loops.yml
@@ -298,7 +298,6 @@ items:
     }
     Console.WriteLine($"The sum is {sum}");
     ```
-
 - title: Congratulations!
   content: |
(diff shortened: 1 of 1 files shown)


IEvangelist:
This might just be a concern with not using top-level statements.

BillWagner:
Hi @DebasisPaul 

I think what's missing is the class.  Here's the correct version:

```csharp
namespace ConsoleApp6
{
    public static class Program
    {
        {
        int sum = 0;
        for (int number = 1; number < 21; number++)
        {
            if (number % 3 == 0)
            {
                sum = sum + number;
            }
        }
        Console.WriteLine($"The sum is {sum}");}
    }
}
```

I'll close this, as the PR is an unrelated whitespace change. I hope this info helps you finish the tutorial.

PR #30521, opened 2022-08-08
Update available-counters.md

magoogli:
The 'Well-known EventCounters in .NET' documentation currently states that the GC Heap Size is measured in bytes, and provides no unit for the Working Set Event Counter.  Both actually measure megabytes as far as I can tell.

See https://github.com/dotnet/runtime/blob/4822e3c3aa77eb82b2fb33c9321f923cf11ddde6/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L79

And 

https://github.com/dotnet/runtime/blob/4822e3c3aa77eb82b2fb33c9321f923cf11ddde6/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L80


IEvangelist (approved):

PR #30530, opened 2022-08-08
Update auto-generated comment

Youssef1313:
The comment is actually unnecessary given the generated file name ends with `.g.cs`. But updating it for a form supported by Roslyn anyway.


BillWagner (approved):
This LGTM @Youssef1313 

I'll :shipit: now.

PR #30633, opened 2022-08-11
Update description of Debug Logging Provider

Nick-Stanton:
## Summary

Current described behavior is inaccurate. The DebugLogger checks if the debugger is attached before writing, so the Linux system logs would not be written to regardless.

Discovered while investigating [this reported issue](https://github.com/dotnet/runtime/issues/53459).


docs/core/extensions/logging-providers.md
```diff
@@ -80,11 +80,6 @@ The `Console` provider logs output to the console.
 
 The `Debug` provider writes log output by using the <xref:System.Diagnostics.Debug?displayProperty=fullName> class, specifically through the <xref:System.Diagnostics.Debug.WriteLine%2A?displayProperty=nameWithType> method. The <xref:Microsoft.Extensions.Logging.Debug.DebugLoggerProvider> creates <xref:Microsoft.Extensions.Logging.Debug.DebugLogger> instances, which are implementations of the `ILogger` interface.
```
tarekgh, line 81: could you please add sentence here that this only happen when the debugger is attached?

tarekgh (approved):

tarekgh:
@BillWagner  any idea how to fix the CI failure?

IEvangelist (approved):

IEvangelist:
> Reference

Hi @tarekgh - I addressed these in #30646

PR #30681, opened 2022-08-15
Fix [!NOTE] formatting

GitHubPang:
Fixed incorrect formatting of the `[!NOTE]`.
https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/exceptions/creating-and-throwing-exceptions


BillWagner (approved):
Thanks again @GitHubPang 

I'll :shipit: now.

PR #30695, opened 2022-08-15
Address other reported security alerts and build issues

IEvangelist:
Address other reported security alerts and build issues.

Fixes all of the recently reported [dependabot alerts](https://github.com/dotnet/docs/security/dependabot).


BillWagner (approved):
This LGTM @IEvangelist 

You can :shipit: when ready.

PR #30702, opened 2022-08-16
Added example for generic types with interfaces or class inheritance

### LINKED ISSUE #18948: Provide example for usage of generic type constraints with interface or class inheritance
Need example for syntax of declaring a class that implements an interface or inherits a class and also uses generic type constraint.
Following code can be shown as an example:
```csharp
public class MyClass<T>: IEnumerable where T : class { }
```

---
#### Document Details

* ID: 74dfcde5-aa2b-1128-fa29-f198a6e437b7
* Version Independent ID: 59aeccf5-364c-77a8-88dc-cb529b76aebc
* Content: [where (generic type constraint) - C# Reference](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/where-generic-type-constraint)
* Content Source: [docs/csharp/language-reference/keywords/where-generic-type-constraint.md](https://github.com/dotnet/docs/blob/master/docs/csharp/language-reference/keywords/where-generic-type-constraint.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

mducharm:
## Summary

Added example for generic types with interfaces or class inheritance. 

I put it here so as to not detract from the initial definition and examples of the `where` clause, but let me know if it should be placed elsewhere.

Fixes #18948

### DIFF
--- a/docs/csharp/language-reference/keywords/where-generic-type-constraint.md
+++ b/docs/csharp/language-reference/keywords/where-generic-type-constraint.md
@@ -14,7 +14,7 @@ helpviewer_keywords:
 ---
 # where (generic type constraint) (C# Reference)
 
-The `where` clause in a generic definition specifies constraints on the types that are used as arguments for type parameters in a generic type, method, delegate, or local function. Constraints can specify interfaces, base classes, or require a generic type to be a reference, value, or unmanaged type. They declare capabilities that the type argument must have.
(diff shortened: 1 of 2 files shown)


BillWagner:
Hi @mducharm 

I like the idea behind this addition. However, rather than the specific examples, I think this might be better as added text in the opening paragraphs. Something to the effect of:

> The `where` clauses for any type parameters must be placed after any declared base class or implemented interfaces.

What do you think?

mducharm:
Hey there - that makes sense to me. I made an adjustment, let me know if this fits with what you had in mind.

BillWagner (approved):
Thanks again @mducharm 

PR #30710, opened 2022-08-16
Add source code for a custom YAML config provider implementation

IEvangelist:
Add source code for a custom YAML config provider implementation.


BillWagner (approved):
Hi @IEvangelist 

The code here is all good.  I didn't see any references from any current articles. Is that coming in a different PR? Or did you forget to add the link here?

IEvangelist:
> Hi @IEvangelist
> 
> The code here is all good. I didn't see any references from any current articles. Is that coming in a different PR? Or did you forget to add the link here?

I was going to write content on this later, but had to deprioritize it for now. I'll resync .NET sample code soon so that this is at least part of the samples browser.

PR #30738, opened 2022-08-18
C# reference: Consolidated articles about checked and unchecked statements

pkulikov:
@BillWagner while working on this, I've also found two typos in the spec: dotnet/csharpstandard#612


docs/csharp/language-reference/builtin-types/numeric-conversions.md
```diff
@@ -77,7 +77,7 @@ The following table shows the predefined explicit conversions between the built-
 
 Also note that:
 
-- When you convert a value of an integral type to another integral type, the result depends on the overflow [checking context](../keywords/checked-and-unchecked.md). In a checked context, the conversion succeeds if the source value is within the range of the destination type. Otherwise, an <xref:System.OverflowException> is thrown. In an unchecked context, the conversion always succeeds, and proceeds as follows:
+- When you convert a value of an integral type to another integral type, the result depends on the [overflow-checking context](../statements/checked-and-unchecked.md). In a checked context, the conversion succeeds if the source value is within the range of the destination type. Otherwise, an <xref:System.OverflowException> is thrown. In an unchecked context, the conversion always succeeds, and proceeds as follows:
```
BillWagner, line 80: ping @tannergooding on this change. There are several locations in this content where we state that in a `checked` context, "an `OverflowException` occurs". Is this still always true, now that developers can write `checked` operators for their type? I'll tag the other locations I found.

If we should change the language, I'll suggest something like:

```suggestion
- When you convert a value of an integral type to another integral type, the result depends on the [overflow-checking context](../statements/checked-and-unchecked.md). In a checked context, the conversion succeeds if the source value is within the range of the destination type. Otherwise, an <xref:System.OverflowException> is thrown. for all built-in integral types. User defined numeric types define the behavior in a `checked` operator. In an unchecked context, the conversion always succeeds, and proceeds as follows:
```

docs/csharp/language-reference/compiler-options/language.md
```diff
@@ -23,13 +23,13 @@ The following options control how the compiler interprets language features. The
 
 ## CheckForOverflowUnderflow
 
-The **CheckForOverflowUnderflow** option specifies whether an integer arithmetic statement that results in a value that is outside the range of the data type causes a run-time exception.  
+The **CheckForOverflowUnderflow** option specifies whether an integer arithmetic expression that results in a value that is outside the range of the destination type causes a run-time exception.
```
BillWagner, line 26: Same comment.

docs/csharp/language-reference/compiler-options/language.md
```diff
@@ -23,13 +23,13 @@ The following options control how the compiler interprets language features. The
...
+An integer arithmetic expression that is in the scope of a [`checked` or `unchecked` keyword](../statements/checked-and-unchecked.md) isn't subject to the effect of the **CheckForOverflowUnderflow** option. If an integer arithmetic expression that isn't in the scope of a `checked` or `unchecked` keyword results in a value outside the range of the destination type, and **CheckForOverflowUnderflow** is `true`, that expression causes an <xref:System.OverflowException> at run time. If **CheckForOverflowUnderflow** is `false`, that expression doesn't cause an exception at run time. The default value for this option is `false`, that is, overflow checking is disabled.
```
BillWagner, line 32: Same comment,

docs/csharp/language-reference/statements/checked-and-unchecked.md
```diff
@@ -0,0 +1,65 @@
...
+  > [!NOTE]
+  > When you convert a `decimal` value to an integral type and the result is outside the range of the destination type, an <xref:System.OverflowException> is always thrown, regardless of the overflow-checking context.
+
+- Beginning with C# 11, user-defined checked operators and conversions. For more information, see the [User-defined checked operators](../operators/arithmetic-operators.md#user-defined-checked-operators) section of the [Arithmetic operators](../operators/arithmetic-operators.md) article.
+
+## Default overflow-checking context
+
+If you don't specify the overflow-checking context, the value of the [**CheckForOverflowUnderflow**](../compiler-options/language.md#checkforoverflowunderflow) compiler option defines the default context for non-constant expressions. By default the value of that option is unset and integral-type arithmetic operations and conversions are executed in an **unchecked** context.
+
+Constant expressions are evaluated by default in a checked context and a compile-time error occurs in the case of an overflow. You can explicitly specify an unchecked context for a constant expression with the `unchecked` statement or operator.
```
BillWagner, line 53: IIRC, this behavior didn't change with the new checked and unchecked operators. Tagging @tannergooding to confirm. (If my memory is correct, this is fine as written)

BillWagner (commented):
Thanks again @pkulikov 

I believe this is all fine. I had a couple questions that I'd like Tanner to verify, and then it will be ready.

docs/csharp/language-reference/builtin-types/numeric-conversions.md
pkulikov: Thank you @BillWagner 
I don't think we need to change this text as this article is already about *built-in* numeric conversions between *built-in* types. Also, "user-defined numeric types" may be a niche use case => mentioning it can be more confusing. I would wait on user feedback to those articles.

As for the original queston:

>  Is this still always true, now that developers can write checked operators for their type?

The docs *recommend* that user-defined checked operators throw, but it's not guaranteed, so it's possible to observe no exception in case of overflow in a checked user-defined operator.

docs/csharp/language-reference/compiler-options/language.md
pkulikov: However, here we should mention user-defined checked operators and possible no-exception case. I'll think about how to do that.

docs/csharp/language-reference/builtin-types/numeric-conversions.md
tannergooding: I think I'd agree with Petr here. Even with user-defined types, there is no guarantee that "overflow" will happen. Some cases like `float` or `double` always "saturate" instead for example.

It really is type and input dependent.

docs/csharp/language-reference/statements/checked-and-unchecked.md
tannergooding: Yep. C# has always evaluated `const T = expr` as if it were checked and want's you to explicitly `uncheck(expr)` in the case it would overflow.

docs/csharp/language-reference/builtin-types/numeric-conversions.md
BillWagner: Thanks both. I'll mark this as resolved.

docs/csharp/language-reference/compiler-options/language.md
pkulikov: @BillWagner at the end, I've rewritten this section. Please check my latest `Addressed feedback` commit

docs/csharp/language-reference/statements/checked-and-unchecked.md
```diff
@@ -0,0 +1,68 @@
...
+
+:::code language="csharp" interactive="try-dotnet-method" source="snippets/checked-and-unchecked/Program.cs" id="MainExample":::
+
+When integer arithmetic overflow occurs, the overflow-checking context defines what happens as follows:
+
+- In a checked context, a <xref:System.OverflowException?displayProperty=nameWithType> is thrown; if overflow happens in a constant expression, a compile-time error occurs.
+- In an unchecked context, the operation result is truncated by discarding any high-order bits that don't fit in the destination type. For example, in the case of addition it wraps from the maximum value to the minimum value, as the preceding example shows.
+
+> [!NOTE]
+> The behavior of user-defined operators and conversions in the case of the overflow of the corresponding result type can differ from the one described in the previous paragraph. In particular, [user-defined checked operators](../operators/arithmetic-operators.md#user-defined-checked-operators) might throw no exception in a checked context.
```
BillWagner, line 26: nit on sentence structure. Then, I'll :shipit:

```suggestion
> The behavior of user-defined operators and conversions in the case of the overflow of the corresponding result type can differ from the one described in the previous paragraph. In particular, [user-defined checked operators](../operators/arithmetic-operators.md#user-defined-checked-operators) might not throw an exception in a checked context.
```

docs/csharp/language-reference/compiler-options/language.md
BillWagner: I had one minor grammar nit, and then this is ready.

Thanks again @pkulikov

BillWagner (approved):
Thanks again @pkulikov 

This is great work, and I'll :shipit: as soon as the build completes.

PR #30765, opened 2022-08-19
Delete no-response.yml

gewarren:
Follow up to #29704. We currently have two versions of no-response.yml. I'm deleting the older one.


BillWagner (approved):
This LGTM @gewarren 

I'll :shipit:

PR #30797, opened 2022-08-23
Correct the note about the .net 5 SDK EOL

marcpopMSFT:
## Summary

We're actually supporting 5.0.2xx until October and 5.0.4xx until December in VS only scenarios so updating the comment about that.


tdykstra (approved):
LGTM

PR #30871, opened 2022-08-26
Fix programs in "specify-era"

GitHubPang:
## Summary

Fix the *third* program in the ["Instantiate a date with an era" section of the "Work with calendars" page](https://docs.microsoft.com/en-us/dotnet/standard/datetime/working-with-calendars#instantiate-a-date-with-an-era).

1. Use the variables `date2` and `date3`, which were defined but unused.
2. Use `M` instead of `m` for month, since `m` is for minute and `M` is for month for *custom date and time formats*, according to [the doc](https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings).
3. Add closing parentheses in the output.


BillWagner (approved):
Thanks again @GitHubPang 

PR #30898, opened 2022-08-28
Remove extra comma at the end of switch expression

alireza-rezaee:
The extra commas at the end of the last case were removed from the switch expression(s).


IEvangelist:
Hi @alireza-rezaee - thank you for posting this PR, however; this is a style preference. It's not going to break anything by leaving them. I'm curious what @BillWagner's take is on this. Personally, I prefer to omit them, as you've done here - but I'm not sure what style preference we've employed here.

BillWagner:
Hi @alireza-rezaee @IEvangelist 

It is a style question. My preference has been to leave the optional comma in place. My reason for this is that I often simplify switch expressions to combine multiple arms, or rearrange the order to make sure none of the cases are subsumed by earlier cases. 

It's a mild preference, I'm not ready to add it to the coding style guidelines, one way or the other.

For now, I'll close this and leave the current style in place.

PR #30923, opened 2022-08-30
Add a few bits of additional context to address #30862

### LINKED ISSUE #30862: Stacked Usings Section is confusing to me
I'm really confused about what issue you are trying to bring to our attention within this section. Here's a couple of questions.

1. Is the problem you are brining up here inherit to all stacked usings or just await usings?

2. Is this only a problem if you use ConfigureAwait? So if you aren't using ConfigureAwait you can use stacked usings without worry?

3. Within the "Unacceptable Pattern", you seem to be implying that there's an inherent issue with stacked usings such that we would otherwise expect the DisposeAysnc to be called but because of this issue it won't be. But in the code I'd never expect the DisposeAysnc methods to be called since the exception is happening before the await usings. This does not seem to be problem with stacked usings at all. I don't understand how this code is an example of using stacked usings the wrong way. 

---
#### Document Details

* ID: 61a961dc-3ebf-2a96-a6eb-2e0b62ab909f
* Version Independent ID: d72276e1-9059-815f-dc14-a03a7544ab87
* Content: [Implement a DisposeAsync method](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync)
* Content Source: [docs/standard/garbage-collection/implementing-disposeasync.md](https://github.com/dotnet/docs/blob/main/docs/standard/garbage-collection/implementing-disposeasync.md)
* GitHub Login: @IEvangelist
* Microsoft Alias: **dapine**

IEvangelist:
## Summary

Add a few bits of additional context to address #30862

Fixes #30862

### DIFF
--- a/docs/standard/garbage-collection/implementing-disposeasync.md
+++ b/docs/standard/garbage-collection/implementing-disposeasync.md
@@ -146,7 +146,7 @@ The highlighted lines in the following code show what it means to have "stacked 
 :::code language="csharp" id="dontdothis" source="snippets/dispose-async/ExamplePatterns.cs" highlight="9-10":::
 
 > [!TIP]
-> Avoid this pattern as it could lead to unexpected behavior.
+> Avoid this pattern as it could lead to unexpected behavior. If you use one of the acceptable patterns, the problem of undisposed objects is non-existent. The clean-up operations are correctly performed when `using` statements aren't stacked.
 
 ## See also
 
--- a/docs/standard/garbage-collection/implementing-disposeasync.md
+++ b/docs/standard/garbage-collection/implementing-disposeasync.md
@@ -3,7 +3,7 @@ title: Implement a DisposeAsync method
 description: Learn how to implement DisposeAsync and DisposeAsyncCore methods to perform asynchronous resource cleanup.
 author: IEvangelist
 ms.author: dapine
-ms.date: 07/26/2022
+ms.date: 08/30/2022
 dev_langs:
   - "csharp"
 helpviewer_keywords:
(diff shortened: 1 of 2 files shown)


gewarren:
@gbrunton Does this clarification help?

docs/standard/garbage-collection/implementing-disposeasync.md
```diff
@@ -141,7 +141,7 @@ In the preceding example, each asynchronous clean up operation is implicitly sco
 
 ### Unacceptable pattern
 
-The highlighted lines in the following code show what it means to have "stacked usings". If an exception is thrown from the `AnotherAsyncDisposable` constructor, neither object is properly disposed of. The variable `objTwo` is never assigned because the constructor did not complete successfully. As a result, the constructor for `AnotherAsyncDisposable` is responsible for disposing any resources allocated before it throws an exception.
+The highlighted lines in the following code show what it means to have "stacked usings". If an exception is thrown from the `AnotherAsyncDisposable` constructor, neither object is properly disposed of. The variable `objTwo` is never assigned because the constructor did not complete successfully. As a result, the constructor for `AnotherAsyncDisposable` is responsible for disposing any resources allocated before it throws an exception. If the `ExampleAsyncDisposable` type has a finalizer, it is eligible for finalization. This problem is non-existent when using one of the acceptable patterns, as the clean up operations are correctly performed when they're not stacked.
```
gewarren, line 144:
```suggestion
The highlighted lines in the following code show what it means to have "stacked usings". If an exception is thrown from the `AnotherAsyncDisposable` constructor, neither object is properly disposed of. The variable `objTwo` is never assigned because the constructor did not complete successfully. As a result, the constructor for `AnotherAsyncDisposable` is responsible for disposing any resources allocated before it throws an exception. If the `ExampleAsyncDisposable` type has a finalizer, it's eligible for finalization.
```

docs/standard/garbage-collection/implementing-disposeasync.md
```diff
@@ -141,7 +141,7 @@ In the preceding example, each asynchronous clean up operation is implicitly sco
...
 
 :::code language="csharp" id="dontdothis" source="snippets/dispose-async/ExamplePatterns.cs" highlight="9-10":::
```
gewarren, line 147: Then I'd put the bit about acceptable patterns as a second bullet point in the TIP block that starts on line 148.

```
- If you use one of the acceptable patterns, the problem of undisposed objects is non-existent. The clean-up operations are correctly performed when `using` statements aren't stacked.
```

BillWagner (approved):
This LGTM @IEvangelist 

You can :shipit: when ready.

gbrunton:
> @gbrunton Does this clarification help?

Do you think this update of documentation answered my 3 questions that I asked you within my issue?

1. Is the problem you are brining up here inherit to all stacked usings or just await usings?

2. Is this only a problem if you use ConfigureAwait? So if you aren't using ConfigureAwait you can use stacked usings without worry?

3. Within the "Unacceptable Pattern", you seem to be implying that there's an inherent issue with stacked usings such that we would otherwise expect the DisposeAysnc to be called but because of this issue it won't be. But in the code I'd never expect the DisposeAysnc methods to be called since the exception is happening before the await usings. This does not seem to be problem with stacked usings at all. I don't understand how this code is an example of using stacked usings the wrong way.

PR #30953, opened 2022-08-31
C# nameof: minor edits

pkulikov:
- Updated example to show various uses
- Made the language less spec-like


BillWagner (approved):
Thanks @pkulikov 

This is a nice improvement. I'll :shipit: now.

Issue #19562, opened 2020-07-17
IFileInfo: don't use physical path anymore when serving from packfiles


jhudsoncedaron:
## IFileInfo: don't use physical path anymore when serving from packfiles

When serving content from packfiles, returning the path of the packfile in IFileInfo.PhysicalPath (for filesystem watcher) doesn't work anymore. It now serves from the physical path rather than the file stream it was given.

In reading the code, I'm convinced the change is intentional.

### Version introduced

3.0

### Old behavior

The fixed part of the static file server in Kestrel would serve the static file using the stream returned in IFileInfo whether or not the path is included. It looks like the path gets used for setting up a FileSystemWatcher that gets used for something though.

### New behavior

The fixed part of the static file server in Kestrel (Microsoft.AspNetcoer.StaticFiles.StaticFileContext ) would serve the static file using the stream returned in IFileInfo only if the path is null.

### Reason for change

Not known to me.

### Recommended action

Don't fill IFileInfo.PhysicalPath unless the physical path refers to a file that contains the exact content to return. Providing the path that could be used to derive the content (for If-Modified-Since or related things) doesn't work.

### Category

- ASP.NET Core

### Affected APIs

Microsoft.Extension.FileProviders.IFileInfo.PhysicalPath

---
#### Issue metadata

* Issue type: breaking-change

danmosemsft:
@maryamariyan is this your area - can you review?

maryamariyan:
yes. It's my area. OK.

maryamariyan:
@jhudsoncedaron do you mind sharing a repro code for this?

jhudsoncedaron:
[pacsrv.zip](https://github.com/dotnet/docs/files/4955730/pacsrv.zip)

Demonstration; see PackFileProvider.cs line 94.

gewarren:
@scottaddie I just noticed this is an ASP.NET Core breaking change. Do you want to take it?

IEvangelist:
Closing as this is dated back to a version that is out-of-support.

Issue #18627, opened 2020-05-25
Desktop runtime missing.


estebangaro:
Please, update "composition" section in order to include Microsoft.WindowsDesktop.App runtime released at NET Core 3.1
---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 581f6cbb-c666-87ee-ab0c-f131baca403c
* Version Independent ID: 9205f603-e1d4-e0eb-1273-f176fcbf520f
* Content: [.NET Core overview](https://docs.microsoft.com/en-us/dotnet/core/about#support)
* Content Source: [docs/core/about.md](https://github.com/dotnet/docs/blob/master/docs/core/about.md)
* Product: **dotnet-core**
* GitHub Login: @tdykstra
* Microsoft Alias: **dotnetcontent**

tdykstra:
cc @Thraka

Thraka:
You are correct, this is missing. Thank you for pointing it out.

PR #30979, opened 2022-09-02
Enable stale Action

gewarren:
Adds stale label after 6 months of no activity and closes after 7 days of adding the label.
Excludes issues with labels `breaking-change` and `Pri1`. Any others?
This should not act on PRs because we're not explicitly giving write permissions to PRs.
Also, this initial configuration sets `debug-only` to true, so it will just do dry runs until that line is deleted/changed to false.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,7 +1,7 @@
 name: Close stale issues
 
 on:
-  schedule: # Run 5 minutes after the hour every hour.
+  schedule: # Run 5 minutes after midnight daily.
   - cron: '5 0 * * *'
 
 jobs:
@@ -14,13 +14,14 @@ jobs:
     steps:
     - uses: actions/stale@v4.1.0
       with:
+        start-date: '2022-03-01T00:00:00Z' # ISO 8601 or RFC 2822
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         days-before-stale: 180
-        days-before-close: 7
+        days-before-close: 14
         stale-issue-label: stale
         close-issue-message: >
             This issue has been automatically closed due to lack of activity.
             If you feel this issue is still important, please reopen it and leave a comment.
-        exempt-issue-labels: breaking-change,Pri0
+        exempt-issue-labels: 'breaking-change,Pri0,:watch: Not Triaged,:checkered_flag: Release: .NET 7,:checkered_flag: Release: .NET 8'
         operations-per-run: 50
         debug-only: true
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+
+    steps:
+    - uses: actions/stale@v4.1.0
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-stale: 120
```
gewarren, line 18: Adds the label after no action for 120 days. But we can change this...

BillWagner:
If I understand this correctly, any issue that ages 120 days would get the "needs-more-info" label, and then be closed in an additional 14 days, correct? 

I want to make sure I understand its workflow correctly.

gewarren:
> If I understand this correctly, any issue that ages 120 days would get the "needs-more-info" label, and then be closed in an additional 14 days, correct?

Yes, that's correct. And that doesn't sound like a good label to use now that you put it that way. Perhaps we do need both the no-response bot (or similar - perhaps FabricBot has this functionality) *and* the stale bot.

IEvangelist (approved):

BillWagner (approved):
This is a great addition @gewarren 

Let's :shipit:

ocdtrekkie:
I would strongly discourage the use of a stale bot on Microsoft Docs repos. As an outside user, I've often found the handling delay is the correct Microsoft employee reaching the issue, not the outside user. Perhaps rather than incorrectly closing issues that were never handled by staff, the bot should increasingly aggressively find a stakeholder to address the issue?

At a random selection of old issues on this repo: https://github.com/dotnet/docs/issues/19562 the requestor answered all requests promptly, providing sample code and all, and it took five months after that step, for someone to recommend someone else review it... who just never showed up at all, and now it's been nearly two years.

Which is to say, stale bots should only be used if they're successful in only targeted issue abandonment by outside users, not internal staff.

Note that the above issue has a tag your stale bot will ignore, so... I picked another at random: https://github.com/dotnet/docs/issues/18627 Two years ago, a Microsoft staff member confirmed the issue was valid, and then nothing was done with it. So it's a confirmed actual issue, unaddressed, and I believe this new stale action will close this issue.

gewarren:
@ocdtrekkie Thanks for your input. I totally agree that the handling delay is usually on our side (Microsoft docs), and all that means is that we simply don't have the time to address the issue when prioritized with all the other issues that also need attention. It's definitely not ideal to close out issues that are still relevant simply because they're old, but it does help us to focus on the most important issues. Any issues closed via this automation will include a note that says to reopen it if still deemed important enough - so the hope is that the bot will help to decrease the number of active issues, and also to resurface forgotten issues that shouldn't have been forgotten.

We have a separate "no-response" workflow that deals with issues that require more information from the original poster in order to be actionable. This "stale" workflow serves a different purpose.

PR #30989, opened 2022-09-02
Update numerics.md

psyarhei:
Fix Minimum value of System.IntPtr (in 32-bit process)

### DIFF
--- a/docs/standard/numerics.md
+++ b/docs/standard/numerics.md
@@ -35,7 +35,7 @@ ms.assetid: dfebc18e-acde-4510-9fa7-9a0f4aa3bd11
 |<xref:System.Int32?displayProperty=nameWithType>|4|-2,147,483,648|2,147,483,647|
 |<xref:System.Int64?displayProperty=nameWithType>|8|-9,223,372,036,854,775,808|9,223,372,036,854,775,807|
 |<xref:System.SByte?displayProperty=nameWithType>|1|-128|127|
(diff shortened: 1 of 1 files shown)


IEvangelist:
Hi @sergeipetrashko8 - thank you for posting this issue. I was curious about this, so after looking into it a bit more, I agree with you. In the runtime repo, in the `IntPtr` code, there is an alias [named `nint_t`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/IntPtr.cs#L19) that uses a compiler directive to determine whether to use 32 or 64 bit. When 32 bit, the alias is represents the `Int32` type. The `IntPtr.MinValue` gets the corresponding `nint_t.MinValue` (in the case of 32-bit, [that would be `Int32.MinValue`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/IntPtr.cs#L174-L179).

And for `Int32.MinValue` that is in fact `-2,147,483,648`, https://docs.microsoft.com/dotnet/api/system.int32.minvalue?view=net-6.0. I also found this SO answer interesting, https://stackoverflow.com/a/31688652/2410379. Thank you for this!

IEvangelist (approved):

CamSoper:
Verified in Raspberry Pi OS (32-bit) using .NET SDK 6.0.400.
image

PR #30994, opened 2022-09-02
Correct event pattern content to align with guides

### LINKED ISSUE #13906: Article doesn't follow important event conventions
Please implement the requirements from your other article https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/event in this one. This article does not follow a few important conventions. 

The problem starts in the first code snipped below. This has the name "On"EventRaised which should NOT be for the delegate "On" SHOULD be for the protected virtual method to raise each event.

```
void OnEventRaised(object sender, EventArgs args);
```

Instead the delegate should be something like:

```
void EventHandler(object sender, EventArgs args);
```

Then you need to add to this article the method for raising events "OnFileFound" and stop invoking FileFound directly. This is more than necessary if classes are ever derived and the whole framework more-or-less follows this convention.

Also, don't just add the method. Explain why it's added (for derived classes) and link to your other article on it that does things correctly: https://docs.microsoft.com/en-us/dotnet/standard/events/

---
#### Document Details

* ID: 90b37f1a-cd59-d9e5-067b-30ea5f26e5cc
* Version Independent ID: d2078536-1ee3-5b60-3872-f2635d245cc2
* Content: [Standard .NET event patterns](https://docs.microsoft.com/en-us/dotnet/csharp/event-pattern#feedback)
* Content Source: [docs/csharp/event-pattern.md](https://github.com/dotnet/docs/blob/master/docs/csharp/event-pattern.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

IEvangelist:
## Summary

- Edit pass
- Format
- Address customer feedback on event patterns

Fixes #13906

## Preview

https://review.docs.microsoft.com/en-us/dotnet/csharp/event-pattern?branch=pr-en-us-30994


BillWagner (approved):
Thanks @IEvangelist 

This LGTM. Let's :shipit:

PR #31018, opened 2022-09-06
Fix emdash in versioning-sdk-msbuild-vs.md

akoeplinger:
It used the wrong format which made it render as the literal `&mdash&`


BillWagner (approved):
Thanks @akoeplinger 

This LGTM. I'll :shipit: now.

PR #31022, opened 2022-09-06
embedded nulls cause issues with event payload

### LINKED ISSUE #52025: Strings with embedded nulls cause issues with event payload decoding
Manifest driven `EventSource`s [mark all strings as `win:UnicodeString` in the manifest](https://github.com/dotnet/runtime/blob/d1fed288f58d89ebc2fbafc9f6aa79c81fbb5611/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs#L5955-L6003). This is interpreted by TraceEvent and other pieces of tracing infrastructure as "a null-terminated string". There is a separate type for "counted strings" that isn't used by `EventSource`.

C#, however, allows for embedded nulls in strings, and `String.Length` will give the full length of string _including_ the embedded nulls.

`EventSource` uses `String.Length` when it is encoding strings, e.g.,
https://github.com/dotnet/runtime/blob/c37257f0a073fed898ed12ce14b6a3ec39d991f2/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs#L1089-L1091

This can lead to problems for the parser if there are things in the payload _after_ the string. For example:

```
public class MySource : EventSource
{
    public static MySource Instance = new MySource();

    public void MyEvent(int foo, string bar, int baz) => WriteEvent(1, foo, bar, baz);
}

// ...

var myString = "aaa$bbb";
myString = myString.Replace('$', '\0');
MySource.Instance.MyEvent(10, myString, -10);
```

will result in the following at the reader end:
image

The event contains the full data:
```
<Event MSec=  "4096.7191" PID="24292" PName="EventSourceNullContent" TID="31320" EventName="MyEvent"
  TimeStamp="04/26/21 17:14:07.575276" ID="1" Version="0" Keywords="0x0000F00000000000" TimeStampQPC="9,698,287,562,667" QPCTime="0.100us"
  Level="Informational" ProviderName="MySource" ProviderGuid="a61ea624-4944-55fc-c2a8-37838829438d" ClassicProvider="False" ProcessorNumber="7"
  Opcode="0" Task="65533" Channel="0" PointerSize="8"
  CPU="7" EventIndex="70874" TemplateType="DynamicTraceEventData">
  <PrettyPrint>
    <Event MSec=  "4096.7191" PID="24292" PName="EventSourceNullContent" TID="31320" EventName="MyEvent" ProviderName="MySource" foo="10" bar="aaa" baz="6,422,626"/>
  </PrettyPrint>
  <Payload Length="24">
       0:   a  0  0  0 61  0 61  0 | 61  0  0  0 62  0 62  0   ....a.a. a...b.b.
      10:  62  0  0  0 f6 ff ff ff |                           b....... 
  </Payload>
</Event>
```

TraceEvent attempts to read a null terminated string, and then interprets the next 4 bytes as the `baz` parameter which is incorrect.

A potential solution would be to truncate input strings to the first null encountered. I imagine this is a very rare occurrence and the overhead of running `String.IndexOf((char)0)` should be minimal for the common use case. We could also push and out of band message that an event payload string was truncated.

CC @sywhang @noahfalk

mikelle-rogers:
decoding.

Adding notes regarding how embedded nulls cause issues with event payload decoding. 
Fixes [52025](https://github.com/dotnet/runtime/issues/52025)

### DIFF
--- a/docs/core/diagnostics/eventsource-instrumentation.md
+++ b/docs/core/diagnostics/eventsource-instrumentation.md
@@ -86,12 +86,7 @@ end of the event declaration, that is, at the end of the list of method paramete
 new ID to replace the old one.
 8. When declaring events methods, specify fixed-size payload data before variably sized data.
 
->[!NOTE]
->Manifest driven `EventSource`s [mark all strings as `win:UnicodeString` in the manifest](https://github.com/dotnet/runtime/blob/d1fed288f58d89ebc2fbafc9f6aa79c81fbb5611/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs#L5955-L6003). This is interpreted by TraceEvent and other pieces of tracing infrastructure as "a null-terminated string". There is a separate type for "counted strings" that isn't used by `EventSource`.
->C#, however, allows for embedded nulls in strings, and `String.Length` will give the full length of string _including_ the embedded nulls.
->`EventSource` uses `String.Length` when it is encoding strings.
->This can lead to problems for the parser if there are things in the payload _after_ the string.
-
+9. Do not use strings containing null characters. When generating the manifest for ETW EventSource will declare all strings as null terminated, even though it is possible to have a null character in a C# String. If a string contains a null character the entire string will be written to the event payload, but any parser will treat the first null character as the end of the string. If there are payload arguments after the string, the remainder of the string will be parsed instead of the intended value.
 ## Typical event customizations
 
 ### Setting event verbosity levels
--- a/docs/core/diagnostics/eventsource-instrumentation.md
+++ b/docs/core/diagnostics/eventsource-instrumentation.md
@@ -86,6 +86,12 @@ end of the event declaration, that is, at the end of the list of method paramete
 new ID to replace the old one.
 8. When declaring events methods, specify fixed-size payload data before variably sized data.
 
+>[!NOTE]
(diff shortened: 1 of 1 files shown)


docs/core/diagnostics/eventsource-instrumentation.md
```diff
@@ -86,6 +86,12 @@ end of the event declaration, that is, at the end of the list of method paramete
 new ID to replace the old one.
 8. When declaring events methods, specify fixed-size payload data before variably sized data.
 
+>[!NOTE]
+>Manifest driven `EventSource`s [mark all strings as `win:UnicodeString` in the manifest](https://github.com/dotnet/runtime/blob/d1fed288f58d89ebc2fbafc9f6aa79c81fbb5611/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs#L5955-L6003). This is interpreted by TraceEvent and other pieces of tracing infrastructure as "a null-terminated string". There is a separate type for "counted strings" that isn't used by `EventSource`.
+>C#, however, allows for embedded nulls in strings, and `String.Length` will give the full length of string _including_ the embedded nulls.
+>`EventSource` uses `String.Length` when it is encoding strings.
+>This can lead to problems for the parser if there are things in the payload _after_ the string.
+
```
davmason, line 94:
```suggestion
9. When generating the manifest for ETW EventSource will declare all strings as null terminated, even though it is possible to have a null character in a C# String. If a string contains a null character the entire string will be written to the event payload, but any parser will treat the first null character as the end of the string. If there are payload arguments after the string, the remainder of the string will be parsed instead of the intended value.\
When working with EventSource it is strongly recommended to not use strings containing null characters in any event payload.
```

mikelle-rogers: Thank you David!

tommcdon (approved):
Lgtm, thanks!

PR #31035, opened 2022-09-07
Add note on building native libraries

MichalStrehovsky:
Just two quick paragraphs in the overview doc.

The sample largely overlaps so I didn't make a full chapter on this. We could make a full chapters depending on how reviewers feel about it.

Question to reviewers: should we also update the docs for UnmanagedCallersOnly attribute and add a note on NativeAOT? UnmanagedCallersOnlyAttribute.EntryPoint is currently a no-op anywhere but Native AOT.

cc @dotnet/ilc-contrib

### DIFF
--- a/docs/core/deploying/native-aot/index.md
+++ b/docs/core/deploying/native-aot/index.md
@@ -96,6 +96,12 @@ The first release of native AOT in .NET 7 has additional limitations. These incl
 - Not all the runtime libraries are fully annotated to be native AOT compatible (that is, some warnings in the runtime libraries are not actionable by end developers).
 - Limited diagnostic support (debugging and profiling).
 
+## Building native libraries
+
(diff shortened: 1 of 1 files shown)


jkotas (approved):

docs/core/deploying/native-aot/index.md
```diff
@@ -96,6 +96,12 @@ The first release of native AOT in .NET 7 has additional limitations. These incl
 - Not all the runtime libraries are fully annotated to be native AOT compatible (that is, some warnings in the runtime libraries are not actionable by end developers).
 - Limited diagnostic support (debugging and profiling).
 
+## Building native libraries
```
IEvangelist, line 99:
```suggestion
## Build native libraries
```

PR #31066, opened 2022-09-08
Remove feature spec

BillWagner:
The PR for `private protected` has been merged into the C# Standard for V7.n. So, remove the private-protected feature spec and redirect traffic to the updated clause in the standard.

See dotnet/csharpstandard#215


tdykstra (approved):
LGTM

PR #31079, opened 2022-09-08
DataContractSerializer breaking change

### LINKED ISSUE #30722: [Breaking change]: DataContractSerializer retains negative when reading "-0" as a float/double
### Description

Going back to the NetFx days, DataContractSerializer (and DataContractJsonSerializer) have both had issues deserializing a negative float/double zero. The sign gets lost. Both serializers have always done the right thing when given "-0.0" as an input, but with an input of "-0", the sign gets lost. This is both inconsistent with the other behavior and a loss of information. And unfortunately "-0" is how these serializer write the data out during serialization.

The original report provides good detail. https://github.com/dotnet/runtime/issues/69019

This incorrect behavior has been in DCS/DCJS since at least 4.8, and probably long before then.

### Version

### Previous behavior

The negative sign is lost when deserializing "-0" as a float or double.

### New behavior

The negative sign of the double is preserved.

### Type of breaking change

### Reason for change

The old behavior is both inconsistent and a technical loss of data.

### Recommended action

No action recommended. If a user was affected by this bug, then they likely will welcome the change. Or have already worked around it in a way that is unlikely to be broken by this change.

### Feature area

Serialization

### Affected APIs

DataContractSerializer.ReadObject
DataContractJsonSerializer.ReadObject

gewarren:
Fixes #30722

### DIFF
--- a/docs/core/compatibility/7.0.md
+++ b/docs/core/compatibility/7.0.md
@@ -90,6 +90,7 @@ If you're migrating an app to .NET 7, the breaking changes listed here might aff
 
 | Title | Binary compatible | Source compatible | Introduced |
 | - | :-: | :-: | - |
+| [DataContractSerializer retains sign when deserializing -0](serialization/7.0/datacontractserializer-negative-sign.md) | ❌ | ✔️ | RC 1 |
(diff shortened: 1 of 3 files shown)


IEvangelist (approved):

PR #31097, opened 2022-09-09
Added a few observed variations to the naming of the dependabot name

IEvangelist:
## Summary

Added a few observed variations to the naming of the dependabot name


BillWagner (approved):
This LGTM @IEvangelist 

Let's :shipit:

PR #31112, opened 2022-09-10
Unsafe code article: removed not used snippets

pkulikov:


BillWagner (approved):
Thanks @pkulikov 

The build is clean, so I'll :shipit:

PR #31117, opened 2022-09-11
Added the unsigned right shift to the table of overloadable operators

pkulikov:
Also added the introduction sentences and commas between operator entries.


BillWagner (approved):
Thanks again @pkulikov 

I'll :shipit: now.

PR #31141, opened 2022-09-13
Add link to fixing-warnings in trimmer warnings

jtschuster:
## Summary

Adds links to fixing-warnings.md for trimmer warnings related to DynamicallyAccessedMembersAttribute.

Adds more in depth example to DynamicallyAccessedMembers section in fixing-warnings.md

Adds IL2017 doc from the linker repository documentation

Fixes https://github.com/dotnet/docs/issues/29619

### DIFF
--- a/docs/core/deploying/trimming/fixing-warnings.md
+++ b/docs/core/deploying/trimming/fixing-warnings.md
@@ -84,24 +84,29 @@ foreach (var m in type.GetMethods())
 }
 ```
 
-In the example above, the real problem is `Console.ReadLine()`. Because *any* type could be read, the trimmer has no way to know if you need methods on `System.Tuple` or `System.Guid` or any other type. On the other hand, the following code would be fine:
+In the example above, the real problem is `Console.ReadLine()`. Because *any* type could be read, the trimmer has no way to know if you need methods on `System.String` or `System.Guid` or any other type. On the other hand, the following code would be fine:
 
 ```csharp
-Type type = typeof(System.Tuple);
+Type type = typeof(string);
 foreach (var m in type.GetMethods())
 {
     Console.WriteLine(m.Name);
 }
 ```
 
-Here the trimmer can see the exact type being referenced: `System.Tuple`. Now it can use flow analysis to determine that it needs to keep all public methods on `System.Tuple`. So where does `DynamicallyAccessMembers` come in? When reflection is split across multiple methods.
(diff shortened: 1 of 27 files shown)


docs/core/deploying/trimming/fixing-warnings.md
```diff
@@ -84,24 +84,29 @@ foreach (var m in type.GetMethods())
...
-Type type = typeof(System.Tuple);
+Type type = typeof(string);
 foreach (var m in type.GetMethods())
 {
     Console.WriteLine(m.Name);
 }
 ```
 
-Here the trimmer can see the exact type being referenced: `System.Tuple`. Now it can use flow analysis to determine that it needs to keep all public methods on `System.Tuple`. So where does `DynamicallyAccessMembers` come in? When reflection is split across multiple methods.
+Here the trimmer can see the exact type being referenced: `System.String`. Now it can use flow analysis to determine that it needs to keep all public methods on `System.String`. So where does `DynamicallyAccessMembers` come in? When reflection is split across multiple methods. In the below code, we can see that the type `string` flows to `Method3` where reflection is used to access `string`'s methods,
```
vitek-karas, line 97: Maybe I'm just paranoid - I fear that the usage of `string` could be confusing. Mainly because we allow DAM annotations on `string param` as well. Can we we choose some other type for this?
(also `string` is rather special in various ways... )

vitek-karas (approved):
The guidance on fixing warnings is looking nice... just one comment about possibly using a different type in the samples.

docs/core/deploying/trimming/fixing-warnings.md
jtschuster: That's a good point, I've updated to use System.DateTime instead.

docs/core/deploying/trimming/trim-warnings/il2092.md
```diff
@@ -27,3 +27,7 @@ public class Derived : Base
   public override void TestMethod([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)] Type type) {}
 }
 ```
+
+## Fixing
+
+See [Fixing Warnings](../fixing-warnings.md#dynamicallyaccessedmembers) for guidance.
```
tlakollo, line 33: I'm not sure how much this documentation would help with the override case, the problem here is not that we dont have annotations, but that the annotations dont match. Likely prompting to combine the annotations and have according to the example `[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods|DynamicallyAccessedMemberTypes.PublicFields)]`
So probably we need something to mention how to fix those scenarios, or remove the guidance from IL2092 to IL2095?

tlakollo (approved):

docs/core/deploying/trimming/trim-warnings/il2092.md
jtschuster: Good point, I think the fix is pretty clear from the warning, so I'll remove that section on IL2092-IL2095.

PR #31182, opened 2022-09-14
added content for missing error message cs1983

peteraritchie:
added content for missing error message cs1983


docs/csharp/language-reference/compiler-messages/cs1983.md
```diff
@@ -0,0 +1,47 @@
...
+class C
+{
+    static async IEnumerable<int> M()
+    {
+        yield return await Task.FromResult(1);
+    }
+}
+```
+
+Since  `IEnumerable.GetEnumerator` returns an `IEnumerator<T>` whose `MoveNext` method does not return a value of `Type<T>`, it is not compatible with an `async` method.
```
BillWagner, line 31: typo:

```suggestion
Since  `IEnumerable.GetEnumerator` returns an `IEnumerator<T>` whose `MoveNext` method does not return a value of `Task<T>`, it is not compatible with an `async` method.
```

BillWagner (approved):
I noticed one typo, that I'll fix, and then I'll :shipit:

PR #31201, opened 2022-09-15
added content for missing error message cs8803

peteraritchie:
added content for missing error message cs8803


BillWagner (approved):
Thanks again @peteraritchie 

All of these are great additions. I really appreciate how much you're helping us here!

This one LGTM. I'll :shipit: now.

PR #31212, opened 2022-09-15
added content for missing error message cs1997

peteraritchie:
added content for missing error message cs1997


docs/csharp/language-reference/compiler-messages/cs1997.md
```diff
@@ -0,0 +1,47 @@
...
+title: "Compiler Error CS1997"
+ms.date: 9/12/2022
+f1_keywords:
+  - "CS1997"
+helpviewer_keywords:
+  - "CS1997"
+---
+# Compiler Error CS1997
+
+Since is an async method that returns 'Task', a return keyword must not be followed by an object expression. Did you intend to return 'Task\<T\>'?
```
BillWagner, line 12: If these are backticks, the `<` and `>` don't need to be escaped. They'll also be formatted as code:

```suggestion
Since is an async method that returns `Task`, a return keyword must not be followed by an object expression. Did you intend to return `Task<T>`?
```

docs/csharp/language-reference/compiler-messages/cs1997.md
```diff
@@ -0,0 +1,47 @@
...
+        await Task.Factory.StartNew(() => 1);
+    }
+```
+
+But, the resulting implementation no longer needs the `async` modifier or the `await` operator.  A more accurate way of correcting this error is not to remove the `return` statement, but to remove the `async` modifier and the `await` operator:
+
+```csharp
+    public static Task F1()
+    {
+        Task.Factory.StartNew(() => 1);
```
BillWagner, line 45: The `return` statement was also removed:

```suggestion
        return Task.Factory.StartNew(() => 1);
```

BillWagner (approved):
This also LGTM, modulo a couple small suggestions.

PR #31253, opened 2022-09-17
added content for missing error message cs1737

peteraritchie:
added content for missing error message cs1737


BillWagner (approved):
Thanks again @peteraritchie 

This LGTM. I'll :shipit: now.

PR #31262, opened 2022-09-18
added content for missing error message cs1763

peteraritchie:
added content for missing error message cs1763


docs/csharp/language-reference/compiler-messages/cs1763.md
```diff
@@ -0,0 +1,42 @@
...
+title: "Compiler Error CS1763"
+ms.date: 9/18/2022
+f1_keywords:
+  - "CS1763"
+helpviewer_keywords:
+  - "CS1763"
+---
+# Compiler Error CS1763
+
+is of type. A default parameter value of a reference type other than string can only be initialized with null
```
BillWagner, line 12: I think we can remove the partial sentence and still have good SEO for the overall error message:

```suggestion
A default parameter value of a reference type other than string can only be initialized with null
```

BillWagner (approved):
This also LGTM, with one suggestion.

PR #31309, opened 2022-09-20
Move STJ articles into subfolder

gewarren:
Move all /serialization/system-text-json-* files to /serialization/system-text-json/* (including snippets).

No changes other than moving files, updating links, and adding redirects.


IEvangelist (commented):

IEvangelist (approved):

PR #31312, opened 2022-09-21
Fix formatting on keywords list

Anumania:
## Summary

fixes space in between "file" and "finally"


BillWagner (approved):
Thanks @Anumania 

I'll :shipit: now.

PR #31317, opened 2022-09-21
Added content for missing error message cs0273

Bielmartin:
## Summary

With these changes, I intended to add a subtopic of "To correct this error" highlighting the solution to the problem, as well as following the pattern of other pages.


BillWagner (approved):
Thanks @Bielmartin 

I'll :shipit: now.

PR #31328, opened 2022-09-21
Fix var keyword redirection link

pkulikov:
There is no "var" section, it's called differently:
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/declarations#implicitly-typed-local-variables


BillWagner (approved):
Thanks @pkulikov 

(Interesting that our link validation tool didn't trigger this.)

PR #31334, opened 2022-09-22
Fix bad code in C# Programming Guide | Methods | Return values

GitHubPang:
## Summary

1. The `ref` keyword on the right hand side was missing
2. The value should be a `double` instead of an `int` so as to be consistent with the wording above, which says:

    > ... the `Planet.GetEstimatedDistance` method returns a Double value by reference ...


BillWagner (approved):
Thanks for spotting this and fixing it @GitHubPang 

I'll :shipit: now.

PR #31368, opened 2022-09-23
Update compiler-directives.md

smoothdeveloper:
minor formatting for example of #nowarn

## Summary

the example was outside the table.


gewarren (approved):
Thanks @smoothdeveloper! I checked the preview and it looks much better.

PR #31371, opened 2022-09-23
Updated detail about indexer and usage by-reference

### LINKED ISSUE #31360: "Using Indexers - C# Programming Guide" should mention `ref returns`
Following statement isn't quite correct.

>An indexer value is not classified as a variable; therefore, you cannot pass an indexer value as a [ref](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/ref) or [out](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/out-parameter-modifier) parameter.

But you can pass an indexer value as a `ref`, if the indexer uses `ref return`, like this.

```c#
        private readonly int[] arr = new int[10];
        public ref int this[int index]
        {
            get
            {
                return ref arr[index];
            }
        }
```

---
#### Document Details

* ID: 5323b38b-8dc5-8f50-f5e7-b58ca0e88ad7
* Version Independent ID: 3d9a83cd-98e3-fc6e-727b-32c8e9228c69
* Content: [Using Indexers - C# Programming Guide](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/indexers/using-indexers)
* Content Source: [docs/csharp/programming-guide/indexers/using-indexers.md](https://github.com/dotnet/docs/blob/main/docs/csharp/programming-guide/indexers/using-indexers.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

peteraritchie:
Updated detail about an indexer unable to be passed by-reference to clarify that it can be passed as a `ref` or `out` parameter if the indexer returns `ref`.  Fixes #31360

### DIFF
--- a/docs/csharp/programming-guide/indexers/using-indexers.md
+++ b/docs/csharp/programming-guide/indexers/using-indexers.md
@@ -32,7 +32,7 @@ For more information about how to use indexers with an interface, see [Interface
 
 The signature of an indexer consists of the number and types of its formal parameters. It doesn't include the indexer type or the names of the formal parameters. If you declare more than one indexer in the same class, they must have different signatures.
 
-An indexer value is not classified as a variable; therefore, you cannot pass an indexer value as a [ref](../../language-reference/keywords/ref.md) or [out](../../language-reference/keywords/out-parameter-modifier.md) parameter.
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks again @peteraritchie 

This LGTM.

PR #31400, opened 2022-09-26
Remove note about C# 11

### LINKED ISSUE #31395: is null on unconstrained generic types is not limited to C# 11
[Enter feedback here]

The statement, "Beginning with C# 11, you can use is null on unconstrained generic types." is not correct.  It does in fact work in C# 10 (don't know about earlier versions).  For example:

bool TestNull<T>(T? test)=>test is null;

in C#10 yields false for TestNull(6) or TestNull("") but true if applied to an int? containing null or a string? containing null.

---
#### Document Details

* ID: 778dcff2-db62-27d5-c0a2-35460e0d72f8
* Version Independent ID: 14ddc828-0812-3c14-60cf-9109bf3b8f7b
* Content: [is operator - C# reference](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/is)
* Content Source: [docs/csharp/language-reference/operators/is.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/operators/is.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BillWagner:
This feature is available in earlier versions.

Fixes #31395

### DIFF
--- a/docs/csharp/language-reference/operators/is.md
+++ b/docs/csharp/language-reference/operators/is.md
@@ -31,7 +31,7 @@ The `is` operator can be useful in the following scenarios:
 
   :::code language="csharp" source="snippets/shared/IsOperator.cs" id="NullCheck":::
 
-  When you match an expression against `null`, the compiler guarantees that no user-overloaded `==` or `!=` operator is invoked. Beginning with C# 11, you can use `is null` on unconstrained generic types.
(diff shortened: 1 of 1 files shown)


tdykstra (approved):
LGTM

PR #31403, opened 2022-09-26
Fix typo in access-modifiers.md

Dan54:
Correct the number of accessibility levels from six to seven, following the addition of the `file` level.


BillWagner (approved):
Thanks @Dan54 

Good catch! I'll :shipit: now.

PR #31468, opened 2022-09-28
Update garbage-collector.md

Maoni0:
the description for the GCName config is incorrect and I got rid of the link to the design doc. it's unreasonable to ask folks to read a design doc in order to use a config.

### DIFF
--- a/docs/core/runtime-config/garbage-collector.md
+++ b/docs/core/runtime-config/garbage-collector.md
@@ -556,9 +556,8 @@ Example:
 
 ## Standalone GC
 
-- Specifies a path to the library containing the garbage collector that the runtime intends to load.
-- For more information, see [Standalone GC loader design](https://github.com/dotnet/runtime/blob/main/docs/design/features/standalone-gc-loading.md).
-
+- Specifies the name of a standalone GC dll that the runtime loads in place of the one it has in the main runtime dll (coreclr.dll). This dll needs to reside in the same directory as coreclr.dll
+- 
 | | Setting name | Values | Version introduced |
 | - | - | - | - |
 | **runtimeconfig.json** | N/A | N/A | N/A |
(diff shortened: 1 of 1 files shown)


docs/core/runtime-config/garbage-collector.md
```diff
@@ -556,9 +556,8 @@ Example:
 
 ## Standalone GC
 
-- Specifies a path to the library containing the garbage collector that the runtime intends to load.
-- For more information, see [Standalone GC loader design](https://github.com/dotnet/runtime/blob/main/docs/design/features/standalone-gc-loading.md).
-
+- Specifies the name of a standalone GC dll that the runtime loads in place of the one it has in the main runtime dll (coreclr.dll). This dll needs to reside in the same directory as coreclr.dll
```
gewarren, line 559:
```suggestion
- Specifies the name of a standalone GC assembly that the runtime loads in place of the one it has in the main runtime assembly (coreclr.dll). This assembly needs to reside in the same directory as *coreclr.dll*
```

docs/core/runtime-config/garbage-collector.md
```diff
@@ -556,9 +556,8 @@ Example:
...
+-
```
gewarren, line 560:
```suggestion

```

gewarren (approved):

kunalspathak (approved):

docs/core/runtime-config/garbage-collector.md
Maoni0: I don't think we'd want to use the word assembly here since there are native dlls?

gewarren: Okay. In that case should it be capitalized to "DLL"?

Maoni0: it can be. I'm fine either way.

gewarren: Okay, I think it's more correct to be capitalized so I updated my suggestion.

Maoni0:
thanks!

PR #31469, opened 2022-09-28
Assignment operators: grammar fixes, edits around =ref

pkulikov:
Grammar fixes and the link fix.

Most important: revision of the ref assignment section. I'm afraid that using the word "reference" when talking about `= ref` might be misleading. Actually [the spec note](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-7.3/ref-local-reassignment) doesn't use the word "reference" at all while describing the feature. It talks about the "lvalue". For example, the right-hand operand of `= ref` must be an lvalue. However, I don't know yet how to say that in the docs. In my case, the examples were essential for understanding of `=ref`. Second thing that helped understanding is the notion of "alias": one variable becoming an alias to another variable. That's why I've removed the sentences that may be confusing and put more emphasis on alias (hope  the code example helps to grasp that, too).

### DIFF
--- a/docs/csharp/language-reference/operators/assignment-operator.md
+++ b/docs/csharp/language-reference/operators/assignment-operator.md
@@ -28,19 +28,17 @@ The following example demonstrates the usage of the assignment operator with a l
 
 :::code language="csharp" source="snippets/shared/AssignmentOperator.cs" id="SnippetSimple":::
 
-The left operand of an assignment receives the *value* of the right hand of the assignment. When the operands are [value types](../builtin-types/value-types.md), the assignment copies the contents of the right-hand operand. When the operands are [reference types](../builtin-types/reference-types.md), assignment copies the reference to the object.
+The left-hand operand of an assignment receives the *value* of the right-hand operand. When the operands are of [value types](../builtin-types/value-types.md), assignment copies the contents of the right-hand operand. When the operands are of [reference types](../builtin-types/reference-types.md), assignment copies the reference to the object.
 
 This is called *value assignment*: the value is assigned.
 
 ## ref assignment
 
-*Ref assignment* assigns the *reference* of its right hand operator to its left hand operand. The left operand is an alias to the right operand. The left operand must be a [ref local](../keywords/ref.md#ref-locals), [ref readonly local](../keywords/ref.md#ref-readonly-locals) or `ref` field in a [`ref struct`]../builtin-types/ref-struct.md). The following example demonstrates the usage of the ref assignment operator:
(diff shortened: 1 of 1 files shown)


docs/csharp/language-reference/operators/assignment-operator.md
```diff
@@ -28,19 +28,17 @@ The following example demonstrates the usage of the assignment operator with a l
...
 
-*Ref assignment* assigns the *reference* of its right hand operator to its left hand operand. The left operand is an alias to the right operand. The left operand must be a [ref local](../keywords/ref.md#ref-locals), [ref readonly local](../keywords/ref.md#ref-readonly-locals) or `ref` field in a [`ref struct`]../builtin-types/ref-struct.md). The following example demonstrates the usage of the ref assignment operator:
+*Ref assignment* `= ref` makes its left-hand operand an alias to the right-hand operand. The left-hand operand must be a [ref local](../keywords/ref.md#ref-locals), [ref readonly local](../keywords/ref.md#ref-readonly-locals), or a `ref` field in a [`ref struct`](../builtin-types/ref-struct.md). Both operands must be of the same type.
 
-:::code language="csharp" source="snippets/shared/AssignmentOperator.cs" id="SnippetRefAssignment":::
-
-The `ref` must be applied to the right operands for a ref assignment. Both operands must be of the same type.
+The following example demonstrates the usage of the ref assignment operator:
 
-This is called *ref assignment*: The variable now refers to a different object.
```
pkulikov, line 43: I didn't understand that sentence, somehow it contradicts the example code.
```csharp
double[] arr = { 0.0, 0.0, 0.0 };
ref double arrayElement = ref arr[0];
```
Here, `arrayElement` becomes an alias to `arr[0]` and they both "refer" to the same (not different) memory location: the first array element.

Or I misinterpreting the sentence?

BillWagner: That was poorly written. The concept I wanted to describe is that *ref reassignment* means the variable becomes an alias to a different location. Perhaps using this example would highlight it better:

```csharp
double[] arr = { 0.0, 0.0, 0.0 };
// arrayElement is an alias for arr[0]:
ref double arrayElement = ref arr[0];
// Make arrayElement an alias to arr[1]:
arrayElement = ref arr[1];
```

What do you think?

BillWagner (commented):
Thanks for making this updates @pkulikov 

I agree that one sentence was poorly worded. I added some thoughts on how to address it. What do you think?

docs/csharp/language-reference/operators/assignment-operator.md
pkulikov: @BillWagner I've added the description of the example. Please check the latest commit.

BillWagner: That's perfect.

BillWagner (approved):
Thanks again @pkulikov 

This looks great. I'll :shipit:

PR #31471, opened 2022-09-28
Correct the title on the C# language reference card

pkulikov:
The link goes to:
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/

that is titled as "C# operators and expressions"


BillWagner (approved):
Thanks @pkulikov 

I'll :shipit: now.

PR #31476, opened 2022-09-28
Update string escape note for raw string literals

333fred:
## Summary

The for string escape sequences had not been updated for raw string literals.


BillWagner (approved):
Thanks @333fred 

I'd missed that text when I made updates.

I'll :shipit: now.

Issue #30484, opened 2022-08-03
Document new System.Text.Json features


gewarren:
- [x] [Polymorphism](https://devblogs.microsoft.com/dotnet/announcing-dotnet-7-preview-5/#polymorphism) - covered by https://github.com/dotnet/docs/issues/30879.
- [x] [Utf8Reader.CopyString](https://devblogs.microsoft.com/dotnet/announcing-dotnet-7-preview-5/#utf8jsonreader-copystring)
- [x] [Json contract customization](https://devblogs.microsoft.com/dotnet/announcing-dotnet-7-preview-6/#json-contract-customization)
- [x] Json required properties (https://github.com/dotnet/runtime/issues/29861)

See https://gist.github.com/krwq/a24138195a47fc248039314e8708dc40 for more info.

PR #31479, opened 2022-09-29
Add contract customization article

gewarren:
Contributes to #30484.

- Add new article on contract customization - [preview](https://review.learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/custom-contracts?branch=pr-en-us-31479).
- Refresh the "migrate from Newtonsoft" and "custom converters" articles to note that polymorphic deserialization and serializing private fields are now possible.

### DIFF
--- a/docs/standard/serialization/system-text-json/custom-contracts.md
+++ b/docs/standard/serialization/system-text-json/custom-contracts.md
@@ -1,0 +1,68 @@
+---
+title: Custom serialization and deserialization contracts
+description: "Learn how to write your own contract resolution logic to customize the JSON contract for a type."
+ms.date: 09/26/2022
+---
+# Customize a JSON contract
+
+The <xref:System.Text.Json?displayProperty=fullName> library constructs a JSON *contract* for each .NET type, which defines how the type should be serialized and deserialized. The contract is derived from the type's shape, which includes characteristics such as its properties and fields and whether it implements the <xref:System.Collections.IEnumerable> or <xref:System.Collections.IDictionary> interface. Types are mapped to contracts either at run time using reflection or at compile time using the source generator.
+
+Starting in .NET 7, you can customize these JSON contracts to provide more control over how types are converted into JSON and vice versa. The following list shows just some examples of the types of customizations you can make to serialization and deserialization:
+
+- Serialize private fields.
+- Support multiple names for a single property (for example, if a previous library version used a different name).
+- Construct new properties for serialization.
+- Ignore properties with a specific name, type, or value.
+- Replace `null` values with some other value.
+- Distinguish between explicit `null` values vs. lack of a value in the JSON payload.
+
+## Other ways to customize
+
+Besides customizing a contract, there are other ways to influence serialization and deserialization behavior, including the following:
+
+- The use of attributes derived from <xref:System.Text.Json.Serialization.JsonAttribute>, for example, <xref:System.Text.Json.Serialization.JsonIgnoreAttribute> and <xref:System.Text.Json.Serialization.JsonPropertyOrderAttribute>.
+- Modifying <xref:System.Text.Json.JsonSerializerOptions>, for example, to set a naming policy or serialize enumeration values as strings instead of numbers.
+- Writing a custom converter that does the actual work of writing the JSON and, during deserialization, constructing an object.
+
+Contract customization is an improvement over these pre-existing customizations because you might not have access to the type to add attributes, and writing a custom converter is complex and hurts performance.
+
+## How to opt in
+
+There are two ways to plug into customization. Both involve obtaining a resolver, whose job is to provide a <xref:System.Text.Json.Serialization.Metadata.JsonTypeInfo> instance for each type that needs to be serialized.
+
+- By calling the <xref:System.Text.Json.Serialization.Metadata.DefaultJsonTypeInfoResolver.%23ctor> constructor to obtain the <xref:System.Text.Json.JsonSerializerOptions.TypeInfoResolver?displayProperty=nameWithType> and adding your custom actions to its <xref:System.Text.Json.Serialization.Metadata.DefaultJsonTypeInfoResolver.Modifiers> property.
+
+  For example:
+
+  ```csharp
+  JsonSerializerOptions options = new()
+  {
+      TypeInfoResolver = new DefaultJsonTypeInfoResolver()
+      {
+          Modifiers =
+          {
+              MyCustomModifier1,
+              MyCustomModifier2
+          }
+      }
+  };
+  ```
+
+  If you add multiple modifiers, they'll be called serially.
+
+- By writing a custom resolver that implements <xref:System.Text.Json.Serialization.Metadata.IJsonTypeInfoResolver>.
+
+  - If a type is not handled, <xref:System.Text.Json.Serialization.Metadata.IJsonTypeInfoResolver.GetTypeInfo%2A?displayProperty=nameWithType> should return `null` for that type.
+  - You can also combine your custom resolver with others, for example, the default resolver. The resolvers will be queried in order until a non-null <xref:System.Text.Json.Serialization.Metadata.JsonTypeInfo> value is returned for the type.
+
+## Configurable aspects
+
+The <xref:System.Text.Json.Serialization.Metadata.JsonTypeInfo.Kind?displayProperty=nameWithType> property for a type indicates which aspects of its JSON contract are configurable. The `JsonTypeInfoKind.Object` kind ...
+
+
+
+## See also
+
+- [JSON contract customization blog post](https://devblogs.microsoft.com/dotnet/announcing-dotnet-7-preview-6/#json-contract-customization)
+
+- Combine reflection-based resolver with source-generated resolver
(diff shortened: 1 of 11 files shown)


docs/standard/serialization/system-text-json/custom-contracts.md
```diff
@@ -0,0 +1,120 @@
...
+# Customize a JSON contract
+
+The <xref:System.Text.Json?displayProperty=fullName> library constructs a JSON *contract* for each .NET type, which defines how the type should be serialized and deserialized. The contract is derived from the type's shape, which includes characteristics such as its properties and fields and whether it implements the <xref:System.Collections.IEnumerable> or <xref:System.Collections.IDictionary> interface. Types are mapped to contracts either at run time using reflection or at compile time using the source generator.
+
+Starting in .NET 7, you can customize these JSON contracts to provide more control over how types are converted into JSON and vice versa. The following list shows just some examples of the types of customizations you can make to serialization and deserialization:
+
+- Serialize private fields and properties.
+- Support multiple names for a single property (for example, if a previous library version used a different name).
+- Ignore properties with a specific name, type, or value.
+- Distinguish between explicit `null` values vs. lack of a value in the JSON payload.
```
eiriktsarpalis, line 15: It might be worth including an example adding support for third-party/custom attributes, as I anticipate this will be a powerful use case for the feature.

docs/standard/serialization/system-text-json/custom-contracts.md
```diff
@@ -0,0 +1,120 @@
...
+          Modifiers =
+          {
+              MyCustomModifier1,
+              MyCustomModifier2
+          }
+      }
+  };
+  ```
+
+  If you add multiple modifiers, they'll be called serially.
```
eiriktsarpalis, line 40: Probably doesn't matter much, but it's not often I've seen the term "serially" being used in this context.
```suggestion
  If you add multiple modifiers, they'll be called sequentially.
```

docs/standard/serialization/system-text-json/custom-contracts.md
eiriktsarpalis: Ah, I just saw this has been incorporated in the private fields example.

docs/standard/serialization/system-text-json/custom-contracts.md
```diff
@@ -0,0 +1,120 @@
...
+Notice in the output that the value of `RoundTrips` is incremented each time the `Product` instance is deserialized.
+
+## Example: Serialize private fields
+
+By default, `System.Text.Json` ignores private fields and properties. This example adds a new class-wide attribute, `JsonIncludePrivateFieldsAttribute`, to change that default. If the modifier finds the attribute on a type, it adds all the private fields on the type as new properties to <xref:System.Text.Json.Serialization.Metadata.JsonTypeInfo>.
+
+:::code language="csharp" source="snippets/custom-contracts/PrivateFields.cs":::
+
+> [!TIP]
+> If your private field names start with underscores, consider adding a JSON naming policy to remove the underscore and capitalize the name.
```
eiriktsarpalis, line 86: I think users might misinterpret this as saying that `JsonSerializerOptions.PropertyNamingPolicy` is being applied to custom defined `JsonPropertyInfo` instances, whereas it is only being consulted by the built-in resolvers.

I would suggest changing the term "JSON naming policy" to something different to remove that connotation.

docs/standard/serialization/system-text-json/snippets/custom-contracts/IgnoreType.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            _ignoredTypes.Add(type);
+        }
+
+        public void ModifyTypeInfo(JsonTypeInfo ti)
+        {
+            if (ti.Kind != JsonTypeInfoKind.Object)
+                return;
+
+            JsonPropertyInfo[] props = ti.Properties.Where((pi) => !_ignoredTypes.Contains(pi.PropertyType)).ToArray();
```
eiriktsarpalis, line 31: Filtering properties is awkward without an `IList<T>.RemoveAll` method, but it can be easily implemented as an extension method:

```C#
class IgnorePropertiesWithType
{
    private readonly Type[] _ignoredTypes;

    public IgnorePropertiesWithType(params Type[] ignoredTypes)
        => _ignoredTypes = ignoredTypes;

    public void ModifyTypeInfo(JsonTypeInfo ti)
    {
        if (ti.Kind != JsonTypeInfoKind.Object)
            return;

        ti.Properties.RemoveAll(prop => _ignoredTypes.Contains(prop.PropertyType));
    }
}

public static class ListHelpers
{
    // IList<T> implementation of List<T>.RemoveAll method
    public static void RemoveAll<T>(this IList<T> list, Predicate<T> predicate)
    {
        for (int i = 0; i < list.Count; i++)
        {
            if (predicate(list[i]))
            {
                list.RemoveAt(i--);
            }
        }
    }
}
```

docs/standard/serialization/system-text-json/custom-contracts.md
```diff
@@ -0,0 +1,120 @@
...
+
+The following example shows how to filter out properties with a specific type, `SecretHolder`. It does this by first clearing the <xref:System.Text.Json.Serialization.Metadata.JsonTypeInfo.Properties?displayProperty=nameWithType> list and then readding only those properties that don't have the specified type. The filtered properties will completely disappear from the contract, which means `System.Text.Json` won't look at them either during serialization or deserialization.
+
+:::code language="csharp" source="snippets/custom-contracts/IgnoreType.cs":::
+
+## Example: Allow int values to be strings
+
+Perhaps your input JSON can contain quotes around one of the numeric types but not on others. If you had control over the class, you could place <xref:System.Text.Json.Serialization.JsonNumberHandlingAttribute> on the type to fix this, but you don't. Before .NET 7, you'd need to write a [custom converter](converters-how-to.md) to fix this behavior, which is hard. Using contract customization, you can customize the number handling behavior for any type.
```
eiriktsarpalis, line 98: Custom converters aren't necessarily hard, but they certainly involve a degree of boilerplate not proportional to what needs to be achieved in this context
```suggestion
Perhaps your input JSON can contain quotes around one of the numeric types but not on others. If you had control over the class, you could place <xref:System.Text.Json.Serialization.JsonNumberHandlingAttribute> on the type to fix this, but you don't. Before .NET 7, you'd need to write a [custom converter](converters-how-to.md) to fix this behavior, which is involved. Using contract customization, you can customize the number handling behavior for any type.
```

docs/standard/serialization/system-text-json/snippets/custom-contracts/ReadIntFromString.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                jsonTypeInfo.NumberHandling = JsonNumberHandling.AllowReadingFromString;
+            }
+        }
+
+        public static void RunIt()
+        {
+            JsonSerializerOptions options = new()
+            {
+                TypeInfoResolver = new DefaultJsonTypeInfoResolver()
```
eiriktsarpalis, line 27:
```suggestion
                TypeInfoResolver = new DefaultJsonTypeInfoResolver
```

docs/standard/serialization/system-text-json/snippets/custom-contracts/IgnoreType.cs
eiriktsarpalis: API proposal created: https://github.com/dotnet/runtime/issues/76375

docs/standard/serialization/system-text-json/snippets/custom-contracts/SerializationCount.cs
```diff
@@ -0,0 +1,88 @@
...
+                        {
+                            if (value != null)
+                            {
+                                // Increment the value by 1.
+                                value = (int)value + 1;
+                            }
+
+                            set(obj, value);
+                        };
+                    }
```
IEvangelist, line 51:
```suggestion
                    Action<object, object?>? setProperty = propertyInfo.Set;
                    if (setProperty is not null)
                    {
                        propertyInfo.Set = (obj, value) =>
                        {
                            if (value != null)
                            {
                                // Increment the value by 1.
                                value = (int)value + 1;
                            }

                            setProperty (obj, value);
                        };
                    }
```

Avoid a variable name that is a contextual C# keyword.

docs/standard/serialization/system-text-json/snippets/custom-contracts/SerializationCount.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        public static void RunIt()
+        {
+            var product = new Product();
+            product.Name = "Aquafresh";
+
+            JsonSerializerOptions options = new()
+            {
+                TypeInfoResolver = new DefaultJsonTypeInfoResolver()
```
IEvangelist, line 63:
```suggestion
                TypeInfoResolver = new DefaultJsonTypeInfoResolver
```

docs/standard/serialization/system-text-json/snippets/custom-contracts/SerializationCount.cs
```diff
@@ -0,0 +1,88 @@
...
+                        };
+                    }
+                }
+            }
+        }
+
+        public static void RunIt()
+        {
+            var product = new Product();
+            product.Name = "Aquafresh";
```
IEvangelist, line 59:
```suggestion
            var product = new Product
            {
                Name = "Aquafresh"
            };
```

IEvangelist (approved):
A few nits, otherwise this looks great!

PR #31489, opened 2022-09-29
added content for missing error message cs9043

peteraritchie:
added content for missing error message cs9043


BillWagner (approved):
This also LGTM @peteraritchie

Issue #31121, opened 2022-09-12
Docs and nullability annotations disagree over the nullability of `JsonSerializerOptions.GetConverter`


bash:
[`JsonSerializerOptions.GetConverter`](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions.getconverter?view=net-6.0) returns a non-null `JsonConverter` according to the nullability attributes.

The XML docs disagree with this:
> ### Returns
> The first converter that supports the given type, or `null` if there is no converter.

The linked guide also disagrees: Both the `Read` and `Write` method in the `JsonConverterFactory` example handle `null` returns from `GetConverter`: 
```cs
// ...

private class DictionaryEnumConverterInner<TKey, TValue> :
    JsonConverter<Dictionary<TKey, TValue>> where TKey : struct, Enum
{
    private readonly JsonConverter<TValue> _valueConverter;

    public DictionaryEnumConverterInner(JsonSerializerOptions options)
    {
        // For performance, use the existing converter if available.
        _valueConverter = (JsonConverter<TValue>)options
            .GetConverter(typeof(TValue));
        // ...
    }

    public override Dictionary<TKey, TValue> Read(
        ref Utf8JsonReader reader,
        Type typeToConvert,
        JsonSerializerOptions options)
    {
        // ...

        if (_valueConverter != null)
        {
            // ...
            _valueConverter.Read(ref reader, _valueType, options)!;
        }
        else
        {
            JsonSerializer.Deserialize<TValue>(ref reader, options)!;
        }

        // ...
    }

    public override void Write(
        Utf8JsonWriter writer,
        Dictionary<TKey, TValue> dictionary,
        JsonSerializerOptions options)
    {
        // ...

        if (_valueConverter != null)
        {
            _valueConverter.Write(...);
        }
        else
        {
            JsonSerializer.Serialize(...);
        }
    }
}
```

What's the real nullability of this method?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 5de37d79-751f-e372-e866-efde28f4908c
* Version Independent ID: 678c4f23-764a-7ccd-0d6d-159a30957477
* Content: [How to write custom converters for JSON serialization - .NET](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?pivots=dotnet-6-0)
[...]

PR #31492, opened 2022-09-29
GetConverter() never returns null

gewarren:
Contributes to #31121.


eiriktsarpalis (approved):

IEvangelist (approved):

PR #31496, opened 2022-09-30
Fix wrong method name

GitHubPang:
The method name should be `GetPersonalInfo` (not `GetPersonInfo`), to be consistent with the text and code samples around it.


BillWagner (approved):
Thanks @GitHubPang 

PR #31520, opened 2022-09-30
added content for missing error message cs8131

peteraritchie:
added content for missing error message cs8131


docs/csharp/language-reference/compiler-messages/cs8131.md
```diff
@@ -0,0 +1,40 @@
...
+        var (x2, y2) = () => { };
+    }
+}
+```
+
+The compiler cannot convert an delegate (`Action`) to a two-element tuple, resulting in CS8131.
+
+## To correct this error
+
+To assign a value to a tuple, ensuring the right-hand-side expression is the same type of tuple as that on the left-hand-side:
```
BillWagner, line 33: typo:

```suggestion
To assign a value to a tuple, ensure the right-hand-side expression is the same type of tuple as that on the left-hand-side:
```

BillWagner (approved):
Thanks again @peteraritchie 

I found one typo, and then it's ready to :shipit:

PR #31530, opened 2022-09-30
Add note about JsonDocument

gewarren:
Related to #20563.

### DIFF
--- a/docs/standard/serialization/system-text-json/overview.md
+++ b/docs/standard/serialization/system-text-json/overview.md
@@ -42,7 +42,7 @@ For information about security threats that were considered when designing <xref
 
 ## Thread safety
 
-The `System.Text.Json` serializer was designed with thread safety in mind. Practically, this means that once locked, <xref:System.Text.Json.JsonSerializerOptions> instances can be safely shared across multiple threads.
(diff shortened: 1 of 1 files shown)


docs/standard/serialization/system-text-json/overview.md
```diff
@@ -42,7 +42,7 @@ For information about security threats that were considered when designing <xref
 
 ## Thread safety
 
-The `System.Text.Json` serializer was designed with thread safety in mind. Practically, this means that once locked, <xref:System.Text.Json.JsonSerializerOptions> instances can be safely shared across multiple threads.
+The `System.Text.Json` serializer was designed with thread safety in mind. Practically, this means that once locked, <xref:System.Text.Json.JsonSerializerOptions> instances can be safely shared across multiple threads. However, <xref:System.Text.Json.JsonDocument> is not thread safe as of .NET 7.
```
eiriktsarpalis, line 45: For intents and purposes the design of `JsonDocument` was to provide an immutable and thus thread-safe DOM representation for JSON values. What impacts thread safety is in fact a concurrency bug impacting specific methods of the API that we should be fixing (and potentially servicing in older releases).

I'm not sure how to best express this in documentation other noting it as a known bug.

eiriktsarpalis (approved):

PR #31535, opened 2022-10-01
Update pre-requisites for native aot publishing

jkotas:
## Summary

Trim to list pre-requisites for native aot publishing to just what is actually required.

Addressed feedback from https://github.com/dotnet/runtime/issues/69739#issuecomment-1264133462

### DIFF
--- a/docs/core/deploying/native-aot/index.md
+++ b/docs/core/deploying/native-aot/index.md
@@ -24,7 +24,7 @@ The following prerequisites need to be installed before publishing .NET projects
 
 On Windows, install [Visual Studio 2022](https://visualstudio.microsoft.com/vs/), including Desktop development with C++ workload.
 
-On Linux, install clang and developer packages for libraries that .NET runtime depends on.
+On Linux, install compiler toolchain and developer packages for libraries that .NET runtime depends on.
 
 - Ubuntu (18.04+)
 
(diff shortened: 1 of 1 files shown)


jkotas:
cc @am11

docs/core/deploying/native-aot/index.md
```diff
@@ -35,7 +35,7 @@ On Linux, install clang and developer packages for libraries that .NET runtime d
 - Alpine (3.15+)
 
     ```sh
-    sudo apk add clang gcc lld musl-dev build-base zlib-dev
+    sudo apk add build-base zlib-dev
```
am11, line 38: Line 27 notes `install clang and developer packages`, but `build-base` on Alpine (and `build-essential` on Debian/Ubuntu) is a meta package which include gcc toolchain but not llvm-toolchain. Should we update the note to `install compiler toolchain and developer packages`.

Alternatively:

```suggestion
    sudo apk add clang build-base zlib-dev
```

am11, line 38: Line 27 notes `install clang and developer packages`, but `build-base` on Alpine (and `build-essential` on Debian/Ubuntu) is a meta package which includes gcc-toolchain but not llvm-toolchain. Should we update the note to `install compiler toolchain and developer packages`.

Alternatively:

```suggestion
    sudo apk add clang build-base zlib-dev
```

jkotas: I have done both. We are doing our testing with clang, so it is what we should tell people to use. If we were to omit clang here, we would need to redo how we do testing.

am11 (approved):
LGTM, thanks!

gewarren (approved):

PR #31543, opened 2022-10-02
chore: migrate to Markdownlint-CLI2

nschonni:
## Summary

Markdownlint-CLI2 and the updated config are the libraries that now power the VSCode. The update JSONC config allows for comments, and directly referencing custom rules if the Docs Authoring ends up splitting them out


BillWagner (approved):
Thanks @nschonni 

This LGTM. 

I'll :shipit: now.

nschonni:
Sure, although, might make sense to create a `https://github.com/dotnet/workflows` and then these common workflows could be centralized, and then re-used in the various repos

PR #31544, opened 2022-10-02
ci: Allow Dependabot to update GitHub Actions

nschonni:
## Summary

Looking through the action files, some of the ones that were pinned to hashes are several major versions behind at this point. This will allow dependabot to create PRs to keep them in sync

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,12 @@
 # https://github.com/dotnet/core/tree/master/samples/dependadotnet
 version: 2
 updates:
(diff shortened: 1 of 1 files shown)


BillWagner:
Adding @terrajobst @jaredpar @richlander 

Thoughts on having dependabot update GitHub actions versions?

terrajobst:
I don't see a reason why we wouldn't allow this. Seems useful.

BillWagner (approved):
Thanks @nschonni 

I wanted to check with a couple others, but I'll :shipit: now.

PR #31578, opened 2022-10-03
ci: cleanup actions usings

nschonni:
## Summary

Dependabot has now unpinned the `usings` to the major versions. This left behind some out of date comments that were helpful for the old git-has versions.
The setup-node might not have been updated in this round, but would likely be bumped in the next dependabot run.
The last job was the remain git-has reference, so just changed it back to the tag name, since it doesn't look like it has a floating major version tag like the others


BillWagner (approved):
Thanks again @nschonni 

This all LGTM.  I'll :shipit: once the build finishes.

PR #31579, opened 2022-10-03
chore: Add Dependabot for GitHub Actions  snippets

nschonni:
## Summary

Think this should work the same as the other subfolder updates config, but I wasn't sure if the `directory` parameter was any different because of the existing special handling for GitHub Actions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,36 @@ updates:
       interval: "weekly"
       day: "wednesday"
     open-pull-requests-limit: 10
+  - package-ecosystem: "github-actions" # Create dotnet GitHub Actions snippet
+    directory: "/docs/devops/snippets/create-dotnet-github-action"
+    schedule:
+      interval: "weekly"
+      day: "wednesday"
+    open-pull-requests-limit: 10
+  - package-ecosystem: "github-actions" # Build dotnet GitHub Actions snippet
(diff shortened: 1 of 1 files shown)


nschonni:
Note that there are some hardcoded copies of these in some of the documents, so I wasn't sure if it would be better to skip this and manually update both places periodically.

BillWagner:
ping @IEvangelist for this one.

IEvangelist (approved):

nschonni:
Thanks 👍 
I'm not seeing any PRs generated like the last one once an update was checked in. Might be worth seeing if you can trigger or if there are errors on https://github.com/dotnet/docs/network/updates.

IEvangelist:
> Thanks 👍 I'm not seeing any PRs generated like the last one once an update was checked in. Might be worth seeing if you can trigger or if there are errors on https://github.com/dotnet/docs/network/updates.

Awesome, yes — there were several failing updates, I removed them here #31584. 👍🏽

nschonni:
Unfortunatly, it doesn't look like this will actually work.
image
One of the folders does have an `action.yml` so it can sync it, but doesn't work for the actual sample ones with other names that aren't in the special `.github/workflows` folder.
I think you might just want to revert this one in your other PR too.
I'll see about just manually updating those jobs to the latest for now

PR #31546, opened 2022-10-02
fix: various typos in code samples

nschonni:
## Summary

Took a pass with cSpell. Kept the MD file changes in another PR. Could split this one between the VB and CS changes if it's easier

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [samples/snippets/standard/async/async-and-await/vb/readme.md](https://github.com/dotnet/docs/blob/9c0302751ffa1fd4f34ac5edbe77014e7dfc54c4/samples/snippets/standard/async/async-and-await/vb/readme.md) | [samples/snippets/standard/async/async-and-await/vb/readme](https://review.learn.microsoft.com/en-us/dotnet/samples/snippets/standard/async/async-and-await/vb/readme?branch=pr-en-us-31546) |

PR #31581, opened 2022-10-03
Fix sample build issues

BillWagner:
A number of samples aren't building in PR #31546

- Where the sample isn't used anywhere in docs, remove it.
- Where the sample is used, upgrade it so it builds.


IEvangelist (approved):

gewarren (approved):
I only see deletions - is there anything else?

BillWagner:
> I only see deletions - is there anything else?

Note for reviewers:

I went through the samples causing errors in #31546 If a sample was never included in any article, I removed it.

The only changes were deleting those files.

adegeo (approved):

PR #31587, opened 2022-10-04
Ported as conceptual article from .NET dev blog: `Regex` enhancements with .NET 7

### LINKED ISSUE #30289: Create new article introducing `Regex` source generation enhancements
Add a new article introducing Regex source generation enhancements near:

- https://docs.microsoft.com/dotnet/standard/base-types/regular-expressions

See blog for reference:

- https://devblogs.microsoft.com/dotnet/regular-expression-improvements-in-dotnet-7

IEvangelist:
## Summary

Ported an existing .NET dev blog from @stephentoub into the .NET docs, as a new conceptual article introducing `Regex` enhancements for .NET 7:

- Rephrased wording from passive to active.
- Replaced phrases such as "I" or "we", with "you" where appropriate.
- Fixed markdown warnings and applied markdown linter fixes where possible.
- Renamed the image files and ran them through image compression and resizing.
- Added to TOC with the corresponding title.

Fixes #30289

## Preview

- [.NET regular expression source generators](https://review.learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators?branch=pr-en-us-31587)


stephentoub:
Thanks, @IEvangelist.  Though FWIW I find this out-of-place.  It's one thing to have conceptual docs about GeneratedRegex, but this post was about things that are new in .NET 7; it's not limited to just the source generator, there's a lot of implementation detail in here, and there's a lot more to regex than just what's discussed in this post, since it's written about the delta from .NET 6 to .NET 7.

IEvangelist:
> Thanks, @IEvangelist. Though FWIW I find this out-of-place. It's one thing to have conceptual docs about GeneratedRegex, but this post was about things that are new in .NET 7; it's not limited to just the source generator, there's a lot of implementation detail in here, and there's a lot more to regex than just what's discussed in this post, since it's written about the delta from .NET 6 to .NET 7.

Thanks, @stephentoub. This is still a _draft_, with lots of plans for stripping it down - I haven't gotten that far yet.  There is a ton of great info in this post, and I want the conceptual parts to remain after I shake away the implementation details. The current title was an initial thought, but this is all in flux right now, but will most likely be hyperfocused on the source generation bits. Give me a bit to work on this more, and I'll tag you for review when it's ready. Is that fair?

stephentoub:
Yup :-)

docs/standard/base-types/snippets/regular-expression-source-generators/Program.cs
```diff
@@ -0,0 +1,64 @@
...
+    private static void Main()
+    {
+        Console.OutputEncoding = Encoding.UTF8;
+
+        new List<string> { "Incubus", "Deftones", "Tool" }.ForEach(EvaluateText);
+
+        // Sample output:
+        //   ❌ "Incubus" doesn't match "abc|def" pattern.
+        //   ✅ "Deftones" matches "abc|def" pattern.
+        //   ❌ "Tool" doesn't match "abc|def" pattern.
```
CamSoper, line 39: I see what you did there 😆

CamSoper (approved):
LGTM, let's :shipit:

I love your fun string examples.

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,396 @@
...
+
+There are, however, several downsides to `RegexOptions.Compiled`. The most impactful is that it incurs much more construction cost than using the interpreter. Not only are all of the same costs paid as for the interpreter, but it then needs to compile that resulting `RegexNode` tree and generated opcodes/operands into IL, which adds non-trivial expense. The generated IL further needs to be JIT-compiled on first use leading to even more expense at startup. `RegexOptions.Compiled` represents a fundamental tradeoff between overheads on the first use and overheads on every subsequent use. The use of <xref:System.Reflection.Emit?displayProperty=nameWithType> also inhibits the use of `RegexOptions.Compiled` in certain environments; some operating systems don't permit dynamically generated code to be executed, and on such systems, `Compiled` will become a nop.
```
gewarren, line 20:
```suggestion
There are, however, several downsides to `RegexOptions.Compiled`. The most impactful is that it incurs much more construction cost than using the interpreter. Not only are all of the same costs paid as for the interpreter, but it then needs to compile that resulting `RegexNode` tree and generated opcodes/operands into IL, which adds non-trivial expense. The generated IL further needs to be JIT-compiled on first use leading to even more expense at startup. `RegexOptions.Compiled` represents a fundamental tradeoff between overheads on the first use and overheads on every subsequent use. The use of <xref:System.Reflection.Emit?displayProperty=nameWithType> also inhibits the use of `RegexOptions.Compiled` in certain environments; some operating systems don't permit dynamically generated code to be executed, and on such systems, `Compiled` will become a no-op.
```

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,396 @@
...
+
+`Regex.CompileToAssembly` itself has problems, however. First, it was never particularly user-friendly. The ergonomics of having to have a utility that would call `CompileToAssembly` to produce an assembly your app would reference resulted in relatively little use of this otherwise valuable feature. On .NET Core, `CompileToAssembly` has never been supported, as it requires the ability to save reflection-emit code to assemblies on disk, which also isn't supported. This is where source generation becomes so valuable.
```
gewarren, line 24:
```suggestion
`Regex.CompileToAssembly` itself has problems, however. First, it's not user friendly. Because a utility was required that called `CompileToAssembly` to produce an assembly your app would reference, there's relatively little use of this otherwise valuable feature. On .NET [Core], `CompileToAssembly` has never been supported, as it requires the ability to save reflection-emit code to assemblies on disk, which also isn't supported. This is where source generation becomes valuable.
```

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,396 @@
...
...roslyn-sdk/index.md) was implemented in C#. As part of this, it exposed object models for the entire compilation pipeline, with APIs the compiler itself uses to parse and understand C# but that is also exposed for arbitrary code to use to do the same. It then also enabled components that could plug into the compiler itself, with the compiler handing these "analyzers" all of the information the compiler had built up about the code being compiled and allowing the analyzers to inspect the data and issue additional "diagnostics" (or warnings). More recently, Roslyn also enabled source generators. Just like an analyzer, a source generator is a component that plugs into the compiler and is handed all of the same information as an analyzer, but in addition to being able to emit diagnostics, it can also augment the compilation unit with additional source code. The .NET 7 SDK includes a new source generator that recognizes the new <xref:System.Text.RegularExpressions.GeneratedRegexAttribute> on a partial method that returns `Regex`, and provides an implementation of that method which implements on your behalf all the logic for the `Regex`. For example, you might have written code like this:
```
gewarren, line 28: ".NET 7 addresses all of this..." - I don't think it's clear what "this" is referring to.

gewarren, line 28:
```suggestion
.NET 7 addresses all of this with the new `RegexGenerator` source generator. When the C# compiler was rewritten as the ["Roslyn" C# compiler](../../csharp/roslyn-sdk/index.md), it exposed object models for the entire compilation pipeline, as well as analyzers. More recently, Roslyn enabled source generators. Just like an analyzer, a source generator is a component that plugs into the compiler and is handed all of the same information as an analyzer, but in addition to being able to emit diagnostics, it can also augment the compilation unit with additional source code. The .NET 7 SDK includes a new source generator that recognizes the new <xref:System.Text.RegularExpressions.GeneratedRegexAttribute> on a partial method that returns `Regex`. The source generator provides an implementation of that method that implements all the logic for the `Regex`. For example, you might have written code like this:
```

gewarren: I cut out a bunch of the compiler history - I think it only adds clutter here.

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,396 @@
...
+private static void EvaluateText(string text)
+{
+    if (s_abcOrDefGeneratedRegex.IsMatch(text))
+    {
+        // Take action with matching text
+    }
+}
+```
+
+You can now write that as the following C# code:
```
gewarren, line 44:
```suggestion
You can now rewrite the previous code as follows:
```

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,396 @@
...
+private static void EvaluateText(string text)
+{
+    if (s_abcOrDefGeneratedRegex.IsMatch(text))
+    {
+        // Take action with matching text
+    }
+}
+```
+
+The generated implementation of `AbcOrDefGeneratedRegex()` similarly caches a singleton `Regex` instance, so no additional caching is needed in consuming code.
```
gewarren, line 61:
```suggestion
The generated implementation of `AbcOrDefGeneratedRegex()` similarly caches a singleton `Regex` instance, so no additional caching is needed to consume code.
```

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,396 @@
...
+
+:::image type="content" source="media/regular-expression-source-generators/cached-instance.png" alt-text="Cached regex static field":::
+
+But as can be seen, it's not just doing `new Regex(...)`. Rather, the source generator is emitting as C# code a custom `Regex`-derived implementation with logic akin to what `RegexOptions.Compiled` emits in IL. You get all the throughput performance benefits of `RegexOptions.Compiled` (more, in fact) and the start-up benefits of `Regex.CompileToAssembly`, but without the complexity of `CompileToAssembly`. The source that's emitted is part of your project, which means it's also easily viewable and debuggable.
```
gewarren, line 65:
```suggestion
But as can be seen, it's not just doing `new Regex(...)`. Rather, the source generator is emitting as C# code a custom `Regex`-derived implementation with logic similar to what `RegexOptions.Compiled` emits in IL. You get all the throughput performance benefits of `RegexOptions.Compiled` (more, in fact) and the start-up benefits of `Regex.CompileToAssembly`, but without the complexity of `CompileToAssembly`. The source that's emitted is part of your project, which means it's also easily viewable and debuggable.
```

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,396 @@
...
...ompiled`, the startup benefits of not having to do all the regex parsing, analysis, and compilation at runtime, the option of using ahead-of-time compilation with the code generated for the regex, better debugability and understanding of the regex, and even the possibility to reduce the size of your trimmed app by trimming out large swaths of code associated with `RegexCompiler` (and potentially even reflection emit itself). When used with an option like `RegexOptions.NonBacktracking` for which it can't yet generate a custom implementation, it will still helpfully emits caching, XML comments describing the implementation, and so on, such that it's still valuable. The main downside of the source generator is that it is emitting additional code into your assembly, so there's the potential for increased size; the more regexes in your app and the larger they are, the more code will be emitted for them. In some situations, just as `RegexOptions.Compiled` may be unnecessary, and so too may be the source generator, e.g. if you have a regex that's needed only rarely and for which throughput doesn't matter, it could be more beneficial to just rely on the interpreter for that sporadic usage.
```
gewarren, line 383:
```suggestion
The general guidance is if you can use the source generator, use it. If you're using `Regex` today in C# with arguments known at compile time, and especially if you're already using `RegexOptions.Compiled` (because the regex has been identified as a hot spot that would benefit from faster throughput), you should prefer to use the source generator. The source generator will give your regex the following benefits:

- All the throughput benefits of `RegexOptions.Compiled`.
- The startup benefits of not having to do all the regex parsing, analysis, and compilation at run time.
- The option of using ahead-of-time compilation with the code generated for the regex.
- Better debuggability and understanding of the regex.
- The possibility to reduce the size of your trimmed app by trimming out large swaths of code associated with `RegexCompiler` (and potentially even reflection emit itself).

When used with an option like `RegexOptions.NonBacktracking` for which the source generator can't generate a custom implementation, it will still emit caching and XML comments that describe the implementation, making it valuable. The main downside of the source generator is that it emits additional code into your assembly, so there's the potential for increased size. The more regexes in your app and the larger they are, the more code will be emitted for them. In some situations, just as `RegexOptions.Compiled` may be unnecessary, so too may be the source generator. For example, if you have a regex that's needed only rarely and for which throughput doesn't matter, it could be more beneficial to just rely on the interpreter for that sporadic usage.
```

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,396 @@
...
+
+:::image type="content" source="media/regular-expression-source-generators/xml-comments.png" lightbox="media/regular-expression-source-generators/xml-comments.png" alt-text="Generated XML comments describing regex":::
+
+## Inside the source-generated files
+
+The initial creation of the source generator was a straight port of the `RegexCompiler` used internally to implement `RegexOptions.Compiled`; line-for-line, it would essentially just emit a C# version of the IL that was being emitted. With .NET 7, both the source generator and `RegexCompiler` were almost entirely rewritten, fundamentally changing the structure of the generated code. In .NET 5, an experimental alternative approach introduced simple patterns that didn't involve any backtracking, the `RegexCompiler` could emit code that was much cleaner, the primary goal being performance. That approach has now been extended to handle all constructs (with one caveat), and both `RegexCompiler` and the source generator still map mostly 1:1 with each other, following the new approach.
```
gewarren, line 78: To be more concise, I don't think we should go into the history here. That belongs well in the blog post though.

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,396 @@
...
+
+The source generator handles everything `RegexCompiler` handles, with one exception. As with handling `RegexOptions.IgnoreCase`, the implementations now use a casing table to generate sets at construction time, and how `IgnoreCase` backreference matching needs to consult that casing table. That table is internal to `System.Text.RegularExpressions.dll`, and for now, at least, the code external to that assembly (including code emitted by the source generator) does not have access to it. That makes handling `IgnoreCase` backreferences a challenge in the source generator. It's possible that this could change to also output the casing table if it's required, but it's quite a hefty chunk of data to blit[^2] into consuming assemblies. So at least for now, `IgnoreCase` backreferences are the one construct not supported by the source generator that is supported by `RegexCompiler`. If you try to use a pattern that has one of these (which is rare), the source generator won't emit a custom implementation and will instead fall back to caching a regular `Regex` instance:
```
gewarren, line 373:
```suggestion
The source generator handles everything `RegexCompiler` handles, with one exception. As with handling `RegexOptions.IgnoreCase`, the implementations now use a casing table to generate sets at construction time, and how `IgnoreCase` backreference matching needs to consult that casing table. That table is internal to `System.Text.RegularExpressions.dll`, and for now, at least, the code external to that assembly (including code emitted by the source generator) does not have access to it. That makes handling `IgnoreCase` backreferences a challenge in the source generator and they aren't supported. This is the one construct not supported by the source generator that is supported by `RegexCompiler`. If you try to use a pattern that has one of these (which is rare), the source generator won't emit a custom implementation and will instead fall back to caching a regular `Regex` instance:
```

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,396 @@
...
+
+[^2]: https://wikipedia.org/wiki/Blit "Wikipedia: Blit — A blit is a block transfer operation that copies data from one memory location to another."
+
+:::image type="content" source="media/regular-expression-source-generators/cached-instance-fallback.png" lightbox="media/regular-expression-source-generators/cached-instance-fallback.png" alt-text="Unsupported regex still being cached":::
+
+Also, neither `RegexCompiler` nor the source generator supports the new `RegexOptions.NonBacktracking`. If you specify `RegexOptions.Compiled | RegexOptions.NonBacktracking`, the `Compiled` flag will just be ignored, and if you specify `NonBacktracking` to the source generator, it will similarly fall back to caching a regular `Regex` instance.
+
+## Summary
```
gewarren, line 381:
```suggestion
## When to use it
```

docs/standard/base-types/regular-expression-source-generators.md
IEvangelist: Fixed

docs/standard/base-types/regular-expression-source-generators.md
IEvangelist: Fixed

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,400 @@
...
+
+When you specify <xref:System.Text.RegularExpressions.RegexOptions.Compiled?displayProperty=nameWithType>, before .NET 7, all of the same construction-time work would be performed. The resulting instructions would be transformed further by the reflection-emit-based compiler into IL instructions that would be written to a few <xref:System.Reflection.Emit.DynamicMethod>s. When a match was performed, those `DynamicMethod`s would be invoked. This IL would essentially do exactly what the interpreter would do, except specialized for the exact pattern being processed. For example, if the pattern contained `[ac]`, the interpreter would see an opcode that said "match the input character at the current position against the set specified in this set description" whereas the compiled IL would contain code that effectively said, "match the input character at the current position against `'a'` or `'c'`". This special casing and the ability to perform optimizations based on knowledge of the pattern are some of the main reasons for specifying `RegexOptions.Compiled` yields much faster-matching throughput than does the interpreter.
```
stephentoub, line 18: "before .NET 7"... this is still the case in .NET 7 and will be in the future.

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,400 @@
...
+
+To help with these issues, .NET provides a method <xref:System.Text.RegularExpressions.Regex.CompileToAssembly%2A?displayProperty=nameWithType>. This method enables the same IL that would have been generated for `RegexOptions.Compiled` to instead be written to a generated assembly on disk, and that assembly can then be referenced as a library from your app. This has the benefit of avoiding the startup overheads involved in parsing, optimizing, and outputting the IL for the expression, as that can all be done ahead of time rather than each time the app is invoked. Further, that assembly could be ahead-of-time compiled with technology like ngen or crossgen, avoiding most of the associated JIT costs as well.
```
stephentoub, line 22: ".NET provides"... CompileToAssembly isn't supported in .NET Core. Consider saying ".NET Framework" here to make it explicit.  Two paragraphs for an unsupported feature is also a lot for the public docs; consider consolidating this.

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,400 @@
...
+}
+```
+
+You can now rewrite the previous code as follows:
+
+```csharp
+private static readonly Regex s_abcOrDefGeneratedRegex = AbcOrDefGeneratedRegex();
+
+[GeneratedRegex("abc|def", RegexOptions.IgnoreCase | RegexOptions.Compiled, "en-US")]
+private static partial Regex AbcOrDefGeneratedRegex();
```
stephentoub, line 50: There's no benefit to caching this into a static field.  The source generator does that itself.  If you look at the generated code for this, you'll see the generated AbcOrDefGeneratedRegex method is essentially just `AbcOrDefGeneratedRegex() => s_cachedField;`.

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,400 @@
...
+private static void EvaluateText(string text)
+{
+    if (s_abcOrDefGeneratedRegex.IsMatch(text))
+    {
+        // Take action with matching text
+    }
+}
+```
+
+The generated implementation of `AbcOrDefGeneratedRegex()` similarly caches a singleton `Regex` instance, so no additional caching is needed to consume code.
```
stephentoub, line 61: Then why does your code example show caching it manually? :)

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,400 @@
...
+
+But as can be seen, it's not just doing `new Regex(...)`. Rather, the source generator is emitting as C# code a custom `Regex`-derived implementation with logic similar to what `RegexOptions.Compiled` emits in IL. You get all the throughput performance benefits of `RegexOptions.Compiled` (more, in fact) and the start-up benefits of `Regex.CompileToAssembly`, but without the complexity of `CompileToAssembly`. The source that's emitted is part of your project, which means it's also easily viewable and debuggable.
+
+:::image type="content" source="media/regular-expression-source-generators/debuggable-source.png" lightbox="media/regular-expression-source-generators/debuggable-source.png" alt-text="Debugging through source-generated Regex code":::
+
+> [!TIP]
+> In Visual Studio, select the project node in **Solution Explorer**, then expand **Dependencies** > **Analyzers** > **System.Text.RegularExpressions.Generator** > **System.Text.RegularExpressions.Generator.RegexGenerator** > _RegexGenerator.g.cs_ to see the generated C# code from this regex generator.
```
stephentoub, line 70: You can also just right-click on your partial method declaration and select "Go To Definition".

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,400 @@
...
+
+## Compiled regular expressions
+
+When you write `new Regex("somepattern")`, a few things happen. The specified pattern is parsed, both to ensure the validity of the pattern and to transform it into an internal `RegexNode` tree that represents the parsed regex. The tree is then optimized in various ways, transforming the pattern into a functionally equivalent variation that can be more efficiently executed. The tree is written into a form that can be interpreted as a series of opcodes and operands that provide instructions to the `RegexInterpreter` engine on how to match. When a match is performed, the interpreter simply walks through those instructions, processing them against the input text. When instantiating a new `Regex` instance or calling one of the static methods on `Regex`, the interpreter is the default engine employed.
```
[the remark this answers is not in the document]
IEvangelist: Good callout, I'll rephrase this to be conceptual.

docs/standard/base-types/regular-expression-source-generators.md
IEvangelist: Removed .NET 7 callout here.

docs/standard/base-types/regular-expression-source-generators.md
IEvangelist: I thought the purpose of this section was to show exactly what the analyzer with the applied fix will do, I can adjust this. The analyzer's applied fix though doesn't behave this way, instead, it will re-write your code like the resulting code above.

docs/standard/base-types/regular-expression-source-generators.md
IEvangelist: Fixed.

docs/standard/base-types/regular-expression-source-generators.md
IEvangelist: Adding this, thanks.

docs/standard/base-types/regular-expression-source-generators.md
IEvangelist: Removed entirely, as the content stands without it. Instead, added the link to the blog post in the see also section.

IEvangelist:
Thanks for the feedback @stephentoub, I've addressed it in https://github.com/dotnet/docs/pull/31758.

docs/standard/base-types/regular-expression-source-generators.md
```diff
@@ -0,0 +1,400 @@
...
+:::image type="content" source="media/regular-expression-source-generators/debuggable-source.png" lightbox="media/regular-expression-source-generators/debuggable-source.png" alt-text="Debugging through source-generated Regex code":::
+
+> [!TIP]
+> In Visual Studio, select the project node in **Solution Explorer**, then expand **Dependencies** > **Analyzers** > **System.Text.RegularExpressions.Generator** > **System.Text.RegularExpressions.Generator.RegexGenerator** > _RegexGenerator.g.cs_ to see the generated C# code from this regex generator.
+
+You can set breakpoints in it, you can step through it, and you can use it as a learning tool to understand exactly how the regex engine is processing your pattern with your input. The generator even generates [triple-slash (XML) comments](../../csharp/language-reference/xmldoc/index.md) to help make the expression understandable at a glance and where it's used.
+
+:::image type="content" source="media/regular-expression-source-generators/xml-comments.png" lightbox="media/regular-expression-source-generators/xml-comments.png" alt-text="Generated XML comments describing regex":::
+
+## Inside the source-generated files
```
stephentoub, line 76: Much of this section is about implementation detail.  Is it all relevant to the docs?

docs/standard/base-types/regular-expression-source-generators.md
stephentoub: The analyzer isn't mentioned in this section.  The description of this code sample states: "You can now rewrite the previous code as follows:"

PR #31598, opened 2022-10-04
Fix typo.

teo-tsirpanis:


BillWagner (approved):
Thanks for spotting this and fixing it @teo-tsirpanis 

I'll :shipit: now.

PR #31603, opened 2022-10-05
Fix wording in LINQ docs

GitHubPang:
## Summary

Remove the words "first two".

Note that the code samples take all items (not just the first two items) from the lists.


BillWagner (approved):
Thanks again for fixing this @GitHubPang 

This LGTM. I'll :shipit: now.

PR #31608, opened 2022-10-05
Link the specific using article when possible

pkulikov:
I'm thinking about removing the [umbrella `using` article](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using), so updating some links to it before doing that.

### DIFF
--- a/docs/csharp/misc/cs0104.md
+++ b/docs/csharp/misc/cs0104.md
@@ -12,7 +12,7 @@ ms.assetid: 1a7e9ae8-308b-441b-ba85-fac974222875
 
 'reference' is an ambiguous reference between 'identifier' and 'identifier'  
   
- Your program contains [using](../language-reference/keywords/using.md) directives for two namespaces and your code references a name that appears in both namespaces.  
+ Your program contains [using](../language-reference/keywords/using-directive.md) directives for two namespaces and your code references a name that appears in both namespaces.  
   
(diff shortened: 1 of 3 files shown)


BillWagner:
Hi @pkulikov 

I like these changes, and I'll :shipit: shortly.

We shouldn't remove the umbrella `using` article though. The `F1` service in Visual Studio sends the single token under the cursor to the docs service. That means for `using`, the service can't disambiguate between a `using` directive, a `static using` directive, or a `using` statement. That's why we have these umbrella pages. (Similar pages exist for `in`, `out`: covariance/contravariance, parameter modifiers, `protected`, `private`, `internal`: compound access modifiers).

BillWagner (approved):
Thanks @pkulikov 

This LGTM. I'll :shipit: now.

pkulikov:
> We shouldn't remove the umbrella `using` article though.

OK, I see, thank you.

The question about TOC, then.
This is the current TOC:
image

How to change TOC when the `using` statement article is moved into the "Statements" folder? Then, the "using statement" TOC entry should move under the "Statements" TOC entry.

If that umbrella page is mostly the F1 end point, then that page can be removed from TOC. As it is now for [the `out` umbrella page](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/out), which is not in TOC.

PR #31609, opened 2022-10-05
ref struct article: minor edits

pkulikov:


BillWagner (approved):
Thanks again @pkulikov 

These changes look great, and I'll :shipit: now.

PR #31610, opened 2022-10-05
Update backward-compatibility-guidelines.md

zeynelok:
The positions of the parameters have been changed

### DIFF
--- a/docs/orleans/grains/grain-versioning/backward-compatibility-guidelines.md
+++ b/docs/orleans/grains/grain-versioning/backward-compatibility-guidelines.md
@@ -73,7 +73,7 @@ public interface IMyGrain : IGrainWithIntegerKey
 public interface IMyGrain : IGrainWithIntegerKey
 {
     // return a - b
-    Task<int> Subtract(int x, int y);
+    Task<int> Subtract(int b, int a);
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/orleans/grains/grain-versioning/backward-compatibility-guidelines.md
```diff
@@ -73,7 +73,7 @@ public interface IMyGrain : IGrainWithIntegerKey
 public interface IMyGrain : IGrainWithIntegerKey
 {
     // return y - x
```
IEvangelist, line 75:
```suggestion
    // return x - y
```

docs/orleans/grains/grain-versioning/backward-compatibility-guidelines.md
```diff
@@ -73,7 +73,7 @@ public interface IMyGrain : IGrainWithIntegerKey
...
-    Task<int> Subtract(int y, int x);
+    Task<int> Subtract(int x, int y);
```
IEvangelist, line 76:
```suggestion
    Task<int> Subtract(int y, int x);
```

IEvangelist (commented):
Let's use the same parameter names, but swap their order to more easily follow the concern here.

IEvangelist (approved):

PR #31584, opened 2022-10-04
Update dependabot.yml

IEvangelist:
These directories don't exist, and thus it fails with an error similar to the following:

> ⚠️ Dependabot couldn't find a <anything>.(cs|vb|fs)proj
Dependabot couldn't find a .(cs|vb|fs)proj.
> 
> Dependabot requires a .(cs|vb|fs)proj to evaluate your .NET dependencies. It had expected to find one at the path: /samples/snippets/csharp/pipelines/<anything>.(cs|vb|fs)proj.
> 
> If this isn't a .NET project, you may wish to disable updates for it in the .github/dependabot.yml config file in this repo.

PR #31622, opened 2022-10-06
Regenerate our dependabot config file

IEvangelist:
## Summary

Regenerate our @dependabot config file. Note, the generator is only aware of the `package-ecosystem: "nuget"`, we can update that later if we determine it's necessary to add `package-ecosystem: "github-actions"`.

Related to https://github.com/dotnet/docs/pull/31584#issuecomment-1267096686 and part of https://github.com/dotnet/docs-tools/pull/55


.github/dependabot.yml
```diff
@@ -1,435 +1,807 @@
 # generated by dependadotnet
-# https://github.com/dotnet/core/tree/master/samples/dependadotnet
+# https://github.com/dotnet/core/tree/main/samples/dependadotnet
 version: 2
 updates:
-  - package-ecosystem: "github-actions" # Core GitHub Actions
```
nschonni, line 5: May want to manually add this one back, since the generator doesn't know about Actions yet

BillWagner (approved):
This LGTM.

I agree with @nschonni 's comment here.

IEvangelist:
> This LGTM.
> 
> I agree with @nschonni 's comment here.

So, I should generate this and then manually add some back? 😕 Might be worthwhile to just add that functionality automatically. I mean each repo that does support GitHub Actions by defining workflow files has them in a standard location enforced by GitHub. We could just check to see if the repo that the executable is running on has a _.github/workflows_ directory and then apply `github-action` entries for those, what do you think?

PR #31626, opened 2022-10-06
Update .repoman.yml

IEvangelist:
Adding appropriate labels to the networking files and directories. @gewarren is this where we've been managing automated label assignments? I can't recall, and I know that fabricbot was doing some as well.


adegeo (approved):

adegeo (approved):
Just one minor fix above to commit

PR #21882, opened 2020-12-08
Deprecate CA1801 in favor of IDE0060

Evangelink:
## Summary

Deprecate CA1801 in favor of IDE0060.

Relates to https://github.com/dotnet/roslyn-analyzers/issues/4498

PR #31659, opened 2022-10-08
 Remove rule CA2109

x789:
Removes documentation of rule CA2109.
Rule is removed by PR dotnet/roslyn-analyzers#6198.
Removal of the rule was discussed in dotnet/roslyn-analyzers#5974.


gewarren:
We'll at least need to add a redirection to https://github.com/dotnet/docs/blob/main/.openpublishing.redirection.fundamentals.json. But I'm wondering if we should also add a new page/section for "deleted" or "obsolete" rules, just so some reference remains. Thoughts?

Youssef1313:
When CA1801 got deprecated, a note was just added. See https://github.com/dotnet/docs/pull/21882

Regardless of the decision, we should be consistent.

gewarren:
> When CA1801 got deprecated, a note was just added. See #21882
> 
> Regardless of the decision, we should be consistent.

Yes, I like that better. Leave the article in place but add a prominent banner at the top like #21882.

x789:
I like the idea that information about the rule is not deleted completely.
From a developer's point of view, the reason why a rule was removed is definitely helpful.

I will adjust the PR so that the information for CA2109 is not removed, but a banner is added (as suggested by @Youssef1313).

Due to my current workload, this may take some days.

x789:
I started over with a new branch to get a clean commit history. See PR #31797.

PR #31660, opened 2022-10-08
Move `file` keyword into contextual keyword list

kyoyama-kazusa:
C# 11 declares a new feature using a new keyword `file` to restrict the accessibility. It should be a contextual keyword, not a predefined keyword.

Page: [C# Keywords](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/)

Content source: [docs/csharp/language-reference/keywords/index.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/keywords/index.md)

---

This is my first PR. If there is something wrong with my operation or modification, please tell me :D

### DIFF
--- a/docs/csharp/language-reference/keywords/index.md
+++ b/docs/csharp/language-reference/keywords/index.md
@@ -45,7 +45,6 @@ The first table in this article lists keywords that are reserved identifiers in 
         [`explicit`](../operators/user-defined-conversion-operators.md)  
         [`extern`](extern.md)  
         [`false`](../builtin-types/bool.md)  
-        [`file`](file.md)  
         [`finally`](try-finally.md)  
         [`fixed`](../statements/fixed.md)  
         [`float`](../builtin-types/floating-point-numeric-types.md)  
@@ -123,6 +122,7 @@ A contextual keyword is used to provide a specific meaning in the code, but it i
         [`descending`](descending.md)  
         [`dynamic`](../builtin-types/reference-types.md)  
         [`equals`](equals.md)  
+        [`file`](file.md)  
         [`from`](from-clause.md)  
     :::column-end:::
     :::column:::


IEvangelist:
> C# 11 declares a new feature using a new keyword `file` to restrict the accessibility. It should be a contextual keyword, not a predefined keyword.
> 
> Page: [C# Keywords](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/)
> 
> Content source: [docs/csharp/language-reference/keywords/index.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/keywords/index.md)
> 
> This is my first PR. If there is something wrong with my operation or modification, please tell me :D

I think this is an access modifier, like `private` or `public`, and it scopes the access to the type within the file itself. Therefore, I believe this belongs next to those other access modifiers. @BillWagner would know for sure.

BillWagner:
The speclet doesn't explicitly say if `file` is a keyword or a contextual keyword. 

I *think* it's a contextual keyword.  Tagging @RikkiGibson to make sure.

RikkiGibson:
`file` is a contextual keyword, but types named `file` are blocked from being declared in the language.

We think it's important for people to still be able to declare a variable like `Stream file = ...;`

SunnieShine:
hi everybody :)

I found this:

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/file

This passage has mentioned that `file` is a contextual keyword.

BillWagner (approved):
Thanks for making this fix, and your patience while I confirmed that it is correct.

I'll :shipit: now.

Youssef1313:
> Beginning with C# 11, the file contextual keyword is a type access modifier.

I believe Roslyn team was always avoiding to say that `file` is an access modifier.

@RikkiGibson Could you confirm please? Thanks!

RikkiGibson:
Yeah, we use the term *file-local* instead of referring to *file accessibility*. We also avoid saying *file scope* because we don't want the feature to be compared to *file-scoped namespaces*.

I would like for us to revisit our language around accessibility, it seems like what we've come up with really doesn't match how users think of it, but that decision should be made by the LDM and for now the docs should probably reflect the existing design.

PR #31668, opened 2022-10-09
chore: Breaking Change schema errors

nschonni:
## Summary

Opening the file in VSCode flagged a few issues for elements that don't match the current schema.
- `labels` should be an array rather than a comma string
- `checkboxes` only allow `required` on the individual options https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#checkboxes

/cc @gewarren


IEvangelist (approved):

PR #31673, opened 2022-10-10
chore: remove unessary markdown-disable comment

nschonni:
Already ignores instance in code blocks/links

## Summary

Some of these rules are now disabled at the root or have no errors when removed anymore


IEvangelist (approved):

PR #31676, opened 2022-10-10
Fix docs wording: 'maybe' to 'may be'

j-d-b:
When reading this documentation page, I noticed "maybe" was used in a context in place of "may be", which would be more correct in the sentence it is used in.


IEvangelist (approved):

PR #31680, opened 2022-10-10
Publish speclet for low-level struct improvements

BillWagner:
Add the spec for C# 11's low level struct improvements.


BillWagner:
Merging this should fix our current build warnings. /cc @dotnet/docs

IEvangelist (approved):

PR #31687, opened 2022-10-10
docs: fix grammar in numbers-in-csharp

eduardopelitti:
## Summary

The sentence "Double precision number have twice the number of binary digits as single-precision" seems to be missing an "s" in the word "numbers". 

It could also be rephrased as "A double precision number has twice the number of binary digits as single-precision".

I also noticed that "double-precision" and "double precision" are used interchangeably within the same paragraph, the same for "single-precision" and "single precision". Perhaps one should be enforced, but no change is introduced regarding this.

Thank you for a great tutorial :)


BillWagner (approved):
Thanks @eduardopelitti 

This really looks great.

I'll :shipit: now.

BillWagner:

PR #31693, opened 2022-10-11
Revise `file` documentation

Youssef1313:


docs/csharp/language-reference/keywords/file.md
```diff
@@ -21,9 +20,9 @@ file class HiddenWidget
 }
 ```
 
-Any types nested within a file-based type are also only visible within the file in which it's declared. Other types in an assembly may use the same name as a file-scoped type. Because the file-scoped type is visible only in the file where it's declared, these types don't create a naming collision.
+Any types nested within a file-scoped type are also only visible within the file in which it's declared. Other types in an assembly may use the same name as a file-scoped type. Because the file-scoped type is visible only in the file where it's declared, these types don't create a naming collision.
```
RikkiGibson, line 23:
```suggestion
Any types nested within a file-local type are also only visible within the file in which it's declared. Other types in an assembly may use the same name as a file-local type. Because the file-local type is visible only in the file where it's declared, these types don't create a naming collision.
```

docs/csharp/language-reference/keywords/file.md
```diff
@@ -1,18 +1,17 @@
...
 helpviewer_keywords: 
   - "file keyword [C#]"
 ---
 # file (C# Reference)
 
-Beginning with C# 11, the `file` contextual keyword is a type access modifier.
+Beginning with C# 11, the `file` contextual keyword is a type modifier.
 
-File-scoped access restricts a top-level type's scope and visibility to the file in which it's declared. The `file` accessor modifier will generally be applied to types written by a source generator. File scoped types provide source generators with a convenient way to avoid name collisions among generated types. The `file` access modifier declares a file-scoped type, as in this example:
+The `file` modifier restricts a top-level type's scope and visibility to the file in which it's declared. The `file` modifier will generally be applied to types written by a source generator. File-scoped types provide source generators with a convenient way to avoid name collisions among generated types. The `file` modifier declares a file-scoped type, as in this example:
```
RikkiGibson, line 14:
```suggestion
The `file` modifier restricts a top-level type's scope and visibility to the file in which it's declared. The `file` modifier will generally be applied to types written by a source generator. File-local types provide source generators with a convenient way to avoid name collisions among generated types. The `file` modifier declares a file-local type, as in this example:
```

docs/csharp/language-reference/keywords/file.md
```diff
@@ -21,9 +20,9 @@ file class HiddenWidget
...
+Any types nested within a file-scoped type are also only visible within the file in which it's declared. Other types in an assembly may use the same name as a file-scoped type. Because the file-scoped type is visible only in the file where it's declared, these types don't create a naming collision.
 
-A file-scoped type can't be the return type or parameter type of any member that is more visible than `file` scope. A file-scoped type can't be a field member of a type that has greater visibility than `file` scope. However, a more visible type may implicitly implement a file-scoped interface type. The type can't [explicitly implement](../../programming-guide/interfaces/explicit-interface-implementation.md) a file scoped interface.
+A file-scoped type can't be the return type or parameter type of any member that is more visible than `file` scope. A file-scoped type can't be a field member of a type that has greater visibility than `file` scope. However, a more visible type may implicitly implement a file-scoped interface type. The type can't [explicitly implement](../../programming-guide/interfaces/explicit-interface-implementation.md) a file-scoped interface.
```
RikkiGibson, line 25:
```suggestion
A file-local type can't be the return type or parameter type of any member that is more visible than `file` scope. A file-local type can't be a field member of a type that has greater visibility than `file` scope. However, a more visible type may implicitly implement a file-scoped interface type. The type can't [explicitly implement](../../programming-guide/interfaces/explicit-interface-implementation.md) a file-local interface.
```

RikkiGibson (commented):
Suggest changing file-scoped to file-local throughout, which we did in the proposal to avoid comparison with file-scoped namespace.

RikkiGibson (approved):

RikkiGibson:
@BillWagner to make sure you're OK with the terms as they stand. I do want us to revisit the term *file-local* versus calling it an access modifier. I'll open an internal thread about it when I get the chance.

BillWagner (approved):
Thanks @Youssef1313 

This LGTM. I'll :shipit: now.

PR #31694, opened 2022-10-11
Do not use smart quotes in code examples

eNeRGy164:
## Summary

Small changes to code samples that use smart quotes instead of regular quotes.
This fixes some samples and should allow for a better copy & paste experience.


gewarren (approved):
Thank you @eNeRGy164

PR #31695, opened 2022-10-11
Fix cancellation by polling example (#31643)

### LINKED ISSUE #31643: Mistake in example for listening for cancellation requests by polling
What happens if the token is cancelled after the if (token.IsCancellationRequested) {}, but before the evaluation of the outer for loop? The outer for loop's conditions won't be met and thus won't execute again. The function will return without ever executing the cleanup code inside the inner if statement (in the example it's only Console.WriteLine stuff, but in real life it might be important stuff like freeing up vast amounts of resources)?

---
#### Document Details

* ID: 51f5ff8f-7d37-af59-4b51-41de21d67b95
* Version Independent ID: ef30d618-f141-3943-b93b-ce697ca8991b
* Content: [Cancellation in Managed Threads](https://learn.microsoft.com/en-us/dotnet/standard/threading/cancellation-in-managed-threads?redirectedfrom=MSDN)
* Content Source: [docs/standard/threading/cancellation-in-managed-threads.md](https://github.com/dotnet/docs/blob/main/docs/standard/threading/cancellation-in-managed-threads.md)
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

JJS:
## Summary

I moved the `token.IsCancellationRequested`-block out of the outer loop to ensure it is checked in all cases.

Fixes #31643

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_Misc/cancellation/cs/cancellation.csproj
+++ b/samples/snippets/csharp/VS_Snippets_Misc/cancellation/cs/cancellation.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
   </PropertyGroup>
 
 </Project>
--- a/samples/snippets/csharp/VS_Snippets_Misc/cancellation/cs/cancellation.csproj
+++ b/samples/snippets/csharp/VS_Snippets_Misc/cancellation/cs/cancellation.csproj
@@ -1,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+</Project>
--- a/samples/snippets/csharp/VS_Snippets_Misc/cancellation/cs/cancellationex11.cs
+++ b/samples/snippets/csharp/VS_Snippets_Misc/cancellation/cs/cancellationex11.cs
@@ -41,23 +41,22 @@ class CancelByPolling
    static void NestedLoops(Rectangle rect, CancellationToken token)
    {
       for (int x = 0; x < rect.columns && !token.IsCancellationRequested; x++) {
+         // Assume that we know that the inner loop is very fast.
+         // Therefore, checking once per row in the outer loop is sufficient.
          for (int y = 0; y < rect.rows; y++) {
             // Simulating work.
             Thread.SpinWait(5000);
             Console.Write("{0},{1} ", x, y);
          }
+      }
 
-         // Assume that we know that the inner loop is very fast.
-         // Therefore, checking once per row is sufficient.
-         if (token.IsCancellationRequested) {
-            // Cleanup or undo here if necessary...
-            Console.WriteLine("\r\nCancelling after row {0}.", x);
-            Console.WriteLine("Press any key to exit.");
-            // then...
-            break;
-            // ...or, if using Task:
-            // token.ThrowIfCancellationRequested();
-         }
+      if (token.IsCancellationRequested) {
+         // Cleanup or undo here if necessary...
+         Console.WriteLine("\r\nCancelling after row {0}.", x);
+         Console.WriteLine("Press any key to exit.");
+
+         // If using Task:
+         // token.ThrowIfCancellationRequested();
       }
    }
    //</snippet3>
--- a/samples/snippets/csharp/VS_Snippets_Misc/cancellation/cs/cancellationex11.cs
+++ b/samples/snippets/csharp/VS_Snippets_Misc/cancellation/cs/cancellationex11.cs
@@ -40,14 +40,13 @@ class CancelByPolling
    //<snippet3>
    static void NestedLoops(Rectangle rect, CancellationToken token)
    {
-      int col = 0;
-      for (; col < rect.columns && !token.IsCancellationRequested; col++) {
+      for (int col = 0; col < rect.columns && !token.IsCancellationRequested; col++) {
          // Assume that we know that the inner loop is very fast.
          // Therefore, polling once per column in the outer loop condition
          // is sufficient.
          for (int row = 0; row < rect.rows; row++) {
             // Simulating work.
-            Thread.SpinWait(5000);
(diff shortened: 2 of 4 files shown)


BillWagner (approved):
Thanks for updating this @JJS 

JJS:
Is there something to do about the failed snippet? The error seems not to be related to the issue

BillWagner:
> Is there something to do about the failed snippet? The error seems not to be related to the issue

Yes, if you create a default .csproj and .vbproj in the two folders with these source files, that will fix the issue.

It's a historical artifact. The source code was imported into GitHub from an internal system that built the samples when edited. There weren't project files, and we've been adding them whenever we edit any of the existing samples.

samples/snippets/csharp/VS_Snippets_Misc/cancellation/cs/cancellationex11.cs
```diff
@@ -41,23 +41,22 @@ static void Main()
...
-            Console.WriteLine("\r\nCancelling after row {0}.", x);
-            Console.WriteLine("Press any key to exit.");
-            // then...
-            break;
-            // ...or, if using Task:
-            // token.ThrowIfCancellationRequested();
-         }
+      if (token.IsCancellationRequested) {
+         // Cleanup or undo here if necessary...
+         Console.WriteLine("\r\nCancelling after row {0}.", x);
```
Youssef1313, line 55: Is `x` in scope here?

Youssef1313: I don't see how cancellation is useful after the heavy work is already done?

JJS: true, `x` is not in scope, I'll fix that.

As the usefulness of the cancellation is concerned, the overall behavior of the code did not change.

The polling example's main part is the `IsCancellationRequested` check in the loop condition, which did not change.

I suppose you could remove the additional `if` condition, which is only there to provide a place for clean up code in case of cancellation.

I am not sure what you mean by "heavy work is already done" as the outer loop can be left at any point when cancellation is requested.

Youssef1313: I could be misunderstanding something (and in fact I'm not sure what's the issue being resolved here), but I see `token.IsCancellationRequested` is done after the loops have already finished and completed the work.

This also doesn't make sense for me with the existing comment that says:

>         // Assume that we know that the inner loop is very fast.
>         // Therefore, checking once per row in the outer loop is sufficient.

What do you think @BillWagner?

JJS: I updated the scope, the comments and some other issues with mixed up columns/rows.

`token.IsCancellationRequested` is called _in the loop condition_ (so polling each column), not only after the loops. So the loops may not have completed their work. As I understand the original example, the call to `token.IsCancellationRequested` after the loops is only for clean up or calling `token.ThrowIfCancellationRequested`.

As the linked issue correctly states, there is an unexpected behavior if the cancellation is requested _after_ the `if` condition is evaluated and _before_ the loop condition is evaluated. In this case, the clean up code is not called.

Youssef1313: > token.IsCancellationRequested is called in the loop condition

Ah I missed that.

samples/snippets/csharp/VS_Snippets_Misc/cancellation/cs/cancellation.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
IEvangelist, line 5:
```suggestion
    <TargetFramework>net6.0</TargetFramework>
```

Let's at least bump to the latest LTS

samples/snippets/csharp/VS_Snippets_Misc/cancellation/cs/cancellationex11.cs
```diff
@@ -40,24 +40,25 @@ static void Main()
...
-      for (int x = 0; x < rect.columns && !token.IsCancellationRequested; x++) {
-         for (int y = 0; y < rect.rows; y++) {
+      int col = 0;
+      for (; col < rect.columns && !token.IsCancellationRequested; col++) {
+         // Assume that we know that the inner loop is very fast.
+         // Therefore, polling once per column in the outer loop condition
+         // is sufficient.
+         for (int row = 0; row < rect.rows; row++) {
             // Simulating work.
             Thread.SpinWait(5000);
```
IEvangelist, line 50:
```suggestion
            Thread.SpinWait(5_000);
```

samples/snippets/visualbasic/VS_Snippets_Misc/cancellation/vb/cancellation.vbproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
IEvangelist, line 5:
```suggestion
    <TargetFramework>net6.0</TargetFramework>
```

IEvangelist (commented):
A few nits

samples/snippets/csharp/VS_Snippets_Misc/cancellation/cs/cancellationex11.cs
```diff
@@ -40,24 +40,25 @@ static void Main()
    //<snippet3>
    static void NestedLoops(Rectangle rect, CancellationToken token)
    {
-      for (int x = 0; x < rect.columns && !token.IsCancellationRequested; x++) {
-         for (int y = 0; y < rect.rows; y++) {
+      int col = 0;
+      for (; col < rect.columns && !token.IsCancellationRequested; col++) {
```
IEvangelist, line 44: I'm not a fan of having the `col` variable being declared outside of the `for` loop simply for the sake of logging it. I'd argue that we should delete line 57 altogether and instead declare this inline.
```suggestion
      for (int col = 0; col < rect.columns && !token.IsCancellationRequested; col++) {
```

PR #31737, opened 2022-10-12
Unpublish speclets

BillWagner:
The following features have been integrated into the C# 7 draft spec:

- local functions
- throw expressions
- readonly struct

This PR removes those speclets, and redirects all link to those speclets.


docs/csharp/toc.yml
```diff
@@ -1344,12 +1344,8 @@ items:
       items:
       - name: Pattern matching
         href: ../../_csharplang/proposals/csharp-7.0/pattern-matching.md
-      - name: Local functions
```
gewarren, line 1347: Was read-only structs not in the TOC?

gewarren (approved):

docs/csharp/toc.yml
BillWagner: It wasn't. That spec was a placeholder that didn't have any content.

PR #31742, opened 2022-10-12
Environment variable prefix breaking change

### LINKED ISSUE #31490: [Breaking change]: Improvement to handling of env prefixes containing __
## Key changes:

* By comparing normalized prefixes to normalized keys, both the un-altered and colon-normalized prefixes match environment variables in either the __ or colon formats.
* There is a change of behaviour that some environment variables which theoretically didn't match the filter before may now match the filter, but I don't expect this to be an issue in practice (I very much doubt anyone configures variables with both a `:` and `__` prefix, and relies on the fact that these don't match)
* Existing behaviour is maintained, to avoid any potential issues such as https://github.com/dotnet/runtime/issues/61577 or any changes of behaviour with the special case connection string matching.

### Version

.NET 7 Preview 4

### Previous behavior

Some environment variables which theoretically didn't match the filter before may now match the filter

### New behavior

By comparing normalized prefixes to normalized keys, both the un-altered and colon-normalized prefixes match environment variables in either the __ or colon formats.

### Type of breaking change

### Reason for change

Confusing environment variable configuration prefix behaviour was reintroduced in 6.0.2. (It worked on 6.0.0 and 6.0.1. Upgrading to 6.0.2 caused the break.) https://github.com/dotnet/runtime/pull/62819 was merged into 6.0.2 to revert a breaking change in 6.0.0, but the reversion caused a breaking change compared to 6.0.0/6.0.1, by reverting the (good) part of https://github.com/dotnet/runtime/pull/62819.

For reference, the [current configuration documentation is here](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-6.0#environment-variables).

### Recommended action

```cs
Environment.SetEnvironmentVariable("MY_PREFIX__ConfigKey, "myFooValue");
var configuration = new ConfigurationBuilder()
    .AddEnvironmentVariables("MY_PREFIX__")
    .Build();

var loadedValue = configuration.GetValue<string?>("ConfigKey", null);

Assert.Equals("myFooValue", loadedValue); // FAILED in 6.0.2 as loadedConfig was null. The prefix of `MY_PREFIX:` can be specified instead now
```

### Feature area

### Affected APIs

_No response_

gewarren:
[Breaking change preview link](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/extensions/7.0/environment-variable-prefix?branch=pr-en-us-31742).

Fixes #31490.

Also removes some confusing content from the environment variables config provider article.

### DIFF
--- a/docs/core/compatibility/7.0.md
+++ b/docs/core/compatibility/7.0.md
@@ -71,6 +71,7 @@ If you're migrating an app to .NET 7, the breaking changes listed here might aff
 | Title | Binary compatible | Source compatible | Introduced |
 | - | :-: | :-: | - |
 | [ContentRootPath for apps launched by Windows Shell](extensions/7.0/contentrootpath-hosted-app.md) | ❌ | ✔️ | Preview 6 |
+| [Environment variable prefixes](extensions/7.0/environment-variable-prefix.md) | ❌ | ✔️ | Preview 4 |
 
 ## Globalization
 
--- a/docs/core/compatibility/extensions/7.0/environment-variable-prefix.md
+++ b/docs/core/compatibility/extensions/7.0/environment-variable-prefix.md
@@ -47,6 +47,7 @@ var loadedValue = config.GetValue<string?>("ConfigKey", null);
 
 Console.WriteLine(String.Equals(myValue, loadedValue));
 // True
+```
 
 ## New behavior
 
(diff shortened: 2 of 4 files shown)


gewarren:
cc @dhedey.

IEvangelist (approved):

PR #31747, opened 2022-10-12
Unsafe package breaking change

### LINKED ISSUE #31682: [Breaking change]: Removal of System.Runtime.CompilerServices.Unsafe package
### Description

We will stop building new versions of [System.Runtime.CompilerServices.Unsafe](https://www.nuget.org/packages/System.Runtime.CompilerServices.Unsafe) package in 7.0. The main reason we did this was that we used to have two different implementations of Unsafe and now we are unifying into the one that is shipping inbox, so no need to continue to ship a package that would just type forward to the inbox one. For more information, please go to https://github.com/dotnet/runtime/pull/64861.

### Version

.NET 7 Preview 3

### Previous behavior

New versions of `System.Runtime.CompilerServices.Unsafe` package would be produced with new versions of .NET Core.

### New behavior

New functionality will now ship as part of the shared framework.

### Type of breaking change

### Reason for change

We used to have two different implementations of the `Unsafe` type, one that was referenced and used inside of our core libraries, and one that would be on top and also ship as a separate NuGet package. This caused code duplication and also caused JIT and AOT maintenance burden so we decided to instead just remove one and unify their implementations. As a side effect, the NuGet package that was produced earlier no longer needs to ship, as the unified implementation of `Unsafe` ships inbox as part of the shared framework.

### Recommended action

You can continue to use older versions of the packages if you are targeting .NET 6.0 or lower, but starting from .NET 7.0 you should stop depending on the package and instead just use the API that comes as part of the shared framework.

### Feature area

Core .NET libraries

### Affected APIs

All of the APIs under the `Unsafe` type.

gewarren:
Fixes #31682.

[Preview link](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/7.0/unsafe-package?branch=pr-en-us-31747).


IEvangelist (approved):

PR #31750, opened 2022-10-12
Update high-performance-logging.md with note

collinstevens:
## Summary

Adds a note informing people to use the compile-time logging source generation available in .NET 6 instead of the `LoggerMessage.Define` API previously required.

The `Important` note is similar to the note found on https://learn.microsoft.com/en-us/dotnet/standard/base-types/composite-formatting.

I figured this should be added here because it seems like the `LoggerMessageAttribute` replaces the `LoggerMessage` class almost entirely per the summary available here https://learn.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator#summary.


IEvangelist (approved):
It's not obsolete, as your PR title states, but this change itself is accurate.

collinstevens:
> It's not obsolete, as your PR title states, but this change itself is accurate.

What's the use case in lieu of the attribute if you don't mind? After reading the summary it seemed as thought it's a complete replacement.

Issue #28810, opened 2022-03-25
[Breaking change]: Stop generating runtimeconfig.dev.json by default


marcpopMSFT:
### Description

In the .NET 6 SDK, we stopped generating the runtimeconfig.dev.json file by default as it was not used in most scenarios and added small amounts of performance overhead to find and load that file.

This was changed as of 6.0.100-preview5. We missed it from our breaking change documentation from what I can tell as it shouldn't affect most customers. This did impact a customer who reported this and requested documentation on the change: https://github.com/dotnet/sdk/issues/24427

### Version

Other (please put exact version in description textbox)

### Previous behavior

Appname.runtimesettings.dev.json would be automatically generated during compile time of the application

### New behavior

Appname.runtimesettings.dev.json is no longer generated by default

### Type of breaking change

- [X] **Binary incompatible**: Existing binaries may encounter a breaking change in behavior, such as failure to load/execute or different run-time behavior.
- [ ] **Source incompatible**: Source code may encounter a breaking change in behavior when targeting the new runtime/component/SDK, such as compile errors or different run-time behavior.

### Reason for change

This provided a small but measureable performance improvement.
https://github.com/dotnet/sdk/issues/16818
https://github.com/dotnet/sdk/pull/17014

### Recommended action

Add `<GenerateRuntimeConfigDevFile>true</GenerateRuntimeConfigDevFile>` to your project if you still require this file.

### Feature area

SDK

PR #31752, opened 2022-10-13
Fix filename in runtimeconfig.dev.json breaking change doc

ericstj:
## Summary

The filename was incorrect in parts of this document.  Fixing it to be correct and consistent.

Related https://github.com/dotnet/docs/issues/28810


gewarren (approved):
Thanks @ericstj

PR #31755, opened 2022-10-13
Delete unused snippets

gewarren:
Found these using http://github.com/gewarren/cleanrepo.


IEvangelist (changes requested):
I'm not sure how I feel about this PR. The first several files are something that should have been included in related content, I'm addressing that concern in https://github.com/dotnet/docs/pull/31757. The next several files were something that I recently added, knowing full well that it wasn't included in the content yet. The idea was that I'd come back to it, and at least have it in dotnet/samples (ideally, want to keep those synched). There are so many files that it's really challenging to ensure that there are more situations like that. I'm torn.

IEvangelist (approved):

PR #31767, opened 2022-10-13
Add an additional resource to rabbitmq-development

awright18:
## Summary
Add NServiceBus as a solution with support for  RabbitMQ


IEvangelist (approved):

PR #31770, opened 2022-10-13
MemoryCache breaking change

### LINKED ISSUE #31739: [Breaking change]: Extend MemoryCacheOptions with TrackLinkedCacheEntries
## Background and Motivation

`MemoryCache` by default tracks linked cache entries to allow for propagating options. 

In the following example, the `expirationToken` added for `child` is also applied to `parent`:

```cs
using (var parent = cache.CreateEntry(key))
{
    parent.SetValue(obj);

    using (var child = cache.CreateEntry(key1))
    {
        child.SetValue(obj);
        child.AddExpirationToken(expirationToken);
    }
}
```

Tracking internally uses `AsyncLocal<T>`, which is expensive and adds non-trivial overhead. The cost of using it has popped out in the performance reports that we have received from our customers (#45436).

Before this feature this behavior could not be disabled. 

## New API

Here we extend the existing `MemoryCacheOptions` with a flag that disables this behaviour.

```diff
namespace Microsoft.Extensions.Caching.Memory
{
    public class MemoryCacheOptions
    {
+        public bool TrackLinkedCacheEntries { get ; set; }
    }
}
```

## Usage Examples

```cs
var options = new MemoryCacheOptions
{
    TrackLinkedCacheEntries = false
};

var cache = new MemoryCache(options);
```

## Risks

We have changed the defaults, it's now disabled by default.

### Version

.NET 7 Preview 1

### Previous behavior

MemoryCache by default tracks linked cache entries to allow for propagating options.

### New behavior

Using `MemoryCacheOptions.TrackLinkedCacheEntries` this behavior can be configured properly.

### Type of breaking change

### Reason for change

Performance improvement

### Recommended action

Refer to usage sample.

### Feature area

### Affected APIs

```cs
namespace Microsoft.Extensions.Caching.Memory
{
    partial class MemoryCacheOptions
    {
        public bool TrackLinkedCacheEntries { get ; set; } = false;
    }
}
```

gewarren:
Fixes #31739.

[Preview link](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/7.0/memorycache-tracking?branch=pr-en-us-31770).

### DIFF
--- a/docs/core/compatibility/core-libraries/7.0/memorycache-tracking.md
+++ b/docs/core/compatibility/core-libraries/7.0/memorycache-tracking.md
@@ -5,7 +5,7 @@ ms.date: 10/13/2022
 ---
 # Tracking linked cache entries
 
-<xref:System.Runtime.Caching.MemoryCache> tracks linked cache entries to that options are propagated. In the following example, the `expirationToken` added for `child` is also applied to `parent`:
+<xref:System.Runtime.Caching.MemoryCache> tracks linked cache entries so that options are propagated. In the following example, the `expirationToken` added for `child` is also applied to `parent`:
 
 ```csharp
 using (var parent = cache.CreateEntry(key))
--- a/docs/core/compatibility/core-libraries/7.0/memorycache-tracking.md
+++ b/docs/core/compatibility/core-libraries/7.0/memorycache-tracking.md
@@ -1,0 +1,61 @@
+---
+title: "Breaking change: Tracking linked cache entries"
+description: Learn about the .NET 7 breaking change in .NET where MemoryCache no longer tracks linked cache entries by default.
+ms.date: 10/13/2022
+---
+# Tracking linked cache entries
+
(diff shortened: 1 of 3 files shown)


docs/core/compatibility/core-libraries/7.0/memorycache-tracking.md
```diff
@@ -0,0 +1,61 @@
+---
+title: "Breaking change: Tracking linked cache entries"
+description: Learn about the .NET 7 breaking change in .NET where MemoryCache no longer tracks linked cache entries by default.
+ms.date: 10/13/2022
+---
+# Tracking linked cache entries
+
+<xref:System.Runtime.Caching.MemoryCache> tracks linked cache entries to that options are propagated. In the following example, the `expirationToken` added for `child` is also applied to `parent`:
```
CamSoper, line 8: I think you meant:

```suggestion
<xref:System.Runtime.Caching.MemoryCache> tracks linked cache entries so that options are propagated. In the following example, the `expirationToken` added for `child` is also applied to `parent`:
```

CamSoper (approved):
I think I found a typo you might check out.

IEvangelist (approved):

PR #31773, opened 2022-10-14
Update Program.cs

WhitWaldo:
## Summary

There is no Parameter type, rather this is intended to be Expression.Parameter. Minor tweak to sample code.

### DIFF
--- a/samples/snippets/csharp/programming-guide/dynamic-linq-expression-trees/Program.cs
+++ b/samples/snippets/csharp/programming-guide/dynamic-linq-expression-trees/Program.cs
@@ -112,7 +112,7 @@ var fixedQry = companyNames.OrderBy(x => x);
 
 {
     // <Factory_method_expression_tree_parameter>
-    ParameterExpression x = Parameter(typeof(string), "x");
+    ParameterExpression x = Expression.Parameter(typeof(string), "x");
     // </Factory_method_expression_tree_parameter>
 
     // <Factory_method_expression_tree_body>


samples/snippets/csharp/programming-guide/dynamic-linq-expression-trees/Program.cs
```diff
@@ -112,7 +112,7 @@
 
 {
     // <Factory_method_expression_tree_parameter>
-    ParameterExpression x = Parameter(typeof(string), "x");
+    ParameterExpression x = Expression.Parameter(typeof(string), "x");
```
IEvangelist, line 115: On line 6 of this file, there is a `using static` making this redundant. If the conceptual article this is referenced in needs to be more explicit, consider also removing line 6 in the C# file to avoid confusion.

```cs
using static System.Linq.Expressions.Expression;
```

IEvangelist (commented):
Nit

samples/snippets/csharp/programming-guide/dynamic-linq-expression-trees/Program.cs
WhitWaldo: You're right - as someone who was following the article working through a local proof of concept, I completely missed that necessary `using static` despite the note about it at the top of the documentation. I might suggest dropping the `using static` bit despite the callout and just using Expression.* throughout as I would it's more likely that the developer has included the `using System.Linq.Expressions` already given the subject matter and less likely that they started reading the page from the very top.

IEvangelist: For the sake of docs, it's fine to be explicit.

IEvangelist (approved):

PR #31806, opened 2022-10-17
[Tiny] Fix one single Event ID

pjanotti:
See https://github.com/dotnet/runtime/blob/4719135a5fd6c9f22ce429ad8a211d696eae0426/src/coreclr/nativeaot/Runtime/EtwEvents.h#L344 for the correct ID.

## Summary

Just fixed the event ID - I didn't check the others, noticed this one reviewing a [PR](https://github.com/signalfx/signalfx-dotnet-tracing/pull/643#discussion_r997324739) that was using id 9 that I didn't see in the docs.


IEvangelist (approved):

PR #31935, opened 2022-10-19
Fix Standalone GC table format

sebastienros:
## Summary

The table markdown is badly formatted and doesn't render as a table right now:
https://learn.microsoft.com/en-us/dotnet/core/runtime-config/garbage-collector#standalone-gc

image

### DIFF
--- a/docs/core/runtime-config/garbage-collector.md
+++ b/docs/core/runtime-config/garbage-collector.md
@@ -557,6 +557,7 @@ Example:
 ## Standalone GC
 
 - Specifies the name of a standalone GC DLL that the runtime loads in place of the one it has in the main runtime DLL (coreclr.dll). This DLL needs to reside in the same directory as *coreclr.dll*
+
 | | Setting name | Values | Version introduced |
 | - | - | - | - |
 | **runtimeconfig.json** | N/A | N/A | N/A |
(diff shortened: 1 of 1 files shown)


docs/core/runtime-config/garbage-collector.md
```diff
@@ -557,6 +557,7 @@ Example:
 ## Standalone GC
 
 - Specifies the name of a standalone GC DLL that the runtime loads in place of the one it has in the main runtime DLL (coreclr.dll). This DLL needs to reside in the same directory as *coreclr.dll*
```
gewarren, line 559:
```suggestion
- Specifies the name of a standalone GC DLL that the runtime loads in place of the one it has in the main runtime DLL (coreclr.dll). This DLL needs to reside in the same directory as *coreclr.dll*.
```

gewarren (approved):
Thank you @sebastienros!

docs/core/runtime-config/garbage-collector.md
sebastienros: Maybe you were trying to be polite, but I would not have minded if you had updated the PR directly in GH. (assuming you know you can do it, even on forks)

gewarren: Yes, I wanted to be polite but would have eventually committed it if necessary. Thanks!

PR #31940, opened 2022-10-20
Clarify naming rule order

patricksadowski:
I came across the problem described in https://github.com/dotnet/roslyn/issues/23884. After some research I found the implementation details. Now it's time to explain it in the documentation.

### DIFF
--- a/docs/fundamentals/code-analysis/style-rules/naming-rules.md
+++ b/docs/fundamentals/code-analysis/style-rules/naming-rules.md
@@ -81,7 +81,7 @@ You can set the following properties for symbol groups, to limit which symbols a
 
 1. Tuple members aren't currently supported in `applicable_kinds`.
 2. The symbol group matches _all_ the modifiers in the `required_modifiers` property.  If you omit this property, no specific modifiers are required for a match. This means a symbol's modifiers have no effect on whether or not this rule is applied.
-3. If your group has `static` or `shared` in the `required_modifiers` property, the group will also include `const` symbols because they are implicitly `static`/`Shared`. However, if you don't want the `static` naming rule to apply to `const` symbols, you can create a new naming rule with a symbol group of `const`.
+3. If your group has `static` or `shared` in the `required_modifiers` property, the group will also include `const` symbols because they are implicitly `static`/`Shared`. However, if you don't want the `static` naming rule to apply to `const` symbols, you can create a new naming rule with a symbol group of `const` which affects the [rule order](#rule-order).
 4. `class` includes [C# records](../../../csharp/language-reference/builtin-types/record.md).
 
 ### Naming style properties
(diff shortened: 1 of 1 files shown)


docs/fundamentals/code-analysis/style-rules/naming-rules.md
```diff
@@ -81,7 +81,7 @@ You can set the following properties for symbol groups, to limit which symbols a
 
 1. Tuple members aren't currently supported in `applicable_kinds`.
 2. The symbol group matches _all_ the modifiers in the `required_modifiers` property.  If you omit this property, no specific modifiers are required for a match. This means a symbol's modifiers have no effect on whether or not this rule is applied.
-3. If your group has `static` or `shared` in the `required_modifiers` property, the group will also include `const` symbols because they are implicitly `static`/`Shared`. However, if you don't want the `static` naming rule to apply to `const` symbols, you can create a new naming rule with a symbol group of `const`.
+3. If your group has `static` or `shared` in the `required_modifiers` property, the group will also include `const` symbols because they are implicitly `static`/`Shared`. However, if you don't want the `static` naming rule to apply to `const` symbols, you can create a new naming rule with a symbol group of `const` which affects the [rule order](#rule-order).
```
gewarren, line 84:
```suggestion
3. If your group has `static` or `shared` in the `required_modifiers` property, the group will also include `const` symbols because they are implicitly `static`/`Shared`. However, if you don't want the `static` naming rule to apply to `const` symbols, you can create a new naming rule with a symbol group of `const`. The new rule will take precedence according to the [rule order](#rule-order).
```

docs/fundamentals/code-analysis/style-rules/naming-rules.md
```diff
@@ -122,7 +122,7 @@ All naming rule properties are required for a rule to take effect.
...
+The order in which naming rules are defined in an EditorConfig file doesn't matter. The naming rules are automatically ordered according to the definition of the rules themselves. More specific rules regarding accessibilities, modifiers, and symbols match ealier than less specific rules. If there is overlap between rules or if the rule ordering causes problems, the intersection of the two rules can be broken out into a new rule that always matches earlier than the broader rules which it was derived. The [EditorConfig Language Service extension](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.EditorConfig) can analyze an EditorConfig file and report cases where the rule ordering in the file is different to what the compiler will use at run time.
```
gewarren, line 125:
```suggestion
The order in which naming rules are defined in an EditorConfig file doesn't matter. The naming rules are automatically ordered according to the definitions of the rules themselves. More specific rules regarding accessibilities, modifiers, and symbols take precedence over less specific rules. If there is overlap between rules or if the rule ordering causes problems, you can break out the intersection of the two rules into a new rule that takes precedence over the broader rules from which it was derived.

The [EditorConfig Language Service extension](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.EditorConfig) can analyze an EditorConfig file and report cases where the rule ordering in the file is different to what the compiler will use at run time.
```

gewarren: It would be really helpful to add an example here to illustrate the intersection and break out info.

gewarren (approved):
Thanks for adding the examples!

PR #31948, opened 2022-10-20
Update README.md

IEvangelist:


CamSoper (approved):
We don't need no stinking badges!

PR #31972, opened 2022-10-22
Fix errors of Buffers Data Access snippet

sunghwan2789:
* fix compile error that no overload matches for GetPosttion(SequencePosition, long)
* fix getting wrong position by using advanced position


BillWagner (approved):
Thanks @sunghwan2789 

I've reviewed this and I'll :shipit: now

PR #32007, opened 2022-10-24
Fix PoliChek hit: Newfoundland

v-thepet:
## Summary

User Story https://dev.azure.com/mseng/TechnicalContent/_workitems/edit/2001296

The term is "Newfoundland". OK when referring to the island but not the province (Newfoundland and Labrador). This image could be a photo of the island, or of a map showing the province. I tracked down the image to Wikimedia Commons and it's actually a photo of an interior with a dresser. So I renamed the label to "dresser".

https://dev.azure.com/msft-skilling/Content/_workitems/edit/14585
https://dev.azure.com/msft-skilling/Content/_workitems/edit/14619
https://dev.azure.com/msft-skilling/Content/_workitems/edit/14609
https://dev.azure.com/msft-skilling/Content/_workitems/edit/14732

### DIFF
--- a/docs/machine-learning/tutorials/snippets/image-classification/csharp/assets/inception/imagenet.csv
+++ b/docs/machine-learning/tutorials/snippets/image-classification/csharp/assets/inception/imagenet.csv
@@ -58,7 +58,7 @@ image57,Lhasa
 image58,impala
 image59,coyote
 image60,Yorkshire terrier
-image61,Newfoundland
+image61,dresser
 image62,brown bear
 image63,red fox
 image64,Norwegian elkhound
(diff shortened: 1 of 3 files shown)


v-thepet:
#sign-off

IEvangelist (approved):

gewarren:
@v-thepet I'm curious how you tracked the image down - can you point me to it? Also, I'm not sure that these assets are actually used anywhere. The link to download the assets seems to point to the dotnet/samples repo now (which still contains the flagged term).
Finally, I'm guessing that "Newfoundland" in this case was referring to the dog breed, since the other images in the vicinity are either dog breeds or animals.
cc @luisquintanilla

v-thepet:
@gewarren @luisquintanilla - Here's what I think is that image: https://commons.wikimedia.org/wiki/File:Newfoundland_108_(2634775216).jpg. To find that, I backtracked from the images that are being used in the tutorial. You are right that most of them aren't being used, but the ones that are, all pointed to locations on wikimedia commons. The image filenames aren't exactly equivalent to the labels in the lists. 

You're also correct that in the sample repo or in other places where this Tensorflow dataset is available, the Newfoundland label remains. The aim in this PoliChek work item is to get rid of the term in the Microsoft docs if possible, to avoid future hits, even if it's being used in a context that isn't technically incorrect. I figured this usage was inconsequential, so I changed it.

Issue #28630, opened 2022-03-11
Article isn't clear on how to handle IAsyncDisposable without IDisposable


bitbonk:
If I read the article right, it is OK to implement _only_ `IAsyncDisposable` without `IDisopsable` (also for non-sealed classes, the article could be clearer in that regard).

But the article does not show a _full_ example on how the async disposable pattern looks like when `IDisposable` is not implemented. The first full example shown (in [Implements the async dispose pattern](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#implement-the-async-dispose-pattern)) already has `IDisposable` implemented.
In fact, right thereafter the article then says:
> You may need to implement both the IDisposable and IAsyncDisposable

Yes, we already did that right in the example above.

I think the article should be a bit clearer on how to handle `IAsncDisposable` alone and include a full example for this.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 61a961dc-3ebf-2a96-a6eb-2e0b62ab909f
* Version Independent ID: d72276e1-9059-815f-dc14-a03a7544ab87
* Content: [Implement a DisposeAsync method](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync)
* Content Source: [docs/standard/garbage-collection/implementing-disposeasync.md](https://github.com/dotnet/docs/blob/main/docs/standard/garbage-collection/implementing-disposeasync.md)
* Product: **dotnet-fundamentals**
* GitHub Login: @IEvangelist
* Microsoft Alias: **dapine**

IEvangelist:
Hi @bitbonk, thank you for posting this issue. I agree with you that this article needs some work, and the points that you raise are valid. I'll work on addressing this.

IEvangelist:
Related to #28630

lg2de:
@IEvangelist, you agreed with @bitbonk that the article must explain the use of `IAsyncDisposable` alone.
But now you have closed this issue together with the PR which reports (only) a potential resource leak.
I still don't find an explanation for the correct implementation of `IAsyncDisposable` WITHOUT `IDisposable` in the article. I think this issue is still to be finished and should be reopened.

bitbonk:
@IEvangelist I agree. Even with the new change, the article still reads that it is OK to _only_ implement `IAsyncDisposable` (without `IDisposable` and yet the article does now show how this is done.

[...]

PR #32040, opened 2022-10-26
Added more examples, corrected existing, and revised impl patterns

### LINKED ISSUE #32039: Revise async dispose pattern concept article
@IEvangelist I agree. Even with the new change, the article still reads that it is OK to _only_ implement `IAsyncDisposable` (without `IDisposable`) and yet the article does  not show how this is done.

There is a header for the async-only dispose pattern: [Implement the async dispose pattern](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#implement-the-async-dispose-pattern).
And there is a header for the sync+async dispose pattern: [mplement both dispose and async dispose patterns](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#implement-both-dispose-and-async-dispose-patterns).

But _both_ examples implement `IAsyncDisposable` _and_ `IDisposable`. Both examples are almost the same.
What is more, the chapter _Implement both dispose and async dispose patterns_ even begins with the text
> You may need to implement both the [IDisposable](https://docs.microsoft.com/en-us/dotnet/api/system.idisposable) and [IAsyncDisposable](https://docs.microsoft.com/en-us/dotnet/api/system.iasyncdisposable) interfaces

Which seems to convey that were are now introducing `IDisposable` to the mix, but we did that aready in the example right before that.

After reading the article now, we still don't now how to implement `IAsyncDisposable` only.

_Originally posted by @bitbonk in https://github.com/dotnet/docs/issues/28630#issuecomment-1228725877_

IEvangelist:
## Summary

Added more examples, corrected existing, and revised impl patterns

Fixes #32039, related to #28630

### DIFF
--- a/docs/standard/garbage-collection/implementing-disposeasync.md
+++ b/docs/standard/garbage-collection/implementing-disposeasync.md
@@ -68,7 +68,7 @@ The `DisposeAsyncCore()` method is intended to perform the asynchronous cleanup 
 
 ## Implement the async dispose pattern
 
-All non-sealed classes should be considered a potential base class, because they could be inherited. If you implement the async dispose pattern for any potential base class, you must provide the `protected virtual ValueTask DisposeAsyncCore()` method. Here is an example implementation of the async dispose pattern that uses a <xref:System.Text.Json.Utf8JsonWriter?displayProperty=nameWithType>.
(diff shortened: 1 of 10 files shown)


CamSoper (approved):
Lgtm

PR #32041, opened 2022-10-26
Creating CS8210

balazsberes:
## Summary

Describe your changes here.


docs/csharp/language-reference/compiler-messages/cs8210.md
```diff
@@ -0,0 +1,35 @@
...
+{
+}
+
+void Test()
+{
+    var x = ("something", Method());
+}
+```
+
+Your method returns void, so that method doesn't return a value. You can't use a method that returns void for a data member of a tuple.
```
BillWagner, line 30: We should code fence `void`:

```suggestion
Your method returns `void`, so that method doesn't return a value. You can't use a method that returns `void` for a data member of a tuple.
```

BillWagner (approved):
This LGTM @balazsberes 

I'll merge it with the one formatting change I've suggested.

PR #32059, opened 2022-10-27
Removed: await finishedTask;

vcrobe:
## Summary

I've removed the `await finishedTask;` from line 244 because the task was awaited on line 231.


IEvangelist (approved):

minsouk:
I think Await finishedTask shouldn't be deleted. Await finishedTask was executed after receiving the already completed Task by Task.WhenAny. Therefore, waiting is not affected by execution time. Otherwise, exceptions that run asynchronously cannot be caught. Test the fire scenario.

PR #32062, opened 2022-10-27
Replaced "synchronous" with "asynchronous"

vcrobe:


IEvangelist:
Hi @vcrobe,
This change is incorrect. The `Begin` and `End` methods are in fact, synchronous. They start (`Begin` asynchronous work, and handle when that work is done (`End`), but they are synchronous.

BillWagner (approved):
Thanks @vcrobe 

This is correct @IEvangelist 

The *operation* is asynchronous. The `Begin` and `End` methods are synchronous. The `Begin` method starts the operation, and the `End` method reports its completion.

I'll :shipit:

PR #32064, opened 2022-10-27
Update cs0246.md

peter-dye:
I had this issue when I added a new project to our solution but by default it was added with a build configuration that we had removed the other projects from (including a referenced project). I've added what worked to fix the issue as a potential fix here.

## Summary

Added missing potential fix for compiler error CS0246.


BillWagner (approved):
That's a great addition @peter-dye 

I'll :shipit: now.

PR #32076, opened 2022-10-28
Modify log scope to use keyvalue pairs

cijothomas:
## Summary

The scopes example currently uses a plain string. This PR proposes to replace that with a KVP collection, containing the key "TransactionId". This matches the example use case that was described right above the example code snippet.

While Scope can accept any `object` including `string`, most backends expect additional enrichments to come in the form of key-value pairs, so this is demonstrating a better practice of using Scopes.


tarekgh (approved):

noahfalk (approved):
LGTM

PR #32114, opened 2022-11-01
Update consuming-the-task-based-asynchronous-pattern.md

vcrobe:
## Summary

- I think it is better to highlight the note at line 93: "Cancellation requests may be initiated from any thread."

- At line 146 we're talking about the same method as another overloaded version but in fact is the same overload method.


BillWagner (approved):
Thanks @vcrobe 

These changes look great. I'll :shipit: now.

PR #32123, opened 2022-11-02
Fix option heading level

patricksadowski:


gewarren (approved):
Thanks @patricksadowski!

PR #32155, opened 2022-11-03
Compiler Warning CS8509

balazsberes:
## Summary
Compiler Warning CS8509

PS: I remade the branch and PR because of a bad remote commit :).


balazsberes:
Hi @BillWagner ,

What do you think about this PR and approach?

If it's ok, in my next PR I would transfer the rest of the warning waves.

BillWagner:
Hi @balazsberes 

First, I'm thankful that you're helping us get better coverage for our compiler warnings and errors. It's an area where we've fallen behind. And badly.

I'm experimenting with a new organization for these pages. I've been updating new pages by coalescing related errors and warnings into a single page. The two areas where I've done this so far is the nullable related warnings, and warning waves. 

I'm not yet convinced this is a better strategy, and I'm measuring a lot of the analytics we use on our site to help me.

So, my first thought is this: Can you update the content you've added so all warning waves are on the same page. I'd like to continue to measure this over a few months, and add a few more of the combined pages before making a conclusion.

What are your thoughts?

BillWagner:
Hi @balazsberes 

I apologize, because I should have noticed this earlier. I admit I'd only skimmed the changes earlier this week because of everything happening with the release.

`CS8509` *isn't* a warning wave 5 warning. It's a level 2 warning that was introduced in C# 8. Warning waves are reserved for introducing warnings that may lead to breaking changes in later versions.

There are a number of warnings introduced related to pattern matching, but I haven't done all the investigation to put them all together in one place (or add many of them as stand along warning pages).

There are a few options.

We could revert back to a single warning page, add it in the TOC, but *not* link it to the warning waves page. That's simplest, and would help. It would also provide the first step toward adding all the pattern matching warnings.

Alternatively, we could expand into creating the page for all warnings related to pattern matching. That would be awesome, but I don't feel comfortable asking you to do that as a volunteer.

Let me know how you want to proceed. And again, thanks for wanting to help us improve. I apologize for missing this earlier and costing you extra time.

balazsberes:
@BillWagner I just updated it. What do you think?

docs/csharp/language-reference/compiler-messages/pattern-matching-warnings.md
```diff
@@ -0,0 +1,20 @@
...
+
+There are several pattern matching warnings. Learn how to address these warnings.
+
+- **CS8509** - *The switch expression does not handle all possible values of its input type (it is not exhaustive). For example, the pattern '...' is not covered.*
+
+The following code snippets generate CS8509:
+
+:::code language="csharp" source="./snippets/pattern-matching-warnings/Switch.cs" id="SwitchNotAllPossibleValues":::
+
+For more information, see [Switch](../statements/selection-statements.md#the-switch-statement).
```
BillWagner, line 20: Let's add a paragraph about how a reader can address this warning:

```suggestion
To address this warning, add switch arms that cover all possible input values. For example:

```csharp
var result = enumValues switch
{
    EnumValues.Value1 => 1,
    EnumValues.Value2 => 2,
    EnumValues.Value3 => 3,
    _ => throw new ArgumentException("Input isn't a valid enum value", nameof(enumValues));
};
````

The `_` pattern matches all remaining values. It's often used to match invalid values, as shown in the preceding example.

For more information, see [Switch](../statements/selection-statements.md#the-switch-statement).
```

BillWagner: The last sentences of the previous comment should be added / restored as well. The first set of backticks is ending the entire suggestion, where I meant just to end the C# sample.

Also, it would be better to update the code file by adding a new method, but that's a lot harder to do as a suggestion in the UI.

docs/csharp/language-reference/compiler-messages/warning-waves.md
```diff
@@ -5,6 +5,7 @@ ms.date: 05/11/2022
 f1_keywords:
   - "CS7023"
   - "CS8073"
+  - "CS8509"
```
BillWagner, line 8: These two changes should be removed. The F1 help for this error shouldn't go to this page anymore.

```suggestion
```

docs/csharp/language-reference/compiler-messages/warning-waves.md
```diff
@@ -22,6 +23,7 @@ f1_keywords:
 helpviewer_keywords: 
   - "CS7023"
   - "CS8073"
+  - "CS8509"
```
BillWagner, line 26: Same as above. THis should be removed:

```suggestion
```

BillWagner (commented):
Overall, this is great. I had a couple of small comments, and then this is ready to :shipit:

Thanks again for all your work, and the conversation to bring this to completion.

balazsberes:
@BillWagner I just updated it. What do you think?

BillWagner (approved):
Thanks @balazsberes 

Let's :shipit:

I like this a lot.

PR #32158, opened 2022-11-03
Correct some grammatical errors

Stancobridge:
It makes sense that we are to include only one act per test, which validates that when a test fails, it is clear which act is failing. But the documentation has a grammatical error about that: When the test fails, it isn't clear which act is failing.

This PR corrects it


IEvangelist (approved):

PR #32161, opened 2022-11-03
24 hour format

neilboyd:
## Summary

Update the [samples](https://learn.microsoft.com/en-us/dotnet/core/extensions/console-log-formatter) to display output in 24 hour format, eg in the afternoon it will display `16:45:56` instead of `04:45:56`.


IEvangelist (approved):

PR #32171, opened 2022-11-03
system.diagnostics config entry breaking change

### LINKED ISSUE #32147: [Breaking change]: app.config cannot have a `<section name="system.diagnostics"` entry
### Description

For applications that have an app.config, the `<configuration><configSection>` entry may contain a `<section name="system.diagnostics"` entry which is no longer allowed. If present, the entry needs to be removed.

Having a `<section name="system.diagnostics"` entry causes the following exception the first time the configuration system is used:
```ConfigurationErrorsException: Section or group name 'system.diagnostics' is already defined. Updates to this may only occur at the configuration level where it is defined. ```

For example, this app.config file:
```
<configuration>
 <configSections>
   <section name="system.diagnostics"
            type="System.Diagnostics.SystemDiagnosticsSection,
            System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089"/>
 </configSections>  
```
contains the unnecessary entry.

### Version

.NET 7 RC 1

### Previous behavior

Specifying `<section name="system.diagnostics"` was allowed and necessary if there was a later `system.diagnostics` configuration section like:
```
<configuration>
    <system.diagnostics>
```

However, System.Diagnostics did not yet support the feature that processes that section in order to add various listeners and other control other diagnostic features, so it was not automatically read. Only applications which had a`<system.diagnostics>` section would require a `<section name="system.diagnostics"` entry.

.NET Framework, however, does support processing of the "system.diagnostics" section and had an `<section name="system.diagnostics"` in the machine.config file.

Thus, this issue likely only occurs for:
- Applications that were using .NET Framework and migrated to .NET.
- Manually added  `<section name="system.diagnostics"` to the app.config to support manual reading of the `<system.diagnostics>` section.
- Upgraded to 7.0 which now has an implicit  `<section name="system.diagnostics"` entry.

### New behavior

System.Diagnostics now supports reading the `<system.diagnostics>` section from the config file and added an implicit `<section name="system.diagnostics"` entry. Having an explicit `<section name="system.diagnostics"` entry in the app.config file causes a duplicate which throws `ConfigurationErrorsException`.

### Type of breaking change

### Reason for change

The implementation to support the feature where System.Diagnostics reads from the app.config required adding the implicit  `<section name="system.diagnostics"` element. See https://github.com/dotnet/runtime/issues/23937 for details on the feature request.

### Recommended action

The unnecessary  `<section name="system.diagnostics"` entry should be removed.

### Feature area

Other (please put exact area in description textbox)

### Affected APIs

_No response_

### PR DESCRIPTION
**Pull request by gewarren**
Fixes #32147.

[Preview link](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/configuration/7.0/diagnostics-config-section?branch=pr-en-us-32171).

### CODE REVIEW

File: docs/core/compatibility/configuration/7.0/diagnostics-config-section.md [hunk 1]
```diff
@@ -0,0 +1,66 @@
+---
+title: "Breaking change: System.diagnostics entry in app.config"
+description: Learn about the .NET 7 breaking change in configuration where the `<section name="system.diagnostics">` entry is no longer allowed in an app.config file.
+ms.date: 11/03/2022
+---
+# System.diagnostics entry in app.config
+
+For applications that have an *app.config* file, the `<configuration><configSection>` entry may contain a `<section name="system.diagnostics">` entry, which is no longer allowed. If present, you must remove the entry.
```
BillWagner, line 8: On L18, `<configSections>` (plural) is used. Are both uses valid? Or which one is correct?

BillWagner (approved):
This LGTM, once we resolve one inconsistency I noticed in the section names.

Then, let's :shipit:

docs/core/compatibility/configuration/7.0/diagnostics-config-section.md
gewarren: Good catch! I believe plural is correct.

DrSherlockWho (approved):

DrSherlockWho (approved):

PR #32172, opened 2022-11-03
Let's try this quest thing out

IEvangelist:
## Summary
Let's try this quest thing out

### DIFF
--- a/.github/workflows/quest.yml
+++ b/.github/workflows/quest.yml
@@ -1,0 +1,42 @@
+name: "quest import"
+on:
+  issues:
+    types:
+      [ labeled ]
+  workflow_dispatch:
+    inputs:
+      reason:
+        description: "The reason for running the workflow"
+        required: true
+        default: "Manual run"
+      issue:
+        description: "The issue number to manually test"
+        required: true
+
+jobs:
+  import:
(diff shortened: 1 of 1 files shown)


.github/workflows/quest.yml
```diff
@@ -0,0 +1,42 @@
...
+        description: "The issue number to manually test"
+        required: true
+
+jobs:
+  import:
+    if: github.event.label.name == join(fromJson('[":", "world_map", ":", " reQUEST"]'), '') || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
```
BillWagner, line 22: This shouldn't need to write pull requests.

BillWagner (approved):
One small comment, and then let's try it.

I'm not seeing how the command line args get to the app, but that might be just my ignorance.

.github/workflows/quest.yml
IEvangelist, line 22:
```suggestion
      issues: write
```

IEvangelist, line 22:
```suggestion
      issues: write
```

PR #32184, opened 2022-11-04
Add defaultconstraint_CSharpKeyword

Youssef1313:
Related to https://github.com/dotnet/roslyn/pull/65229

Let me know if there is a documentation page that you think is a better fit.


BillWagner (approved):
This LGTM @Youssef1313 

In time, I plan to reorganize some of this content. It may move then, but for now, this is the best destination.

PR #32190, opened 2022-11-04
Update quest.yml

IEvangelist:
Update quest trigger events.

### DIFF
--- a/.github/workflows/quest.yml
+++ b/.github/workflows/quest.yml
@@ -15,7 +15,7 @@ on:
 
 jobs:
   import:
-    if: github.event.label.name == join(fromJson('[":", "world_map", ":", " reQUEST"]'), '') || github.event_name == 'workflow_dispatch'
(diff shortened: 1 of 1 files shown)


.github/workflows/quest.yml
```diff
@@ -2,7 +2,7 @@ name: "quest import"
 on:
   issues:
     types:
-      [ labeled ]
+      [ labeled, closed, reopened, assigned, unassigned ]
```
BillWagner, line 5: With these additional events, the logic currently on L18 needs to be updated.

IEvangelist: I'm not sure how, could you tell me what the rules are for that scenario? I thought the idea would be to only run when there is a trigger label present regardless.

IEvangelist: So, add the inquest label as part of the condition too - right?

IEvangelist: Updated

BillWagner (approved):
This LGTM. Let's :shipit:

PR #32196, opened 2022-11-04
Update selection-statements.md

Youssef1313:


BillWagner (approved):
True. That's a bit of an edge case, but it's certainly correct.

PR #32199, opened 2022-11-04
Update quest.yml

IEvangelist:
Optimize workflow. This change will spin up the `ubuntu` VM if there is a Quest-specific label, or it is a manual run.

### DIFF
--- a/.github/workflows/quest.yml
+++ b/.github/workflows/quest.yml
@@ -18,7 +18,9 @@ jobs:
     if: |
       github.event_name == 'workflow_dispatch' ||
       (contains(github.event.label.name, 'world_map') && contains(github.event.label.name, 'reQUEST')) ||
-      (contains(github.event.label.name, 'pushpin') && contains(github.event.label.name, 'seQUESTered'))
+      (contains(github.event.label.name, 'pushpin') && contains(github.event.label.name, 'seQUESTered')) ||
+      (contains(github.event.issue.labels.*.name, 'world_map') && contains(github.event.issue.labels.*.name, 'reQUEST')) ||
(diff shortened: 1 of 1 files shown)


.github/workflows/quest.yml
```diff
@@ -15,7 +15,12 @@ on:
 
 jobs:
   import:
-    # if: github.event.label.name == join(fromJson('[":", "world_map", ":", " reQUEST"]'), '') || github.event_name == 'workflow_dispatch'
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      (contains(github.event.label.name, 'world_map') && contains(github.event.label.name, 'reQUEST')) ||
+      (contains(github.event.label.name, 'pushpin') && contains(github.event.label.name, 'seQUESTered')) ||
+      (contains(github.event.issue.labels.*.name, 'world_map') && contains(github.event.issue.labels.*.name, 'reQUEST')) ||
+      (contains(github.event.issue.labels.*.name, 'pushpin') && contains(github.event.issue.labels.*.name, 'seQUESTered'))
```
BillWagner, line 23: I think we can safely simplify this (and it will work in the repos that prefer not using the emojis in the labels):

```suggestion
      contains(github.event.label.name, 'reQUEST') ||
      contains(github.event.label.name, 'seQUESTered') ||
      contains(github.event.issue.labels.*.name, 'reQUEST') ||
      contains(github.event.issue.labels.*.name, 'seQUESTered')
```

BillWagner (approved):
This looks great.

I wonder if this simplification is safe?

.github/workflows/quest.yml
IEvangelist: Nice, check it out now!

PR #32200, opened 2022-11-04
Update dotnet-nuget-push.md

james-perfectserve:
The description for the --skip-duplicate option was phrased in a way that makes it easy to misunderstand. "so that the push can continue" makes it sound like the flag will let you overwrite/replace a published nuget package when there is a conflict

### DIFF
--- a/docs/core/tools/dotnet-nuget-push.md
+++ b/docs/core/tools/dotnet-nuget-push.md
@@ -69,7 +69,7 @@ The command pushes an existing package. It doesn't create a package. To create a
 
 - **`--skip-duplicate`**
 
-  When pushing multiple packages to an HTTP(S) server, treats any 409 Conflict response as a warning so that the push can continue.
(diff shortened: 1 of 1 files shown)


docs/core/tools/dotnet-nuget-push.md
```diff
@@ -69,7 +69,7 @@ The command pushes an existing package. It doesn't create a package. To create a
 
 - **`--skip-duplicate`**
 
-  When pushing multiple packages to an HTTP(S) server, treats any 409 Conflict response as a warning so that the push can continue.
+  When pushing multiple packages to an HTTP(S) server, treats any 409 Conflict response as a warning so that the other push(s) can continue.
```
tdykstra, line 72:
```suggestion
  When pushing multiple packages to an HTTP(S) server, treats any 409 Conflict response as a warning so that other pushes can continue.
```

tdykstra (commented):
Thanks! I made a suggestion to tweak the wording slightly.

IEvangelist (approved):

PR #32205, opened 2022-11-04
remove async main speclet

BillWagner:
This can be removed, as the updates were approved in the last C# standards meeting.

See dotnet/csharpstandard#70


IEvangelist (approved):

BillWagner:
@gewarren 

This is ready. The warning looks like a false positive. I've added a redirection for the removed file.

Issue #32225, opened 2022-11-07
[Breaking change]: Removal of .NET 6 Backward Compatibility Console Key Parsing


adamsitnik:
### Description

The ability to use the legacy `Console.ReadKey` implementation exposed via `System.Console.UseNet6CompatReadKey` and `DOTNET_SYSTEM_CONSOLE_USENET6COMPATREADKEY` settings has been removed.

### Version

.NET 8 Preview 1

### Previous behavior

It was possible to request .NET 6 console key parsing via JSON config fille and/or environment variable.

### New behavior

It's impossible to request the .NET 6 compatibility mode for `Console.ReadKey`.

### Type of breaking change

- [ ] **Binary incompatible**: Existing binaries may encounter a breaking change in behavior, such as failure to load/execute or different run-time behavior.
- [X] **Source incompatible**: Source code may encounter a breaking change in behavior when targeting the new runtime/component/SDK, such as compile errors or different run-time behavior.

### Reason for change

This mode was introduced in case the [Console.ReadKey rewrite](https://devblogs.microsoft.com/dotnet/console-readkey-improvements-in-net-7/) has introduced some bugs, as a safety switch. So far only one bug was reported, and it was fixed in .NET 7 RC2 so there is no need to keep the old implementation.

### Recommended action

If for some reason the new key parsing does not work as expected, please report a bug in [dotnet/runtime](https://github.com/dotnet/runtime/issues) so it gets fixed in .NET 8.

### Feature area

Core .NET libraries

### Affected APIs

`Console.ReadKey`

---
[Associated WorkItem - 61230](https://dev.azure.com/msft-skilling/Content/_workitems/edit/61230)

PR #32229, opened 2022-11-07
Remove auto release labeling of breaking changes

gewarren:
The labeling mechanism is too coarse-grained. For example, if it's a .NET 8 breaking change that talks about some functionality that was introduced in .NET 7, then the "Release: .NET 7" label gets added to the issue, and if you remove it, it gets readded.

E.g. https://github.com/dotnet/docs/issues/32225


IEvangelist (approved):

PR #32465, opened 2022-11-09
Fix escaping of term including backticks

nycdotnet:
## Summary

The existing text surrounded by back-ticks does not work since there is intentionally a single back-tick in the content.  Using the triple back-tick on the outside of the text does work.


nycdotnet:
Thanks!

PR #32503, opened 2022-11-10
Show extension method support in logger-message-generator.md

MattKotsenas:
## Summary

When introducing generated LoggerMessage templates to developers, many don't know that the template can be defined as an extension method. Some developers prefer the extension method syntax, as it maps more closely to the existing `LogInformation`, `LogWarning` extension methods.

I've updated the docs to also show an example of defining a logger message template as an extension method.

### DIFF
--- a/docs/core/extensions/logger-message-generator.md
+++ b/docs/core/extensions/logger-message-generator.md
@@ -28,7 +28,21 @@ public static partial class Log
 }
 ```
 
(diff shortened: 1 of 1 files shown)


MattKotsenas:
If there's a better place or way to give an example, please let me know. Thanks for taking a look!

IEvangelist (approved):

PR #32526, opened 2022-11-11
Fix one typo

hoang-himself:
## Summary

This typo is very insignificant, but a typo is a typo.


adegeo (approved):
Thank you for fixing this!

PR #32552, opened 2022-11-12
Add section about nulls

gusty:
## Summary

Based on this discussion https://github.com/dotnet/fsharp/discussions/14005 and some additional exchanges with @dsyme

### DIFF
--- a/docs/fsharp/style-guide/conventions.md
+++ b/docs/fsharp/style-guide/conventions.md
@@ -631,7 +631,19 @@ Aside from the single point of mutation in the middle of the lambda expression, 
 
 ## Nulls and default values
 
-Nulls should generally be avoided in F#. By default F#-declared types do not support the use of the `null` literal, and all values and objects are fully and correctly initialized. However, some common .NET APIs return or accept nulls, and some common .NET-declared types such as arrays and strings allow nulls. However, the occurrence of `null` values is very rare in F# programming and one of the benefits of using F# is to avoid null reference errors in most cases.
+Nulls should generally be avoided in F#. By default F#-declared types do not support the use of the `null` literal, and all values and objects are initialized. However, some common .NET APIs return or accept nulls, and some common .NET-declared types such as arrays and strings allow nulls. However, the occurrence of `null` values is very rare in F# programming and one of the benefits of using F# is to avoid null reference errors in most cases.
+
+### Avoid the use of the `AllowNullLiteral` attribute
+
+By default F#-declared types do not support the use of the `null` literal. You can manually annotate F# types with `AllowNullLiteral` to allow this. However, it is almost always better to avoid doing this.
+
+### Avoid the use of the `Unchecked.defaultof<_>` attribute
+
+It is possible to generate a `null` or zero-initialized value for an F# type by using `Unchecked.defaultof<_>`. This can be useful when initializing storage for some data structures, or in some high-performance coding pattern, or in interoperability. However the use of this construct should be avoided.
+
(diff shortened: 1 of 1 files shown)


docs/fsharp/style-guide/conventions.md
```diff
@@ -527,6 +527,10 @@ It's common to define single-case Discriminated Unions like this for domain mode
 
 Although the previous example showed that a struct Discriminated Union yielded better performance, it is common to have larger Discriminated Unions when modeling a domain. Larger data types like that may not perform as well if they are structs depending on the operations on them, since more copying could be involved.
 
+### Functional programming and nulls
+
+Nulls should normally be avoided, most F# types don't allow nulls, however some .NET types are frequently used if F# and they allow nulls, this is the case for arrays and strings. In practice there's no need to check for nulls but if you do fail fast, this applies to FSharp libraries as well. Although for legacy reasons some string functions in FSharp.Core still treat nulls as empty strings we shouldn't take them as guidance, that is we would not encourage attributing any semantic meaning to "null".
```
KathleenDollard, line 532: I really like that we are adding this, and my suggestions here are to add clarity to your points. This is sufficiently important that we can spend more space.

```suggestion
Nulls should generally be avoided in F#. F# supports nulls because some common .NET APIs return or accept nulls and some common types such as arrays and strings allow nulls. However, one of the benefits of using F# is to avoid null reference errors in most cases.

In practice there's no need to check for nulls but if you do fail fast, this applies to FSharp libraries as well. Although for legacy reasons some string functions in FSharp.Core still treat nulls as empty strings we shouldn't take them as guidance, that is we would not encourage attributing any semantic meaning to "null".
```

I am confused by "most F# types don't have nulls". Members? 

Can we clarify this? _In practice there's no need to check for nulls but if you do fail fast, this applies to FSharp libraries as well._ If you are working with .NET APIs, this does not necessarily hold. 

_Although for legacy reasons some string functions in FSharp.Core still treat nulls as empty strings we shouldn't take them as guidance, that is we would not encourage attributing any semantic meaning to "null"._

KathleenDollard: @vzarytovskii

gusty: from my conversation with @dsyme :
> ... libraries should fail fast for sure
> Should be added to the coding guidelines - "in practice no need to check for nulls but if you do fail fast

dsyme: I'll commit the suggestion and edit a little

dsyme:
I updated thanks, @gusty @KathleenDollard could you talk a look? Thanks!

dsyme (approved):

docs/fsharp/style-guide/conventions.md
```diff
@@ -629,6 +629,46 @@ let kernels =
...
+        let mutable result = false
+        let mutable i = 0
+        while not state && i < array.Length do
+            result <- value = array[i]
+            i <- i + 1
+        result
+```
+
+For legacy reasons some string functions in FSharp.Core still treat nulls as empty strings and do not fail on null arguments. However do not take this as guidance, and do not adopt coding patterns that attribute any semantic meaning to "null".
+
```
BillWagner, line 671: fix lint issue:

```suggestion

```

BillWagner (approved):
Thanks everyone.  This LGTM, modulo a couple grammar nits.

I'll fix those, and then :shipit:

gusty:
Looks good to me, thanks !

PR #32553, opened 2022-11-12
Add missing program output

selaskar:
## Summary

Adding missing output to the example. 
About to create additional pull request to update wording on the documentation page to accommodate test results.


selaskar:
On a second thought, these added results are not inline with the expected behavior described in the documentation section: https://learn.microsoft.com/en-us/dotnet/standard/base-types/backtracking-in-regular-expressions#backtracking-with-nested-optional-quantifiers

They expect unsuccessful match to take longer. But when I run the example in .NET 7/6/5 with VS 2022, the case actually takes much shorter than successful match case.

BillWagner (approved):
Thanks @selaskar 

This LGTM. I'll :shipit: now.

PR #32579, opened 2022-11-15
Add latest OS version in system requirements

ychikazawa:
## Summary

Added Windows 10 22H2.
Remove Windows 11 version. (This means including 22H2.)

Windows 11 is written without a version number in the following document.
This will reduce maintenance costs, but feel free to change it if it is better to note the version as well.
https://learn.microsoft.com/en-us/dotnet/framework/migration-guide/versions-and-dependencies
image


gewarren (approved):
Thanks @ychikazawa!

PR #32628, opened 2022-11-16
Improve surrogate examples

pentp:
Surrogate types should use fields instead of properties because `IConverter` and `IPopulator` pass them using read-only references and accessing properties could force the compiler to create local copies for each property access.


IEvangelist (approved):

PR #32664, opened 2022-11-18
Fix links Nov 2022

gewarren:
Contributes to #32569


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

PR #32667, opened 2022-11-18
Use updated whatsnew

BillWagner:
This version reuses the files rather than always using a new file. This means we won't have the redirection problem in the future.


gewarren (approved):
LGTM

PR #32670, opened 2022-11-18
CompressionLevel verification

### LINKED ISSUE #32620: [Breaking change]: BrotliStream ctor no longer allows values not defined in CompressionLevel enum
### Description

BrotliStream ctor no longer allows values not defined in CompressionLevel enum, an ArgumentException is thrown if that's the case.

### Version

.NET 7

### Previous behavior

BrotliStream allowed to pass in the level to the ctor by casting the desired level directly to CompressionLevel, like this:
```cs
BrotliStream brotli = new BrotliStream(baseStream,
                                       CompressionMode.Compress,
                                       leaveOpen,
                                       bufferSize,
                                       (CompressionLevel)5); // Use level 5
```

### New behavior

BrotliStream only allows the values defined in CompressionLevel, passing an undefined defined will result in an `ArgumentException`.

### Type of breaking change

### Reason for change

The purpose of the CompressionLevel enumeration is to let folks use compression algorithms without needing to understand the meaning of their tuning parameters.

If an arbitrary level was provided, that was passed through as-is to the underlying library, resulting in inconsistent and potentially unexpected behavior. With this change the behavior is aligned with other compression streams e.g: `DeflateStream`. 

With the tuning of the CompressionLevel values made in https://github.com/dotnet/runtime/pull/72266 and with the addition of `CompressionLevel.SmallestSize` https://github.com/dotnet/runtime/pull/41960, it is now possible to have a variety of trade-offs in the compression algorithms and users can keep relying in CompressionLevel as being an abstraction of such trade-off.

### Recommended action

If you were relying in passing undefined values as the CompressionLevel, be advised that you need to re-visit your use case and decide which [documented value](https://learn.microsoft.com/dotnet/api/system.io.compression.compressionlevel?view=net-7.0) is the most optimal for it.

### Feature area

Core .NET libraries

### Affected APIs

```cs
public BrotliStream (System.IO.Stream stream, System.IO.Compression.CompressionLevel compressionLevel);
public BrotliStream (System.IO.Stream stream, System.IO.Compression.CompressionMode mode);
public BrotliStream (System.IO.Stream stream, System.IO.Compression.CompressionLevel compressionLevel, bool leaveOpen);
public BrotliStream (System.IO.Stream stream, System.IO.Compression.CompressionMode mode, bool leaveOpen);
```

gewarren:
Fixes #32620 

[Internal preview](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/7.0/brotlistream-ctor?branch=pr-en-us-32670).

### DIFF
--- a/docs/core/compatibility/7.0.md
+++ b/docs/core/compatibility/7.0.md
@@ -40,6 +40,7 @@ If you're migrating an app to .NET 7, the breaking changes listed here might aff
 | [API obsoletions with default diagnostic ID](core-libraries/7.0/obsolete-apis-with-default-diagnostic.md) | ✔️ | ❌ | Preview 3 |
 | [API obsoletions with non-default diagnostic IDs](core-libraries/7.0/obsolete-apis-with-custom-diagnostics.md) | ✔️ | ❌ | Preview 1 |
 | [BinaryFormatter serialization APIs produce compiler errors](core-libraries/7.0/binaryformatter-apis-produce-errors.md) | ✔️ | ❌ | RC 1 |
+| [BrotliStream no longer allows undefined CompressionLevel values](core-libraries/7.0/brotlistream-ctor.md) | ❌ | ✔️ | |
 | [C++/CLI projects in Visual Studio](core-libraries/7.0/cpluspluscli-compiler-version.md) | ✔️ | ❌ | Preview 3 |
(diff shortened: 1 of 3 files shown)


jozkee (approved):
Thanks.

BillWagner (approved):
LGTM.

PR #32680, opened 2022-11-19
Add double-dash to install/uninstall options.

1offdev:
Examples for install/uninstall produce the following error:
```console
Error: Invalid option(s):
  ./
```

## Summary

Describe your changes here.


1offdev:
The examples are correct for the current SDK version (7.0.100) per the Note at the top of the https://learn.microsoft.com/dotnet/core/tools/dotnet-new-install page.

PR #32733, opened 2022-11-23
Orleans - update reentrancy.md to clarify `[AlwaysInterleave]` behavior

ReubenBond:
Requests for `[AlwaysInterleave]` methods always interleave and can always be interleaved, even by non-`[AlwaysInterleave]` methods.


IEvangelist (approved):

PR #32762, opened 2022-11-28
Document building against the Microsoft.NETFramework.ReferenceAssemblies NuGet package

ladipro:
## Summary

This pull request adds a new page documenting an alternative approach to targeting .NET Framework in non-SDK-style projects. This can be used for example when the official developer pack is no longer available, or if installing it on the machine is not feasible (e.g. CI scenarios).

Once merged, the plan is to link the new document from the .NET SDK download page (https://dotnet.microsoft.com/en-us/download/visual-studio-sdks) for the .NET Framework versions that don't have the corresponding developer pack available anymore.

VS flow:
1. User opens a legacy project in VS, VS shows a dialog advising the user to either upgrade or download developer pack.
2. User chooses to download developer pack, lands on the .NET SDK download page.
3. Instead of a download link, the .NET SDK download page contains a link to this doc.
4. User reads the doc, modifies the project accordingly, keeps using latest VS.

Command line flow:
1. User builds a legacy project, sees an MSBuild error advising the user to download the developer pack.
2. User follows the link in the error message.

then the same as 3. and 4. above.


ladipro:
CI is complaining about warnings in unrelated files. Am I doing something wrong?

docs/framework/install/reference-assemblies.md
```diff
@@ -0,0 +1,29 @@
...
+
+When you target a particular version of .NET Framework, by default your application is built by using the reference assemblies that are included with that version's developer pack. In scenarios where the matching developer pack cannot be installed on the computer, you can alternatively build against reference assemblies distributed via a NuGet package.
+
+## Update project files
+
+Each project that should build against the reference assemblies NuGet package needs to include the following package reference.
+
+```xml
+<PropertyGroup>
+  <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" PrivateAssets="All" />
```
zivkan, line 18: Doesn't the .NET SDK already automatically add this reference? I don't believe there should be a need for customers to ever need to do this themselves.

ladipro: Indeed, SDK-based projects should be getting the reference by default. I am primarily interested in non-SDK-style scenarios where the user would have to add the reference explicitly, just because that's what most legacy projects use. It looks like the guidance should be different for SDK and non-SDK here. Thank you!

BillWagner:
> CI is complaining about warnings in unrelated files. Am I doing something wrong?

@ladipro That should be fixed now. Once you make any new commits based on reviews, the next build should clear it.

BillWagner:

BillWagner:
Hi @ladipro 

Overall this looks good. I'd like to see a couple changes before we merge it.

First, the new article needs to be included in the TOC somewhere. (Otherwise, readers won't find it, and if they land on it from another link, they will be lost without a TOC for reference.)

Second, since the additional NuGet package is only needed for non-SDK style projects, we should add a note about that fact.

ladipro:
@BillWagner thank you for reviewing! We have a call scheduled to discuss some of these changes later this week. I will update the PR once everything is finalized.

ladipro:
@BillWagner, @Thraka, I have made changes as requested.
- Reverted the change to the net-framework-vs.md blurb.
- Clarified that it applies to non-SDK-style projects only.
- Added to TOC and moved to Migration guide.

I have chosen to put it under Migration guide because it doesn't really fit anywhere else and likely doesn't warrant a new top-level link/section. It's going to be linked from SDK Downloads, specifically from unsupported versions, so it kind of fits into the concept of "migration". Looking forward to your feedback!

adegeo:
I would also include a full section on how to do this in framework-style projects, since it's highly probable that customers who run into this issue are indeed using that project style.

docs/framework/migration-guide/reference-assemblies.md
```diff
@@ -0,0 +1,32 @@
...
+```xml
+<ItemGroup>
+  <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" PrivateAssets="All" />
+</ItemGroup>
+```
+
+This can be done with the NuGet Package Manager UI in Visual Studio, with [dotnet add package](../../core/tools/dotnet-add-package.md) on the .NET SDK command line, or by manually editing the project file. The package contains reference assemblies for many versions of .NET Framework. The version to be actually used is determined by the TargetFramework (TargetFrameworks) property.
+
+> [!NOTE]
+> Projects that contain a project reference need to be restored before they can be built. When building with Visual Studio, either in the IDE or on the command line with *MSBuild.exe*, make sure to explicitly execute the *restore* target prior to building the project. This is done on the Developer Command Prompt by executing `msbuild /t:restore` in the project directory.
```
adegeo, line 25: Avoid the work "execute" if you can
```suggestion
> Projects that contain a project reference need to be restored before they can be built. When building with Visual Studio, either in the IDE or on the command line with *MSBuild.exe*, make sure to explicitly run the *restore* target prior to building the project. This is done on the Developer Command Prompt with the `msbuild /t:restore` command from the project directory.
```

adegeo, line 25: Also, running `msbuild /t:restore` is really something someone would run if they were using Framework-style projects and not SDK-style projects. So along with my other suggestion of including instructions for both project styles, I would update this specifically to be `dotnet restore` and then use `msbuild` in the Framework-style section.

adegeo (changes requested):

docs/framework/migration-guide/reference-assemblies.md
ladipro: This guidance is primarily for non-SDK-style ("Framework-style" as you wrote) projects. SDK-style works out of the box because they already have the `PackageReference` included from the .NET SDK.

SDK-style still needs to run restore, but that's something they have to do anyways, so it looks as though including it here would be redundant. Although on second thought, it makes sense that SDK-style devs would land here too, and the step to restore may not be obvious, especially since "Restore NuGet Packages" in VS doesn't cut it.

OK, let me restructure it a bit. Thank you!

ladipro: I ran through the scenarios again and have updated the text.
- The build prefers the machine-wide targeting pack in all cases. So I have removed the "non-SDK-style" part from the first paragraph. The new text is more correct and concise.
- The section about updating project files explains the difference between SDK and non-SDK. It doesn't recommend `dotnet add-package` anymore, because it's not needed for SDK and doesn't work for non-SDK.
- A new section explains how to restore on the command line. The difference here is not between SDK and non-SDK, but between VS and dotnet. While likely equivalent in most cases, I think it's better to recommend `msbuild /t:restore` even for SDK-style projects if the developer is in VS. @zivkan, could you please comment on this? Thank you!

adegeo (changes requested):
A few more questions I have about this.

docs/framework/migration-guide/reference-assemblies.md
```diff
@@ -0,0 +1,32 @@
...
+
+Each project that should build against the reference assemblies NuGet package needs to include the following package reference.
+
+```xml
+<ItemGroup>
+  <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" PrivateAssets="All" />
+</ItemGroup>
+```
+
+SDK-style projects include this reference by default. For traditional, non-SDK-style projects, the reference can be added with the NuGet Package Manager UI in Visual Studio or by manually editing the project file. The package contains reference assemblies for many versions of .NET Framework. The version to be actually used is determined by the TargetFramework (TargetFrameworks) property.
```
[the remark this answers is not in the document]
ladipro: Thank you, the suggested change is awesome. I copied the "traditional" from https://github.com/dotnet/docs/blame/main/docs/csharp/language-reference/preprocessor-directives.md#L116. On second thought I agree it is not appropriate.

Great catch on `TargetFramework`. All add `TargetFrameworkVersion` and qualify with SDK / non-SDK.

docs/framework/migration-guide/reference-assemblies.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Update project files
+
+Each project that should build against the reference assemblies NuGet package needs to include the following package reference.
+
+```xml
+<ItemGroup>
+  <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" PrivateAssets="All" />
+</ItemGroup>
+```
```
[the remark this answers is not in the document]
ladipro: Non-SDK supports package references in project files as well:
https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#project-type-support

I believe that the default in VS is still packages.config for compatibility reasons but can be flipped in options somewhere. Let me check if one project can use both (existing packages.config + `PackageReference`). If not, I guess I'll have to document both ways here.

docs/framework/migration-guide/reference-assemblies.md
zivkan: Some (most? all?) .NET Framework (non-SDK) projects can't be built, or restored, by the `dotnet` CLI. The project files import props/targets that only ship in VS. However, VS and msbuild is able to discover the .NET SDK. So, `msbuild -t:restore` will work for anyone with VS installed, but `dotnet restore` will only work for solutions where every* project in the solution is SDK style. Just one non-SDK style project (like a Visual Studio Extension VSIX project) will cause `dotnet` to fail MSBuild evaluation.

So, while I personally do recommend customers with all SDK style projects in their solution to use the dotnet cli (just so the skills are transferrable to Linux & Mac, no technical reason), it is safe to tell anyone on Windows with VS installed to use msbuild.

*I think that very basic console and classlib projects might work with the dotnet cli, but it's just easier to say something like "non-SDK style project, you must use msbuild".

docs/framework/migration-guide/reference-assemblies.md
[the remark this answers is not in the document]
ladipro: Turns out NuGet doesn't support mix-and-match. The project is either `packages.config`-based or `PackageReference`-based. The former has poor command-line support so I think it's better to recommend only the VS Package Manager to add the reference. Oh, this is so convoluted :)

docs/framework/migration-guide/reference-assemblies.md
ladipro: Thank you. The text now says "When building with VS ... use `nuget restore` or `msbuild /t:restore`" and "When building with .NET SDK ... use `dotnet restore`"). That should be safe / work for everyone, unless I misunderstood.

docs/framework/migration-guide/reference-assemblies.md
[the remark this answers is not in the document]
ladipro: I've added a `packages.config` snippet as you suggested.

docs/framework/migration-guide/reference-assemblies.md
```diff
@@ -0,0 +1,44 @@
...
+
+Projects that contain a package reference need to be restored before they can be built.
+
+When building with Visual Studio, either in the IDE or on the command line with *MSBuild.exe*, make sure to explicitly restore packages prior to building the project. For projects using _packages.config_, this done with [nuget restore](/nuget/reference/cli-reference/cli-ref-restore). For projects using `<PackageReference>`, this is done on the Developer Command Prompt with the `msbuild /t:restore` command from the project directory.
+
+When building with the .NET SDK, make sure to run the [dotnet restore](../../core/tools/dotnet-restore.md) command from the project directory.
+
+## References
+
+- [.NET Framework Targeting Pack Nuget Packages](https://github.com/microsoft/dotnet/tree/main/releases/reference-assemblies)
```
adegeo, line 44: What do you think of this change? This rewrites the section to be explicit in the steps someone needs to take to restore their project, which from your description, seems to be a key to getting it to work.

I also deleted the **## References** section, as we don't really those anymore.

I also changed the .NET SDK text, since .NET Core 2.0, a restore action happens every time the project is run or compiled.

```suggestion

## Restore the project

Projects that contain a package reference must be restored before they can be built.

After adding the **Microsoft.NETFramework.ReferenceAssemblie** NuGet package to your project, you must explicitly run the restore action in one of the following ways:

- If your project is a non SDK-style project and uses the _packages.config_ file to reference NuGet packages:

  01. Install the [NuGet CLI tool](https://learn.microsoft.com/en-us/nuget/install-nuget-client-tools#nugetexe-cli), making sure _nuget.exe_ is in the PATH variable.
  01. Open a command prompt.
  01. Navigate to the directory that contains your project file.
  01. Run _nuget.exe restore_.

- If your project is a non SDK-style project and uses `<PackageReference>` settings in the project file to reference NuGet packages:

  01. Open **Developer Command Prompt for VS 2022**. The name of this app may be different based on which version of Visual Studio you've installed.
  01. Navigate to the directory that contains your project file.
  01. Run _msbuild /t:restore_.

- If your project is an SDK-style project, you don't need to do anything. The NuGet restore action is automatically run when the project is built.
```

docs/framework/migration-guide/reference-assemblies.md
```diff
@@ -0,0 +1,44 @@
+---
+title: "Build apps against Microsoft.NETFramework.ReferenceAssemblies"
+description: "Developers can target .NET Framework by referencing the official reference assemblies NuGet package."
+ms.date: 11/25/2022
```
adegeo, line 4: I altered the description a bit. I think this is more search engine friendly. I also updated the article date 😁 
```suggestion
description: "Learn how to reference the .NET Framework Reference Assemblies in your project, when you can't install a matching developer pack."
ms.date: 01/19/2023
```

adegeo (approved):
This is looking great now! Much more clear. I have two final suggestions for you, you can simply accept them in line if you wish, or decline them.

docs/framework/migration-guide/reference-assemblies.md
ladipro: 👍

docs/framework/migration-guide/reference-assemblies.md
ladipro: Much better, thank you!

(I have edited your suggestion to fix a typo `Microsoft.NETFramework.ReferenceAssemblie` -> `Microsoft.NETFramework.ReferenceAssemblies` before committing.)

Issue #30452, opened 2022-08-02
How to leave feedback on new features?


TobiasKnauss:
The notice at the top of the page says
> These are currently preview features. [...] These features may not all be released in C# 11. Some may remain in a preview phase for longer based on feedback on the feature.

Where can I leave feedback about the features? I guess it is somewhere in Github, but honestly I don't like to waste time searching for the issues in which the respective feature is planned and discussed, if that issue was the right one at all.
Therefore please add the related Github issue for leaving feedback to each feature. Thanks!

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1f94e527-6240-7b69-9d98-da8405790b1c
* Version Independent ID: a3d4d1bb-cfb0-3306-3ce7-5977fb199255
* Content: [What's new in C# 11 - C# Guide](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-11#generic-attributes)
* Content Source: [docs/csharp/whats-new/csharp-11.md](https://github.com/dotnet/docs/blob/main/docs/csharp/whats-new/csharp-11.md)
* Product: **dotnet-csharp**
* Technology: **csharp-whats-new**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

---
[Associated WorkItem - 31002](https://dev.azure.com/msft-skilling/Content/_workitems/edit/31002)

BillWagner:
That's a great suggestion @TobiasKnauss 

I'll update this article to include a link to the planning issue in the dotnet/csharplang repo for each feature.

BillWagner:
/cc @KathleenDollard 

Let's determine where I should be routing people for comment on each C# 11 feature.

PR #32766, opened 2022-11-28
standardize feedback in later releases

BillWagner:
Follow up to #30452

Use an include file to save work in future releases.


IEvangelist (approved):

PR #32783, opened 2022-11-29
Small spelling update

kendaleiv:


BillWagner (approved):
Thanks for spotting and fixing this @kendaleiv 

I'll :shipit: now.

PR #32802, opened 2022-11-30
Address issues from #32672

### LINKED ISSUE #32672: "Options pattern - .NET" is not clear about when updated options are read
[Enter feedback here]

In section "Bind hierarchical configuration", the example code for `ConfigurationBinder.Bind` is immediately followed by the comment, "In the preceding code, changes to the JSON configuration file after the app has started are read."  Since the example code is specifically for _Program.cs_, the program has just started and there has hardly been any time for the JSON configuration file to be changed, so it's not clear what is meant by the comment.

The following example for `ConfigurationBinder.Get<T>` is also followed by the same comment.  In this case, the code is not set specifically in _Program.cs_, but it does start with `configuration.Build()` which presumably reads the configuration file.  Does the comment mean that in order to read changes to the configuration since the app started, one needs to rebuild the configuration?

My understanding is that the JSON file configuration provider will update the built configuration when it detects changes to the file.  If so, then perhaps it's not the call to `configuration.Build()` that needs to be repeated, but only the call to `ConfigurationBinder.Get<T>`.  Or is there some magic in `Bind()` and `Get<T>()` that will maintain a reference to the returned object and update it should the configuration change? Unfortunately, the comment doesn't make any of that clear.

The section then continues with the use of `.Configure<T>` to inject configuration into a DI container, providing both the code to use when configuring the DI container as well as class that takes an `IOptions<TransientFaultHandlingOptions>` in the constructor and assigns its `.Value` to a readonly field.  The example is followed by the comment, "In the preceding code, changes to the JSON configuration file after the app has started are ***not*** read. To read changes after the app has started, use IOptionsSnapshot."  In this case, what is the value of the `ExampleService` class taking an `IOptions<TransientFaultHandlingOptions>` class instead of simply taking `TransientFaultHandlingOptions`?

The documentation should clearly answer the following questions regarding how updated configuration values can be read:
1. What reads updated configuration values?
2. What calls must be repeated to get the updated values?
3. With dependency injection, what interfaces need to be used to get updated values?

---
#### Document Details

* Content: [Options pattern - .NET](https://learn.microsoft.com/en-us/dotnet/core/extensions/options#options-interfaces)
* Content Source: [docs/core/extensions/options.md](https://github.com/dotnet/docs/blob/main/docs/core/extensions/options.md)
* GitHub Login: @IEvangelist
* Microsoft Alias: **dapine**

---
[Associated WorkItem - 39307](https://dev.azure.com/msft-skilling/Content/_workitems/edit/39307)

IEvangelist:
Fixes #32672

### DIFF
--- a/docs/core/extensions/options.md
+++ b/docs/core/extensions/options.md
@@ -3,7 +3,7 @@ title: Options pattern
 author: IEvangelist
 description: Learn how to use the options pattern to represent groups of related settings in .NET apps.
 ms.author: dapine
-ms.date: 05/12/2022
+ms.date: 11/30/2022
 ---
 
 # Options pattern in .NET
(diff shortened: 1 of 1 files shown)


luisquintanilla (approved):
LGTM

PR #32806, opened 2022-11-30
Add alert about keyboard shortcuts

### LINKED ISSUE #30362: Pressing F11 maximises the console window, instead of stepping to the next statement
Hi there,

Just a little documentation issue, that might cause a beginner some confusion.

In the section on "Step through a program", at the point after the user has entered their the keyboard focus will still be on the console window. The immediate next tutorial instruction (step 5) is to press F11 to step to the next program statement, but this is sent to the console window which causes it to be maximised or restored.

There needs to be an instruction to switch the keyboard focus from the console window back to Visual Studio before pressing F11

Regards.

John.

---
[Associated WorkItem - 44861](https://dev.azure.com/msft-skilling/Content/_workitems/edit/44861)

IEvangelist:
Fixes #30362

### DIFF
--- a/docs/core/tutorials/debugging-with-visual-studio.md
+++ b/docs/core/tutorials/debugging-with-visual-studio.md
@@ -1,7 +1,7 @@
 ---
 title: Debug a .NET console application using Visual Studio
 description: Learn how to debug a .NET console app using Visual Studio.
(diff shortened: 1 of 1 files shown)


luisquintanilla (approved):
LGTM

PR #32825, opened 2022-12-01
create what's new for Nov

BillWagner:
There is an issue with the TOC processing (now that we've added a node for the release What's new).

I ran the tool to test my fix, and here's the results.


BillWagner:
closing. My local repo was stale and missed some updates.

PR #32887, opened 2022-12-05
Type discriminator order

### LINKED ISSUE #32789: Add a note that the discriminator type property needs to be first in the json for System.Text.Json polymorphic serialization
[Enter feedback here]
Hi,

Can you please add a note that the discriminator type property must be the first property in the json when deserializing as mentioned in these comments?

https://github.com/dotnet/runtime/issues/63747#issuecomment-1184404967
https://github.com/dotnet/runtime/issues/63747#issuecomment-1158624112

Thanks.
---
#### Document Details

* ID: 179cc49a-a37d-155e-5ecf-deca4f219a5a
* Version Independent ID: 17511810-6572-ad88-6eaf-4f9bded720b7
* Content: [How to serialize properties of derived classes with System.Text.Json](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/polymorphism)
* Content Source: [docs/standard/serialization/system-text-json/polymorphism.md](https://github.com/dotnet/docs/blob/main/docs/standard/serialization/system-text-json/polymorphism.md)
* GitHub Login: @gewarren

gewarren:
Fixes #32789.

### DIFF
--- a/docs/standard/serialization/system-text-json/polymorphism.md
+++ b/docs/standard/serialization/system-text-json/polymorphism.md
@@ -289,7 +289,7 @@ Console.WriteLine(value is WeatherForecastWithCity); // True
 ```
 
 > [!NOTE]
-> The type discriminator must be the first property in the JSON for deserialization to succeed.
+> The type discriminator must be placed at the start of the JSON object, grouped together with other metadata properties like `$id` and `$ref`.
 
 ```vb
 Dim value As WeatherForecastBase = JsonSerializer.Deserialize(json)
--- a/docs/standard/serialization/system-text-json/polymorphism.md
+++ b/docs/standard/serialization/system-text-json/polymorphism.md
@@ -288,6 +288,9 @@ WeatherForecastBase value = JsonSerializer.Deserialize<WeatherForecastBase>(json
 Console.WriteLine(value is WeatherForecastWithCity); // True
 ```
 
+> [!NOTE]
+> The type discriminator must be the first property in the JSON for deserialization to succeed.
+
 ```vb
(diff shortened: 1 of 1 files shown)


docs/standard/serialization/system-text-json/polymorphism.md
```diff
@@ -288,6 +288,9 @@ WeatherForecastBase value = JsonSerializer.Deserialize<WeatherForecastBase>(json
 Console.WriteLine(value is WeatherForecastWithCity); // True
 ```
 
+> [!NOTE]
+> The type discriminator must be the first property in the JSON for deserialization to succeed.
```
eiriktsarpalis, line 292: This is true for any metadata property that STJ emits, it doesn't strictly speaking need to be the first property, however all metadata properties need to be placed before any regular properties.
```suggestion
> The type discriminator must be placed at the start of the JSON object, grouped together with other metadata properties like `$id` and `$ref`.
```

eiriktsarpalis (approved):
Other than one piece of feedback, LGTM otherwise.

QuintinWillison:
It's only this morning that I've come across this page in the documentation and given it a read through...

I find it really surprising and disappointing that the words "at the start of the JSON object" were conceived and approved -- and, furthermore, that this polymorphic feature was ever released in the fragile form it has been! A JSON object is an unordered collection of name/value pairs (see [json.org](https://www.json.org/json-en.html)). This feature is not standards compliant and cannot be guaranteed to survive intermediary processing that uses JSON I/O. I can't imagine how many software systems are going to be badly architected by those trusting the underlying data encodings because this was "rubber stamped" by the official .NET / Microsoft team. 🤯

PR #32936, opened 2022-12-07
Linux install instructions update due to 3.1 EOL

ayakael:
* Removes .NET 3.1 due to EOL
* Updates Alpine Linux

@adegeo


adegeo:
Tables were changed to be simplified via recommendation of the product team. But to their credit, the original tables were designed back when we were documenting more of the unsupported distros too, but now that we just remove the distros when they're no longer supported, it doesn't make sense to have complex tables with ✔️ and ❌

adegeo (changes requested):
Tables were changed to be simplified via recommendation of the product team. But to their credit, the original tables were designed back when we were documenting more of the unsupported distros too, but now that we just remove the distros when they're no longer supported, it doesn't make sense to have complex tables with ✔️ and ❌

So I would revert those changes 😁

svick:
The Linux install instructions pages also include the `includes/versions-not-supported.md` file. I think that should also be updated.

ayakael:
> The Linux install instructions pages also include the `includes/versions-not-supported.md` file. I think that should also be updated.

Many thanks, done :)

ayakael:
> Tables were changed to be simplified via recommendation of the product team. But to their credit, the original tables were designed back when we were documenting more of the unsupported distros too, but now that we just remove the distros when they're no longer supported, it doesn't make sense to have complex tables with heavy_check_mark and x
> 
> So I would revert those changes grin

Reverted those changes, once tests passes I consider this ready.

adegeo (approved):
Thanks!

PR #32937, opened 2022-12-07
remove no-loc directive in metadata

BillWagner:
It turns out the `no-loc` directive isn't allowed in the metadata values.

Provide a better description for this page.

I searched the current repo, and I didn't find any other pages with this issue.


gewarren (approved):
Thanks!

PR #32939, opened 2022-12-07
Remove legacy code suggestions, and instead point to config docs

### LINKED ISSUE #32746: "Running the app" needs to be updated for Orleans 7.0
The page on https://learn.microsoft.com/en-us/dotnet/orleans/deployment/ reference configuration method that doesn't exist, even in 3.x

The samples should just use `UseLocalhostClustering()` instead.

---
[Associated WorkItem - 46076](https://dev.azure.com/msft-skilling/Content/_workitems/edit/46076)

IEvangelist:
## Summary
Remove legacy code suggestions, and instead point to config docs

Fixes #32746 /cc @ReubenBond @bradygaster 

[Preview](https://review.learn.microsoft.com/en-us/dotnet/orleans/deployment/?branch=pr-en-us-32939)

### DIFF
--- a/docs/orleans/deployment/index.md
+++ b/docs/orleans/deployment/index.md
@@ -1,62 +1,22 @@
 ---
 title: Run an Orleans application
 description: Learn how to run an Orleans app in .NET.
-ms.date: 03/09/2022
+ms.date: 12/07/2022
 ---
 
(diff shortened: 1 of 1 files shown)


tdykstra (approved):
LGTM

PR #32963, opened 2022-12-08
Correcting video embed URL

brstegge:
## Summary

The video in this article uses an invalid embed URL and is likely to break. Correcting to use the properly formatted embed URL.


brstegge:
@BillWagner Small fix. Let me know if you have any questions? The correct embed URL can be found on the "Embed" tab at the bottom of the video's episode page: https://learn.microsoft.com/en-us/shows/CSharp-101/What-is-C?tab=tab-embed

IEvangelist (approved):

PR #32997, opened 2022-12-13
Update `ref` performance article(s) for C# 11

### LINKED ISSUE #27175: Avoid defensive copies – outdated example (in C# 10+ there are no _defensive copies_)
In chapter **Avoid defensive copies** there is

> The Point3D structure is not a read-only struct. There are six different property access calls in the body of this method. On first examination, you may think these accesses are safe. After all, a get accessor shouldn't modify the state of the object. But there's no language rule that enforces that. It's only a common convention. Any type could implement a get accessor that modified the internal state.

```csharp
private static double CalculateDistance(in Point3D point1, in Point3D point2)
{
    double xDifference = point1.X - point2.X;
    double yDifference = point1.Y - point2.Y;
    double zDifference = point1.Z - point2.Z;

    return Math.Sqrt(xDifference * xDifference + yDifference * yDifference + zDifference * zDifference);
}
```
Next paragraph starts with note abote `readonly` modifier

> Without some language guarantee, the compiler must create a temporary copy of the argument before calling any member not marked with the readonly modifier. 

Because this example does not show exact implementation of `Point3D` chapter maybe could be enriched with [!NOTE] about C# 8+ autogenerated `readonly` modifiers for autoimplemented-property getters.

---

Excerpts from chapter **Declare readonly members for mutable structs**

> If you use auto-implemented properties, the compiler adds the readonly modifier to the get accessor for read-write properties.

> The compiler adds the readonly modifier to the auto-implemented property declarations for properties with only a get accessor.

---
#### Document Details

* Content: [Write safe and efficient C# code](https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#avoid-defensive-copies)
* Content Source: [docs/csharp/write-safe-efficient-code.md](https://github.com/dotnet/docs/blob/main/docs/csharp/write-safe-efficient-code.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

---
[Associated WorkItem - 54740](https://dev.azure.com/msft-skilling/Content/_workitems/edit/54740)

### LINKED ISSUE #27199: Declare immutable structs as readonly – readonly struct rules lack _init-only_ setter option
> All properties must be read-only, including auto-implemented properties.

C# 9.0 init-only setter option is not mentioned here.

From [readonly struct](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/struct#readonly-struct)
> All data members of a readonly struct must be read-only as follows:
> * Any field declaration must have the readonly modifier
> * Any property, including auto-implemented ones, must be read-only. In C# 9.0 and later, a property may have an init accessor.

---
#### Document Details

* Content: [Write safe and efficient C# code](https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#declare-immutable-structs-as-readonly)
* Content Source: [docs/csharp/write-safe-efficient-code.md](https://github.com/dotnet/docs/blob/main/docs/csharp/write-safe-efficient-code.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

---
[Associated WorkItem - 54747](https://dev.azure.com/msft-skilling/Content/_workitems/edit/54747)

### LINKED ISSUE #27201: Declare readonly members for mutable structs – props with init-only setter getter also get `IsReadOnlyAttribute`
> If you use auto-implemented properties, the compiler adds the readonly modifier to the get accessor for read-write properties. The compiler adds the readonly modifier to the auto-implemented property declarations for properties with only a get accessor.

Properties with  _init-only_ setter are considered readonly/immutable. If they are autoimplemented, compiler generates _readonly modifier_. 

```csharp
 object Obj { get; init; }

.method /* 060000D7 */ private hidebysig specialname instance object get_Obj () cil managed 
{
	.custom instance void [System.Runtime]System.Runtime.CompilerServices.IsReadOnlyAttribute::.ctor() = ( 01 00 00 00 )
	.custom instance void [System.Runtime]System.Runtime.CompilerServices.CompilerGeneratedAttribute::.ctor() = ( 01 00 00 00 )
	// Method begins at RVA 0x35a0
	// Code size 7 (0x7)
	.maxstack 8

…
```

`System.Runtime.CompilerServices.IsReadOnlyAttribute` expresses `readonly` modifier.

Usually by _readonly property_ is meant property without any _setter_. So it is better to explicitly denote that autogenerated _RW-, readonly- and with init-only-setter_- properties get readonly modifier to _getter_.

---
#### Document Details

* Content: [Write safe and efficient C# code](https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#declare-readonly-members-for-mutable-structs)
* Content Source: [docs/csharp/write-safe-efficient-code.md](https://github.com/dotnet/docs/blob/main/docs/csharp/write-safe-efficient-code.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

---
[Associated WorkItem - 54742](https://dev.azure.com/msft-skilling/Content/_workitems/edit/54742)

### LINKED ISSUE #27202: Use ref readonly return statements – duplicity
Sentences

> Notice that the readonly modifier must be part of the declaration of the variable. The reference to which it refers can't be modified. Attempts to do so result in a compile-time error.

are followed by new paragraph

> The readonly modifier is required on the declaration of originReference.

which is in turn followed by next paragraph starting with

> The compiler enforces that the caller can't modify the reference. . Attempts to assign the value directly generate a compile-time error.

It is duplicity actuating disturbance in comprehensive content of article.

I guess removing first quote and merging middle paragraph with next can be fine.

---
#### Document Details

* Content: [Write safe and efficient C# code](https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#use-ref-readonly-return-statements)
* Content Source: [docs/csharp/write-safe-efficient-code.md](https://github.com/dotnet/docs/blob/main/docs/csharp/write-safe-efficient-code.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

---
[Associated WorkItem - 54743](https://dev.azure.com/msft-skilling/Content/_workitems/edit/54743)

### LINKED ISSUE #27203: The out, ref, and in keywords – misworded statement
> Add the in modifier to pass an argument by reference and declare your design intent to pass arguments by reference to avoid unnecessary copying. You don't intend to modify the object used as that argument.

I guess this statement was not about to repeat double _to pass (an) argument by reference_ and then directly address reader. 

It should be something like 

_Add the `in` modifier to pass an argument by reference and declare your design intent to preserve referenced object state._

---
#### Document Details

* Content: [Write safe and efficient C# code](https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#the-out-ref-and-in-keywords)
* Content Source: [docs/csharp/write-safe-efficient-code.md](https://github.com/dotnet/docs/blob/main/docs/csharp/write-safe-efficient-code.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

---
[Associated WorkItem - 54745](https://dev.azure.com/msft-skilling/Content/_workitems/edit/54745)

### LINKED ISSUE #27205: The out, ref, and in keywords – description of state preservation rules is defective
>  First of all, the called method can't directly assign to an in parameter. It can't directly assign to any field of an in parameter when that value is a struct type. In addition, you can't pass an in parameter to any method using the ref or out modifier. These rules apply to any field of an in parameter, provided the field is a struct type and the parameter is also a struct type. In fact, these rules apply for multiple layers of member access provided the types at all levels of member access are structs. The compiler enforces that struct types passed as in arguments and their struct members are read-only variables when used as arguments to other methods.

Statement “_In fact, these rules apply for multiple layers of member access provided the types at all levels of member access are structs._” uses term `member` while it should be `field`. When method (like property or indexer getter) return value, it is copied. So readonly protection (CS8332) is enforced only on `field` access chains.

```csharp
struct Outer { public Inner Inner { get; set; } }
struct Inner { public object Obj { get; set; } }
class Tests
{
  void Test ( in Outer outer )
  {
    Inner inner = outer.Inner;
    inner.Obj = default; // No compiler error – working with Inner copy.

    // Error CS1612 Cannot modify the return value of 'Outer.Inner' because it is not a variable
    outer.Inner.Obj = default;
  }
}


```
Sample show that when _value type_ is returned, it gets copied. Since it is not assigned locally, cannot be assigned to its member (CS1612). 

╍╍ ╍╍╍╍ ╍╍╍╍ ╍╍╍╍ ╍╍╍╍ ╍╍

Since `struct` cannot return `by-ref`, **AFAIK** there is no option how get-access to _struct-type_ member of `struct` could return without copy of that member value. That means that _layer access protection_ applies only to `fields`.

```csharp
struct RefStruct
{
  int _int;
  // Error CS8170 Struct members cannot return 'this' or other instance members by reference
  public ref int Int => ref _int;
}
```
---

Statement “_The compiler enforces that struct types passed as in arguments and their struct members are read-only variables when used as arguments to other methods._” is rather unfathomable. Again it “suddenly” discusses members while `field`s were in question. Let assume fields. If so, this statement tells nothing untold. Otherwise it implies impossible – property, indexer, event, delegate, type, …, cannot be passed

I guess this was about to note _defensive copy_ logic instead of recapitulating deep immutability characteristics of readonly value type reference.

---
---

This description better got reworded. It tries to explain something what is quite simple in a way that is not simple, or at least in not well-arranged way. Imagine that even if `in` modifier causes _pass-by-ref_, type kind is still treated as if was not pointer, i.e. it is `reference` or `value`. Then:

1. For reference is not possible only to assign to `in` param. This satisfies the condition, **reference** (of reference type) at original storage is preserved, i.e. it is perfectly exhibited well-known shallow immutability behavior of readonly reference variables.
2. Value type follows same protection since otherwise original location would start pointing to another value and of course it is not possible to alter any field since this would alter the value – image if some DateTime like value was added 2 days internally. Similarly to 1. shallow immutability applies to reference type fields.

```csharp
struct MyStruct
{
  public object Object;
  public int Int;
}

void Test (in MyStruct myStruct)
{
  // Error CS8332 Cannot assign to a member of variable 'in MyStruct' because it is a readonly variable
  myStruct.Int = default;
  // Error CS8332 Cannot assign to a member of variable 'in MyStruct' because it is a readonly variable
  myStruct.Object = default;
}

class N { public object obj; }
void Test ( in N n )
{
  n.obj = default;      
  n = default; // Error CS8331 Cannot assign to variable 'in Container.N' because it is a readonly variable
}
```
FMPOV it is better to introduce term shallow and deep immutability and rather replace behavior  description with more separated information like table or separated list for `in` param of reference type and `in` param of value type. 

---
#### Document Details

* Content: [Write safe and efficient C# code](https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#the-out-ref-and-in-keywords)
* Content Source: [docs/csharp/write-safe-efficient-code.md](https://github.com/dotnet/docs/blob/main/docs/csharp/write-safe-efficient-code.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

---
[Associated WorkItem - 54741](https://dev.azure.com/msft-skilling/Content/_workitems/edit/54741)

### LINKED ISSUE #27208: Optional use of in at call site – bad formulation
> Omitting the in modifier at the call site informs the compiler that it's allowed to make a copy of the argument for any of the following reasons:
>
> * There exists an implicit conversion but not an identity conversion from the argument type to the parameter type.
> * The argument is an expression but doesn't have a known storage variable.
> * An overload exists that differs by the presence or absence of in. In that case, the by value overload is a better match.
>
>These rules are useful as you update existing code to use read-only reference arguments. Inside the called method, you can call any instance method that uses by-value parameters. In those instances, a copy of the in parameter is created.

Paragraph “_These rules are useful …_” likely was about to link just ability to omit `in` modifier at caller site since rules above are directly linked to caller site characteristic rather than `in` parameter at callee site.

---
#### Document Details

* Content: [Write safe and efficient C# code](https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#optional-use-of-in-at-call-site)
* Content Source: [docs/csharp/write-safe-efficient-code.md](https://github.com/dotnet/docs/blob/main/docs/csharp/write-safe-efficient-code.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

---
[Associated WorkItem - 54746](https://dev.azure.com/msft-skilling/Content/_workitems/edit/54746)

### LINKED ISSUE #32594: "Don't pass a nullable value type as an `in` argument."
> Don't pass a nullable value type as an `in` argument. The Nullable<T> type isn't declared as a read-only struct.  That means the compiler must generate defensive copies for any nullable value type argument passed to a method using the `in` modifier on the parameter declaration.

I think this guidance is a bit too strong.  Roslyn (v3.2 and later) does not make defensive copies when [accessing some members of `Nullable<T>`](https://github.com/dotnet/roslyn/blob/v4.2.0/src/Compilers/CSharp/Portable/CodeGen/EmitExpression.cs#L1731-L1733) even when targeting .NET Framework (where none of its members are `readonly`).

---
#### Document Details

* Content: [Write safe and efficient C# code](https://learn.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#feedback)
* Content Source: [docs/csharp/write-safe-efficient-code.md](https://github.com/dotnet/docs/blob/main/docs/csharp/write-safe-efficient-code.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

---
[Associated WorkItem - 54744](https://dev.azure.com/msft-skilling/Content/_workitems/edit/54744)

### LINKED ISSUE #32668: This needs reworking for C# 11
Since this was first written, we've done quite a bit to extend the use of `ref`.  `ref struct` types can now include `ref` fields. `ref` variables can include the `scoped` modifier. There have been enhancements to analysis for `in` parameters and `readonly` struct variables. 

This should get updated to tell the C# 11 story, with new scenarios and new examples.

---
#### Document Details

* Content: [Write safe and efficient C# code](https://learn.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code)
* Content Source: [docs/csharp/write-safe-efficient-code.md](https://github.com/dotnet/docs/blob/main/docs/csharp/write-safe-efficient-code.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

---
[Associated WorkItem - 39263](https://dev.azure.com/msft-skilling/Content/_workitems/edit/39263)

BillWagner:
This article was originally created in the C# 7 wave. Several improvements in this area have happened since then. We've added many features around the general concept of using `ref` variables and `ref assignment. Rework this article into an overview and a tutorial that cover the feature.

The overview describes the the problem space and the techniques to improve performance using `struct` types and references instead of values.

The tutorial starts with a program that creates numerous unnecessary allocations. Each step demonstrates one of the techniques to remove allocations, preserve the original semantics, and clarify expectations on data modification.

Fixes #32668

Several other issues related to this issue are also fixed in this PR

Fixes #27175 - Add note about `readonly` modifier on auto-implemented properties.
Fixes #27199 - Include init-only properties as well for readonly.
Fixes #27201 - add note about init only properties marked as readonly.
Fixes #27202 - remove duplicate notes about ref assignment and readonly ref assignment.
Fixes #27203 - reword sentence.
Fixes #27205 - reword description, remove duplicate explanation
Fixes #27208 - clarify internal reference.
Fixes #32594 - lower the strength of the specific recommendation.

----

- [Internal preview - overview](https://review.learn.microsoft.com/en-us/dotnet/csharp/advanced-topics/performance/?branch=pr-en-us-32997)
- [Internal preview - tutorial](https://review.learn.microsoft.com/en-us/dotnet/csharp/advanced-topics/performance/ref-tutorial?branch=pr-en-us-32997)


docs/csharp/advanced-topics/performance/ref-tutorial.md
```diff
@@ -0,0 +1,215 @@
...
+
+Then, you modify the `Room` class following the same technique you used for the `Debounce` property. The `Average` property returns a `readonly ref` to the private field for the average measurement. The `AddMeasurement` method modifies the internal fields.
+
+:::code language="csharp" source="./snippets/ref-tutorial/IntruderAlert-finished/Room.cs" id="AverageMeasurement":::
+
+## Avoid boxing
+
+There's one final change to improve performance. The main program is printing stats for the room, including the risk assessment:
+
+:::code language="csharp" source="./snippets/ref-tutorial/IntruderAlert-starter/Program.cs" id="RoomStatus":::
```
CamSoper, line 194: You never tell them to change this to 

```csharp
Console.WriteLine(room.ToString());
````

**Review by CamSoper — CHANGES_REQUESTED** [review]
In the tutorial, my program had unexpected output. Lot's of values were 0 or `NaN`. Totally willing to believe I missed something in the text but I'm not seeing it.

image

On further review (thanks, WinMerge!) it looks like there were changes that I was supposed to make that weren't clear in *Program.cs* and *Room.cs*.

image


image

**Comment by BillWagner** [comment]
Thanks @CamSoper 

I've addressed all the feedback.

**Review by CamSoper — APPROVED** [review]
LGTM

File: docs/csharp/advanced-topics/performance/index.md [hunk 2]
```diff
@@ -0,0 +1,94 @@
...
+
+## Ref safe to escape scope
+
+C# includes rules for `ref` expressions to ensure that a `ref` expression can't be accessed where the storage it refers to is no longer valid. Consider the following example:
+
+```csharp
+public ref int CantEscape()
+{
+    int index = 42;
+    return ref index; // Error: error's ref safe to escape scope is the body of CantEscape
```
cston, line 44: Should the comment refer to _index's_ ref safe to escape instead?

docs/csharp/advanced-topics/performance/index.md
```diff
@@ -0,0 +1,94 @@
...
+}
+```
+
+The compiler reports an error because you can't return a reference to a local variable from a method. The caller can't access the storage being referred to. The *ref safe to escape scope* defines the scope in which a `ref` expression is safe to access or modify. For non-ref locals, the *ref safe to escape* scope is the body of the method. In other words, a non-ref local can't be returned as a `ref` expression. For a variable to have a *ref safe to escape scope* of the entire method, it must be accessible and assignable outside the method. The following snippet shows two examples. A member field can be accessed from the scope calling a method, so a class or struct field's *ref safe to escape scope* is the entire method. The *ref safe to escape scope* for a parameter with the `ref`, `in`, or `out` modifiers is also the entire method. Both can be `ref` returned from a member method:
```
cston, line 48: >Both can be `ref` returned from a member method: [](http://example.com/codeflow?start=844&length=48)

The _ref safe to escape scope_ of a class field is the calling method, so a ref to a class field can be returned from a member method. But the _ref safe to escape scope_ of a non-ref struct field is the current method, so a ref to a struct field cannot be returned.

cston: For parameters, in C#11 the _ref safe to escape scope_ of an `out` parameter is the current method, so an `out` parameter cannot be returned by reference.

docs/csharp/advanced-topics/performance/index.md
```diff
@@ -0,0 +1,94 @@
...
+:::code language="csharp" source="./snippets/ref-safety/EscapeScopes.cs" id="RefSafeToEscapeScopes":::
+
+> [!NOTE]
+> When the `in` modifier is applied to a parameter, that parameter can be returned by `ref readonly`, not `ref`.
+
+The compiler ensures that a reference can't escape its *ref safe to escape scope*. You can use `ref` parameters, `ref return` and `ref` local variables safely because the compiler detects if you've accidentally written code where a `ref` expression could be accessed when its storage isn't valid.
+
+## Safe to escape scope and ref structs
+
+`ref struct` types require more rules to ensure they can be used safely. A `ref struct` type may include `ref` fields. That requires the introduction of a *safe to escape scope*. For most types, the *safe to escape scope* is the entire method. In other words, a value that's not a `ref struct` can always be returned from a method.
```
cston, line 59: >the *safe to escape scope* is the entire method [](http://example.com/codeflow?start=195&length=47)

For types other than ref structs, the safe to escape scope is the _calling method_. (A value with safe to escape scope of _calling method_ is safe to return from the method; a value with safe to escape scope of current method is not.)

docs/csharp/advanced-topics/performance/index.md
```diff
@@ -0,0 +1,94 @@
...
+    {
+        numbers[i] = i;
+    }
+    return numbers; // Error! numbers can't escape this method.
+}
+```
+
+## Unify memory types
+
+The introduction of <xref:System.Span%601?displayProperty=fullName> and <xref:System.Memory%601?displayProperty=fullName> provide a unified model for working with memory. <xref:System.ReadOnlySpan%601?displayProperty=fullName> and <xref:System.ReadOnlyMemory%601?displayProperty=fullName> provide readonly versions for accessing memory. They all provide an abstraction over a block of memory storing an array of similar elements. The difference is that `Span<T>` and `ReadOnlySpan<T>` are `ref struct` types whereas `Memory<T>` and `ReadOnlyMemory<T>` are `struct` types. Spans contain a `ref field`. Therefore instances of a span can't leave its *ref safe to escape* context. The implementation of `Memory<T>` and `ReadOnlyMemory<T>` remove this restriction. You use these types to directly access memory buffers.
```
cston, line 82: >Therefore instances of a span can't leave its *ref safe to escape* context. [](http://example.com/codeflow?start=601&length=75)

Instances of a span can't leave its _safe to escape_ context (which is the _ref safe to escape_ of the ref field within the span).

docs/csharp/advanced-topics/performance/index.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Improve performance with ref safety
+
+Using these features to improve performance involves these tasks:
+
+- *Avoid allocations*:  When you change a type from a `class` to a `struct`, you change how it's stored. Local variables are stored on the stack. Members are stored inline when the container object is allocated. This change means fewer allocations and that decreases the work the garbage collector does. It may also decrease memory pressure so the garbage collector runs less often.
+- *Preserve reference semantics*: Changing a type from a `class` to a `struct` changes the semantics of passing a variable to a method. Code that modified the state of its parameters needs modification. Now that the parameter is a `struct`, the method is modifying a copy of the original object. You can restore the original semantics by passing that parameter as a `ref` parameter. After that change, the method modifies the original `struct` again.
+- *Avoid copying data*: Copying larger `struct` types can impact performance in some code paths. You can also add the `ref` modifier to pass larger data structures to methods by reference instead of by value.
```
cston, line 90: >You can also add the `ref` modifier [](http://example.com/codeflow?start=97&length=35)

You can also add the `ref` _or `in`_ modifier ...

docs/csharp/advanced-topics/performance/index.md
BillWagner: I restructured this paragraph as a table that states the *ref safe to escape* scopes for all different declaration types, based on these notes. I think I got it correct, but it's worth another look.

docs/csharp/advanced-topics/performance/index.md
BillWagner: I took this comment verbatim. I think that's fine here.

I'll see if I get questions on how we define the *safe to escape* scope for a `ref struct` with multiple `ref` fields. Intuitively, it would be the "smallest" scope of the multiple *ref safe to escape* scopes. Do we already have formal language for that?

BillWagner (commented):
Thanks a lot @cston This is really helpful.

I addressed all the feedback. One set of changes was to rewrite a paragraph. Can you give that change in the final commit another look?

Thanks again.

docs/csharp/advanced-topics/performance/index.md
```diff
@@ -0,0 +1,106 @@
...
+|-----------------------------|-------------------------------|
+| non-ref local               | block where local is declared |
+| non-ref parameter           | current method                |
+| `ref`, `in` parameter       | calling method                |
+| `out` parameter             | current method                |
+| `class` field               | calling method                |
+| non-ref `struct` field      | current method                |
+| `ref` field of `ref struct` | calling method                |
+
+A variable can be `ref` returned if its *ref safe to escape scope* is the calling method. If its *ref safe to escape scope* is the current method or a block, `ref` return is disallowed. The following snippet shows two examples. A member field can be accessed from the scope calling a method, so a class or struct field's *ref safe to escape scope* is the entire method. The *ref safe to escape scope* for a parameter with the `ref`, or `in` modifiers is the entire method. Both can be `ref` returned from a member method:
```
cston, line 60: >is the entire method [](http://example.com/codeflow?start=451&length=20)

The table above looks correct. The comment here that the ref safe to escape scope for a `ref` or `in` parameter is the entire method might be confusing though since the table says _calling method_ and the sentence following this one says the `ref` can be returned from the method. Consider using _calling method_ here instead of _entire method_ to reduce the chance of confusion.

PR #33006, opened 2022-12-14
Adding mention that ctor types and prop must match

### LINKED ISSUE #32637: System.Text.Json - Immutable types and Records - Docs don't mention that ctor types and prop types must match
Documentation mentions

> "The parameter names of a parameterized constructor must match the property names. Matching is case-insensitive...

But it has a caveat that had me runing around, it should say "must match the property names **and types**"

The following minimal reproducible example will fail, but will work fine if MyProperty is changed to an int and .ToString() is removed

```c#
internal class Program
{
    private static void Main(string[] args)
    {
        var json = $$"""{"MyProperty": 61}""" ;
        var value = JsonSerializer.Deserialize<MyClass>(json);
    }

    public class MyClass
    {
        public MyClass(int myProperty)
        {
            this.MyProperty = myProperty.ToString();
        }
        
        public string MyProperty { get; }
    }
}
```

---
#### Document Details

* ID: 31f52750-1f3a-3cba-9790-bf36a5ae30a5
* Version Independent ID: cf9fe1d5-1079-025c-73cb-668cc9cbd520
* Content: [How to use immutable types and non-public accessors with System.Text.Json](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/immutability)
* Content Source: [docs/standard/serialization/system-text-json/immutability.md](https://github.com/dotnet/docs/blob/main/docs/standard/serialization/system-text-json/immutability.md)
* GitHub Login: @gewarren

Viniciusap:
## Summary

Adding mention that ctor types and prop types must match

Fixes #32637

### DIFF
--- a/docs/standard/serialization/system-text-json/immutability.md
+++ b/docs/standard/serialization/system-text-json/immutability.md
@@ -27,7 +27,7 @@ This article shows how to use immutable types, public parameterized constructors
 :::code language="csharp" source="snippets/system-text-json-how-to-5-0/csharp/ImmutableTypes.cs" highlight="12":::
 :::code language="vb" source="snippets/system-text-json-how-to-5-0/vb/ImmutableTypes.vb" :::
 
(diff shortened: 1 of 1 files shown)


Viniciusap:
@gewarren I'd appreciate it if you could please take a look at this simple PR. Thank you in advance!

IEvangelist (approved):

PR #33007, opened 2022-12-14
Giving a better resolution for CS1591

Viniciusap:
## Summary

Giving a better resolution for CS1591.

Fixes ##32339


BillWagner (approved):
Thanks for making this suggestion @Viniciusap 

I'll :shipit: now.

PR #24630, opened 2021-06-10
Migration docs revamp.

StephenBonikowsky:
Related to #23561

Phase 1 is supposed to be the smallest possible set of changes to get the existing documentation mapped to the new [TOC design](https://github.com/dotnet/docs/issues/23215).

The modifications to docs/fundamentals/toc.yml reflect this minimal set of changes.

These changes and additional content related changes are tracked with the [Migration guide project](https://github.com/dotnet/docs/projects/128).

Next steps are to make the content related changes as listed in the 'Phase 1' column of the project and start getting reviews and feedback.

[Existing Migration Docs](https://docs.microsoft.com/en-us/dotnet/core/porting/)
[Preview Migration Docs](https://review.docs.microsoft.com/en-us/dotnet/core/porting/?branch=pr-en-us-24630)

PR #33111, opened 2022-12-20
Reorganize breaking changes TOC

gewarren:
- Move ".NET Framework to .NET (Core)" breaking change articles from "Compatibility" table of contents (TOC) to "Migration guide" in main .NET TOC, since the Migration guide is all about migrating from .NET Framework to .NET (Core) - [preview link](https://review.learn.microsoft.com/en-us/dotnet/core/porting/breaking-changes?branch=pr-en-us-33111)
- Re-add link to compatibility TOC to main .NET TOC under "Versioning" - [preview link](https://review.learn.microsoft.com/en-us/dotnet/core/versions/?branch=pr-en-us-33111). (The compatibility TOC [was (probably unintentionally) orphaned in July 2021](https://github.com/dotnet/docs/pull/24630/files#r1054837339) as part of the migration guide work.
- Move compatibility definitions and change rules to bottom of compatibility TOC under "Resources" (and expand that node by default) - [preview link](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/categories?branch=pr-en-us-33111)

### DIFF
--- a/docs/csharp/whats-new/version-update-considerations.md
+++ b/docs/csharp/whats-new/version-update-considerations.md
@@ -9,15 +9,15 @@ ms.date: 09/19/2018
 
 Compatibility is a very important goal as new features are added to the C# language. In almost all cases, existing code can be recompiled with a new compiler version without any issue.
 
-More care may be required when you adopt new language features in a library. You may be creating a new library with features found in the latest version and need to ensure apps built using previous versions of the compiler can use it. Or you may be upgrading an existing library and many of your users may not have upgraded versions yet. As you make decisions on adopting new features, you'll need to consider two variations of compatibility: source-compatible and binary-compatible.
+More care may be required when you adopt new language features in a library. You may be creating a new library with features found in the latest version and need to ensure apps built using previous versions of the compiler can use it. Or you may be upgrading an existing library and many of your users may not have upgraded versions yet. As you make decisions on adopting new features, you'll need to consider two variations of compatibility: source compatibility and binary compatibility.
 
 ## Binary-compatible changes
 
-Changes to your library are **binary-compatible** when your updated library can be used without rebuilding applications and libraries that use it. Dependent assemblies are not required to be rebuilt, nor are any source code changes required.
+Changes to your library are **binary-compatible** when your updated library can be used without rebuilding applications and libraries that use it. Dependent assemblies don't need to be rebuilt, and no source code changes are required.
 
 ## Source-compatible changes
 
-Changes to your library are **source-compatible** when applications and libraries that use your library do not require source code changes, but the source must be recompiled against the new version to work correctly.
+Changes to your library are **source-compatible** when applications and libraries that use your library don't require source code changes when they're recompiled against the new version.
 
 ## Incompatible changes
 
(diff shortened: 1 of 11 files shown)


docs/csharp/whats-new/version-update-considerations.md
```diff
@@ -9,15 +9,15 @@ ms.date: 09/19/2018
...
 
 ## Binary-compatible changes
 
-Changes to your library are **binary-compatible** when your updated library can be used without rebuilding applications and libraries that use it. Dependent assemblies are not required to be rebuilt, nor are any source code changes required.
+Changes to your library are **binary-compatible** when your updated library can be used without rebuilding applications and libraries that use it. Dependent assemblies don't need to be rebuilt, and no source code changes are required.
 
 ## Source-compatible changes
 
-Changes to your library are **source-compatible** when applications and libraries that use your library do not require source code changes, but the source must be recompiled against the new version to work correctly.
```
gewarren, line 20: @BillWagner This definition seemed incorrect to me - a source-incompatible change is one that appears *if* you recompile. A binary-incompatible change is where the source must be recompiled.

BillWagner: Maybe we need to update the wording, but the current language is correct for the C# language definition of "source breaking" and "binary breaking" change.  If a change is "source compatible", code that compiled in version N still compiles in version N + 1. The example on line 47 shows an example. You must recompile. If you don't, you're app won't run using the new libraries. It's a binary break. But, if you recompile, your code adapts to the library change without any edits to the source.

BillWagner (approved):
This LGTM @gewarren 

We should resolve the one conversation so we use consistent language for all types of breaking changes.

docs/csharp/whats-new/version-update-considerations.md
BillWagner: @gewarren 

I see your point. This article probably needs a fair amount of rework now. I first wrote this to address specific feedback where customers were choosing a newer language version while targeting an older TFM (typically .NET Standard 2.0). That's not a supported scenario.

Today, customers choose a language version / TFM as a pair. That should change the focus to be more about asking what changes a developer can make that are:

- compatible API changes
- binary breaking, but source compatible API changes
- source breaking API changes

I haven't thought it through, but my first impression is that a source breaking, but binary compatible API change can't exist. 

Those changes feel like scope creep for this PR. Do you want to make the edits, or open an issue and I'll address it after break?

gewarren: > I haven't thought it through, but my first impression is that a source breaking, but binary compatible API change can't exist.

You could have a new obsoletion as an error (instead of a warning) - that would be source incompatible but binary compatible.

I'll open a separate issue for this article then. Thanks Bill!

docs/fundamentals/toc.yml
```diff
@@ -1791,6 +1791,8 @@ items:
       href: ../core/versions/index.md
     - name: .NET version selection
       href: ../core/versions/selection.md
+    - name: Compatibility
```
gewarren, line 1794: I also just realized that the entire compatibility TOC is orphaned as of July 2021 (see https://github.com/dotnet/docs/pull/24630/files#r1054837339). I will figure out a place to slot it back into this TOC.

PR #33113, opened 2022-12-20
Updated SampleClass example.

MatsM16:
I belive a generic class is followed by a single backtid and then the count of generic parameters.

## Summary

Changed from a double backtick to a single backtick.

### DIFF
--- a/docs/csharp/language-reference/xmldoc/index.md
+++ b/docs/csharp/language-reference/xmldoc/index.md
@@ -143,7 +143,7 @@ The compiler observes the following rules when it generates the ID strings:
 - For conversion operators only (`op_Implicit` and `op_Explicit`), the return value of the method is encoded as a `~` followed by the return type. For example:
      `<member name="M:System.Decimal.op_Explicit(System.Decimal arg)~System.Int32">` is the tag for the cast operator `public static explicit operator int (decimal value);` declared in the `System.Decimal` class.
 - For generic types, the name of the type is followed by a backtick and then a number that indicates the number of generic type parameters. For example:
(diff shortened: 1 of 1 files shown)


docs/csharp/language-reference/xmldoc/index.md
```diff
@@ -143,7 +143,7 @@ The compiler observes the following rules when it generates the ID strings:
 - For conversion operators only (`op_Implicit` and `op_Explicit`), the return value of the method is encoded as a `~` followed by the return type. For example:
      `<member name="M:System.Decimal.op_Explicit(System.Decimal arg)~System.Int32">` is the tag for the cast operator `public static explicit operator int (decimal value);` declared in the `System.Decimal` class.
 - For generic types, the name of the type is followed by a backtick and then a number that indicates the number of generic type parameters. For example:
-     `<member name="T:SampleClass``2">` is the tag for a type that is defined as `public class SampleClass<T, U>`.
+     `<member name="T:SampleClass`2">` is the tag for a type that is defined as `public class SampleClass<T, U>`.
```
BillWagner, line 146: We need to escape the single embedded backtick:  ``<member name="T:SampleClass`2">``  (Note that the inline code is marked with double-backticks.

```suggestion
     ``<member name="T:SampleClass`2">`` is the tag for a type that is defined as `public class SampleClass<T, U>`.
```

BillWagner (commented):
Thanks for making this fix @MatsM16 

The original is incorrect. However, to get the result to render correctly, we need to enclose that inline code example in double-backticks.

BillWagner (approved):
Thanks again @MatsM16 

I'll :shipit: now.

PR #33120, opened 2022-12-20
Duplicate line removed

surfmuggle:
The current line 61 `Console.WriteLine(string.Join(" ", numbers));  // output: 5 0` was repeated on line 64. I removed 64. 

The line 63 was `numbers.Add(a ??= 0);` i changed it to `numbers.Add(a ??= 3);` because in my opinion an int > 0 conveys clearly that it is not `null`.  On a side note in dutch / flemish and german zero `0` is called  `Nul` or `Null` which is pretty close to `null`

## Summary

Duplicate line removed, integer changed from 0 to 3

### DIFF
--- a/docs/csharp/language-reference/operators/snippets/shared/NullCoalescingOperator.cs
+++ b/docs/csharp/language-reference/operators/snippets/shared/NullCoalescingOperator.cs
@@ -58,7 +58,7 @@ public static class NullCoalescingOperator
         int? a = null;
 
         Console.WriteLine((numbers is null)); // expected: true
-        // if numbers is null add 5 to numbers
+        // if numbers is null, initialize it. Then, add 5 to numbers
         (numbers ??= new List<int>()).Add(5);
         Console.WriteLine(string.Join(" ", numbers));  // output: 5
         Console.WriteLine((numbers is null)); // expected: false        
--- a/docs/csharp/language-reference/operators/snippets/shared/NullCoalescingOperator.cs
+++ b/docs/csharp/language-reference/operators/snippets/shared/NullCoalescingOperator.cs
@@ -60,9 +60,8 @@ public static class NullCoalescingOperator
         (numbers ??= new List<int>()).Add(5);
         Console.WriteLine(string.Join(" ", numbers));  // output: 5
 
-        numbers.Add(a ??= 0);
-        Console.WriteLine(string.Join(" ", numbers));  // output: 5 0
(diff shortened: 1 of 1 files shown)


BillWagner:
Hi @surfmuggle 

I don't want to take this change as-is, but let's discuss what could make this example more clear.

First, while the line of code on L64 is duplicated from L61, the output is different. The `Add` method changed the contents of the collection.

Second, the value of `0` was chosen deliberately. Using 0 shows that 0 *is not null* for an `int?`. Maybe a good change would be to add a third operation: assign `a` to `0`,  then add `a ?? 3`. Thoughts?

surfmuggle:
How about this?

```cs
List<int> numbers = null;
int? a = null;
	
Console.WriteLine((numbers is null)); // expected: true
// if numbers is null add 5 to numbers
(numbers ??= new List<int>()).Add(5);
Console.WriteLine((numbers is null)); // expected: false
	
Console.WriteLine(string.Join(" ", numbers));  // output: 5

// if a is null then assign 0 to a and add a to the list
Console.WriteLine((a is null)); // expected: true
numbers.Add(a ??= 0);
Console.WriteLine((a is null)); // expected: false
Console.WriteLine(string.Join(" ", numbers));  // output: 5 0
Console.WriteLine(a);  // output: 0
```

surfmuggle:
Regarding your suggestion

> add a third operation: assign a to 0, then add a ?? 3

I really like the idea. 

How about this?

```cs
Console.WriteLine((a is null)); // expected: true
Console.WriteLine((a ?? 3)); // expected: 3 since a is null
```

docs/csharp/language-reference/operators/snippets/shared/NullCoalescingOperator.cs
```diff
@@ -57,12 +57,20 @@ private static void NullCoalescingAssignment()
         List<int> numbers = null;
         int? a = null;
 
+        Console.WriteLine((numbers is null)); // expected: true
+        // if numbers is null add 5 to numbers
```
BillWagner, line 61: Let's change this comment to be more explicit about the action:

```suggestion
        // if numbers is null, initialize it. Then, add 5 to numbers
```

BillWagner (approved):
Thanks @surfmuggle 

I have one small addition that I'll make and then merge this.

Thanks for your patience. I had just gone on holiday when you'd made this update.

PR #33174, opened 2022-12-22
Fix naming rules in dynamic2.cs

lindexi:
## Summary

Fix the name rule.


BillWagner (approved):
Thanks for making these updates @lindexi 

I'll :shipit: now.

lindexi:
Thank you @BillWagner

PR #33191, opened 2022-12-23
Fix Orleans samples links

poulinjulien:
* Updated the image source of the TLS screenshot to point to the correct URL
* Fix the link to the Stocks sample app


IEvangelist (approved):

PR #33259, opened 2022-12-28
Typo fix - update to, method parameters - C# reference, article

mahab339:
## Summary

Change (The *ref_safe_to_escape* scope defines when a variable can *ref* assigned or *ref* reassigned.) to (... can be *ref* assigned or...).


IEvangelist (approved):

PR #33267, opened 2022-12-29
Edit Inheritance.cs

zipperer:
in comment, replace 'originalItemId' -> 'originalItemID'


adegeo (approved):
Thank you for spotting and fixing that!

PR #33308, opened 2023-01-04
Minor rephrasing to improve readability

S1GGEN:
## Summary

Rephrase from increases run-time performance_ to _improves run-time performance_. When skimming through this text, one might parse that this change will increase the run-time, when it in fact will decrease it. This change decreases ambiguity when skimming through the text.

image


IEvangelist (approved):

PR #33326, opened 2023-01-04
Reinstate dispose pattern guidance

gewarren:
The design guidelines aren't supposed to be altered as they're copyrighted. Reverts #9572.


BillWagner (approved):
Thanks @gewarren 

This LGTM. Let's :shipit:

PR #33328, opened 2023-01-05
Change misleading text about .NET Framework compatibility mode

reflectronic:
.NET Framework compatibility mode is supported on all .NET platforms, not just Windows.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [docs/core/porting/index.md](https://github.com/dotnet/docs/blob/b2c336bc864bed32d4670508fe61b24c23542f63/docs/core/porting/index.md) | [docs/core/porting/index](https://review.learn.microsoft.com/en-us/dotnet/core/porting/index?branch=pr-en-us-33328) |


richlander:
Thanks for the contribution. I don't think it is quite right. Here's my understanding of .NET Framework compatibility mode:

- Works on all OSes, in general.
- Some types are Windows-specific (like Registry APIs) -- AKA not cross-platform.
- .NET APIs work if they exist in the .NET Core version that the app is using.
- The compiler gives you some but limited feedback (types in method signatures are checked, but not those used only in the method body)
- Has no relationship to .NET Standard.

Is that right @terrajobst?

reflectronic:
The prose I copied in is straight from our other docs (though, maybe I agree that it could be tweaked a little bit):

https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-1-0#net-framework-compatibility-mode:
> Starting with .NET Standard 2.0, the .NET Framework compatibility mode was introduced. This compatibility mode allows .NET Standard projects to reference .NET Framework libraries as if they were compiled for .NET Standard.

Maybe we should tweak both of these places; the behavior is really more like "as if they were compiled for the version of .NET used at runtime."

In any case, the text as it is currently isn't right either.

terrajobst:
I agree with everything @richlander wrote except for the last part. For better or worse, we branded the .NET Framework Compatibility mode as something that comes from .NET Standard. The way to think about this is that any .NET runtime that implements .NET Standard 2.0 also supports the .NET Framework Compatibility mode, but some runtimes may support a larger chunk of .NET Framework than others.

I'm not sure why the text has any reference to Windows; the .NET Framework compatibility mode is generally a best effort kind of compatibility bridge: it works when it works and it doesn't when it doesn't. My proposal on wording would be:

> The .NET Framework compatibility mode was introduced in .NET Standard 2.0. The compatibility mode allows .NET Standard and (and all .NET runtimes implementing it, including .NET Core) projects to reference .NET Framework libraries as if they were compiled for it. Referencing .NET Framework libraries doesn't work for all projects, such as if the library uses WPF APIs, but it does unblock many porting scenarios. For more information, see the [Analyze your dependencies to port code from .NET Framework to .NET](third-party-deps.md#net-framework-compatibility-mode).

richlander:
> as if they were compiled for it.

I may be wrong/missing some context. It is my understanding that there are APIs that work for the compat mode that are not in .NET Standard.

richlander:
That text looks good to me. It's a more complicated description but I think much more accurate and descriptive.

docs/core/porting/index.md
```diff
@@ -42,7 +42,7 @@ For more information, see [Use the Windows Compatibility Pack to port code to .N
 
 ## .NET Framework compatibility mode
 
-The .NET Framework compatibility mode was introduced in .NET Standard 2.0. The compatibility mode allows .NET Standard and .NET (including .NET Core 3.1) projects to reference .NET Framework libraries on Windows only. Referencing .NET Framework libraries doesn't work for all projects, such as if the library uses WPF APIs, but it does unblock many porting scenarios. For more information, see the [Analyze your dependencies to port code from .NET Framework to .NET](third-party-deps.md#net-framework-compatibility-mode).
+The .NET Framework compatibility mode was introduced in .NET Standard 2.0. The compatibility mode allows .NET Standard and .NET (including .NET Core) projects to reference .NET Framework libraries as if they were compiled for .NET Standard. Referencing .NET Framework libraries doesn't work for all projects, such as if the library uses WPF APIs, but it does unblock many porting scenarios. For more information, see the [Analyze your dependencies to port code from .NET Framework to .NET](third-party-deps.md#net-framework-compatibility-mode).
```
terrajobst, line 45: Based on the conversation I propose this:

```suggestion
The .NET Framework compatibility mode was introduced in .NET Standard 2.0. The compatibility mode allows projects to reference .NET Framework libraries as if they were compiled for the project's target framework. The .NET Framework compatibility mode is not limited to .NET Standard 2.0 but extends to any .NET runtime that implements .NET Standard 2.0, such as .NET Core. However, some .NET implementations may support a larger chunk of .NET Framework than others. For example, .NET Core 3.0 extends the .NET Framework compatibility mode to Windows Forms and WPF.

Referencing .NET Framework libraries doesn't work in all cases, as it depends on which .NET Framework APIs were used and whether or not these APIs are supported by the project's target framework. Also, some of the .NET Framework APIs will only work on Windows. The .NET Framework compatibility mode unblocks many porting scenarios but you should test your projects to ensure that they also work at runtime. For more information, see the [Analyze your dependencies to port code from .NET Framework to](third-party-deps.md#net-framework-compatibility-mode).
```

reflectronic: The only issue with this, I think, is that the the article "for more information" now has less information than this one. English is hard...

Let's transplant the text from that page onto this one:
```suggestion
Starting with .NET Standard 2.0, the .NET Framework compatibility mode was introduced. This compatibility mode allows .NET Standard and .NET Core projects to reference .NET Framework libraries. Referencing .NET Framework libraries doesn't work for all projects, such as if the library uses Windows Presentation Foundation (WPF) APIs, but it does unblock many porting scenarios. For more information, see the [Analyze your dependencies to port code from .NET Framework to .NET](third-party-deps.md#net-framework-compatibility-mode).
```

And we can move this new text that you've proposed to the "more information" article, instead. The we're editing here seems like a "light on specifics, lets you easily skim the problems you might have" type of thing, while the linked one seems to go into much more detail about the nuances of porting.

Sound like a plan?

docs/core/porting/index.md
```diff
@@ -64,6 +66,7 @@ An SDK-style project uses a different property to identify the target framework,
 
 For a list of all target monikers, see [Target frameworks in SDK-style projects](../../standard/frameworks.md#supported-target-frameworks).
 
+
```
BillWagner, line 69:
```suggestion
```

BillWagner (approved):

PR #33356, opened 2023-01-06
Update features.md

rogeriomc:
## Summary

In the topic "String Interpolation", the format string "MM-DD-YYYY" for the property weatherData.Date is wrong. The correct must be "MM-dd-yyyy".


BillWagner (approved):
Thanks again @rogeriomc 

I'll approve this and :shipit: now.

PR #33357, opened 2023-01-06
Update Features.cs

rogeriomc:
## Summary

In the topic "String Interpolation", the format string "MM-DD-YYYY" for the property weatherData.Date is wrong. The correct must be "MM-dd-yyyy".


BillWagner (approved):
Thanks for spotting these and making the fix. I'll merge both PRs now.

PR #33360, opened 2023-01-06
Update ClassesObjects.cs

rogeriomc:
## Summary

My suggestion is to modify the indexer `set` accessor code of the `MyList<T>` class to only assign a value to an item at a specific index and call the `OnChange()` virtual method if the value of the item has really been changed. This modification will make the code more confident with the event name.


BillWagner (approved):
Thanks again @rogeriomc 

I'll :shipit: now.

PR #33361, opened 2023-01-06
Update features.md

rogeriomc:
## Summary

The current version of the previous example doesn't have a static method called `Square`, but instead it has a lambda expression to create an anonymous function to calculate the square of a value: `(x) => x * x`.


BillWagner (approved):
Thanks again @rogeriomc 

This is better wording.

I'll :shipit: now.

PR #33365, opened 2023-01-06
Document that PDBs are produced by default in Release mode too

### LINKED ISSUE #66103: DebugType does not default to "none" for Release configuration
When building an application or library in Release mode, a .pdb file is generated with its absolute path embedded into the output assembly. This is absolutely not ideal, and borders on a major security flaw, as it exposes some potentially sensitive information such as filesystem structure and - depending on the build path - developer's real names. I've written [a blog post](https://blog.oliverbooth.dev/2022/12/22/a-hidden-csharp-compiler-setting-that-might-kind-of-dox-you/) which explains this behaviour in some more detail, but suffice to say that even in cases where no PDB file is published to the public (and rightfully so), the path to that file is still in the assembly, and in one case I was able to discover a developer's real first name because the build was published to their Desktop folder.

An example of how this appears in dotPeek when examining the assembly for Stardew Valley, a game made with XNA:
image

This seems to massively contradict [what the documentation states](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/code-generation#debugtype) about this compiler option.
> The **DebugType** option causes the compiler to generate debugging information and place it in the output file or files. Debugging information is added by default for the _Debug_ build configuration. ___It is off by default for the _Release_ build configuration.___

As far as I can tell, this is flat out false. I believe this is a huge oversight, possibly a bug.

**Steps to Reproduce**:

1. Create any type of application
2. Build it under Release configuration (or even dotnet publish it)
3. Examine dll metadata in application such as dotPeek or ILSpy

**Expected Behavior**:
Debug symbols should not be emitted in Release configuration. More importantly, the absolute path to the .pdb file should not be embedded into the assembly at all.

**Actual Behavior**:
It does do that. Developer's filesystem structure is publicly exposed in a huge number of cases.

jcouv:
## Summary

Fixes https://github.com/dotnet/roslyn/issues/66103
See https://github.com/dotnet/roslyn/issues/66103#issuecomment-1374199318 for conclusion of our discussion. 
FYI @BillWagner

### DIFF
--- a/docs/csharp/language-reference/compiler-options/code-generation.md
+++ b/docs/csharp/language-reference/compiler-options/code-generation.md
@@ -21,7 +21,7 @@ The following options control code generation by the compiler. The new MSBuild s
 
 ## DebugType
 
-The **DebugType** option causes the compiler to generate debugging information and place it in the output file or files. Debugging information is added by default for the *Debug* build configuration. It is off by default for the *Release* build configuration.
+The **DebugType** option causes the compiler to generate debugging information and place it in the output file or files. Debugging information is added by default.
 
 ```xml
 <DebugType>pdbonly</DebugType>


tmat (approved):

BillWagner (approved):
Thanks @jcouv 

I'll :shipit: now.

PR #33372, opened 2023-01-07
Delete Vista and Server 2008 prereqs

jkotas:
These OSes are not supported by .NET Core for quite a while.


IEvangelist (approved):

PR #33379, opened 2023-01-09
CreateObjectReference is not awaitable

bjorkstromm:
## Summary

CreateObjectReference is not awaitable. See docs:
https://learn.microsoft.com/en-us/dotnet/api/orleans.igrainfactory.createobjectreference?view=orleans-7.0


IEvangelist (approved):

PR #33381, opened 2023-01-09
docs: Consistent ClientBase type

mycomycul:
ClientBase requires a type
Set all CliantBase.CacheSettings to ClientBase<ITest>.CacheSettings


IEvangelist (approved):

PR #33395, opened 2023-01-10
Update Observers docs

bjorkstromm:
## Summary

- ObserverManager now part of Orleans.
- Clarify that IGrainObserver also can return Task

### DIFF
--- a/docs/orleans/grains/observers.md
+++ b/docs/orleans/grains/observers.md
@@ -9,9 +9,9 @@ ms.date: 03/16/2022
 There are situations in which a simple message/response pattern is not enough, and the client needs to receive asynchronous notifications.
 For example, a user might want to be notified when a new instant message has been published by a friend.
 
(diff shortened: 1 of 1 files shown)


docs/orleans/grains/observers.md
```diff
@@ -47,7 +47,10 @@ public class Chat : IChat
 }
 ```
 
-On the server, we should next have a Grain that sends these chat messages to clients. The Grain should also have a mechanism for clients to subscribe and unsubscribe themselves for notifications. For subscriptions, the grain can use a copy of the utility class [`ObserverManager<T>`](https://github.com/dotnet/orleans/blob/e997335d2d689bb39e67f6bcf6fd70862a22c02f/test/Grains/TestGrains/ObserverManager.cs#L12).
+On the server, we should next have a Grain that sends these chat messages to clients. The Grain should also have a mechanism for clients to subscribe and unsubscribe themselves for notifications. For subscriptions, the grain can use an instance of the utility class <xref:Orleans.Utilities.ObserverManager%601>.
+
```
IEvangelist, line 51:
```suggestion
On the server, we should next have a Grain that sends these chat messages to clients. The Grain should also have a mechanism for clients to subscribe and unsubscribe themselves for notifications. For subscriptions, the grain can use an instance of the utility class <xref:Orleans.Utilities.ObserverManager%601>.

```

IEvangelist (approved):

PR #33420, opened 2023-01-11
Change IGrainIdentity to GrainId.

NzKyle:
## Summary

In the example ``DataService`` type - the constructor is using ``IGrainIdentity`` when it should be using ``GrainId`` - this small changes updates that - so that others don't get confused like it did :)


IEvangelist:
Hi @NzKyle - thanks for the PR, but this is a slightly larger issue, and this PR would need to be updated to use the appropriate zones for the entire code block. I'll update this in a separate PR and close this when that one is merged. Let's treat this more like an issue, rather than a PR.

PR #33426, opened 2023-01-11
Make BinaryFormatter examples more secure

ladipro:
## Summary

Per https://liquid.microsoft.com/Web/Object/Read/ms.security/Requirements/Microsoft.Security.SystemsADM.10010#Zguide, `SerializationBinder` subclasses must not return null from their `BindToType` and `BindToName` methods. The PR makes the code snippets compliant with this SDL requirement. The new code is functionally equivalent with the old code.


gewarren (approved):
Thanks @ladipro

PR #33517, opened 2023-01-13
Automatic RID for publish command only

### LINKED ISSUE #33051: [Breaking change]: .NET 7.0.2xx SDK Will No Longer Give Automatic `RuntimeIdentifier` Under Certain Circumstances Outside of `Publish`, including `Restore`
### Description

This is because of this PR:  https://github.com/dotnet/sdk/pull/29031 
As noted in https://github.com/dotnet/docs/issues/32816 : In the 7.0 SDK, we added a change that gives an automatic `RuntimeIdentifier` for properties that require a `RuntimeIdentifier`. These properties are as follows: 

- `SelfContained`
- `PublishAot`
- `PublishReadyToRun`
- `PublishSingleFile`
- `PublishSelfContained` ( In 7.0.200+)

However, some of these properties are `Publish` only properties. Namely, - `PublishAot`, `PublishReadyToRun`, `PublishSingleFile`, and `PublishSelfContained`. Yet we still added the implicit `RuntimeIdentifier` even when doing any `dotnet operation`, such as `dotnet build`, if these properties were in the project file or specified in the dotnet invocation.

Now, we have limited the automatic RID for these properties to only appear during publish, and only when doing a dotnet publish from the CLI, not from VS, and not from MSBuild, as those are separate mechanisms, and VS should give the RID in its own codebase.
 
This means there was a short period of time in 7.0 (RC1 to 7.0.2xx) where doing a build with these properties would get a RID, now they don't get a RID. It also affects `--no-restore`, as you got an implicit RID when doing a restore, but now you need to give the RID like `dotnet restore -r RID` if you want `dotnet publish --no-restore` with one of these properties for it to work.

image

 

### Version

.NET 7

### Previous behavior

Previous behavior is if you specified the above properties you would get the RID, now you only get it while doing `dotnet publish`. This also means any restore that was added without a RID in 7.0, that was used to restore for a `publish --no-restore` using one of the 4 above properties will now require a RID.  

### New behavior

The RID is only automatically added in the event of doing a `dotnet publish`. 

### Type of breaking change

- [X] **Binary incompatible**: Existing binaries may encounter a breaking change in behavior, such as failure to load or execute, and if so, require recompilation.

### Reason for change

The RID change was a breaking change and it doesnt make sense for a publish property to affect dotnet build and friends. 

### Recommended action

For something like `restore`, `publish --no-restore`, you need to add the rid via `dotnet restore -r RID` . It would be better to then also be explicit when publishing so the publish has the same RID (`dotnet publish -r RID`), or just remove `--no-restore` from the publish.

For everything else, no action is needed, unless you want to keep the RID, in which case you can add it to the project file like `<RuntimeIdentifier>Foo</RuntimeIdentifier>`

### Feature area

SDK

### Affected APIs

_No response_

---
[Associated WorkItem - 59319](https://dev.azure.com/msft-skilling/Content/_workitems/edit/59319)

gewarren:
Fixes #33051 

[Preview](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/7.0/automatic-rid-publish-only?branch=pr-en-us-33517)

### DIFF
--- a/docs/core/compatibility/sdk/7.0/automatic-rid-publish-only.md
+++ b/docs/core/compatibility/sdk/7.0/automatic-rid-publish-only.md
@@ -1,0 +1,50 @@
+---
+title: "Breaking change: "
+description: Learn about a breaking change in the .NET 7 SDK where a RuntimeIdentifier is automatically added to projects only for the dotnet publish command.
+ms.date: 01/12/2023
+---
+# Automatic RuntimeIdentifier for publish only
+
+In the 7.0.100 SDK, there was [a change](automatic-runtimeidentifier.md) to automatically add `<RuntimeIdentifier>` to projects with properties that require a runtime identifier (RID). Those properties are as follows:
+
+- `SelfContained`
+- `PublishAot`
+- `PublishReadyToRun`
+- `PublishSingleFile`
+- `PublishSelfContained`
+
+However, all of these properties except for `SelfContained` are only used for publishing. Yet the implicit `<RuntimeIdentifier>` was added for any [`dotnet` operation](../../../tools/dotnet.md) if these properties were in the project file or specified as part of the `dotnet` command.
+
+Now, the automatic RID for these properties is limited to only be added during publish. In addition, the automatic RID is only added when using the `dotnet publish` CLI command. It's not added when you publish from Visual Studio or `msbuild`, as those are separate mechanisms, and Visual Studio should provide its own RID.
+
+## Version introduced
+
+.NET 7.0.200 SDK
+
+## Previous behavior
+
+Previously, if you specified any of the mentioned properties, the RID was automatically added to the project.
+
+## New behavior
+
+The RID is only automatically added for the `dotnet publish` command.
+
+If you performed a restore without an RID in .NET 7 and use it to restore for a `publish --no-restore` command on a project with one of the mentioned properties, you'll need to specify an RID using `restore -r <RID>`.
+
+## Type of breaking change
+
+This change can affect [binary compatibility](../../categories.md#binary-compatibility) and [source compatibility](../../categories.md#source-compatibility).
+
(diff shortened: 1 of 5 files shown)


BillWagner (approved):
This LGTM @gewarren 

Let's :shipit:

docs/core/compatibility/sdk/7.0/automatic-rid-publish-only.md
```diff
@@ -0,0 +1,50 @@
...
+
+- `SelfContained`
+- `PublishAot`
+- `PublishReadyToRun`
+- `PublishSingleFile`
+- `PublishSelfContained`
+
+However, all of these properties except for `SelfContained` are only used for publishing. Yet the implicit `<RuntimeIdentifier>` was added for any [`dotnet` operation](../../../tools/dotnet.md) if these properties were in the project file or specified as part of the `dotnet` command.
+
+Now, the automatic RID for these properties is only added during publish. In addition, the automatic RID is only added when using the `dotnet publish` CLI command. It's not added when you publish from Visual Studio or `msbuild`, as those are separate mechanisms, and Visual Studio should provide its own RID.
```
nagilson, line 18: `Now, the automatic RID for these properties is only added during publish.` 


```suggestion
Now, the automatic RID for these properties, excluding `SelfConained`, is only added during publish. In addition, the automatic RID is only added when using the `dotnet publish` CLI command. It's not added when you publish from Visual Studio or `msbuild`, as those are separate mechanisms, and Visual Studio should provide its own RID.
```

It may be better to be explicit here.

nagilson (approved):
Looks good, I see you changed the existing RID doc too. I think this is a... nuanced breaking change, but you described the nuance well and in high clarity. Thanks for taking the time for this :smile:

PR #33544, opened 2023-01-15
[csc.exe cmd line - "response" file] Extra '-' character

glenn-slayden:
When specifying a "response" (input) file for `csc.exe` with the `@` character, there mustn't be a dash in front.

cscrsp


BillWagner (approved):
Thanks for spotting this and fixing it @glenn-slayden 

I'll :shipit: now.

PR #33552, opened 2023-01-16
Specify where to actually set assembly attributes

olfek:
"Set assembly attributes" page ironically doesn't say where to set assembly attributes. Ha.

### DIFF
--- a/docs/standard/assembly/set-attributes.md
+++ b/docs/standard/assembly/set-attributes.md
@@ -15,7 +15,7 @@ dev_langs:
 
 # Set assembly attributes
 
-Assembly attributes are values that provide information about an assembly. The attributes are divided into the following sets of information:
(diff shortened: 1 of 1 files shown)


docs/standard/assembly/set-attributes.md
```diff
@@ -15,7 +15,7 @@ dev_langs:
 
 # Set assembly attributes
 
-Assembly attributes are values that provide information about an assembly. The attributes are divided into the following sets of information:
+Assembly attributes are values that provide information about an assembly. They are usually set in a `AssemblyInfo.cs` file. The attributes are divided into the following sets of information:
```
IEvangelist, line 18:
```suggestion
Assembly attributes are values that provide information about an assembly. They're usually set in an _AssemblyInfo.cs_ file. The attributes are divided into the following sets of information:
```

IEvangelist (approved):
I like this proposal, thanks for adding this. I have a nit about the text and the file reference. Files should be italicized, otherwise, this is great.

PR #33566, opened 2023-01-17
Fix architecture breadcrumb and dotnet command TOC node

gewarren:
- Nest .NET Architecture breadcrumb under .NET
- Add TOC node named "dotnet command reference" for dotnet commands (note that [we already mark these articles as ms.topic = reference](https://github.com/dotnet/docs/blob/adfc9215485d155ea8148ae74b8be2766edada15/docfx.json#L233))

### DIFF
--- a/docs/fundamentals/toc.yml
+++ b/docs/fundamentals/toc.yml
@@ -266,139 +266,141 @@ items:
     - name: Overview
       displayName: '.net cli,command-line interface,cli,tool'
       href: ../core/tools/index.md
-    - name: dotnet
-      href: ../core/tools/dotnet.md
-    - name: dotnet add/list/remove package
+    - name: dotnet command reference
       items:
-      - name: dotnet add package
-        href: ../core/tools/dotnet-add-package.md
-      - name: dotnet list package
-        href: ../core/tools/dotnet-list-package.md
-      - name: dotnet remove package
-        href: ../core/tools/dotnet-remove-package.md
-    - name: dotnet add/list/remove reference
-      items:
-      - name: dotnet add reference
-        href: ../core/tools/dotnet-add-reference.md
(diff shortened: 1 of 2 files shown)


docs/fundamentals/toc.yml
```diff
@@ -266,139 +266,141 @@ items:
...
-        href: ../core/tools/dotnet-tool-uninstall.md
-      - name: dotnet tool update
-        href: ../core/tools/dotnet-tool-update.md
-    - name: dotnet vstest
-      href: ../core/tools/dotnet-vstest.md
-    - name: dotnet watch
-      displayName: hot reload
-      href: ../core/tools/dotnet-watch.md
-    - name: dotnet workload
+    - name: dotnet command reference
```
tdykstra, line 269: I see the logic behind grouping these nodes under a "command reference" node, but the nodes under this new node are already recognizable without the new node in that all of them have names that begin with `dotnet`. And doing this adds an extra level of nesting in a TOC that is already two levels deep at that point. Have you seen any data indicating that an extra node is needed for clarity?

tdykstra: We used to have a Reference node like this but removed it to flatten the TOC, see https://github.com/dotnet/docs/pull/25222

gewarren: We discussed this in the sprint meeting this morning and agreed that the long list of `dotnet xyz` entries obscures the other nodes under CLI, such as tab completion and templates. My thought is that the extra level is worth it to have them all organized together. Also, in the proposed breakout of TOCs, tools and diagnostics would be a separate TOC and this would likely go back to only 3 levels deep. cc @adegeo

tdykstra (approved):

PR #33595, opened 2023-01-18
fix: aligned enum  to proto version

linkdotnet:
## Summary

The enum used in the protobuf enumerations example maps twice to the value 4.
As no other sample uses the `Cancelled`, I removed it.


IEvangelist (approved):

PR #33668, opened 2023-01-22
Add an example of suppressing CS0067, when event is unused intentionally

### LINKED ISSUE #33495: Warning still appears when the event is only there because of an interface
This warning makes sense in the normal case, but we're seeing it in a situation we might not have generally expected it.

Suppose you have this interface:

```cs
public interface IThing
{
    event Action? E;
}
```
Implemented by this class:
```cs
public class Thing : IThing
{
    public event Action? E;
}
```

Suppose there are many implementations of `IThing`, most of which make good use of the `E` event. And this one implementation never encounters a situation where it is needed. Perhaps it is an event about some bit of data changing, but where this implementation never changes that data. There would be no need to use the event, but the event is still required because `Thing` implements the `IThing` interface. The warning is misleading.

We'll likely just `#pragma warning disable` it, but that's kind of ugly. This seems like a situation where the event going unused is justified, because it doesn't compile otherwise.

What am I missing?

---
#### Document Details

* ID: 951244ec-7b1c-d067-aba6-e5074b01fe19
* Version Independent ID: fcb87ea8-0040-004d-14a7-808afd939c30
* Content: [Compiler Warning (level 3) CS0067](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0067)
* Content Source: [docs/csharp/misc/cs0067.md](https://github.com/dotnet/docs/blob/main/docs/csharp/misc/cs0067.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BartoszKlonowski:
This pull request closes #33495
It adds the example of a situation where the event is left unused intentionally and shows a way of suppressing the warning.

This example leaves the original part intact. The example is added as a part below so it's treated as additional detail worth mentioning.
The example is taken from [this comment](https://github.com/dotnet/docs/issues/33495#issuecomment-1380344707) added by @Youssef1313

### DIFF
--- a/docs/csharp/misc/cs0067.md
+++ b/docs/csharp/misc/cs0067.md
@@ -37,4 +37,21 @@ class MyClass
    {  
    }  
 }  
+```  
+  
+If the event is unused intentionaly (for example when it's part of an interface a class implements) then one way to suppress the warning is:  
+  
+```csharp  
+using System;  
+  
+public interface IThing  
+{  
+   event Action? E;  
+}  
+  
+public class Thing : IThing  
+{  
+   // no CS0067 though the event is left unused  
+   public event Action? E { add { } remove { } }  
+}  
 ```
--- a/docs/csharp/misc/cs0067.md
+++ b/docs/csharp/misc/cs0067.md
@@ -39,7 +39,7 @@ class MyClass
 }  
 ```  
   
-If the event is unused intentionaly (for example when it's part of an interface a class implements) then one way to suppress the warning is:  
+If the event is unused intentionally, for example, when it's part of an interface implementation, then you can avoid emitting an unnecessary field as follows:  
   
 ```csharp  
 using System;  


docs/csharp/misc/cs0067.md
```diff
@@ -37,4 +37,21 @@ class MyClass
    {  
    }  
 }  
+```  
+  
+If the event is unused intentionaly (for example when it's part of an interface a class implements) then one way to suppress the warning is:
```
Youssef1313, line 42:
```suggestion
If the event is unused intentionally, for example, when it's part of an interface implementation, then you can avoid emitting an unnecessary field as follows:  
```

BillWagner (approved):
This LGTM @BartoszKlonowski 

I'll :shipit: now

PR #33672, opened 2023-01-22
Remove `win-arm` RIDs from the RID catalog page.

alexrp:
`win-arm` went out of support with .NET 5 and so probably doesn't need to be mentioned so prominently.


tdykstra (approved):
Thanks for the update!

PR #33698, opened 2023-01-24
Add notes about native unions.

AaronRobinsonMSFT:
## Summary

Doc feedback from https://github.com/dotnet/runtime/issues/80973

/cc @gewarren @tannergooding @dotnet/interop-contrib

### DIFF
--- a/docs/standard/native-interop/customize-struct-marshalling.md
+++ b/docs/standard/native-interop/customize-struct-marshalling.md
@@ -333,7 +333,7 @@ struct Currency
 
 ### Unions
 
-A union is a data type that can contain different types of data atop the same memory. It is a common form of data in C. A union can be expressed in .NET using `LayoutKind.Explicit`. It recommended to use structs when defining a union in .NET. Using classes can cause issues with layout and produce unpredictable behavior.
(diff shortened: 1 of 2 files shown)


docs/standard/native-interop/customize-struct-marshalling.md
```diff
@@ -329,6 +331,66 @@ struct Currency
 };
 ```
 
+### Unions
+
+A union is a data type that can contain different types of data atop the same memory. It is a common form of data in C. A union can be expressed in .NET using `LayoutKind.Explicit`. It recommended to use structs when defining a union in .NET. Using classes can cause issues with layout and produce unpredictable behavior.
```
gewarren, line 336:
```suggestion
A union is a data type that can contain different types of data atop the same memory. It's a common form of data in the C language. A union can be expressed in .NET using `LayoutKind.Explicit`. It's recommended to use structs when defining a union in .NET. Using classes can cause layout issues and produce unpredictable behavior.
```

gewarren (approved):
Left a couple things to consider.

jkoritzinsky (approved):
LGTM other than other review comments

tannergooding (approved):

PR #33714, opened 2023-01-25
Extensions breaking changes

### LINKED ISSUE #31785: [Breaking change]: `ActivatorUtilities.CreateInstance` behaves consistently independent of ctor definition order
### Description

We introduced a breaking change in order to fix the bug with behavior dependence on order of constructor definition.

As part of this fix we also made the behavior of `CreateInstance` more coherent with `CreateFactory`, such that when `IServiceProviderIsService` is not present in the DI container then `CreateInstance` falls back to a `CreateFactory` logic, where only one constructor is allowed to match with all the required given input parameters.

In a more general case when we expect `IServiceProviderIsService` to be present, the `CreateInstance` API prefers the longest constructor overload that has all its arguments either available as input to the API or as registered in the container or available with default values to the constructor itself.

Example:

```cs
public class A
{
   A(B b, C c, string st = "default string") { }
   A() { }
}
```

For the above class definition, with `IServiceProviderIsService`, the expectation is for `ActivatorUtilities.CreateInstance<A>(serviceProvider, new C())` to instantiate `A` by picking the first constructor taking `B`, `C` and `string`.

### Version

.NET 8 Preview 1

### Previous behavior

`ActivatorUtilities.CreateInstance` had cases where it behaved unexpectedly.

The API would make sure all required instances passed to it exist in the constructor being picked. But the constructor selection was buggy and not reliable.

### New behavior

General idea:

`CreateInstance` tries to find the longest constructor matching all parameters based on the behavior of IServiceProviderIsService.
If none are found, try to fallback to `CreateFactory`, if it finds more than one constructor it throws
if `IServiceProviderIsService` is not present it falls back to `CreateFactory` logic.
If `IServiceProviderIsService` is configured wrong, or does not exist, it is OK for the API to function incorrectly or ambiguously.

### Type of breaking change

### Reason for change

To fix the bug with inconsistent behavior when constructor overload definition order changes

### Recommended action

Carefully examine constructor definitions if for the given instance type if you are suddenly getting failure now after updating to .NET 8 Preview 1. (Refer to new behavior above).

### Feature area

### Affected APIs

`ActivatorUtilities.CreateInstance`

Fixed in PR https://github.com/dotnet/runtime/pull/75846

---
[Associated WorkItem - 60828](https://dev.azure.com/msft-skilling/Content/_workitems/edit/60828)

### LINKED ISSUE #32175: [Breaking change]: ConfigurationBinder raises exceptions if ErrorOnUnknownConfiguration is set
### Description

If `BinderOptions.ErrorOnUnknownConfiguration` is set and a value in specified in config that cannot be bound to the corresponding property in the bound model, an `InvalidOperationException` is now thrown.

Previously, `BinderOptions.ErrorOnUnknownConfiguration` was solely used to raise an exception if a value existed in config but didn't exist in the model being bound to.

As part of PR https://github.com/dotnet/runtime/pull/77708 (which fixes #73915), this property is now also used to throw exceptions if the value in config cannot be converted to the type of value in the model being bound to.

### Version

.NET 8 Preview 1

### Previous behavior

Previously, this would silently swallow the exceptions for the fields containing invalid enums:

```csharp
public enum TestSettingsEnum
{
    Option1,
    Option2,
}

public class MyModelContainingArray
{
    public TestSettingsEnum[] Enums { get; set; }
}

public void SilentlySwallowsInvalidItems()
{
    var dic = new Dictionary<string, string>
        {
            {"Section:Enums:0", "Option1"},
            {"Section:Enums:1", "Option3"}, // invalid - ignored
            {"Section:Enums:2", "Option4"}, // invalid - ignored
            {"Section:Enums:3", "Option2"},
        };

    var configurationBuilder = new ConfigurationBuilder();
    configurationBuilder.AddInMemoryCollection(dic);
    var config = configurationBuilder.Build();
    var configSection = config.GetSection("Section");

    var model = configSection.Get<MyModelContainingArray>(o => o.ErrorOnUnknownConfiguration = true);

    // we previously would end up with just Option1 and Option2 in the bound collection
}
```

### New behavior

Now, if the value cannot be converted to the type of the value in the model, an `InvalidOperationException` is thrown.

### Type of breaking change

### Reason for change

In the bug report, users were setting `BinderOptions.ErrorOnUnknownConfiguration` and expecting that to throw if there were any issues in binding configuration.

But the original intention of it was to just throw if something was _missing_ from the bound model.

Perhaps if I'd originally chosen a clearer name, this might've avoided this confusion. I suggested [adding another property in BinderOptions](https://github.com/dotnet/runtime/issues/73915#issuecomment-1297731003), but we went instead with [making the existing property dual purpose](https://github.com/dotnet/runtime/issues/73915#issuecomment-1297832467).

### Recommended action

The recommended action is to change or remove the configuration values that cannot be converted to the properties in the bound model.

Alternatively, change `BinderOptions.ErrorOnUnknownConfiguration` from `true` to `false`

### Feature area

Core .NET libraries, Serialization

### Affected APIs

All of the following in `ConfigurationBinder`:

*  `public static T? Get<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] T>(this IConfiguration configuration, Action<BinderOptions>? configureOptions)`
* `public static object? Get(this IConfiguration configuration, Type type, Action<BinderOptions>? configureOptions)`
* `public static void Bind(this IConfiguration configuration, object? instance, Action<BinderOptions>? configureOptions)`

---
[Associated WorkItem - 61240](https://dev.azure.com/msft-skilling/Content/_workitems/edit/61240)

### LINKED ISSUE #34158: [Breaking change]: ActivatorUtilities.CreateInstance(...) no longer allows `null` for the `provider` parameter
### Description

For the class `Microsoft.Extensions.DependencyInjection.ActivatorUtilities`, the [two CreateInstance() methods](https://learn.microsoft.com/dotnet/api/microsoft.extensions.dependencyinjection.activatorutilities.createinstance) will now throw `ArgumentNullException` when the `provider` parameter is `null`.

### Version

.NET 8 Preview 1

### Previous behavior

A `null` value was allowed for the `provider` parameter and did create the specified type correctly for some cases.

### New behavior

When `provider` is `null`, `ArgumentNullException` is thrown.

### Type of breaking change

### Reason for change

Various constructor-matching issues were fixed along with parameter validation to match the intended purpose of `CreateInstance()`. The `CreateInstance()` methods have a non-nullable `provider` parameter, so it was generally expected that a `null` provider was not allowed.

See also the breaking change https://github.com/dotnet/docs/issues/31785 which covers the constructor-matching changes.

### Recommended action

Pass a non-null `IServiceProvider` for the `provider` argument. If the provider also implements `IServiceProviderIsService` then constructor arguments may be obtained through that.

Alternatively, if your scenario doesn't require dependency injection, since `IServiceProvider` is `null`, use [Activator.CreateInstance](https://learn.microsoft.com/dotnet/api/system.activator.createinstance) instead.

### Feature area

Core .NET libraries, Extensions

### Affected APIs

On the class `Microsoft.Extensions.DependencyInjection.ActivatorUtilities`:
```
public static object CreateInstance (IServiceProvider provider, Type instanceType, params object[] parameters);
public static T CreateInstance<T> (IServiceProvider provider, params object[] parameters);
```

Changed in PR https://github.com/dotnet/runtime/pull/75846

gewarren:
Fixes #31785 
Fixes #32175
Fixes #34158 

Previews:

- [CreateInstance behavior](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/extensions/8.0/activatorutilities-createinstance-behavior?branch=pr-en-us-33714)
- [CreateInstance argument validation](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/extensions/8.0/activatorutilities-createinstance-null-provider?branch=pr-en-us-33714)
- [ConfigurationBinder exceptions](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/extensions/8.0/configurationbinder-exceptions?branch=pr-en-us-33714)

@SteveDunn


docs/core/compatibility/extensions/8.0/activatorutilities-createinstance-behavior.md
```diff
@@ -0,0 +1,56 @@
...
+
+- If no constructors are found or if <xref:Microsoft.Extensions.DependencyInjection.IServiceProviderIsService> isn't present, it falls back to <xref:Microsoft.Extensions.DependencyInjection.ActivatorUtilities.CreateFactory(System.Type,System.Type[])> logic.
+- If it finds more than one constructor, it throws an <xref:System.InvalidOperationException>.
+
+> [!NOTE]
+> If <xref:Microsoft.Extensions.DependencyInjection.IServiceProviderIsService> is configured incorrectly or doesn't exist, `CreateInstance` may function incorrectly or ambiguously.
+
+## Type of breaking change
+
+This change is a [behavioral change](../../categories.md#behavioral-change).
```
gewarren, line 44: The issue said "source incompatible", but I didn't see how it affected compilation, so please double check this.

IEvangelist: The interface was added, and the `CallSiteFactory` as part of service lookup now inherits it, thus making it source incompatible, I believe. @davidfowl wrote this and could confirm.

docs/core/compatibility/extensions/8.0/configurationbinder-exceptions.md
```diff
@@ -0,0 +1,73 @@
...
+
+public void SilentlySwallowsInvalidItems()
+{
+    var dictionary = new Dictionary<string, string>
+        {
+            {"Section:Enums:0", "Option1"},
+            {"Section:Enums:1", "Option3"}, // invalid - ignored
+            {"Section:Enums:2", "Option4"}, // invalid - ignored
+            {"Section:Enums:3", "Option2"},
+        };
```
IEvangelist, line 38:
```suggestion
    var dictionary = new Dictionary<string, string>
    {
        ["Section:Enums:0"] = "Option1",
        ["Section:Enums:1"] = "Option3", // invalid - ignored
        ["Section:Enums:2"] = "Option4", // invalid - ignored
        ["Section:Enums:3"] = "Option2",
    };
```

IEvangelist (approved):
Looks good, but have a few comments and suggestions.

PR #33723, opened 2023-01-26
derived class has access to all virtual members

dkj468:
Earlier description was bit confusing as derived class doesn't have access to all the fields or properties. If this is in context to inheritance then it should mention public or protected access modifier. Since this article is related to polymorphism using Virtual keyword is more appropriate

## Summary

Describe your changes here.

### DIFF
--- a/docs/csharp/fundamentals/object-oriented/polymorphism.md
+++ b/docs/csharp/fundamentals/object-oriented/polymorphism.md
@@ -32,7 +32,7 @@ In C#, every type is polymorphic because all types, including user-defined types
 
 ### Virtual members
 
-When a derived class inherits from a base class, it includes all the members of the base class. All the behavior declared in the base class is part of the derived class. That enables objects of the derived class to be treated as objects of the base class. Access modifiers (`public`, `protected`, `private` and so on) determine if those members are accessible from the derived class implementation. Virtual methods gives the designer choice of different behavior of the derived class:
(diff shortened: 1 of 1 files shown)


dkj468:
hi @BillWagner  Please review the PR

docs/csharp/fundamentals/object-oriented/polymorphism.md
```diff
@@ -32,7 +32,7 @@ In C#, every type is polymorphic because all types, including user-defined types
 
 ### Virtual members
 
-When a derived class inherits from a base class, it gains all the methods, fields, properties, and events of the base class. The designer of the derived class has different choices for the behavior of virtual methods:
+When a derived class inherits from a base class, it gains all the virtual methods, virtual properties, and virtual events of the base class. The designer of the derived class has different choices for the behavior of virtual methods:
```
BillWagner, line 35: This paragraph was making a different point. I agree that the previous text was confusing. How about this instead:

```suggestion
When a derived class inherits from a base class, it includes all the members of the base class. All the behavior declared in the base class is part of the derived class. That enables objects of the derived class to be treated as objects of the base class. Access modifiers (`public`, `protected`, `private` and so on) determine if those members are accessible from the derived class implementation. Virtual methods gives the designer different choices for the behavior of the derived class:
```

BillWagner (changes requested):
Hi @dkj468 

I agree that the previous text caused confusion. The text was trying to explain a different aspect of derived classes. I made a suggestion for different wording. Let me know if you think that's more clear.

docs/csharp/fundamentals/object-oriented/polymorphism.md
dkj468: I have amended the PR with new wording. I've made slight change to your suggested wording. Please review it again

BillWagner (approved):

BillWagner:
Thanks @dkj468 

I've approved this PR, and I'll :shipit: as soon as the build completes. You should see the changes on the live site in the next day or so, on our regular publishing cycle.

PR #33171, opened 2022-12-21
Fsharp Stroustrup style guide

josh-degraw:
## Summary

This is an attempt to update the style guide to allow for "Stroustrup" formatting as discussed in https://github.com/fsharp/fslang-design/issues/706.

There may still be several points which have not been agreed upon or that need further refinement. For now this is very much still considered a work in progress.

CC: @nojaf

PR #33730, opened 2023-01-26
Fix some minor issues in FSharp docs

josh-degraw:
## Summary
This includes the suggesions from @gewarren on #33171 that were left just after it got merged.

### DIFF
--- a/docs/fsharp/style-guide/formatting.md
+++ b/docs/fsharp/style-guide/formatting.md
@@ -626,7 +626,7 @@ list[..^1]
 Indentation of conditionals depends on the size and complexity of the expressions that make them up.
 Write them on one line when:
 
-* `cond`, `e1`, and `e2` are short
+* `cond`, `e1`, and `e2` are short.
 * `e1` and `e2` are not `if/then/else` expressions themselves.
 
 ```fsharp
@@ -646,16 +646,6 @@ if a then
 if a then ()
 ```
 
-If any of the expressions are multi-line or `if/then/else` expressions.
-
-```fsharp
-// ✔️ OK
-if cond then
-    e1
-else
-    e2
-```
-
 Multiple conditionals with `elif` and `else` are indented at the same scope as the `if` when they follow the rules of the one line `if/then/else` expressions.
 
 ```fsharp
@@ -671,6 +661,7 @@ If any of the conditions or expressions is multi-line, the entire `if/then/else`
 ```fsharp
 // ✔️ OK
 if cond1 then
+    let e1 = something()
     e1
 elif cond2 then
     e2
@@ -826,7 +817,7 @@ let pascalsTriangle =
     |]
 ```
 
-If a list or array expression is the right-hand side of a binding, you may prefer to use "Stroustrup" style:
+If a list or array expression is the right-hand side of a binding, you may prefer to use `Stroustrup` style:
 
 ```fsharp
 // ✔️ OK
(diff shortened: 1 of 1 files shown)


docs/fsharp/style-guide/formatting.md
```diff
@@ -646,14 +646,21 @@ if a then
 if a then ()
 ```
 
-If any of the expressions are multi-line or `if/then/else` expressions.
+If any of the expressions are multi-line, each conditional branch should be multi-line.
```
josh-degraw, line 649: I wasn't sure if I should update this for both of these cases, since they're essentially describing the same thing, one just includes some `elif` branches, so I aired on the side of including both.

docs/fsharp/style-guide/formatting.md
```diff
@@ -646,14 +646,21 @@ if a then
...
 
 ```fsharp
 // ✔️ OK
 if cond then
+    let e1 = something()
```
josh-degraw, line 654: Both of these sections were describing how to format expressions when the content was multi-line, but neither actually had a multi-line expression in them, so that's why I added these `let e1 = something()` lines.

I wasn't sure if I should update this for both of these cases, since they're essentially describing the same thing, one just includes some `elif` branches, but I aired on the side of including both for completeness' sake.

docs/fsharp/style-guide/formatting.md
```diff
@@ -1689,7 +1705,7 @@ type PostalAddress =
     }
 ```
 
-You may also put the `{` at the end of the type declaration line like ("stroustrup" style):
+You may also put the `{` at the end of the type declaration line like (`Stroustrup` style):
```
gewarren, line 1708:
```suggestion
You can also put the `{` at the end of the first line of the type definition (`Stroustrup` style):
```

docs/fsharp/style-guide/formatting.md
```diff
@@ -1667,7 +1683,7 @@ For any custom operator that starts with `*` and that has more than one characte
 
 ### Formatting record declarations
 
-For record declarations, by default you should indent `{` in type definition by four spaces, start the field list on the same line and align any members with the `{` token:
+For record declarations, by default you should indent `{` in type definitions by four spaces, start the field list on the same line, and align any members with the `{` token:
```
gewarren, line 1686:
```suggestion
For record declarations, by default you should indent the `{` in the type definition by four spaces, start the label list on the same line, and align members, if any, with the `{` token:
```

docs/fsharp/style-guide/formatting.md
```diff
@@ -1667,7 +1683,7 @@ For any custom operator that starts with `*` and that has more than one characte
...
 
 ```fsharp
 // ✔️ OK
```
gewarren, line 1689: Should line 1696 say "...with **labels** indented by an additional four spaces"?

gewarren (commented):
I suggested the use of "label" based on https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/records.

docs/fsharp/style-guide/formatting.md
josh-degraw: No, it's referring to additional members as described [here](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/records#records-and-members). I'll update the example to include one for clarity

gewarren (approved):
Thanks @josh-degraw

PR #33759, opened 2023-01-29
Update to ReadOnlyProperty  Counter

sannyg0701:
The following code snippet results in the error "The modifier 'readonly' is not valid for this item" in VS2022
```private int counter;
public int Counter
{
    readonly get => counter;
    set => counter = value;
}
```
Removing both the getter and the setter makes the error go away

## Summary

Describe your changes here.


BillWagner:
Hi @sannyg0701 

The original sample code is correct. You can read about the feature [here](https://learn.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-8.0/readonly-instance-members).

Is your project targeting .NET Framework? That would cause the compiler to use C# 7.3, which doesn't support this feature.

I'll close this PR, as the original code is correct.

PR #33762, opened 2023-01-30
snippet fix: need 'ref' passing Utf8JsonReader

glenn-slayden:
The existing snippet code is obviously incorrect (since `Utf8JsonReader` is a `ref struct`), but in the example as shown it doesn't matter because `GetMoreBytesFromStream` is never called, since the json string in the demo is less than 4096 bytes.

This PR would fix the C# version, but note that I did not test it.

I assume there is a similar error in the VB version of this code but I don't know how to fix it.

Please feel free to process this PR as desired without my further participation.


gewarren:
@tdykstra @eiriktsarpalis Can you review this?

eiriktsarpalis (approved):

gewarren:
Thank you @glenn-slayden!

PR #33789, opened 2023-01-31
Update migration mark down with note

AaronRobinsonMSFT:
Add alternative approaches for C++/CLI

### DIFF
--- a/docs/core/porting/cpp-cli-alts.md
+++ b/docs/core/porting/cpp-cli-alts.md
@@ -5,7 +5,7 @@ author: AaronRobinsonMSFT
 ms.date: 01/30/2023
 ---
 
-# Migration from C++/CLI to ISO C++ and/or C#
+# Migration from C++/CLI to ISO C++ and/or C\#
 
 The C++/CLI language has been around since 2005 and has enabled users to easily integrate C++ and a .NET language into a single binary. The utility of C++/CLI remains as an innovative way to integrate C++ with the .NET runtime on the Windows platform. There are however limitations to using C++/CLI and as ISO C++ and .NET platforms evolve these limitations are expected to grow.
 
@@ -154,7 +154,7 @@ static unsafe partial IMyClass get_myclass();
 
 Defining COM interfaces for a C++ API can be daunting and expensive. There is a lot of boilerplate code for COM interop and does represent an older C++ style that would be pushed into codebases that are forward looking. COM and by extension, COM interop, carry opinionated design considerations and behavior which may make adoption difficult in some scenarios.
 
-## Port C++/CLI to C#
+## Port C++/CLI to C\#
 
 Converting a C++/CLI project into C# where possible and then calling system APIs via P/Invoke is a common option and the preferred mechanism for the .NET runtime team. This means converting as much C++/CLI as possible into C# and then calling C exports that group logical units of C++. This approach is similar to providing a C API for the entire C++ API previously mentioned.
 
--- a/docs/core/porting/cpp-cli-alts.md
+++ b/docs/core/porting/cpp-cli-alts.md
@@ -2,26 +2,24 @@
 title: Migrating C++/CLI projects to ISO C++ and/or .NET
 description: Learn about alternatives to C++/CLI
 author: AaronRobinsonMSFT
-ms.date: 01/30/2023
+ms.date: 02/10/2023
 ---
 
 # Migration from C++/CLI to ISO C++ and/or C\#
 
-The C++/CLI language has been around since 2005 and has enabled users to easily integrate C++ and a .NET language into a single binary. The utility of C++/CLI remains as an innovative way to integrate C++ with the .NET runtime on the Windows platform. There are however limitations to using C++/CLI and as ISO C++ and .NET platforms evolve these limitations are expected to grow.
+The C++/CLI language has been around since 2005 and has enabled users to integrate C++ and a .NET language into a single binary. The utility of C++/CLI remains as an innovative way to integrate C++ with the .NET runtime on the Windows platform. There are however limitations to using C++/CLI and as ISO C++ and .NET platforms evolve these limitations are expected to grow.
 
 **Current Limitations**
 
 - Only operates on the Windows platform.
-
 - Limited functional support for the latest ISO C++.
-
 - Limited support for newer .NET types and patterns (for example, [`Span<T>`](/dotnet/api/system.span-1) and [default interface methods](/dotnet/csharp/language-reference/proposals/csharp-8.0/default-interface-methods)).
 
 If the above limitations are a concern then this document is intended to provide options on how to transtition from C++/CLI to ISO C++ and/or .NET. The approaches described represent a preference for cross-platform and access to the respective language's entire feature set.
 
 ## Expose C++ as C exports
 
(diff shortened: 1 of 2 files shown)


docs/core/porting/cpp-cli.md
```diff
@@ -1,13 +1,16 @@
...
 
-Beginning with .NET Core 3.1 and Visual Studio 2019, [C++/CLI projects](/cpp/dotnet/dotnet-programming-with-cpp-cli-visual-cpp) can target .NET Core. This support makes it possible to port Windows desktop applications with C++/CLI interop layers to .NET Core/.NET 5+. This article describes how to port C++/CLI projects from .NET Framework to .NET Core 3.1.
+Beginning with .NET Core 3.1 and Visual Studio 2019, [C++/CLI projects](/cpp/dotnet/dotnet-programming-with-cpp-cli-visual-cpp) can target .NET Core. This support makes it possible to port Windows desktop applications with C++/CLI interop layers to .NET Core/.NET. This article describes how to port C++/CLI projects from .NET Framework to the latest .NET.
+
+> [!NOTE]
+> C++/CLI is a fully supported language, however, we do not expect any future improvements. For example, we do not expect to add support for new .NET features like `Span<T>` or for new C++ standard versions. C++/CLI users should consider if this level of support is sufficient or if transitioning to ISO C++ and another .NET language, for example, C#, is a better approach for their needs. Migration approaches can be found [here](./cpp-cli-alts.md).
```
AaronRobinsonMSFT, line 13: > or for new C++ standard versions.

@tgani-msft should this instead be clarified to mean "no functional support" from newer C++ standard versions? I recall some mention of import support for modules but not export. Threading the needle on phrasing here is proving difficult.

richlander: > or if transitioning to ISO C++ and another .NET language, for example, C#, is a better approach

I think this phrasing is a bit confusing, for two reasons:

- It says "another .NET language", but it is ambiguous which one was the other one. A naive read would think it is ISO C++ but you mean C++/CLI.
- I think you should simply say C# and not worry about the few folks that will pick another language instead. Also, your language is already sufficiently loose. It is not prescriptive.

Here's an actual suggestion.

```suggestion
> C++/CLI is a fully supported language, however, we do not expect any future improvements. For example, we do not expect to add support for new .NET features like `Span<T>` or for new C++ standard versions. C++/CLI users should consider if this level of support is sufficient or if transitioning to ISO C++ and C# is a better approach for their needs. Migration approaches can be found [here](./cpp-cli-alts.md).
```

AaronRobinsonMSFT: I agree. Thanks.

docs/core/porting/cpp-cli-alts.md
```diff
@@ -0,0 +1,175 @@
+---
+title: Migrating C++/CLI projects to ISO C++ and/or .NET
+description: Learn about alternatives to C++/CLI
+author: AaronRobinsonMSFT
+ms.date: 01/30/2023
+---
+
+# Migration from C++/CLI to ISO C++ and/or C\#
+
+The C++/CLI language has been around since 2005 and has enabled users to easily integrate C++ and a .NET language into a single binary. The utility of C++/CLI remains as an innovative way to integrate C++ with the .NET runtime on the Windows platform. There are however limitations to using C++/CLI and as ISO C++ and .NET platforms evolve these limitations are expected to grow.
```
richlander, line 10: > easily

We typically try to avoid this word since it is very subjective. Also, we don't want to oversell something that might not be the best forward-looking choice.

docs/core/porting/cpp-cli-alts.md
```diff
@@ -0,0 +1,175 @@
...
+
+**Current Limitations**
+
+- Only operates on the Windows platform.
```
richlander, line 14: There usually isn't a blank line between markdown bullets.

docs/core/porting/cpp-cli-alts.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Expose C++ as C exports
+
+.NET has supported directly calling C code since .NET Framework 1.0 through the [Platform Invoke](/dotnet/standard/native-interop/pinvoke) (P/Invokes) mechanism. All C++ APIs that are expected to be called from .NET could be wrapped in a C-style function. This approach has always worked with .NET but prior to .NET 7, the recommendation was to use [`DllImportAttribute`](/dotnet/api/system.runtime.interopservices.dllimportattribute). Since .NET 7, the [`LibraryImportAttribute`](/dotnet/standard/native-interop/pinvoke-source-generation?source=recommendations) should be prefered in all cases. As it relates to C++/CLI, the `LibraryImportAttribute` mechanism provides a higher performance solution to calling C exports than previously available. Furthermore, the extensibility provided by `LibraryImportAttribute` enables developers to create customized marshalling schemes where they are in full control of how and when data is marshalled across the interop boundary. This is in stark contrast to using `DllImportAttribute` that imposes significant performance penalties if custom marshalling is needed.
```
richlander, line 24: > supported since .NET Framework 1.0

I get it, but it is unnecessary to provide this historical context.

```suggestion
.NET supports directly calling C code through the [Platform Invoke](/dotnet/standard/native-interop/pinvoke) (P/Invokes) mechanism. All C++ APIs that are expected to be called from .NET could be wrapped in a C-style function. This approach has always worked with .NET but prior to .NET 7, the recommendation was to use [`DllImportAttribute`](/dotnet/api/system.runtime.interopservices.dllimportattribute). Since .NET 7, the [`LibraryImportAttribute`](/dotnet/standard/native-interop/pinvoke-source-generation?source=recommendations) should be prefered in all cases. As it relates to C++/CLI, the `LibraryImportAttribute` mechanism provides a higher performance solution to calling C exports than previously available. Furthermore, the extensibility provided by `LibraryImportAttribute` enables developers to create customized marshalling schemes where they are in full control of how and when data is marshalled across the interop boundary. This is in stark contrast to using `DllImportAttribute` that imposes significant performance penalties if custom marshalling is needed.
```

docs/core/porting/cpp-cli-alts.md
```diff
@@ -0,0 +1,175 @@
...
+static unsafe partial int call_get_value(void* mc);
+```
+
+### Concerns
+
+The wrapping of complex or large C++ APIs to C is tedious and error prone. Source generation techniques can be employed to help create wrapped C APIs, but even this may represent significant effort.
+
+## Expose C++ as `IUnknown` based APIs
+
+Instead of reducing object-oriented concepts down to what can be expressed in C, using [COM interop](/dotnet/standard/native-interop/cominterop) permits keeping some aspects of C++. The .NET platform has a long history of reducing friction when using COM style interfaces. There is existing tools that can help generate COM interface definitions and [cross-platform support](/dotnet/standard/native-interop/tutorial-comwrappers) through the [`ComWrappers`](/dotnet/api/system.runtime.interopservices.comwrappers) API for `IUnknown` interop was enabled in .NET 6.
```
richlander, line 82:
```suggestion
Instead of reducing object-oriented concepts down to what can be expressed in C, using [COM interop](/dotnet/standard/native-interop/cominterop) permits keeping some aspects of C++. The .NET platform has a long history of reducing friction when using COM style interfaces. There are tools that can help generate COM interface definitions and [cross-platform support](/dotnet/standard/native-interop/tutorial-comwrappers) through the [`ComWrappers`](/dotnet/api/system.runtime.interopservices.comwrappers) API for `IUnknown` interop.
```

Not necessary to mention .NET 6. It is now the oldest version we support.

adegeo:
@AaronRobinsonMSFT @richlander Do we want to set an absolute date for getting this marked as ready for review. Just peeking at this I noticed:

- This is almost 2 years old.
- The links used in the article are using absolute html links, but they should be relative file links.
- Merge conflict.
- We'll need to run it through acrolinx for an editing pass.

AaronRobinsonMSFT:
I'll close this for now. There was some pushback from the C++ side so that needs to be handled offline first.

PR #33796, opened 2023-01-31
ITypeDescriptorContext properties marked nullable

### LINKED ISSUE #33225: [Breaking change]: Nullable annotation corrected for System.ComponentModel.ITypeDescriptorContext
### Description

`System.ComponentModel.ITypeDescriptorContext` has 3 properties that were previously annotated as being non-nullable, but they were indeed nullable in practice. The nullable annotations for these properties have been updated to indicate that they are nullable. This can result in new build warnings related to use of nullable members.

* [Container](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.itypedescriptorcontext.container?view=net-7.0)
* [Instance](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.itypedescriptorcontext.instance?view=net-7.0)
* [PropertyDescriptor](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.itypedescriptorcontext.propertydescriptor?view=net-7.0)

### Version

.NET 8 Preview 1

### Previous behavior

The properties were previously annotated as not being nullable. Consuming their values and assuming they could not be null would not produce any warnings during build.

### New behavior

The properties are now annotated as being nullable. Consuming the values without null checks will result in build warnings.

### Type of breaking change

### Reason for change

As was reported and fixed in dotnet/runtime#78287, the annotations of these properties were incorrect. This change applies the appropriate behavior for the properties and ensures callers understand the values can be null.

### Recommended action

Update calling code to guard against null for these properties.

### Feature area

Core .NET libraries

### Affected APIs

* [System.ComponentModel.ITypeDescriptorContext.Container](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.itypedescriptorcontext.container?view=net-7.0)
* [System.ComponentModel.ITypeDescriptorContext.Instance](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.itypedescriptorcontext.instance?view=net-7.0)
* [System.ComponentModel.ITypeDescriptorContext.PropertyDescriptor](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.itypedescriptorcontext.propertydescriptor?view=net-7.0)

---
[Associated WorkItem - 57957](https://dev.azure.com/msft-skilling/Content/_workitems/edit/57957)

gewarren:
Fixes #33225 

[Preview](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/8.0/itypedescriptorcontext-props?branch=pr-en-us-33796)


jeffhandley (approved):
Sorry for the delayed review on this. Looks good!

PR #33822, opened 2023-02-01
Typo fix - update to inheritance in C# and .NET article

mahab339:
## Summary

* Add closing quotation mark to ("<member> cannot override inherited member <member> because it is not marked virtual, abstract, or override.").
* Change ( "cannot derive from sealed type <typeName>".) to (...<typeName>."). This moves last dot to inside of quotation mark, as other quoted compiler errors through out the article.


BillWagner (approved):
Thanks @mahab339 

You should see the updates on our live site in the next day or two, on our regular publishing cycle.

Issue #33826, opened 2023-02-01
Update "What's New" config to match changes in "Fundamentals" TOC


BillWagner:
See https://github.com/dotnet/docs/pull/33797#discussion_r1093494919

Now that the '.NET Fundamentals' TOC area has been split into multiple areas, the What's new config should reflect that.

It should be just modifying this section of the config file: https://github.com/dotnet/docs/blob/main/.whatsnew.json#L49-L52

---
[Associated WorkItem - 61980](https://dev.azure.com/msft-skilling/Content/_workitems/edit/61980)

gewarren:
@BillWagner I will update the config. However, I'm worried that we'll still get "catch all" duplicates. For example, "core/compatibility" files get put under ".NET breaking changes", but then they're also caught by "core" for ".NET fundamentals". How can we avoid that?

PR #33831, opened 2023-02-01
Add fine-grained feature areas

gewarren:
Contributes to #33826.

### DIFF
--- a/.whatsnew.json
+++ b/.whatsnew.json
@@ -47,6 +47,47 @@
             "heading": ".NET breaking changes"
         },
         {
+            "names": [ "standard/linq", "standard/data" ],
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks @gewarren 

Let's :shipit:

We can run the tool again after this is merged, and update the existing new article. What do you think?

gewarren:
@BillWagner I think there is still some work we need to do in the tool logic to exclude files that have already been captured - see https://github.com/dotnet/docs/issues/33826#issuecomment-1412711488.

BillWagner:
Thanks @gewarren 

Let's go ahead and merge this. I can use it to test the code changes then.

PR #33833, opened 2023-02-02
DateTime.Add precision breaking change

### LINKED ISSUE #33492: [Breaking change]: DateTime Add methods precision change
### Description

In .NET 6 and earlier versions only, the value parameter of the DateTime Add methods is rounded to the nearest millisecond. In .NET 7 and later versions, the full System.Double precision of the value parameter is used. However, due to the inherent imprecision of floating-point math, the resulting precision will vary.

### Version

.NET 7

### Previous behavior

The value parameter of the DateTime Add methods is rounded to the nearest millisecond.

### New behavior

Get results with a better System.Double precision depending on inherented imprecision of floating-point math.

### Type of breaking change

### Reason for change

Community request to enhance the precision inside DateTime. [Request Example](https://github.com/dotnet/runtime/issues/66815).

### Recommended action

No specific action needed more than if having code depends on the precision of the Add methods, will need to review this code and retest it to avoid any surprises with the precision change. 

### Feature area

Core .NET libraries

### Affected APIs

DateTime all Add methods.

gewarren:
Fixes #33492.

[Preview](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/7.0/datetime-add-precision?branch=pr-en-us-33833)


tarekgh (approved):
Thanks @gewarren, LGTM!

BillWagner (approved):
This LGTM @gewarren 

Let's :shipit:

PR #33849, opened 2023-02-03
Example Code fix

robertmuehsig:
## Summary

The sample code is wrong, it should be `armClient`.


scottaddie:
Fixed in https://github.com/dotnet/docs/pull/33997

PR #33857, opened 2023-02-04
dotnet pack/publish breaking changes

### LINKED ISSUE #33642: [Breaking change]: `dotnet publish` uses `Release` configuration by default
### Description

NOTE: Working on a corresponding document for `dotnet pack` and `PackRelease`. EDIT: is it done now here: https://github.com/dotnet/docs/issues/33643.)
Dependent on the changes in this PR: https://github.com/dotnet/sdk/pull/29155

`dotnet publish` uses `Release` instead of `Debug` in 8.0+ TargetFrameworks.

### Version

.NET 8 Preview 1

### Previous behavior

Previously `dotnet publish` would use `Debug` configuration unless told otherwise. 
`PublishRelease` was added in 7.0 as a path forward to this breaking change.  

### New behavior

When using the NET 8.0+ SDK:

Performing `dotnet publish` will use the `Release` Configuration for any projects that have a `TargetFramework` of NET 8.0 or higher.
If the project is multi-targeted, you must specify a target framework to publish, which was an existing behavior. The target framework specified, say, with `-f`, will be what determines the configuration in this case.

If the object under publish is a solution file, then the following will occur:
For 8.0+ projects in the solution: The value of `PublishRelease`, the property that sets the configuration, will be implicitly set to `true` if undefined.
Yet, all projects in the solution must agree on their value of `PublishRelease`. If they do not agree we cannot determine the correct `Configuration` to use. So users must add `PublishRelease` as `false` in the new projects if they had it specified as `false` in older projects. 

In addition:

The environment variable / 7.0 feature flag `DOTNET_CLI_ENABLE_PUBLISH_RELEASE_FOR_SOLUTIONS` has been removed. It was documented here https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props. This behavior is enabled by default now for solutions and that granular control is no longer necessary. 

***People may have CICD scripts/programs that break when they bump the TargetFramework to 8.0 Because `Debug` can be hardcoded in the output paths of `dotnet`.*** Also, for people who have wrote code that is unstable in `Release` configuration (yikes), they may not expect this change if their app only works in `Debug` and they go to `publish` it. 

### Type of breaking change

### Reason for change

Customers have asked us to use `Release` for publish for a long time, nearly since the inception of SDK CLI. Also, VS has had this behavior for many years. It makes sense if you do a publish you want your code to be optimized and don't need debugging information in your app to make it smaller. 

### Recommended action

If a user's solution publish is broken:

One thing the user can do is go to each project file and specify `PublishRelease` like 
```xml
<PublishRelease>true</PublishRelease>
```
(or false, if they also add false to the 8.0+ projects, if they want the old behavior.)

They can also, if the projects don't define `PublishRelease`, create a Directory.Build.Props file (https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022) and specify the value of `PublishRelease` there for every project in the solution.
Note that if this is false, the 8.0+ projects must specify it as false explicitly. This does not solve the case where `PublishRelease` is specified as `false` in some projects and `true` in other projects.

The new environment variable `DOTNET_CLI_DISABLE_PUBLISH_AND_PACK_RELEASE` can be used to disable this feature entirely.

They can also specify a configuration with `-c` or `--configuration` to disable this break. Customers who already did this will not be broken.
This means those who have already set `-c Release` won't be affected, but those who did not decide on a `Configuration` will be made more aware of the change.

If the user's pipelines / scripts / code etc is broken after upgrading:

They will need to update their code to expect `Release` instead of `Debug`, or disable the feature or go back to `Debug` using one of the above methods.

### Feature area

SDK

### Affected APIs

_No response_

cc @richlander 

---
[Associated WorkItem - 62391](https://dev.azure.com/msft-skilling/Content/_workitems/edit/62391)

### LINKED ISSUE #33643: [Breaking change]: `dotnet pack` now uses the `Release` `Configuration` by default in 8.0+ SDKs
### Description

`dotnet pack` will use `Release` over `Debug` by default now. 
This is rather similar to https://github.com/dotnet/docs/issues/33642 in most aspects. However, there are a few differences I will note below.

 cc @richlander 

### Version

.NET 8 Preview 1

### Previous behavior

`dotnet pack` will use `Debug` by default. 
`PackRelease` was added in 7 as a way to enable this default toward the breaking change. It is still around. https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#packrelease

* Likely, the most important thing is that customers may expect `Debug` to be in the output path of `dotnet`, which is where we expect most people to be broken, as it will instead become `Release`. (CI/CD and scripts, users trying to debug a packed app, etc.)

### New behavior

`dotnet pack` will use `Release` unless told otherwise with an option like `--configuration`.  All projects will default with `PackRelease=true`.

NOTE: `dotnet publish` has a very similar breaking change but that change is gated on the `TargetFramework`. This breaking change will occur regardless of the `TargetFramework`, and merely by updating the .NET SDK to an 8.0 version.

When multitargeting with `publish` you must give one TFM (`TargetFramework`) to publish for. Pack is different in that it can pack for many TFM at once. This causes a conflict where you could have multiple TFM, some with a default of packing as `Release`, but others with the default of packing as `Debug`. But such an implementation is not possible without rewriting much of the MSBuild code in the SDK and breaking customers even further. We expect less customers to be broken by a global SDK (as opposed to TFM break) for `dotnet pack` as opposed to `dotnet publish`, which is why these two will behave differently. 

Packing a solution behaves similarly to `dotnet publish`'s breaking change documented here: https://github.com/dotnet/docs/issues/33642

However, because all projects will default with `PackRelease=true`, if the old behavior is desired, _every_ project, not just 8.0 TFM projects, must set `PackRelease=false` to pack as `Debug` by default without any other settings needed. 

`DOTNET_CLI_ENABLE_PACK_RELEASE_FOR_SOLUTIONS` has been deprecated like the publish alternative.

We also expect a performance regression in `dotnet pack` due to this.

### Type of breaking change

### Reason for change

Same as https://github.com/dotnet/docs/issues/33642

### Recommended action

Same as https://github.com/dotnet/docs/issues/33642 (note this environment variable disables BOTH features, unlike the deprecated one which was singular to each feature.) 

And replace the `PublishRelease` or `publish` with `PackRelease` or `pack`. There is the added nuance above with solutions because this is not a TFM gated breaking change. Please reach out if I can add more clarity on this.

### Feature area

SDK

### Affected APIs

_No response_

---
[Associated WorkItem - 62395](https://dev.azure.com/msft-skilling/Content/_workitems/edit/62395)

gewarren:
Fixes #33642
Fixes #33643 

Previews:

- [dotnet pack](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/dotnet-pack-config?branch=pr-en-us-33857)
- [dotnet publish](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/dotnet-publish-config?branch=pr-en-us-33857)

### DIFF
--- a/docs/core/compatibility/sdk/8.0/dotnet-pack-config.md
+++ b/docs/core/compatibility/sdk/8.0/dotnet-pack-config.md
@@ -5,6 +5,57 @@ ms.date: 02/02/2023
 ---
 # 'dotnet pack' uses Release configuration
 
+The [`dotnet pack` command](../../../tools/dotnet-pack.md), which packs code into a NuGet package, now uses the `Release` configuration instead of the `Debug` configuration by default.
+
+## Previous behavior
+
+Previously, `dotnet pack` used the `Debug` configuration unless the configuration was specified explicitly or `PackRelease` was set to `true`.
+
+The [`PackRelease` property](../../../project-sdk/msbuild-props.md#packrelease) was added in .NET 7 as a path forward to this breaking change. Previously, you could set the `DOTNET_CLI_ENABLE_PACK_RELEASE_FOR_SOLUTIONS` environment variable to use `PackRelease` in a project that was part of a Visual Studio solution.
+
+## New behavior
+
+If you're developing with the .NET 8 SDK or a later version, `dotnet pack` uses the `Release` configuration by default for *all* projects. If you have a CI/CD script where you've hardcoded `Debug` into an output path, this change may break your workflow. Also, you won't be able to debug a packed app unless the `Debug` configuration was explicitly specified (for example, using `dotnet pack --configuration Debug`.
+
+`dotnet pack` can pack for multiple target framework monikers (TFM) at the same time. If your project targets multiple versions and you have different `PackRelease` values for different targets, you can have a conflict where some TFMs pack the `Release` configuration and others pack the `Debug` configuration.
+
+### Projects in a solution
+
+`dotnet pack` can pack all the projects in a Visual Studio solution if given a solution file. For each project in the solution, the value of `PackRelease` is implicitly set to `true` if it's undefined. In order for `dotnet pack` to determine the correct configuration to use, all projects in the solution must agree on their value of `PackRelease`.
+
+The `DOTNET_CLI_ENABLE_PACK_RELEASE_FOR_SOLUTIONS` environment variable is no longer recognized. By default, the executable project's `PackRelease` value takes precedence and is flowed to other projects in the solution.
+
+## Version introduced
+
+.NET 8 Preview 1
+
+## Type of breaking change
+
+This change can affect [source compatibility](../../categories.md#source-compatibility) and is also a [behavioral change](../../categories.md#behavioral-change).
+
+## Reason for change
+
+In most cases when you create a package, you want your code to be optimized and can keep the package smaller by excluding debugging information.
+
+The `DOTNET_CLI_ENABLE_PACK_RELEASE_FOR_SOLUTIONS` environment variable was removed since the behavior it enabled is now the default behavior and the granular control is no longer necessary.
+
+## Recommended action
+
+- To disable the new behavior entirely, you can set the `DOTNET_CLI_DISABLE_PUBLISH_AND_PACK_RELEASE` environment variable to `true` (or any other value). This environment variable affects both `dotnet publish` and `dotnet pack`.
+
+- To explicitly specify the `Debug` configuration for packing, use the `-c` or `--configuration` option with `dotnet pack`.
+
+- If your CI/CD pipeline is broken due to hardcoded output paths, update the paths to `Release` instead of `Debug`, disable the new behavior using the `DOTNET_CLI_DISABLE_PUBLISH_AND_PACK_RELEASE` environment variable, or specify that the `Debug` configuration should be used.
+
+- If you're packing a solution and it's broken because one or more projects explicitly sets a value for `PackRelease`, you should explicitly set `PackRelease` to `false` in each project:
+
+  ```xml
+  <PropertyGroup>
+    <PackRelease>false</PackRelease>
+  </PropertyGroup>
+  ```
+
+- If you're packing a solution and the performance has regressed, you can set the `DOTNET_CLI_LAZY_PUBLISH_AND_PACK_RELEASE_FOR_SOLUTIONS` environment variable to `true` (or any other value) to remove the regression.
 
 ## See also
 
--- a/docs/core/compatibility/sdk/8.0/dotnet-pack-config.md
+++ b/docs/core/compatibility/sdk/8.0/dotnet-pack-config.md
@@ -1,0 +1,11 @@
+---
+title: "Breaking change: 'dotnet pack' uses Release configuration"
+description: Learn about a breaking change in the .NET 8 SDK where 'dotnet pack' uses the 'Release' configuration by default.
+ms.date: 02/02/2023
+---
+# 'dotnet pack' uses Release configuration
+
+
+## See also
+
+- ['dotnet publish' uses Release configuration](dotnet-publish-config.md)
--- a/docs/core/compatibility/sdk/8.0/dotnet-pack-config.md
+++ b/docs/core/compatibility/sdk/8.0/dotnet-pack-config.md
@@ -15,7 +15,7 @@ The [`PackRelease` property](../../../project-sdk/msbuild-props.md#packrelease) 
 
 ## New behavior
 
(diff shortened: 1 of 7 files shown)


docs/core/compatibility/sdk/8.0/dotnet-publish-config.md
```diff
@@ -0,0 +1,66 @@
...
+
+For projects in a solution:
+
+- `dotnet publish` can publish all the projects in a Visual Studio solution if given a solution file. For the solution projects that target .NET 8 or later, the value of `PublishRelease` is implicitly set to `true` if it's undefined. However, in order for `dotnet publish` to determine the correct configuration to use for the solution, all projects in the solution must agree on their value of `PublishRelease`. If an older project in the solution has `PublishRelease` set to `false`, you should explicitly set the property to `false` for any new .NET 8+ projects as well.
+
+- This change might cause the performance of `dotnet publish` to regress, especially for solutions that contain many projects. To address this, a new environment variable `DOTNET_CLI_LAZY_PUBLISH_AND_PACK_RELEASE_FOR_SOLUTIONS` has been introduced.
+
+- The `DOTNET_CLI_ENABLE_PUBLISH_RELEASE_FOR_SOLUTIONS` environment variable is no longer recognized. By default, the executable project's `PublishRelease` value takes precedence and is flowed to other projects in the solution.
```
nagilson, line 28:
```suggestion
- The `DOTNET_CLI_ENABLE_PUBLISH_RELEASE_FOR_SOLUTIONS` environment variable is no longer recognized.
```

See the comment on the other copy.

docs/core/compatibility/sdk/8.0/dotnet-pack-config.md
```diff
@@ -0,0 +1,64 @@
...
+
+- If you're packing a solution and it's broken because one or more projects explicitly sets a value for `PackRelease`, you should explicitly set `PackRelease` to `false` in each project:
+
+  ```xml
+  <PropertyGroup>
+    <PackRelease>false</PackRelease>
+  </PropertyGroup>
+  ```
+
+- If you're packing a solution and the performance has regressed, you can set the `DOTNET_CLI_LAZY_PUBLISH_AND_PACK_RELEASE_FOR_SOLUTIONS` environment variable to `true` (or any other value) to remove the regression. This variable affects both `dotnet publish` and `dotnet pack`.
```
nagilson, line 60:
```suggestion
- If you're packing a solution and the performance has regressed, you can set the `DOTNET_CLI_LAZY_PUBLISH_AND_PACK_RELEASE_FOR_SOLUTIONS` environment variable to `true` (or any other value) to remove the regression. However, if any project defines `PackRelease`, when using this variable, all projects must, or you can use a _Directory.Build.Props_ file. This variable affects both `dotnet publish` and `dotnet pack`.
```

docs/core/compatibility/sdk/8.0/dotnet-pack-config.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Previous behavior
+
+Previously, `dotnet pack` used the `Debug` configuration unless the configuration was specified explicitly or `PackRelease` was set to `true`.
+
+The [`PackRelease` property](../../../project-sdk/msbuild-props.md#packrelease) was added in .NET 7 as a path forward to this breaking change. Previously, you could set the `DOTNET_CLI_ENABLE_PACK_RELEASE_FOR_SOLUTIONS` environment variable to use `PackRelease` in a project that was part of a Visual Studio solution.
+
+## New behavior
+
+If you're developing with the .NET 8 SDK or a later version, `dotnet pack` uses the `Release` configuration by default for *all* projects. If you have a CI/CD script where you've hardcoded `Debug` into an output path, this change may break your workflow. Also, you won't be able to debug a packed app unless the `Debug` configuration was explicitly specified (for example, using `dotnet pack --configuration Debug`.
```
nagilson, line 18:
```suggestion
If you're developing with the .NET 8 SDK or a later version, `dotnet pack` uses the `Release` configuration by default for *all* projects. If you have a CI/CD script, tests, and or code, where you've hardcoded `Debug` into an output path, this change may break your workflow. Also, you won't be able to debug a packed app unless the `Debug` configuration was explicitly specified (for example, using `dotnet pack --configuration Debug`.
```

Consider this change as well for the other document.

nagilson (approved):
Overall, looks great. I added a few suggestions that should be looked at before merging. Thanks 😄

PR #33922, opened 2023-02-08
Clarify TDD process

### LINKED ISSUE #33896: Incorrect text
The text has the following:
"A popular approach in test driven development (TDD) is to write a test before implementing the target code"

This is obviously incorrect.  Writing a failing test before implementing the behaviour is the first step in TDD.  One might argue that it is very, very slightly correct to claim that writing a test first is not a popular approach in TDD because writing a test is always the first step in TDD.  But to claim that it's popular, is obviously false.  And it is specifically writing a failing test that is the first step, not just writing a test.

---
#### Document Details

* ID: e39c2f1b-f41a-bdf9-2646-09379941453d
* Version Independent ID: 422811cb-6287-03af-ba8d-1daebee9744d
* Content: [Unit testing C# code in .NET Core using dotnet test and xUnit - .NET](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-with-dotnet-test)
* Content Source: [docs/core/testing/unit-testing-with-dotnet-test.md](https://github.com/dotnet/docs/blob/main/docs/core/testing/unit-testing-with-dotnet-test.md)
* GitHub Login: @ardalis
* Microsoft Alias: **wiwagn**

ardalis:
Note that tests written via TDD should always be written so that they fail first.

Fixes #33896

### DIFF
--- a/docs/core/testing/unit-testing-with-dotnet-test.md
+++ b/docs/core/testing/unit-testing-with-dotnet-test.md
@@ -121,7 +121,7 @@ Follow the instructions for "Replace the code in *PrimeService.cs* with the foll
 
 ## Create a test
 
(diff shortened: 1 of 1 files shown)


IEvangelist (approved):

PR #33925, opened 2023-02-08
update SDK Docker publish docs to talk about auth, RIDs and misc updates

baronfel:
## Summary

The 7.0.200 release of the SDK coincides with the 0.3.2 release of the container support package, which brings all kinds of improvements to supporting multiple platforms and authenticating to remote registries. I've updated the docs here with descriptions of the new runtime property and how it interacts with the `-r` option, examples of supported registries, and links to other in-repo documentation.

### DIFF
--- a/docs/core/docker/publish-as-container.md
+++ b/docs/core/docker/publish-as-container.md
@@ -1,7 +1,7 @@
 ---
 title: Containerize an app with dotnet publish
 description: In this tutorial, you'll learn how to containerize a .NET application with dotnet publish.
-ms.date: 01/04/2023
+ms.date: 02/08/2023
 ms.topic: tutorial
 ---
 
@@ -119,10 +119,11 @@ Determining projects to restore...
   All projects are up-to-date for restore.
   DotNet.ContainerImage -> .\Worker\bin\Release\net7.0\linux-x64\DotNet.ContainerImage.dll
   DotNet.ContainerImage -> .\Worker\bin\Release\net7.0\linux-x64\publish\
-  Pushed container 'dotnet-worker-image:1.0.0' to registry 'docker://'
+  Building image 'dotnet-worker-image' with tags 1.0.0 on top of base image mcr.microsoft.com/dotnet/aspnet:7.0
+  Pushed container 'dotnet-worker-image:1.0.0' to Docker daemon
 ```
 
-This command compiles your worker app to the *publish- folder and pushes the container to your local docker registry.
+This command compiles your worker app to the *publish* folder and pushes the container to your local docker registry.
 
 ## Configure container image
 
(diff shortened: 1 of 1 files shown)


docs/core/docker/publish-as-container.md
```diff
@@ -147,26 +148,53 @@ The tag of the image is inferred to be the numeric component of your chosen `Tar
...
+```
+
+### `ContainerRuntimeIdentifier`
+
+The container runtime identifier property controls the operating system and architecture used by your container if your [`ContainerBaseImage`](#containerbaseimage) supports more than one platform. For example, the `mcr.microsoft.com/dotnet/runtime` image currently supports `linux-x64`, `linux-arm`, `linux-arm64` and `win10-x64` images all behind the same tag, so the tooling needs a way to be told which of these versions you intend to use.  By default, this will be set to the value of the `RuntimeIdentifier` that you chose when you published the container.  This property rarely needs to be set explicitly - instead use the `-r` option to the `dotnet publish` command.  If the image you've chosen doesn't support the `RuntimeIdentifier` you've chosen, you'll get an error that describes the RuntimeIdentifiers the image does support.
+
+You can always set the `ContainerBaseImage` property to a fully qualified image name, including the tag, to avoid needing to use this property at all.
+
+```xml
+<PropertyGroup>
```
baronfel, line 163: We got some feedback on our in-repo docs that it wasn't clear which knobs were properties and which were items, so we made the decision to 'wrap' them in every example for clarity.

docs/core/docker/publish-as-container.md
```diff
@@ -147,26 +148,53 @@ The tag of the image is inferred to be the numeric component of your chosen `Tar
...
+    <ContainerRuntimeIdentifier>linux-arm64</ContainerRuntimeIdentifier>
+</PropertyGroup>
 ```
 
 ### `ContainerRegistry`
 
-The container registry property controls the destination registry, the place that the newly created image will be pushed to. Be default, it's pushed to the local Docker daemon (`docker://`), but you can also specify a remote registry. For example, consider the following XML example:
+The container registry property controls the destination registry, the place that the newly created image will be pushed to. Be default it's pushed to the local Docker daemon, but you can also specify a remote registry.  When using a remote registry that requires authentication, you authenticate using the well-known `docker login` mechanisms. See [Authenticating to container registries](https://aka.ms/dotnet/containers/auth) for more details. For a concrete example of using this property, consider the following XML example:
```
baronfel, line 170: This URL points to in-repo docs - @IEvangelist should we have dedicated pages here for talking about docker authentication?

IEvangelist: Hi Chet (@baronfel), I think that's a great idea. We should create a new issue to track the work for a new article. For transparency, this would work will likely be prioritized next sprint as there are a lot of work for this sprint already.

IEvangelist (approved):

PR #33948, opened 2023-02-09
Added missing parenthesis to IDE0018 rule sample

KristofferStrube:
## Summary

Added missing parenthesis to IDE0018 rule sample.


gewarren (approved):
Thank you @KristofferStrube

PR #34005, opened 2023-02-14
Update garbage-collector.md

Maoni0:
## Summary

clarifies that GC configs are rarely ever set per machine based on feedback from a customer. also made another clarification on when the configs are read.

### DIFF
--- a/docs/core/runtime-config/garbage-collector.md
+++ b/docs/core/runtime-config/garbage-collector.md
@@ -12,8 +12,8 @@ Settings are arranged into groups on this page. The settings within each group a
 
 > [!NOTE]
 >
-> - These settings can also be changed dynamically by the app as it's running, so any configuration options you set may be overridden.
-> - Some settings, such as [latency level](../../standard/garbage-collection/latency.md), are typically set only through the API at design time. Such settings are omitted from this page.
+> - These configurations are only read by the runtime when the GC is initialize (usually this means during the process startup time). So if you change an environment variable when a  process is already running, that change will not be reflected in that process. Settings that can be changed through APIs during at run time such as [latency level](../../standard/garbage-collection/latency.md), are omitted from this page.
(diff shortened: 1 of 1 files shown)


docs/core/runtime-config/garbage-collector.md
```diff
@@ -12,8 +12,8 @@ Settings are arranged into groups on this page. The settings within each group a
 
 > [!NOTE]
 >
-> - These settings can also be changed dynamically by the app as it's running, so any configuration options you set may be overridden.
-> - Some settings, such as [latency level](../../standard/garbage-collection/latency.md), are typically set only through the API at design time. Such settings are omitted from this page.
+> - These configurations are only read by the runtime when the GC is initialize (usually this means during the process startup time). So if you change an environment variable when a  process is already running, that change will not be reflected in that process. Settings that can be changed through APIs during at run time such as [latency level](../../standard/garbage-collection/latency.md), are omitted from this page.
```
gewarren, line 15:
```suggestion
> - These configurations are only read by the runtime when the GC is initialized (usually this means during the process startup time). If you change an environment variable when a process is already running, the change won't be reflected in that process. Settings that can be changed through APIs at run time, such as [latency level](../../standard/garbage-collection/latency.md), are omitted from this page.
```

docs/core/runtime-config/garbage-collector.md
```diff
@@ -12,8 +12,8 @@ Settings are arranged into groups on this page. The settings within each group a
...
+> - Because GC is per process, it rarely ever makes sense to set these configurations on the machine level. For example, you would not want to set every .NET process to use Server GC, or the same heap hard limit on a machine.
```
gewarren, line 16:
```suggestion
> - Because GC is per process, it rarely ever makes sense to set these configurations at the machine level. For example, you wouldn't want every .NET process on a machine to use server GC or the same heap hard limit.
```

gewarren (approved):

PR #34031, opened 2023-02-14
Use `required` keyword to avoid `null` as default when instantiating record types

### LINKED ISSUE #33851: Suggested code for "standard property syntax" is bad practice?
The suggested code given for a "standard property syntax" record is:

    public record Person
    {
        public string FirstName { get; init; } = default!;
        public string LastName { get; init; } = default!;
    };

This feels like it's encouraging people to do something bad - create a class that you can happily just construct and, with no warning, will introduce `null`-valued non-nullable reference fields into your codebase.

(At first sight I thought it was being clever, that the `default!` would only be evaluated if no value was passed in an object initialiser, and that it would immediately throw a NRE because `default` is null, but apparently that's neither how `!` nor default property values work. [I'm also not the first person to be confused by this](https://github.com/dotnet/docs/issues/28413).)

Absent a null-checker warning for "you've not initialised these properties and by default they are illegal" at `new Person()` constructor sites, would it not be better either for the example to drop the default values and leave the `CS8618` warnings in to warn people or, at worst, to have legal but obviously default values in?

---
#### Document Details

* Content: [Records - C# reference](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BartoszKlonowski:
This pull request fixes #33851

It changes the way the fields and properties are declared within the `record` classes.
The structs in these examples are left intact, because the `double` type will get 0 by default comparing to `null` for strings.

### DIFF
--- a/docs/csharp/language-reference/builtin-types/snippets/shared/RecordType.cs
+++ b/docs/csharp/language-reference/builtin-types/snippets/shared/RecordType.cs
@@ -41,8 +41,8 @@ namespace builtin_types
         // <ImmutableRecord>
         public record Person
         {
-            public string FirstName { get; init; } = default!;
-            public string LastName { get; init; } = default!;
+            public required string FirstName { get; init; }
+            public required string LastName { get; init; }
         };
         // </ImmutableRecord>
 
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks @BartoszKlonowski 

These changes are great. I'll :shipit: now.

PR #34124, opened 2023-02-15
Add breaking change doc for --output when used with solutions

baronfel:
## Summary

We introduced a breaking change to the CLI in 7.0.2000 around the use of `--output` with solution files. This documents the break so users have a source of truth to figure out next steps.

### DIFF
--- a/docs/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid.md
+++ b/docs/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid.md
@@ -1,0 +1,44 @@
+---
+title: "Solution-level `--output` option no longer valid for build-related commands"
+description: Learn about a breaking change in the .NET 7.0.200 SDK where using the `--output` option is no longer valid when using a solution file
+ms.date: 02/15/2023
+---
+# Solution-level `--output` option no longer valid for build-related commands
+
+In the 7.0.200 SDK, there was [a change]([automatic-runtimeidentifier.md](https://github.com/dotnet/sdk/pull/29065)) to no longer accept the `--output`/`-o` option when using a solution file with the following commands:
+
+* build
+* clean
+* pack
+* publish
+* store
+* test
+* vstest
+
+This is because the semantics of the `OutputPath` property, which is controlled by the `--output`/`-o` option, are not well-defined for solutions - projects built in this way will have their output placed in the same directory, which is not consistent and has lead to a number of user-reported issues in the past.
+
+## Version introduced
+
+.NET 7.0.200 SDK
+
+## Previous behavior
+
+Previously, if you specified `--output`/`-o` when using a solution file, the output for all built projects would be placed in the specified directory in an undefined and inconsistent order.
+
+## New behavior
+
+The `dotnet` CLI will error if the `--output`/`-o` option is used with a solution file.
+
+## Type of breaking change
+
+This change requires changes to build scripts and Continuous Integration pipelines.
+
+## Reason for change
+
+This change was taken because the semantics of the `OutputPath` property, which is controlled by the `--output`/`-o` option, are not well-defined for solutions - projects built in this way will have their output placed in the same directory, which is not consistent and has lead to a number of user-reported issues in the past.
+
+For examples of how this presents in practice, see the discussion on [dotnet/sdk#15607](https://github.com/dotnet/sdk/issues/15607).
+
+## Recommended action
+
+The general recommendation is to perform the action that you previously took _without_ the `--output`/`-o` option, and then move the output to the desired location after the command has completed. It is also possible to perform the action at a more specific level (i.e. at a specific project) and still apply the `--output`/`-o` option, as that has more well-defined semantics.
--- a/docs/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid.md
+++ b/docs/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid.md
@@ -15,7 +15,7 @@ In the 7.0.200 SDK, there was [a change](https://github.com/dotnet/sdk/pull/2906
 * test
 * vstest
 
-This is because the semantics of the `OutputPath` property, which is controlled by the `--output`/`-o` option, are not well-defined for solutions - projects built in this way will have their output placed in the same directory, which is not consistent and has lead to a number of user-reported issues in the past.
+This is because the semantics of the `OutputPath` property, which is controlled by the `--output`/`-o` option, are not well-defined for solutions - projects built in this way will have their output placed in the same directory, which is not consistent and has led to a number of user-reported issues in the past.
 
 ## Version introduced
 
@@ -35,10 +35,10 @@ This change requires changes to build scripts and Continuous Integration pipelin
 
 ## Reason for change
 
-This change was taken because the semantics of the `OutputPath` property, which is controlled by the `--output`/`-o` option, are not well-defined for solutions - projects built in this way will have their output placed in the same directory, which is not consistent and has lead to a number of user-reported issues in the past.
+This change was taken because the semantics of the `OutputPath` property, which is controlled by the `--output`/`-o` option, are not well-defined for solutions - projects built in this way will have their output placed in the same directory, which is not consistent and has led to a number of user-reported issues in the past.
 
 For examples of how this presents in practice, see the discussion on [dotnet/sdk#15607](https://github.com/dotnet/sdk/issues/15607).
 
 ## Recommended action
 
-The general recommendation is to perform the action that you previously took _without_ the `--output`/`-o` option, and then move the output to the desired location after the command has completed. It is also possible to perform the action at a more specific level (i.e. at a specific project) and still apply the `--output`/`-o` option, as that has more well-defined semantics.
+The general recommendation is to perform the action that you previously took _without_ the `--output`/`-o` option, and then move the output to the desired location after the command has completed. It is also possible to perform the action at a specific project and still apply the `--output`/`-o` option, as that has more well-defined semantics.
--- a/docs/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid.md
+++ b/docs/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid.md
@@ -31,7 +31,7 @@ The `dotnet` CLI will error if the `--output`/`-o` option is used with a solutio
 
 ## Type of breaking change
 
-This breaking change may require modifications to build scripts and continuous integration pipelines.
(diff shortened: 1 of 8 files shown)


docs/core/tools/dotnet-build.md
```diff
@@ -118,6 +118,10 @@ The project or solution file to build. If a project or solution file isn't speci
 
   Directory in which to place the built binaries. If not specified, the default path is `./bin/<configuration>/<framework>/`.  For projects with multiple target frameworks (via the `TargetFrameworks` property), you also need to define `--framework` when you specify this option.
 
+  - .NET 7.0.200 SDK and later
+
+    If you specify the `--output` option the CLI will emit an error due to the unclear semantics of the output path (as described below).
```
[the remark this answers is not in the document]
baronfel: good spot - the guidance existed for publish, but not the other commands. I made a brief explainer and applied it to all the other commands.

docs/core/tools/dotnet-build.md
```diff
@@ -118,6 +118,10 @@ The project or solution file to build. If a project or solution file isn't speci
 
   Directory in which to place the built binaries. If not specified, the default path is `./bin/<configuration>/<framework>/`.  For projects with multiple target frameworks (via the `TargetFrameworks` property), you also need to define `--framework` when you specify this option.
 
+  - .NET 7.0.200 SDK and later
+
+    If you specify the `--output` option the CLI will emit an error due to the unclear semantics of the output path. All outputs of all built projects will be copied into the specified directory, which isn't compatible with multi-targeted projects, as well as projects that have different versions of direct and transitive dependencies.
```
tdykstra, line 123:
```suggestion
    If you specify the `--output` option when running this command on a solution, the CLI will emit an error due to the unclear semantics of the output path. The `--output` option is disallowed because all outputs of all built projects would be copied into the specified directory, which isn't compatible with multi-targeted projects, as well as projects that have different versions of direct and transitive dependencies.
```

docs/core/tools/dotnet-clean.md
```diff
@@ -52,6 +52,10 @@ The MSBuild project or solution to clean. If a project or solution file is not s
 
   The directory that contains the build artifacts to clean. Specify the `-f|--framework <FRAMEWORK>` switch with the output directory switch if you specified the framework when the project was built.
 
+  - .NET 7.0.200 SDK and later
+
+    If you specify the `--output` option the CLI will emit an error due to the unclear semantics of the output path. All outputs of all built projects will be copied into the specified directory, which isn't compatible with multi-targeted projects, as well as projects that have different versions of direct and transitive dependencies.
```
tdykstra, line 57:
```suggestion
    If you specify the `--output` option when running this command on a solution, the CLI will emit an error due to the unclear semantics of the output path. The `--output` option is disallowed because all outputs of all built projects would be copied into the specified directory, which isn't compatible with multi-targeted projects, as well as projects that have different versions of direct and transitive dependencies.
```

docs/core/tools/dotnet-pack.md
```diff
@@ -97,6 +97,10 @@ You can provide MSBuild properties to the `dotnet pack` command for the packing
 
   Places the built packages in the directory specified.
 
+  - .NET 7.0.200 SDK and later
+
+    If you specify the `--output` option the CLI will emit an error due to the unclear semantics of the output path. All outputs of all built projects will be copied into the specified directory, which isn't compatible with multi-targeted projects, as well as projects that have different versions of direct and transitive dependencies.
```
tdykstra, line 102:
```suggestion
    If you specify the `--output` option when running this command on a solution, the CLI will emit an error due to the unclear semantics of the output path. The `--output` option is disallowed because all outputs of all built projects would be copied into the specified directory, which isn't compatible with multi-targeted projects, as well as projects that have different versions of direct and transitive dependencies.
```

docs/core/tools/dotnet-publish.md
```diff
@@ -155,6 +155,10 @@ For more information, see the following resources:
   <DefaultItemExcludes>$(DefaultItemExcludes);publishoutput**</DefaultItemExcludes>
   ```
 
+  - .NET 7.0.200 SDK and later
+
+    If you specify the `--output` option the CLI will emit an error due to the unclear semantics of the output path (as described below).
```
tdykstra, line 160:
```suggestion
    If you specify the `--output` option when running this command on a solution, the CLI will emit an error due to the unclear semantics of the output path. The `--output` option is disallowed because all outputs of all built projects would be copied into the specified directory, which isn't compatible with multi-targeted projects, as well as projects that have different versions of direct and transitive dependencies.
```

docs/core/tools/dotnet-test.md
```diff
@@ -196,6 +196,10 @@ Where `Microsoft.NET.Test.Sdk` is the test host, `xunit` is the test framework.
 
   Directory in which to find the binaries to run. If not specified, the default path is `./bin/<configuration>/<framework>/`.  For projects with multiple target frameworks (via the `TargetFrameworks` property), you also need to define `--framework` when you specify this option. `dotnet test` always runs tests from the output directory. You can use <xref:System.AppDomain.BaseDirectory%2A?displayProperty=nameWithType> to consume test assets in the output directory.
 
+  - .NET 7.0.200 SDK and later
+
+    If you specify the `--output` option the CLI will emit an error due to the unclear semantics of the output path. All outputs of all built projects will be copied into the specified directory, which isn't compatible with multi-targeted projects, as well as projects that have different versions of direct and transitive dependencies.
```
tdykstra, line 201:
```suggestion
    If you specify the `--output` option when running this command on a solution, the CLI will emit an error due to the unclear semantics of the output path. The `--output` option is disallowed because all outputs of all built projects would be copied into the specified directory, which isn't compatible with multi-targeted projects, as well as projects that have different versions of direct and transitive dependencies.
```

tdykstra (approved):
One more set of suggestions 😄

docs/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid.md
```diff
@@ -0,0 +1,44 @@
+---
```
gewarren, line 1: This should be added to two locations in https://github.com/dotnet/docs/blob/main/docs/core/compatibility/toc.yml.

docs/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid.md
```diff
@@ -0,0 +1,44 @@
...
+title: "Solution-level `--output` option no longer valid for build-related commands"
+description: Learn about a breaking change in the .NET 7.0.200 SDK where using the `--output` option is no longer valid when using a solution file
+ms.date: 02/15/2023
+---
+# Solution-level `--output` option no longer valid for build-related commands
+
+In the 7.0.200 SDK, there was [a change](https://github.com/dotnet/sdk/pull/29065) to no longer accept the `--output`/`-o` option when using a solution file with the following commands:
+
+* build
```
gewarren, line 10: These should probably all be code-fenced so they don't get localized.

docs/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid.md
```diff
@@ -0,0 +1,44 @@
...
+* clean
+* pack
+* publish
+* store
+* test
+* vstest
+
+This is because the semantics of the `OutputPath` property, which is controlled by the `--output`/`-o` option, are not well-defined for solutions - projects built in this way will have their output placed in the same directory, which is not consistent and has led to a number of user-reported issues in the past.
```
gewarren, line 18:
```suggestion
This is because the semantics of the `OutputPath` property, which is controlled by the `--output`/`-o` option, aren't well defined for solutions. Projects built in this way will have their output placed in the same directory, which is inconsistent and has led to a number of user-reported issues.
```

docs/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid.md
```diff
@@ -0,0 +1,44 @@
...
+
+Previously, if you specified `--output`/`-o` when using a solution file, the output for all built projects would be placed in the specified directory in an undefined and inconsistent order.
+
+## New behavior
+
+The `dotnet` CLI will error if the `--output`/`-o` option is used with a solution file.
+
+## Type of breaking change
+
+This change requires changes to build scripts and Continuous Integration pipelines.
```
gewarren, line 34:
```suggestion
This breaking change may require modifications to build scripts and continuous integration pipelines.
```

docs/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Reason for change
+
+This change was taken because the semantics of the `OutputPath` property, which is controlled by the `--output`/`-o` option, are not well-defined for solutions - projects built in this way will have their output placed in the same directory, which is not consistent and has led to a number of user-reported issues in the past.
```
gewarren, line 38:
```suggestion
This change was made because the semantics of the `OutputPath` property, which is controlled by the `--output`/`-o` option, aren't well defined for solutions. Projects built in this way will have their output placed in the same directory, which is inconsistent and has led to a number of user-reported issues.
```

docs/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid.md
```diff
@@ -0,0 +1,44 @@
...
+
+For examples of how this presents in practice, see the discussion on [dotnet/sdk#15607](https://github.com/dotnet/sdk/issues/15607).
+
+## Recommended action
+
+The general recommendation is to perform the action that you previously took _without_ the `--output`/`-o` option, and then move the output to the desired location after the command has completed. It is also possible to perform the action at a specific project and still apply the `--output`/`-o` option, as that has more well-defined semantics.
```
gewarren, line 44:
```suggestion
The general recommendation is to perform the action that you previously took _without_ the `--output`/`-o` option, and then move the output to the desired location after the command has completed. It's also possible to perform the action at a specific project and still apply the `--output`/`-o` option, as that has more well-defined semantics.
```

docs/core/compatibility/sdk/7.0/solution-level-output-no-longer-valid.md
gewarren: Also, based on the boxes you checked in 7.0.md, this should say that the change affects both binary and source compatibility.

baronfel:
Thanks @gewarren - applied the fixes you suggested.

PR #34142, opened 2023-02-16
Remove trailing backslash in default-values.md

ascpixi:
This PR removes a trailing backslash after the "Default constructors" link in the default values documentation.


BillWagner (approved):
Thanks @ascpixi 

This LGTM, and I'll :shipit: as soon as the build finishes.

PR #34144, opened 2023-02-17
Clarify ownership of IDisposable objects

### LINKED ISSUE #33202: Question about using IDisposable
Under section **IDisposable instance members** it's said:
`If a class holds an IDisposable implementation as an instance member, either a field or a property, the class should also implement IDisposable. For more information, see implement a cascade dispose.`
Following the link "[implement a cascade dispose](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#cascade-dispose-calls)" we can read following:
`If your class owns a field or property, and its type implements IDisposable, the containing class itself should also implement IDisposable. A class that instantiates an IDisposable implementation and storing it as an instance member, is also responsible for its cleanup.`

I might be wrong, but this seems to mean that there are 2 cases when class should implement `IDisposable` and call `Dispose` on another object:
- the class has instance field or property holding the object
- the class creates the object

IMHO, in the first case the class should not dispose the object that has been created elsewhere even if the class stores the object in an instance member.

Here is a code demonstrating that disposing the object created outside of the class might break program:

```c#
public static class Program
{
    public static async Task Main()
    {
        using Observer observer = new();

        while (true)
        {
            using Worker worker = new(observer);

            worker.DoWork();

            await Task.Delay(1000);
        }
    }
}

public sealed class Worker : IDisposable
{
    private readonly Observer _observer;

    public Worker(Observer observer)
    {
        _observer = observer;
    }

    public void DoWork()
    {
        Console.WriteLine("Doing some work");
        _observer.Notify();
    }

    #region IDisposable
    private bool _disposed;

    public void Dispose()
    {
        if (_disposed)
        {
            return;
        }

        _observer.Dispose();

        _disposed = true;
    }
    #endregion
}

public sealed class Observer : IDisposable
{
    public void Notify()
    {
        if (_disposed)
        {
            throw new InvalidOperationException("Observer is already disposed");
        }

        Console.WriteLine("Notified");
    }

    #region IDisposable
    private bool _disposed;

    public void Dispose()
    {
        if (_disposed)
        {
            return;
        }

        _disposed = true;
    }
    #endregion
}

/*
Output:
Doing some work
Notified
Doing some work
Unhandled exception. System.InvalidOperationException: Observer is already disposed
   at Observer.Notify() in .../Program.cs:line 56
   at Worker.DoWork() in .../Program.cs:line 30
   at Program.Main() in .../Program.cs:line 11
   at Program.<Main>()
*/

```

This code crashes on the second loop step, because the first created instance of `Worker` class disposed `Observer` object (as it holds instance member of `Observer`).

Am I missing something or have I misunderstood the documentation?

---
#### Document Details

* ID: b83e20ec-96de-0340-0813-27968ae4ca89
* Version Independent ID: db8ef20b-08da-cbac-0d96-91dcef8abe56
* Content: [Using objects that implement IDisposable](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/using-objects#idisposable-instance-members)
* Content Source: [docs/standard/garbage-collection/using-objects.md](https://github.com/dotnet/docs/blob/main/docs/standard/garbage-collection/using-objects.md)
* GitHub Login: @gewarren

gewarren:
Fixes #33202 

[Preview](https://review.learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose?branch=pr-en-us-34144#cascade-dispose-calls)

### DIFF
--- a/docs/standard/garbage-collection/implementing-dispose.md
+++ b/docs/standard/garbage-collection/implementing-dispose.md
@@ -1,7 +1,7 @@
 ---
 title: "Implement a Dispose method"
 description: In this article, learn to implement the Dispose method, which releases unmanaged resources used by your code in .NET.
-ms.date: 09/29/2021
+ms.date: 02/16/2023
 dev_langs:
   - "csharp"
   - "vb"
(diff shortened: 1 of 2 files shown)


docs/standard/garbage-collection/implementing-dispose.md
```diff
@@ -81,24 +83,26 @@ If the method call comes from a finalizer, only the code that frees unmanaged re
...
 
 :::code language="csharp" source="../../../samples/snippets/csharp/VS_Snippets_CLR/conceptual.disposable/cs/Foo.cs":::
 :::code language="vb" source="../../../samples/snippets/visualbasic/VS_Snippets_CLR/conceptual.disposable/vb/Foo.vb":::
 
 > [!TIP]
-> There are cases when you may want to perform `null`-checking in a finalizer (which includes the `Dispose(false)` method invoked by a finalizer), one of the primary reasons is if you're unsure whether the instance got fully initialized (for example, an exception may be thrown in a constructor).
+>
+> - If your class has an <xref:System.IDisposable> field or property but doesn't *own* it, meaning the class doesn't create the object, then the class doesn't need to implement <xref:System.IDisposable>.
```
gewarren, line 93: This is where I added the clarification.

IEvangelist (approved):

PR #34179, opened 2023-02-20
Update fundamentals.md

ramronn:
hi,

i'm new here and i think i found out some mistake in documentation at row 61 where it states that garbage collector allocates memory but it should be the runtime right?

great documentation otherwise :)

happy to be part of .NET community

Ronald

## Summary

Describe your changes here.


IEvangelist:
Tagging @Maoni0

Maoni0:
I think it's fine. the paragraph describes the idea of a bump pointer allocator but of course that's not what actually happens (nobody does this simple thing). but there's benefits to describe the idea without getting into the details. so it's ok to make these changes. it is actually the GC that reserves the space, but it's not the GC that allocates objects in the memory handed out most of the time. but the GC is part of the runtime so it's fine to just use "runtime" to cover all here.

gewarren (approved):

PR #34257, opened 2023-02-24
Add support for automated clean repo tool execution

IEvangelist:
## Summary

Adds monthly calls to the Clean Repo tool.

Related to https://github.com/dotnet/docs-tools/issues/121 and https://github.com/dotnet/docs-tools/pull/130

### DIFF
--- a/.github/workflows/clean-repo.yml
+++ b/.github/workflows/clean-repo.yml
@@ -36,6 +36,7 @@ jobs:
       - name: create-pull-request
         uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04
         with:
+          branch: create-cleanrepo-pull-request/patch
           title: "Monthly chores: automated clean repo tool"
           commit-message: 'Bot 🤖 generated "Clean repo tool run"'
           body: "Automated creation of Clean repo updates."
--- a/.github/workflows/clean-repo.yml
+++ b/.github/workflows/clean-repo.yml
@@ -17,7 +17,7 @@ jobs:
       # Call clean repo
       - name: Clean repo
         id: clean-repo-step
-        uses: docs-tools/cleanrepo/CleanRepo@main
+        uses: dotnet/docs-tools/cleanrepo/CleanRepo@main
         with:
           start-directory: '.'
           docset-root: './docs'
--- a/.github/workflows/clean-repo.yml
+++ b/.github/workflows/clean-repo.yml
@@ -23,6 +23,7 @@ jobs:
           start-directory: '.'
           docset-root: './docs'
           repo-root: '.'
+          base-path: '/dotnet'
           delete: true
           orphaned-topics: true
           orphaned-images: true
--- a/.github/workflows/clean-repo.yml
+++ b/.github/workflows/clean-repo.yml
@@ -27,7 +27,7 @@ jobs:
           delete: true
           orphaned_topics: true
           orphaned_images: true
-          orphaned_snippet: true
+          orphaned_snippets: true
           orphaned_includes: true
           replace_redirects: true
           relative_links: true
--- a/.github/workflows/clean-repo.yml
+++ b/.github/workflows/clean-repo.yml
@@ -8,6 +8,7 @@ jobs:
   clean-repo:
     runs-on: ubuntu-latest
     permissions:
+      contents: write
       issues: write
       pull-requests: write
 
--- a/.github/workflows/clean-repo.yml
+++ b/.github/workflows/clean-repo.yml
@@ -20,18 +20,18 @@ jobs:
         id: clean-repo-step
         uses: dotnet/docs-tools/cleanrepo/CleanRepo@main
         with:
-          start-directory: '.'
-          docset-root: './docs'
-          repo-root: '.'
-          base-path: '/dotnet'
+          start_directory: '.'
+          docset_root: './docs'
+          repo_root: '.'
+          base_path: '/dotnet'
           delete: true
-          orphaned-topics: true
-          orphaned-images: true
-          orphaned-snippet: true
-          orphaned-includes: true
-          replace-redirects: true
-          relative-links: true
-          remove-hops: true
+          orphaned_topics: true
+          orphaned_images: true
+          orphaned_snippet: true
+          orphaned_includes: true
+          replace_redirects: true
+          relative_links: true
+          remove_hops: true
         
       # Create the PR for the work done by the "clean repo" tool
       - name: create-pull-request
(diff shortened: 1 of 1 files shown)


.github/workflows/clean-repo.yml
```diff
@@ -0,0 +1,40 @@
+name: "clean repo"
+
+on:
+  schedule:
+    - cron: "0 1 1 * *" # Runs at 01:00, on day 1 of the month
```
BillWagner, line 5: Should we pick a different day so this runs on a different day than the What's new action?

BillWagner (commented):
This is awesome!

commenting so @gewarren can take a look when she returns.

.github/workflows/clean-repo.yml
IEvangelist: The .NET Version Sweeper was the first tool in our repo to be setup as a CRON job. I bumped it up an hour from that time. I think the What's new tool should be bumped another hour.

.github/workflows/clean-repo.yml
```diff
@@ -0,0 +1,40 @@
...
+      issues: write
+      pull-requests: write
+
+    steps:
+      - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
+
+      # Call clean repo
+      - name: Clean repo
+        id: clean-repo-step
+        uses: docs-tools/cleanrepo/CleanRepo@main
```
IEvangelist, line 20:
```suggestion
        uses: dotnet/docs-tools/cleanrepo/CleanRepo@main
```

.github/workflows/clean-repo.yml
```diff
@@ -0,0 +1,40 @@
...
+          orphaned-includes: true
+          replace-redirects: true
+          relative-links: true
+          remove-hops: true
+        
+      # Create the PR for the work done by the "clean repo" tool
+      - name: create-pull-request
+        uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04
+        with:
+          title: "Monthly chores: automated clean repo tool"
```
BillWagner, line 38: based on what we learned with the what's new tool, let's declare a unique branch here:

```suggestion
          branch: create-cleanrepo-pull-request/patch
          title: "Monthly chores: automated clean repo tool"
```

.github/workflows/clean-repo.yml
```diff
@@ -0,0 +1,40 @@
...
+
+on:
+  schedule:
+    - cron: "0 1 1 * *" # Runs at 01:00, on day 1 of the month
+
+jobs:
+  clean-repo:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
```
IEvangelist, line 11:
```suggestion
      contents: write
      issues: write
```

.github/workflows/clean-repo.yml
```diff
@@ -0,0 +1,42 @@
...
+      # Call clean repo
+      - name: Clean repo
+        id: clean-repo-step
+        uses: dotnet/docs-tools/cleanrepo/CleanRepo@main
+        with:
+          start-directory: '.'
+          docset-root: './docs'
+          repo-root: '.'
+          delete: true
+          orphaned-topics: true
```
gewarren, line 27: @IEvangelist Does it call the tool once with all of these arguments? If so, we'll need probably need to call it separately for each function, or I'll have to change the logic in the tool itself to call multiple functions successively.

IEvangelist: Yeah, whenever the action workflow is invoked (one time per month), it wakes up and calls your CleanRepo.dll once with all these switches. If you need help updating the app to accept multiple switches, I could help, that might be a more straightforward update. Otherwise, we'd end up re-creating the container for each call which is resource intensive.

BillWagner: We could add a bash script to execute the following:
1. Build cleanrepo
2. iterate and call dotnet run with each of the switches in turn

IEvangelist: I'm not sure how that would work in this scenario and calling a bash script (while possible) differs from calling the CleanRepo GitHub Action. The CleanRepo tool is now exposed as a GitHub Action, via the _Dockerfile_. The image is built on demand, and the container is executed through the _action.yml_ - where does the bash script come into play?

BillWagner: I didn't explain what I was thinking well enough.

Suppose the entry point in the action was the shell script. 

The first seven lines of the [Dockerfile](https://github.com/dotnet/docs-tools/blob/main/cleanrepo/CleanRepo/Dockerfile#L1-L7) would remain the same.

The remainder would be something like:

```txt
FROM mcr.microsoft.com/dotnet/sdk:7.0
COPY --from=build-env /app/out .
# Assume the build environment has a shell script named `clean-repo.sh`
ENTRYPOINT [ "/bin/sh", "/app/clean-repo.sh" ]
```

The `clean-repo.sh` script runs `dotnet /CleanRepo.dll` repeatedly with the different arguments.

IEvangelist: Ah, ok. That makes more sense now. Thank you for the added context. Let's ask @gewarren for her thoughts and preferences on this. I'm happy to try to update this when we hear back from her.

.github/workflows/clean-repo.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      # Call clean repo
+      - name: Clean repo
+        id: clean-repo-step
+        uses: dotnet/docs-tools/cleanrepo/CleanRepo@main
+        with:
+          start-directory: '.'
+          docset-root: './docs'
+          repo-root: '.'
+          delete: true
```
gewarren, line 26:
```suggestion
          base-path: '/dotnet'
          delete: true
```

.github/workflows/clean-repo.yml
gewarren: Yes, let's try to loop through here rather than in the app itself.

gewarren:
This PR needs to be merged before this automation will work properly: https://github.com/dotnet/docs-tools/pull/128

.github/workflows/clean-repo.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: dotnet/docs-tools/cleanrepo/CleanRepo@main
+        with:
+          start_directory: '.'
+          docset_root: './docs'
+          repo_root: '.'
+          base_path: '/dotnet'
+          delete: true
+          orphaned_topics: true
+          orphaned_images: true
+          orphaned_snippet: true
```
gewarren, line 30:
```suggestion
          orphaned_snippets: true
```

BillWagner (approved):
This LGTM. I'm excited to see this happen.

gewarren (approved):

PR #34258, opened 2023-02-24
Update what's new action to use the updated workflow

BillWagner:
The action yml file now uses the action configured in the docs-tools repository instead of the nuget package installed in this repo.


.github/workflows/whats-new.yml
```diff
@@ -14,53 +10,29 @@ on:
...
-    # The type of runner that the job will run on
     runs-on: ubuntu-latest
     permissions:
-      contents: write
       pull-requests: write
 
-    # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
```
IEvangelist, line 20:
```suggestion
```

This is no longer needed.

IEvangelist (commented):
Looks good! I think we can remove the checkout, shouldn't be needed.

IEvangelist (approved):

PR #34264, opened 2023-02-24
restore contents: write permissions

BillWagner:
add back write perms, and create a unique name for the dependabot config action


IEvangelist (approved):

IEvangelist:
Sure, let's try this :)

PR #34272, opened 2023-02-25
Fix spaces around colons in signatures and type annotations in F# style guide

dawedawe:
## Summary

As far as I understand the style guide, the space-wrapped colon (` : `) should only be used for return type annotations of functions and members, not for val signatures or normal type annotations.


BillWagner:
Thanks @dawedawe 

Checking with @dsyme before merging.

dsyme (approved):

PR #34274, opened 2023-02-25
Describe the example of user-defined types used as fields declarations of anonymous types

### LINKED ISSUE #33918: missing information
[Enter feedback here]

AnotherObject x = new AnotherObject();
x.Field3 = "zzz";
new {field1 = "xxx",field2 = "yyy", x}
What happens now? Why dont you add this info to this page? Where do i find this info otherwise?

---
#### Document Details

* ID: d1b30ebe-b640-449b-3602-9891a55794db
* Version Independent ID: dd937177-98f0-a797-d49b-29332aae0614
* Content: [Anonymous Types](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/types/anonymous-types)
* Content Source: [docs/csharp/fundamentals/types/anonymous-types.md](https://github.com/dotnet/docs/blob/main/docs/csharp/fundamentals/types/anonymous-types.md)
* Product: **dotnet-csharp**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

BartoszKlonowski:
This pull request fixes #33918 
It adds the explanation of what happens when a variable holding object of user-defined type is used as a declaration of anonymous types field.

The code snippet presents both class case as well as another anonymous type case used.
The `product` var is of type `Product` which is already defined in the `anonymous_types` namespace, but it is not mentioned in the additional text, as the same assumption about `Product` type already existing is already made in the same article.

**NOTE:** I'm not sure about the linking of `ID="snippet03"` and `<Snippet03>` when it comes to capital letter - I can see that the same case is for `ID="snippet02"` linking correctly with `<Snippet02>` so I decided to stick with that approach, but I'll be super happy for reviewers to correct me if there might be any issue with linking. 🙏

### DIFF
--- a/docs/csharp/fundamentals/types/anonymous-types.md
+++ b/docs/csharp/fundamentals/types/anonymous-types.md
@@ -37,6 +37,10 @@ If you don't specify member names in the anonymous type, the compiler gives the 
 > [!TIP]
 > You can use .NET style rule [IDE0037](../../../fundamentals/code-analysis/style-rules/ide0037.md) to enforce whether inferred or explicit member names are preferred.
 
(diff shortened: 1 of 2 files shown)


BillWagner (approved):
Thanks for making this addition @BartoszKlonowski 

This looks great, and the snippets render correctly.

I'll :shipit: now.

PR #34284, opened 2023-02-27
Fix json sample

WeihanLi:
## Summary

Fixes the json sample, it caused an error likes 

```
Member names cannot be the same as their enclosing type
```


gewarren (commented):
Thanks @WeihanLi, good catch. Can you also fix line 51 to match?

WeihanLi:
@gewarren updated thanks

gewarren (approved):

PR #34300, opened 2023-02-28
reviewed for FY23 freshness

alexwolfmsft:
## Summary

Describe your changes here.

### DIFF
--- a/docfx.json
+++ b/docfx.json
@@ -387,7 +387,7 @@
                 "docs/architecture/**/**.md": "nanil",
                 "docs/architecture/grpc-for-wcf-developers/**/**.md": "jamesnk",
                 "docs/architecture/modernize-desktop/**/**.md": "oliag",
-                "docs/azure/**/*.*": "casoper",
+                "docs/azure/**/*.*": "alexwolf",
                 "docs/orleans/**/*.*": "dapine",
                 "docs/core/**/**.md": "dotnetcontent",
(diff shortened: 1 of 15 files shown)


CamSoper (commented):
Looks fine, but we should check one thing...

@IEvangelist -- I thought there was a place to specify the author for an entire folder. I can't seem to find it. Am I hallucinating?

IEvangelist:
> Looks fine, but we should check one thing...
> 
> @IEvangelist -- I thought there was a place to specify the author for an entire folder. I can't seem to find it. Am I hallucinating?

You're not hallucinating, it can be done with the _docfx.json_ file, [see `fileMetadata`](https://github.com/dotnet/docs/blob/main/docfx.json#L172). You can specify each desired metadata value there, in fact Alex is already set as [`author`](https://github.com/dotnet/docs/blob/main/docfx.json#L295). Using the Metadata Explorer that @gewarren and I wrote, you can see this value in place.

image

Ideally, we'd remove the explicit `author` and `ms.author` values from each article and declaratively update the _docfx.json_.

alexwolfmsft:
Thanks @IEvangelist I remember this now from a long time ago. I removed all of the individual file updates for the author and adjusted the docfx file.

CamSoper (approved):
Thank you, @IEvangelist! 

Looks good to me, @alexwolfmsft!

Issue #34310, opened 2023-03-01
`unmanaged` types should include `nint` and `nuint`


just-ero:
The "[Unmanaged types (C# reference)](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/unmanaged-types)" page currently does not include the `unmanaged` native types `nint` and `nuint`. They are blittable, `unmanaged` types and should therefore be included.

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 93e5d75f-07a3-e7ee-2842-ac754bdd2e41
* Version Independent ID: 4a3622fb-7d09-78ec-ba38-b88001475018
* Content: [Unmanaged types - C# reference](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/unmanaged-types)
* Content Source: [docs/csharp/language-reference/builtin-types/unmanaged-types.md](https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/builtin-types/unmanaged-types.md)
* Product: **dotnet-csharp**
* Technology: **csharp-language-reference**
* GitHub Login: @BillWagner
* Microsoft Alias: **wiwagn**

PR #34311, opened 2023-03-01
Add `nint` and `nuint` as `unmanaged` types (#34310)

ero-qt:
## Summary

This PR adds the blittable native types `nint` and `nuint` to the "[Unmanaged types (C# reference)](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/unmanaged-types)" page.

Issue: https://github.com/dotnet/docs/issues/34310

### DIFF
--- a/docs/csharp/language-reference/builtin-types/unmanaged-types.md
+++ b/docs/csharp/language-reference/builtin-types/unmanaged-types.md
@@ -9,7 +9,7 @@ helpviewer_keywords:
 
 A type is an **unmanaged type** if it's any of the following types:
 
-- `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, or `bool`
+- `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `nint`, `nuint`, `char`, `float`, `double`, `decimal`, or `bool`
 - Any [enum](enum.md) type
 - Any [pointer](../unsafe-code.md#pointer-types) type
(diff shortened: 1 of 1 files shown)


tannergooding (approved):
LGTM

These were previously covered (from the C# perspective) under
> Any user-defined [struct](struct.md) type that contains fields of unmanaged types only.

However, now that they are keywords adding them to the explicit keyword list makes sense to me.

just-ero:
I didn't think about this before, but should I order them differently, maybe?
```diff
- `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool`, `nint`, or `nuint`
+ `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `nint`, `nuint`, `char`, `float`, `double`, `decimal`, or `bool`
```

tannergooding:
Don't think the order matters too terribly much. But someone else may have a different opinion.

BillWagner:
> I didn't think about this before, but should I order them differently, maybe?
> 
> ```diff
> - `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool`, `nint`, or `nuint`
> + `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `nint`, `nuint`, `char`, `float`, `double`, `decimal`, or `bool`
> ```

I like this order better. The `nint` and `nuint` are integral types, so it makes sense to locate them together. I'll merge as soon as you make that update.

BillWagner (approved):
Thanks again @just-ero 

PR #34361, opened 2023-03-03
Added missing double dashes in front of parameter.

LennartCode:
Now, the command can be copied to the console as is.

## Summary

Added missing double dashes in front of the `list` parameter to allow copying the command to the console.


LennartCode:
@dotnet-policy-service agree

tdykstra (commented):
See the Note in the purple box at the end of the Description section of the [`dotnet new list`](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-new-list) article -- In SDK 7.0 the `list` option became a subcommand.

PR #34362, opened 2023-03-03
Use the Quest cron job

BillWagner:
This does a couple things.

First, run the quest bulk action every night, at a time when it's likely that none of us are working. Second, run the single issue quest *only* as a workflow_dispatch to run when needed.

### DIFF
--- a/.github/workflows/quest-bulk.yml
+++ b/.github/workflows/quest-bulk.yml
@@ -1,5 +1,7 @@
 name: "bulk quest import"
 on:
+  schedule:
+    - cron: '0 10 * * *' # UTC time, that's 5:00 am EST, 2:00 am PTC.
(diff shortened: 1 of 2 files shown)


IEvangelist (approved):

.github/workflows/quest-bulk.yml
```diff
@@ -1,5 +1,7 @@
 name: "bulk quest import"
 on:
+  schedule:
+    - cron: '0 10 * * *' # UTC time, that's 5:00 am EST, 2:00 am PTC.
```
gewarren, line 4:
```suggestion
    - cron: '0 10 * * *' # UTC time, that's 5:00 am EST, 2:00 am PST.
```

?

gewarren (approved):

PR #34391, opened 2023-03-06
Update worker snippets to use `Host.CreateApplicationBuilder`

### LINKED ISSUE #34247: Update worker service documentation to use Host.CreateApplicationBuilder
We're planning on updating the worker template to start using `Host.CreateApplicationBuilder(args)` in .NET 7 (https://github.com/dotnet/aspnetcore/issues/43113).

There's some discussion around whether need to add an `IHostBuilder` adapter for legacy third-party extension methods targeting it, but this looks like something that's going to happen. Even though we're not obsoleting `Host.CreateDefaultBuilder`, we should update the docs to align with our templates.

So, for instance, this:

https://github.com/dotnet/docs/blob/bbaf869e8bee41cd121be7b87b0df512883c1644/docs/core/extensions/workers.md?plain=1#L61-L72

Would become this:

```csharp
using App.WorkerService;

HostApplicationBuilder builder = Host.CreateApplicationBuilder(args);
builder.Services.AddHostedService<Worker>();

IHost host = builder.Build();
host.Run();
```

@Rick-Anderson should we open a similar issue in https://github.com/dotnet/AspNetCore.Docs? There seems to be a lot of overlap.

@eerhardt

IEvangelist:
## Summary

Update worker snippets to use `Host.CreateApplicationBuilder`

Fixes #34247

Previews:

- [Deploy a Worker Service to Azure](https://review.learn.microsoft.com/en-us/dotnet/core/extensions/cloud-service?branch=pr-en-us-34391&pivots=visualstudio)
- [Create a Queue Service](https://review.learn.microsoft.com/en-us/dotnet/core/extensions/queue-service?branch=pr-en-us-34391)
- [Use scoped services within a `BackgroundService`](https://review.learn.microsoft.com/en-us/dotnet/core/extensions/scoped-service?branch=pr-en-us-34391)
- [Implement the `IHostedService` interface](https://review.learn.microsoft.com/en-us/dotnet/core/extensions/timer-service?branch=pr-en-us-34391)
- [Create a Windows Service using `BackgroundService`](https://review.learn.microsoft.com/en-us/dotnet/core/extensions/windows-service?branch=pr-en-us-34391)
- [Worker Services in .NET](https://review.learn.microsoft.com/en-us/dotnet/core/extensions/workers?branch=pr-en-us-34391)

### DIFF
--- a/docs/core/extensions/snippets/workers/windows-service/Program.cs
+++ b/docs/core/extensions/snippets/workers/windows-service/Program.cs
@@ -2,25 +2,21 @@
 using Microsoft.Extensions.Logging.Configuration;
 using Microsoft.Extensions.Logging.EventLog;
 
-using IHost host = Host.CreateDefaultBuilder(args)
-    .UseWindowsService(options =>
-    {
-        options.ServiceName = ".NET Joke Service";
-    })
-    .ConfigureServices(services =>
-    {
-        LoggerProviderOptions.RegisterProviderOptions<
-            EventLogSettings, EventLogLoggerProvider>(services);
+HostApplicationBuilder builder = Host.CreateApplicationBuilder(args);
+builder.Services.AddWindowsService(options =>
+{
+    options.ServiceName = ".NET Joke Service";
+});
 
-        services.AddSingleton<JokeService>();
-        services.AddHostedService<WindowsBackgroundService>();
-    })
-    .ConfigureLogging((context, logging) =>
-    {
-        // See: https://github.com/dotnet/runtime/issues/47303
-        logging.AddConfiguration(
-            context.Configuration.GetSection("Logging"));
-    })
-    .Build();
+LoggerProviderOptions.RegisterProviderOptions<
+    EventLogSettings, EventLogLoggerProvider>(builder.Services);
 
-await host.RunAsync();
+builder.Services.AddSingleton<JokeService>();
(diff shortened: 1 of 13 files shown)


docs/core/extensions/snippets/workers/windows-service/Program.cs
```diff
@@ -2,25 +2,21 @@
...
-    })
-    .ConfigureServices(services =>
-    {
-        LoggerProviderOptions.RegisterProviderOptions<
-            EventLogSettings, EventLogLoggerProvider>(services);
+HostApplicationBuilder builder = Host.CreateApplicationBuilder(args);
+builder.Services.AddWindowsService(options =>
+{
+    options.ServiceName = ".NET Joke Service";
+});
```
BillWagner, line 9: I don't think the brackets are necessary here:

```suggestion
builder.Services.AddWindowsService(options => options.ServiceName = ".NET Joke Service";);
```

BillWagner (approved):
This LGTM. I had one optional suggestion, if you want.

docs/core/extensions/snippets/workers/windows-service/Program.cs
IEvangelist: I think for readability, I'm going to leave this as is. Also, there is specific line highlighting that is in the markdown file too that would also have to be updated.

PR #34394, opened 2023-03-06
Touchups to .NET Fx version article

gewarren:
Follow-up to #34383.


IEvangelist (approved):

PR #34399, opened 2023-03-06
Update index.yml

Rick-Anderson:
See https://github.com/dotnet/AspNetCore.Docs/issues/28562

## Summary

Describe your changes here.

### DIFF
--- a/docs/navigate/migration-guide/index.yml
+++ b/docs/navigate/migration-guide/index.yml
@@ -1,18 +1,18 @@
 ### YamlMime:Landing
 
-title: .NET migration documentation
-summary: Learn about migrating apps from .NET Framework to .NET.
+title: .NET upgrade documentation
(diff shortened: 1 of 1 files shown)


gewarren (commented):
Thanks. I feel like this should be part of a larger PR that updates all the affected files, e.g. https://github.com/dotnet/docs/blob/main/docs/navigate/migration-guide/toc.yml, https://github.com/dotnet/docs/blob/main/docs/index.yml, and any affected articles under /docs/core/porting. Also perhaps the docs/navigate/migration-guide directory should be renamed (it only contains two files so redirecting isn't a big deal).

cc @StephenBonikowsky

Rick-Anderson:
@gewarren but this band-aid is better than nothing. Right now the migration docs are a real mess and we should at least get started. See also #34545

PR #34445, opened 2023-03-07
Update single-file docs

agocke:
Two changes:
 - Remove docs about extra steps for debugging. The latest VS releases no longer need any extra steps to debug single-file binaries.
 - Add docs about post-build processing. This is a public extension point that can be used to support things like signing.

#### Internal previews

|:--|:--|
| _docs/core/deploying/single-file/overview.md_ | [Preview: docs/core/deploying/single-file/overview](https://review.learn.microsoft.com/en-us/dotnet/core/deploying/single-file/overview?branch=pr-en-us-34445) |

### DIFF
--- a/docs/core/deploying/single-file/overview.md
+++ b/docs/core/deploying/single-file/overview.md
@@ -156,25 +156,25 @@ Managed C++ components aren't well suited for single file deployment. We recomme
 
 ### Native libraries
 
-Only managed DLLs are bundled with the app into a single executable. When the app starts, the managed DLLs are extracted and loaded in memory, avoiding the extraction to a folder. This approach means that the managed binaries are embedded in the single file bundle, but the native binaries of the core runtime itself are separate files.
+Only managed DLLs are bundled with the app into a single executable. When the app starts, the managed DLLs are extracted and loaded in memory, avoiding the extraction to a folder. With this approach, the managed binaries are embedded in the single file bundle, but the native binaries of the core runtime itself are separate files.
 
-To embed those files for extraction and get one output file, set the property `IncludeNativeLibrariesForSelfExtract` to `true`. For more information about extraction, see [Including native libraries](#include-native-libraries).
+To embed those files for extraction and get one output file, set the property `IncludeNativeLibrariesForSelfExtract` to `true`.
 
 Specifying `IncludeAllContentForSelfExtract` extracts all files, including the managed assemblies, before running the executable. This may be helpful for rare application compatibility problems.
 
- [!NOTE]
+ [!IMPORTANT]
 > If extraction is used, the files are extracted to disk before the app starts:
 >
-> - If environment variable `DOTNET_BUNDLE_EXTRACT_BASE_DIR` is set to a path, the files are extracted to a directory under that path.
-> - Otherwise, if running on Linux or MacOS, the files are extracted to a directory under `$HOME/.net`.
+> - If the `DOTNET_BUNDLE_EXTRACT_BASE_DIR` environment variable is set to a path, the files are extracted to a directory under that path.
+> - Otherwise, if running on Linux or macOS, the files are extracted to a directory under `$HOME/.net`.
 > - If running on Windows, the files are extracted to a directory under `%TEMP%/.net`.
 >
-> To prevent tampering, these directories should not be writable by users or services with different privileges. Don't use _/tmp_ or _/var/tmp_ on most Linux and MacOS systems.
+> To prevent tampering, these directories shouldn't be writable by users or services with different privileges. Don't use _/tmp_ or _/var/tmp_ on most Linux and macOS systems.
 
 > [!NOTE]
-> In some Linux environments, such as under `systemd`, the default extraction does not work because `$HOME` is not defined. In such cases, we recommend that you set `$DOTNET_BUNDLE_EXTRACT_BASE_DIR` explicitly.
+> In some Linux environments, such as under `systemd`, the default extraction doesn't work because `$HOME` isn't defined. In such cases, it's recommended that you set `$DOTNET_BUNDLE_EXTRACT_BASE_DIR` explicitly.
 >
-> For `systemd`, a good alternative seems to be defining `DOTNET_BUNDLE_EXTRACT_BASE_DIR` in your service's unit file as `%h/.net`, which `systemd` expands correctly to `$HOME/.net` for the account running the service.
+> For `systemd`, a good alternative is to define `DOTNET_BUNDLE_EXTRACT_BASE_DIR` in your service's unit file as `%h/.net`, which `systemd` expands correctly to `$HOME/.net` for the account running the service.
 >
 > ```text
 > [Service]
@@ -213,23 +213,23 @@ We have some recommendations for fixing common scenarios:
 
 Some workflows require post-processing of binaries before bundling. A common example is signing. The dotnet SDK provides MSBuild extension points to allow processing binaries just before single-file bundling. The available APIs are:
 
-* A target `PrepareForBundle` that will be called before `GenerateSingleFileBundle`
-* An ItemGroup `FilesToBundle` containing all files that will be bundled
-* A Property `AppHostFile` that will specify the apphost template. Post-processing might want to exclude the apphost from processing.
+- A target `PrepareForBundle` that will be called before `GenerateSingleFileBundle`
+- An `<ItemGroup><FilesToBundle /></ItemGroup>` containing all files that will be bundled
+- A Property `AppHostFile` that will specify the apphost template. Post-processing might want to exclude the apphost from processing.
 
-The way to plug into this would typically involve creating a target that will be executed between PrepareForBundle and GenerateSingleFileBundle
+To plug into this involves creating a target that will be executed between `PrepareForBundle` and `GenerateSingleFileBundle`.
 
-Example:
+Consider the following .NET project `Target` node example:
 
 ```xml
 <Target Name="MySignedBundledFile" BeforeTargets="GenerateSingleFileBundle" DependsOnTargets="PrepareForBundle">
 ```
 
-It is possible that tooling will need to copy files in the process of signing. That could happen if the original file is a shared item not owned by the build, for example the file comes from a nuget cache. In such case it is expected that the tool will modify the path of the corresponding FilesToBundle item to point to the modified copy.
+It's possible that tooling will need to copy files in the process of signing. That could happen if the original file is a shared item not owned by the build, for example, the file comes from a NuGet cache. In such a case, it's expected that the tool will modify the path of the corresponding `FilesToBundle` item to point to the modified copy.
 
 ### Compress assemblies in single-file apps
 
-Single file apps can be created with compression enabled on the embedded assemblies. Set the `EnableCompressionInSingleFile` property to `true`. The single file that's produced will have all of the embedded assemblies compressed, which can significantly reduce the size of the executable.
+Single-file apps can be created with compression enabled on the embedded assemblies. Set the `EnableCompressionInSingleFile` property to `true`. The single file that's produced will have all of the embedded assemblies compressed, which can significantly reduce the size of the executable.
 
 Compression comes with a performance cost. On application start, the assemblies must be decompressed into memory, which takes some time. We recommend that you measure both the size change and startup cost of enabling compression before using it. The impact can vary significantly between different applications.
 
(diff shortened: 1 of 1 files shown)


agocke:
@vsadov for review

docs/core/deploying/single-file/overview.md
```diff
@@ -193,48 +209,27 @@ We have some recommendations for fixing common scenarios:
...
+### Post-processing binaries before bundling
 
-Without this file, Visual Studio might produce the error: "Unable to attach to the process. A debug component is not installed." VS Code might produce the error: "Failed to attach to process: Unknown Error: 0x80131c3c."
+Some workflows require post-processing of binaries before bundling. A common example is signing. The dotnet SDK provides MSBuild extension points to allow processing binaries just before single-file bundling. The available APIs are:
 
-To fix these errors, _mscordbi_ needs to be copied next to the executable. _mscordbi_ is `publish`ed by default in the subdirectory with the application's runtime ID. So, for example, if you publish a self-contained single file executable using the `dotnet` CLI for Windows using the parameters `-r win-x64`, the executable would be placed in _bin/Debug/net5.0/win-x64/publish_. A copy of _mscordbi.dll_ would be present in _bin/Debug/net5.0/win-x64_.
+* A target `PrepareForBundle` that will be called before `GenerateSingleFileBundle`
+* An ItemGroup `FilesToBundle` containing all files that will be bundled
```
IEvangelist, line 217:
```suggestion
- An `<ItemGroup><FilesToBundle /></ItemGroup>` containing all files that will be bundled
```

docs/core/deploying/single-file/overview.md
```diff
@@ -158,13 +154,33 @@ Single file applications have all related PDB files alongside the application, n
...
+To embed those files for extraction and get one output file, set the property `IncludeNativeLibrariesForSelfExtract` to `true`. For more information about extraction, see [Including native libraries](#include-native-libraries).
 
-Starting with .NET 5, only managed DLLs are bundled with the app into a single executable. When the app starts, the managed DLLs are extracted and loaded in memory, avoiding the extraction to a folder. On Windows, this approach means that the managed binaries are embedded in the single file bundle, but the native binaries of the core runtime itself are separate files.
+Specifying `IncludeAllContentForSelfExtract` extracts all files, including the managed assemblies, before running the executable. This may be helpful for rare application compatibility problems.
 
-To embed those files for extraction and get one output file, like in .NET Core 3.x, set the property `IncludeNativeLibrariesForSelfExtract` to `true`. For more information about extraction, see [Including native libraries](#include-native-libraries).
+ [!NOTE]
```
IEvangelist, line 165:
```suggestion
 [!IMPORTANT]
```

When we have two alerts back-to-back, it's best to vary their styling to help differentiate between them, if they cannot be otherwise avoided.

docs/core/deploying/single-file/overview.md
```diff
@@ -158,13 +154,33 @@ Single file applications have all related PDB files alongside the application, n
...
+> If extraction is used, the files are extracted to disk before the app starts:
+>
+> - If environment variable `DOTNET_BUNDLE_EXTRACT_BASE_DIR` is set to a path, the files are extracted to a directory under that path.
```
IEvangelist, line 168:
```suggestion
> - If the `DOTNET_BUNDLE_EXTRACT_BASE_DIR` environment variable is set to a path, the files are extracted to a directory under that path.
```

docs/core/deploying/single-file/overview.md
```diff
@@ -158,13 +154,33 @@ Single file applications have all related PDB files alongside the application, n
...
+> - Otherwise, if running on Linux or MacOS, the files are extracted to a directory under `$HOME/.net`.
```
IEvangelist, line 169:
```suggestion
> - Otherwise, if running on Linux or macOS, the files are extracted to a directory under `$HOME/.net`.
```

docs/core/deploying/single-file/overview.md
```diff
@@ -158,13 +154,33 @@ Single file applications have all related PDB files alongside the application, n
...
+> - If running on Windows, the files are extracted to a directory under `%TEMP%/.net`.
+>
+> To prevent tampering, these directories should not be writable by users or services with different privileges. Don't use _/tmp_ or _/var/tmp_ on most Linux and MacOS systems.
```
IEvangelist, line 172:
```suggestion
> To prevent tampering, these directories shouldn't be writable by users or services with different privileges. Don't use _/tmp_ or _/var/tmp_ on most Linux and macOS systems.
```

docs/core/deploying/single-file/overview.md
```diff
@@ -158,13 +154,33 @@ Single file applications have all related PDB files alongside the application, n
...
+
+> [!NOTE]
+> In some Linux environments, such as under `systemd`, the default extraction does not work because `$HOME` is not defined. In such cases, we recommend that you set `$DOTNET_BUNDLE_EXTRACT_BASE_DIR` explicitly.
```
IEvangelist, line 175:
```suggestion
> In some Linux environments, such as under `systemd`, the default extraction doesn't work because `$HOME` isn't defined. In such cases, it's recommended that you set `$DOTNET_BUNDLE_EXTRACT_BASE_DIR` explicitly.
```

docs/core/deploying/single-file/overview.md
```diff
@@ -158,13 +154,33 @@ Single file applications have all related PDB files alongside the application, n
...
+>
+> For `systemd`, a good alternative seems to be defining `DOTNET_BUNDLE_EXTRACT_BASE_DIR` in your service's unit file as `%h/.net`, which `systemd` expands correctly to `$HOME/.net` for the account running the service.
```
IEvangelist, line 177:
```suggestion
> For `systemd`, a good alternative is to define `DOTNET_BUNDLE_EXTRACT_BASE_DIR` in your service's unit file as `%h/.net`, which `systemd` expands correctly to `$HOME/.net` for the account running the service.
```

docs/core/deploying/single-file/overview.md
```diff
@@ -193,48 +209,27 @@ We have some recommendations for fixing common scenarios:
...
+### Post-processing binaries before bundling
 
-Without this file, Visual Studio might produce the error: "Unable to attach to the process. A debug component is not installed." VS Code might produce the error: "Failed to attach to process: Unknown Error: 0x80131c3c."
+Some workflows require post-processing of binaries before bundling. A common example is signing. The dotnet SDK provides MSBuild extension points to allow processing binaries just before single-file bundling. The available APIs are:
 
-To fix these errors, _mscordbi_ needs to be copied next to the executable. _mscordbi_ is `publish`ed by default in the subdirectory with the application's runtime ID. So, for example, if you publish a self-contained single file executable using the `dotnet` CLI for Windows using the parameters `-r win-x64`, the executable would be placed in _bin/Debug/net5.0/win-x64/publish_. A copy of _mscordbi.dll_ would be present in _bin/Debug/net5.0/win-x64_.
+* A target `PrepareForBundle` that will be called before `GenerateSingleFileBundle`
```
IEvangelist, line 216:
```suggestion
- A target `PrepareForBundle` that will be called before `GenerateSingleFileBundle`
```

docs/core/deploying/single-file/overview.md
```diff
@@ -193,48 +209,27 @@ We have some recommendations for fixing common scenarios:
...
+* An ItemGroup `FilesToBundle` containing all files that will be bundled
+* A Property `AppHostFile` that will specify the apphost template. Post-processing might want to exclude the apphost from processing.
 
-### Include native libraries
+The way to plug into this would typically involve creating a target that will be executed between PrepareForBundle and GenerateSingleFileBundle
 
-Single file deployment doesn't bundle native libraries by default. On Linux, the runtime is prelinked into the bundle and only application native libraries are deployed to the same directory as the single file app. On Windows, only the hosting code is prelinked and both the runtime and application native libraries are deployed to the same directory as the single file app. This approach is to ensure a good debugging experience, which requires native files to be excluded from the single file.
+Example:
```
IEvangelist, line 222:
```suggestion
Consider the following .NET project `Target` node example:
```

docs/core/deploying/single-file/overview.md
```diff
@@ -193,48 +209,27 @@ We have some recommendations for fixing common scenarios:
...
-> ```text
-> [Service]
-> Environment="DOTNET_BUNDLE_EXTRACT_BASE_DIR=%h/.net"
-> ```
+It is possible that tooling will need to copy files in the process of signing. That could happen if the original file is a shared item not owned by the build, for example the file comes from a nuget cache. In such case it is expected that the tool will modify the path of the corresponding FilesToBundle item to point to the modified copy.
 
 ### Compress assemblies in single-file apps
 
-Starting with .NET 6, single file apps can be created with compression enabled on the embedded assemblies. Set the `EnableCompressionInSingleFile` property to `true`. The single file that's produced will have all of the embedded assemblies compressed, which can significantly reduce the size of the executable.
+Single file apps can be created with compression enabled on the embedded assemblies. Set the `EnableCompressionInSingleFile` property to `true`. The single file that's produced will have all of the embedded assemblies compressed, which can significantly reduce the size of the executable.
```
IEvangelist, line 232:
```suggestion
Single-file apps can be created with compression enabled on the embedded assemblies. Set the `EnableCompressionInSingleFile` property to `true`. The single file that's produced will have all of the embedded assemblies compressed, which can significantly reduce the size of the executable.
```

docs/core/deploying/single-file/overview.md
```diff
@@ -158,13 +154,33 @@ Single file applications have all related PDB files alongside the application, n
 
 Managed C++ components aren't well suited for single file deployment. We recommend that you write applications in C# or another non-managed C++ language to be single file compatible.
 
-### Output differences from .NET 3.x
+### Native libraries
+
+Only managed DLLs are bundled with the app into a single executable. When the app starts, the managed DLLs are extracted and loaded in memory, avoiding the extraction to a folder. This approach means that the managed binaries are embedded in the single file bundle, but the native binaries of the core runtime itself are separate files.
 
-In .NET Core 3.x, publishing as a single file produced one file, consisting of the app itself, dependencies, and any other files in the folder during publish. When the app starts, the single file app was extracted to a folder and run from there.
+To embed those files for extraction and get one output file, set the property `IncludeNativeLibrariesForSelfExtract` to `true`. For more information about extraction, see [Including native libraries](#include-native-libraries).
```
IEvangelist, line 161:
```suggestion
To embed those files for extraction and get one output file, set the property `IncludeNativeLibrariesForSelfExtract` to `true`.
```

I'm suggesting that we remove:

>
> ```markdown
> For more information about extraction, see [Including native libraries](#include-native-libraries).
> 

As there is no heading that this links to. Either add the appropriate heading or remove this.

docs/core/deploying/single-file/overview.md
```diff
@@ -158,13 +154,33 @@ Single file applications have all related PDB files alongside the application, n
 
 Managed C++ components aren't well suited for single file deployment. We recommend that you write applications in C# or another non-managed C++ language to be single file compatible.
 
-### Output differences from .NET 3.x
+### Native libraries
+
+Only managed DLLs are bundled with the app into a single executable. When the app starts, the managed DLLs are extracted and loaded in memory, avoiding the extraction to a folder. This approach means that the managed binaries are embedded in the single file bundle, but the native binaries of the core runtime itself are separate files.
```
IEvangelist, line 159:
```suggestion
Only managed DLLs are bundled with the app into a single executable. When the app starts, the managed DLLs are extracted and loaded in memory, avoiding the extraction to a folder. With this approach, the managed binaries are embedded in the single file bundle, but the native binaries of the core runtime itself are separate files.
```

IEvangelist (approved):

docs/core/deploying/single-file/overview.md
```diff
@@ -158,13 +154,33 @@ Single file applications have all related PDB files alongside the application, n
...
 
-Starting with .NET 5, only managed DLLs are bundled with the app into a single executable. When the app starts, the managed DLLs are extracted and loaded in memory, avoiding the extraction to a folder. On Windows, this approach means that the managed binaries are embedded in the single file bundle, but the native binaries of the core runtime itself are separate files.
+To embed those files for extraction and get one output file, set the property `IncludeNativeLibrariesForSelfExtract` to `true`.
 
-To embed those files for extraction and get one output file, like in .NET Core 3.x, set the property `IncludeNativeLibrariesForSelfExtract` to `true`. For more information about extraction, see [Including native libraries](#include-native-libraries).
+Specifying `IncludeAllContentForSelfExtract` extracts all files, including the managed assemblies, before running the executable. This may be helpful for rare application compatibility problems.
+
+ [!IMPORTANT]
```
IEvangelist, line 165:
```suggestion
> [!IMPORTANT]
```

PR #34482, opened 2023-03-09
Added .sln files

IEvangelist:
## Summary

- Remove `*.sln` from the _.gitignore_
- Add local solution files, that were otherwise ignored


BillWagner (approved):
I'm good with this.

I'd like to make sure others on the team weigh in before we merge though.

tdykstra:
You can include a .sln file with --force -- what's the motivation for making it included by default?

gewarren:
I'd rather not allow .sln files by default. I think more people are going to add one by default when it's not needed (i.e. there's only one project), and hopefully Snippets 5000 will catch when a .sln file is missing but needed.

Also, are all the .sln files here necessary (i.e. there's more than one project)?

IEvangelist:
I have updated the PR based on everyone's feedback thus far.

gewarren (approved):

PR #34490, opened 2023-03-09
Update check-for-build-warnings.yml

IEvangelist:


.github/workflows/check-for-build-warnings.yml
```diff
@@ -10,6 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
         statuses: write
+        contents: write
```
IEvangelist, line 13:
```suggestion
        contents: write
        issues: write
```

BillWagner (approved):
:shipit:

Let's get it working, then if possible, let's dial down the `contents: write` as an experiment.

PR #34497, opened 2023-03-10
RID-specific app breaking change

### LINKED ISSUE #33726: [Breaking change]: RuntimeSpecific Apps in .NET Will No Longer Be `SelfContained` by default, and instead Framework Dependent By Default
### Description

RuntimeSpecific Apps (.NET Apps with a `RuntimeIdentifier`) Will No Longer Be `SelfContained` by default, and instead Framework Dependent By Default.
https://github.com/dotnet/sdk/pull/30038

cc @richlander.

### Version

.NET 8 Preview 5

### Previous behavior

Previously, if `RuntimeIdentifier` (RID) was specified, the app would be `SelfContained` (SC) if `SelfContained` was not explicitly specified. 

### New behavior

Now, for apps with `TargetFramework` 8.0 or higher, `RuntimeIdentifier` no longer implies `SelfContained` by default. Instead, apps that give a runtime identifier will be dependent on the .NET Framework by default (FDD). Apps for .NET 7 or below will not see this change in behavior.

#

This is a break for any customer that deployed, distributed, published, etc, their app, and desired their app to not require the .NET Framework to be installed on the machine for it to work, but did not explicitly add `SelfContained` to get this behavior. They may have relied on the old behavior to produce a .NET Framework INDEPENDENT app by default. 

This is a break for anyone relying on the IL Link tool. Users will have to take the steps described below to use IL Link again.
Note: Some `Publish` properties like `PublishTrimmed`, and `PublishSingleFile`, `PublishAot`, _currently_ require `SelfContained` to work and using these properties will require `SelfContained` to be added now. 

This is a break for Blazor WASM apps as they relied on this behavior. We expect the Blazor WASM team to side-step this breaking change in their SDK by adding `SelfContained` automatically for all customers, so Blazor customers should be unaffected.

### Type of breaking change

- [x] **Binary incompatible**: Existing binaries may encounter a breaking change in behavior, such as failure to load or execute, and if so, require recompilation.
- [x] **Source incompatible**: When recompiled using the new SDK or component or to target the new runtime, existing source code may require source changes to compile successfully.

### Reason for change

Previously we warned of this breaking change with the following warning in .NET 6:

```
warning NETSDK1179: One of '--self-contained' or '--no-self-contained' options are required when '--runtime' is used. [/app/app.csproj]
```

We have given customers time to add `SelfContained` explicitly.

As for why the change is being made:

1. This is currently what Visual Studio does by default.  So it aligns the dotnet SDK with VS. 
2. It will save space to have apps be framework dependent by default as  there won't be copies of .NET stored in each app.
3. Command line options should be ideally orthogonal. In this case the tooling supports both RuntimeIdentifier specific SelfContained Deployment and RID specific Framework Dependent Deployments. So it's really weird that no-RID -> defaults to FDD and RID -> defaults to SCD. It has made even internal people at MSFT think that you cant have a Non SelfContained app (FDD) if a RID is given, which is NOT true.

### Recommended action

Customers using .NET 7 and below who relied on the old behavior where `SelfContained` was inferred will see this warning:

```
For projects with TargetFrameworks >= 8.0, RuntimeIdentifier no longer automatically gives a SelfContained app. To continue creating a .NET framework independent app after upgrading to 8.0, consider setting SelfContained explicitly.
```

They should follow the guidance of the warning if they want to continue to produce `SelfContained` apps. If they would like to move to the new default, they can set `SelfContained` as `false`, in the project file, like 
```xml
<SelfContained>false</SelfContained
```

or in the call to `dotnet,` like `dotnet publish --no-self-contained`. 

For .NET 8, nothing needs to be specified unless the old default is desired, in which case the customer must add `SelfContained` as `true`. 

### Feature area

SDK

### Affected APIs

_No response_

---
[Associated WorkItem - 67115](https://dev.azure.com/msft-skilling/Content/_workitems/edit/67115)

gewarren:
Fixes #33726 

#### Internal previews

|:--|:--|
| _docs/core/compatibility/8.0.md_ | [Preview: docs/core/compatibility/8.0](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/8.0?branch=pr-en-us-34497) |
| _docs/core/compatibility/sdk/8.0/runtimespecific-app-default.md_ | [Preview: docs/core/compatibility/sdk/8.0/runtimespecific-app-default](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/runtimespecific-app-default?branch=pr-en-us-34497) |

### DIFF
--- a/docs/core/compatibility/sdk/8.0/runtimespecific-app-default.md
+++ b/docs/core/compatibility/sdk/8.0/runtimespecific-app-default.md
@@ -1,0 +1,63 @@
+---
+title: "Breaking change: Runtime-specific apps no longer self-contained"
+description: Learn about a breaking change in the .NET 8 SDK where apps that specify a runtime identifier are no longer self-contained by default.
+ms.date: 03/09/2023
+---
+# Runtime-specific apps no longer self-contained
+
+Runtime-specific apps, or .NET apps with a `RuntimeIdentifier`, are no longer [self-contained](../../../deploying/index.md#publish-self-contained) by default, and instead [framework-dependent](../../../deploying/index.md#publish-framework-dependent) by default.
+
+This is a break in the following situations:
+
+- If you deployed, distributed, or published your app and didn't explicitly add the `SelfContained` property, but also didn't require that the .NET runtime be installed on the machine for it to work. In this case, you may have relied on the previous behavior to produce a non-framework-dependent app by default.
+
+- If you relied on the IL Link tool. In this case, you'll have to take the steps described under [Recommended action](#recommended-action) to use IL Link again.
+
+  > [!NOTE]
+  > Some publish properties, like `PublishTrimmed`, `PublishSingleFile`, and `PublishAot`, currently require `SelfContained` to work. If you use these properties, you'll need to add the `SelfContained` property.
+
+- For Blazor WASM apps, because they relied on the previous behavior. However, the Blazor WASM team may side-step this breaking change in their SDK by adding `SelfContained` automatically for all apps, so Blazor customers shouldn't be affected.
+
+## Previous behavior
+
+Previously, if a runtime identifier (RID) was specified (via [RuntimeIdentifier](../../../project-sdk/msbuild-props.md#runtimeidentifier)), the app was published as self-contained, even if `SelfContained` wasn't explicitly specified.
+
+## New behavior
+
+Starting in .NET 8, for apps that target .NET 8 or a later version, `RuntimeIdentifier` no longer implies `SelfContained` by default. Instead, apps that specify a runtime identifier will be dependent on the .NET runtime by default (framework-dependent). Apps that target .NET 7 or earlier versions aren't affected.
+
+## Version introduced
+
+.NET 8 Preview 2
+
+## Type of breaking change
+
+This change can affect [source compatibility](../../categories.md#source-compatibility) and [binary compatibility](../../categories.md#binary-compatibility).
+
+## Reason for change
+
+.NET 6 alerted users to this breaking change with the following warning:
+
+**warning NETSDK1179: One of '--self-contained' or '--no-self-contained' options are required when '--runtime' is used.**
+
+Now that customers have had time to add `SelfContained` explicitly, we felt we could introduce the break.
+
+As for why the change is being made:
+
+- The new .NET SDK behavior aligns with Visual Studio behavior.
+- Framework-dependent apps are now smaller by default, since there aren't copies of .NET stored in each app.
+- Ideally, command-line options are orthogonal. In this case, the tooling supports both RID-specific self-contained deployment (SCD) and RID-specific framework-dependent deployment (FDD). So it didn't make sense that no RID defaulted to FDD and RID defaulted to SCD. This behavior was often confusing for users.
+
+## Recommended action
+
+If you're using .NET 7 or an earlier version and relied on the previous behavior where `SelfContained` was inferred, you'll see this warning:
+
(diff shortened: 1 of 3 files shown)


docs/core/compatibility/sdk/8.0/runtimespecific-app-default.md
```diff
@@ -0,0 +1,61 @@
+---
+title: "Breaking change: Runtime-specific apps no longer self-contained"
+description: Learn about a breaking change in the .NET 8 SDK where apps that specify a runtime identifier are no longer self-contained by default.
+ms.date: 03/09/2023
+---
+# Runtime-specific apps no longer self-contained
+
+Runtime-specific apps, or .NET apps with a `RuntimeIdentifier`, are no longer [self-contained](../../../deploying/index.md#publish-self-contained) by default. Instead, they are [framework-dependent](../../../deploying/index.md#publish-framework-dependent) by default.
+
+This is a break in the following situations:
```
IEvangelist, line 10:
```suggestion
This is a breaking change in the following situations:
```

docs/core/compatibility/sdk/8.0/runtimespecific-app-default.md
```diff
@@ -0,0 +1,61 @@
...
+
+- If you deployed, distributed, or published your app and didn't explicitly add the `SelfContained` property, but also didn't require that the .NET runtime be installed on the machine for it to work. In this case, you may have relied on the previous behavior to produce a non-framework-dependent app by default.
```
IEvangelist, line 12:
```suggestion
- If you've deployed, distributed, or published your app and didn't explicitly add the `SelfContained` property, but also didn't require that the .NET runtime be installed on the machine for it to work. In this case, you may have relied on the previous behavior to produce a non-framework-dependent app by default.
```

docs/core/compatibility/sdk/8.0/runtimespecific-app-default.md
```diff
@@ -0,0 +1,61 @@
...
+
+- If you relied on the IL Link tool. In this case, you'll have to take the steps described under [Recommended action](#recommended-action) to use IL Link again.
```
IEvangelist, line 14:
```suggestion
- If you've relied on the IL Link tool. In this case, you'll have to take the steps described under [Recommended action](#recommended-action) to use IL Link again.
```

docs/core/compatibility/sdk/8.0/runtimespecific-app-default.md
```diff
@@ -0,0 +1,61 @@
...
+
+  > [!NOTE]
+  > Some publish properties, like `PublishTrimmed`, `PublishSingleFile`, and `PublishAot`, currently require `SelfContained` to work. If you use these properties, you'll need to add the `SelfContained` property.
+
+- For Blazor WASM apps, because they relied on the previous behavior. However, the Blazor WASM team may side-step this breaking change in their SDK by adding `SelfContained` automatically for all apps, so Blazor customers shouldn't be affected.
```
IEvangelist, line 19:
```suggestion
- For Blazor WebAssembly apps, because they relied on the previous behavior. However, the Blazor WASM team may side-step this breaking change in their SDK by adding `SelfContained` automatically for all apps, so Blazor customers shouldn't be affected.
```

WASM is incorrect, it's either Wasm or WebAssembly. Source: https://webassembly.org/

IEvangelist (approved):
A few nits, otherwise it looks great.

docs/core/compatibility/sdk/8.0/runtimespecific-app-default.md
gewarren:
```suggestion
- If you rely on the IL Link tool. In this case, you'll have to take the steps described under [Recommended action](#recommended-action) to use IL Link again.
```

gewarren:
Thanks @IEvangelist. Love the preview table too!!

PR #34539, opened 2023-03-13
A couple more updates

gewarren:
Follow up for #33726.

#### Internal previews

|:--|:--|
| _docs/core/compatibility/sdk/8.0/runtimespecific-app-default.md_ | [Preview: docs/core/compatibility/sdk/8.0/runtimespecific-app-default](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/runtimespecific-app-default?branch=pr-en-us-34539) |

### DIFF
--- a/docs/core/compatibility/sdk/8.0/runtimespecific-app-default.md
+++ b/docs/core/compatibility/sdk/8.0/runtimespecific-app-default.md
@@ -16,7 +16,7 @@ This is a breaking change in the following situations:
   > [!NOTE]
   > Some publish properties, like `PublishTrimmed`, `PublishSingleFile`, and `PublishAot`, currently require `SelfContained` to work. If you use these properties, you'll need to add the `SelfContained` property.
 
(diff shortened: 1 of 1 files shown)


gewarren:
@nagilson I meant to tag you in the original PR. Does the preview look okay to you?

nagilson:
@gewarren 

Ah shoot, thanks for letting me know. The article looks good, BUT we ended up in a very circular discussion about the component of it that delayed it for preview 1, so while the work is done, it's not in preview 2, nor will it be in preview 3... And don't know when we will have a final decision on it. That is to say, we are still going to make the break, so it was not for naught to write the change doc. But:

The disagreement is about this part:
> Note: Some Publish properties like PublishTrimmed, and PublishSingleFile, PublishAot, currently require SelfContained to work and using these properties will require SelfContained to be added now.

So, we should change that ASAP as its not in preview 2. Maybe title it upcoming or hide it for now.

PR #34544, opened 2023-03-13
Update breaking-changes.md

BoldAsLove:
changed "almost any change in behavior could result breaking someone" to be "almost any change in behavior could cause a logic error for a consumer". 

## Summary
I think the author intended to convey the idea that a change to behavior in a library can cause breaking changes for the consumer of that library, that's why they wrote 'breaking someone'. However, that doesn't read well nor convey the idea well. It reads as though the 'someone' is the thing that is breaking, not their code. I prefer the change to read "almost any change in behavior could cause a logic error", leaving off the 'for a consumer', but will leave that on the change to keep it as close to what I believe was the authors intent.

#### Internal previews

|:--|:--|
| _docs/standard/library-guidance/breaking-changes.md_ | [Preview: docs/standard/library-guidance/breaking-changes](https://review.learn.microsoft.com/en-us/dotnet/standard/library-guidance/breaking-changes?branch=pr-en-us-34544) |

### DIFF
--- a/docs/standard/library-guidance/breaking-changes.md
+++ b/docs/standard/library-guidance/breaking-changes.md
@@ -44,7 +44,7 @@ Because a source breaking change is only harmful when developers recompile their
 
 ### Behavior breaking change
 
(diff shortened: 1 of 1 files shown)


BoldAsLove:
@dotnet-policy-service agree

IEvangelist (approved):

PR #34654, opened 2023-03-19
Verbatim reference: put string usage first

pkulikov:
I've almost always met `@` used with verbatim strings; I assume that's the general case. That's why putting verbatim strings as the first use case for `@`.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [docs/csharp/language-reference/tokens/verbatim.md](https://github.com/dotnet/docs/blob/7b4c7729b963e890e18511835b81a3397f47cea7/docs/csharp/language-reference/tokens/verbatim.md) | [docs/csharp/language-reference/tokens/verbatim](https://review.learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/verbatim?branch=pr-en-us-34654) |

### DIFF
--- a/docs/csharp/language-reference/tokens/verbatim.md
+++ b/docs/csharp/language-reference/tokens/verbatim.md
@@ -11,11 +11,7 @@ helpviewer_keywords:
 ---
 # Verbatim text - `@` in variables, attributes, and string literals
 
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks @pkulikov 

That makes sense to me. I'll :shipit: now.

PR #34711, opened 2023-03-23
.NET MAUI eBook - .NET 7 Messenger Migration

michaelstonis:
## Summary

This update replaces the MessagingCenter content with the MVVM Toolkit replacement IMessenger. Content has been updated to replace MessagingCenter.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [docs/architecture/maui/communicating-between-components.md](https://github.com/dotnet/docs/blob/311fa0287de90dcfefc0b1b3c432e92935fa7f3e/docs/architecture/maui/communicating-between-components.md) | [docs/architecture/maui/communicating-between-components](https://review.learn.microsoft.com/en-us/dotnet/architecture/maui/communicating-between-components?branch=pr-en-us-34711) |


docs/architecture/maui/communicating-between-components.md
```diff
@@ -1,115 +1,115 @@
...
+There are two implementations of the `IMessenger` interface that come with the `CommunityToolkit.Mvvm` package. The `WeakReferenceMessenger` uses weak references which can result in easier cleanup for message subscribers. This is a good option if your subscribers do not have a clearly defined lifecycle. The `StrongReferenceMessenger` uses strong references which can result in better performance and a more clearly controlled lifetime of the subscription. If you have a workflow with a very controlled lifetime (e.g. a subscription that is bound to a page's `OnAppearing` and `OnDisappearing` methods), the `StrongReferenceManager` may be a better option, if performance is a concern. Both of these implementations are available with default implementations ready to use by referencing either `WeakReferenceMessenger.Default` or `StrongReferenceMessenger.Default`.
```
IEvangelist, line 31:
```suggestion
There are two implementations of the `IMessenger` interface that come with the `CommunityToolkit.Mvvm` package. The `WeakReferenceMessenger` uses weak references which can result in easier cleanup for message subscribers. This is a good option if your subscribers do not have a clearly defined lifecycle. The `StrongReferenceMessenger` uses strong references which can result in better performance and a more clearly controlled lifetime of the subscription. If you have a workflow with a very controlled lifetime (for example, a subscription that is bound to a page's `OnAppearing` and `OnDisappearing` methods), the `StrongReferenceManager` may be a better option, if performance is a concern. Both of these implementations are available with default implementations ready to use by referencing either `WeakReferenceMessenger.Default` or `StrongReferenceMessenger.Default`.
```

docs/architecture/maui/communicating-between-components.md
```diff
@@ -1,115 +1,115 @@
...
-    // Add product to basket
-    public const string AddProduct = nameof(AddProduct);
+    public AddProductMessage(int count) : base(count)
+    {
+    }
 }
 ```
 
-In this example, messages are defined using constants. The advantage of this approach is that it provides compile-time type safety and refactoring support.
+The base class is defined using `ValueChangedMessage<T>` where T can be of any type needed to pass data. Both message publishers and subscribers can expect messages of a specific type (i.e. `AddProductMessage`). This can help ensure that both parties have agreed to a messaging contract and that the data provided with that contract will be consistent. Additionally, this approach provides compile-time type safety and refactoring support.
```
IEvangelist, line 60:
```suggestion
The base class is defined using `ValueChangedMessage<T>` where `T` can be of any type needed to pass data. Both message publishers and subscribers can expect messages of a specific type (for example, `AddProductMessage`). This can help ensure that both parties have agreed to a messaging contract and that the data provided with that contract will be consistent. Additionally, this approach provides compile-time type safety and refactoring support.
```

Per style guide, avoid things like `i.e.` or `e.g.`.

docs/architecture/maui/communicating-between-components.md
```diff
@@ -1,115 +1,115 @@
...
 
 ```csharp
-MessagingCenter.Unsubscribe<CatalogViewModel>(
-    this, MessengerKeys.AddProduct);
+WeakReferenceMessenger.Default.Unregister<Messages.AddProductMessage>(this);
 ```
 
-In this example, the Unsubscribe method syntax reflects the type arguments specified when subscribing to receive the AddProduct message.
+> [!NOTE]
+> In this example, it is not fully necessary to call `Unregister` as the `WeakReferenceMessenger` will allow unused objects to be garbage collected. If the `StrongReferenceMessenger` were used, it would be advised to call `Unregister` for any subscriptions that are no longer in use.
```
IEvangelist, line 109:
```suggestion
> In this example, it isn't fully necessary to call `Unregister` as the `WeakReferenceMessenger` will allow unused objects to be garbage collected. If the `StrongReferenceMessenger` were used, it would be advised to call `Unregister` for any subscriptions that are no longer in use.
```

IEvangelist (approved):
Looks good, just a few minor nits.

PR #34878, opened 2023-04-04
New console encoding breaking change

### LINKED ISSUE #34471: [Breaking change]: .NET SDK No Longer Changes Encoding Upon Exit
### Description

Also applies to 7.0.300. Work done: https://github.com/dotnet/sdk/pull/30963

Similar to https://github.com/dotnet/docs/issues/34250 but the bug mentioned where the SDK changed the encoding of an entire console has been fixed. Its possible users began to rely on the SDK changing the encoding of the command line console upon exit even for other programs ran on that console, but that is incorrect behavior.

The other change is that we no longer change the encoding as mentioned in https://github.com/dotnet/docs/issues/34250 for old builds of Windows that don't officially support it, and instead require opt-in.

### Version

.NET 8 Preview 3

### Previous behavior

* The SDK would change the encoding of a terminal after running a command such as `dotnet build`.
* The SDK would use the UTF 8 encoding to correctly render non-en characters, but it did so on versions of windows 10 that did not officially support UTF-8, so the behavior was undefined on those versions.

### New behavior

* The SDK does not change the encoding after exit for other programs.
* The SDK needs to be opted in to use UTF8 for windows versions which don't support it. See more below.

### Type of breaking change

- [X] **Binary incompatible**: Existing binaries may encounter a breaking change in behavior, such as failure to load or execute, and if so, require recompilation.

### Reason for change

First
A) We changed the encoding for older versions of windows 10 that had undefined behavior as those early builds didn't support utf-8 (builds before the Nov 2019 update.) Since it is not officially supported, we don't want this behavior by default and instead provide an opt-in.

B) We fixed https://github.com/dotnet/msbuild/issues/1596 which was an existing issue in .NET 7 where the SDK affected the encoding on the console for other programs. Running the .NET SDK should not impact the output of other programs.

### Recommended action

Applications that need to change the codepage can do so by running a process to invoke the `chcp` command on Windows. They shouldn't rely on the SDK to do so.

If you would like to keep the encoding change inside the SDK (but it still won't stay after the SDK is finished executing a command) for non-en languages in Windows 10 versions that don't officially support UTF-8, which is any build of before the Windows 10 (November 2019 update), you should set the environment variable `DOTNET_CLI_FORCE_UTF8_ENCODING`. This should also be mentioned here. https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables 

### Feature area

SDK

### Affected APIs

_No response_

---
[Associated WorkItem - 82741](https://dev.azure.com/msft-skilling/Content/_workitems/edit/82741)

gewarren:
Fixes #34471.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [docs/core/compatibility/7.0.md](https://github.com/dotnet/docs/blob/665fa155d73d94f60015841138c1f23a4fa37d27/docs/core/compatibility/7.0.md) | [docs/core/compatibility/7.0](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/7.0?branch=pr-en-us-34878) |
| [docs/core/compatibility/8.0.md](https://github.com/dotnet/docs/blob/665fa155d73d94f60015841138c1f23a4fa37d27/docs/core/compatibility/8.0.md) | [docs/core/compatibility/8.0](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/8.0?branch=pr-en-us-34878) |
| [docs/core/compatibility/sdk/8.0/console-encoding-fix.md](https://github.com/dotnet/docs/blob/665fa155d73d94f60015841138c1f23a4fa37d27/docs/core/compatibility/sdk/8.0/console-encoding-fix.md) | [docs/core/compatibility/sdk/8.0/console-encoding-fix](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/console-encoding-fix?branch=pr-en-us-34878) |
| [docs/core/compatibility/sdk/8.0/console-encoding.md](https://github.com/dotnet/docs/blob/665fa155d73d94f60015841138c1f23a4fa37d27/docs/core/compatibility/sdk/8.0/console-encoding.md) | [docs/core/compatibility/sdk/8.0/console-encoding](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/console-encoding?branch=pr-en-us-34878) |
| [docs/core/tools/dotnet-environment-variables.md](https://github.com/dotnet/docs/blob/665fa155d73d94f60015841138c1f23a4fa37d27/docs/core/tools/dotnet-environment-variables.md) | [docs/core/tools/dotnet-environment-variables](https://review.learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables?branch=pr-en-us-34878) |

### DIFF
--- a/docs/core/compatibility/7.0.md
+++ b/docs/core/compatibility/7.0.md
@@ -125,6 +125,7 @@ If you're migrating an app to .NET 7, the breaking changes listed here might aff
 | [Automatic RuntimeIdentifier for certain projects](sdk/7.0/automatic-runtimeidentifier.md) | ✔️ | ❌ | 7.0.100 |
 | [Automatic RuntimeIdentifier for publish only](sdk/7.0/automatic-rid-publish-only.md) | ❌ | ❌ | 7.0.200 |
 | [CLI console output uses UTF-8](sdk/8.0/console-encoding.md) | ❌ | ❌ | 7.0.3xx |
+| [Console encoding not UTF-8 after completion](sdk/8.0/console-encoding-fix.md) | ❌ | ✔️ | 7.0.300 |
 | [MSBuild serialization of custom types in .NET 7](sdk/7.0/custom-serialization.md) | ❌ | ❌ | 7.0.100 |
 | [Side-by-side SDK installations](sdk/7.0/side-by-side-install.md) | ❌ | ❌ | 7.0.100 |
 | [Version requirements for .NET 7 SDK](sdk/7.0/vs-msbuild-version.md) | ✔️ | ✔️ | 7.0.100 |
--- a/docs/core/compatibility/8.0.md
+++ b/docs/core/compatibility/8.0.md
@@ -60,6 +60,7 @@ If you're migrating an app to .NET 8, the breaking changes listed here might aff
 | Title                                                                           | Type of change                                   | Introduced |
 | ------------------------------------------------------------------------------- | ------------------------------------------------ | ---------- |
 | [CLI console output uses UTF-8](sdk/8.0/console-encoding.md)                    | Behavioral change/Source and binary incompatible | Preview 1  |
+| [Console encoding not UTF-8 after completion](sdk/8.0/console-encoding-fix.md)  | Behavioral change/Binary incompatible            | Preview 3  |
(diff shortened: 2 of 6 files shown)


nagilson:
LGTM

nagilson (approved):

PR #35007, opened 2023-04-14
Update 7.0.md with implicit Required for ASP.NET Core 7 model binding

johnkors:
## Summary

This hit us when going from ASP.NET Core 6 to 7.

Nullability: enabled

A model sent to an API controller  containing a `string` field was not treated as [Required] in ASP.NET Core 6, but was implicitly [Required] in 7.

Mark: The docs say this is already present in ASP.NET Core 6, but this only hit us after migrating to 7, so something does not add up.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [docs/core/compatibility/7.0.md](https://github.com/dotnet/docs/blob/98ab1ecb278e66bd9a693e84d057df35079dc73e/docs/core/compatibility/7.0.md) | [Breaking changes in .NET 7](https://review.learn.microsoft.com/en-us/dotnet/core/compatibility/7.0?branch=pr-en-us-35007) |

### DIFF
--- a/docs/core/compatibility/7.0.md
+++ b/docs/core/compatibility/7.0.md
@@ -34,6 +34,7 @@ If you're migrating an app to .NET 7, the breaking changes listed here might aff
 | [MVC's detection of an empty body in model binding changed](aspnet-core/7.0/mvc-empty-body-model-binding.md) | ❌ | ✔️ | Preview 3 |
 | [Output caching API changes](aspnet-core/7.0/output-caching-renames.md) | ❌ | ❌ | RC 2 |
 | [SignalR Hub methods try to resolve parameters from DI](aspnet-core/7.0/signalr-hub-method-parameters-di.md) | ✔️ | ❌ | Preview 2 |
+| [Model binding: Nullability enabled means implicitly Required fields]([aspnet-core/7.0/signalr-hub-method-parameters-di.md](https://learn.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-6.0#non-nullable-reference-types-and-the-required-attribute)) | ❓ | ❓ | No idea, plz halp |
 
 ## Core .NET libraries
 


gewarren:
Could it be that nullable context wasn't enabled when you were targeting .NET 6?

@Rick-Anderson or @brunolins16 If this model validation change is considered a breaking change, can you create a breaking change issue as per the usual workflow?

Rick-Anderson:
Must have been missing
`<Nullable>enable</Nullable>`

See [Non-nullable reference types and [Required] attribute](https://learn.microsoft.com/en-us/aspnet/core/mvc/models/validation?branch=pr-en-us-29000&view=aspnetcore-7.0#non-nullable-reference-types-and-required-attribute)

This changed in 6.0, not 7.0

johnkors:
Nope, I have a repro here: 

https://github.com/johnkors/aspnetcore7-compat-repro

Rick-Anderson:
@johnkors  Thanks for the awesome test project, we'll get right on this.

brunolins16:
> The following example controller also demonstrates the issue, which appears to be specific to the type being generic:
> 
> ```cs
> [ApiController]
> [Route("/")]
> public class HomeController : ControllerBase
> {
>     [HttpPost]
>     public IActionResult Post(TestModel<object> testModel)
>         => Ok(testModel);
> }
> 
> public class TestModel<T>
> {
>     public object Value { get; set; }
> }
> ```
> 
> The same example, without generics, is consistent between .NET 6 and .NET 7:
> 
> ```cs
> [ApiController]
> [Route("/")]
> public class HomeController : ControllerBase
> {
>     [HttpPost]
>     public IActionResult Post(TestModel testModel)
>         => Ok(testModel);
> }
> 
> public class TestModel
> {
>     public object Value { get; set; }
> }
> ```
> 
> From an initial investigation, I believe this may have changed in [dotnet/aspnetcore#39219](https://github.com/dotnet/aspnetcore/pull/39219). Specifically, the .NET 6 code has [this](https://github.com/dotnet/aspnetcore/pull/39219/files#diff-5d956f36da57378cce17b92a268e2cc418502fabbf709aa2c147b881da7fcf3dL523), which decides that the type isn't nullable, and therefore doesn't need an implicit `[Required]`:
> 
> ```cs
> // For generic types, inspecting the nullability requirement additionally requires
> // inspecting the nullability constraint on generic type parameters. This is fairly non-triviial
> // so we'll just avoid calculating it. Users should still be able to apply an explicit [Required]
> // attribute on these members.
> if (containingType.IsGenericType)
> {
>     return false;
> }
> ```
> 
> In .NET 7, the approach changes to use `NullabilityInfoContext`, which declares that the type _is_ nullable and adds the implicit `[Required]`. I believe that decision is made [here](https://github.com/dotnet/aspnetcore/blob/release/7.0/src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs#L451):
> 
> ```cs
> var isOptional = nullability != null && nullability.ReadState != NullabilityState.NotNull;
> ```
> 
> I was able to prove, in a local project, that `nullability.ReadState` _is_ set to `NullabilityState.NotNull`, which evaluates to the attribute being added.

@serpent5  you are totally right. We have updated the docs months ago.
https://github.com/dotnet/AspNetCore.Docs/issues/25840

johnkors:
Nice! [That note](https://github.com/dotnet/AspNetCore.Docs/issues/25840) is exactly what I didn't find from the "breaking changes" sections. If it could be included, it would be helpful for us & others migrating.

gewarren:
Closing this PR in favor of https://github.com/dotnet/AspNetCore.Docs/issues/29059.

